PR #1298, opened 2016-03-11
keyTyped() is preventing keyboard input

limzykenneth:
The search bar can't be used if the `return false;` wasn't commented out. Quick fix but should work.


lmccart:
thanks!

Issue #1314, opened 2016-03-22
text is not a function


SaintPepsi:
I might not be a pro at coding but i do know that i've used this function yesterday and i think around 14^14 times and nothing changed since yesterday. is it my fault or is this something that's wrong in p5?

full details:

draw	  @	sketch.js:71
p5.redraw	@	p5.js:16412
(anonymous function)	@	p5.js:11813
(anonymous function)	@	p5.js:11714
p5	@	p5.js:11976
_globalInit	@	p5.js:8482

the p5.js editor gives a different error:
72: Uncaught TypeError: string is not a function

The line of code that's being the problem:
text("test",100,200);

SaintPepsi:
the p5.js editor gives a different error:

72: Uncaught TypeError: string is not a function

SaintPepsi:
This is the code it's in:

`var population = [];
var populationSize = 10;
var initialChrLength = 10;
function coder(idx, dschrl) {
    this.x = 10;
    this.y = idx * 15 + 95; //index
    this.name = "coder" + idx;
    this.fitness = 0;
    this.chromosome = "";
    this.desiredChromosomeLength = dschrl;
    this.chromosomeBuilder = function (){
        var chrbuilt = "";
        var possible = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789 ,''$(){}";
        for( var i=0; i < this.desiredChromosomeLength; i++ )
            text += possible.charAt(Math.floor(Math.random() * possible.length));

        return chrbuilt;
    }
    this.fitnessTester = function () {
        
    };
    
}
function setPopulation() {
    for (var pop = 0; pop < populationSize; pop++) {
        population[pop] = new coder(pop, initialChrLength);
        population[pop].chromosomeBuilder();
    }
}

function setup() {
    createCanvas(900,300);
    setPopulation();
}

function draw() {
    background(210,210,210);
    fill(0,0,0,100);
    textSize(15);
    for (var i = 0; i < population.length; i++) {
       text("test",100,200);
    }
}`

dhowe:
can you include the simplest bit of code that demonstrates the problem, and also list the version of p5.js you're using? thanks

SaintPepsi:
This is where the text function is:

```
function draw() {
    background(210,210,210);
    fill(0,0,0,100);
    textSize(15);
    for (var i = 0; i < population.length; i++) {
       text("test",100,200);
    }
}
```

[...]

PR #1318, opened 2016-03-23
Log helpful warnings if user overwrites p5 global functions.

### LINKED ISSUE #1317: Consider defining p5 methods as non-writable properties on window (in global mode)
I was just looking at #1314, where a user accidentally overrwrote p5's `text` function, and realized that one potential way to prevent this from happening is by defining p5's methods (in global mode) to be **non-writable** properties of the `window` object. Here's an example:

```js
Object.defineProperty(window, 'text', {
   value: function() {
     console.log("TEXT WAS CALLED");
   },
  writable: false
});

text += "blah";

text();
```

In the above example, the assignment to `text` will silently fail, and the subsequent calling of `text()` will succeed.

Two alternatives that I *think* are possible:

  1. Throw a helpful error if the value is overwritten, instead of failing silently.
  2. Permit overwriting the value, but log a helpful warning telling the user that they just overwrote a core p5 method.

I guess alternative (2) might be best if we don't want to break backwards compatibility... Or we can just leave the current behavior as-is, of course.

toolness:
This is an attempt to fix #1317 and prevent future occurrences of #1314 by allowing the overwriting of p5 globals, but logging helpful warning messages informing the user of what just happened.

### Example 1

```js
var text = "blarg";

function setup() {
}
```

This will cause the following message to be logged to the console:

```
p5 had problems creating the global function "text", possibly because your
code is already using that name as a variable. You may want to rename
your variable to something else.
```

### Example 2

```js
function setup() {
  text += "blarg";
}
```

This will cause the following message to be logged to the console:

```
You just changed the value of "text", which was a p5 function. This could cause
problems later if you're not careful.
```

### Limitations

* Modifying a global p5 function via [`delete`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete) or [`Object.defineProperty()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) won't log any warnings, but it's unlikely that users--especially novices who could really use the assistance--will be using these.
* Warnings are only logged for p5 global *functions*, so the user won't see anything if e.g. their code conflicts with `mouseX` or `mouseIsPressed`. We might want to tackle those in a separate PR if we want to address them, since I think the implementation will be a bit different and independent of the changes in this PR.
* There's a potential performance hit from the fact that this mechanism adds the overhead of a function call every time a global p5 function is accessed. It'd be nice if we could add a build-time conditional that only adds this extra functionality for non-minified/non-optimized builds, as per the friendly error system. ~~I know how to do this kind of thing with webpack but not browserify.~~ Update: this is actually a feature of UglifyJS, which I mention in https://github.com/processing/p5.js/pull/1318#issuecomment-202081479.
* ~~This PR doesn't currently have any unit tests. I'd like to add some!~~ We have unit tests now!

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -442,6 +442,56 @@ var p5 = function(sketch, node, sync) {
     }
   }, this);
 
+  var friendlyBindGlobal = function(prop, value) {
+    if (typeof(value) === 'function') {
+      try {
+        // Because p5 has so many common function names, it's likely
+        // that users may accidentally overwrite global p5 functions with
+        // their own variables. Let's allow this but log a warning to
+        // help users who may be doing this unintentionally.
+        //
+        // For more information, see:
+        //
+        //   https://github.com/processing/p5.js/issues/1317
+        Object.defineProperty(window, prop, {
+          configurable: true,
+          enumerable: true,
+          get: function() {
+            return value;
+          },
+          set: function(newValue) {
+            Object.defineProperty(window, prop, {
+              configurable: true,
+              enumerable: true,
+              value: newValue,
+              writable: true
+            });
+            console.log(
+              'You just changed the value of "' + prop + '", which was ' +
+              'a p5 function. This could cause problems later if you\'re ' +
+              'not careful.'
+            );
+          }
+        });
+      } catch (e) {
+        // It's likely that this is a TypeError due to us trying to
+        // redefine a non-configurable property, which can happen if
+        // the user already declared the property name as a global
+        // variable. Let's log a warning and fall back to our legacy
+        // behavior.
+        console.log(
+          'p5 had problems creating the global function "' + prop + '", ' +
+          'possibly because your code is already using that name as ' +
+          'a variable. You may want to rename your variable to something ' +
+          'else.'
+        );
+        window[prop] = value;
+      }
+    } else {
+      window[prop] = value;
+    }
+  };
+
   // If the user has created a global setup or draw function,
   // assume "global" mode and make everything global (i.e. on the window)
   if (!sketch) {
@@ -451,16 +501,16 @@ var p5 = function(sketch, node, sync) {
       if(typeof p5.prototype[p] === 'function') {
         var ev = p.substring(2);
         if (!this._events.hasOwnProperty(ev)) {
-          window[p] = p5.prototype[p].bind(this);
+          friendlyBindGlobal(p, p5.prototype[p].bind(this));
         }
       } else {
-        window[p] = p5.prototype[p];
+        friendlyBindGlobal(p, p5.prototype[p]);
       }
     }
     // Attach its properties to the window
     for (var p2 in this) {
       if (this.hasOwnProperty(p2)) {
-        window[p2] = this[p2];
+        friendlyBindGlobal(p2, this[p2]);
       }
     }
 
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -556,7 +556,7 @@ p5.prototype._createFriendlyGlobalFunctionBinder = function(options) {
   };
 
   return function(prop, value) {
-    if (typeof(value) === 'function') {
+    if (typeof(IS_MINIFIED) === 'undefined' && typeof(value) === 'function') {
       try {
         // Because p5 has so many common function names, it's likely
         // that users may accidentally overwrite global p5 functions with
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -442,55 +442,7 @@ var p5 = function(sketch, node, sync) {
     }
   }, this);
 
-  var friendlyBindGlobal = function(prop, value) {
-    if (typeof(value) === 'function') {
-      try {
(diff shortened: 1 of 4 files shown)


toolness:
Ok @lmccart, I think this PR is ready for review!

src/core/core.js
```diff
@@ -540,4 +542,73 @@ p5.prototype.registerMethod = function(name, m) {
...
+        // that users may accidentally overwrite global p5 functions with
+        // their own variables. Let's allow this but log a warning to
+        // help users who may be doing this unintentionally.
+        //
+        // For more information, see:
+        //
+        //   https://github.com/processing/p5.js/issues/1317
+
+        if (prop in globalObject && !(prop in propsToForciblyOverwrite)) {
+          throw new Error('global "' + prop + '" already exists');
```
lmccart: Can you explain a little more about what's happening here? It looks like the log call on line 587 will display the console.log warning, which is the behavior we want (as noted in line 563). So in which case would an error get thrown?

lmccart:
This looks great, one question inline.

Limitations 1 and 2 sound find for now, your notes make sense.

Regarding the potential performance hit... is it possible to do something similar like we were doing with the original friendly errors, where at build time in the minified version only, the argument checker function was overwritten so it wouldn't slow things down. 

I am imagining something like overwriting `p5.prototype._createFriendlyGlobalFunctionBinder`, though I guess we'd need to change/remove the assignment on [line 445](https://github.com/processing/p5.js/pull/1318/files#diff-dfe07c186488ee794e2200e24b2bd54aR445).

src/core/core.js
toolness: Oh, good question... I actually ran into this problem before adding the explicit `throw` in lines 570-571, when I used the following sketch:

```
var text;

function setup() {}
```

Basically browsers threw an actual error when I was trying to call `Object.defineProperty`, claiming that I was trying to "redefine a non-configurable property". So I originally added the try/catch to detect for that, but I ended up later adding lines 570-571, which check for such a condition for us, so we might be able to get rid of the try/catch...

That said, because this functionality will *always* be called in global mode, and because it's sort of "nice to have" functionality rather than "must have" functionality, I thought it might be a good idea to leave the try/catch with the fallback to legacy behavior in there just in case there are really weird edge cases that we don't know about... I can remove it if you want though!

toolness:
Yep, overwriting `_createFriendlyGlobalFunctionBinder` is definitely one way to do it...

Do you know if p5's build system uses UglifyJS under the hood? If it does, we can actually use uglify's [Global Definitions](http://lisperator.net/uglifyjs/compress) functionality to conditionally define `_createFriendlyGlobalFunctionBinder` at build-time. This way the "friendly" version of the binder won't even take up extra space in the minified JS file!

lmccart:
indeed we are using uglify, that's how it worked in the previous version of friendly errors as well, I believe:
https://github.com/processing/p5.js/blob/master/Gruntfile.js#L253

src/core/core.js
lmccart: Makes sense. Would you mind just adding a little line note here so we remember what's going on and this doesn't cause confusion for others?

toolness:
Oh cool! Sure, I will add the conditional compilation bit and the comment.

toolness:
Ok, in 365cc78 I added a commit that only includes this fancy behavior in un-minified builds. Unlike with `_validateParameters` and `_friendlyFileLoadError`, however, this is not done by overwriting the functions to be no-ops in the uglify `footer` options.

Instead, it's done by adding an uglifyjs `global_defs` option that sets a global called `IS_MINIFIED` for minified builds. The implementation of `_createFriendlyGlobalFunctionBinder` checks for the existence of `IS_MINIFIED` in its implementation. The nice thing about this is that because uglifyjs knows that `IS_MINIFIED` has a constant value, it uses its dead code removal functionality to excise the unused code from the minified version entirely. This can be evidenced by running `grep 'not careful' lib/p5.min.js`--the string "not careful" is part of one of our warning messages--and verifying that no matches are found.

I recommend using this kind of feature for `_validateParameters` and `_friendlyFileLoadError`, too, since it will reduce the size of the minified build by eliminating the whole friendly debugger from it (which will be increasingly useful as the friendly debugger's functionality grows). I also like the approach because it makes it more obvious to readers that the functionality won't be available in minified builds. If you agree, I can file a bug for it.

**Note:** I originally wanted the variable to be called `DEBUG` instead of `IS_MINIFIED`, but one of the challenges of this was defaulting `DEBUG` to true; because p5's build system first generates `p5.js` and uses that as the source for `p5.min.js`, doing this proved difficult, so I opted to call the variable `IS_MINIFIED` and *only* define it in minified builds.

lmccart:
awesome! yes, I agree this approach would be preferable for `_validateParameters` and `_friendlyFileLoadError`, as well. this all looks great, going ahead with the merge.

PR #1321, opened 2016-03-24
Fix push/pop for text

### LINKED ISSUE #1320: push() and pop() do not properly save and restore text rendering state
Originally uncovered by molleindustria/p5.play#54

The [`p5.prototype.push()`](https://github.com/processing/p5.js/blob/a385d1ca7a1232b0e4d20066a395706ef88d0e7c/src/core/structure.js#L161) and [`p5.prototype.pop()`](https://github.com/processing/p5.js/blob/a385d1ca7a1232b0e4d20066a395706ef88d0e7c/src/core/structure.js#L229) methods are designed to "allow you to change the style and transformation settings and later return to what you had." They claim to store style information related to text rendering, including `fill` and `stroke` properties.

However, they fail to save/restore text fill and stroke properties correctly, [as illustrated in this p5.play pull request](https://github.com/molleindustria/p5.play/pull/55#issuecomment-200067742).  They miss the renderer's `_fillSet` and `_strokeSet` properties, which are set by [`fill()`](https://github.com/processing/p5.js/blob/f56df5deab1df1ae9d104ded71e47e1a8cca0ced/src/color/setting.js#L373) and [`stroke()`](https://github.com/processing/p5.js/blob/f56df5deab1df1ae9d104ded71e47e1a8cca0ced/src/color/setting.js#L541) and [influence text rendering](https://github.com/processing/p5.js/blob/20289b05e7df453c4e809320c7356b0f5e33f0bb/src/core/p5.Renderer2D.js#L1143) but don't seem to affect anything else (searching p5.js for [_fillSet](https://github.com/processing/p5.js/search?utf8=%E2%9C%93&q=_fillSet) or [_strokeSet](https://github.com/processing/p5.js/search?utf8=%E2%9C%93&q=_strokeSet)).

You can see the resulting behavior [in this JSBin](https://jsbin.com/bujikudapu/edit?html,output): The text renders normally (with a default style) but after you click the mouse it changes to rendering with a stroke and no fill, which is clearly not what the code specifies.

It should be a fairly easy fix to have `push()` track `_fillSet` and `_strokeSet` (I've tested it locally myself) but I don't know the original intent of these properties or whether this would be considered a breaking change.  Can someone please take a look and let me know if this seems like a safe fix to make?

Thank you!
Brad

islemaster:
Fixes #1320 

Make `p5.prototype.push()` track the `_strokeSet` and `_fillSet` properties which can impact text rendering.  Adds a simple test to show that these properties are now saved and restored to their original state by push/pop.

### DIFF
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -162,7 +162,9 @@ p5.prototype.push = function () {
   this._renderer.push();
   this._styles.push({
     _doStroke: this._renderer._doStroke,
+    _strokeSet: this._renderer._strokeSet,
     _doFill: this._renderer._doFill,
+    _fillSet: this._renderer._fillSet,
     _tint: this._renderer._tint,
     _imageMode: this._renderer._imageMode,
     _rectMode: this._renderer._rectMode,
--- a/test/test-minified.html
+++ b/test/test-minified.html
@@ -36,6 +36,7 @@
   <script src="unit/core/core.js" type="text/javascript" ></script>
   <!--<script src="unit/core/2d_primitives.js" type="text/javascript" ></script>-->
   <script src="unit/core/curves.js" type="text/javascript" ></script>
+  <script src="unit/core/structure.js" type="text/javascript" ></script>
 
   <script src="unit/math/calculation.js" type="text/javascript" ></script>
   <script src="unit/math/random.js" type="text/javascript" ></script>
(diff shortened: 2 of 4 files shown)


islemaster:
Note: I've realized I forgot to add my new test file to test.html, so they don't actually run - and when I did add them they don't work yet.  Fixing now...

islemaster:
That's better!  The tests actually run now :fearful: and I replaced the very invasive/targeted tests I had with a more thorough approach that does a deep comparison of the render state before and after using push/pop to undo a particular operation.  I added a test for every operation that's supposedly supported, according to the documentation on `push()` (except for `textMode` which apparently isn't a real thing?).

lmccart:
right on, thank you!

PR #1326, opened 2016-03-26
trim() parameter not optional

limzykenneth:
Tiny fix to [`trim()`](http://p5js.org/reference/#/p5/trim)'s documentation.


lmccart:
good catch, thanks!

PR #1330, opened 2016-03-29
Improve documentation for random()

toolness:
This is an attempt to make some minor improvements to `random()`:

* Mention the fact that the lower bound is *inclusive* while the upper bound is *exclusive*. At least, um, that's how I *think* `random()` is working--someone may want to double-check, but it appears to be the way this works, especially given the third example in the docs. It's also how most random functions I know of work (e.g. in [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random), [Python](https://docs.python.org/2/library/random.html#random.random), and [Ruby](http://ruby-doc.org/core-2.2.0/Random.html#method-i-rand)).
* Mark both arguments as optional.
* Mention that the return value is always a floating-point number. I'm not sure if "floating-point" might confuse beginners, but I couldn't think of a better word, and I thought it was important to note. @iamjessklein and I were actually thinking it might be nice to have a glossary in the docs, which terms like this can directly link to?

### DIFF
--- a/src/math/random.js
+++ b/src/math/random.js
@@ -72,13 +72,13 @@ p5.prototype.randomSeed = function(seed) {
  * Return a random floating-point number.
  *
  * Takes either 0, 1 or 2 arguments.
- * 
+ *
  * If no argument is given, returns a random number from 0
  * up to (but not including) 1.
- * 
+ *
  * If one argument is given, returns a random number from 0 up to
  * (but not including) the number.
- * 
+ *
  * If two arguments are given, returns a random number from the
  * first argument up to (but not including) the second argument.
  *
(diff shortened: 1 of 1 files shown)


limzykenneth:
For floating point number, I think [this video](https://www.youtube.com/watch?v=PZRI1IfStY0) explains it perfectly especially for `0.1+0.2=0.30000000000000004` which could be a pit fall when it comes to floating point number.

toolness:
Ah yeah, that would be a great video to link-to from a glossary!

lmccart:
Thanks! A glossary is a great idea.
Note to self, or anyone: this is a method that could really benefit from the multiple signature documentation being added.

PR #1335, opened 2016-03-31
Expose the global p5 instance as `p5.instance` on `window`

yangsu:
This allows the p5 instance created in global mode to be accessed and manipulated outside of the sketch code.

See use case [here](https://github.com/yangsu/p5-live/blob/master/main.js)


mactkg:
LGTM.

I'm developing application that use p5.js such as IDE, editor. Sometimes I want to access p5 instance that is treated as 'global mode'.

ping @lmccart

lmccart:
makes sense, thanks for the ping and for the PR @yangsu

PR #1342, opened 2016-04-01
Delete docs/progress.txt

toolness:
I noticed that this file is 2 years old so it's probably out of date? Should we just remove it from the repository? Or would it be better to somehow bring it up-to-date?


lmccart:
yes very out of date, good call, let's remove!

PR #1295, opened 2016-03-07
Document overloaded methods nicely

toolness:
**This is a [work-in-progress pull request](http://vrybas.github.io/blog/2014/04/11/wip-pull-requests/), please don't merge it yet.**

This is ~~an experimental proof-of-concept to see how feasible it may be~~ a PR to document overloaded methods in a way similar to [Java Processing's documentation](https://processing.org/reference/color_.html), as per @lmccart's explanation in https://github.com/processing/p5.js/pull/1287#issuecomment-192735393.

Here's what the new [`p5.color()`](http://p5js.org/reference/#/p5/color) docs look like in this PR: 

overloads

The good news is that YUIDoc already has a mechanism for recording information about overloaded methods: if you define a method *twice* with different params, it just adds both versions to `data.json`. We just need to add logic to our view layer to ensure that our index page doesn't contain a separate link for each overloaded version, and that the item detail view properly lists all the different method signatures and all the params. That's what this PR attempts to do.

Limitations:

* This PR makes things work with overloaded *methods*, but it may not work with overloaded *constructors* or other kinds of callables.
* Ideally I think that all the optional params would be listed after all the required params. Shouldn't be hard to fix.
* Unlike the [original Processing docs for `color()`](https://processing.org/reference/color_.html), I documented `[alpha]` as an optional param, which resulted in only two different signatures rather than four. There's no reason we can't document it the way Processing does--I was being a bit lazy, I guess. :grin:
* ~~A lot of logic for the overloading is done during render time, rather than during build time. It'll be better for QA to raise errors and such at build time instead. I'm thinking that we could do some kind of transformation to `data.json`, maybe, so that the item view just displays information rather than doing a bunch of the logic I've added in this PR. This will also probably make the information easier for the [experimental type-checker](https://github.com/processing/p5.js/pull/1287) and friendly error system to use, too.~~
* I'm not really sure how much I like the whole "just document multiple versions of the same method in separate comment blocks" approach of YUIDoc. As can be seen from the example of `p5.color()` that I've done for this PR, it reads a bit weird to have a comment block with just one version of the method's params and tons of examples, and then have little overloaded versions documented in separate comment blocks right after it. We might be able to experiment with this. But another alternative, if we *really* don't like this approach, may be to invent our own syntax for describing overloaded methods and bypass YUIDoc entirely.

Issue #1190, opened 2016-01-02
Documentation style guide and YUIDoc markdown


limzykenneth:
To build on the point about style guide for the inline documentation mentioned in #1173, at the moment the way to add line breaks or create new paragraphs is only by adding `<br />` tags. I think that's too messy and way too easy to miss.

According to YUIDoc's [syntax reference](http://yui.github.io/yuidoc/syntax/index.html#basic-requirements), YUIDoc supports markdown and also recognizes two line breaks as start of a new paragraph. 

I'm not that familiar with YUIDoc myself but if possible I think it will be great to minimize the amount of html tags that needs to be inserted into the inline docs, removing the need for the `<br />` tags.

On a side note, I was thinking about removing the need for the `<div>` and `<code>` tags for the example but I think that means we cannot have a `norender` class on the example code `<div>` tags?

toolness:
Yeah, a style guide would be great! One of the particular things that always gnaws at me is the fact that JS symbols like functions/variables/constants are never `monospaced` when they're used in sentences. Most of the time this is just my OCD at work, but sometimes I think it genuinely hampers readability and makes things a bit confusing.

I generally agree regarding the `<div>` and `<code>` stuff but as you mentioned, we'd need to figure out what to do about `norender`. One option is to add a specially-formatted comment to the beginning of the sample like `// NORENDER`, which our documentation pipeline notices, removes from the code shown to the user, and then adds to the rendered HTML.

I also address some of this in https://github.com/processing/p5.js/pull/1144#issuecomment-160419254, regarding browser capabilities required for certain examples to run.

kaganjd:
+1 to documentation style guide! I'm working on autocomplete this summer. The YUIDoc data.json file is the starting point of a process for generating the file that Tern.js uses to do autocomplete. 

In data.json, I'm running into lots of typos (#2042, #2016 ), missing fields ([color_conversion](https://github.com/processing/p5.js/blob/37f1e69e8027c1745adfb7c18c50b1fbbb7c0d49/src/color/color_conversion.js), for example, doesn't list any `@method`s which means that when I generate the Tern.js file, color_conversion methods don't show up), and inline HTML tags that all ultimately affect how autocomplete works.

anhtuann:
@kaganjd any news on the tern autocompletion file ? I would love to get it to work with p5js

PR #1343, opened 2016-04-01
Render documentation descriptions as Markdown

### LINKED ISSUE #1337: Allow cross-references to other parts of the documentation
It would be helpful if there were an easy way to cross-reference to other parts of the documentation from within it. For instance, the documentation for [`fill()`](http://p5js.org/reference/#/p5/fill) mentions [`colorMode()`](http://p5js.org/reference/#/p5/colorMode):

colormode

It would be great if that was actually a link to the documentation for `colorMode()` rather than just static text.

YUIDoc does seem to have a mechanism for [cross-referencing modules and classes](http://yui.github.io/yuidoc/syntax/index.html#cross-referencing-modules-and-classes), but as far as I can tell, it assumes a certain syntax for documentation URLs that p5 doesn't conform to. I'm also not a huge fan of how verbose it is.

So, one option is to come up with our own alternative formatting. I guess one way we could do it, which might even work without having to write any additional tooling/preprocessing, is simply to use Markdown to relatively link to the appropriate place, e.g.:

```markdown
depending on the current [`colorMode()`](#/p5/colorMode).
```

I like this approach because it takes advantage of the nice URL fragment syntax the docs already use. It also doesn't really require any explicit documentation, and contributors can easily figure out what the fragment is by just visiting the proper section on the website and copy-pasting the fragment.

If we did things this way, I could probably even add a step to our test suite that checks to make sure that none of these relative links are broken.

Thoughts?

toolness:
While trying to fix #1337, I realized that our documentation descriptions aren't currently rendering as Markdown. p5's use of YUIDoc had led me to assume otherwise.

I believe the reason for this is because YUIDoc is built for *statically* generating documentation, and it renders Markdown during this static generation phase. However, p5's static generation phase is actually extremely simple, because its documentation is a single-page app, rendered at [`main.handlebars`](https://github.com/processing/p5.js/blob/master/docs/yuidoc-p5-theme/layouts/main.handlebars).

Instead, the majority of page rendering is actually done on the client-side via the Backbone app contained in [`docs/yuidoc-p5-theme-src`](https://github.com/processing/p5.js/tree/master/docs/yuidoc-p5-theme-src). This app looks at `data.json` directly to retrieve all the reference documentation data, and *that* file has the original, raw descriptions for our documentation--*not* rendered Markdown.

This is why we have to have `<br><br>`'s between every paragraph in our documentation, rather than simply inserting a blank line, as YUIDoc leads us to believe (which @limzykenneth has noted in #1190).

So this PR attempts to ease this difficulty by using our [YUIDoc preprocessor](https://github.com/processing/p5.js/pull/1295#issuecomment-201275694) to convert all descriptions in `data.json` to Markdown at build time (via the [marked](https://github.com/chjj/marked) npm module). I decided to take this approach rather than rendering the Markdown on the client-side for a few reasons:

* There isn't any dynamic functionality pertaining to Markdown rendering that actually *requires* it to be on the client side.
* Rendering it on the client-side means shipping a Markdown-rendering library to the user's browser, which isn't great given that the [docs already load slowly](https://github.com/processing/p5.js-website/issues/143). (According to GitHub, [`marked.min.js`](https://github.com/chjj/marked/blob/master/marked.min.js) weighs in at 19K.)
  * Actually, in retrospect, there is a tradeoff here: because Markdown is generally more compact than HTML, the `data.json` blob will be larger if we render the Markdown at build time. So not sure if this ends up being a net gain.
* Rendering it on the client-side potentially slows down the client/drains battery life.
* Rendering it at build time means we can (eventually) statically analyze it and ensure its correctness, e.g. that any internal links aren't broken.

That said, because this PR now renders everything as Markdown, it's quite likely that some documentation doesn't appear exactly like it used to.

### DIFF
--- a/docs/yuidoc-p5-theme/assets/reference.css
+++ b/docs/yuidoc-p5-theme/assets/reference.css
@@ -1,0 +1,23 @@
+/**
+ * Note: most of the styling for the reference documentation is actually
+ * contained in the p5.js-website repository:
+ *
+ * https://github.com/processing/p5.js-website/blob/master/css/main.css
+ *
+ * However, some of it is located here so we can easily iterate on it
+ * in the same repository.
+ */
+
+#reference .params table p {
+  /* Recently-added support for Markdown means that every parameter
+   * description is wrapped in a paragraph element. (Previously, they weren't
+   * wrapped in any kind of element.)
+   *
+   * We may eventually want to display paragraphs as blocks, so that we
+   * can have lengthy descriptions for parameters, but for now we'll
+   * keep our pre-existing behavior and essentially make paragraphs
+   * "invisible" by rendering them inline. */
+  display: inline;
+
+  font-size: inherit;
+}
--- a/docs/preprocessor.js
+++ b/docs/preprocessor.js
@@ -1,3 +1,5 @@
+var marked = require('marked');
+
 var DocumentedMethod = require('./yuidoc-p5-theme-src/scripts/documented-method');
 
 function smokeTestMethods(data) {
@@ -91,9 +93,32 @@ function mergeOverloadedMethods(data) {
   });
 }
 
+function renderItemDescriptionsAsMarkdown(item) {
+  if (item.description) {
+    item.description = marked(item.description);
+  }
+  if (item.params) {
+    item.params.forEach(renderItemDescriptionsAsMarkdown);
+  }
+}
+
+function renderDescriptionsAsMarkdown(data) {
+  Object.keys(data.modules).forEach(function(moduleName) {
+    renderItemDescriptionsAsMarkdown(data.modules[moduleName]);
+  });
+  Object.keys(data.classes).forEach(function(className) {
+    renderItemDescriptionsAsMarkdown(data.classes[className]);
+  });
+  data.classitems.forEach(function(classitem) {
+    renderItemDescriptionsAsMarkdown(classitem);
+  });
+}
+
 module.exports = function(data, options) {
+  renderDescriptionsAsMarkdown(data);
   mergeOverloadedMethods(data);
   smokeTestMethods(data);
 };
 
 module.exports.mergeOverloadedMethods = mergeOverloadedMethods;
+module.exports.renderDescriptionsAsMarkdown = renderDescriptionsAsMarkdown;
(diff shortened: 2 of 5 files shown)


docs/yuidoc-p5-theme/assets/reference.css
```diff
@@ -0,0 +1,23 @@
+/**
+ * Note: most of the styling for the reference documentation is actually
+ * contained in the p5.js-website repository:
+ *
+ * https://github.com/processing/p5.js-website/blob/master/css/main.css
+ *
+ * However, some of it is located here so we can easily iterate on it
+ * in the same repository.
```
toolness: Um, sorry if this is a terrible idea :sweat:

Would it be useful to actually move all the reference-related CSS from [p5.js-website's `main.css`](https://github.com/processing/p5.js-website/blob/master/css/main.css) over here, though? Or is some of that reference CSS used elsewhere on the p5.js website?

limzykenneth:
I'm a little bit confused...with this PR will description in data.json be in markdown or rendered html from markdown?

toolness:
It will be in rendered HTML from markdown. We can change it to contain the original markdown, but this will mean rendering the markdown on the client-side (in the Backbone app), where we will incur the opposite of the tradeoffs I listed in the description above.

Even if we do that, though, we'll likely (eventually) want to at least *parse* the markdown at build time to ensure there aren't any broken links and such.

limzykenneth:
hmm...do you have a branch where the data.json will be in markdown and we parse markdown on client side? I'd like to do some tests and comparisons see what works best.

toolness:
Ah, nope, unfortunately I do not currently have such a branch. :/

docs/yuidoc-p5-theme/assets/reference.css
lmccart: The main.css file holds all the css for the whole site. I suppose we could separate it out, but there are some things we want to ensure keep consistent between the reference and the rest of the site, and it seems like this might just end up in duplicated code?

lmccart:
This all looks good, let's give it a try. Thanks @toolness. @limzykenneth if you do any further testing in this direction let us know, this is all a work very much in progress as you know. :)

limzykenneth:
Cool, will report back as soon as I find some extra time to do some test on this... :sob:

PR #1351, opened 2016-04-02
Don't trigger warnings on preload().

### LINKED ISSUE #1350: using preload() triggers warning messages
Here's a sample:

```javascript
var img;

function preload() {
  img = loadImage("assets/rockies.jpg")
}

function setup() {
  image(img, 0, 0);
}
```

The code runs fine but the console output reads:
```
You just changed the value of "loadJSON", which was a p5 function. This could cause problems later if you're not careful.
p5.js:12099 You just changed the value of "loadImage", which was a p5 function. This could cause problems later if you're not careful.
p5.js:12099 You just changed the value of "loadStrings", which was a p5 function. This could cause problems later if you're not careful.
p5.js:12099 You just changed the value of "loadXML", which was a p5 function. This could cause problems later if you're not careful.
p5.js:12099 You just changed the value of "loadTable", which was a p5 function. This could cause problems later if you're not careful.
p5.js:12099 You just changed the value of "loadFont", which was a p5 function. This could cause problems later if you're not careful.
```

I'm guessing we just need to account for the preload maneuvering in the recent overwriting warnings we added. cc @toolness.

toolness:
This fixes #1350.

It's not a completely optimal fix, because it actually completely disables warning messages for overwriting preload methods. This means that even in the case of legitimate user error like this:

```js
function preload() {
  loadJSON += "blah";
  loadJSON('http://stuff');
}

function setup() {}
```

the user won't get a warning. However, my hunch is that users are unlikely to accidentally overwrite the preload methods, since their names are more unusual than other p5 functions and thus a lot less likely to accidentally be overwritten. I could be wrong, though, in which case we could implement something a bit more complex (well, harder to unit test, at least) like https://github.com/processing/p5.js/issues/1350#issuecomment-204802589.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -556,7 +556,9 @@ p5.prototype._createFriendlyGlobalFunctionBinder = function(options) {
   };
 
   return function(prop, value) {
-    if (typeof(IS_MINIFIED) === 'undefined' && typeof(value) === 'function') {
+    if (typeof(IS_MINIFIED) === 'undefined' &&
+        typeof(value) === 'function' &&
+        !(prop in p5.prototype._preloadMethods)) {
       try {
         // Because p5 has so many common function names, it's likely
         // that users may accidentally overwrite global p5 functions with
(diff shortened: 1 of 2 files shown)


lmccart:
good point. this is great, thanks!

Issue #1316, opened 2016-03-22
Vertically centering multiline text


pete-rai:
In old processing.js textAlign(CENTER, CENTER) used to vertically center the text on the specified x, y point even when it was multiline (i.e it included "\n"). In p5js, it appears to always vertically center align the first line only. The other lines just follow below. It this the intended new functionality? If it is, then thats a lot of extra jiggling we will all need to do in our own code :(

As an example, I would expect the text in the example below to be in the center of the circle:

    ellipse(100,100,100,100);
    textAlign(CENTER, CENTER);
    text("ABCD\nEFGH", 100, 100);

lmccart:
Nope, this is not intentional, definitely a bug, thanks! I will check it out.

kaganjd:
When I try to replicate this, I see both lines affected by the textAlign arguments, but RIGHT and LEFT are reversed. I've attached some screen shots. Is this what you get?
<img width="470" alt="screen shot 2016-04-01 at 1 24 55 pm" src="https://cloud.githubusercontent.com/assets/9204835/14214719/94bd3fe6-f80c-11e5-8bd1-1c6e907075cd.png">
<img width="463" alt="screen shot 2016-04-01 at 1 24 42 pm" src="https://cloud.githubusercontent.com/assets/9204835/14214720/94bd7e7a-f80c-11e5-9dc9-f985de395ac4.png">

kaganjd:
Just kidding- I understand the problem now. Working on it!

lmccart:
fixed with https://github.com/processing/p5.js/pull/1369

dhowe:
I dont think this actually fixed the problem, but it is better described now in #1410

ProgrammingLife:
It doesn't fixed yet. Now it works wrong for horizontal center alignment.

PR #1369, opened 2016-04-22
fix for vertically centering multiline text (issue #1316)

kaganjd:
also updated documentation for textAlign() function


lmccart:
woohoo!!

PR #1287, opened 2016-03-05
[WIP] Add experimental type-checker for reference tests

toolness:
**This is a [work-in-progress pull request](http://vrybas.github.io/blog/2014/04/11/wip-pull-requests/), please don't merge it yet.**

@lmccart: let me know if it's annoying to have a WIP PR lying about here--it's fine if that's the case, I can just move it over to my fork. I thought it might be nice to have it in the p5.js PR list so others can participate.

This is an attempt to use p5's own yuidocs to actively type-check its own examples. This can be beneficial for a number of reasons:
- It ensures that our documentation is accurate. For example, if `lerpColor()` can actually take a `p5.Color` as a first argument--_and_ it's likely that users will want to do this--but it isn't documented, then we should document it. Or if the argument type `Number` is misspelled as `Nimber`, we should throw an error at build or test time.
- It means that we can _potentially_ use our yuidoc metadata to build a more robust friendly error system, as per #759.

Limitations of this PR:
- Eventually, we should support some way of providing custom validation for methods that have complicated or problematic signatures, like [`save()`](http://p5js.org/reference/#/p5/save). There's a number of ways we can do this but for now we'll use one of two strategies:
  - Store a blacklist that excludes such methods, or even entire files, from validation entirely. This is defined as `FILES_TO_IGNORE_FOR_NOW` and `METHODS_TO_IGNORE_FOR_NOW` in `test/js/typeChecker.js`.
  - If the custom validation isn't very complex, we can write our own custom validator for now by adding an entry to the `customValidators` mapping in `test/js/typeChecker.js`.
- Ideally, we'd also type-check against the standard test suite, but this introduces complications to the build system, since it means that we'd have to generate yuidocs metadata before running tests, so we'll deal with that later if this PR is successful.
- This doesn't actually implement #759, it just lays some groundwork for doing so, while also providing the benefit of ensuring the continual correctness of our documentation. After all, before actually implementing #759, we need to make sure that the advice we're giving users in our friendly error system is actually accurate! :grin: 
- This PR doesn't currently check that the _return value_ of methods is actually what they're documented to be; this can be added in a future PR too.

Other concerns:
[...]

PR #1370, opened 2016-04-22
Merges in webgl geometry refactor

indefinit:
feature additions:
obj loader - loadModel()
more 3d and 2d primitives in webgl (ellipse, rect, triangle)
modularized p5.Geometry for ad hoc mesh creation.


indefinit:
Forgot to add this note.  Upon merging in the geometry refactor branch I get this warning from Grunt:

Warning: for p5.ellipse() defined in src/core/2d_primitives.js:133, all overloads must be defined in the same file ("src/3d/primitives.js" !== "src/core/2d_primitives.js") Used --force, continuing.
Warning: Cannot read property 'project' of undefined Used --force, continuing.

For now, I've forced the grunt compilation but I'm wondering if anyone has a recommendation for structuring the overloaded methods from 2d/3d.

indefinit:
It looks like the documentation comments are what's throwing the overloads warning.  Maybe I should just add a webgl flag to each overloaded method documentation?

lmccart:
hey @indefinit wow what a ton of work, it looks great! a couple comments:

* can you remove `lib/addons/p5.dom.min.js` from the PR? it is listed in the .gitignore and shouldn't be included in the repo.
* regarding your question about documentation. the best way to do this would be rather than making an additional entry (like you have [here](https://github.com/processing/p5.js/pull/1370/files#diff-33cc0bccd71995d911c080eb1e185610R634)), modify the entry in the [top level definition](https://github.com/processing/p5.js/blob/master/src/core/2d_primitives.js#L152) to include additional parameters and signatures (so we end up with an entry similar to what we see in the [processing ref](
https://processing.org/reference/line_.html). recently, @toolness added functionality that supports the display of multiple signatures which should help make things more clear, info [here](https://github.com/processing/p5.js/wiki/Inline-documentation#additional-signatures).

other than that, all looks good to go!

indefinit:
Ok thanks.  I'll work on this.  Question about _validateParameters.  Should I just delete this code in the p5.prototype methods since it's known to be disabled?

indefinit:
So close to passing all checks.  Now Travis fails on: 1) src/typography/attributes.js textStyle documentation example #1 works:
     code took too long to execute

Any ideas?  I don't recall ever working on src/typography so I don't think the error is related to my refactor.

lmccart:
hm good question about `_validateParameters`, @toolness what do you say?

is the test failing when you run grunt locally or just with the travis CI test? when I looked at the error log from travis, it looked like it passed on 3/4 different node builds. sometimes I notice that travis will randomly fail for one of the node versions, but then work again the next time the tests are run. I'm not sure what we need to do to fix this, but I usually just ignore it and things work out ok. :) ?

toolness:
> hm good question about `_validateParameters`, @toolness what do you say?

I need to blog more about this, but my basic philosophy right now is to transparently validate the vast majority of parameters through an automated mechanism based on our yuidoc metadata like e.g. #1287, and to resort to hand-crafted, artisanal type-checking for functions with unusual type signatures like [`save()`](http://p5js.org/reference/#/p5/save). So I think that it's okay to just remove `_validateParameters` now!

indefinit:
@lmccart yea build passes tests fine on grunt but fails in Travis.  I'll go ahead and ignore.  

@toolness sounds good to me.  I'll go into the 2d primitives and remove _validateParameters for now.  As soon as I remove and push those changes, we'll be good to go with the merge.  Woohoo!

indefinit:
Sweet! It looks like all checks have passed.  @lmccart Should I merge or do you want to do it?

lmccart:
go for it!

PR #1372, opened 2016-04-23
Fixed pixels[] calculation in get()-docs

tiktuk:
X should be multiplied by four to get the right byte for each RGBA-component, right?


lmccart:
good catch, thanks! I think we actually want the *4 outside the parenthesis (like [this](http://p5js.org/examples/examples/Color_Brightness.php)), I'll merge this and make the change.

PR #1397, opened 2016-05-09
get() returns blank image if metadata not loaded and no arguments

shiffman:
`get()` return as `p5.Image` if no arguments are passed in.  For examples that are copying from a video element, this fix allows `get()` to return a blank image while waiting for the `loadedmetadata`


shiffman:
@lmccart I'm not sure if this is a "good" solution to the issue I was running into.  This is my testing example:

```javascript
var video;

function setup() {
  createCanvas(640, 240);
  video = createCapture(VIDEO);
  video.size(320, 240);
}

function draw() {
  image(video, 0, 0);
  var video2 = video.get();
  image(video2, 320, 0);
}
```

lmccart:
looks reasonable to me, thanks!

PR #1401, opened 2016-05-10
Quad bug fixes

mindofmatthew:
@indefinit: I noticed that webgl quad() is broken as of 0.5.0. This PR fixes that, by replacing the call to calculateUVs with hand-coded UVs (pinning a corner of the texture to each corner of the shape, but we can also adopt a different approach).


indefinit:
@mindofmatthew whoops!  Good catch!  Your code looks good to me.  Not sure if I fully understand why the Travis build is failing.  Do you mind git pulling the upstream to make sure it's up to date and then grunt again?  I've had this problem before and after git pull and another push Travis passed all the checks.

mindofmatthew:
@indefinit I just checked and Github says that I'm up to date with the master branch. Looking at the logs, it just looks like that build job was delayed for some reason which caused the log to get larger than Travis's limit (the Nyan Cat test runner generates a lot of log code).

Would you mind re-running the Travis build, just in case it was a transient issue? It looks like I can't re-run it myself without commit access.

indefinit:
Reran the build but it's failing on the following unit test:
  1) time and date p5.prototype.second second() should return this second:
     expected 6 to equal 5

Given your changes aren't touching any time functions I think we can ignore and merge the PR.  @lmccart what do you think?

lmccart:
yeah, sounds good. I'll look into the failing test and see what's up.

Issue #1413, opened 2016-05-13
Text size bug when drawing opentype font with multiple renderers


mikewesthad:
I was using an offscreen renderer with an opentype font and ran into a bug.  When drawing to the offscreen renderer, the text size that is being applied is from the main renderer.  The problem looks to be coming from [here](https://github.com/processing/p5.js/blob/master/src%2Ftypography%2Fp5.Font.js#L228).  When getting path data from the font, the parent of the font is used for the text size - rather than the current renderer.

[Codepen demo](http://codepen.io/mikewesthad/pen/zqXdNQ)

I'll submit a PR with a proposed fix.

dhowe:
I think the way to support more use-cases like your example is to allow a renderer to be passed in (as part of the options object) for all functions in p5.Font, the way it is now in _renderPath, and in your PR. I can't think of any disadvantages to this offhand.

mikewesthad:
@dhowe right, good point.  I just took a look to see what else could use an options parameter:

`p5.Font.textBounds` & `p5.Font.textToPoints` have fontSize as an explicit parameter.  Since those are part of the public API, an options parameter might confuse beginners reading the docs.

`p5.Font._renderPath` has options, and `p5.Font._getPath` will have it if the PR is okay.

`p5.Font._handleAlignment` is being passed a reference to a specific p instance, so in theory, it should have the renderer.  But it is calling `p5.Font._textWidth`, etc. without passing in an argument, so those methods are looking up fontSize from `this.parent` rather than the specific p instance.  That seems like it could be problematic, so fontSize should probably be pulled from the p instance's renderer.  (The p variable here could be nixed in favor of an options parameter - for consistency with _renderPath and _getPath.)

`p5.Font._textWidth`, `p5.Font._textAscent`, `p5.Font._textDescent` &  `p5.Font._scale` already have fontSize as a parameter and don't seem to need a reference to the renderer.  That said p5.Renderer2D is calling those methods without supplying a fontSize.  That also seems like it could be a problem, but since fontSize is a parameter, that seems like a place where the 2D renderer should just pass in its font size.

mikewesthad:
[Codepen demo](http://codepen.io/mikewesthad/pen/NNmYvr) for bug with `p5.Renderer.textWidth`.

[Codepen demo](http://codepen.io/mikewesthad/pen/MyRVEv) for bug with `p5.Font._handleAlignment`.

[...]

Issue #1410, opened 2016-05-12
Incorrect vertical alignment of multi-line text


dhowe:
It appears that multiline text is not following vertical alignment rules.

Image: pink is Processing, bottom is p5js, test code [here](https://github.com/processing/p5.js/blob/master/test/manual-test-examples/p5.Font/custom/sketch.js#L277)...

![screen shot 2016-05-13 at 4 03 52 am](https://cloud.githubusercontent.com/assets/737638/15228784/76ba8222-18c0-11e6-9a49-23212230b1d5.png)

mikewesthad:
@xyfeng - I see @dhowe assigned this to you both. Are you working on it? If not, I'll have some time this weekend to check it out.

lmccart:
fixed with https://github.com/processing/p5.js/pull/1433

PR #1415, opened 2016-05-13
Fixes for #1413: bugs when drawing opentype fonts with multiple renderers

mikewesthad:
Fixes for #1413 (bug demos in issue): 

These changes fix some bugs that arise when using a p5.Font with multiple renderers.  The issues come from the font assuming it's `parent` property is the appropriate p5 instance to use to look up the font size & font alignment.

Changes:

- `Font._renderPath` now passes along the renderer that it receives (via `options`) to `Font._getPath`, so that the appropriate text size can be looked up and applied.  Without this, `Font._getPath` defaults to the renderer that is stored when the font was created (`font.parent`).  This caused the text size and text alignment to always be based on `font.parent`.
- `Font._getPath` was being called with an extra (old?) argument for text size that wasn't being handled, so I removed that.
- `Renderer2D.textWidth` was not specifying a fontSize when calling `Font.textWidth`.  This caused textWidth to always be based on the font's `parent`.
- `Font._handleAlignment` was not using it's p5 instance parameter to look up the font size.  Alignment was thus always based on `this.parent`'s font size.

I tested this locally and everything seems to be working.  Demos from #1413 now all work as expected.

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -1195,7 +1195,7 @@ p5.Renderer2D.prototype.textWidth = function(s) {
 
   if (this._isOpenType()) {
 
-    return this._textFont._textWidth(s);
+    return this._textFont._textWidth(s, this._textSize);
   }
 
   return this.drawingContext.measureText(s).width;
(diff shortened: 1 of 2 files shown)


lmccart:
this looks good to me, @dhowe what do you think? feel free to merge if it's set.

dhowe:
Indeed. I believe that the only open alignment issue now is https://github.com/processing/p5.js/issues/1410

Issue #1416, opened 2016-05-14
p5.Font.textBounds does not take into account text alignment


mikewesthad:
Hopefully I'm not flooding things with `p5.Font` bugs.  I was working on a demo using a `p5.Font` to create text particles, and I noticed that `p5.Font.textBounds` doesn't adjust the bbox's (x, y) coordinates based on the renderer's text alignment settings.  It defaults to opentype's positioning scheme: left & baseline.

[Codepen demo](http://codepen.io/mikewesthad/pen/QNPxLz).  Below is a snapshot from the demo.  The orange dot is the (x, y) anchor where the text is being drawn.  The orange box is the bbox.  The gray text displays the alignment settings ("horizontal-vertical") for the text.

I'm looking at `p5.Font` now.  I think it's just a matter of shifting the (x, y) at the top of `Font.textBounds` (using `Font._handleAlignment`) and making sure that caching stays intact.  I'll work on a suggested PR.

![image](https://cloud.githubusercontent.com/assets/1131802/15269770/821744ce-19cf-11e6-8039-e110fc71a5f3.png)

dhowe:
Nice work! This looks really close but the dot-position for valign=TOP looks incorrect to me. Here is what we get in Processing (which is our target, but doesn't have a built-in textBounds function):

![screen shot 2016-05-15 at 8 46 50 pm](https://cloud.githubusercontent.com/assets/737638/15274123/5147dc36-1ade-11e6-97d4-6fd6fc93a9b3.png)

dhowe:
I wonder also if we need to consider #766 here...

mikewesthad:
For posterity purposes, summarizing the discussion from #1417 here: Vertical alignment looks to be an independent issue, opened #1421 to deal with that.

lmccart:
closed with https://github.com/processing/p5.js/pull/1417, #1421 remains to address vertical text

PR #1417, opened 2016-05-14
Fix for #1416: Update Font.textBounds to take text's alignment and baseline into account

mikewesthad:
Fix for #1416:

Updating `Font.textBounds` to take into account the text's baseline and alignment when calculating the bounding box:

- `textBounds` uses `_handleAlignment` to shift the bbox's x, y based on baseline and alignment.  This required that `_handleAlignment` take an optional `textWidth` parameter (to prevent recursion with `_handleAlignment` calling `textBounds` to find the width).
- Caching now uses the renderer's baseline and alignment in the key.

Caching tested via code like this:

```js
var string = "test";
textAlign(LEFT, BASELINE);
fill("#00A8EA");
text(string, 100, 100); // No cache hit
text(string, 100, 100); // Cache hit
textAlign(CENTER, CENTER);
fill("#FF8132");
text(string, 100, 100); // No cache hit
text(string, 100, 100); // Cache hit
```

The bug [demo](http://codepen.io/mikewesthad/pen/QNPxLz) on codepen now looks like this when tested against this PR.  The orange dots represent the (x, y) coordinates of text draw command and the boxes are the text bounds.

image

### DIFF
--- a/src/typography/p5.Font.js
+++ b/src/typography/p5.Font.js
@@ -98,7 +98,17 @@ p5.Font.prototype.textBounds = function(str, x, y, fontSize, options) {
   y = y !== undefined ? y : 0;
   fontSize = fontSize || this.parent._renderer._textSize;
 
-  var result = this.cache[cacheKey('textBounds', str, x, y, fontSize)];
+  // Check cache for existing bounds. Take into consideration the text alignment
+  // settings. Default alignment should match opentype's origin: left-aligned &
+  // alphabetic baseline.
+  var p = (options && options.renderer && options.renderer._pInst) ||
+    this.parent,
+    ctx = p._renderer.drawingContext,
+    alignment = ctx.textAlign || constants.LEFT,
+    baseline = ctx.textBaseline || constants.BASELINE;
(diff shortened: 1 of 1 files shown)


dhowe:
nice!

dhowe:
Nice work! This looks really close but the dot-position for valign=TOP looks incorrect to me. Here is what we get in Processing (which is our target):

screen shot 2016-05-15 at 8 46 50 pm

mikewesthad:
@dhowe thanks for checking that! Hmm, none of the changes I made should effect the dot placement relative to the text. The [codepen demo](http://codepen.io/mikewesthad/pen/QNPxLz) is using v0.5.0 p5 and has the same vertical alignment of text mismatch with processing.

Since I was just focusing in on the bbox, think the vertical alignment mismatch is an existing bug that warrants a separate issue/PR? Happy to help & look into a fix.

dhowe:
mmm, I wondered if that was unrelated -- so, yes please do...

mikewesthad:
Opened #1421 - investigating the issue of font vertical alignment there

Issue #1412, opened 2016-05-13
model() not displaying .obj files


k808a:
Been trying to re-create the use of model() and loadModel() from the P5.js examples page (  [https://p5js.org/reference/#/p5/model](url) ) without success.

All I get is the gray canvas, without the spinning teapot. Results are the same with the P5 editor app and on my Apache server. (The editor app doesn't display any errors when I run the code, BTW.)

The code I'm using (straight copy-paste from the examples page) is this: 

`
//draw a spinning teapot
var teapot;

function setup(){
  createCanvas(100, 100, WEBGL);

  teapot = loadModel('assets/teapot.obj');
}

function draw(){
  background(200);
  rotateX(frameCount * 0.01);
  rotateY(frameCount * 0.01);
  model(teapot);
}
`

I'm running p5.js v0.5.0 May 02, 2016 (which does seem to have code for loadModel() and model() implemented) with a 264KB teapot.obj file I found from the web. (The .obj file shows up properly in Photoshop, so I'm fairly confident it's OK. I also tried other .obj files, same result.)

I have tried moving teapot.obj out of the assets folder to the same level as sketch.js, also no luck.

Any insights?

indefinit:
Do you have a link to teapot.obj by any chance?  If you aren't seeing errors, then the only thing off the top of my head is that maybe the model scale is very small?  Also, which 3D program did you use to generate teapot.obj?

indefinit:
Sorry, just read again and saw you said the teapot was found on the web.  If you have a Dropbox or other link to the file I can take a look

k808a:
Here's a direct link: [http://www.subdude-site.com/WebPages_Local/RefInfo/Computer/Linux/LinuxGuidesByBlaze/apps3Dtools/3D_viewers-converters/3Dtestfiles/obj/teapot.obj] 
I found it on this page, which was created in 2011: [http://www.subdude-site.com/WebPages_Local/RefInfo/Computer/Linux/LinuxGuidesByBlaze/apps3Dtools/3D_viewers-converters/3Dtestfiles/3D_testfiles_withTOC.htm](url)
I tried three other files from other sources as well; like teapot.obj, they all displayed fine in Photoshop, but not when I ran the example code in my P5 editor or on the server. 
Unfortunately I haven't been able to find the original used on p5js.org

lmccart:
here's the one from the reference examples:
http://p5js.org/reference/assets/teapot.obj

lmccart:
[...]

PR #1418, opened 2016-05-14
Normalize model import

mindofmatthew:
In response to the discussion surrounding #1412, I went ahead and implemented a normalization feature for loadModel, following @indefinit's suggested behavior.


src/webgl/loading.js
```diff
@@ -38,11 +47,15 @@ require('./p5.Geometry');
  * </code>
  * </div>
  */
-p5.prototype.loadModel = function ( path ) {
+p5.prototype.loadModel = function ( path, normalized ) {
```
indefinit: I wonder whether we should use 'normalize', instead of 'normalized'.  My thinking is that since we want to normalize the model as an action, it is kinda happening in the present/future tense instead of 'normalized' which feels like the model has already been normalized.   What do you think?

indefinit:
Other than my little nomenclature comment, this looks great and is a really cool addition to p5.Geometry!

mindofmatthew:
That makes sense. Do we have any other variables that are similarly named? Unless there's some other precedent, "normalize" is fine with me.

indefinit:
I can't seem to find any precedence, but I did find that almost all of the other loader functions have callback arguments.  I'm a fan of terseness, so I'm not sure if we want to go down the route of 3 arguments (last two optional).  That said, accepting a callback would make this function consistent with the rest of the p5 api.  Let's see what @lmccart or @shiffman think :)

lmccart:
is loadModel() asynchronous? if so, then yes, I think it'd be good to match the model for the other loadXX() functions and support it in preload and with a callback.

shiffman:
If it's asynchronous I agree it needs the callback.   How loading media works is quite a source of confusion for students so it's important for us to be consistent.   Is "normalize" something you would do once its loaded and could this just be a single callback?

mindofmatthew:
I checked in a version with callbacks. Right now, it accepts both: `loadModel(path, normalize, successCallback, failureCallback)` and `loadModel(path, successCallback, failureCallback)`. I wasn't sure if there was a good way to indicate this in documentation, so I just said that everything other than path was optional.

lmccart:
great!

Issue #1421, opened 2016-05-15
Vertical text alignment doesn't match processing


mikewesthad:
In #1417, @dhowe pointed out that the vertical alignment (specifically vertical align = TOP) for text in p5 doesn't match processing.

Snapshot from a [codepen demo](http://codepen.io/mikewesthad/pen/QNPxLz) in p5
![image](https://cloud.githubusercontent.com/assets/1131802/15276092/436a26f4-1aa3-11e6-8ab3-8edd1bfc36ef.png)

@dhowe's snapshot from processing
![image](https://cloud.githubusercontent.com/assets/1131802/15276089/2fe68104-1aa3-11e6-8c77-0a75cd5ffe62.png)

I'm checking this out now, but my guess is that the problem comes from this [line](https://github.com/processing/p5.js/blob/master/src%2Ftypography%2Fp5.Font.js#L426) in `p5.Font._handleAlignment`.  When the text is set to be drawn with baseline top, the y coord is shifting by `textHeight` but should instead be shifted by `textAscent`.  As far as I can tell, that's what Processing does.

lmccart:
looks like this one is set then? or is there more to do here still?

mikewesthad:
Yeah, #1421 fixes this for most fonts, so this can be closed. There's still going to be some small amount of inconsistency between processing vs p5 system font vs p5 opentype font since they all use different methods of calculating baselines (sampling a raster vs browser canvas vs values specified by font designer).

dhowe:
I believe that we still need to deal with vertical alignment for multi-line text however (see https://github.com/processing/p5.js/issues/1410)

mikewesthad:
@dhowe, right - yeah this issue/PR was single line focused, so mutli-line is still a problem.  I can pitch in over on that issue if no one else is already working on it.

dhowe:
If you have the bandwidth, that would be great -- I'm in semester grading so won't get to it for a while...

PR #1423, opened 2016-05-16
Fix for #1421: p5.Font vertical alignment doesn't match processing

mikewesthad:
This issue (#1421) is a little thorny.  I made a manual test (following @dhowe's example) e3ee70d24f28a0d5614e965389271859efaae7c8 that highlights the problem:

Four different fonts are rendered with all possible combinations of horizontal and vertical alignment.  On the left side is processing's text rendering, and on the right, p5's text rendering.  The red triangles highlight the places where there is a sizable mismatch between p5 and processing: vertical align top & vertical align middle.  Everything else is passable.

before

This is fairly new to me, so I'm going to throw the background and reasoning for the changes in the PR here:

**What processing does:**

Ascent & descent are calculated through a rasterization approach. "p" is drawn and then scanned to find the descent. "d" is used for the ascent.  (Inline [comments](https://github.com/processing/processing/blob/master/core%2Fsrc%2Fprocessing%2Fcore%2FPFont.java#L96) describing this; [code](https://github.com/processing/processing/blob/master/core%2Fsrc%2Fprocessing%2Fcore%2FPFont.java#L1089) that does the calculation).

**What p5 does:**

Ascent/descent are coming from the font file headers (via opentype). There are actually multiple variants of ascent/descent metrics inside of the headers, and these are set by the font designer so there's a bit of variability here. p5 is using the ascent/descent metrics from the Horizontal Header table (hhea). These values represent the "distance from baseline of highest ascender/descender" (source: [apple dev](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6hhea.html)).

~~Processing's raster approach is closer in value to a different font metric - the sTypoAscender & sTypoDescender from the OS/2 table in the font header (ref: [apple dev](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6OS2.html)). These values (at least for Western fonts) are smaller than the hhea values and line up well with processing's measurements of "p" and "d" height.~~

**EDIT: after additional tests with more fonts, the original ascender/descender from the hhea table was the best option.  The real issue was that _handleAlignment didn't match processing.**

**Changes**

- ~~Use OS/2 sTypoAscender & sTypoDescender when available, default to hhea ascender & descender if OS/2 is unavailable.~~
- Update `p5.Font._handleAlignment` to match processing's coordinates (relevant [source](https://github.com/processing/processing/blob/f6fb24b750743ac95c9060b73cf0f955a84967a9/core%2Fsrc%2Fprocessing%2Fcore%2FPGraphics.java#L4537))

It won't be a perfect match without adopting a raster approach, but this seems much better (tested in Chrome & Firefox on PC):

after


dhowe:
This looks great. We might also want to double-check these alignments with system fonts (times, arial, georgia, etc.). As in [this code](https://github.com/processing/p5.js/blob/master/test/manual-test-examples/p5.Font/system/sketch.js#L237).

mikewesthad:
@dhowe I'll follow up in this PR this evening about that.  The system font drawing shouldn't be affected by any changes here, but it does make sense to zoom out to get a bird's eye view of vertical alignment in general.

mikewesthad:
I added a some tests for both system fonts and opentype fonts.  After looking at more fonts wrt vertical text alignment, I realized that the original descender/ascender metric used is a better overall fit.  The real issue was just the `_handleAlignment` math not matching processing's math.  Everything should be good to go now.  This PR brings p5 closer to processing, but it's not going to line up perfectly because of the difference in approaches to calculating ascent/descent (raster vs metrics in font headers).  The font headers metrics are set by the designers so there are no guarantees that they will be consistent.

**Tests**

The new tests are a little abrasive on the eyes, but they make comparing p5 to processing easier.  p5 is in red in the foreground, processing is in black in the background.

p5.Font vertical alignment focused testing (textVertAlignmentSketch in p5.Font/custom):

image

p5.Font all alignments testing (textAlignmentSketch in p5.Font/custom):

image

Re: the system font alignment question.  Nothing changes with this PR when you render text the system font way (textAllAlignmentsSketch in p5.Font/system).  For good measure, the right side shows the system fonts loaded via p5.Font after the PR (by using the ttfs).

image

dhowe:
Excellent work

mikewesthad:
Thanks

lmccart:
wonderful!

PR #1428, opened 2016-05-18
Improve docs load speed

limzykenneth:
My attempt at improving load speed of the docs. This is still ongoing but I'm filing this PR just to make sure the changes I made is compatible with the website itself. Also for other people to chip in ideas. So don't merge this yet 😉


docs/yuidoc-p5-theme/layouts/main.handlebars
```diff
@@ -38,16 +38,22 @@
     <!--<script src="//ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>-->
     <script>window.jQuery || document.write('{{p5SiteRoot}}/js/vendor/jquery-1.9.1.min.js"><\/script>')</script>
 
+    <!-- underscore-->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3/underscore-min.js"></script>
+
+    <!-- backbone-->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/backbone.js/1.3.3/backbone-min.js"></script>
```
limzykenneth: @lmccart Would something like this work with the website or would it create duplicates?

limzykenneth:
@lmccart Actually if everything looks ok with this feel free to merge it. I was thinking about more optimizations but changing structure of data.json might be too hard right now and gziping the files should be done on the backend, I would prefer to wait for the nodejs version of the website to be ready before doing that.

kadamwhite:
@limzykenneth +1, from the diff it looks like these changes will be useful regardless.

May I ask the naive question of why we're using an ajax-loading theme rather than trying to use one of the yuidoc options that compiles static HTML?

limzykenneth:
@kadamwhite Not entirely sure I understand your question...

Changes I made are all quite simple, first to simply minify data.json and use minified versions of the libs; second is to remove underscore and backbone from requirejs build step and request them on page load, that way reference.js will be smaller and underscore and backbone can be downloaded in parallel.

kadamwhite:
@limzykenneth My Q should have been asked in #143, as it is unrelated to your work; I get the changes you're proposing :) Sorry for the confusion!

limzykenneth:
@kadamwhite Oh yeah, I see, misunderstanding on my part.

limzykenneth:
@lmccart Actually this is ready to merge whenever. I'm working on some ideas for offline docs at the moment and not really changing anything here anymore.

lmccart:
This looks great, thanks @limzykenneth! 
@kadamwhite oh, I didn't even realize that yuidoc could compile to static html pages. Maybe this is the right next step for further improvements.

PR #938, opened 2015-09-23
Remove uses of Array.prototype.slice.call(arguments) in order to take…

plural:
… advantage of more JS compiler optimizations.  See https://dockyard.com/blog/2014/09/22/javascript-performance-for-the-win for some more details.

My methodology here was:
 - run some of my own sketches and p5.js examples
 - record profiles in the chrome developer tools
 - look for the little yellow "Optimization disabled" warnings.

The tests all still pass and things look otherwise OK.

I'd *really* prefer to have some sort of macro expansion for taking care of this, but i am not super comfortable with grunt and its workflows, so I did the straightforward literal thing in the pile of files that had these warnings.

PR #1431, opened 2016-05-21
bugs fixed for lights, material, and geometry

karenpeng:
-bugs fixed:
 - color parameters parsing in light
 - refine sphere and torus normals


indefinit:
Hey @karenpeng thanks for submitting this.  Going to make some comments inline.  Generally speaking, I'm wondering if we can avoid splitting our modelview matrix into two matrices as it adds a bit more overhead on both the CPU and shader sides.  Also, perhaps we can try to emulate the existing Processing api here:  https://processing.org/reference/beginCamera_.html

Lastly, if you could provide more info in your commit message so I can better understand what the bugs are being fixed that would be very helpful!  Perhaps a link to a test of some sort?

src/webgl/light.js
```diff
@@ -150,16 +150,17 @@ p5.prototype.directionalLight = function(v1, v2, v3, a, x, y, z) {
 
   var _x, _y, _z;
 
-  if(typeof arguments[arguments.length-1] === 'number'){
-    _x = arguments[arguments.length-3];
-    _y = arguments[arguments.length-2];
-    _z = arguments[arguments.length-1];
+  var argus = Array.prototype.slice.call(arguments);
```
indefinit: check out primitives.js for naming convention regarding arguments to Array.  variable should be var args instead of argus

src/webgl/p5.RendererGL.js
```diff
@@ -44,7 +44,8 @@ p5.RendererGL = function(elt, pInst, isMainCanvas) {
    * model view, projection, & normal
    * matrices
    */
-  this.uMVMatrix = new p5.Matrix();
+  this.uVMatrix = new p5.Matrix();
```
indefinit: Explained more in the general comment, but I'd prefer not to have separate view matrices and model matrices if possible.

src/webgl/p5.RendererGL.js
```diff
@@ -478,26 +488,36 @@ p5.RendererGL.prototype.rotateZ = function(rad) {
...
 
 p5.RendererGL.prototype.resetMatrix = function() {
-  this.uMVMatrix = p5.Matrix.identity();
+  this.uMMatrix = p5.Matrix.identity();
+  this.uVMatrix = p5.Matrix.identity();
   this.translate(0, 0, -800);
   return this;
 };
 
+p5.RendererGL.prototype.cameraTranslate = function(x, y, z) {
```
indefinit: perhaps we can adopt an approach similar to: https://processing.org/reference/beginCamera_.html where translations and rotations are simply applied to a matrix in our MVMatrix stack?

src/webgl/material.js
```diff
@@ -178,7 +178,7 @@ p5.prototype.ambientMaterial = function(v1, v2, v3, a) {
   gl.useProgram(shaderProgram);
   shaderProgram.uMaterialColor = gl.getUniformLocation(
     shaderProgram, 'uMaterialColor' );
-  var colors = this._renderer._applyColorBlend(v1,v2,v3,a);
+  var colors = this._renderer._applyColorBlend.apply(this._renderer, arguments);
```
indefinit: I'm not sure if I see where the bug is here.  apply is called on the current renderer context within _applyColorBlend.  If we want to call apply outside of the _applyColorBlend function, we should rename _applyColorBlend to _blendColor and tweak that function so as not to apply separate context inside.

karenpeng: @indefinit, for example `specularMaterial(250)`: `v1, v2, v3, a` is parsed in, the arguments in `_applyColorBlend` will become `[250, undefined, undefined, undefined]` instead of `[250]`. It will be translated to be red, instead of gray.
That's why we are seeing a red color in here http://p5js.org/examples/examples/3D_Multiple_Lights.php.

src/webgl/p5.RendererGL.js
karenpeng: I'll explain it in more details later. I think it could be a great conversation.

src/webgl/material.js
indefinit: @karenpeng Sorry for the delay getting back to you on this.  Yes, good point.  Could you pull this bug fix into a separate PR and please submit?  I think the camera changes are a larger conversation, but you're 100% right about the _applyColorBlend bug and I'd like to get this patched ASAP.  Thanks!

karenpeng:
@indefinit, 
Here's the modified diff with just the bugs fixed, the separation of model and view matrix is removed.

indefinit:
Hi @karenpeng thanks for making the changes.  This is almost there!  Only comment now is please replace calls to Array.prototype.slice with the original:
```
   args[i] = arguments[i];
}
```
The latter is more performant than Array.prototype.slice.

karenpeng:
@indefinit, [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) suggests this:
Using slice on arguments prevents optimizations in some JavaScript engines (V8 for example - more information). If you care for them, try constructing a new array by iterating through the arguments object instead. An alternative would be to use the despised Array constructor as a function:
```js
var args = (arguments.length === 1 ? [arguments[0]] : Array.apply(null, arguments));
```
How about we go for that? It's more clean and readable.

indefinit:
I think we should follow the general conventions set in p5 already.  Here's a PR from last year discussing arguments optimizations: https://github.com/processing/p5.js/pull/938

Ultimately we ended up with the existing args approach with for loop instead of array.prototype.slice

karenpeng:
@indefinit , ahh thanks for pointing out the conversation. Fixed it.

lmccart:
thanks @karenpeng! @indefinit feel free to merge this whenever you feel it's ready.

PR #1432, opened 2016-05-21
use saveJSON() in its example

jvolker:
The example of saveJSON() was using saveJSONObject() instead of the saveJSON().


lmccart:
thanks!

PR #1439, opened 2016-05-27
Fix rect() argument handling when rectMode === CORNERS

### LINKED ISSUE #1438: rectMode(CORNERS) doesn't work for negative points
Check the output of this:

```
rectMode(CORNERS);  
fill(100);  // Set fill to gray
rect(25, 25, -50, -50); 
```

It seems that the code is taking the absolute value of the third and fourth arguments, which is correct for rectMode(CORNER) but not for this mode.

joecridge:
From what I understand of WebGL it doesn’t care about `rectMode` so [line 437](https://github.com/joecridge/p5.js/blob/120e85cb682e16b4a64f53f9885c94cf70481006/src/core/2d_primitives.js#L437) fine as is, but let me know if this is wrong...


lmccart:
looks right, thanks!

PR #1440, opened 2016-05-29
Fix ellipsoid crashing

limzykenneth:
Fix [#254](https://github.com/processing/p5.js-website/issues/254) of p5.js-website

### DIFF
--- a/src/webgl/primitives.js
+++ b/src/webgl/primitives.js
@@ -428,10 +428,12 @@ p5.prototype.cone = function(){
  * @param  {Number} radiusx           xradius of circle
  * @param  {Number} radiusy           yradius of circle
  * @param  {Number} radiusz           zradius of circle
- * @param  {Number} [detail]          number of segments,
+ * @param  {Number} [detailX]         optional: number of segments,
  *                                    the more segments the smoother geometry
- *                                    default is 24. Avoid detail number above
- *                                    150. It may crash the browser.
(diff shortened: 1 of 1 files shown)


chiunhau:
Hi @limzykenneth 
I think it's just caused by YUIDoc's markdown-like parser (if it exist).
Because the 3rd line of `[detail]`  coincidentally start with `150. `, which is a markdown syntax. Generally it will be compiled into a list item (but in this case it's a code block hmmm don't know why).



I discovered that, delete the period of `150. ` and run yuidoc again, it's well formatted and everything is just fine. See:

limzykenneth:
@chiunhau It's not to do with those actually, it's just the code for ellipsoid is trying to read the third argument as the X detail level where it should have been the Z coordinate. When the detail level is too high as we have found previously is that it will either execute for very long or just give up and crash. That's why there's a warning on not using details level above 150.

chiunhau:
@limzykenneth ya ya i agree with your commits, they do really fix the bug that "cause crashing". Sorry for poor English. What I found and mentioned above is a common issue, it can be anywhere in this doc, not only for this case. 

Yes It's nothing to do with any programming issues, but will makes yuidoc wrongly compile markup. This can be avoided manually when editing docs though 😄

lmccart:
thanks @limzykenneth! @chiunhau I will look into the misformatting, thanks.

PR #1442, opened 2016-05-30
Improve example for p5.prototype.focused

chiunhau:
The original example isn't wrong but a little bit confusing:
When we switch to other window, `focused` turns false and a red cross shows which overlaps the circle. But then if switch back, `focused` turns true again right. However, the red cross is still here, the only difference is, the green circle is now on top of red cross.

I think it's hard to notice the difference between two state for people, I just thought it's a bug when I first play with this example. So I try to improve it to be more logical and visual appealing.
slice1


lmccart:
looks great, this is more clear, thanks!

PR #1443, opened 2016-05-31
Perspective function bug fix

indefinit:
typo in boolean property prevented perspective camera from resetting projection matrix in webgl mode.


lmccart:
👍

PR #1456, opened 2016-06-09
add random(Array) to get a random element from an array of choices

### LINKED ISSUE #1331: Make it possible to pass an Array into random()
I'm not sure how frequently people want to choose a random item from a list--perhaps it's not uncommon, since it's the third example for [`random()`](http://p5js.org/reference/#/p5/random)--but I thought it might be neat for the function to accept an Array as its first argument, so that the third example could be rewritten like so:

```js
// Get a random element from an array
var words = [ "apple", "bear", "cat", "dog" ];
text(random(words),10,50);  // Displays one of the four words
```

Previously this might have been confusing to document, but now that we can [document overloaded methods nicely](https://github.com/processing/p5.js/pull/1295), it might not be so bad.

Anyways, just a thought. Happy to submit a PR if others like the idea.

nthitz:
Apologies if I'm stepping on toes but I think this addresses the desired outcome in #1331. Was a little unsure if I got the inline documentation correct (and may make sense to remove the previous array example?) but things looked good locally.

### DIFF
--- a/src/math/random.js
+++ b/src/math/random.js
@@ -88,7 +88,7 @@ p5.prototype.randomSeed = function(seed) {
  * @method random
  * @param  {Number} [min]   the lower bound (inclusive)
  * @param  {Number} [max]   the upper bound (exclusive)
- * @return {Number|Object} the random number or a random element in choices
+ * @return {Number|mixed} the random number or a random element in choices
  * @example
  * <div>
  * <code>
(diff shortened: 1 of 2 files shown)


limzykenneth:
I would not say ` @return {Object} the random element from the array` since object have special meaning in javascript (or most programming languages really) maybe element? Not sure, need to dig around existing docs and see what the conventions are.

nthitz:
Yeah that's a good point. Possibly could also be `@return {mixed}` or `@return {any}` maybe? Given that the array could contain any type, the return value from this function could be pretty much anything. Happy to change that documentation to whatever makes the most sense.

toolness:
Thanks @nthitz ! I agree that it would make sense to remove the previous array example.

Regarding the use of the word "element", it does seem like MDN uses this word in the way @nthitz is using it, e.g. in the docs for [`indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf). But I agree that some kind of content style guide might be useful here so that all the docs speak in the same voice.

kadamwhite:
Not to throw a shoe into the works, but other libraries usually call this function "sample" or some such (e.g. lodash) and I wonder whether making a new function that does this by wrapping `random` rather than overloading the existing function might be less confusing.

toolness:
Thanks @kadamwhite !  We discussed it a bit in #1331 but it was decided to overload `random()`.  I don't personally know what the best solution is, but it would be interesting to somehow observe new programmers and see what they found easiest.

Personally I think the function should have the name "random" *somewhere* in it, so that someone quickly scanning the API would know that it involves randomness. I personally found lodash's use of the name `sample()` confusing, and had to read the documentation to learn what it actually did, but that may just be me!

kadamwhite:
@toolness makes sense! I'd missed that discussion, I appreciate the clarification and do not object to the logic :)

nthitz:
Ok my current plans are to remove the old Array example and change the documentation to `@return {mixed}`. I think the `{}` are supposed to be type annotations so `element` doesn't really indicate type to me. Not that this project uses flow, but [mixed or any](https://www.flowtype.org/docs/builtins.html) seem to be the recommended solution. (also hi @kadamwhite hope you got to play some pinball at eyeo!)

lmccart:
looks great, thanks @nthitz! I agree, let's keep random() for now and we can revisit if it seems to cause a lot of confusion.

PR #1457, opened 2016-06-11
Link to p5.dom/p5.sound docs from main reference

toolness:
This fixes https://github.com/processing/p5.js-website/issues/184 by adding a little note at the top of the main reference section:

### DIFF
--- a/docs/yuidoc-p5-theme-src/scripts/tpl/menu.html
+++ b/docs/yuidoc-p5-theme-src/scripts/tpl/menu.html
@@ -1,3 +1,10 @@
+<p>
+  <small>
+    Can't find what you're looking for? You may want to check out
+    <a href="#/libraries/p5.dom">p5.dom</a> or
+    <a href="#/libraries/p5.sound">p5.sound</a>.
+  </small>
+</p>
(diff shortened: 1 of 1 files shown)


docs/yuidoc-p5-theme-src/scripts/tpl/menu.html
```diff
@@ -1,3 +1,10 @@
+<p>
+  <small>
+    Can't find what you're looking for? You may want to check out
+    <a href="#/libraries/p5.dom">p5.dom</a> or
+    <a href="#/libraries/p5.sound">p5.sound</a>.
+  </small>
+</p>
```
toolness: I *think* that this `menu.html` template is only used from the main page of the reference documentation... let me know if I'm wrong though!

lmccart:
looks great, thanks!

docs/yuidoc-p5-theme-src/scripts/tpl/menu.html
lmccart: yes, this is just used on the main page, should be all good.

PR #1460, opened 2016-06-11
Rearrange if statement in _hsbaToRGBA, fixes #1459

### LINKED ISSUE #1459: Inconsistent colorMode behavior
Haven't tried other colorModes but I'm pretty sure this is a bug. Seems that if I pass 1000 the max for all HSB values, passing 1000 for the hue breaks things while passing 1000 for saturation/brightness gets me the max as I expect.
```
// Assume 1000 is the maximum allowed value (that is, inclusive) passed in to color() as docs suggest
colorMode(HSB, 1000, 1000, 1000, 1000);
color(0, 0, 0); // RGB is 0, 0, 0 as expected
color(0, 1000, 1000); // RGB is 255, 0, 0 as expected
color(999, 1000, 1000); // RGB is 255, 0, 1.53; ok
color(1000, 1000, 1000); // RGB is 255, 0, 255. Huh???
// So is 999 the true max allowed value?
color(0, 999, 999); // RGB is 254.745, 0.2547, 0.2547 so that's not right
```

joecridge:


kadamwhite:
Code looks great, haven't tested locally but +1 on the method of the fix

PR #1462, opened 2016-06-12
Remove side-effects in p5.Color toString method

### LINKED ISSUE #1461: p5.Color toString has side-effects on alpha value
The toString method of the p5.Color class changes the alpha value of the color:

p5.Color.prototype.toString = function() {
  var a = this.levels;
  a[3] = this._array[3];  // String representation uses normalized alpha.
  return 'rgba('+a[0]+','+a[1]+','+a[2]+','+ a[3] +')';
};

This should never be done, because it has side-effects when reusing a Color object (for example, to make a lerpColor calculation).

leonardo-fernandes:
Fixes #1461

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -53,8 +53,8 @@ p5.Color = function(renderer, vals) {
 
 p5.Color.prototype.toString = function() {
   var a = this.levels;
-  a[3] = this._array[3];  // String representation uses normalized alpha.
-  return 'rgba('+a[0]+','+a[1]+','+a[2]+','+ a[3] +')';
+  var alpha = this._array[3];  // String representation uses normalized alpha.
+  return 'rgba('+a[0]+','+a[1]+','+a[2]+','+ alpha +')';
 };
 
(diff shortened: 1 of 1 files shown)


src/color/p5.Color.js
```diff
@@ -53,8 +53,8 @@ p5.Color = function(renderer, vals) {
 
 p5.Color.prototype.toString = function() {
   var a = this.levels;
-  a[3] = this._array[3];  // String representation uses normalized alpha.
-  return 'rgba('+a[0]+','+a[1]+','+a[2]+','+ a[3] +')';
+  var alpha = this._array[3];  // String representation uses normalized alpha.
+  return 'rgba('+a[0]+','+a[1]+','+a[2]+','+ alpha +')';
```
kadamwhite: This is an awesome catch, thanks for the PR! I've added some unit tests for this behavior, feel free to pull them in from #1463 if you'd like otherwise the two PRs can merge in as siblings. This implementation looks good, 👍 !

PR #1463, opened 2016-06-12
Add unit tests for p5.Color.toString()

kadamwhite:
This PR adds unit tests for `p5.Color.prototype.toString()`, prompted by the bug identified by @leonardo-fernandes in #1462. The basic behavior of toString is also validated.


lmccart:
very helpful, thanks!

PR #1466, opened 2016-06-14
Added RADIANS condition for _angleMode in transform.js, rotate function.

danazkari:
Added a condition in src/core/transform.js to the rotate function to include the case where _angleMode is RADIANS.


lmccart:
looks great, thanks for the fast fix!

Issue #1269, opened 2016-02-23
touchX/touchY or mouseX/mouseY reporting wrong in iOS 7.0.x


noisefactor:
I believe I've discovered a bug that affects iOS (at least version 7.0.x) and possibly other mobile contexts.  When I try to use touchStart() (or mousePressed()) to get the current touchX or mouseX it reports the previous touch position instead of the current touch position.  

Tapping an iPad or iPhone like mine with this simple demo page shows the x and y coordinates reported in the title area.  (The problem doesn't occur in the desktop/laptop view of the page.)

   http://artcontext.org/act/96/draward/foo.html

function setup(){
   createCanvas(500,500);
}
function draw(){
	background("blue");
}
function touchStarted(){
     document.title=(touchX+" "+touchY);
}

hkirat:
you are using the wrong function..
It shouldn't work in any device whatsoever.
Use the touchEnded function instead of touchStarted function..
Should work fine.

noisefactor:
Perhaps some background would help.  My line drawing tools have init() and draw() methods.  What I do (that works for desktop and laptop modes of p5) is I use the mousePressed() function to trigger the init() method and this establishes the current, starting position of the line.  Subsequent mouseMove (drag) events then continue the line until the mouseReleased() event fires.

I think your interpretation of the problem is that I'm just trying to click someplace and find out where the click occurred.  But I am not seeking a position from a tap (click) event, really: I'm seeking the starting point of a new line (and I assume the touch event continues from the point of the tap).  I don't see how your suggestion could solve the problem that I'm having, or why the touchStarted() function should be returning stale positional data. I'm drawing lines and the problem that I have is that (only on iOS that I know of so far) both mousePressed and touchStarted fail to give me the actual position of the touch, which is needed by the line drawing that follows the initial touch.

noisefactor:
Adding the **bold** lines below to p5.js made things work the way I think they should.  It would appear someone thinks it's fine to wait until the draw() method updates these values, but I would tend to disagree with that strategy.  I don't fully understand the event model, but I do think it's logical that the touchX, touchY, mouseX, mouseY variables are updated when the ontouchstart function fires.

p5.prototype._ontouchstart = function(e) {
  var context = this._isGlobal ? window : this;
  var executeDefault;
  this._updateNextTouchCoords(e);
  this._updateNextMouseCoords(e);
  **this._updateMouseCoords(e);
  this._updateTouchCoords(e);**

[...]

Issue #1176, opened 2015-12-16
pmouseX always equal to mouseX


BarbaraAlmeida:
When using p5.js v0.4.20  from December 11, 2015 pmouseX/Y is always equal to mouseX/Y.

I've tried the p5.js files on [this issue] (https://github.com/processing/p5.js/issues/1035):

In p5.js v0.4.17 from October 25, 2015 (the one that solved the issue) they are also behaving weirdly: pmouseX/Y is not equal to mouseX/Y when the mouse stops moving (seems to be the same problem reported [here] (https://github.com/processing/p5.js/issues/1162)) and ```line(mouseX, mouseY, pmouseX, pmouseY);``` produces a dashed line.

It seems to work fine with p5.js v0.4.17 from October 13, 2015.

Code:

```javascript
function setup() {
  createCanvas(500, 500);
  background(240);
}


function draw() {  
  line(mouseX, mouseY, pmouseX, pmouseY);
  
  fill(255);
  rect(0, 0, 300, 50);
  
  fill(0);
  text("mouseX = " + mouseX, 30, 20);
  text("pmouseX = " + pmouseX, 30, 35);
  
  text("mouseY = " + mouseY, 180, 20);
  text("pmouseY = " + pmouseY, 180, 35);
  
  if (mouseX == pmouseX) {
    ellipse(10, 20, 10, 10);
  }
  
   if (mouseX == pmouseX) {
    ellipse(160, 20, 10, 10);
  }
}


function mousePressed() {
  background(240);
}
```

lmccart:
This is due to the fact that the mouse coords are update do mousemove, but the pmouse coords are updated each frame, so often the order of events causes them to end up the same. I would suggest instead doing the drawing inside mouseMoved() function:

```javascript

function mouseMoved() {
  console.log(mouseX, pmouseX, mouseY, pmouseY);
}
```

KingWalrus:
I recently reported an issue with pmouseX/Y not updating when the mouse stops moving, but in Processing the mouseX/Y positions update every frame (https://github.com/processing/p5.js/issues/1162), which was fixed by updating pmouseX/Y every frame, but now it seems that not updating current mouseX/Y every frame does not allow for pmousX/Y to be used in the main draw function, which is how a lot of people have been accustomed to using it. For example, the following code form Daniel Shiffman no longer works because of this functionality. 

````javascript
function setup() {
  createCanvas(600,600);
  background(255);
  
}

function draw() {
  stroke(0);
[...]

PR #1471, opened 2016-06-16
Change mouseX/mouseY handling

mattdesl:
Hello!

This changes the way mouse and touch values are stored to address the following two tickets:
https://github.com/processing/p5.js/issues/1176
https://github.com/processing/p5.js/issues/1269

Currently, `mouseX` and `mouseY` values are always lagging behind the real system values when the mouse is moving.

The new approach works like this:

- On mouse events, set the `mouseX` and `mouseY` values so that they represent the actual mouse value at that event
- After we `draw()` a frame, we set the `pmouseX` and `pmouseY` to the value of `mouseX` and `mouseY`, the next time the draw is called it will have the previous mouse position
- The first time we change `mouseX` and `mouseY`, we also set the `pmouseX/pmouseY` to the same value. This way they don't start at (0, 0) which is a bit arbitrary.
- Same logic applied for touch events and `pwinMouseX` / `pwinMouseY`

Some tests, I'm using the p5js editor.

## Example 1

```js
function setup() {
}

function draw () {
  line(mouseX, mouseY, pmouseX, pmouseY);
}
```

Before PR:  
screen shot 2016-06-16 at 12 03 32 pm

After PR:  
screen shot 2016-06-16 at 12 00 34 pm

## Example 2

```js
function setup() {
}

function mouseMoved() {
  ellipse(mouseX, mouseY, 50, 50);
}
```

Before PR:  
screen shot 2016-06-16 at 12 03 09 pm

After PR:  
screen shot 2016-06-16 at 12 00 51 pm

Let me know what you think! :smile:

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -345,9 +345,9 @@ var p5 = function(sketch, node, sync) {
       }
 
       this._setProperty('frameCount', this.frameCount + 1);
+      this.redraw();
       this._updateMouseCoords();
       this._updateTouchCoords();
-      this.redraw();
       this._frameRate = 1000.0/(now - this._lastFrameTime);
       this._lastFrameTime = now;
     }
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -13,15 +13,12 @@ var p5 = require('../core/core');
 var constants = require('../core/constants');
 
 /*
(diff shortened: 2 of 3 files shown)


lmccart:
this is great and makes a lot of sense, thank you!

PR #1484, opened 2016-06-23
fixes #1482

### LINKED ISSUE #1482: Inaccurate documentation for degrees()
The [documentation ](https://p5js.org/reference/#/p5/degrees) states

    println(rad + " radians is " + deg + " degrees");
    // Prints: 45 degrees is 0.7853981633974483 radians

I seriously doubt that println method would produce that output. :)

chiunhau:
Fixes inaccurate documentation for degrees():  #1482


limzykenneth:
Might be better if we change this line

```
println(rad + " radians is " + deg + " degrees");
```

to

```
println(deg + " degrees is " + rad + " radians");
```

instead and keep the comments as it is.

lmccart:
thanks!

Issue #1485, opened 2016-06-23
createCanvas() followed by get() in draw() hogs memory; crashes


hobzcalvin:
Consider this minimal example:
```
<html>
  <body>
    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.2/p5.js"></script>
    <script type="text/javascript">
      function setup() {
        createCanvas(600, 600);
      }
      function draw() {
        get();
      }
    </script>
  </body>
</html>
```
Running this in Chrome, I see memory usage climb as high as 1GB before dropping to a few hundred MB. Sometimes this crashes with:
`p5.js:15460 Uncaught RangeError: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': Out of memory at ImageData creation`
Other times it crashes the whole tab with the "Aw, Snap!" message, no console errors.
It crashes all the time on a Nexus 9 tablet; crashes sometimes on Windows; seems fine on a Mac or iPad. Apparently this is part luck of the draw, part memory limits for the particular browser/device.
Memory usage drops eventually, but it gets dangerously high beforehand.

hobzcalvin:
Sorry, I should've used a hosted p5.js:
`<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.2/p5.js"></script>`

I whittled down [a rather large p5js project](http://hobzcalvin.github.io/fadeclient/) with this issue into that minimal example on a Google Nexus 9, but now I can't reproduce it on my Windows machine. However, that large project now crashes on Windows Chrome with this exception (vastly preferable to the whole Chrome tab crashing):
`p5.js:15460 Uncaught RangeError: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': Out of memory at ImageData creation`

So, a memory leak, then?

limzykenneth:
I think it is sensible to advise the user against calling memory intensive operations on every frame of the draw calls instead of quietly handling it in the background. There are potentially numerous ways to create memory leaks and to try and account for them all is futile, awareness on the user side, especially the ability to diagnose it is probably more helpful. That's what I think anyway...

hobzcalvin:
Well, as you may see from the feed, I tried [a commit](https://github.com/hobzcalvin/fadeclient/commit/07c9b26de5c3c263411ea221c26e17e3308da805) where I removed the get()/image() calls from my draw() function. But it still crashes on a Nexus 9!

[...]

PR #1488, opened 2016-06-24
Streamline Renderer2D.get()

hobzcalvin:
I'm new here! But this seemed like a pretty obvious improvement.

Don't call loadPixels(); we don't need it.
When getting a single pixel value, just use getImageData with
width/height 1.

Tests pass, and there's a get() test, but it seems to never run??
I could never get it to fail, so there's something not right there.
I tested with a local sketch and things seem fine.

An attempt to help with:
https://github.com/processing/p5.js/issues/1485


lmccart:
this looks good to me, thanks!

PR #1494, opened 2016-06-29
Fix #1479 (extend support for windowWidth and windowHeight)

### LINKED ISSUE #1479: windowHeight returns 0 in Firefox
Using the basic example:  
`console.log(windowHeight);` returns 0 in Mozilla Firefox 47.0 (tested with p5.js v0.5.2 June 17, 2016)  
So does `document.documentElement.clientHeight`

Although `console.log(window.innerHeight);` returns an expected value.

nok:
I guess that the [onresize](https://github.com/processing/p5.js/blob/master/src/core/environment.js#L342-L353) event doesn't be triggered initially. Because of that the default values for [windowWidth](https://github.com/processing/p5.js/blob/master/src/core/environment.js#L308) and [windowHeight](https://github.com/processing/p5.js/blob/master/src/core/environment.js#L319) are based on the `document.documentElement.clientHeight` and `document.documentElement.clientWidth` values (≠ `window.innerHeight` and `window.innerWidth`). This pull request fixes this behaviour.

### DIFF
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -305,7 +305,7 @@ p5.prototype.displayHeight = screen.height;
  * createCanvas(windowWidth, windowHeight);
  * </code></div>
  */
-p5.prototype.windowWidth = document.documentElement.clientWidth;
+p5.prototype.windowWidth = getWindowWidth();
 /**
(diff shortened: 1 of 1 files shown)


lmccart:
thanks!

Issue #1389, opened 2016-05-04
Minor feature request - redraw()


emilyxxie:
I often find myself writing:
```
for (var i = 0; i <= n; i++) {
   redraw();
}
```
....for _n_ times I'd like to see something be redrawn, either for debugging purposes, or as part of the final sketch. Not sure if anyone else has adopted this habit, but if it's a popular enough pattern, I think it'd be great to be able to optionally pass in a positive integer argument into redraw() and only redraw for that given number of times----provided that such a feature does not go against any best practices.

I would not mind trying my hand at implementing this small feature myself either. Seems like a fun and quick way to start getting involved in contributing to the source code.

limzykenneth:
@emilyxxie I would say go on ahead and implement this yourself! Might even be easier for us to see how useful for most people this could be once it is implemented. Cheers.

emilyxxie:
Awesome---thanks. Just saw this response. I'll dive into implementing this as soon as I square away another project. Excited about this!

nok:
I like your idea @emilyxxie, because with this feature the developers have more flexibility and control about changing variables. So I created this minor feature for us in PR #1498. Again, great idea and request!

---

There is still a failed test with the exception "code took too long to execute". On my local machine all tests passed successfully. I will check it.

lmccart:
closed with https://github.com/processing/p5.js/pull/1498

PR #1498, opened 2016-07-01
Add param to redraw() to control the number of redraws #1389

nok:
I added a minor feature related to the idea or request of @emilyxxie in issue https://github.com/processing/p5.js/issues/1389.

By the way I fixed the indentations of the prior example code:

screen shot 2016-07-02 at 01 05 10

### DIFF
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -257,8 +257,12 @@ p5.prototype.popStyle = function() {
  * <br><br>
  * The redraw() function does not work properly when called inside draw().
  * To enable/disable animations, use loop() and noLoop().
+ * <br><br>
+ * In addition you can set the number of redraws per method call. Just
+ * add an integer as single parameter for the number of redraws.
  *
  * @method redraw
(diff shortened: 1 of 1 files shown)


nok:
There is still a failed test with the exception "code took too long to execute". On my local machine all tests passed successfully. I will check it.

limzykenneth:
Probably just need to rerun the test on travis

lmccart:
looks good, thanks for this. two minor suggestions for the documentation before we merge this in.
1. rather than just updating the example, let's include both examples, as they demonstrate the two different uses of the  function.
2. could you add a note in the function description that explains this optional use, and add an optional parameter to the method documentation. you can read more about inline documentation here: https://github.com/lmccart/p5.js/wiki/inline-documentation, or see [this example](https://github.com/processing/p5.js/blob/master/src/core/p5.Element.js#L117) for showing optional arguments.

nok:
Thanks for the notes, @lmccart. I hope the documentation is comprehensible.

lmccart:
looks great, thanks!

Issue #1481, opened 2016-06-22
Is there any problems in preload?


mikashin:
Hi!
I used preload() to load my table in PC. But it didn't work.
So I tried the example code in your website (http://p5js.org/reference/#/p5/loadTable ) and found that preload() doesn't work. It caused  "22462: Uncaught TypeError: object is not a function"  error.  
So I ran a script from the example code as below.

```javascript
var table;

//function preload(){
// table=loadTable("http://p5js.org/reference/assets/mammals.csv", "csv", "header");
//}

function setup() {
  a=1;
  print(a);
  table=loadTable("http://p5js.org/reference/assets/mammals.csv", "csv", "header");
  print(table.getColumn("name"));
  print("hi");
  
  for (var r = 0; r < table.getRowCount(); r++)
    for (var c = 0; c < table.getColumnCount(); c++) {
      print(table.getString(r, c));
    }
  
 
}
```

The outcome was
1
[]
"hi"

As you see, the example data didn't come out.
I think preload() and loading components in table doesn't work properly. Could you confirm this?

lmccart:
Preload seems to be working for me. The code above as printed gives the output because you are trying to print parts of the table before the table has fully loaded. This is the code that I'm using. I'm working locally and have downloaded the mammals.csv file locally as well and placed in a folder called assets.

```javascript
var table;

function preload(){
  table=loadTable("assets/mammals.csv", "csv", "header");
}

function setup() {
  print(table.getColumn("name"));
  for (var r = 0; r < table.getRowCount(); r++)
    for (var c = 0; c < table.getColumnCount(); c++) {
      print(table.getString(r, c));
    }
}
```

outputs:
```
["Goat", "Leopard", "Zebra"]
0
Capra hircus
Goat
1
Panthera pardus
Leopard
2
Equus zebra
Zebra
```

toolness:
@mikashin, one thing I've noticed about your code is that it's attempting to retrieve http://p5js.org/reference/assets/mammals.csv, which will fail because p5js.org isn't serving those files with [Cross-Origin Resource Sharing (CORS)](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) headers. This is part of why @lmccart downloaded the file locally and her code retrieves it via a relative URL... It's probably also why your original attempt at using `loadTable()` within `preload()` didn't work.

[...]

PR #1500, opened 2016-07-03
When loadTable() errors, call callback if and only if it's a function

toolness:
In #1481, @mikashin was understandably confused when their `loadTable()` call failed with the error `22462: Uncaught TypeError: object is not a function`.

This is actually a bug in p5 and it's happening because the code in `loadTable` that checks to see if a callback was passed-in is checking to see if it's not `undefined`; however, internally the function defaults the value to `null`, which isn't the same thing as `undefined`, so the browser was trying to call the null callback and failing.

This PR fixes the bug by checking to see if `callback` is specifically a function. It also improves the documentation by documenting what's passed in to the callback if an error occurred.


lmccart:
good find, thanks for the fix!

PR #1503, opened 2016-07-05
Added removeAttribute() function. Should fix #1499

### LINKED ISSUE #1499: Failing to enable a DOM button after disabling it with "disabled" attribute
Hi!

I'm having some problems with the `disabled` attribute because I can't find a way of removing it within p5js.

I [can see](http://www.w3schools.com/tags/att_input_disabled.asp) that this attribute is boolean and acts when it is present, regardless of its value.

Using p5js, I can create and disable a button as follows:
```javascript
var button = createButton('Press me');
button.attribute('disabled','true');
```

Something funny is that in the previous snippet, `'true'` could actually be any other word (_including **false** and an empty string_) as the browser apparently will only care about the presence of the `disabled` attribute (I'm using Google Chrome in Ubuntu 16.04, but I guess that's the normal behavior). 

Based on this, apparently there's no way of re-enabling the button using the `attribute` function. Please correct me if I'm wrong or if there's another way.

sejomagno:
This is a simple wrapper for the [HTML DOM removeAttribute method](http://www.w3schools.com/jsref/met_element_removeattribute.asp). Apparently, using this method would be the only way of removing the [`disabled` attribute](http://www.w3schools.com/tags/att_input_disabled.asp) to re-enable an input.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1385,6 +1385,21 @@
       return this;
     }
   };
+  
+  
+  /**
+   *
+   * Removes an attibute on the specified element.
+   *
+   * @method removeAttribute
+   * @param  {String} attr       attribute to remove
+   * @return {Object/p5.Element} 
+   *                             
+   */
+  p5.Element.prototype.removeAttribute = function(attr) {
+    this.elt.removeAttribute(attr);
+    return this;
+  };  
 
 
   /**


lmccart:
looks great, thanks!

toolness:
Yay, congrats on your [first PR](https://twitter.com/p5xjs/status/750628291237728257)! 🎉 🎉 🎉 (And thanks for the contribution!)

PR #1510, opened 2016-07-06
Define param as optional

nok:
Currently I'm developing a library for p5.js. For that I tested different kinds of instantiation. Thereby I noticed, that the second parameter `node` of the method `window.p5()` is optional. So I changed the internal comments to improve the documentation quality. In fact it's a small detail.

Should we extend the current [example](https://p5js.org/examples/examples/Instance_Mode_Instantiation.php) with these examples?

Here you see the tested variants:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title></title>
  <script language="javascript" type="text/javascript" src="https://ashleyjamesbrown.com/p5js/p5.js"></script>
</head>
<body>
  <script>
  var sketch = function(p) {
    p.setup = function(){
      p.createCanvas(100, 100);
      p.background(0);
    }
  };
  new window.p5(sketch);  // <--- That works.
  </script>
</body>
</html>
```

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title></title>
  <script language="javascript" type="text/javascript" src="https://ashleyjamesbrown.com/p5js/p5.js"></script>
</head>
<body>
  <div id="container"></div>
  <script>
  var sketch = function(p) {
    p.setup = function(){
      p.createCanvas(100, 100);
      p.background(0);
    }
  };
  new window.p5(sketch, window.document.getElementById('container'));  // <--- That works.
  </script>
</body>
</html>
```

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title></title>
  <script language="javascript" type="text/javascript" src="https://ashleyjamesbrown.com/p5js/p5.js"></script>
</head>
<body>
  <script>
  var sketch = function(p) {
    p.setup = function(){
      p.createCanvas(100, 100);
      p.background(0);
    }
  };
  var node = document.createElement("div");
  // console.log('node: ', typeof(node), node.__proto__, node);
  // HTMLDivElement <- HTMLElement
  window.document.getElementsByTagName('body')[0].appendChild(node);
  new window.p5(sketch, node);  // <--- That works.
  </script>
</body>
</html>
```

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title></title>
  <script language="javascript" type="text/javascript" src="https://ashleyjamesbrown.com/p5js/p5.js"></script>
</head>
<body>
  <script>
  var sketch = function(p) {
    p.setup = function(){
      p.createCanvas(100, 100);
      p.background(0);
    }
  };
  var node = document.createElement("div");
  // console.log('node: ', typeof(node), node.__proto__, node);
  // HTMLDivElement <- HTMLElement
  new window.p5(sketch, node);  // <--- Logically that works as well.
  window.document.getElementsByTagName('body')[0].appendChild(node);
  </script>
</body>
</html>
```

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -29,8 +29,8 @@ var constants = require('./constants');
  * @param  {Function}    sketch a closure that can set optional preload(),
  *                              setup(), and/or draw() properties on the
  *                              given p5 instance
- * @param  {HTMLElement|boolean} node element to attach canvas to, if a
- *                                    boolean is passed in use it as sync
+ * @param  {HTMLElement|boolean} [node] element to attach canvas to, if a
+ *                                      boolean is passed in use it as sync
  * @param  {boolean}     [sync] start synchronously (optional)
  * @return {p5}                 a p5 instance
  */


lmccart:
This is correct, thank you. I think it would be great to add more examples to the examples page under hte heading "Instance Mode". If you'd be willing to add these, the instructions are here: https://github.com/processing/p5.js-website/wiki/Adding-examples

nok:
I would like to but I don't want to change the general html template. The examples show either how to specify a `div` where the canvas will be created or how to create a new `div` which will be append to the Document Object Model. 

> Node is the tool we use to automatically build the examples page and individual pages for each example.

Can we add a flag (e.g. `@skip`) to skip the HTML generation for the current example? Then we can create unusual or special examples by hand. What do you think?

lmccart:
yep definitely. we're currently in the process of rebuilding the website to support internationalization (https://github.com/mayaman/p5js-website) which should launch sometime this summer, so I might just add these in along with the `@skip` flag myself just because I'm worried if we merge it into the current website repo it could get lost in the mix. so I'll go ahead with this if it sounds ok?

nok:
Sure, step for step, conflicts are a nightmare. Maybe there is a better solution for that issue. Nevertheless it makes fun to contribute, thanks @lmccart. 👍

lmccart:
hey @nok I added support for a @norender tag and HTML code in the examples, and your examples have been added here: http://staging.p5js.org/examples/instance-mode-instance-container.html. they'll be visible on the main site as soon as we launch (soon).

nok:
Looks great! By the way should we add the basic tags to provide support for _older_ browsers?

```html
<meta charset="utf-8">
<meta http-equiv="x-ua-compatible" content="ie=edge">
```

The first line specifies the character encoding for the HTML document in general.

The second line will force the IE to use the latest mode as possible to support the generated `canvas` tag and users don't need to change the compatibility mode from developer tools. 

But maybe it's the wrong place for it, I prefer more the clear examples. So you can close that issue. 👍

Issue #1514, opened 2016-07-09
Unable to set Radio value to 0


KyleNehman:
Some example code:
```
function setup() {

  createCanvas(windowWidth, windowHeight);

  var radioOptions = ['radio', 'button', 'test', 'values'];
  var thisRadio = createRadio();
  var x = 100;
  var y = 100;

  for (var i = 0; i < radioOptions.length; i++) {
    var option = radioOptions[i];

    thisRadio.option(option, i);
  }
  thisRadio.position(x, y);

  thisRadio.value(0);
}

```

If the radio buttons' values are set to ints [0, n-1] calling thisRadio.value(0) does nothing. Calling the same function with arguments [1, n-1] works as intended and allows me to set any button besides the first one to be filled.

lmccart:
thanks for the report, fixed with https://github.com/processing/p5.js/pull/1515, it will be in the next release

PR #1515, opened 2016-07-10
Fix value() index bug (#1514)

nok:
Related issue: https://github.com/processing/p5.js/issues/1514

Short description: The number `0` as value in an if-condition will be interpreted as `false`. Because of that nothing happens. Instead the fix uses the length of passed arguments.


lmccart:
perfect, thanks!

Issue #1414, opened 2016-05-13
Issues with WebGL 2d Primitives


mindofmatthew:
I've noticed that the y-axis currently points in opposite directions between the 2d canvas and WebGL modes. Should we flip the y-axis in the WebGL renderer?

This is how Processing works, and I think that this will make a lot of sense as more 2d functions are implemented in WebGL mode. However this is different than the way that many 3d tools/libraries work, as discussed in #820.

indefinit:
Hey @mindofmatthew, do you mean positioning with regards to the 2d primitive functions in webgl mode?  We should have parity with Processing's P3D.

mindofmatthew:
@indefinit Yes. I suppose I assumed that there was a larger issue that impacted translations and everything, but that's clearly not the case. Renamed the issue accordingly.

Here are some specific things I found:

- WebGL quad() and rect() produce shapes that are flipped vertically compared with those same shape calls in 2d/Processing
- x and y parameters don't seem to have any bearing on the position of rect()
- Overall, all shapes are slightly smaller in WebGL then they ought to be

indefinit:
Hey, sorry I let this one slip.  Let's make sure the 2d primitives operate the same as they do in 2d canvas.  

Re: 
* Overall, all shapes are slightly smaller in WebGL then they ought to be

I think this might have to do with the initial model view matrix.  Definitely a worthwhile endeavor to overhaul our camera controls, but probably a separate issue to the primitives axis.

lmccart:
fixed with https://github.com/processing/p5.js/pull/1518

PR #1518, opened 2016-07-12
Light 2d primitives refactor; addresses #1414

indefinit:
- Refactors and removes unnecessary duplicate code between canvas and webgl modes.  -- 
- Implements all rect* and ellipse* modes in webgl.
- drawing 2d primitives in webgl mode follows same api as canvas mode and Processing P3D.  **Warning: potential breaking change for anyone using x,y,z coords in their webgl 2d primitives**
Ultimately, the removal of z coords in 2d primitives will make for less buggy, cross renderer-ready sketches.


lmccart:
looks great!

Issue #1512, opened 2016-07-07
random() and sin() significantly slower compared to Math.random() and Math.sin()


ofZach:
Test code: 

    // performance test ("random()")
    start = new Date();
    randomTest = 0;
    for (i = 0; i < 100000; i++){
        randomTest += random();
    }
    end = new Date();
    millisecondsElapsed = end - start;
    console.log("random() : " + millisecondsElapsed);
    
    // Math.random();
    start = new Date();
    randomTest = 0;
    for (i = 0; i < 100000; i++){
        randomTest += Math.random();
    }
    end = new Date();
    millisecondsElapsed = end - start;
    console.log("Math.random() : " + millisecondsElapsed);
    
    // sin();
    start = new Date();
    randomTest = 0;
    for (i = 0; i < 100000; i++){
        randomTest += sin(i);
    }
    end = new Date();
    millisecondsElapsed = end - start;
    console.log("sin() : " + millisecondsElapsed);
    
    // Math.sin();
    start = new Date();
    randomTest = 0;
    for (i = 0; i < 100000; i++){
        randomTest += Math.sin(i);
    }
    end = new Date();
    millisecondsElapsed = end - start;
    console.log("Math.sin() : " + millisecondsElapsed);
    

I find about a 10x speed improvement using Math.random() and Math.sin()

ofZach:
note @shiffman may have more insight here, we were going back and forth about this on the SFPC slack -- since we are doing some pixel operations in class the slowness of these functions was really felt (ie, it's the difference between 2 fps and 30 fps  on an updating image of static noise, etc.

limzykenneth:
Very interesting problem! I tried to isolate the problem but just couldn't as it seem the very act of assigning `random()` to map to `Math.random()` is where the performance hit occurs. I suspect the way the browser optimise these things could have something to do with it and here's my findings:

On Safari:
![screen shot 2016-07-07 at 12 53 06 pm](https://cloud.githubusercontent.com/assets/7543950/16652335/c5a93ac0-4441-11e6-8dd2-cc880e87bec4.png)
~10 times the difference

On Firefox:
![screen shot 2016-07-07 at 12 53 57 pm](https://cloud.githubusercontent.com/assets/7543950/16652351/e319d1a0-4441-11e6-9cc0-083f0062ed53.png)
Erm...pretty bad...

On Chrome:
[...]

PR #1519, opened 2016-07-12
Avoid .bind when assigning some methods to global object

kadamwhite:
Per #1512, there are significant slowdowns when using certain methods in global mode, such as `random()`. Part of the origin of this performance issue is that [methods are .bound to the p5 instance when being assigned to the global object](https://github.com/processing/p5.js/blob/d0cf66e2d2de3d26a2d3091cb0f0471a15675249/src/core/core.js#L467), and the use of the .bind (which is not needed for a pure function like `random`) impacts the method's speed. ([see explanation](http://stackoverflow.com/questions/17638305/why-is-bind-slower-than-a-closure))

This PR takes a crack at a solution proposed in #1512, specifically decorating methods which can be assigned to the global context without the use of `.bind` with a `._context = false` property. Core.js is then modified to respect this flag by only using `.bind` if `method._context !== false`.

This could use some thorough testing, I've only spot-checked&mdash;and this currently only accounts for the math utilities in the Calculation, Noise, Random and Trigonometry files. The `_context` argument can likely be rolled out more broadly, across any method of `p5.prototype` which does not utilize `this`.

Unfortunately, most trigonometry functions cannot be decorated in this way because they depend on the instance property `_angleMode` in order to return the proper value.

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -28,6 +28,8 @@ var p5 = require('../core/core');
  * </code></div>
  */
 p5.prototype.abs = Math.abs;
+// Avoid `.bind`ing this method in global mode
+p5.prototype.abs._context = false;
 
 /**
  * Calculates the closest int value that is greater than or equal to the
@@ -64,6 +66,8 @@ p5.prototype.abs = Math.abs;
  * </code></div>
  */
 p5.prototype.ceil = Math.ceil;
+// Avoid `.bind`ing this method in global mode
+p5.prototype.ceil._context = false;
 
 /**
  * Constrains a value between a minimum and maximum value.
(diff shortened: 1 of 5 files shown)


src/math/calculation.js
```diff
@@ -28,6 +28,8 @@ var p5 = require('../core/core');
  * </code></div>
  */
 p5.prototype.abs = Math.abs;
+// Avoid `.bind`ing this method in global mode
+p5.prototype.abs._context = false;
```
kadamwhite: Before we commit to this approach I would like several people to thoroughly test whether `abs` in this branch is faster than `abs` in master (both in global mode), as well as the same test for a custom-implementation function like `random`.

kadamwhite:
After comparative testing in #1520 and further discussion in #1512 I am increasingly convinced the effects of the `.bind` are negligible _vs_ the object property getters used to set them on the global object, and in that context I actually think this approach adds unneeded complexity for little gain. But we should profile it to see.

lmccart:
hi @kadamwhite, are there any updates on this? or what's the next step here? it looks like maybe we want to set up a profile test and compare the effects of this across browsers?

lmccart:
I'm going to close this since it's been inactive for a while. Please reopen if there are further updates. Thanks!

PR #1520, opened 2016-07-12
Improve performance of Random method by not using "arguments"

kadamwhite:
In #1512 there's been a lot of discussion about the performance of various methods in P5, and what might contribute to them being faster or slower. While investigating, I did identify one major area of potential performance improvement on the random method specifically: the use of `arguments` significantly impacts the runtime speed of a method, and I recommend using the named arguments for the random method in lieu of the `arguments` keyword.

To test the hypothesis, I prepared this test harness:

http://bl.ocks.org/kadamwhite/raw/0e3240d7c475a91cb0fbbbc4a084e351/

using the code in [this gist](https://gist.github.com/kadamwhite/0e3240d7c475a91cb0fbbbc4a084e351)

If you visit that URL and open the console, you will see the browser run six tests:

1. The existing `random` implementation, called directly
2. The existing `random` implementation, called through Object lookup<sup>+</sup>
3. The existing `random` implementation, `.bind`ed to `window`
4. The new propose `random` implementation, called directly
5. The new propose `random` implementation, called through Object lookup<sup>+</sup>
6. The new propose `random` implementation, `.bind`ed to `window`

<sup>+</sup> There is additional overhead on this one of a second function being called; prior tests designed to avoid that caveat showed directionally similar results, so I opted in favor of code brevity

In Chrome, the avoidance of `arguments` drops the average run time (across many iterations) for the basic case (calling the method directly) from about 60ms to about 30ms, a 2x improvement.

Safari showed an improvement from about ~80ms to ~60ms, not as notable but still an improvement.

But on Firefox, the avoidance of `arguments` drops the average run time (for me, on OSX using Firefox ) from **250ms or more to less than 10ms**, over 40x faster.

This should be independently tested, but I propose the use of the named arguments is actually more readable than the use of the `arguments` keyword, and the performance seems to be dramatically improved through this approach.

Interestingly (given the discussion in #1512), the use of `.bind` had almost completely negligible effects on the function's execution time.

Sample output of script for Chrome:
```
(index):131 Average runtime for object lookup: 57.94ms
(index):131 Average runtime for bound function: 51.01ms
(index):131 Average runtime for basic without "arguments": 24.86ms
(index):131 Average runtime for object lookup without "arguments": 37.05ms
(index):131 Average runtime for bound function without "arguments": 24.94ms
```
Safari:
```
[Log] Average runtime for basic: 75.18ms
[Log] Average runtime for object lookup: 115.98ms
[Log] Average runtime for bound function: 79.85ms
[Log] Average runtime for basic without "arguments": 57.31ms
[Log] Average runtime for object lookup without "arguments": 92.23ms
[Log] Average runtime for bound function without "arguments": 57.37ms
```
Firefox:
```
Average runtime for basic: 248.97ms
Average runtime for object lookup: 246.36ms
Average runtime for bound function: 226.46ms
Average runtime for basic without "arguments": 5.89ms
Average runtime for object lookup without "arguments": 13.34ms
Average runtime for bound function without "arguments": 6.79ms
```

### DIFF
--- a/src/math/random.js
+++ b/src/math/random.js
@@ -131,13 +131,12 @@ p5.prototype.random = function (min, max) {
   } else {
     rand = Math.random();
   }
-
-  if (arguments.length === 0) {
+  if (typeof min === 'undefined') {
     return rand;
   } else
-  if (arguments.length === 1) {
-    if (arguments[0] instanceof Array) {
-      return arguments[0][Math.floor(rand * arguments[0].length)];
+  if (typeof max === 'undefined') {
+    if (min instanceof Array) {
+      return min[Math.floor(rand * min.length)];
     } else {
       return rand * min;
     }


nok:
This thread results in a similar result and recommendation: http://stackoverflow.com/a/13299598/1293700 
I am also for it, because the performance of the core library matters.
Here we find the relevant parts: https://github.com/processing/p5.js/search?utf8=%E2%9C%93&q=arguments

rwaldron:
@kadamwhite asked me to review this change, but since I am not an active participant I will keep my notes as general as possible.

src/math/random.js
```diff
@@ -131,13 +131,12 @@ p5.prototype.random = function (min, max) {
...
   }
-
-  if (arguments.length === 0) {
+  if (typeof min === 'undefined') {
     return rand;
   } else
-  if (arguments.length === 1) {
-    if (arguments[0] instanceof Array) {
-      return arguments[0][Math.floor(rand * arguments[0].length)];
+  if (typeof max === 'undefined') {
```
rwaldron: Will you ever need to guard against `null`?

rwaldron:
Two notes: 

1. `instanceof` will fail across realms, so if there is any possibility that `random(...)` might be called with input from an iframe or called in iframe with input from a parent frame, then it will fail. With a minor perf hit on check for the cross-realm arrays, changing the condition to the following will resolve the issue:

  ```js
  if (min instanceof Array || Array.isArray(min)) {
  ```
  ...However it might just be that this is a use case that's not intended to be supported, in which case: freely ignore :)

2. It might be worthwhile to add some kind of grunt-task benchmark testing, @shama's [`grunt-benchmark`](https://github.com/shama/grunt-benchmark) seems like a good place to start :)

rwaldron:
One more thing! Is your CI not hooked up to pull requests?

src/math/random.js
kadamwhite: @lmccart What're your thoughts, should we be doing anything preventative against `null` here? The old code did not do anything particular around null inputs, they would still have triggered the `arguments.length === 1` path.

That said, there's a difference between `arguments.length === 1` and `typeof secondArg === 'undefined'`; technically the former would be represented by `typeof firstArg !== 'undefined' && typeof secondArg === 'undefined'`, to capture the case where e.g. 

```js
var obj = {
  max: 7
};
random( obj.min, obj.max ); // obj.min -> undefined
```

I lean towards leaving things as they are for efficiency, since it's much more likely that `min` will be something non-mathematical that throws an error than that it will be a valid reference which is undefined, but I wanted to vet the importance of guarding against edge cases vs maintaining the existing code's efficiency.

kadamwhite:
100% in favor of wiring up grunt-benchmark; added a note to #1524.

I don't think the iframe realm question is cause for concern, but that's a @lmccart _et al_ question. Same with the final Q; I assume PR's be triggering CI, but don't have the full context there.

src/math/random.js
lmccart: I agree, in this case I think we can prioritize efficiency. Generally, I think we want to be taking measures to catch common errors, but don't need to safeguard against every possible one.

lmccart:
Yeah, I don't think iframe realm crossing is a major use case so this seems fine to me. PRs do trigger CI, but it takes a moment to run, so a PR may show up initially as mergeable if there are no conflicts and later display an error. (I'm pretty sure? Is there something you saw that seems like it's not working properly?)

kadamwhite:
@lmccart What more needs to be done here for this to be merge-able?

lmccart:
ack sorry, this one got lost in the mix of travels and emails. thanks for the bump, and for the PR!

PR #1526, opened 2016-07-14
Added example to removeAttribute()

sejomagno:
I added an example demonstrating how to enable/disable a button using the `changed()` method in a checkbox. It shows the maybe-non-intuitive way of disabling an input adding the `disabled` attribute without the need of arguments, and the maybe-less-intuitive way of re-enabling it by removing such attribute.


lmccart:
thanks!

Issue #1524, opened 2016-07-13
add performance testing and benchmarking


lmccart:
We need some way to test performance of various p5 functions, cross-browser and cross-platform to determine when optimizations are actually that, and when they add unnecessary complexity.
This is a follow up from #1512.

kadamwhite:
In #1520 @rwaldron proposed using [grunt-benchmark](https://github.com/shama/grunt-benchmark) for this. It seems we'd need to do a little extra wiring to execute them in multiple browsers but that's probably a good place to start.

Jared-Sprague:
@lmccart I would like to take this one.  Performance benchmarking get's me exited! :laughing: 

It sounds like the goal here is to have an npm task to benchmark performance optimizations when implementing them to know that they are helping.  I think it sounds like more of a dev task than part of the automated build.

If it doesn't need to run on every build, but a task that developers kick off to test their performance optimizations,  then I recommend using [karma](https://www.npmjs.com/package/karma), and [karma-benchmark](https://www.npmjs.com/package/karma-benchmark).

Karma lets you run code in multiple real browsers on your local machine,  and karma-benchmark uses [benchmark.js](https://www.npmjs.com/package/benchmark) (used by jsPerf) to execute and output benchmarks in multiple browsers.

Here is example output of a test I ran on my laptop:
```
Chrome 62.0.3202 (Linux)  Array iteration: util.each at 73838720 ops/sec
Chrome 62.0.3202 (Linux)  Array iteration: Array.forEach at 248844823 ops/sec
Chrome 62.0.3202 (Linux)  Array search: util.contains at 25346940 ops/sec
Chrome 62.0.3202 (Linux)  Array search: Array.indexOf at 12303891 ops/sec
Chrome 62.0.3202 (Linux)
  Array iteration: Array.forEach at 248844823 ops/sec (3.37x faster than util.each)
  Array search: util.contains at 25346940 ops/sec (2.06x faster than Array.indexOf)

Running "karma:firefox" (karma) task
WARN [karma]: Port 9876 in use
INFO [karma]: Karma v0.10.10 server started at http://localhost:9877/
INFO [launcher]: Starting browser Firefox
INFO [Firefox 56.0.0 (Fedora)]: Connected on socket ZgSjYh_osO5rZacRazqw
Firefox 56.0.0 (Fedora)  Array iteration: util.each at 116097316 ops/sec
[...]

Issue #1329, opened 2016-03-28
Use IS_MINIFIED to disable friendly debugger in minified builds


toolness:
Currently, the friendly debugger is disabled by overwriting all relevant private p5 methods (basically `_validateParameters` and `_friendlyFileLoadError`) in the footer of the minified build (done via the gruntfile configuration for `uglify`).

However, a disadvantage of this is that the _code_ for the friendly debugger is still present in minified builds--it's just overwritten at run-time to be a no-op.

In #1318, I introduced a new uglifyJS-based global defined only in minified builds called `IS_MINIFIED`. If we use this variable anywhere in p5's source code like so:

``` js
if (typeof(IS_MINIFIED) === 'undefined') {
  // We're in an unminified build, do friendly debugging logic etc. here
} else {
  // We're in a minified build, probably just do nothing.
}
```

then uglifyJS will realize that the conditional evaluates to a constant and will _remove_ the first branch when it generates the minified build. This is called "dead code removal" and it's nice because it means the minified build will really be as small as possible.

More background on how this works is explained here: https://github.com/processing/p5.js/pull/1318#issuecomment-202450591

lmccart:
Thanks @toolness! I can make this change unless you're already on it.

toolness:
Oh cool, thanks! Sure, you're welcome to make the change if you like, I was just filing the issue to make sure it wasn't forgotten about.

Issue #971, opened 2015-10-04
friendly error system fixes, updates, and improvements


lmccart:
The friendly error system was first implemented at the p5.js contributors conference. It is an effort to provide more human readable and friendly error messages in console. It also does some minimal argument type checking since JS does not do this. The nice thing about it is that it is only added to the p5.js build of the library, and eliminated from p5.min.js to improve performance speed. However, there are a number of issues and extensions that are required before it's fully up and running. 

I think the FEs would benefit from one person taking this on as a sort of concentrated project. It could be a good Processing Fellowship or Google Summer of Code project. For organization purposes, I am aggregating all the Friendly Error related issues here for now. I will close but link to all other issues. If someone were to tackle this project, the first step would be to read through all the issues and devise a roadmap and plan of attack.
- [validateParameters can't support multiple-type parameter](https://github.com/processing/p5.js/issues/868) - for example, methods dealing with color often can take either a color object or a number as an argument. however, the `validateParameters()` function doesn't support this.
- [Friendly Error Messaging for unsupported browsers](https://github.com/processing/p5.js/issues/657) - it would be nice to have messages that alert the user if they are trying to use functions of p5 that are not supported by their browser.
- [Friendly Error Messaging should show sketch line number](https://github.com/processing/p5.js/issues/750)
- [Generate Friendly Error Messaging from existing documentation](https://github.com/processing/p5.js/issues/759) - it would be nice to auto-generate the parameters accepted by each function from the reference, rather than having to manually add them to each function in the code.

These two pages describe some more thinking around it, currrent status, and future plans/ideas:
- https://github.com/processing/p5.js/wiki/Friendly-Debugger
- https://p5jscon.hackpad.com/Error-Reporting-6UAUU37d6j9

shiffman:
I'll just add a short note here that the recent Processing 3 implemented custom error messages for the first time (developed originally as part of "PDE X" by @Manindra29) -- the goals were to give more specific info about syntax errors as well as provide friendlier language and reduce scary Java-specific (and often misleading) messages.  If / when this project gets more development time, I'd be happy to help offer resources and information about some of the questions and solutions that came up during the Processing development.

For example, one thing that was helpful was just maintaining a spreadsheet of all the common errors, what causes them, and ideas for new language.

darkcoderrises:
HI, I would like to work on this. Can someone mentor me on this?

lmccart:
[...]

PR #1527, opened 2016-07-14
Adding flag for disabling the friendly error system

mikewesthad:
**Summary from #1512, #1329, #971:**  

The friendly error logging system causes a significant slowdown (e.g. 10x slowdown in Chrome for simple functions like `random` or `sin`).  

In global instance mode with unminified p5, p5 properties are added to the global scope using getters/setters via `Object.defineProperty` in [core.js#L593](https://github.com/processing/p5.js/blob/master/src/core/core.js#L593).  This makes it possible to log an error if a beginner (or not-so-beginner) inadvertently overrides a p5 property (e.g. `text`). This is a really helpful feature, but comes with a hefty performance cost.

**Proposed solution:**

Add a property to the p5 constructor function `disableFriendlyErrors` which can be used to prevent the friendly error getters/setters.  That way beginners get the friendly errors by default, but adding `p5.disableFriendlyErrors = true` to the top of a sketch will boost performance.  (I attached this property to the p5 constructor over p5.prototype - seems to make more sense that way.)

**Some Results:**

tl;dr: Huge gains in Chrome, reasonable gains in p5 editor & FF, no real gains in IE/Edge. I'm opening this PR as a solution to part of #1512, but it's disappointing that it's not a magic bullet for the p5 editor slowdown.  (Maybe that's because the chromium engine powering the editor is >1 year old?)

```
Calling p5 methods in global mode 10000000x times. 

Chrome, Windows:

Friendly Errors:  enabled       ->       disabled
random:           1518.62ms     ->       320.28md
abs:              1210.32ms     ->       91.99ms
sin:              1712.18ms     ->       478.58ms

Same thing in p5 Editor (built against this PR's p5.js):

Friendly Errors:  enabled       ->       disabled     
random:           3077.91ms     ->       1798.99ms
abs:              3153.38ms     ->       21935.76ms
sin:              3579.56ms     ->       2560.79ms


Times vary in FF but performance boost ranges from 2x-10x.
IE11/Edge minimal to zero gains.
```

Could use a better test (#1524), but for now, something manual:

```js

var iterations = 10000000;

function setup() {
  console.log("Friendly Errors: " +
    (p5.disableFriendlyErrors) ? "disabled" : "enabled");

  var start = performance.now();
  for (var i = 0; i < iterations; i += 1) {
    random();
  }
  var elapsed = performance.now() - start;
  console.log("random: " + elapsed.toFixed(2) + "ms");

  var start = performance.now();
  for (var i = 0; i < iterations; i += 1) {
    abs(i);
  }
  var elapsed = performance.now() - start;
  console.log("abs: " + elapsed.toFixed(2) + "ms");

  var start = performance.now();
  for (var i = 0; i < iterations; i += 1) {
    sin(i/iterations * (15 * 3.14));
  }
  var elapsed = performance.now() - start;
  console.log("sin: " + elapsed.toFixed(2) + "ms");
}
```

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -515,6 +515,10 @@ var p5 = function(sketch, node, sync) {
   }
 };
 
+// Allows for the friendly error system to be turned off when creating a sketch,
+// which can give a significant boost to performance when needed.
+p5.disableFriendlyErrors = false;
+
 // attach constants to p5 prototype
 for (var k in constants) {
   p5.prototype[k] = constants[k];
@@ -565,7 +569,8 @@ p5.prototype._createFriendlyGlobalFunctionBinder = function(options) {
   };
 
   return function(prop, value) {
-    if (typeof(IS_MINIFIED) === 'undefined' &&
+    if (!p5.disableFriendlyErrors &&
+        typeof(IS_MINIFIED) === 'undefined' &&
         typeof(value) === 'function' &&
         !(prop in p5.prototype._preloadMethods)) {
       try {


lmccart:
this is great! my one question is whether is makes more sense to have: 

```javascript

function setup() {
  createCanvas(100, 100);
  // etc
}
```

or a function call within setup, something like:
```javascript
function setup() {
  friendlyErrors(false); // or disableFriendlyErrors();
  createCanvas(100, 100);
}
```

maybe the second is a little more in line with the flow people use to think about and write p5, ie it is similar to the way you would [disable pixel density scaling](http://p5js.org/reference/#/p5/pixelDensity). or, do you feel this is on a meta level above the code and should be outside of setup?

mikewesthad:
Yeah, I do agree that having it be a method on p5.prototype would be more processing-like.  

The reason why I put it on the p5 constructor is because the value needs to be set before `setup` is called. Basically, the value needs to set before the page triggers a load event.  That's when the global p5 instance is created and the getters/setters are put in place.

I figured this would be cleaner than providing a way to rebind all the p5 properties/methods to the window on-the-fly via `friendlyErrors(false)`.  I was worried that rebinding might cause some weird edge cases, but let me take a look and see.

lmccart:
ah, I was thinking that might be the case. ok let's stick with this then.

PR #1531, opened 2016-07-17
Remove debug console log from p5.dom

JoshuaStorm:
I noticed when I was creating radio buttons in a personal project that I was getting a `console.log`. I imagine this isn't intended, so I just figured I'd put this PR out there.

Cheers


lmccart:
thanks!

PR #1533, opened 2016-07-19
Coordinate fixes

mindofmatthew:
This addresses one final point that I briefly mentioned in #1414. Now, the default camera distance adjusts so that 1 coordinate unit = 1 pixel for anything drawn on the plane z=0 (and therefore the sizes of 2d primitives are as expected).

I also updated the canvas resizing code so that it recalculates the camera aspect ratio if (and only if) the default camera is in use.

@indefinit, please take a look and let me know if you see any issues with how I approached this.


mindofmatthew:
Not sure why the build failed—as far as I can tell an unrelated test timed out, but I'll investigate further.

lmccart:
@mindofmatthew looks like the test is fine, I think it's just a timeout.

src/webgl/p5.RendererGL.js
```diff
@@ -38,7 +38,7 @@ p5.RendererGL = function(elt, pInst, isMainCanvas) {
   this.directionalLightCount = 0;
   this.pointLightCount = 0;
   //camera
-  this._isSetCamera = false;
+  this._setCamera = undefined;
```
indefinit: instead of assigning undefined, consider changing to null

src/webgl/p5.RendererGL.js
```diff
@@ -86,18 +86,18 @@ p5.RendererGL.prototype._initContext = function() {
 //detect if user didn't set the camera
 //then call this function below
 p5.RendererGL.prototype._setDefaultCamera = function(){
-  if(!this._isSetCamera){
+  if(this._setCamera === undefined){
```
indefinit: same comment about null as before

src/webgl/camera.js
```diff
@@ -89,7 +89,7 @@ p5.prototype.perspective = function(fovy,aspect,near,far) {
   );
   this._renderer.uPMatrix = p5.Matrix.identity();
   this._renderer.uPMatrix.perspective(fovy,aspect,near,far);
-  this._renderer._isSetCamera = true;
+  this._renderer._setCamera = 'custom';
```
indefinit: to me, the name setCamera seems like a setter function but in reality it's just a property of the renderer class.  Consider renaming to _camera, or _cam

indefinit:
made some code comments inline but otherwise looks great!  Thanks!

mindofmatthew:
I thought that "_curCamera" would be in keeping with some of the other renderer properties.

src/webgl/p5.RendererGL.js
```diff
@@ -383,6 +383,11 @@ p5.RendererGL.prototype.resize = function(w,h) {
   var gl = this.GL;
   p5.Renderer.prototype.resize.call(this, w, h);
   gl.viewport(0, 0, gl.drawingBufferWidth, gl.drawingBufferHeight);
+  // If we're using the default camera, update the aspect ratio
+  if(this._curCamera === 'default') {
+    this._curCamera = null;
```
indefinit: Do we really need to set the _curCamera to null if we are just resetting it to default in the _setDefaultCamera funct?

mindofmatthew: Yes, because _setDefaultCamera checks the value of _curCamera to determine whether or not to do anything. It's one of these functions that you can call liberally, because it only actually executes once (except now we need to override that behavior in this instance).

lmccart:
thanks @mindofmatthew!

PR #1550, opened 2016-08-02
readability changes to argument in limit function.

AkashGutha:
readability changes to the limit function for better readability.


lmccart:
I think this is somewhat confusing because the function name itself is already `limit`.

AkashGutha:
Yes. It's confusing . Any other ideas ??

lmccart:
you could use max, this is how it's documented in the ref for users.

AkashGutha:
did the changes. @lmccart

lmccart:
looks great, thanks!

PR #1552, opened 2016-08-03
added a gitignore for vscode folder.

AkashGutha:
for the dev's who are using VSCODE for development, to ensure that the .vscode folder isn't picked up by the git tracker adding a gitignore to the .gitignore file like this will work.


AkashGutha:
@lmccart  is a new line required at the end of file ? as being shown by the github's diff view .

lmccart:
no, it's not necessary. thanks for this!

Issue #1556, opened 2016-08-13
windowHeight / height is broken in Firefox 48 with p5.js 0.52


i3games:
`windowHeight / height` return 0 when canvas is created with (windowWidth, windowHeight).  Firefox 48.0 / Mac, p5.js 0.52. Works in Chrome 52.

## To reproduce: 

```
function setup() {
  // createCanvas(400, 400);
  createCanvas(windowWidth, windowHeight);
  print(windowWidth);
  print(windowHeight);
  print(width);
  print(height);
  print(displayWidth);
  print(displayHeight);
}

function draw() {
  ellipse(50, 50, 80, 80);
}

function mousePressed() {
  print(windowWidth);
  print(windowHeight);
}
```

## Possible fix:

replace 

```
p5.prototype.windowWidth = document.documentElement.clientWidth;
p5.prototype.windowHeight = document.documentElement.clientHeight;
```

with
 
```
p5.prototype.windowWidth = document.documentElement.clientWidth || document.body.clientWidth;
p5.prototype.windowHeight = document.documentElement.clientHeight ||  document.body.clientHeight;
```

This would require to include p5.js in the body, not in the head, however.

AkashGutha:
@i3games  was unable to reproduce the error on FireFox 48.0 / Windows.

Could you please share more info ?

limzykenneth:
Cannot reproduce on Mac as well

nok:
This is a duplicated and already fixed issue: https://github.com/processing/p5.js/issues/1479

Nevertheless thanks for noticing that issue, @i3games.

I guess, that the fix will roll out with the next release, probably `0.53`?

PR #1558, opened 2016-08-13
fix windowWidth & windowHeight properties

AkashGutha:
issue #1556 
this fix has been done as suggested by the article at [w3Schools] (http://www.w3schools.com/js/js_window.asp) 
this fix makes sure that it covers all browsers independent of their versions.

### DIFF
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -53,7 +53,7 @@ if (window.console && console.log) {
         var newArgs = JSON.parse(JSON.stringify(args));
         console.log(newArgs);
       }
-    } catch(err) {
+    } catch (err) {
       console.log(args);
     }
   };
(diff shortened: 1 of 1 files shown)


limzykenneth:
I don't think `document.body.clientWidth`/`document.body.clientWidth` will be ever used and is a good idea. `document.documentElement.clientWidth` is already supported on all browsers all the way back to IE6 [ref](http://caniuse.com/#search=documentelement). `document.body.clientHeight` can also be higher than the expected window size because it is looking at the size of the `<body>` HTML element which can be higher (page with scroll bar).

According to issue thread the fix should already be in the source, just need to wait for the next release. Personally, I would take w3schools with a pinch of salt as most of the time they don't give explanation behind solutions, making it hard to know precisely what is happening and vet the accuracy of it.

AkashGutha:
that sounds good. @limzykenneth

lmccart:
yes, thanks all this is correct. the fix is in there already and will be included in 0.5.3. hoping to get this out tomorrow. sorry for the delay and confusion.

PR #1563, opened 2016-08-20
Make winMouseX/winMouseY relative to window instead of document

gemedet:
This PR updates the variables `winMouseX/winMouseY` to be relative to the window instead of the document, by using `clientX/clientY` instead of `pageX/pageY`.

It also adds new corresponding touch variables: `winTouchX/winTouchY` and `pwinTouchX/pwinTouchY`.

### DIFF
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -311,20 +311,26 @@ p5.prototype.isMousePressed = false; // both are supported
 p5.prototype._updateNextMouseCoords = function(e) {
   var x = this.mouseX;
   var y = this.mouseY;
+  var winX = this.winMouseX;
+  var winY = this.winMouseY;
   if(e.type === 'touchstart' ||
      e.type === 'touchmove' ||
      e.type === 'touchend' || e.touches) {
     x = this.touchX;
     y = this.touchY;
+    winX = this.winTouchX;
+    winY = this.winTouchY;
   } else if(this._curElement !== null) {
     var mousePos = getMousePos(this._curElement.elt, e);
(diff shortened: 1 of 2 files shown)


src/events/mouse.js
```diff
@@ -311,20 +311,26 @@ p5.prototype.isMousePressed = false; // both are supported
...
     y = mousePos.y;
+    winX = mousePos.winX;
+    winY = mousePos.winY;
   }
   this._setProperty('mouseX', x);
   this._setProperty('mouseY', y);
-  this._setProperty('winMouseX', e.pageX);
-  this._setProperty('winMouseY', e.pageY);
+  this._setProperty('winMouseX', winX);
+  this._setProperty('winMouseY', winY);
```
lmccart: this makes sense to use clientX/Y instead of pageX/Y, but can you explain the rest of these changes a bit? would it work to just update these two lines here?

gemedet: Yeah, I think if we just updated these two lines to use clientX/Y, it'd work fine for the mouse position.

The issue I ran into was that touch events don't have clientX/Y, so it has to be grabbed from the touches array (lines 196-197 in touch.js). Then everything else is just to pass that to winMouseX/Y. You'll notice I mimicked the way that touchX/Y is passed to mouseX/Y.

Looking at it now, we could just let the touch events populate the mouse properties directly. Then we could drop touchX/Y, as it's essentially a duplicate of mouseX/Y, right?

lmccart:
ah got it. I think you're right...I'll merge this now and open another issue for that.

PR #1568, opened 2016-08-25
Custom shaders support

mindofmatthew:
I'm not ready for this to be merged yet—I still need to rename some uniforms (bringing them in line with the PShader API) and finish method documentation.

That said, @indefinit, it would be great if you could start looking over this as I continue wrapping up my tutorials.


lmccart:
oooh this is exciting

src/webgl/p5.Shader.js
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * p5 Shader class
+ * @constructor
+ * @param  {String} fragShader Source code of a fragment shader as
+ * @param  {String} vertShader Source code of a vertex shader as a string
+ *
+ */
+p5.Shader = function(fragSource, vertSource){
+  //TODO: this key should be the hash of the optional fragSource and vertSource
+  this.shaderKey = 'customShader';
```
indefinit: so if I understand correctly, this.shaderKey would essentially be an id for lookup in a material hash?

src/webgl/light.js
```diff
@@ -39,32 +39,16 @@ var p5 = require('../core/core');
...
-    shaderProgram,
-    'uAmbientColor[' + this._renderer.ambientLightCount + ']');
-
   var color = this._renderer._pInst.color.apply(
     this._renderer._pInst, arguments);
-  var colors = color._array;
-
-  gl.uniform3f( shaderProgram.uAmbientColor,
-    colors[0], colors[1], colors[2]);
+  var colors = color._array.slice(0, 3);
```
indefinit: consider renaming var as rgb, since we already have a var color on line 42

src/webgl/light.js
```diff
@@ -39,32 +39,16 @@ var p5 = require('../core/core');
...
 
-  //in case there's no material color for the geometry
-  shaderProgram.uMaterialColor = gl.getUniformLocation(
-    shaderProgram, 'uMaterialColor' );
-  gl.uniform4f( shaderProgram.uMaterialColor, 1, 1, 1, 1);
+  this._renderer._setUniform('uAmbientColor[' +
```
indefinit: I wonder whether we can leverage the p5.Shader object in the core so that we don't have to call a private method _setUniform on the renderer object.  It would be great if we could get the shader program and then call set() on that program.

src/webgl/light.js
```diff
@@ -139,22 +123,13 @@ p5.prototype.directionalLight = function(v1, v2, v3, a, x, y, z) {
...
-  gl.useProgram(shaderProgram);
-  shaderProgram.uDirectionalColor = gl.getUniformLocation(
-    shaderProgram,
-    'uDirectionalColor[' + this._renderer.directionalLightCount + ']');
-
   //@TODO: check parameters number
   var color = this._renderer._pInst.color.apply(
     this._renderer._pInst, [v1, v2, v3]);
-  var colors = color._array;
+  var colors = color._array.slice(0, 3);
```
indefinit: same comment re: var naming as above

src/webgl/material.js
```diff
@@ -343,4 +323,34 @@ p5.RendererGL.prototype._applyColorBlend = function(v1,v2,v3,a){
   return colors;
 };
 
+
+p5.prototype.loadShader = function(fragShader, vertShader) {
+  var loadedShader = new p5.Shader();
```
indefinit: why not use the shader key as argument to the p5.Shader() constructor?

src/webgl/p5.RendererGL.js
```diff
@@ -41,15 +41,30 @@ p5.RendererGL = function(elt, pInst, isMainCanvas) {
...
    */
-  this.uMVMatrix = new p5.Matrix();
-  this.uPMatrix  = new p5.Matrix();
-  this.uNMatrix = new p5.Matrix('mat3');
+  this._uniforms = Object.create(null);
+  this._setUniform('uResolution', RESOLUTION);
+  this._setUniform('uModelViewMatrix', new p5.Matrix());
+  this._setUniform('uProjectionMatrix', new p5.Matrix());
+  this._setUniform('uNormalMatrix', new p5.Matrix('mat3'));
+  //TODO: Possibly Normal Matrix doesn't work in immediate mode? Investigate.
```
indefinit: yea we didn't originally have a normal matrix in immediate mode

src/webgl/p5.RendererGL.Retained.js
```diff
@@ -98,26 +85,37 @@ p5.RendererGL.prototype.createBuffers = function(gId, obj) {
...
+      var attribLocation = gl.getAttribLocation(shaderProgram, attribute.name);
+      if(attribLocation !== -1) {
+        gl.enableVertexAttribArray(attribLocation);
+        gl.vertexAttribPointer(attribLocation,
+                               attribute.size, attribute.type, false, 0, 0);
+      }
+    }
+  }
+
+  //TODO: This re-binds the textures each render call, which could be more
```
indefinit: yea, let's discuss more.  I agree this seems like it could be a big bottleneck especially if we're performing video texture manipulation.

src/webgl/p5.RendererGL.js
```diff
@@ -127,151 +144,178 @@ p5.RendererGL.prototype.background = function() {
...
-  gl.useProgram(shaderProgram);
+/**
+ * @param {String} name The name of the uniform.
+ * @param {any} data The data to set in the uniform. This can take many
+ *                   different forms. See p5.Shader.setUniform for full
+ *                   documentation.
+ */
+p5.RendererGL.prototype._setUniform = function()
+{
+  var args = Array.prototype.slice.call(arguments);
```
indefinit: use 
```
    args[i] = arguments[i];
  }
```
instead of slice

indefinit:
Awesome work!  There's some seriously heavy lifting in this PR.  I added a few comments inline.  I think the code is pretty readable.  I guess my only high level suggestion is what you already know per your PR description - that we should try to leverage the p5.Shader object internally as much as we can.  I'll keep testing this to see if there's anything breaking but otherwise I think you're very close to merge time.

src/webgl/p5.RendererGL.js
mindofmatthew: Will do

src/webgl/light.js
mindofmatthew: Will do

src/webgl/light.js
mindofmatthew: Will do

src/webgl/light.js
mindofmatthew: I didn't want to assume that the shader program necessarily exists yet when this function is called, so instead I went with a system where uniform data is saved first, and then applied to the shader program at render-time.

Right now, we've got a distinction between renderer._setUniform, which can set the uniform for all shader programs, and shader.setUniform, which only sets it for a specific program. I'd like to keep this distinction, but think that there's probably a way to make it more understandable, either through different function naming, or better documentation, or maybe moving all global uniform setting to the p5.Shader class. Any thoughts?

src/webgl/p5.Shader.js
mindofmatthew: Yes, although this particular shaderKey may not be used anymore. Let me check into that.

src/webgl/material.js
mindofmatthew: This gets into a break with the Processing PShader API that we should probably discuss: in Processing, the arguments for the constructor are the frag and vert file paths. In p5.Shader, I switched to using the actual frag and vert source code as creation arguments, which works better for default shaders (and in the case where, for example, someone wanted to make a p5 project that pulled shader source from somewhere else, like a DOM node). However, it means that loadShader has to return an empty shader object at first, and then asynchronously populate the sources of that shader, once those sources actually load.

indefinit: Ok this makes sense.  I think we should keep your implementation as is.  Perhaps a very small change should be to rename the loadedShader variable as shaderProgram, since technically the shader isn't loaded until the frag and vert sources have finished loading asynchronously.

src/webgl/light.js
indefinit: Hmm, I'm not sure if I see the problems of explicitly checking whether the shader program exists and if not, create one, then apply the uniform.

I think you're right that having a clear distinction between a global uniform setting vs. program specific uniform set is very important, but also setting global uniforms in a p5.Shader instance could be dangerous and potentially confusing since we really want to expose as much as the p5.Shader api as possible to the user.

indefinit:
Hi Matthew, Thanks for all of the work on this so far.  I've been going back and forth on this for the last couple of weeks and I think we should move the uniforms and defines' getters and setters into the p5.Shader class.  The main reason to pursue the refactor is because of long term readability.  We should really keep all shader related "properties", like uniforms and defines, in the p5.Shader class.  Another potentially confusing area is the shader.js module, which instantiates our default materials.  Perhaps we can rename shader.js to shaderMacro.js, or another name that doesn't risk confusion with p5.Shader.

I started quickly refactoring your PR on a local branch just to make sure I wasn't crazy for suggesting it.

Also, I tested your branch against some primitives and it seems like box disappears if you call cone, cylinder, or torus afterward.  It's probably another reason to keep the various uniforms and defines objects together with their shader just in case there's any naming collisions.

I'm happy to continue the conversation offline as well if you want to go more in depth.

src/webgl/p5.RendererGL.js
```diff
@@ -1,9 +1,10 @@
 'use strict';
 
 var p5 = require('../core/core');
-var shader = require('./shader');
+var builtInShaders = require('./shader');
```
indefinit: hey, just a reminder we should change this to defaultShaders.js or shaderBuilder.js

src/webgl/p5.RendererGL.js
```diff
@@ -127,149 +143,113 @@ p5.RendererGL.prototype.background = function() {
...
+/**
+ * Copy data from renderUniforms to specific shader uniforms.
+ */
+p5.RendererGL.prototype._copyRenderUniformsToShader = function(shader)
+{
+  for(var uniformName in this.renderUniforms) {
+    shader.set(uniformName, this.renderUniforms[uniformName]);
   }
-};
+}
```
indefinit: missing semicolon

src/webgl/material.js
```diff
@@ -115,64 +117,57 @@ p5.prototype.normalMaterial = function(){
...
-  var args = new Array(arguments.length);
-  for (var i = 0; i < args.length; ++i) {
-    args[i] = arguments[i];
-  }
-  var gl = this._renderer.GL;
-  var shaderProgram = this._renderer._getShader('lightVert',
-    'lightTextureFrag');
-  gl.useProgram(shaderProgram);
+p5.prototype.texture = function(tex){
+  this._renderer.shaderUniforms.uSampler = tex;
```
indefinit: when calling texture(), I'm getting shaderUniforms undefined

src/webgl/material.js
```diff
@@ -245,23 +240,12 @@ p5.RendererGL.prototype._bind = function(tex, data){
...
-
-  gl.uniform4f(shaderProgram.uMaterialColor,
-    colors[0], colors[1], colors[2], colors[3]);
-
-  shaderProgram.uSpecular = gl.getUniformLocation(
-    shaderProgram, 'uSpecular' );
-  gl.uniform1i(shaderProgram.uSpecular, false);
-
-  gl.uniform1i(gl.getUniformLocation(shaderProgram, 'isTexture'), false);
+  this._renderer.shaderUniforms.uMaterialColor = colors;
```
indefinit: I think this should be renderUniforms, right?

maxkolyanov:
This is a pretty old PR, but is this feature still on the roadmap? 
I'd really like to see custom shader support in p5.js and would love to help. 
I've experimented and learned a lot about WebGL lately and feel pretty confident I can pick up this branch and finish it to some extent.

Or maybe it would be better to start off from scratch?
Appreciate any input from core contributors on where would be best place to start.
Thanks! ✨

mlarghydracept:
@maxkolyanov hi! It is great that you want to build off this PR. Just one quick thing to note is that @kjhollen and I will be working on the WebGL for summer of code so we may be looking at different design approaches for something like custom shaders. So if you do decide to take a crack at it, especially if you want to start from scratch, it would be great to have a chat about it first!

mindofmatthew:
Hi @mlarghydracept, @maxkolyanov, and @kjhollen!

Matthew here—I worked on this last summer up until the point where I ran into some issues where I couldn't quite make my design work with the p5 api and got stalled out. Unrelatedly, I got busy with other projects and kind of orphaned this branch.

I'm glad to see other folks working to push p5 WebGL forward! If anyone would like, I'd be happy to discuss the design that I was attempting here, which parts could be potentially adapted to other approaches, what issues I ran into, etc. I'll have some more free time this summer again, so let me know if there's anything I can do to help out, but I'm happy to let the summer of code folks take the lead in terms of figuring out how p5 WebGL should proceed.

maxkolyanov:
Great to hear that a lot of people are interested in making WebGL and p5.js best friends! YAY
I agree with @mindofmatthew, since custom shaders will depend on a lot of core p5 WebGL code, I think @mlarghydracept and @kjhollen should lead this. It just will be hard to mix two approaches together, if they are being developed by different people and I think I just won't have time/resources to keep up with GSoC.

tolkanabroski:
I writing this, becorce i also have trouble with a work of preprocessor (my Teacher)
I think P5js is a learning envirompent. I came across Processing few weeks ago, and still fighting against the default [(things that just happen under the hood) setup. ](https://github.com/processing/processing/wiki/Advanced-OpenGL) Whether i a newbie or not, i will be the best in the thing, that i love.  Usually i develop my skills and gain experience, when i break things into the basic atoms and see how they work.. People who are interessted in learning shaders get inpriration from site like shadertoy and will come sooner or later across WebGL.

I would like to see  minimal WebGL example.

If someone write a comstom shaders for p5js. Please leave everything minimal and raw as possible. 

For me it's okay that their are people that use Processing 
and PShader = loadShader(".frag");
 
But i don't think that [any of thouse realy use shaders](https://forum.processing.org/two/categories/glsl-shaders), maybe a Copy+Paste Work just to add a filter, if you dealing with shaders, theirs is not way, you don't read one or other tutorial that cracks up the default loadShader method. 

In other words:
A student needs the red pill, he will enter the Matrix.

http://alpha.editor.p5js.org/nabr/sketches/Sk-hp7slb

Best

tolkanabroski:
@mindofmatthew @maxkolyanov 

I would love to see shaders in P5js!


My english is not so well to explain things. 

The good thing about javascript is that only the last call in stack is valid. So its easy to override and we can use the editor.

Later we can run unit tests, first lets get started:

http://alpha.editor.p5js.org/nabr/sketches/S1clfrTeb

lmccart:
going to close this for organizational purposes... but we'll be looking at all of this closely as we work on developing webgl mode further this summer. thanks @mindofmatthew!

PR #1577, opened 2016-08-31
Canvas Resizing (Fix windowWidth and windowHeight)

craftoid:
The implementation for assigning window dimensions to `windowWidth` and `windowHeight` introduced in #1494 does not work as expected.  As a result down-resizing the browser window to a smaller size will not resize the canvas appropriately, when using this snippet of code:

```js
function windowResized() {
}
```

The `windowWidth` and `windowHeight` variables should refer to the size of the viewport.
So we must use `window.innerWidth` and `window.innerHeight`.

Both `document.documentElement.clientWidth` and `document.body.clientHeight`
may be used as polyfill, but they refer to the dimensions of the body and the html element,
which may be greater than those of the actual viewport (!).

### DIFF
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -392,14 +392,17 @@ p5.prototype._onresize = function(e){
 };
 
 function getWindowWidth() {
-  return Math.max(
-    document.documentElement.clientWidth,
-    window.innerWidth || 0);
+  return window.innerWidth ||
+         document.documentElement && document.documentElement.clientWidth ||
+         document.body && document.body.clientWidth ||
+         0;
 }
+
 function getWindowHeight() {
-  return Math.max(
-    document.documentElement.clientHeight,
-    window.innerHeight || 0);
+  return window.innerHeight ||
+         document.documentElement && document.documentElement.clientHeight ||
+         document.body && document.body.clientHeight ||
+         0;
 }
 
 /**


limzykenneth:
Hmm...I don't quite understand why `document.documentElement && document.documentElement.clientWidth` is used instead of just `document.documentElement.clientWidth` since if `document.documentElement` is not available, `document.documentElement.clientWidth` should fall through to the next case either way. Also I don't think there's much reason to include `document.body.clientHeight` as part of the polyfill as the previous cases should already cover by all browsers all the way to IE6 and as you mentioned it might give unexpected values.

craftoid:
@limzykenneth The `document.documentElement` part is to make sure the `documentElement`` really exists. I assumed that `document.documentElement.clientWidth`would correspond to `undefined.clientWidth` and might yield an error, such as`TypeError: undefined has no properties` (https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Errors/No_properties
) But maybe this is just a Firefox thing (I haven't really done any cross browser testing)

I agree that including both `documentElement` and `body` in the polyfill may be overdoing it, unless you want browser compatibility down to IE4 ;)

However note that both `documentElement` (refering to the `html` tag) and `body` may yield values which are greater than the actual viewport dimensions, resulting in potential problems when resizing the window.

lmccart:
sorry for the delay, this looks good to me. thank you!

Issue #1580, opened 2016-09-03
Slider does not support floats properly


brmscheiner:
The createSlider function accepts floating point arguments but handles them in unexpected ways. For example, `createSlider(0.1, 3.7, 2)` creates a slider with initial value 2.1 (?!) that can only take the following four values: 0.1, 1.1, 2.1 and 3.1. 

One potential fix would be to refactor the slider so that it can return any floating point number between the min and max (rather than taking steps of 1). Once there's some consensus about what to do I would be happy to implement the solution =).

golanlevin:
I've seen libraries for other toolkits in which the sliders' constructors also take a bool argument indicating whether they are continuous or 'discrete' (returning integers only).

lmccart:
@brmscheiner I think it'd be great to change this so we can support floats as well. I'd suggest a 4th argument (currently step size) = 0 to indicate continuous. Otherwise I think we can default to discrete with step size 1. 

As an addition, we could do some checking, like if the user has entered floating point numbers as the first or second argument, we automatically assume it's continuous. I'm not sure if this adds room for confusion, in which case I'm fine with defaulting to discrete even if the behavior in the example above is weird without the 4th argument of "0".

brmscheiner:
Sure! I think it makes sense to default to step=1, especially so that people's old code will still work the way it did before

lmccart:
fixed with https://github.com/processing/p5.js/pull/1590

PR #1590, opened 2016-09-14
Continuous slider behavior when step=0

brmscheiner:
Relevant issue: [issue 1580](https://github.com/processing/p5.js/issues/1580).
HTML range input doesn't have a continuous mode out of the box, so I did a little bit of testing ([jsfiddle](https://jsfiddle.net/brmscheiner/p2agoodt/9/)) and found the smallest valid slider increment. Step takes that value when the user sets step=0.


lmccart:
very cool, thank you!

brmscheiner:
no problem!

PR #1592, opened 2016-09-17
Better error-handling on failed font load

dhowe:
I'm not sure if there is some unified way to do error-handling when things (images, fonts, JSON, etc) fail to load, but currently we get a rather unhelpful error message. This PR adds a better console error message and also calls p5. _friendlyFileLoadError, though this doesn't produce any output in the browser.

### DIFF
--- a/src/core/error_helpers.js
+++ b/src/core/error_helpers.js
@@ -210,7 +210,12 @@ var errorCases = {
   '3': {
     fileType: 'text file',
     method: 'loadStrings'
-  }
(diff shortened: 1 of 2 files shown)


toolness:
Thanks @dhowe !  I volunteered to fix `_friendlyFileLoadError` a little while ago but I think I might have gotten stuck at writing unit tests for it or something... I need to get back onto that!

lmccart:
thanks!

Issue #1570, opened 2016-08-28
Document/unit test ability for clients to initialize global mode explicitly


toolness:
The TL;DR here is that it'd be nice to be able to initialize p5 global mode *explicitly* at any point in a web page's lifetime.

Currently there is no *documented* way to do this, but there is an *undocumented* way to do it: simply call `new p5()`, without any arguments.

So this issue isn't actually about adding any new functionality--it's just about documenting and/or adding unit tests so that tools or sketches that depend on it don't break in the future.

## Use case for the p5 widget

In the [p5 widget](https://github.com/toolness/p5.js-widget), we're currently trying to support scenarios where widget embedders may want to [specify p5 addon libraries to load](https://github.com/toolness/p5.js-widget/issues/53).

Currently, the p5 widget's preview frame works like this:

1. Dynamically add a `<script>` element to the page containing the user's sketch code (which expects to be run in global mode).
2. Dynamically add a `<script>` element with its `src` attribute pointing at a version of p5.

As soon as p5.js loads, it sees the `setup`/`draw` functions defined by the user's sketch and initializes global mode.

However, because the p5 widget loads p5 asynchronously, adding support for p5 addon libraries poses a conundrum:

* We can't load p5 addon libraries *before* loading p5 because they depend on e.g. [`p5.prototype.registerPreloadMethod()`](https://github.com/processing/p5.js/wiki/Libraries#creating-a-new-library) to function properly.

* It's difficult to load p5 addon libraries *after* loading p5 because p5 will have noticed that `setup`/`draw` exists in the global namespace and will start the sketch immediately, before we have a chance to asynchronously load any addon libraries.

While there are some ways the p5 widget could [change its architecture](https://github.com/toolness/p5.js-widget/issues/53#issuecomment-240387563) to work around this, the easiest solution would be the following:

1. Asynchronously load p5.js, followed by any add-on libraries.
2. Add a `<script>` element to the page containing the user's p5 global mode sketch.
3. Activate global mode to start the sketch.

Step 3 is where we need ongoing support from p5. While it's possible *right now*, it'd be nice to have the functionality documented and/or unit tested, so that the widget doesn't break when used with newer versions of p5.

I've written a working proof-of-concept of this approach at https://github.com/toolness/p5.js-widget/pull/63.

## Other benefits

Aside from making the p5 widget's implementation easier, I can think of a few other benefits to being able to explicitly start p5 global mode:

[...]

PR #1593, opened 2016-09-18
Support and test on-demand global mode

toolness:
This adds support and unit tests for on-demand global mode, fixing #1570.

In particular, it fixes a bug whereby p5 would actually initialize itself *twice*: once when on-demand global mode was triggered via an explicit `new p5()`, and again when the page's `load` event fired (i.e., "non-on-demand global mode").  This would then cause a bunch of warnings from #1318 to be logged.

Now `p5.instance` is always written when `_isGlobal` is set, and it's checked before initializing non-on-demand global mode.

Since there weren't already tests for non-on-demand global mode, I also added some.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -471,6 +471,7 @@ var p5 = function(sketch, node, sync) {
   // assume "global" mode and make everything global (i.e. on the window)
   if (!sketch) {
     this._isGlobal = true;
+    p5.instance = this;
     // Loop through methods on the prototype and attach them to the window
     for (var p in p5.prototype) {
       if(typeof p5.prototype[p] === 'function') {
@@ -535,6 +536,10 @@ var p5 = function(sketch, node, sync) {
     }
   }
 };
+
+// This is a pointer to our global mode p5 instance, if we're in
+// global mode.
+p5.instance = null;
 
 // Allows for the friendly error system to be turned off when creating a sketch,
 // which can give a significant boost to performance when needed.
--- a/test/unit/color/p5.Color.js
+++ b/test/unit/color/p5.Color.js
@@ -781,7 +781,7 @@ suite('p5.Color', function() {
     });
   });
 
-  suite.only('p5.Color.prototype.toString', function() {
+  suite('p5.Color.prototype.toString', function() {
     var colorStr;
 
     setup(function() {
--- a/src/app.js
+++ b/src/app.js
@@ -74,9 +74,10 @@ var _globalInit = function() {
   if (!window.PHANTOMJS && !window.mocha) {
     // If there is a setup or draw function on the window
     // then instantiate p5 in "global" mode
-    if((window.setup && typeof window.setup === 'function') ||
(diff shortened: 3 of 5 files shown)


test/unit/color/p5.Color.js
```diff
@@ -781,7 +781,7 @@ suite('p5.Color', function() {
     });
   });
 
-  suite.only('p5.Color.prototype.toString', function() {
+  suite('p5.Color.prototype.toString', function() {
```
toolness: Looks like this was introduced in #1463. It didn't seem to affect running tests from the command-line, but when run in-browser it'd configure mocha to [only run this suite](http://jaketrent.com/post/run-single-mocha-test/) and no others.

toolness:
Er... I am really confused by the test failure.  As far as I can tell from looking at the source, the test that's failing *should* have started failing as of f024aab406eafbc58bb1ea0a0d72936cb0a62f7e...

toolness:
Hmm, in fact, that test *is* failing when I run it in my browser on `master`, but it's not failing when run through the command-line.  This is really weird!

test/unit/color/p5.Color.js
toolness: Oh, I was wrong about this not affecting tests from the command-line.  It *does* affect tests from the command-line!  In fact, f024aab406eafbc58bb1ea0a0d72936cb0a62f7e broke the tests, but they didn't show up in Travis CI because of this line :disappointed:

toolness:
Blerg, figured it out--see https://github.com/processing/p5.js/pull/1593#r79306305.

@lmccart any idea what to do about the failing test introduced in f024aab406eafbc58bb1ea0a0d72936cb0a62f7e?

limzykenneth:
Was just about to say it was that commit. I think we can just fix [this line](https://github.com/processing/p5.js/blob/master/test/unit/io/files_input.js#L28) in the test definition.

toolness:
Cool, do you think we should just change it to test if the result is an `object` instead of an `array`?

lmccart:
oops, yes we can just change that to object. I wonder why it didn't fail when I ran it. sorry about that!

toolness:
No worries!  It didn't fail when you ran it because of the [`suite.only()`](https://github.com/processing/p5.js/pull/1593#r79305812) debug call that was accidentally left in as part of #1463, I think.

Just committed a fix, so this should be ready for review!

src/core/core.js
```diff
@@ -471,6 +471,7 @@ var p5 = function(sketch, node, sync) {
   // assume "global" mode and make everything global (i.e. on the window)
   if (!sketch) {
     this._isGlobal = true;
+    p5.instance = this;
```
toolness: If we want to be really friendly, we could actually add a check here to see if `p5.instance` is non-`null`, and log a friendly warning if so.  Hopefully that should never happen, but it could potentially cause some hard-to-catch bugs if it ever does.

lmccart:
this looks all good, very helpful stuff!!

PR #1605, opened 2016-09-28
Update documentation of .mouseWheel()

### LINKED ISSUE #1604: Canvas.mouseWheel event is undefined
I'm trying to run this example from the reference: http://p5js.org/reference/#/p5.Element/mouseWheel

This works fine in Chrome and Edge, but not in Firefox or IE.

In Firefox I get `ReferenceError: event is not defined`. In IE I don't get any errors, but if I print the value of `event` I see that it is `undefined`.

I found [this closed issue](https://github.com/processing/p5.js/issues/503) which might be related, and [this closed issue](https://github.com/processing/p5.js/issues/39) that makes me wonder if `mouseWheel()` was finished for non-chrome browsers?

I should also note that the regular (not using the Canvas) `mouseWheel()` examples works fine on all browsers: http://p5js.org/reference/#/p5.Element/mouseWheel

Is this expected behavior?

Originally reported on the Processing forum: https://forum.processing.org/two/discussion/comment/75533

limzykenneth:
Fixes #1604. Do have a look through the changes since this is basically only updating the documentation to reflect latest API. Don't think anything broke in the code itself so I didn't change anything.

### DIFF
--- a/src/core/p5.Element.js
+++ b/src/core/p5.Element.js
@@ -195,11 +195,17 @@ p5.Element.prototype.mousePressed = function (fxn) {
 /**
  * The .mouseWheel() function is called once after every time a
  * mouse wheel is scrolled over the element. This can be used to
- * attach element specific event listeners.<br><br>
- * The event.wheelDelta or event.detail property returns negative values if
- * the mouse wheel if rotated up or away from the user and positive in the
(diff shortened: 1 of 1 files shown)


lmccart:
thanks!

PR #1611, opened 2016-09-30
Implement events to createCanvas, resizeCanvas and noCanvas

### LINKED ISSUE #1602: createCanvas and resizeCanvas EVENTS
I'm creating a lib for p5 and would be great if the `createCanvas` and `resizeCanvas` methods triggered an event, so my lib can be adjusted to the canvas new size.

daniloprates:
I have made this implementation to make it possible to bind functions to the `createCanvas`, `resizeCanvas` and `noCanvas` actions.

E.g.:

```js

function setup() {
	p5.onCreateCanvas = function(w,h) {
		console.log('HABEMUS CANVAS!', w, h)
	}
	p5.onResizeCanvas = function(w,h) {
		console.log('CANVAS RESIZED!', w, h)
	}
	p5.onNoCanvas = function() {
		console.log('NO CANVAS AT ALL!')
	}

}

```

It would be very useful for a library I'm working on, but please feel free to ignore if it's not of your interest or suggest amendments.

Thanks!

If accepted, it closes #1602

### DIFF
--- a/src/core/rendering.js
+++ b/src/core/rendering.js
@@ -113,6 +113,11 @@ p5.prototype.createCanvas = function(w, h, renderer) {
   if (isDefault) { // only push once
     this._elements.push(this._renderer);
   }
+
+  if (typeof p5.onCreateCanvas === 'function') {
+    p5.onCreateCanvas.call(this, w, h, r);
+  }
+
   return this._renderer;
 };
 
@@ -159,6 +164,10 @@ p5.prototype.resizeCanvas = function (w, h, noRedraw) {
     if (!noRedraw) {
       this.redraw();
     }
+
+    if (typeof p5.onResizeCanvas === 'function') {
+      p5.onResizeCanvas.call(this, w, h, noRedraw);
+    }
   }
 };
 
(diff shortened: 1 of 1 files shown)


lmccart:
Thanks. Can you give more explanation of the use case here? If you are calling createCanvas() anyway, couldn't you just call whatever additional functions you wanted after that line?

lmccart:
Ah, sorry I just saw issue #1602. I understand the use case now. I think this sounds fine, but the event handlers should be added via the `registerMethod()` infrastructure seen here: https://github.com/processing/p5.js/wiki/Libraries#use-registermethod-to-register-functions-with-p5-that-should-be-called-at-various-times.

daniloprates:
Should I create this?
```js
p5.prototype.onCreateCanvas = function(fxn) {
  attachListener('createcanvas', fxn, this);
}
```
how would I trigger `createcanvas` from the the `createCanvas` function?

daniloprates:
would `onCreateCanvas`, `onResizeCanvas` and `onNoCanvas` be method names, like `pre` and `post`?

daniloprates:
Just to explain a bit better what I'm trying to do:

This lib I'm creating relies in the canvas size. So I just need a way to proper check it.

I could have a `pre` prototype function, so I would know when `width` or `height` have changed, but I'm not sure this is a good way of doing it.

daniloprates:
Closing it as `registerMethod` will do the job =]

thanks @lmccart.

PR #1615, opened 2016-10-02
Fix clipping of WebGL objects on small canvas

### LINKED ISSUE #1544: webgl: 2d primitives disappear with canvas size 100x100
Referencing: Issue #1505

Seems to be related to camera perspective on canvas sizes exactly less than 116x116.  Sample code:
```
//works!
function setup() {
  createCanvas(116, 116, WEBGL);
}
function draw() {
  fill('black');
  rect(0,0,10,10);
}

//doesn't work
function setup() {
  createCanvas(115, 115, WEBGL);
}
function draw() {
  fill('black');
  rect(0,0,10,10);
}
```
No errors produced.

mindofmatthew:
This fixes #1544: the issue is that the default camera now moves closer and further away from the scene depending on canvas size, so that the size in pixels of an object is consistent across different sizes of canvas. What the code neglects to do is adjust the near and far clipping planes, which means that for smaller canvases, the camera gets so close that the objects in the scene (if they're at or near z=0) are technically "in front" of the rendered region.

For now, I've changed the default near plane and verified that it fixes the issue for canvas sizes down to and including 1x1.

Processing solves this by dynamically altering the near and far planes, but I couldn't get the default values that Processing uses to work here, which suggests some inconsistency between how Processing and p5.js deal with perspective matrices. I can investigate that further, but for now I think this is a fine fix for the problem at hand.


indefinit:
Thanks!

Issue #1588, opened 2016-09-13
 "video.get()" not working


eglrd:
```javascript
var eCam;
var eButton;
var eSnapshots = [];

function setup() {
  createCanvas(320, 240);
  background(51);
  eCam = createCapture(VIDEO);
  eCam.size(320, 240);
  eCam.hide();
  eButton = createButton('snap');
  eButton.mousePressed(takeSnap);
}

function takeSnap(){
  eSnapshots.push(eCam.get());
}

function draw() {
  for(var i=0;i < eSnapshots.lenght; i++){
    image(eSnapshots[i], 0, 0, 50, 50);
  }
}
```

triggers
p5.js:12060 Uncaught TypeError: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The provided value is not of type '(HTMLImageElement or HTMLVideoElement or HTMLCanvasElement or ImageBitmap)'

on macos with 10.11.6 hosting with node.js 4.5.0, and p5js v0.5.3﻿

shiffman:
I did some digging on this and tested the following code.  It works properly in 0.5.2 but fails in 0.5.3.  I'm going to see if I can pinpoint what changed caused the problem.

```javascript
function setup() {
  createCanvas(320, 240);
  video = createCapture(VIDEO);
  video.size(320, 240);
}

function draw() {
  background(51);
  var snapshot = video.get();
  image(snapshot, 0, 0);
}
```

shiffman:
Ok, after some digging it's this commit that breaks `get()` 8952a7c6869ebeeb385ce5e97adf8efb6fc26dc0.

I'll file a pull request to fix and add `loadPixels()` back in.  (But maybe there's something I am missing, feel free to let me know!)

shiffman:
@hobzcalvin Adding you to this thread as I noticed the commit I'm referencing above (https://github.com/processing/p5.js/commit/8952a7c6869ebeeb385ce5e97adf8efb6fc26dc0) was from your pull request.  Let me know if I've got something wrong here!

lmccart:
fixed with https://github.com/processing/p5.js/pull/1616

Juarez-BioIT:
I tried this code today in the p5.js web editor and is still not working

lmccart:
there's a typo in the above code `eSnapshots.lenght` should read `eSnapshots.length`
when I made this change, the code runs in the editor

```js
var eCam;
var eButton;
var eSnapshots = [];

function setup() {
[...]

Issue #1490, opened 2016-06-27
copy and element created with createCapture


vanevery:
This issue seems to be related to #995 and could probably be solved in the same way or it could be fixed in the createCapture code in the DOM library.

In any case it _seems_ that this should work but it fails on reading the width capture object.  Perhaps there is a better method to do this that I am missing:

```javascript
var cameraFeed;
var myImage;
var button;

function takeSnapshot() {
    myImage.copy(cameraFeed,0,0,320,240,0,0,320,240);
}

function setup() {
    cameraFeed = createCapture(VIDEO);
//     cameraFeed.size(320,240);
//     cameraFeed.canvas = {width: 320, height: 240};
//     console.log(cameraFeed);
    button = createButton('SNAP');
    button.mousePressed(takeSnapshot);
   
	myImage = createImage(320,240);
}

function draw() {
    image(myImage,0,245,320,240);
}
```

lmccart:
should work now... I just had to change your last line so that it shows up on the canvas (which is 100x100 by default)
`image(myImage, 0, 0, 320, 240);`

vanevery:
Wonderful!  Thanks!

PR #1616, opened 2016-10-02
adds loadPixels back in re: #1588

shiffman:
@hobzcalvin tagging you in case you want to take a look as this reverts part of  8952a7c6869ebeeb385ce5e97adf8efb6fc26dc0

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -999,8 +999,13 @@
     // set width and height onload metadata
     elt.addEventListener('loadedmetadata', function() {
       elt.play();
-      c.width = c.elt.width = elt.videoWidth;
-      c.height = c.elt.height = elt.videoHeight;
+      if (elt.width) {
+        c.width = elt.videoWidth = elt.width;
+        c.height = elt.videoHeight = elt.height;
+      } else {
+        c.width = c.elt.width = elt.videoWidth;
+        c.height = c.elt.height = elt.videoHeight;
+      }
       c.loadedmetadata = true;
     });
     return c;
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -214,6 +214,7 @@ p5.Renderer2D.prototype.get = function(x, y, w, h) {
   }
 
   var ctx = this._pInst || this;
+  ctx.loadPixels();
 
   var pd = ctx._pixelDensity;
 
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -188,9 +188,7 @@ p5.Renderer2D.prototype.copy = function () {
 
 p5.Renderer2D._copyHelper =
 function (srcImage, sx, sy, sw, sh, dx, dy, dw, dh) {
-  if (!srcImage.canvas) {
-    srcImage.loadPixels();
-  }
+  srcImage.loadPixels();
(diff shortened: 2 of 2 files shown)


shiffman:
@lmccart while working through this issue I also noticed that your commit 657d46e088948c81ebdfb47297dcbd37c79a3cbc results in an incorrect image size after calling `get()`.   You changed:

```javascript
c.width = elt.videoWidth = elt.width;
c.height = elt.videoHeight = elt.height;
```

to

```javascript
c.width = c.elt.width = elt.videoWidth;
c.height = c.elt.height = elt.videoHeight;
```

And the following example I get a 640x480 image instead of a 320x240 one.

```javascript
  var video = createCapture(VIDEO);
  video.size(320, 240);
  var snapshot = video.get(); 
```

Should I revert back as well?

lmccart:
thanks for looking into this! argh the width/height thing is another one of those blocks of code that gets perpetually changed back and forth based on a few different use cases that break it in different ways. I'm not sure what to do. 🙁

lmccart:
I guess this is the other use case involving createCapture(): https://github.com/processing/p5.js/issues/1490. So I think we want to test both and see if we can come up with a solution that works for both?

shiffman:
@lmccart yeah, heh, let me look into it more and see what i can come up with!

shiffman:
Ok, the issue is that we need the element's canvas to have a width and height whether `size()` is called or not.  When `size()` is called we want those dimensions, when it's not we want the native dimensions.  This seems to work now in both cases:

```javascript
      if (elt.width) {
        c.width = elt.videoWidth = elt.width;
        c.height = elt.videoHeight = elt.height;
      } else {
        c.width = c.elt.width = elt.videoWidth;
        c.height = c.elt.height = elt.videoHeight;
      }
```

I admit to being confused between all the different widths and heights so not sure if the above is still missing a possible use case?

However, there's a new problem where only the first call to `copy()` works and subsequent calls do not update the image being copied onto.   @lmccart if the above seems right to you, I'll add that in and try to determine what the other problem is.

shiffman:
I could definitely use a sanity check on this but I believe I have resolved all the issues including:

1. `get()` works properly after `size()` has been called. #1588 
2. `copy()` works whether or not `size()` has been called. #1490 
3. `copy()` works more than just the first time.

hobzcalvin:
Thanks for looping me in. I was new to the codebase with 8952a7c6869ebeeb385ce5e97adf8efb6fc26dc0 and couldn't run the tests properly, so I'm perfectly willing to believe that removing `loadPixels()` was a bad idea. I think my confusion was that `loadPixels()` doesn't seem to do anything that would affect this code as it's defined here: https://github.com/processing/p5.js/blob/8952a7c6869ebeeb385ce5e97adf8efb6fc26dc0/src/core/p5.Renderer2D.js#L248 That is, it updates `imageData` and `pixels` properties but these properties are completely unused by `get()`. I think what I overlooked is that `get()` in p5.dom.js *does* do substantive stuff that will affect `get()`, namely changing width/height values: https://github.com/processing/p5.js/blob/master/lib/addons/p5.dom.js#L1799

Anyway, not sure if that's helpful or not, but this change seems fine to me. You could perhaps add a comment noting that `loadPixels()` is needed because `p5.MediaElement`'s version does meaningful stuff.

lmccart:
This looks good, thanks for sorting this out @shiffman!

Issue #1617, opened 2016-10-03
Links broken on the README file


pizzels:
The link for the "Learn" list item on the "Learning" section points to the wrong URL. It should link to the tutorials page on the web.

Also does the "contribute section" link under "Get Involved". For what I see all the info to contribute to the project is now on the "Development" wiki page, but maybe the best would be to link to the "Community" page on the website.

pizzels:
I'd like to fix this myself as my first contribution to the project

lmccart:
fixed with https://github.com/processing/p5.js/pull/1618. welcome @pizzels :)

PR #1618, opened 2016-10-03
Fix broken links on the README

pizzels:
A couple of links were leading to a 404 page. They have been updated to match available pages from the website.

I'm not completely sure the new link on the 'Get involved' section is the one that should be there, but seems the most appropiate one.

This fixes issue #1617


lmccart:
thank you!

Issue #1622, opened 2016-10-07
Issues with positioning point lights


mindofmatthew:
Right now, point light positioning doesn't use the same coordinate space as everything else. In the Processing example below, the light is positioned between the two spheres. In the p5 version of the same example, the light is positioned far to the left of the spheres (I'm pretty sure that its effective x-position is -40000, rather than -40).

Processing example:
![lights in Processing](https://cloud.githubusercontent.com/assets/911429/19174689/c7a1b974-8bfe-11e6-9716-ff6745830f23.png)

``` processing
size(200, 100, P3D);

noStroke();
fill(255);
background(0);

ortho();

translate(width / 2, height / 2, 0);

pointLight(255, 255, 255, -40, 0, 0);

sphere(20);

pushMatrix();
translate(-80, 0, 0);
sphere(20);
popMatrix();
```

p5 Example:
![lights in p5.js](https://cloud.githubusercontent.com/assets/911429/19174910/5bd1a6ee-8c00-11e6-8f74-8f087885b02d.png)

``` javascript
function setup() {
  createCanvas(200, 100, WEBGL);
}

function draw() {
  background(0);
  ambientMaterial(255, 255, 255);

  pointLight(255, 255, 255, -40, 0, 0);

  sphere(20);

  push();
  translate(-80, 0, 0);
  sphere(20);
  pop();
}
```

kjhollen:
I think this one will be handled with #2139!

kjhollen:
Just an update to say that this wasn't fixed with the coordinate system like I'd hoped. I'll take a closer look at the lighting shader to try and see what's up.

Spongman:
here's another example that demonstrates this problem: https://codepen.io/Spongman/pen/xpEpqd?editors=0010

the large sphere is lit by a point light at the location of the small sphere. the small sphere is on the x axis. only the right half of the large sphere should be lit. Also note that the lower sphere is lit identically to the other one.

![image](https://user-images.githubusercontent.com/1088194/34290409-07fff252-e6ac-11e7-95d1-760d455a17e8.png)

I believe the problem is caused by the fact that the uPointLightLocation positions are in model space and the rest of the calculation is done in camera. Unfortunately this cannot simply be fixed by multiplying the uPointLightLocation by the MV matrix as this will cause the lights to also be affected by the translate/rotate calls.

i believe the final fix involve separating the view & model matrices.

Spongman:
ok, this is now fixed now that #2428 has been merged.

[...]

PR #1624, opened 2016-10-07
Remove uResolution uniform

### LINKED ISSUE #1621: Inconsistencies between Processing and p5 clipping plane parameters
As I mentioned in #1615, p5 WebGL's camera clipping planes (ie, the "near" and "far" parameters to `perspective()` and `ortho()`) don't quite work as expected. See my test case below. Basically, it draws three spheres at different depths, and if you explicitly set the clipping planes through one of the commented-out camera function calls, only the middle sphere is contained within the renderable area and the other two disappear. This works in Processing, but currently fails in p5.

(I've got a PR in the works that addresses this, but I thought I'd go ahead and clearly document the issue here for posterity.)

Processing test case:
```processing
size(200, 100, P3D);

noStroke();
fill(255);

float fov = PI/3.0;
float cameraZ = (height/2.0) / tan(fov/2.0);

//Uncomment either of these lines and only the middle sphere should be visible
//perspective(fov, float(width)/float(height), cameraZ - 20, cameraZ + 20);
//ortho(-width/2, width/2, -height/2, height/2, cameraZ - 20, cameraZ + 20);

translate(width / 2, height / 2, 0);

sphere(20);

pushMatrix();
translate(-40, 0, 40);
sphere(20);
popMatrix();

pushMatrix();
translate(40, 0, -40);
sphere(20);
popMatrix();
```

p5.js translation:
```javascript
function setup() {
  createCanvas(200, 100, WEBGL);

  var fov = PI/3.0;
  var cameraZ = (height/2.0) / tan(fov/2.0);

  //Uncomment either of these lines and only the middle sphere should be visible
  //perspective(fov, float(width)/float(height), cameraZ - 20, cameraZ + 20);
  //ortho(-width/2, width/2, -height/2, height/2, cameraZ - 20, cameraZ + 20);
}

function draw() {
  background(128);
  
  sphere(20);

  push();
  translate(-40, 0, 40);
  sphere(20);
  pop();
  
  push();
  translate(40, 0, -40);
  sphere(20);
  pop();
}
```

mindofmatthew:
The WebGL renderer and shaders apply a constant scaling factor called "resolution" to (almost) all position values. I can't quite tell what issues this was meant to resolve (most likely, they've been fixed in other ways as well), and in the interest of simplifying our shaders, I think we'd be better off removing it entirely.

This change does affect two existing pieces of API behavior, both of which I consider to be bugs. This PR fixes #1621 outright and slightly improves #1622, though there's still work to be done to fully address that issue.

@indefinit, let me know if you can think of any side-effects to this change that I may have missed. Thanks!

### DIFF
--- a/src/webgl/camera.js
+++ b/src/webgl/camera.js
@@ -148,10 +148,6 @@ p5.prototype.ortho = function(left,right,bottom,top,near,far) {
     args,
       ['Number', 'Number', 'Number', 'Number', 'Number', 'Number']
   );
-  left /= this.width;
-  right /= this.width;
-  top /= this.height;
-  bottom /= this.height;
   this._renderer.uPMatrix = p5.Matrix.identity();
   this._renderer.uPMatrix.ortho(left,right,bottom,top,near,far);
   this._renderer._curCamera = 'custom';
--- a/src/webgl/camera.js
+++ b/src/webgl/camera.js
@@ -64,7 +64,9 @@ p5.prototype.camera = function(x, y, z){
  * //you will see there's a vanish point
  * function setup(){
  *   createCanvas(100, 100, WEBGL);
- *   perspective(60 / 180 * PI, width/height, 0.1, 100);
+ *   var fov = 60 / 180 * PI;
+ *   var cameraZ = (height/2.0) / tan(fov/2.0);
+ *   perspective(60 / 180 * PI, width/height, cameraZ * 0.1, cameraZ * 10);
  * }
  * function draw(){
  *  background(200);
@@ -117,7 +119,7 @@ p5.prototype.perspective = function(fovy,aspect,near,far) {
  * //there's no vanish point
  * function setup(){
  *   createCanvas(100, 100, WEBGL);
- *   ortho(-width/2, width/2, height/2, -height/2, 0.1, 100);
+ *   ortho(-width/2, width/2, height/2, -height/2, 0, 500);
  * }
  * function draw(){
  *  background(200);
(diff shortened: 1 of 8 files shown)


indefinit:
Originally the resolution uniform was meant to handle canvas resolution (width/height).  Most of the time I've seen resolution uniforms inside the pixel shader because of per/pixel operations on textures / calculation of the fragCoord.  However I do agree that it's not really necessary inside of these vert shaders.

mindofmatthew:
I definitely agree that we need a canvas width/height uniform for the reasons you mentioned, but the currently implementation of "uResolution" doesn't really seem to be that (for one thing, it's currently set to a constant value of 1000). My thinking here was to remove that old implementation to clear the way for adding in a new resolution uniform, and I can go ahead and add that in now.

One API question then: the closest thing in Processing is a uniform called "viewport" that's a vec4 with the values (0, 0, <width>, <height>). I'm not sure why it's a vec4—it seems to be mostly used by the wireframe shader. Do we want to implement that, or would it make more sense to implement a separate vec2 "resolution" uniform?

indefinit:
@mindofmatthew I think a vec2 is sufficient but just in case I'm pinging @codeanticode for his weigh-in on vec2 vs. vec4 for viewport res uniform.

codeanticode:
I agree, using a vec2 would make more sense. I don't see any place in the Processing code where a viewport with x and y different from zero could be used, so makes me think if we should replace it by resolution as well.

codeanticode:
went ahead and added a resolution uniform to PShader. Viewport is still supported, but won't be set if cannot be found in the shader code. Use of resolution in pixel shaders is widespread enough for this to be included.

mindofmatthew:
@indefinit As it turns out, I've already added a "resolution" uniform defined as vec2(width, height) to the shader branch (and then decided to go back and sort out the older "uResolution"). It sounds like that uniform is everything we'll want—and that will get added with the rest of the custom shader functionality.

Anything else that needs to happen in this PR before it can be merged?

Issue #1619, opened 2016-10-06
Create a vertex using a vector object


Niels-NTG:
I'm currently refactoring some old personal project of mine, this time trying to be fully object oriented. I was somewhat disappointed that I couldn't do this:

``` javascript
var someVector = new p5.Vector(4, 1);
beginShape();
vertex(someVector);
endShape();
```

Instead I have to do this:

``` javascript
var someVector = new p5.Vector(4, 1);
beginShape();
vertex(someVector.x, someVector.y);
endShape();
```

Does seem like a missed opportunity to me.

joshuaalm:
Not sure how to connect the issue with my commit, but I copied the code from the current vertex() function and made the args variable contain the contents of vertex.array().  Let my education begin.

lmccart:
thanks for the pull request. after looking into this, I'd like to stick to the original signature and not accept p5.Vector as arguments here. I think it opens a can of worms for maintaining consistency across other methods that I'm not prepared to deal with right now.

Niels-NTG:
Understandably. It seemed like a good idea at the time, but since JavaScript doesn't really have proper support for overloading, the code turns into a big mess.

PR #1626, opened 2016-10-08
Add way of using vector objects to create vertexes

Niels-NTG:
I implemented a way to use `p5.Vector` objects for creating vertexes instead of only separate coordinates. This addresses this issue https://github.com/processing/p5.js/issues/1619 I submitted earlier this week.

### DIFF
--- a/src/core/vertex.js
+++ b/src/core/vertex.js
@@ -583,7 +583,8 @@ p5.prototype.quadraticVertex = function(cx, cy, x3, y3) {
  * endShape() functions.
  *
  * @method vertex
- * @param  {Number} x x-coordinate of the vertex
+ * @param  {Number|p5.Vector} x x-coordinate of the vertex, or a p5.Vector
+ * object containing values for both the x- and y-coordinate
  * @param  {Number} y y-coordinate of the vertex
  * @return {Object}   the p5 object
  * @example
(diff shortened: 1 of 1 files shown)


src/core/vertex.js
```diff
@@ -596,40 +597,65 @@ p5.prototype.quadraticVertex = function(cx, cy, x3, y3) {
...
  *
  * @alt
  * 4 black points in a square shape in middle-right of canvas.
  *
  */
 p5.prototype.vertex = function(x, y, moveTo) {
-  var args = new Array(arguments.length);
-  for (var i = 0; i < args.length; ++i) {
-    args[i] = arguments[i];
-  }
```
Niels-NTG: I also took the liberty in removing this somewhat redundant loop.

limzykenneth: I believe this loop is created so that we can have an actual array of arguments (with associated array functions) instead of an array-like object. See [here](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/arguments). Whether it is needed or not, I can't really say although it seem all the code beneath are not using `args`, I can't really say whether it will be needed in the future, is it worth keeping for now?

Niels-NTG: I would say it could easily be reimplemented if needed in the future. As long it is not absolutely necessarily it's best to keep the code not unnecessarily bloated.

Niels-NTG:
Todo:

- [ ] `bezierVertex`
- [ ] `curveVertex`
- [ ] `quadraticVertex`

src/core/vertex.js
limzykenneth: Yeah I agree

lmccart:
I think the line where the arguments are defined needs to be changed, and potentially have the named arguments removed completely. Right now it expects (x, y, moveTo), moveTo being an internally used variable not something the user passes in. But if you pass in (x, y, z) or (p5.Vector) this could get mixed up.

lmccart:
closing this, the updated pr is here: https://github.com/processing/p5.js/pull/1665#issuecomment-264973269

Issue #1579, opened 2016-09-02
URGENT: print() command collides with browser/OSX print() command


golanlevin:
With apologies, I have marked this URGENT because fall classes have just started and this bug is about to slay dozens of new programmers everywhere. 

Something has changed between 0.5.2 and 0.5.3, such that the following code, 

var foo = 0; 
print(foo); 

no longer prints diagnostic data to the console. Instead, in version 0.5.3, it pauses the p5 environment and opens up the operating system's Print dialog, for the printer! I have tested this in OSX 10.9.5 and 10.11.5. I consider this problem urgent because printing data to the console is one of the easiest ways to begin debugging. 

Thank you!
Golan

dribnet:
Confirmed that in 0.5.2 "print" was a p5.js function, but no longer is in 0.5.3. As a workaround, the following can be added to the top of any 0.5.3 sketch:
```js
p5.prototype.print = p5.prototype.println;
```

dribnet:
So `print` was intentionally removed in ecf8ee0e and this was documented in the [0.5.3 release notes](https://github.com/processing/p5.js/releases/tag/0.5.3).

lmccart:
thanks @dribnet, this is correct, it was removed as of 0.5.3. the main reason is that having it in there overwrites the native browser command `window.print()` that opens a print dialog, which could be useful functionality. in addition, we already had `println()` which was essential an alias for the same method. as of 0.5.3, `println()` should be used instead. I'll look into adding a warning message per @golanlevin's suggestion if users do try to call this.

for reference, here's the original discussion of this change.
https://github.com/processing/p5.js/issues/1491

golanlevin:
Thanks for your replies @dribnet & @lmccart.  I'm embarrassed not to have checked the changelog/releasenotes to see that this change was actually _deliberate_ (though I _had_ checked to see whether an issue had been filed :) It's good to know that `println()` and `console.log` (see [https://developer.mozilla.org/en-US/docs/Web/API/Console/log](https://developer.mozilla.org/en-US/docs/Web/API/Console/log), via @brysonian) are working solutions.

So, a lot of my older pedagogic examples use print(), and so do my typing habits :/ I'd like to advise that a warning for print() should be issued in p5.js (but perhaps not in p5.min.js), which says something to the effect of "Did you mean to use println() or console.log()?" I consider it much more likely that someone is mistakenly trying to use `print()` to do some debugging, than they are the rare individual who actually is trying to kick open the printer. 

[...]

PR #1629, opened 2016-10-10
Updated docs from print() to println()

phazonoverload:


lmccart:
hi, thanks for this. however, after some discussion, we have settled on print() rather than println() as it's simpler to write and remember. https://github.com/processing/p5.js/issues/1579

phazonoverload:
But it throws an error? I was suggesting moving the documention inline to what works in practice.

lmccart:
are you using the most recent release (0.5.4). the inline documentation you've modified is directly above the source code which is the function definition for print(), so it should work.

PR #1630, opened 2016-10-10
Fix webkitImageSmoothingEnabled is deprecated

zenozeng:
Fix `'CanvasRenderingContext2D.webkitImageSmoothingEnabled' is deprecated` warning caused by p5.prototype.resizeCanvas.

### DIFF
--- a/src/core/rendering.js
+++ b/src/core/rendering.js
@@ -146,6 +146,11 @@ p5.prototype.resizeCanvas = function (w, h, noRedraw) {
     // save canvas properties
     var props = {};
     for (var key in this.drawingContext) {
+      if (key === 'webkitImageSmoothingEnabled') {
+        if ('imageSmoothingEnabled' in this.drawingContext) {
+          continue;
+        }
+      }
       var val = this.drawingContext[key];
       if (typeof val !== 'object' && typeof val !== 'function') {
         props[key] = val;


src/core/rendering.js
```diff
@@ -146,6 +146,11 @@ p5.prototype.resizeCanvas = function (w, h, noRedraw) {
     // save canvas properties
     var props = {};
     for (var key in this.drawingContext) {
+      if (key === 'webkitImageSmoothingEnabled') {
```
lmccart: since it's deprecated, it seems like we culd just ignore this key completely and continue after this line?

zenozeng:
imageSmoothingEnabled  is supported by Chrome with a webkit prefix until version 29. 

And since noSmooth and smooth support both `imageSmoothingEnabled` and `webkitImageSmoothingEnabled`, I think maybe we should be consistent here. 

```
p5.Renderer2D.prototype.noSmooth = function() {
  if ('imageSmoothingEnabled' in this.drawingContext) {
    this.drawingContext.imageSmoothingEnabled = false;
  }
  else if ('mozImageSmoothingEnabled' in this.drawingContext) {
    this.drawingContext.mozImageSmoothingEnabled = false;
  }
  else if ('webkitImageSmoothingEnabled' in this.drawingContext) {
    this.drawingContext.webkitImageSmoothingEnabled = false;
  }
  else if ('msImageSmoothingEnabled' in this.drawingContext) {
    this.drawingContext.msImageSmoothingEnabled = false;
  }
  return this;
};
```

See also: 
- https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled
- https://github.com/processing/p5.js/blob/7a9d866b40a185c50096a0fc5616c605e0bdcdf9/src/core/p5.Renderer2D.js#L900-L902

sakshamsaxena:
(de-lurking) Are we headed anywhere with this ? :sweat_smile:

lmccart:
:) I'd like to resolve this one and merge it in if it still makes sense. I've neglected to make the time to research the current browser status with the property. feel free to take a look if you'd like, or I will look into it this weekend!

sakshamsaxena:
[The latest living standard spec](https://html.spec.whatwg.org/multipage/canvas.html#image-smoothing) does include this property, and there's an entry about it at [ChromeStatus](https://www.chromestatus.com/features/5639849247768576) too. Since FF51, it has been [unprefixed](https://developer.mozilla.org/en-US/Firefox/Releases/51#Canvas) and [MS IE](https://msdn.microsoft.com/en-us/library/dn265062(v=vs.85).aspx) supports it well.

If I followed the idea of this PR correctly, then at this stage, we might need to update the smooth and noSmooth definitions, and ignore the changes in rendering.js, as this property is indeed supported well now. Makes sense ?

lmccart:
@sakshamsaxena yes, this sounds like the right direection

lmccart:
thank you for this and apologies for the big delay. this has been incorporated here: https://github.com/processing/p5.js/pull/2228

PR #1641, opened 2016-10-21
rgbaToHSLA fix, closes #1640

### LINKED ISSUE #1640: Color conversion to HSL/HSB
Creating an RGB color and then using it´s hue, sat, and lightning for a new color should create a color close to the  orginal RGB one. Like this:

`function setup() {
  createCanvas(300, 200);
  colorMode(RGB);
  var col = color(247,192,31);
  console.log(red(col), green(col), blue(col));
  fill(col);
  rect(100,100,100,100);
  var h = hue(col);
  var s = saturation(col);
  var l = lightness(col);
  var b = brightness(col);
  colorMode(HSL);
  var col2 = color(h,s,l);
  fill(col2);
  rect(200,100,100,100);
  console.log(red(col2), green(col2), blue(col2));
}`

but I get a different one:
247 192 31
224.22448979591837 180.8231292517007 53.775510204081606

Using HSB produce a similar result. Seems that their is an issue with the color conversion routine in P5. When I use the routines from tinyColor.js I get the same RGB color. I tried to debug to see the difference but then I got lost in P5 color handling. I was using: /*! p5.js v0.5.4 October 01, 2016 */ .Time for a real color class in P6 :-)

joecridge:
This was a due to a typo that hadn’t been picked up by our unit tests. Ideally we should have a more comprehensive test suite that covers the various branches of colour conversion functions, but I don’t have time to implement that right now.

### DIFF
--- a/src/color/color_conversion.js
+++ b/src/color/color_conversion.js
@@ -229,7 +229,7 @@ p5.ColorConversion._rgbaToHSLA = function(rgba) {
     if (li < 1) {
       sat = chroma / li;
     } else {
-      sat = chroma / (2 - chroma);
+      sat = chroma / (2 - li);
     }
     if (red === val) {  // Magenta to yellow.
       hue = (green - blue) / chroma;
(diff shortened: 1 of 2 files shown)


lmccart:
aha, great!

PR #1647, opened 2016-10-24
Added DELETE method to HttpDo(), fixes #653

### LINKED ISSUE #653: HttpDo() not calculating content length for PUT
HttpDo() appears not to calculate the content length or pass the content when you use a PUT method. May be true for other methods, I could only test PUT with my application.   A test follows:

    var payload = {};
    payload.bri = 88;
    setLight(thisLight, payload);	

    function setLight(lightNumber, data) {
      var url = 'http://example.com/
      var path = url + lightNumber + '/state';		// assemble the full URL
      var content = JSON.stringify(data);		// convert JSON obj to string

      // HttpDo below sends a request with content length of 0 and no body,
      // according to the JS console in Chrome, and WireShark:

      httpDo( path, 'PUT', content, 'json', function(response) {
        println(response);
      });

     // in contrast, the following jQuery request works perfectly:

    /*
       var request = $.ajax({
       	type: "PUT",					// use the PUT method
       	url: path,						// URL to call
       	data: content,					// body of the request
       	dataType: 'text/json'		// data type of the body
           }, function(response) {		// callback function
      		println('response: ' + response);		// server response
        });
    */

sakshamsaxena:
Went through #653 and reproduced it. Saw that GET, POST, PUT are the only ones considered, and other method or no mention of method defaults to GET. So, added DELETE to the same as well, and made a little stub too to test the same [here](https://github.com/sakshamsaxena/p5-io-test). Thanks to @tigoe!

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -862,7 +862,7 @@ p5.prototype.httpDo = function () {
     for (var i = 1; i < arguments.length; i++) {
       var a = arguments[i];
       if (typeof a === 'string') {
-        if (a === 'GET' || a === 'POST' || a === 'PUT') {
+        if (a === 'GET' || a === 'POST' || a === 'PUT' || a === 'DELETE') {
           method = a;
         } else {
           type = a;


lmccart:
looks good, thanks for working on this!

tigoe:
Just to confirm, this is not yet in the release on the main site, right?  Just downloaded that, and GET, POST, PUT all work with that now, but not DELETE yet. Not urgent, just following up.

lmccart:
correct, not yet. I'll ping this thread again when it's deployed so you can test.

sakshamsaxena:
Hi. It did get merged a while back. Not out yet for production on main site. But you can build from the latest commit. :-)

On 25-Oct-2016 4:44 PM, tigoe <notifications@github.com> wrote:

Just to confirm, this is not yet in the release on the main site, right? Just downloaded that, and GET, POST, PUT all work with that now, but not DELETE yet. Not urgent, just following up.

tigoe:
After some digging, I found the problem that caused #653 (this issue didn't actually solve it, but it did add some useful things, thanks). 

Turns out that when httpDo() formats JSON as a string, it formats true and false as strings, not as values. And strict interpreters of the request will read the JSON as invalid. 

I found this by doing the following to a Philips Hue:

````
  var path = url + lightNumber + '/state';		// assemble the full URL
  var foo = {"on": false};
  httpDo( path, 'PUT', foo, 'json', getLights);
`````

I got back: 
````
{"error":{"type":2,"address":"/lights/3/state","description":"body contains invalid json"}}
````

So I put WireShark on it. When I use JQuery to send the same request, like so:

````
var requestParams = {
    type: "PUT",		// use the PUT method
    url: path,			// URL to call
    data: content,		// body of the request
    dataType: 'text'		// data type of the body
  };

 var request = $.ajax(requestParams, getLights);
}
````

It works, and the Wireshark capture shows this for the body:
````
{"on":false}
````

But when I do it with the p5.js at the top, I get:
````
HTML Form URL Encoded: application/x-www-form-urlencoded
{"on" = "false"}
````

So the question is, how do I send out valid JSON with httpDo()? Is there a way to customize the JSON? Or is this still a bug?

limzykenneth:
@tigoe I'm finishing up #1812, after that is merged in it should work as expected

tigoe:
Great, tag me on it when you're ready and I'll be glad to test it. Thanks!

lmccart:
@tigoe it's been merged if you'd like to pull the latest, otherwise there will be a new release later this week.

Issue #1623, opened 2016-10-07
Add default parameters to perspective() and ortho()


mindofmatthew:
Right now, the WebGL `perspective()` and `ortho()` camera functions require all of their parameters to be set, even though both have well-defined default values, and should be callable without passing in any parameters.

In addition, the documentation for both of these methods needs to be expanded.

hkirat:
@mindofmatthew Right now the ortho function takes 6 parameters that define the window. 
However, it assumes that we are taking a front view ortho projection. 
However, orthographic projections need not necessarily have the view normal alligned to the z axis as assumed by the ortho function.
Shouldn't the parameters taken by the function be as follows : 
{x, y, z} of the camera
{x1, y1, z1} defining the view normal (z axis in this case)
{x2, y2, z2} defining the view up normal (y axis in this case)

PR #1691, opened 2016-12-01
Added default data for ortho

hkirat:
In reference to #1623

### DIFF
--- a/src/webgl/camera.js
+++ b/src/webgl/camera.js
@@ -123,6 +123,12 @@ p5.prototype.perspective = function(fovy,aspect,near,far) {
  *
  */
 p5.prototype.ortho = function(left,right,bottom,top,near,far) {
+  left = left || (-this.width/2);
+  right = right || (this.width/2);
+  bottom = bottom || (-this.height/2);
+  top = top || (this.height/2);
+  near = near || 0;
+  far = far || Math.max(this.width, this.height);
   this._renderer.uPMatrix = p5.Matrix.identity();
   this._renderer.uPMatrix.ortho(left,right,bottom,top,near,far);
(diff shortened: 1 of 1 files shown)


src/webgl/camera.js
```diff
@@ -141,6 +141,12 @@ p5.prototype.perspective = function(fovy,aspect,near,far) {
  *
  */
 p5.prototype.ortho = function(left,right,bottom,top,near,far) {
+  left = arguments[0] = arguments[0] || (-this.width/2);
```
lmccart: these could be written more simply as:
```left = left || (-this.width/2);```

hkirat: Might create a problem later on if I do this
https://github.com/processing/p5.js/blob/master/src/webgl/camera.js#L146

lmccart:
ah I see. I've just removed all the validateParameters code so I think we can go with the simpler `left = left || (-this.width/2);`

hkirat:
There might be another problem now.
If I pass a string as a parameter, 
left = left || (-this.width/2);
will retain the string parameter.
This might create a problem later on.

lmccart:
only numbers are intended to be passed into this function. we may attempt parameter type validation again at a later date but for now it was too sporadic, buggy, and slowing down the code so I had to remove it until someone has the bandwidth to give it a focused look. it is fine to proceed with this function with the assumption only numbers will be passed in.

hkirat:
Great!
I'll modify the code and remove the merge conflicts.

hkirat:
Should be good to merge @lmccart

lmccart:
wonderful, thank you

jpsxlr8:
.

Issue #1681, opened 2016-11-29
randomSeed does not reset the "previous"-variable used by randomGaussian


highyield:
The randomGaussian-function used in the current version of p5.js is using the polar form of the box muller transformation. In the implementation a boolean called "previous" is used to return a value calculated in the previous call of randomGaussian.

Now the problem with this is that randomSeed is not resetting this "previous"-variable to false. As a result in some cases the first time you call randomGaussian() after calling randomSeed(seed), you retrieve a number that has been calculated on the basis of the previously calculated seed (because "previous" is still set to true) and not on the new one.

I do not know if this is intentional or not, but it can lead to major problems when you expect to get the same random-numbers every time after you have called randomSeed(seed). This has only happened to me in rare cases however and for the bug to show up it relies on how often you call randomGaussian() before calling randomSeed()

shiffman:
@lmccart I believe I implemented this originally and will take a look.

lmccart:
fixed with https://github.com/processing/p5.js/pull/1697

PR #1697, opened 2016-12-03
fixes seeding for randomGaussian() #1681

shiffman:
`randomGaussian()` does not produce consistent values with `randomSeed()` unless `previous` is reset to `false` every time it's seeded.


lmccart:
thanks!

PR #1698, opened 2016-12-04
Fixes an issue where get() return single pixel when caller asks for image

shiffman:
If you ask for a section of the image like in the following code, when the video is not ready, `get()` now returns a single pixel which breaks a call to `image()`.  This could be solved with error handling in the end-user's code, but I'm adding something where `get()` returns a blank image until it's ready.  

```javascript
  var img = video.get(x, y, w, h);
  image(img, 0, 0);
```

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1839,6 +1839,8 @@
       return p5.Renderer2D.prototype.get.call(this, x, y, w, h);
     } else if (!x) {
       return new p5.Image(1, 1);
+    } else if (x && w > 1) {
+      return new p5.Image(x, y, w, h);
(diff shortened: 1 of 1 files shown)


shiffman:
This line:

```javascript
} else if (!x) {
```

should also probably be:

```javascript
} else if (x === undefined) {
```

or

```javascript
} else if (arguments.length === 0) {
```

Since when `x` equals `0` it returns false, but `0` is a valid value for `x`.

Let me know what you prefer @lmccart.

lmccart:
this check should be 
```javascript
else if (typeof x === 'undefined')
```

I'll make the change right now, thanks for this!

PR #1704, opened 2016-12-12
Fix for misusing println() instead of print()

sbagdat:
Replace Java's println() method with p5's built-in print() method.


lmccart:
thanks!

PR #1729, opened 2016-12-29
Implement fillRule - NONZERO or EVENODD

fogleman:
This PR adds a `fillRule()` function and two constants, `NONZERO` and `EVENODD`, for controlling the fill rule used during fill operations. It is only implemented in the 2D context.

First PR for p5.js, let me know if more is needed.

Nonzero (default):

nonzero

Evenodd:

evenodd

### DIFF
--- a/src/core/attributes.js
+++ b/src/core/attributes.js
@@ -326,4 +326,21 @@ p5.prototype.strokeWeight = function(w) {
   return this;
 };
 
+/**
+ * Sets the fill rule used when filling paths, either NONZERO or EVENODD.
+ * The default fill rule is NONZERO.
+ *
+ * @method fillRule
+ * @param  {Number/Constant} fillRule either NONZERO, EVENODD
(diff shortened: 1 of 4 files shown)


lmccart:
awesome! this is useful but beyond the scope of the p5.js library at this time. perhaps this could be the basis of creating an addon library with some additional drawing helper functions. (if you're interested in that, see more info [here](https://github.com/processing/p5.js/wiki/Libraries))

in general, it's best to make an issue about a new feature before implementing so we don't have to close PRs unexpectedly.

fogleman:
Really? Fill rule is a very common feature in any drawing library. Indeed, it's just an extra parameter to the `CanvasRenderingContext2D.fill()` function. Why not expose that in p5.js? I don't see it being more at home in an add-on library. Can you explain your rationale for it being out of scope?

subhog:
Are there any news about this feature?

Issue #1707, opened 2016-12-16
can I has multidimensional support in dist()


Smilebags:
I am looking at using the dist function in finding the theoretical distance between multidimensional points much like @shiffman used in his video on object tracking by using rgb channels as dimensions in dist().

It would be great to have 4,5,6 (read n)-dimensional support in the function, allowing it to be used in more abstract ways.

I am happy to implement the change in a branch if I could have a discussion with someone about the best way to do this without interfering with the current functionality of the function. I propose the function looks at the number of (dont know the correct term) properties of the objects passed in before deciding the dimension to calculate the distance in.

I am not at home so not sure if this is how the function works but I thought I would throw the idea in there and offer my hand to get it working.

limzykenneth:
@Smilebags Not sure how I missed this but I think this is an interesting idea. I'm not sure if multidimensional distance will be useful enough to be included in the core p5.js library but it definitely can be a useful addon library for some people.

If you are still interested in working on this I don't mind working out some of the details with you. As javascript doesn't have function overloading so yes, looking at the number of arguments passed in is one of the ways to achieve the same effect.

Smilebags:
Yeah, I came to the same conclusion myself; it seems like after looking at the way the code works with the vectors etc, this would be more like an optional extra, but I thought I may as well offer it if I'm going to be writing it anyway.

The conclusion I came to is that it isn't going to play nice with the standard p5 vector, since that only has the 3 dimensions in it. I also noticed the distance function calls the vector subtract function.

The thing I would like to ask is what would be the best format for the object being passed in? Just an object with n properties is assumed to have that many dimensions?

limzykenneth:
If you are thinking of creating an addon library, it would be best to not mess with the existing `dist` function unless you want to optimize the algorithm (but the exposed functionality should not change). It will be much easier to have a create a new function such as `multiDist` or `dist4D`, etc. 

Which leads onto the next point which is you have a few options:
1. Have one function that looks at how many arguments are passed to it and calculate the corresponding distance.
2. Have several functions, one for each dimension, ie. `dist4D`, `dist5D` etc. That way it's a bit more explicit and simpler to implement.

The two above are potentially very messy, ie. distance between 5D vectors will need 10 arguments.

3. You can create new vector objects that can be 4D, 5D or even nD, and write a generic `dist` function to calculate nD distance (you will need to figure out the maths here). This way you can also easily implement other multi-dimensional maths if you so wish.

[...]

PR #1735, opened 2017-01-01
Multidimensional Support for dist()

max0410:
Dist() function now returns the euclidean distance of the inputs, allowing the function to calculate any dimensional distance.

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -118,16 +118,18 @@ p5.prototype.constrain = function(n, low, high) {
 };
 
 /**
- * Calculates the distance between two points.
+ * Calculates the euclidean distance between two points.
  *
  * @method dist
  * @param  {Number} x1 x-coordinate of the first point
- * @param  {Number} y1 y-coordinate of the first point
+ * @param  {Number} [y1] y-coordinate of the first point
  * @param  {Number} [z1] z-coordinate of the first point
+ * @param  {Number} [...] any spacial corordiante of the first point
  * @param  {Number} x2 x-coordinate of the second point
- * @param  {Number} y2 y-coordinate of the second point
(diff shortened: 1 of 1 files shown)


lmccart:
looks great, thanks!

lmccart:
I'm sorry, I had to roll back these changes because it didn't pass the unit tests. Can you please try running `grunt` and confirming all the tests are successful, then create a new PR? Thanks! More info about grunt [here](https://github.com/processing/p5.js/wiki/development).

max0410:
Will do!

Smilebags:
Not sure if this was in response to my question at #1707 or not, but seems to be addressing the same feature request. I just thought I would notify you that I've created a repo to be used as an addon, as suggested by @limzykenneth. https://github.com/Smilebags/p5.dimensions.js

The code doesn't seem as clean as yours, @max0410, so if you wanted to implement more efficient code in my repo I'd be more than willing to merge it. It might help the unit test failures here, since I'm not changing any p5 core functions.

P.S. is creating a new repo the right way to create an addon for p5?? I'm new to all of this.

max0410:
Sure, I'd love to implement it into your repo! Though I would still have to edit the unit tests, because the old unit tests are configured to the "Math.hypot()" function. The only problem is that my code can't pass the overflowing and under flowing tests, though I don't believe they are that important. What do you think about the under and over flowing tests before I make a Pull Request for your repo?

max0410:
@Smilebags I created the pull request, I shortened both the distance function, and the vector object initiator.

Smilebags:
Yeah, that's one of the reasons I decided to create an addon rather than pushing to p5.

Thanks for the PR, I will have a look now.

Issue #1728, opened 2016-12-29
p5.Vector.prototype.rotate(a) isn't rotating the correct angle when using `angleMode(DEGREES)`


jbcrowley:
If I use `v.rotate()` using radians (default) it works fine. If I use `angleMode(DEGREES)` and rotate by the same angle in degrees, it doesn't. The code below demonstrates the issue. You will have to comment/uncomment the relevant lines below. Run in radians mode, the lines look like 9:00. When run in degrees mode, it looks like 10:00.

    function setup() {
      createCanvas(300, 300);
      strokeWeight(3);
  
      var v = createVector(0, 1);
      v.setMag(10);
      var x1 = 50;
      var y1 = 50;
      var x2 = x1 + v.x;
      var y2 = y1 + v.y;
      line(x1, y1, x2, y2);
      x1 = x2;
      y1 = y2;

      print(radians(90));
      print(HALF_PI);
  
      //v.rotate(HALF_PI);

      angleMode(DEGREES);
      v.rotate(90); // <-- bug is here when using DEGREES
  
      stroke(255, 0, 0);
      line(x1, y1, x1 + v.x, y1 + v.y);
      noLoop();
    }

I looked at the code in p5.Vector.js and nothing jumped out at me. I checked `radians()` to make sure it was returning the correct value, `1.57...` and it was.

DonKarlssonSan:
I agree.

I made an executable [repro on CodePen](http://codepen.io/DonKarlssonSan/pen/vgYYpR) with values from the [documentation for rotate()](https://p5js.org/reference/#/p5.Vector/rotate)
````
  angleMode(DEGREES);
  var v2 = createVector(10.0, 20.0);
  v2.rotate(90);
  console.log(v2.x, v2.y, v2.z);
````
expected: [-20.0, 9.999999, 0.0]
actual: [-12.682861232479269 18.415890718552674 0]

It looks like the unit tests for Vector was commented out in [this commit](https://github.com/processing/p5.js/commit/62e8a966f68f2ab6fe31019abe7a939982010a67)

DonKarlssonSan:
Yes, I agree.

I think I know where the bug is: in rotate() when using degrees, a is converted to radians. Then when newHeading is calculated with
`var newHeading = this.heading() + a;`
heading() returns degrees while a is radians => ERROR!

lmccart:
fixed with https://github.com/processing/p5.js/pull/1738

PR #1738, opened 2017-01-03
Fixing Vector.rotate() #1728

DonKarlssonSan:
The problem was that angles in both radians and degrees were mixed.

Adding unit test also


lmccart:
thanks!

PR #1743, opened 2017-01-05
Adds default parameters for Perspective, Closes #1623

### LINKED ISSUE #1623: Add default parameters to perspective() and ortho()

hkirat:


lmccart:
thanks!

PR #1748, opened 2017-01-07
Edited tests to euclidean function

max0410:


test/unit/math/calculation.js
```diff
@@ -57,27 +57,19 @@ suite('Calculation', function() {
...
       test('should return correct distance', function() {
         result = dist(0, 0, 0, 2, 3, 5);
-        assert.equal(result, 6.164414002968977); // Math.hypot(2, 3, 5)
+        assert.equal(result, 6.164414002968976); 
       });
       test('should return positive  distance', function() {
         result = dist(0, 0, 0, -2, -3, 5);
-        assert.equal(result, 6.164414002968977); // Math.hypot(2, 3, 5)
-      });
-      test('should not underflow', function() {
```
lmccart: I think we want to have these two tests in there, and make sure that the function behaves similarly to Math.hypot() (which does not over or under flow)

lmccart:
closing this for now, please reopen if you are able to resolve the comment with the tests. thanks!

Issue #1755, opened 2017-01-09
Using saveCanvas() within mouseClicked() creates an infinite loop that downloads a bunch of files


alterebro:
Hi,
there is some problem with the `saveCanvas()` function that creates a loop when it is called inside the `mouseClicked()` function.
Chrome and Opera (OSX 10.11) download files until the browser throws a `'Maximum call stack size exceeded'` error message but Firefox (OSX and Ubuntu) doesn't stop the execution of the script, it downloads a huge amount of files, it freezes and then gets stuck.
Safari doesn't have the problem as it opens the image in different window.

This is the code that creates the issue :

```js
function mouseClicked() {
	saveCanvas('my_image', 'jpg');
}
```

The following codepen using the latest version of the p5 library (0.5.6) exposes the problem : http://codepen.io/anon/pen/LxpXPa?editors=0010

juansrx:
@alterebro : From this : link https://p5js.org/reference/#/p5/mouseClicked it seems that not all browsers handle this function the same, but they suggest you can "return false" at the end of the method to avoid default behaviors. I tried it on your example and it works (Chromium Version 55.0.2883.87 Built on Ubuntu , running on Ubuntu 16.04 made only one download). I hope this can help you.

alterebro:
Hola @juansrx ,
thanks for the feedback and pointing out the 'return false' thing, that certainly fixes the multiple files download issue which is cool but keeps the loop running :(

lmccart:
fixed in https://github.com/processing/p5.js/pull/1768

PR #1768, opened 2017-01-17
avoids infinite loop when saving canvas within mousepressed #1755

alterebro:
Fix to avoid the loop described in the issue #1755


lmccart:
thanks!

Issue #1769, opened 2017-01-19
keyIsPressed Issue


jkrobertson88:
So it seems to recognize when the a single key is still being pressed but not necessarily that a key is still being pressed. 

Example:
I press the up arrow it returns true while I am holding the up arrow down.
While the up arrow is being pressed I press the down arrow.
I release the up arrow but maintain the down arrow being pressed down. 
keyIsPressed now returns false even though the down arrow is pressed.

lmccart:
can you tell me which browser you're using?

jkrobertson88:
Chrome Version 55.0.2883.87 m

tstefanich:
I think this may be a problem with the _keyup function 

```
p5.prototype._onkeyup = function (e) {
  var keyReleased = this.keyReleased || window.keyReleased;
  this._setProperty('isKeyPressed', false);
  this._setProperty('keyIsPressed', false);
  this._setProperty('_lastKeyCodeTyped', null);
  downKeys[e.which] = false;
  //delete this._downKeys[e.which];
  var key = String.fromCharCode(e.which);
  if (!key) {
    key = e.which;
  }
  this._setProperty('key', key);
  this._setProperty('keyCode', e.which);
  if (typeof keyReleased === 'function') {
    var executeDefault = keyReleased(e);
    if(executeDefault === false) {
      e.preventDefault();
    }
  }
};
```

You can see on the 3rd line of the function that it always sets keypressed to false always on key up but it doesn't check to see if any other buttons on still being pressed. I did a small amount of testing and this started to get the correct behavior.  You can see I moved 3 lines of code to the bottom of the function and checked against other key presses. You can see what I added after //TJS 

```
p5.prototype._onkeyup = function (e) {
  var keyReleased = this.keyReleased || window.keyReleased;
  var keysStillPressed = false;
  downKeys[e.which] = false;
  //delete this._downKeys[e.which];
  var key = String.fromCharCode(e.which);
  if (!key) {
    key = e.which;
  }
  this._setProperty('key', key);
  this._setProperty('keyCode', e.which);
  if (typeof keyReleased === 'function') {
    var executeDefault = keyReleased(e);
    if(executeDefault === false) {
      e.preventDefault();
    }
  }

  // TJS
  for (var i in downKeys) {
    console.log(downKeys[i])
    if (downKeys[i] === true) {
        console.log('test');
        keysStillPressed = true;
        //break;
    }
   }
   if(keysStillPressed == false){
[...]

PR #1783, opened 2017-01-27
Check all keys are released before setting keyIsPressed to false. #1769

aemonm:
Adds a function to check all keys are depressed before setting the keyPressed flags to false. Applied the function to onKeyUp to fix issue #1769.

### DIFF
--- a/src/events/keyboard.js
+++ b/src/events/keyboard.js
@@ -225,12 +225,16 @@ p5.prototype._onkeydown = function (e) {
  */
 p5.prototype._onkeyup = function (e) {
   var keyReleased = this.keyReleased || window.keyReleased;
-  this._setProperty('isKeyPressed', false);
-  this._setProperty('keyIsPressed', false);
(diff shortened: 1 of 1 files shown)


src/events/keyboard.js
```diff
@@ -354,4 +358,22 @@ p5.prototype.keyIsDown = function(code) {
...
+/**
+ * The checkDownKeys function returns a boolean true if any keys pressed
+ * and a false if no keys are currently pressed.
+
+ * Helps avoid instances where a multiple keys are pressed simultaneously and
+ * releasing a single key will then switch the
+ * keyIsPressed property to true.
+ *
+**/
+p5.prototype.areDownKeys = function() {
```
lmccart: I think this could be an internal function rather than attached to p5.prototype

```javascript
function areDownKeys() {
  // etc
```

aemonm: Good call. Changed.

lmccart:
thanks!

PR #1787, opened 2017-01-28
Update mouse.js

dabe:
The property "touches" of the TouchEnd event is empty (tested with chrome). In this case the function getMousePos raises an exception. So this modification checks if touches contains elements.


lmccart:
thanks!

PR #1789, opened 2017-01-31
Cursor function not working as expected #1773

### LINKED ISSUE #1773: CSS cursors are not working
CSS default cursors, like explained here:

http://www.w3schools.com/cssref/pr_class_cursor.asp

Are not working.

limzykenneth:
Fix #1773 The equality test was written as an inequality test so it is always incorrect when a string is passed in. Added a `https://` rule as well.


limzykenneth:
#1773 is a fix, my bad

PR #1793, opened 2017-02-04
Handle ‘numeric’ strings for 2D primitives, closes #1792

### LINKED ISSUE #1792: Odd rendering when passing string width/height values to ellipse()
Hello,

When passing a string argument for width and/or height, the ellipse() method interprets the string as a very high value, instead of parsing it to an integer. 

Here is an example with a comparison to the rect() method, which does not have this problem : 
image

joecridge:
- **Modifies `canvas.modeAdjust()` and `canvas.arcModeAdjust()` so that they always implicitly convert their arguments to numbers**  
This closes #1792 by ensuring that the calculations within the 2D primitive functions (which start by calling mode adjust) deal only with actual numbers. It also fixes some mode adjust-specific quirks that you might currently see if you pick unfortunate combinations of string arguments.  

- **Adds some mode adjust unit tests**  
Mode adjust isn’t currently tested because it resides in it’s own ‘canvas’ module that doesn’t get exported to `p5`. I had to export the mode adjust functions (e.g. `p5.prototype._modeAdjust()`) so that they could be tested for now (not sure if there is a better alternative for doing this?), but the tests cover the basic behaviour as well as the string argument handling.

### DIFF
--- a/src/core/2d_primitives.js
+++ b/src/core/2d_primitives.js
@@ -13,6 +13,10 @@ var constants = require('./constants');
 var canvas = require('./canvas');
 require('./error_helpers');
 
+// Make mode adjust functions available for testing.
+p5.prototype._modeAdjust = canvas.modeAdjust;
+p5.prototype._arcModeAdjust = canvas.arcModeAdjust;
+
 /**
  * Draw an arc to the screen. If called with only a, b, c, d, start, and
  * stop, the arc will be drawn as an open pie. If mode is provided, the arc
--- a/src/core/canvas.js
+++ b/src/core/canvas.js
@@ -8,25 +8,25 @@ module.exports = {
 
   modeAdjust: function(a, b, c, d, mode) {
     if (mode === constants.CORNER) {
-      return { x: a, y: b, w: c, h: d };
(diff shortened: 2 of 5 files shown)


src/core/2d_primitives.js
```diff
@@ -13,6 +13,10 @@ var constants = require('./constants');
 var canvas = require('./canvas');
 require('./error_helpers');
 
+// Make mode adjust functions available for testing.
+p5.prototype._modeAdjust = canvas.modeAdjust;
+p5.prototype._arcModeAdjust = canvas.arcModeAdjust;
```
lmccart: is this just being attached to p5.prototype for testing purposes? is there another way to do this? I'd rather avoid adding anything to the prototype that isn't required for functioning of the library, just to avoid bloat of the object.

joecridge: Yeah it’s just for testing… I spent quite a long time trying to find a more elegant solution but ultimately gave up. If anyone can think of an alternative that would be great!

lmccart:
closing this per discussion and plan to instead focus on friendly error checking of arguments for every function: https://github.com/processing/p5.js/issues/1792#issuecomment-300564863

PR #1794, opened 2017-02-05
Fixes textBound() bug on fonts that don't have the proper Glyph Name
Tables.

tstefanich:
Fixes textBound() bug on fonts that don't have the proper Glyph Name
Tables. Added to a conditional inside of textBounds to either look for
the glyph.name ‘space’ && glyph.unicode 32 to insure proper
calculations of the text bounding box when spaces are used.


dhowe:
looks good to me... thanks @tstefanich !

Issue #1784, opened 2017-01-27
grunt hangs on "mocha:yui" (mocha) task on OSX


kjhollen:
Hi folks! First time building the library from source today in a while. I ran into some issues running grunt—the last thing I see is this:

```
Running "mocha:yui" (mocha) task
Testing: http://localhost:9001/test/test-reference.html
```

and then the build hangs. I let this run for 10 minutes before stopping it. Another participant in the workshop in LA ran into this issue and Lauren said it cropped up last week in NY, too.

More details from a verbose run of grunt:

```
Running "mocha:yui" (mocha) task
Verifying property mocha.yui exists in config...OK
File: [no files]
Options: log, reporter="Nyan", timeout=5000, inject="/Users/kjhollen/src/p5.js/node_modules/grunt-mocha/phantomjs/bridge.js", phantomScript="/Users/kjhollen/src/p5.js/node_modules/grunt-mocha/phantomjs/main.js", urls=["http://localhost:9001/test/test-reference.html"], bail=false, logErrors, growlOnSuccess, run=false
Options: {
  "log": true,
  "reporter": "Nyan",
  "timeout": 5000,
  "inject": "/Users/kjhollen/src/p5.js/node_modules/grunt-mocha/phantomjs/bridge.js",
  "phantomScript": "/Users/kjhollen/src/p5.js/node_modules/grunt-mocha/phantomjs/main.js",
  "urls": [
    "http://localhost:9001/test/test-reference.html"
  ],
  "bail": false,
  "logErrors": true,
  "growlOnSuccess": true,
  "run": false
}
Phantom options: {
  "timeout": 5000,
  "inject": "/Users/kjhollen/src/p5.js/node_modules/grunt-mocha/phantomjs/bridge.js",
  "phantomScript": "/Users/kjhollen/src/p5.js/node_modules/grunt-mocha/phantomjs/main.js",
  "logErrors": true,
  "growlOnSuccess": true,
  "run": false
}
Testing: http://localhost:9001/test/test-reference.html
```

Here are some version numbers that may or may not be relevant:
grunt-cli v1.2.0
grunt v0.4.5
npm 3.10.8

Let me know if there are other helpful details I can provide!

aemonm:
Hi - You need to open that URL in a browser - http://localhost:9001/test/test-reference.html - for the tests to run to completion.

limzykenneth:
[...]

PR #1804, opened 2017-02-10
Implement fillRule - NONZERO or EVENODD

fogleman:
This PR adds a fillRule() function and two constants, NONZERO and EVENODD, for controlling the fill rule used during fill operations.

@lmccart I have added an example and a unit test for each fill rule. Unfortunately I'm unable to run the docs locally. I get this when following the instructions:

screen shot 2017-02-09 at 8 50 26 pm

(I'm also running into issue #1784, where grunt hangs on mocha:yui) 

For the unit tests, I figured we'd want to do an image comparison like those in `test/unit/core/2d_primitives.js` - but then I discovered these tests are not being run - it's commented out in `test.html`. I left my test in, but it won't be run because of this. (And since I can't view the docs to see my examples, I don't have ground-truth imagery to compare to yet.)

### DIFF
--- a/src/core/attributes.js
+++ b/src/core/attributes.js
@@ -332,7 +332,39 @@ p5.prototype.strokeWeight = function(w) {
  *
  * @method fillRule
  * @param  {Number/Constant} fillRule either NONZERO, EVENODD
- * @return {p5}                   the p5 object
+ * @return {p5}                       the p5 object
+ * @example
+ * <div>
+ * <code>
+ * noStroke();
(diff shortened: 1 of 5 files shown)


lmccart:
Thanks. Can you also check in the image files for testing? 
unit/assets/renders/evenodd.png
unit/assets/renders/nonzero.png

kindanduseful:
Hi, was this merged or integrated? Thanks!

Issue #1694, opened 2016-12-02
replace reqwest with fetch API?


lmccart:
per @brysonian's [comment](https://github.com/processing/p5.js/issues/968#issuecomment-264197472)

look into replacing reqwest with [fetch api](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). this may make things smoother with load functions and get/put/post functions, and remove need to rely on an external library.

limzykenneth:
Is anyone currently working on this? If not, I have some free time to look into this.

lmccart:
go for it!

lmccart:
that would be great, go for it!

giltayar:
@limzykenneth - did you take this? I would love to work on it.

limzykenneth:
@giltayar I haven't started on this yet but going to soon. I'm quite happy to work on this with you if you want. I haven't have a closer look yet so I'm not sure how to break this down though.

giltayar:
Nope - I finally didn't take it. All yours...

limzykenneth:
haha, alright 😄

limzykenneth:
I need some second opinion on a particular implementation. I have managed to very basically replace reqwest with fetchAPI calls and now I'm going through each functions to make them feature complete and add the header fields etc. I'm starting on `loadJSON()` and there are a couple problems.

First, fetch API doesn't support JSONP out of the box so in order to support that, a polyfill will be required. Second, from #968 the purpose of this change is also to make passing headers easier, it is pretty trivial to provide header fields and even POST body fields to the request. I just need a decision on what the valid signature for the function calls will be. Currently, `loadJSON()` has the following singature `loadJSON(path,[callback],[errorCallback],[datatype])`, I'm just wondering how to add the additional fields into the signature, we may be looking at a breaking change? (Also datatype won't be needed in this case if `loadJSON()` doesn't support JSONP).

lmccart:
closed with https://github.com/processing/p5.js/pull/1812

Issue #968, opened 2015-10-04
Setting headers for httpGet() / httpDo()


shiffman:
This I think follows up on the discussion in #385.   We added `httpGet()`, `httpPost()`, and `httpDo()` to handle generic HTTP Requests and the functions are working great for me.  For one of my examples for [A2Z week 4](https://github.com/shiffman/A2Z-F15/tree/gh-pages/week4) I need to add a header to the request for authentication.  I'm able to get this to work with JQuery as follows (the API here is [Clarifai](http://www.clarifai.com/)).

``` javascript
  $.ajax({
    url: 'https://api.clarifai.com/v1/tag/',
    type: 'GET',
    beforeSend: function(xhr) {
      xhr.setRequestHeader('Authorization', 'Bearer ACCESS_TOKEN');
    },
    data: {
      url: 'http://www.clarifai.com/img/metro-north.jpg'
    },
    success: function (response) { 
      console.log(response);
    },
    error: function (err) { 
      console.log(err);
    },
  });
```

`httpDo()` does everything above with the exception of `beforeSend`.   My question is:

Is an optional argument for headers something we should add to our API or is this a scenario beyond the scope of p5?

lmccart:
I think potentially we could add it only to `httpDo()` which is already our base case http method (the put and get methods call this under the hood). we could also think about supporting passing in an arg object similar to jquery just for this one. but I do worry that might be opening a large can of wormy api discussions...

brysonian:
Maybe passing an object just treats it as a direct wrapper for `reqwest`? I also notice it is missing an error callback, if something goes wrong in the request it'll be really hard to debug. Should we add another callback? Break existing api and pass the first arg as err (which is consistent with node-style) or maybe just pass a second success arg?

shiffman:
I think if reqwest accepts an object, this could be something of a hidden advanced feature, i.e. if you want these advanced features just make the object yourself and pass it in.  We could do this with `httpDo()` only as @lmccart suggests.  

I think having an optional "error" callback would be useful.  Is there a discussion about this with `loadJSON()` or does this already exist and I just missed it (as often seems to be the case)?

I can open a separate issue, and would suggest something like:

```javascript
loadJSON(url, callback); 
loadJSON(url, callback, errorCallback);  // 3rd arg optional
```

lmccart:
I think this makes sense!

vorg:
[...]

PR #1812, opened 2017-02-18
Fetch API implementation 

limzykenneth:
Work in progress. **Do not merge this yet!**

As per discussed in #1694 and #968, this PR replaces all dependencies of reqwest with the new [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch), using [github's polyfill](https://github.com/github/fetch) where it is not available.

So far, `httpDo` `httpGet` `httpPost` `loadJSON` `loadXML` `loadTable` are all using the Fetch API. 
- everything here resolves to `httpDo` calls, hopefully that can help with unifying the API.
- for `loadJSON`, JSONP is not supported. A polyfill to implement JSONP support will be required if it is deemed necessary.
- `loadStrings` can be adapted from its current XMLHttpRequest to using Fetch API if desired. (It will make the source simpler I think)
- to get more advanced features (pass in headers, change cors rules etc) will require passing in a `Request` object to `httpDo`, syntax as follows:
```javascript
var myRequest = new Request("./pathToFile.txt", {
  method: "GET",
  mode: "cors",
  headers: new Headers({
    "Content-Type": "text/plain"
  })
});

httpDo(myRequest, function(data){
  // Work with returned data
}, function(err){
  // optional error callback
});
```
The only difference between the above and using the Fetch API directly is the above uses callback instead of Fetch API's promise and this utilises the automatic type check for JSON, XML and all others as text so the user don't need to worry about extra parsing steps.

---

I have yet to fully edit the tests and documentation yet as I think there are a lot of ways to implement this and I want the decision more or less made before those are changed. Do take some time to have a look. Let me know if there are some other preferred way some things are done, suggestion on function signatures, use cases, complete overhauls etc.

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -275,30 +275,6 @@ p5.prototype.loadJSON = function () {
     }
   }, errorCallback);
 
-  // fetch(path)
-  //   .then(function(res){
-  //     if(res.ok){
-  //       return res.json();
-  //     }
-
-  //     if (errorCallback) {
-  //       errorCallback(res);
-  //     } else { // otherwise log error msg
-  //       throw new Error(res.statusText);
-  //     }
-  //   })
-  //   .then(function(resp){
-  //     for (var k in resp) {
-  //       ret[k] = resp[k];
-  //     }
-  //     if (typeof callback !== 'undefined') {
-  //       callback(resp);
-  //     }
-  //     if (decrementPreload && (callback !== decrementPreload)) {
-  //       decrementPreload();
-  //     }
-  //   });
-
   return ret;
 };
 
@@ -757,33 +733,6 @@ p5.prototype.loadXML = function (path, callback, errorCallback) {
     }
   }, errorCallback);
 
-  // fetch(path)
-  //   .then(function(res){
-  //     if(res.ok){
-  //       return res.text();
-  //     }
-  //     if (errorCallback) {
-  //       errorCallback(res);
-  //     } else { // otherwise log error msg
-  //       throw new Error(res.statusText);
-  //     }
-  //   })
-  //   .then(function(resp){
-  //     var parser = new DOMParser();
-  //     var intermediate = parser.parseFromString(resp, 'text/xml');
-  //     var xml = parseXML(intermediate.documentElement);
-
-  //     for(var key in xml) {
-  //       ret[key] = xml[key];
-  //     }
-  //     if (typeof callback !== 'undefined') {
-  //       callback(ret);
-  //     }
-  //     if (decrementPreload && (callback !== decrementPreload)) {
-  //       decrementPreload();
-  //     }
-  //   });
-
   return ret;
 };
 
@@ -930,11 +879,13 @@ p5.prototype.httpDo = function () {
       if (typeof a === 'string') {
         if (a === 'GET' || a === 'POST' || a === 'PUT' || a === 'DELETE') {
           method = a;
-        } else {
+        } else if(a === 'json' || a === 'jsonp' || a === 'xml' || a === 'text') {
           type = a;
+        } else {
+          data = a;
         }
       } else if (typeof a === 'object') {
-        data = a;
+        data = JSON.stringify(a);
       } else if (typeof a === 'function') {
         if (!callback) {
           callback = a;
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -245,8 +245,7 @@ p5.prototype.loadJSON = function () {
   var decrementPreload = p5._getDecrementPreload.apply(this, arguments);
 
   var ret = {}; // object needed for preload
-  // assume jsonp for URLs
-  var t = 'json'; //= path.indexOf('http') === -1 ? 'json' : 'jsonp';
+  var t = 'json';
 
   // check for explicit data type argument
   for (var i = 2; i < arguments.length; i++) {
@@ -260,11 +259,17 @@ p5.prototype.loadJSON = function () {
     }
   }
 
+  if(t === 'jsonp'){
+    console.log('JSONP polyfill required');
+    return ret;
+  }
+
   fetch(path)
     .then(function(res){
       if(res.ok){
         return res.json();
       }
+
       if (errorCallback) {
         errorCallback(res);
       } else { // otherwise log error msg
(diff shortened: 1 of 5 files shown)


limzykenneth:
New idea:
use the following signatures
```
httpDo(path, [callback], [errorCallback]);
httpDo(path, method, [data], [datatype], [callback], [errorCallback]);
httpDo(path, requestObject, [callback], [errorCallback]);
```
The `requestObject` parameter is the second parameter passed to the Fetch API's Request object constructor. In a way, this parameter gives `httpDo` a similar syntax to the original Fetch API, potentially making it a better option for beginner's transition to using the native API.

lmccart:
thanks for laying this all out so clearly! here are my thoughts on the questions you raised:

- everything here resolves to httpDo calls, hopefully that can help with unifying the API.
  > Great! Let's adapt loadStrings to use Fetch API and resolve to httpDo as well so it is all uniform
- to get more advanced features (pass in headers, change cors rules etc) will require passing in a Request object to httpDo
  > Sounds good. I like the most updated signatures you proposed.
- for loadJSON, JSONP is not supported. A polyfill to implement JSONP support will be required if it is deemed necessary.
  > I think we should add this in. as I recall instagram uses jsonp? I'm not sure if it still does. I think users will still expect this is possible. However, I'm wondering if this should be collapsed as an option in the requestObject to simplify? In this case, the signatures would look like this:
```
httpDo(path, [callback], [errorCallback]);
httpDo(path, [method], [data], [callback], [errorCallback]);
httpDo(path, requestObject, [callback], [errorCallback]);
```

limzykenneth:
There is a Fetch API JSONP polyfill available [here](https://github.com/camsong/fetch-jsonp), it uses roughly the same patterns as the Fetch API so it should be easier to maintain on our side, I feel this is better than trying to implement a JSONP polyfill myself.

In terms of detecting the user is looking to use JSONP, I have a check in place that will see if then passed in type option is `jsonp` and run different steps if so. We will probably still need the `datatype` option for `httpDo(path, method, [data], [datatype], [callback], [errorCallback]);` unless we want to fully rely on URL type detection. For `httpDo(path, requestObject, [callback], [errorCallback]);` I'm wondering if it is easier to pass in the `type` as a value in the requestObject or let it be another optional parameter after requestObject, if we put it in the requestObject then the signature is not strictly speaking the same as Fetch API's.

limzykenneth:
@lmccart more or less done here, just need decisions on the issues raised above then I can move onto implementing them (or not) and fixing the tests. If you can, have a look at my findings for #1784 and see if you want to do anything about it

lmccart:
Let's use the jsonp polyfill, and add promises polyfill for IE. I realize we're not covering everything but I think we'll get close. Let's keep `datatype` as is for now, meaning don't combine with requestObject. I think that lets use merge this first chunk and then revisit this item later if we want to, instead of having to sort out all the signatures again right now.

limzykenneth:
@lmccart Ready to go on this one, you can merge this whenever. I would like to write more tests for this part of the library but that can be another PR. I'm not expecting a lot of breaking changes, the only one that may break are sketches that uses reqwest options when using `httpDo` so you may want to take note of that when publishing this in a release.

May have bugs or better ways to do things but as you said, better to revisit them later if required, not sure how people are going to use this anyway.

lmccart:
Awesome! I will review in the next day or two. Excited about this!

limzykenneth:
@lmccart Tests covers almost everything that uses the new API now.

lmccart:
this all looks good, thank you so much for all your work on this much needed change! it will be in the next release :)

brysonian:
Yay! Perhaps we should return the promise generated by fetch in case a user wants to tap into that? Would be great for users that understand promises to have that around.

src/io/files.js
```diff
@@ -820,55 +794,103 @@ p5.prototype.httpGet = function () {
...
+
+/**
+ * @method httpDo
+ * @param  {String}        path
+ * @param  {Object}        options   Request object options as documented in the
+ *                                    "fetch" API
+ * <a href="https://developer.mozilla.org/en/docs/Web/API/Fetch_API">reference</a>
+ * @param  {Function}      [callback]
+ * @param  {Function}      [errorCallback]
+ */
```
lmccart: could we add an example(s) using the request object?

limzykenneth: Yeah, sure. I think to highlight what would be a good circumstance to use the request object (needing to set headers), maybe have it request a server that uses some headers. Any API you can think of might be a good example?

limzykenneth:
@brysonian Currently the `loadSomething()` functions all return the data that the user is loading so they can do:
```javascript
var myObject = loadJSON(url);
```
and use `myObject` directly. The only set that can return promises are the `httpDo` family of functions (`httpGet` and `httpPost`) since they don't return data but rather rely exclusively on callbacks for fetched data.

brysonian:
ahh right of course, sorry!

nengine:
An example of httpDo with body and Content-Type: 'application/json' something like below would be greatly appreciated.  I tried with version 0.5.8 and not getting it right. Thanks.

```js
data = {"images": [{ "image": '11111111111111' }] }
"method": "POST"
"body": data,
"headers": {
"Content-Type": "application/json"
}
```

limzykenneth:
@nengine the `body` attribute for sending JSON data needs to be stringified. Stringify literally means convert to a string, that's how JSON data is transferred through the internet and the utility function to do that is `JSON.stringify(object)`. To use the native options of `httpDo` you will need to take care of these yourself, so if you are just sending a simple POST request with data, it is best to just use the parameters provided for you as part of `httpDo` or `httpPost`. Below is an example code that should work with 0.5.8.

```javascript
httpDo(path, {
  method: "post",
  body: JSON.stringify(data),
  headers: new Headers({
    "Content-Type": "application/json"
  })
}, function(data){
  console.log(data);
});
```

**Having said that**, I seemed to have forgotten to do that myself! 🤦‍♂️  @lmccart Because I was not testing POST requests I haven't thought of this problem, if you are planning to release 0.5.9 soon may I suggest to put it on hold for a couple days while I make sure it works correctly?

nengine:
Yayyy! It works now. Many thanks.

DiggidyDale:
Hi Guys,
Wondering if someone can help, I don't understand how we are meant to pass headers into a get request to load a JSON file. I need to grab the just giving API and it keeps failing, i think it ius becasue i need to pass in the headers 'Accept' : 'application/json' but i have no idea how I am meant to do so. Sorry if it is dumb question...also thanks in advance

brysonian:
@DiggidyDale you would need to use httpDo and pass the headers in the options. Something like:

```
httpDo('/path/to/resource', {
  method: 'GET',
  headers: {
    'Content-Type': 'application/json'
});
```

That said, should `loadJSON` set the content-type header by default?

limzykenneth:
@brysonian yes it should and it should be fixed in the repo now or is in a PR pending merger. Should expect it to be fixed in the next version probably.

lmccart:
yes, sorry for the delay! i'm aiming to merge and have a new version in the next few days.

brysonian:
Awesome @limzykenneth, not sure if this is in there too, but maybe return fetch's promise from httpDo and return the xml res as well as calling the cb? That way folks could use the promise syntax if they like. Although maybe people who would do so would just use fetch directly.

limzykenneth:
@brysonian returning a promise is not a bad idea although I feel at that point maybe the user can start exploring using Fetch itself which is extremely easy (Safari just added native support as well so most people won't need a polyfill anymore). I'm quite happy to make those changes but I'm a bit tied up recently so unless someone wants to have a go, I don't want to hold up the release of the next version as there are quite a few significant bug fixes pending the next release.

brysonian:
@limzykenneth makes sense to me on encouraging users to look at fetch. Thanks!

PR #1819, opened 2017-02-23
Update core.js for passive MouseWheel event support

ancula:
Passive event support for Chrome. 

Starting with MouseWheel as Chrome Dev 58.0.3018.3 is suggesting (via Violation text in the console) a possible performance improvement from it.  

Spot checked in Firefox and Edge without issues.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -510,7 +510,7 @@ var p5 = function(sketch, node, sync) {
     var f = this['_on'+e];
     if (f) {
       var m = f.bind(this);
-      window.addEventListener(e, m);
+      window.addEventListener(e, m, {passive: true});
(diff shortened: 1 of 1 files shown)


feedzh:
This is very bad solution and break other events.

It turns all events as passive events.

ancula:
OKay, I'll trap for just the mousewheel

feedzh:
On the line 185 is the object of this array. Perhaps you should add a second object like this._passiveEvents={'wheel':true} and then check if it exists in the loop.

feedzh:
You should also check if client supports passive events: https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md

Issue #1815, opened 2017-02-22
mousePressed and touchStarted broken in chrome mobile


reijovosu:
Hello,

On the issue #1811 there is problem when counter counts double in chrome. Although the solution was a hack there is bigger underlining issue.

Chrome changed how they handle touchstart. To make scrolling faster they now ignore e.preventDefault inside touchstart event. And just tapping is firing mouse and touch events.

At the beginning I thought to prevent this happening, just delete mousePressed from touchStarted event. This will brake mobile safari instead. That fires touchstart and mousedown as a bubbled event.

So on this case I think this is more underlining issue of the framework, that I don't have authorization of fixing.

For reading:
https://developers.google.com/web/updates/2017/01/scrolling-intervention
https://github.com/WICG/interventions/issues/18

Long story short. Everybody that uses mousePressed in the mobile chrome atm - the function is fired twice.

Reijo

jacorre:
Adding touchstarted helps alleviate the extra touch being generated. However, it doesn't seem to help with listeners attached to dom elements. For example, I have an icon that when clicked should do something. The click is triggered twice. Does this mean the p5 dom library also needs changes? Thank you!

lmccart:
this has been resolved with #1820, it will be in the next release. thanks!

mlarghydracept:
@jacorre what kind of event listener is attached to your dom element? 

I am asking because I am unable to recreate your problem.

jacorre:
I was just questioning if the dom library also needed changes. After making the update to the newest release, things are working as expected. Thank you!

jacorre:
Looks like it's still as issue in 0.5.11 specifically for mobile on Android device. I'm using Chrome. See my pen at https://codepen.io/anon/pen/dRaYWj. If you tap on the test button it will alert twice. The button uses the mousepressed function.

KevinWorkman:
I'm having the same problem on Android chrome with version 0.5.11.

Adding a `touchStarted()` function doesn't seem to do anything for me, but this hack seems to work okay in the meantime:

```javascript
var released = true;

function mouseReleased(){
	released = true;
	return false;
}

function mousePressed(){
	
	if(!released){
		return;
	}
	released = false;

	//rest of your code
}
```

luistoledo:
I suffer the same issue. Cant make it work well on an ios & browser cordova app, that also uses a 3rd party alert library.
Already tried every combination of mouse|touch released|pressed, returning false|true

[...]

Issue #1811, opened 2017-02-18
After update to Chrome 56, mousePressed is having issues


jacorre:
Everything was working fine for a while with a program I have until Chrome was recently updated to v56. On the desktop, it still seems to work fine, but if you're on mobile or use the emulator in the browser, there seems to be an issue with mousePressed. 

When I press anywhere on the canvas, I get the following in the console:

Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080

Uncaught TypeError: Cannot read property 'clientX' of undefined
    at d (p5.min.js:5)
    at e._updateNextMouseCoords (p5.min.js:5)
    at e._ontouchend (p5.min.js:6)

I do return false at the end of the mousePressed function. If I comment that out, I no longer receive the warning about preventDefault but I still get the Uncaught TypeError.

lmccart:
which version are you using? I think this should be fixed in the most recent version 0.5.7

jacorre:
I was using 0.5.6. I updated to 0.5.7 and the Uncaught TypeError is no longer appearing. However, the issue about the "Unable to preventDefault" is still appearing.

jacorre:
Did some further testing. Didn't notice any issues on iPad with Chrome or Safari. Seems to be an issue with Android on Chrome. Treats a tap as two.

jacorre:
Set up a simple example at http://codepen.io/jacorre/pen/zNVBvy.

Tapping canvas increases count by two when it should be one.

jacorre:
Has anyone else been able to reproduce this on Chrome for Android? Please view my [pen](http://codepen.io/jacorre/pen/zNVBvy) for an example. I have two applications that have been impacted by this and I'd really like to resolve this soon if possible. Thanks in advance for any help provided.

feedzh:
Hi,

It seems that to make life easier, browser also populates both events: mouseEvent and touchEvent. If you put return false; in the end of mousePressed function, it will start working for mobile. It is still counting 2 in the browser emulation (Since you are manipulating it with a mouse).

P5 tries to make sure both of these also. Unfortunately Chrome fires 2 events. So p5 will fire 2 events also. That is the reason, it is still broken in chrome emulator.

jacorre:
I forgot to include return false in my pen example. I updated the example. 

[...]

PR #1820, opened 2017-02-23
fixing chrome mobile 56 mousedown event fire issue

feedzh:
As describe on issue #1811 & #1815

Preventing google chrome mobile for firing duplicated mousePressed event when taping on element.

### DIFF
--- a/src/events/touch.js
+++ b/src/events/touch.js
@@ -234,6 +234,9 @@ p5.prototype._ontouchend = function(e) {
     if(executeDefault === false) {
       e.preventDefault();
     }
+  } else {
+    e.preventDefault();
+    return false;
   }
 };
 


lmccart:
looks good

src/events/touch.js
```diff
@@ -234,6 +234,9 @@ p5.prototype._ontouchend = function(e) {
     if(executeDefault === false) {
       e.preventDefault();
     }
+  } else {
```
DoctorBud: This change causes problems with iOS (Safari). Specifically, once a sketch is created, it and the entire HTML page it is contained in will be unresponsive to actions that rely on `ontouchend`, such as link-clicking and button pushes. Commenting out the above else-branch makes everything work again.

I recommend that the above code is made conditional for the environment where it is needed (Chrome Mobile 56).

Meanwhile, I'm monkeypatching around the above 'fix'.

tigoe:
Thanks, I'll give it a try and report back.

lmccart:
Ok, I updated the behavior of p5.js to only add duplicate firings if they are not already supported by the browser. Currently, chrome already fires off both mouse and touch events for a touch, firefox and safari do not. It is in the spec though, so we can expect that all browsers will eventually support this and the check for (window.PointerEvent) will be unnecessary. Here's more reading:
http://caniuse.com/#search=pointer%20events
https://www.w3.org/TR/pointerevents/
https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent

The p5.js behavior has shifted slightly due to this. If you define a mouse related event (ie mousePressed, mouseMoved, etc), it will be triggered by either a mouse or a touch. It will trigger only once.

However, if you define a touch related event (ie touchStarted, touchEnded, etc), it will be triggered by a touch, but *not* by a mouse click. In the past mouse and touch functions were completely interchangeable, now they go one direction. So if you are trying to define a function that will trigger on either mouse or touch, use the mouse named one. If you want to define a function that *only* triggers on touch, use the touch related ones.

This preserves the functionality of select/option and any other elements that depend on the 'click' event.

PR #1833, opened 2017-03-01
Rename 3D to webgl in test html files 

iashris:
A minor correction in the test-*.html files to reflect the renaming of 3D folder to webgl


lmccart:
thanks!

Issue #1392, opened 2016-05-07
Hope there will be a place for typescript definition file of p5js


LujunWeng:
I hope there will be a place for typescript definition files of p5js, so others can contribute to these definition files. 

I've made a Definition Request [here](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/9245).

@toolness has made a [tool](http://processing.toolness.org/general/2016/03/16/typescript-to-the-rescue.html) to generate definition files automatically. I think we can start from these generated definition files.

toolness:
Thanks for the comment!

I'd actually like to add my tool itself to the p5.js repository and have those definitions continue to be auto-generated if possible, as it keeps things [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) while also serving as a way to test our own documentation.

There will definitely be some functions, like [`save()`](http://p5js.org/reference/#/p5/save), whose method signatures will be too complex to describe accurately using yuidoc and which therefore will need to have custom definitions in the typescript definitions file, but I think we can work out a way to "merge" such custom definitions with the auto-generated ones.

LujunWeng:
Right now, I'm using the generated definition files. It works well for autocomplete in my IDE. Once there is any mismatch the IDE shows, I turn to the definition files and make modifications manually.

Like yesterday, I added `constructor(sketch: (p:any)=>void, node?: HTMLElement|boolean, sync?: boolean);` to the `p5` class so that `new p5(sketch)` would not cause an error any more.

PR #1849, opened 2017-03-08
fix some yuidoc annotations

infernalmaster:
Fixed big part of trivial yuidoc syntax errors, but not all.

Most of them I found when tried to generate typescript typings using this tool https://github.com/processing/p5.js/issues/1392


lmccart:
thanks!

Issue #1850, opened 2017-03-08
updatePixels() needed when using set() + get()


jbakse:
I'm not sure if this is a bug or working as designed, but it appears that `updatePixels()` is need after every call to `set()` if also using `get()`.

expected: all pixels turned red.
observed: only the last pixel is updated.
including the updatePixels() or removing the get() will fix

I *think* that this may be a bug, but I'm not sure the intended design. An `updatePixels()` is needed when done working with the pixels, but I *think* the one after the loop is meant to be enough. It is not entirely clear in the documentation if users are expected to call .updatePixels() after every .set() if also using .get() or just when done, before drawing the image.

Seen in p5.js v0.5.6

```javascript
function setup() {
	createCanvas(500, 500);
}

function draw() {
	background(0);

	img = createImage(10, 10);
	for (var y = 0; y < img.height; y++) {
		for (var x = 0; x < img.width; x++) {
			img.get(x, y); // would normally do something with this...
			var out_color = color(255, 0, 0, 255);
			img.set(x, y, out_color);

			// img.updatePixels();

		}
	}

	img.updatePixels();

	noSmooth();
	image(img, 0, 0, width, height);

	noLoop();
}
```

lmccart:
This is correct behavior, it's documented on the reference page for set:
http://p5js.org/reference/#/p5/set

jbakse:
Thanks so much for the very quick reply!

Just double checking this, because I think the documentation doesn't match up with the example i'm posting.

The docs say:
> After using set(), you must call updatePixels() for your changes to appear. This should be called once all pixels have been set.

But the example code requires updatePixels() after every single set(). Which is a performance drag.

Before I posted about this I looked through old issues to see if there was discussion and found this (issue #521), which led me to think that it wasn't intended to have to call so often:

> yes the thinking was... it's more difficult to figure out the correct pixel position with this new pixel scheme, so set() might be used more often. however due to canvas limitations, calling updatePixels() after every set call could really drag if you were setting multiple pixels rather than one. 

I just noticed that that quote was from you actually. :) 

It seems like setting pixels this way could work fine as long as you were not relying on get() being correct for set() pixels before updatePixels is called. In the example above the get() checks only pixels that haven't been set yet.

Maybe an extra note in the docs is in order? Something like:

After using set(), you must call updatePixels() for your changes to appear. This should be called once all pixels have been set, and must be called before calling .get() or drawing the image.

Thank you very much for your help. Sorry for dragging this ticket out.

[...]

PR #1854, opened 2017-03-11
Updated set() documentation

Trion129:
**Issue in consideration:**
#1850 

**Changes:**
Added the proposed documentation change in the pixels.js file's inline documentation


lmccart:
thanks!

PR #1856, opened 2017-03-14
added p5.Vector.average function

Durisvk:
I was coding p5.play Pixel Perfect Collision Detection and I was lacking this Vector.average function, so I was thinking it might be helpful for someone else.

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -1060,6 +1060,38 @@ p5.Vector.mag = function (vecT){
  */
 
 p5.Vector.average = function () {
+  var array = null;
+  if ( arguments.length === 0 ) {
+    return new p5.Vector( 0, 0, 0 );
+  } else if ( arguments.length === 1 && arguments[0] instanceof Array &&
+              arguments[0].length > 0) {
+    array = arguments[0];
+  }
+  else if ( arguments.length > 1 ) {
+    array = [].slice.call(arguments);
+  }
+  else {
+    return new p5.Vector(0, 0, 0);
+  }
+
+  if ( array ) {
+    return new p5.Vector(
+      array.reduce( function(sum, vector) {
+        return sum + vector.x;
+      }, 0 ) / array.length,
+      array.reduce( function(sum, vector) {
+        return sum + vector.y;
+      }) / array.length,
+      array.reduce( function(sum, vector) {
+        if ( vector.z ) {
+          return sum + vector.z;
+        }
+        return sum;
+      }) / array.length
+    );
+  }
+
+  return new p5.Vector(0, 0, 0);
 };
 
 
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -1050,4 +1050,17 @@ p5.Vector.mag = function (vecT){
   return Math.sqrt(magSq);
 };
 
+
+/**
+ * Takes the average of the vectors passed as array or as individual argument.
+ * Returns back the new vector.
+ * @static
+ * @param  {p5.Vector|p5.Vector[]} v1 a p5.Vectors to take average from
+ * @return {p5.Vector} the resulting p5.Vector
+ */
+
+p5.Vector.average = function () {
+};
+
+
 module.exports = p5.Vector;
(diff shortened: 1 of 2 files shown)


lmccart:
@shiffman do you want to take a look at this one? feel free to merge it in if you think it makes sense and passes review. or we can discuss if you don't find it necessary for p5.Vector, I think you have the best sense of where this object is currently at.

Durisvk:
Btw @shiffman I've watched all of your Coding Challenges, I really liked them...

shiffman:
Thanks for this @Durisvk, it looks great! One thing I think that is important to be mindful of is the size of the reference. It can be overwhelming to beginners when they look at, say, the `p5.Vector` page and see so many functions. That said, `p5.Vector` is separate from the main reference and it's also generally used by more experienced p5 users. 

Is it only `average()` that is missing or are there are lot of other vector math operations that we might want to include in which case does an add-on make sense?

Can you provide a couple code snippet samples of how you see it being used in examples?

Durisvk:
Yes you are right, it's not good to have a one big file with a lot of methods. I was mainly using `p5.Vector` as `Point`, and I had a problem with the array of points (btw I was making a `Pixel Perfect Collision detection` so I stored the `pixels` as points. I was missing the functions which would provide me the full control of this array of points. It was not only `average()`, I was missing those functions too: `getTopestPoint(arrayOfPoints)`, `getRightestPoint(arrayOfPoints)`, `getBottomestPoint(arrayOfPoints)`, `getLeftestPoint(arrayOfPoints)` => of course the naming is wrong, I am not good in english :) , but you surely know what I mean by those names.

Sample of usage of `average` would be getting the middle of convex shape from pixels/points. In p5.play I did this:

```javascript
this.findMiddle = function(xs, ys, filled) {
    var cx = this.average(xs);
    var cy = this.average(ys);


    var minDist = Infinity;
    var minX = -1;
    var minY = -1;
    for(var i = 0; i < filled.length; i++) {
      var d = dist(filled[i].x, filled[i].y, cx, cy);
      if(minDist > d) {
        minDist = d;
        minX = filled[i].x;
        minY = filled[i].y;
      }
    }
    return createVector(minX, minY);
  };

```

where `xs` is an array of x-coordinates in PNG image of non-transparent pixels,
and `ys` is an array of y-coordinates in this image.

average function looks like this:

```javascript
this.average = function(arr) {
    return arr.reduce(function(sum, item) {
      return sum + item;
    }, 0) / arr.length;
  };
```

**This will find the middle of a convex shape in PNG file with transparency.**

I also used function `this.averageVector` when I did find collision points and wanted to find the middle one.


-----------------------------------

If you want not to have the one big p5.Vector class, I can propose this:

I can create the `p5.VectorsList` -> or something like that (once more I am not good in naming the functions and objects)
and in that `p5.VectorsList` we can have the function for manipulation with arrays Vectors.
What do you say about my idea?
@shiffman

shiffman:
I think this is an excellent idea! I would suggest first making it as a p5 add-on library separate from p5 core (p5 geometry? p5 vertices?). This is way it can be more experimental and can be evaluated and tested by users more quickly. 

The next step would be to make a new repo for this idea, link from here and close the issue?

Durisvk:
Ok tomorrow I will create a repo and provide a link, then you can close the issue.

Durisvk:
here it is: https://github.com/Durisvk/p5.geometry

for now it's empty but tomorrow I would like to add some things to it

lmccart:
awesome! thanks @Durisvk

Durisvk:
Ok so here's the status from yesterday:
I added `VectorList` with which you can simply do multiple vectors operation. One of that operations I want to pin point is this:

You have a `p5.Vector` for position, velocity and acceleration, like @shiffman had in almost every video, what you can do is to store them into a `VectorList` like that:

```javascript
this.vectors = new VectorList( new p5.Vector(width/2, height/2), new p5.Vector(0, 1), new p5.Vector(0, 0);
```

And then in update function you can call this:

```javascript
this.vectors.chainAdd();
```

which will add velocity into position and acceleration into velocity. :)

shiffman:
@durisvk thanks for this really interesting work! I'll follow your other repo and look forward to trying it out sometime.

Issue #1864, opened 2017-03-20
Using select() and option() with a <select> HTMLelement


shiffman:
This is going to be confusing due to the word "select" meaning two different things, let me see if I can explain!

The following works:

```javascript
var dropdown = createSelect();
dropdown.option('apple');
dropdown.option('mango');
```

However, if I have an existing HTML element like:

```html
<select id="fruitlist></select>
```

The following does not work:

```javascript
var dropdown = select('#fruitlist');
dropdown.option('apple');
dropdown.option('mango');
```

It seems that the `p5.Element` created is not properly noted as a `<select>` element and therefore does not have the `option()` method attached?

I am happy to investigate this further, feedback or thoughts welcome.

JimishF:
I'm working on this issue. It can be resolved by adding methods of select prototype at the time of returning `wrapElemet()`

tigoe:
An added detail: 

This also doesn't work when you use p5.js in Cordova/PhoneGap. It used to, prior to 0.5.4, but in that version there was a problem  that caused touch events to be generated twice, e.g. every button press would generate two touchEnded() events. The  touchEnded fix  may have broken the select methods?

I've only tested on Android, but so far, no love.

Here's a test for this in PhoneGap:

1. Follow [Jjiashan's instructions](https://github.com/OhJia/p5Mobile/wiki/Phonegap) to make a p5 PhoneGap project. 
2. Add p5.dom.js to the list of script includes in the index.html as well. 
3. Put the following in the sketch.js:

````
var responseDiv, opts;
function setup() {
  console.log('starting');
  opts = createSelect();
  opts.position(10, 10);
  opts.size(150, 30)
  opts.option('Pick a device', '');
  opts.option('fee', '');
  opts.option('fie', '');
  opts.option('fo', '');
  opts.changed('fum');

  responseDiv = createDiv('tap the scan button to begin');
  responseDiv.position(10, 100);
}

function draw() {
}

function fum() {
  responseDiv.html(opts.value());
}
````

4. Run the sketch on the phone, then try to touch the select. You'll get no action.

JimishF:
`changed()` is from p5.Element.prototype, which is simply attaching change listener to particular element. I guess `opts.changed(fxn);` is accepting a function rather than a string. 
Try `opts.changed(function(e){ console.log( e.target ) })`

tigoe:
[...]

PR #1868, opened 2017-03-24
Fixed select() issue with existing dropdown

JimishF:
#1864 has been resolved via assigning select-box methods at the time of `wrapElement()`

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -528,10 +528,11 @@
 
   /**
    * Creates a dropdown menu &lt;select&gt;&lt;/select&gt; element in the DOM.
-   * It also helps to asssign select-box methods to p5.Element when selecting existing select box
+   * It also helps to assign select-box methods to p5.Element when selecting existing select box
    * @method createSelect
-   * @param {boolean} [multiple] true if dropdown should support multiple selections
-   * @return {Object|p5.Element} pointer to p5.Element holding created node
+   * @param {boolean} [multiple] true if dropdown should support multiple selections OR
+   * @param {Object  [P5.Element] if wrapping methods to existing select
+   * @return {Object|p5.Element} pointer to p5.Element holding created or existing node
    * @example
    * <div><code>
    * var sel;
@@ -555,17 +556,18 @@
    * </code></div>
    */
 
-  p5.prototype.createSelect = function(mult) {
-
-    if( typeof mult === 'object' && mult.elt.nodeName == 'SELECT' ){
-      var self = mult;
-      var elt = this.elt = mult.elt;
+  p5.prototype.createSelect = function() {
+    var elt,self;
+    var arg = arguments[0];
+    if( typeof arg === 'object' && arg.elt.nodeName == 'SELECT' ){
+      self = arg;
+      elt = this.elt = arg.elt;
     }else{
-      var elt = document.createElement('select');
-      if( mult && typeof mult === 'boolean' ){
+      elt = document.createElement('select');
+      if( arg && typeof arg === 'boolean' ){
         elt.setAttribute('multiple', 'true');
       }
-      var self = addElement(elt, this);
+      self = addElement(elt, this);
     }
     self.option = function(name, value){
       var opt = document.createElement('option');
@@ -586,7 +588,7 @@
         }
         return this;
       }else{
-        if (mult){
+        if (arg){
           for (var i = 0; i < this.elt.selectedOptions.length; i++){
             arr.push(this.elt.selectedOptions[i].value);
           }
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -188,7 +188,10 @@
       return converted;
     } else if (elt.tagName === "VIDEO" || elt.tagName === "AUDIO") {
       return new p5.MediaElement(elt);
-    } else {
+    } else if ( elt.tagName === "SELECT" ){
+      return createSelect( new p5.Element(elt) );
+    } 
+    else {
       return new p5.Element(elt);
     }
   }
@@ -525,6 +528,7 @@
 
   /**
    * Creates a dropdown menu &lt;select&gt;&lt;/select&gt; element in the DOM.
+   * It also helps to asssign select-box methods to p5.Element when selecting existing select box
    * @method createSelect
    * @param {boolean} [multiple] true if dropdown should support multiple selections
    * @return {Object|p5.Element} pointer to p5.Element holding created node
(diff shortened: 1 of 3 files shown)


JimishF:
Hello there, any changes to be done in this pull?  Please review..

lib/addons/p5.dom.js
```diff
@@ -550,12 +554,19 @@
...
+
+    if( typeof mult === 'object' && mult.elt.nodeName == 'SELECT' ){
+      var self = mult;
+      var elt = this.elt = mult.elt;
+    }else{
+      var elt = document.createElement('select');
+      if( mult && typeof mult === 'boolean' ){
+        elt.setAttribute('multiple', 'true');
+      }
+      var self = addElement(elt, this);
```
lmccart: This is a local variable, so the self.option may not attach correctly. Instead `var self;` should e declared before the if/else, then assigned in each part.

lib/addons/p5.dom.js
```diff
@@ -550,12 +554,19 @@
...
    */
+
   p5.prototype.createSelect = function(mult) {
-    var elt = document.createElement('select');
-    if (mult){
-      elt.setAttribute('multiple', 'true');
+
+    if( typeof mult === 'object' && mult.elt.nodeName == 'SELECT' ){
+      var self = mult;
+      var elt = this.elt = mult.elt;
```
lmccart: I'm not quite sure what the intention is with this line, can you explain what's happening here?

lib/addons/p5.dom.js
```diff
@@ -550,12 +554,19 @@
    * }
    * </code></div>
    */
+
```
lmccart: There should be an additional method signature identifying the option to pass in a DOM element rather than a boolean. https://github.com/processing/p5.js/wiki/Inline-documentation#additional-signatures

lib/addons/p5.dom.js
```diff
@@ -550,12 +554,19 @@
...
   p5.prototype.createSelect = function(mult) {
```
lmccart: This variable name should be changed to something that could refer to either the multiple boolean value or a DOM object. Or we could leave it empty and grab the object out of arguments[].

lib/addons/p5.dom.js
```diff
@@ -188,7 +188,10 @@
       return converted;
     } else if (elt.tagName === "VIDEO" || elt.tagName === "AUDIO") {
       return new p5.MediaElement(elt);
-    } else {
+    } else if ( elt.tagName === "SELECT" ){
+      return createSelect( new p5.Element(elt) );
+    }
```
lmccart: I don't think we should create the p5.Element here.. there's already a part in createSelect() that calls addElement() which does this. I would suggest just passing through the DOM element itself.

JimishF: If the tag name is select here, the addElement will never invoked insides createSelect. Because we're not going to add a new select.

lib/addons/p5.dom.js
JimishF: Okay

lib/addons/p5.dom.js
JimishF: If create select will be called via wrapElement, that means the argument mult should be a p5element. And if it is, than we just need the two methods of select box instead of creating a whole new select in Dom.

lib/addons/p5.dom.js
JimishF: Okay sure.

lib/addons/p5.dom.js
JimishF: Ah, I forgot that to mention but I'll append it.

lib/addons/p5.dom.js
```diff
@@ -528,10 +528,11 @@
 
   /**
    * Creates a dropdown menu &lt;select&gt;&lt;/select&gt; element in the DOM.
-   * It also helps to asssign select-box methods to p5.Element when selecting existing select box
+   * It also helps to assign select-box methods to p5.Element when selecting existing select box
    * @method createSelect
-   * @param {boolean} [multiple] true if dropdown should support multiple selections
-   * @return {Object|p5.Element} pointer to p5.Element holding created node
+   * @param {boolean} [multiple] true if dropdown should support multiple selections OR
+   * @param {Object  [P5.Element] if wrapping methods to existing select
```
limzykenneth: Typo here, probably should be `{Object|p5.Element} if wrapping methods to existing select`

JimishF: okay Updating.

JimishF:
Any status on merging ?

lib/addons/p5.dom.js
```diff
@@ -525,9 +528,11 @@
 
   /**
    * Creates a dropdown menu &lt;select&gt;&lt;/select&gt; element in the DOM.
+   * It also helps to assign select-box methods to p5.Element when selecting existing select box
    * @method createSelect
-   * @param {boolean} [multiple] true if dropdown should support multiple selections
-   * @return {Object|p5.Element} pointer to p5.Element holding created node
+   * @param {boolean} [multiple] true if dropdown should support multiple selections OR
+   * @param {Object|p5.Element} if wrapping methods to existing select
```
lmccart: This should go in a different block, see "additional signatures" section here: https://github.com/processing/p5.js/wiki/Inline-documentation#additional-signatures

lib/addons/p5.dom.js
```diff
@@ -550,12 +555,20 @@
...
    */
-  p5.prototype.createSelect = function(mult) {
-    var elt = document.createElement('select');
-    if (mult){
-      elt.setAttribute('multiple', 'true');
+
+  p5.prototype.createSelect = function() {
+    var elt,self;
+    var arg = arguments[0];
+    if( typeof arg === 'object' && arg.elt.nodeName == 'SELECT' ){
```
lmccart: This could create an error if an object other than a p5.Element is passed in. Instead this check could be:
if( arg.elt && arg.elt.nodeName == 'SELECT' ){

JimishF: Okay, i'll replace that.

lib/addons/p5.dom.js
JimishF: Oh okay got it.

lmccart:
awesome thanks for this! sorry it took so long to merge. I resolved all remaining review comments. this will be in the next release.

JimishF:
okay no problem. I was out of work actually. :)

shiffman:
Yay, glad to see this merged, thank you @JimishF and @lmccart!

R0dri:
Thanks for the update! I just read about this merge so i updated my libraries and my sketch begun to work marvelously! I was having trouble with the .changed() callback on a createSelect() var. So again, thank you very much for the work! 

I was getting an error like this:

> .changed() is not a function

PR #1874, opened 2017-04-03
Fixed error handling

### LINKED ISSUE #1871: undefined variable error + preload()/loadJSON()
If I use a variable that I've never defined like so:

```javascript
function setup() {
  var test1 = 5;
  var test3 = test1 + test2;
  console.log(test3);
}
```

I get the error I expect in the console:

screen shot 2017-04-01 at 2 08 19 pm

However, with that same code plus a call to `loadJSON()` in `preload()`, the error is not reported but rather is (??) bottled up somewhere in the `httpDo()` function:

screen shot 2017-04-01 at 2 09 54 pm

The corresponding code at line 23278 is in `src/io/files.js`

https://github.com/processing/p5.js/blob/0efd01151a17770b6c81baa7f338cb90da1c5dba/src/io/files.js#L640

With a complex sketch, this can make finding your typo quite difficult!

limzykenneth:
Fixed #1871. Also fixed errors with passing error callbacks, I forgot to take into account that `decrementPreload` injects an argument into the `arguments` list.

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -262,7 +262,7 @@ p5.prototype.loadJSON = function () {
       if (arg === 'jsonp' || arg === 'json') {
         t = arg;
       }
-    } else if (typeof arg === 'function') {
+    } else if (typeof arg === 'function' && arg !== decrementPreload) {
       if(!callback){
         callback = arg;
       }else{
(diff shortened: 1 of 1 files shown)


limzykenneth:
Add fix for bug mentioned in #653 as well

lmccart:
thanks! I logged the failed test issue here: https://github.com/processing/p5.js/issues/1881. will look into it further.

Issue #1888, opened 2017-04-11
handling table.getNum() with non-numerical cell values


lmccart:
```javascript
var data;

function preload() {
  data = loadTable("test.csv");
}

function setup() {
  console.log(data.getNum(2,1));
}
```

```csv
"joy",13,"blue eyes"
"gabriel",23,"black eyes"
"yu",,"black eyes"
```

this throws an error: Uncaught (in promise) Error: undefined is NaN (Not a Number), or sometimes fails silently with output "undefined undefined". 

I'm not sure what the best behavior is. if you call `int(undefined)` you get undefined, and if you call `int("")` you get NaN. maybe it should conform to this?
In any case, it should not throw an error that breaks the sketch, as it currently does.

lmccart:
@mlarghydracept?

PR #1889, opened 2017-04-12
handling table.getNum() with non-numerical cell values #1888

mlarghydracept:
I just fixed the error-throwing that breaks the sketch. 

I didn't handle undefined because I think that all the checks for calls outside the bounds of the table should catch any undefined cells. I could be mistaken though.

I know it goes against a bit of the design but it might useful to pass the current row to the getNum function so that if it does hit a blank cell it can throw more useful error messages (ie "The cell at row __ column __ is blank)


lmccart:
thanks! hm yeah, perhaps something for @almchung to think about if she starts work on friendly errors

almchung:
I will add this to my list! Thank you!

Issue #1826, opened 2017-02-27
hsb colors and map-ing


alexandrdmitrievich:
https://alpha.editor.p5js.org/full/SkmmVoZ9x
Why does HSB coloring working differently?

joecridge:
@alexandrdmitrievich could you post the code you’re using to draw the squares?

alexandrdmitrievich:
function setup() { 
  createCanvas(400, 400);
} 

function draw() { 
 // noStroke();  // Don't draw a stroke around shapes
c = color('hsb('+map(mouseX,0,width,0,255)+',75%,100%)');
fill(c); 
rect(10, 10, 35, 35); 
  
c = color('hsb('+round(map(mouseX,0,width,0,255))+',75%,100%)');
fill(c); 
rect(10, 55, 35, 35); 

colorMode(HSB, 100);
c = color(map(mouseX,0,width,0,255),125,255);
fill(c);
rect(55, 10, 35, 35);  
}

joecridge:
Okay, thanks.

Using an `'hsb(123, 45%, 67%)'` string to create colours only works [if the hue is an integer](https://github.com/processing/p5.js/blob/master/src/color/p5.Color.js#L378)\*, which is why the top left square is almost always white. Hue *always* goes from 0 to 360 (rather than 0 to 255) in these strings, which is why you don’t see the full spectrum in the bottom left square.

When you set `colorMode(HSB, 100)` before drawing the right square, you make it so that hue (*as a number*, going straight into `color()`) goes from 0 to 100, so you see even less of the spectrum. Additionally,  you are making it so that saturation and lightness also go from 0 to 100 (again, *as numeric arguments*), so the 125 and the 255 both get maxed out at 100 behind the scenes.

Note also that the colour mode persists once it has been set, so after the first draw loop the colour mode is always HSB (because you never change it to anything else), but this doesn’t affect the left two rectangles because *calling `color()` with a string always ignores `colorMode()`*.

---

Putting that all together, you would need something like this (to get the bottom left square to match the right one, and to make them both show the full spectrum):

```javascript
function setup() {
  createCanvas(400, 400);

  // We can set `colorMode()` up here because it’s the same every loop.
  // This is the same as calling `colorMode(HSB, 360, 100, 100)`.
  colorMode(HSB);
}

function draw() {
  // Hue in strings goes from 0 to 360.
  // This one will be white when string hue is decimal…
[...]

PR #1891, opened 2017-04-12
Constrain HSB values after parsing for pattern #1826

mlarghydracept:
i


lmccart:
great! 
FYI rather than putting it in the PR title, if you write "issue #1826" (or whichever issue you are referencing), it will automatically create a link between this PR and the issue. you can also write "closes #1826" and it will auto-close it when it's merged.

PR #1899, opened 2017-04-18
Adds handling for scientific notation in nf()

### LINKED ISSUE #1879: Unexpected nf()-formatting
Hi,

when using the nf()-function, the number gets formatted regardless of scientific notation. This makes very small or very big numbers indistinguishable.

Consider this example
```
function setup() { 
    createCanvas(400, 400);
} 

function draw() {
    background(220);

    smallNumber = 3.8944423423433e-14;
    text(smallNumber, 50, 50);
    text(nf(smallNumber, 1, 3), 50, 65);
	
    mediumNumber = 3.8944423423433;
    text(mediumNumber, 50, 100);
    text(nf(mediumNumber, 1, 3), 50, 115);
	
    bigNumber = 3.8944423423433e+50;
    text(bigNumber, 50, 150);
    text(nf(bigNumber, 1, 3), 50, 165);
}
```

Please let me know if the information provided is sufficient. And since this is my first contribution, I'm also open to general suggestions for improvement.

Best

mlarghydracept:
**Not ready to merge.**

I think that I should add the conversion of scientific notation to the other number formatting functions first. Currently I wrote in a helper function to handle the numbers. Javascript doesn't convert numbers with a magnitude higher than 21 or lower than -7. Currently these numbers are just silently ignored but the handleExponents() processes them. Not sure if writing in a helper function is the best idea. Open to feedback!

closes #1879

### DIFF
--- a/src/utilities/string_functions.js
+++ b/src/utilities/string_functions.js
@@ -197,11 +197,11 @@ p5.prototype.nf = function () {
 
 function doNf() {
   var num = arguments[0];
+  num = handleExponents(num);
   var neg = num < 0;
-  var n = neg ? num.toString().substring(1) : num.toString();
-  var decimalInd = n.indexOf('.');
-  var intPart = decimalInd !== -1 ? n.substring(0, decimalInd) : n;
-  var decPart = decimalInd !== -1 ? n.substring(decimalInd + 1) : '';
+  var decimalInd = num.indexOf('.');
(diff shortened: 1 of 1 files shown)


lmccart:
I guess the one other question is whether this causes much of a performance effect, and if so, if that's justified by a common enough use case. I looked up the Processing reference and it looks like they don't support scientific notation, so it is an option to forgo support in p5.js as well. What do you think?

mlarghydracept:
I agree 100% with this option!

Issue #1898, opened 2017-04-17
httpPost() not passing data


shiffman:
I believe this commit (c1d70b93632032c99bbcc367d7212cb953694cca) is causing an issue in 0.5.8 not present in 0.5.7. I am making a call to `httpPost()` like so:

```javascript
var data = {
  seed: 'hello'
}
httpPost('/upload', data, success, error);
```

I am console logging `arguments` in `httpDo()` in both 0.5.7 and 0.5.8. Note how the data is missing in 0.5.8.

0.5.7 (correct behavior):
<img width="342" alt="screen shot 2017-04-17 at 7 35 23 pm" src="https://cloud.githubusercontent.com/assets/191758/25108613/0b28f9c6-23a5-11e7-8bbd-7b7514fef8f4.png">

0.5.8 (incorrect behavior):
<img width="289" alt="screen shot 2017-04-17 at 7 36 14 pm" src="https://cloud.githubusercontent.com/assets/191758/25108625/29d89dea-23a5-11e7-981d-8bd51ab357a7.png">

I'm happy to investigate further and/or make a simplified example to test with.  cc @limzykenneth 

Full example is here: https://github.com/shiffman/NOC-S17-2-Intelligence-Learning/tree/master/week5-cnn-rnn-tensorflow/04b_rnn_flask_js

limzykenneth:
@shiffman Yes, I think you are right. I wonder if the latest fix has already fixed the data object missing but actually even then it probably won't work as mentioned in #1812, I forgot to stringify the object into proper JSON so POSTing data won't work yet.

I'm planning to get a PR together today or tomorrow and if there's any more bugs, do let me know. In the meantime, if at all possible, using `httpDo` with native options is one way to get around a lot of the problems.

limzykenneth:
@shiffman the data in your case is missing because the headers content type wasn't explicitly set to `application/json`, I should have known. @lmccart also turns out the data is stringified after all but still expect a PR for this issue.

lmccart:
fixed with https://github.com/processing/p5.js/pull/1900

PR #1900, opened 2017-04-18
Fix missing content type headers

limzykenneth:
This should take care of #1898. @shiffman it will be great if you can test this out!

A few things to note (maybe add to docs as well): using `httpDo` if the second argument is an object, it will be treated as using native request object and NOT a GET request with body content, I feel that is the correct behaviour although httpDo defaults to GET, GET request is not supposed to have a body content while POST and PUT would have body content will require the method argument as the second argument which works fine.

Sidenote: @lmccart Do you think it is worth having a HEAD build hosted somewhere like cdnjs or Amazon AWS just so that it will be easier for people to test out the HEAD library build and potentially spot bugs like these before they went live? It probably can be automated whenever there's a commit to the master branch.

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -792,12 +792,8 @@ p5.prototype.selectInput = function () {
  *                                    in as first argument
  */
 p5.prototype.httpGet = function () {
-  var args = new Array(arguments.length);
-  args[0] = arguments[0];
-  args[1] = 'GET';
-  for (var i = 1; i < arguments.length; ++i) {
-    args[i+1] = arguments[i];
-  }
+  var args = Array.prototype.slice.call(arguments);
+  args.splice(1, 0, 'GET');
   p5.prototype.httpDo.apply(this, args);
 };
 
@@ -818,12 +814,8 @@ p5.prototype.httpGet = function () {
  *                                    in as first argument
  */
 p5.prototype.httpPost = function () {
-  var args = new Array(arguments.length);
-  args[0] = arguments[0];
-  args[1] = 'POST';
-  for (var i = 1; i < arguments.length; ++i) {
-    args[i+1] = arguments[i];
-  }
+  var args = Array.prototype.slice.call(arguments);
+  args.splice(1, 0, 'POST');
   p5.prototype.httpDo.apply(this, args);
 };
 
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -6,6 +6,7 @@
  */
 
 /* globals Request: false */
+/* globals Headers: false */
 
 'use strict';
 
@@ -863,6 +864,7 @@ p5.prototype.httpDo = function () {
   var request;
   var jsonpOptions = {};
   var cbCount = 0;
+  var contentType = 'text/plain';
   // Trim the callbacks off the end to get an idea of how many arguments are passed
   for (var i = arguments.length-1; i > 0; i--){
     if(typeof arguments[i] === 'function'){
@@ -914,6 +916,7 @@ p5.prototype.httpDo = function () {
           }
         }else{
           data = JSON.stringify(a);
+          contentType = 'application/json';
         }
       } else if (typeof a === 'function') {
         if (!callback) {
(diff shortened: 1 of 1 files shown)


lmccart:
thanks for the fixes @limzykenneth! I'll test this out in the next few days, @shiffman please share your results too if you get a chance.

regarding the idea of a dev build, we'd previously worried having in progress copies of the dev build floating around could cause confusion... but I was thinking about it more, and I think if we just named it differently this would be less likely to cause problems and would be very helpful. I'll look into this next chance I get.

limzykenneth:
To avoid confusion for a dev build I think it is important that the file or package name does not contain any version number and clearly marked as a dev build. I quite like how OF does a nightly build that is kept on github but linked to from their website, with p5.js though we can go a bit further and have direct links to the js files instead of zip packages.

shiffman:
This didn't work for me. I think there is something amiss in the call to `httpDo()` in `httpPost()` here:

```javascript
p5.prototype.httpPost = function () {
  args[0] = arguments[0];
  args[1] = 'POST';
  for (var i = 2; i < args.length; ++i) {
    args[i] = arguments[i];
  }
  p5.prototype.httpDo.apply(this, args);
};
```

I believe this should be:

```javascript
p5.prototype.httpPost = function () {
  var args = new Array(arguments.length+1); // +1
  args[0] = arguments[0];
  args[1] = 'POST';
  for (var i = 2; i < args.length; ++i) {
    args[i] = arguments[i-1]; // -1
  }
  p5.prototype.httpDo.apply(this, args);
};
```

Otherwise the data is overwritten with `args[1] = 'POST'`. With this change everything worked out, but for whatever reason I'm now getting a "bad path" error. I can continue to investigate, but am a bit stumped at the moment.

kadamwhite:
After testing out the function, I agree with @shiffman. However, I'd potentially recommend a more concise function implementation using `Array#splice`:
```js
p5.prototype.httpPost = function () {
  var args = Array.prototype.slice.call(arguments);
  args.splice(1, 0, 'POST');
  p5.prototype.httpDo.apply(this, args);
};
```
this has the same result and avoids the need to manually iterate to build the array
image

limzykenneth:
@kadamwhite I quite like that solution, I'm going to make those changes, please do test them when I pushed it @shiffman Thanks.

limzykenneth:
@shiffman I've changed the implementation of `httpPost` and `httpGet` over to @kadamwhite's solution, and for my tests, they worked fine. I'll provide some sample test code below.

```javascript
function setup(){
  httpPost("http://localhost:8080/", "some data, can be JSON as well", function(data){
    console.log(data);
  }, function(err){
    console.log(err);
  });
}
```

On `localhost:8080` I have a minimal express server running with the following route and setting

```javascript
app.use(bodyParser.json());
app.use(bodyParser.text());

app.post("/", function(req, res){
  console.log(req.body);
  res.send(req.body);
});

app.get("/", function(req, res){
  res.send("Success!");
});
```

@shiffman The link you provided in your original issue (#1898) doesn't exist anymore, can I get a copy of both your p5 code and server code if you are still experiencing issue? May be some problems I didn't anticipate with my setup.

PS. @lmccart I'm thinking of writing more tests using something like [sinon.js](http://sinonjs.org) which will at least assure the basic signature works. Won't be able to get around to it in awhile though

shiffman:
@limzykenneth sorry, I refactored that repo. Here is the example I am working with. I can try to test later, but feel free to go ahead!

https://github.com/shiffman/NOC-S17-2-Intelligence-Learning/tree/master/week6-rnn-tensorflow/01b_rnn_flask_js

limzykenneth:
@shiffman I just realized that you probably were looking at the wrong version of the library, the argument splicing problem should be mixed with https://github.com/processing/p5.js/commit/c1d70b93632032c99bbcc367d7212cb953694cca 

If you can, try a build with my fork, see if everything works as expected. Your sketch looks simple enough where it is calling `httpPost`, I didn't have python setup on my machine so I didn't test them but go ahead and try it, see if it works.

lmccart:
this is great, thanks @limzykenneth. it seems to work for me @shiffman let us know if your example is still failing if you build with the latest.

shiffman:
Thank you for fixing this, it all checks out!

PR #1903, opened 2017-04-19
Opentype.js update

### LINKED ISSUE #1902: Need to update included opentype.js to current version
https://github.com/nodebox/opentype.js/releases

Short-term, this will give us better font-support (specifically fonts using format 12 cmap tables) and thus eliminate at least some of the problems people are having with certain fonts not loading or rendering correctly.

Longer-term, as opentype.js continues to develop (support for Chinese fonts is still weak, for example), we should integrate this into our dev-ops workflow so it happens on a somewhat regular basis.

mlarghydracept:
All tests work as expected

closes #1902

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "brfs": "^1.4.0",
     "es6-promise": "^4.0.5",
     "fetch-jsonp": "^1.0.6",
-    "opentype.js": "^0.4.9",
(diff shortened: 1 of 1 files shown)


lmccart:
this looks good, @dhowe is there anything more to review/check here before we merge?

dhowe:
if automated and manual tests are fine, should be good to go...

PR #1907, opened 2017-04-20
isPointInCircle method added

jelic98:
Goal: Adding a bunch of methods to simplify usage of analytical geometry

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -118,6 +118,34 @@ p5.prototype.constrain = function(n, low, high) {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


lmccart:
this is cool. generally, it's a good idea to open an issue before submitting a PR so we can be sure we're able to add the feature you're working on. in this case, I think the function is super useful but a bit beyond the scope of this library. it could make sense as part of an addon library though. there's a document here that explains how to make one, if you're interested. https://github.com/processing/p5.js/wiki/Libraries

jelic98:
Sure. I'll check that out.

Issue #1745, opened 2017-01-05
Canvas width should be 320 in capture video example


DonKarlssonSan:
The [capture video example](https://github.com/processing/p5.js/blob/4dd42f2136400d33b4eca521b218b07636e16011/test/manual-test-examples/addons/p5.dom/capture_video/sketch.js) creates a canvas with width 390 instead of 320:

````
var capture;

function setup() {
  createCanvas(390, 240);
  capture = createCapture(VIDEO);
  capture.size(320, 240);
  //capture.hide();
}

function draw() {
  background(255);
  image(capture, 0, 0, 320, 240);
  filter('INVERT');
}
````

Here is the link to the live documentation: https://p5js.org/examples/dom-video-capture.html

Issue #1673, opened 2016-11-23
capture object drawn into graphics object is scaled incorrectly on retina


kylemcdonald:
Hi, I'm trying to work with video from a webcam at two different resolutions. As far as I know, it's not possible to open the webcam at two resolutions, so I tried creating a graphics object and drawing into that:

```javascript

var captureBig;
var captureSmall;
var smallWidth = 160, smallHeight = 120;

function setup() {
  createCanvas(640, 480);
  captureBig = createCapture(VIDEO);
  captureBig.size(width, height);
  captureBig.hide();
  captureSmall = createGraphics(smallWidth, smallHeight);
}

function draw() {
  image(captureBig, 0, 0);
  captureSmall.image(captureBig, 0, 0, smallWidth, smallHeight); // renders incorrectly on retina
  // captureSmall.image(captureBig, 0, 0, smallWidth/2, smallHeight/2); // renders correctly on retina
  image(captureSmall, 0, 0);
}
```

When this is working correctly, you should see a small "picture in picture" at the top left. When it's working incorrectly, that PIP crops out the top left quarter of the image.

Correct:

<img width="641" alt="screen shot 2016-11-23 at 2 55 51 pm" src="https://cloud.githubusercontent.com/assets/157106/20581315/f8a62c36-b18c-11e6-9b3f-b03a16b267c9.png">

Incorrect:

<img width="640" alt="screen shot 2016-11-23 at 2 55 35 pm" src="https://cloud.githubusercontent.com/assets/157106/20581324/ffe6827a-b18c-11e6-9dbc-72160f7395dd.png">

I tested disabling retina by using [RDM](https://github.com/avibrazil/RDM) and selecting a non-retina resolution, and the scaling worked correctly.

krongeroso:
I have the same problem with a similar project. I use the "createGrapichs()" function to draw by layers (like photoshop). The strange thing is that it works well on my computer, but I have the same problem as you when I run it from my smartphone. For some strange reason the image is scaled, so use a cheap solution... scale the image for 1/3 and work ... but only on some phones. In others the scaling factor is another, it is very rare. I do not know why it is scaled or because it is different in some cases. :(

krongeroso:
@kylemcdonald 
I solve the problem!!
Only you need to use the function pixelDensity(1) like this:

pg = createGraphics(200, 200);
pg.pixelDensity(1);

Check this example: 
https://github.com/processing/p5.js/issues/1673

Issue #1079, opened 2015-11-06
p5.MediaElement size()


shiffman:
Filing an issue @lmccart and I just talked about related to #1022.   We adjusted the associated canvas size when calling `loadPixels()` so that the pixel array would match the DOM element size.  I'm noticing users confused even when just calling `image()` to draw the image to the canvas that the size doesn't match what is set with `size()`.

shiffman:
I'm not sure how this happened but I'm looking at p5.dom.js 0.2.7 and the code change from https://github.com/processing/p5.js/commit/317fdd17bd507acb35b7e0666dd6b301b9fef0dd no longer seems to be in there.  This is causing the pixels not to be resized appropriately when calling `size()` of a `p5.MediaElement`.

I believe here [line 1679](https://github.com/processing/p5.js/blob/master/lib/addons/p5.dom.js#L1679):

```javascript
if (this.canvas.width !== this.elt.width) {
  this.canvas.width = this.elt.videoWidth;
  this.canvas.height = this.elt.videoHeight;
  this.width = this.canvas.width;
  this.height = this.canvas.height;
}
```

should be:

```javascript
if (this.canvas.width !== this.elt.width) {
  this.canvas.width = this.elt.width;
  this.canvas.height = this.elt.height;
  this.width = this.canvas.width;
  this.height = this.canvas.height;
}
```

Or maybe something should be happening when `size()` is called that makes this work with the existing code? I'm not sure but it is broken again.  Happy to pull request the above, but perhaps some other things changed and I'm missing something.

lmccart:
thanks for catching this! i think there was something tricky with this one... let me look into it tomorrow. what would be super helpful is a unit test for this. i think it could just check the size of elements or a test a few pixels in the pixels array.

lmccart:
very weird, i can't tell how this got lost but you're right it should def be the second version. i just made the change. thanks!

shiffman:
This particular fix was reverted due to issue #1274 and the pixels array length does not match the specified `size()` anymore.  You can see how the commit 07d154c641baf1757e464d10218fbd232bdabb1b changes `this.elt.width` back to `this.elt.videoWidth`.  We seem to be in an endless cycle of changing this code back and forth.  How do we best resolve both of these issues?  

I am making a video about this question and when I upload that video I will post a link here.

Just to be clear again, the code that resolves this issue is:

```javascript
if (this.canvas.width !== this.elt.width) {
  this.canvas.width = this.elt.width;
[...]

PR #1910, opened 2017-04-25
fixing problems with video element scaling

### LINKED ISSUE #1848: Video on canvas has incorrect width.
```javascript
var vid;

function setup() {
  createCanvas(320, 240);
  background(51);
  vid = createCapture(VIDEO);
  vid.size(320, 240);
  vid.hide();
}

function draw() {
  image(vid, 0, 0, width, height);
}
```

This is producing a video on canvas with the width of video source divided in half. 
I have the latest version of p5js: 0.5.7
Anyone else see this issue?

mlarghydracept:
Closes #1848 

The implementation is kind of unseemly but this handles every combination of scaling with image(), size(), and changes to pixel density. Hopefully we can make image() more readable in the future but for now this works as expected in all the cases I could test.

Also tested on example code in #1745, #1673, and #1079

### DIFF
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -198,10 +198,8 @@ p5.prototype.image =
   var defH = img.height;
 
   if (img.elt && img.elt.videoWidth && !img.canvas) { // video no canvas
-    var actualW = img.elt.videoWidth;
-    var actualH = img.elt.videoHeight;
     defW = img.elt.videoWidth;
-    defH = img.elt.width*actualH/actualW;
+    defH = img.elt.videoHeight;
   }
 
   var _dx = dx;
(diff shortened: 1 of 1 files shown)


src/image/loading_displaying.js
```diff
@@ -223,7 +221,14 @@ p5.prototype.image =
   var pd = 1;
 
   if (img.elt && img.elt.videoWidth && img.elt.style.width && !img.canvas) {
-    pd = img.elt.videoWidth / parseInt(img.elt.style.width, 10);
+    //if img.elt.size() has been used and no width passed to image()
+    if(img.elt.style.width && !dWidth){
+      pd = img.elt.videoWidth / parseInt(img.elt.style.width, 10);
```
lmccart: to clean this a little, we can pull out the `parseInt(img.elt.style.width, 10);` part and do that division at the end. that is, reformat lines 223-235 to first check img.elt, then do the other checks inside that if structure, then doing the final division at the end.

```javascript
if (img.elt) {
  if (img.elt.videoWidth && ...
  } ...
  pd /= parseInt(img.elt.style.width, 10);
}

lmccart:
i commend your ability to forge past the cringeworthiness of this general code area and make a patch!

PR #1911, opened 2017-04-26
Cleaning up code from PR #1910

mlarghydracept:
Cleaning up prior merged code from PR #1910 with more clarity as per @lmccart 's [suggestion](https://github.com/processing/p5.js/pull/1910#pullrequestreview-34736627)

PS Let me know If there is a better way to add to a PR that has already been merged!


lmccart:
looks great. this is the correct way to add to an already merged PR. if it's an unmerged PR, you can just push changes to the same branch and they'll automatically add on.

PR #1913, opened 2017-04-26
Immediate mode shapes with texturing

### LINKED ISSUE #1561: WEBGL: texture not being set when using shapes
p5js version: 0.5.3

Example:  https://jsfiddle.net/7nxcL1tf/19/

But if box() is called prior to endShape() then the texture is applied to the shape correctly.

mlarghydracept:
closes #1561

### DIFF
--- a/src/webgl/p5.RendererGL.Immediate.js
+++ b/src/webgl/p5.RendererGL.Immediate.js
@@ -76,7 +76,7 @@ function(mode, isCurve, isBezier,isQuadratic, isContour, shapeKind){
     this.immediateMode.vertexPositions,
     this.immediateMode.vertexColors);
   if(mode){
-    if(this.drawMode === 'fill' || 'texture'){
+    if(this.drawMode === 'fill' || this.drawMode ==='texture'){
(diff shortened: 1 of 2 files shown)


src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -76,7 +76,7 @@ function(mode, isCurve, isBezier,isQuadratic, isContour, shapeKind){
     this.immediateMode.vertexPositions,
     this.immediateMode.vertexColors);
   if(mode){
-    if(this.drawMode === 'fill'){
+    if(this.drawMode === 'fill' || 'texture'){
```
lmccart: does this work? I think this would get evaluated as 
`if ( (this.drawMode === 'fill') || 'texture')` which would always result in "true", since the second part of the OR is a non-falsey value. I could be wrong..
rewriting as:
`if ( this.drawMode === 'fill' || this.drawMode === 'texture')` removes any ambiguity.

mlarghydracept:
Thanks for catching that @lmccart! Switching between languages confused me. I also caught another problem. Everything should be good now.

PR #1917, opened 2017-04-28
Minor API change: textFont() should return current font

dhowe:
For consistency with other text functions (textSize(), textAlign(), textLeading()), textFont() called without arguments should return the current font (object or string) rather than 'this'. There are also use-cases where this call is needed as the current font cannot otherwise be accessed.


lmccart:
looks like there's one small lint error, could you fix this before I merge? thanks!

PR #1926, opened 2017-05-06
Fixed select() for select elements

stevengreens10:
I'm not sure if this solution is optimal as this is my first time ever contributing to p5.js, but I tested it locally, and it seemed to work.


stevengreens10:
Resolves issue #1864

lmccart:
thanks for this! we actually had a [similar PR](https://github.com/processing/p5.js/pull/1868) open for addressing this same thing, which was just merged.

Issue #1937, opened 2017-05-10
Nice to have for httpDo


limzykenneth:
I just noticed some missing feature in `httpDo()` `httpPost()`, specifically the `data` option currently do not accept arrays or numbers which I think it should. 

I can only get to this earliest next week so if anyone else wants to tackle this before I get to it, do feel free.

lmccart:
@mlarghydracept want to check this one out?

JMandel1027:
hi! I had some free time this evening and made a fix on `httpDo()` `httpPost` I'm going to run some more tests and then make a pull request~!! (my first ever n_n)

mlarghydracept:
@jay-manday still want to work this one out? I'll unassign myself if so.

jay-manday:
@mlarghydracept yup! @limzykenneth and I have been working on this 🌮

lmccart:
this should be all set from https://github.com/processing/p5.js/commit/41dfb9a395c55d4170d1cc082ee098bda8d77843. httpGet is covered by the change to httpDo

jay-manday:
yup! thanks again @lmccart & @limzykenneth was great working with you both on this!

PR #1940, opened 2017-05-11
added support for int's and array's to 'httpDo'

JMandel1027:
First  pull request!! made `integer` and `array` type handling for `httpDo`as requested in #1937   n_n

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -177,6 +177,8 @@ p5.prototype.loadFont = function (path, onSuccess, onError) {
  *                                    there is an error, response is passed
  *                                    in as first argument
  * @return {Object|Array}             JSON data
+ * @return {array}                    Array
+ * @return {integer}                   Integer
  * @example
  *
  * <p>Calling loadJSON() inside preload() guarantees to complete the
@@ -249,14 +251,17 @@ p5.prototype.loadJSON = function () {
         t = arg;
       }
     } else if (typeof arg === 'function' && arg !== decrementPreload) {
-      if(!callback){
+      if(!callback) {
         callback = arg;
-      }else{
+        return callback;
+      } else {
         errorCallback = arg;
+        return errorCallback;
       }
-    } else if (typeof arg === 'object' && arg.hasOwnProperty('jsonpCallback')){
+    } else if (typeof arg === 'object' && arg.hasOwnProperty('jsonpCallback')) {
       t = 'jsonp';
       options = arg;
+      return options;
     }
   }
 
@@ -839,7 +844,7 @@ p5.prototype.httpDo = function () {
   for (var i = arguments.length-1; i > 0; i--) {
     if(typeof arguments[i] === 'function') {
       cbCount++;
-    } else{
+    } else {
       break;
     }
   }
@@ -908,7 +913,7 @@ p5.prototype.httpDo = function () {
         type = 'json';
       } else if (path.indexOf('xml') !== -1) {
         type = 'xml';
-      }  else if (path.indexOf('array') !== -1) {
+      } else if (path.indexOf('array') !== -1) {
         type = 'array';
       } else if (path.indexOf('integer') !== -1) {
         type = 'integer';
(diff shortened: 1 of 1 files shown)


src/io/files.js
```diff
@@ -869,7 +877,17 @@ p5.prototype.httpDo = function () {
       if (typeof a === 'string') {
         if (a === 'GET' || a === 'POST' || a === 'PUT' || a === 'DELETE') {
           method = a;
-        } else if(a === 'json' || a === 'jsonp' || a === 'xml' || a === 'text') {
```
stevengreens10: Why did you take away the || (or)'s if all of the else if statements do the same thing?

JMandel1027: I removed the or || because they were breaking the code style guides, otherwise would have left as is

JMandel1027: I removed the or || because they were breaking the code style guides bc the line got to long, otherwise would have left as is

src/io/files.js
```diff
@@ -177,6 +177,8 @@ p5.prototype.loadFont = function (path, onSuccess, onError) {
  *                                    there is an error, response is passed
  *                                    in as first argument
  * @return {Object|Array}             JSON data
+ * @return {array}                    Array
+ * @return {integer}                   Integer
  * @example
  *
```
JMandel1027: I added error handling so if `resp` received a `type: array` or `type: integer` it would be parsed as a string... I realized this was a little naive since array's have a bit more to them than that. Would adding a `map('text/array')` functionality to `loadJSON` to store the `array` this seems like a good solution. @lmccart @limzykenneth curious to hear your feedback on this :)

limzykenneth:
@JMandel1027 I'm afraid you are a bit too heavy handed with the changes here. A lot of the stuff don't really need to be changed for this functionality and there are several errors that were made as part of those changes. I don't have a good internet connection right now to do a complete code review but I'll try to get back to you about this tomorrow.

JMandel1027:
@limzykenneth for sure for sure, my initial attempt was targeted at `else if` statements inside `httpDo()` starting at line 835. but i realize now that the `data` option where that `for()` loop checks the `data` arguments might be the actual solution instead of the checks for `text/array` or `text/integer`. I realized I went to far and left as is 😵 i'll sit tight on further changes

limzykenneth:
@JMandel1027  Ok, so for this change you only need to look at [this set of code](https://github.com/processing/p5.js/blob/master/src/io/files.js#L869-L892). I have totally forgotten that array have `typeof` `object` so there is no need to detect whether the user passed in an array or not, it should be covered already. That leaves only Number, do note that in Javascript there is no differentiation between integer and floats, all numbers are 64-bit floating point number and the type is called `Number`. I'll leave some line notes in your commits to clarify what needs to be changed.

src/io/files.js
```diff
@@ -777,7 +780,9 @@ p5.prototype.httpGet = function () {
...
  * @method httpPost
  * @param  {String}        path       name of the file or url to load
- * @param  {String}        [datatype] "json", "jsonp", "xml", or "text"
+ * @param  {String}        [datatype] "json", "jsonp", "xml",
+ * "text", "array", or "integer"
```
limzykenneth: This should not be changed, the `datatype` option here refers to the return datatype from the server

src/io/files.js
```diff
@@ -803,7 +808,9 @@ p5.prototype.httpPost = function () {
  * @param  {String}        path       name of the file or url to load
  * @param  {String}        [method]   either "GET", "POST", or "PUT",
  *                                    defaults to "GET"
- * @param  {String}        [datatype] "json", "jsonp", "xml", or "text"
+ * @param  {String}        [datatype] "json", "jsonp", "xml",  "text",
+ *"array", or "integer"
```
limzykenneth: Same as above, should not be changed

src/io/files.js
```diff
@@ -854,11 +861,15 @@ p5.prototype.httpDo = function () {
         type = 'json';
       } else if (request.url.indexOf('xml') !== -1) {
         type = 'xml';
-      } else {
+      } else if (request.url.indexOf('text') !== -1) {
         type = 'text';
+      } else if (request.url.indexOf('array') !== -1) {
+        type = 'array';
+      } else if (request.url.indexOf('integer') !== -1) {
+        type = 'integer';
```
limzykenneth: This block should not be changed as well for one this is to detect return type from the server and not the datatype the request is sending, for another, detecting if the URL string contains `text`, `array` or `integer` doesn't make sense...

src/io/files.js
limzykenneth: This change should be reverted as well as this also concerns return data from the server. If line length is too long, break the line at `||` (double pipe), breaking it into multiple `if`s may have performance penalties.

src/io/files.js
```diff
@@ -897,8 +918,12 @@ p5.prototype.httpDo = function () {
         type = 'json';
       } else if (path.indexOf('xml') !== -1) {
         type = 'xml';
-      } else {
+      } else if (path.indexOf('text') !== -1) {
```
limzykenneth: Same as the `request.url` part, not needed and doesn't really make sense...

limzykenneth:
My advice would be to close this PR and start over, most of the changes are not needed and to implement the new functionality is really basic, don't overthink it. Sorry about that, I should have given some hints over in the issue before you started working on this.

src/io/files.js
JMandel1027: Thanks for this feedback! I was definitely way way to heavy handed with the scope of my edits, I've been looking at the set you identified in an earlier message and reverted to the current official p5 build of `files.js` as I start another attempt. I hope I'm not wasting to much of your time with this! Thank you again for your patience and feedback

JMandel1027:
Of course! Thank you for this feedback and for taking your time to look over my work. I also appreciate the opportunity to finish adding this simple functionality. I was definitely way way to heavy handed with the scope of my edits, I've been looking at the set you identified in an earlier message and reverted to the current official p5 build of files.js as I start another attempt.

lmccart:
@jay-manday thanks for your patience and effort in revising this. sorry we weren't able to merge this one, but hope we can merge your second attempt. @limzykenneth thanks for your really significant feedback and guidance, it is really great to see this is a place that people can learn and develop their code

jay-manday:
@lmccart of course! I really appreciate you guys giving me the opportunity to contribute in a really constructive environment, thanks again @limzykenneth for all the guidance! I learnt a ton working with him on this, @limzykenneth and I have made these needed changes opened a new PR 😄

PR #1947, opened 2017-05-18
angleBetween3D()

### LINKED ISSUE #1893: Suggestion for p5.Vector.angleBetween() change
p5.Vector.angleBetween() is useful for 2D, but not very useful for 3D transformations which require 3 angles. So I suggest creating a 3D version like this one:

_function angleBetween(p1,p2) {
	var rotX = atan2(p2.z - p1.z, p2.y - p1.y);
	var rotY = atan2(p2.x - p1.x, p2.z - p1.z);
	var rotZ = atan2(p2.y - p1.y, p2.x - p1.x);
	var v = createVector(rotX,rotY,rotZ);
	// if angle in degrees
	// v.mult(180/PI);
	return v;
}
rotateX(v.x);
rotateY(v.y);
rotateZ(v.z);_

Regards

mlarghydracept:
closes #1893


mlarghydracept:
please don't merge yet realized that I need to do some editing to the example

PR #1961, opened 2017-05-24
cached get image reference

### LINKED ISSUE #1935: calls to get() create new p5.Images, hurting performance
https://github.com/processing/p5.js/blob/e51fa1cbafe654c9c8c5d54aa67d12892c484be7/src/core/p5.Renderer2D.js#L244

can we just create one and update it?

mlarghydracept:
I am still not 100% sure that this is what is being looked for.

I tested this with
```
var b;
var c;

function setup() {
	createCanvas(600, 200);
	background(0);
	fill(0,0,255);
	rect(0,0,200,200);
	b = get();
	c = get();
	c.loadPixels();
	for (i = 0; i < 100; i++) {
	  for (j = 0; j < 100; j++) {
	    c.set(i, j, color(255));
	  }
	}
   	c.updatePixels();
}

function draw() {
    image(b, 200, 0);
    image(c, 400, 0);
}
```
and the original p5.image stored in `b` was not changed.

Let me know if this is way off base.

closes #1935

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -242,11 +242,10 @@ p5.Renderer2D.prototype.get = function(x, y, w, h) {
     var sw = dw * pd;
     var sh = dh * pd;
 
-    var region;
     if(this.cachedGetImage === null) {
       this.cachedGetImage = new p5.Image(dw, dh);
     }
-    region = this.cachedGetImage;
+    var region = this.cachedGetImage;
     region.canvas.getContext('2d').drawImage(this.canvas, sx, sy, sw, sh,
       0, 0, dw, dh);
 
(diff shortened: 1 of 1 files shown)


src/core/p5.Renderer2D.js
```diff
@@ -241,7 +242,12 @@ p5.Renderer2D.prototype.get = function(x, y, w, h) {
     var sw = dw * pd;
     var sh = dh * pd;
 
-    var region = new p5.Image(dw, dh);
+    var region;
+    if(this.cachedGetImage === null) {
```
lmccart: does this.cachedGetImage ever get set?

mlarghydracept: ah yeah oops! I just fixed it so that it does but now `b` is modified along with `c` in the above example

lmccart:
thanks for looking into this!

Issue #1878, opened 2017-04-05
WebGL texture transparent PNG


andyinabox:
I'm attempting to use a transparent PNG as a texture, but the alpha channel doesn't register. Is this expected? Here's some sample code:

```javascript
var sketch = function($p5){

  var img;

  $p5.preload = function() {
    img = $p5.loadImage("data/shape3.png");
  }

  $p5.setup = function() {
    $p5.createCanvas(500, 500, $p5.WEBGL);
  }

  $p5.draw = function() {
    $p5.background(200);
    $p5.texture(img);
    $p5.plane(300, 300);
  }
}
new p5(sketch);
```

The image I'm using is:
![shape3](https://cloud.githubusercontent.com/assets/151084/24698429/99371928-19be-11e7-8496-a5116906b5cc.png)

And it renders like this:
![screenshot 2017-04-05 05 10 19](https://cloud.githubusercontent.com/assets/151084/24698460/bd18c63e-19be-11e7-9526-a3f3dd1141d9.png)

As you can see, anything that has an alpha value seems to show up as white.

mlarghydracept:
As far as I can tell this expected behavior. The default material isn't going to detect the alpha of your texture image. You can try using [specularMaterial()](https://p5js.org/reference/#/p5/specularMaterial), but from what I understand this won't necessarily achieve your desired effect either.

lmccart:
Yes, this is currently the intended behavior.

sixhat:
I've been dealing with this for a while and found that you need to enable the Blend mode of your webgl canvas if you need transparent textures to work. 

You can either do 

```javascript
gl = this._renderer.GL;
gl.enable(gl.BLEND);
gl.blendFunc(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA);
```
or take advantage that `fill` has a side effect of setting this for you if you issue a `fill` with a color that is not opaque. 

in your code, if you just put `$p5.fill(0,0,0,0);` before `$p5.texture(img);` you should get what you look for. 

The example below produces 

![output-alpha](https://cloud.githubusercontent.com/assets/614881/26454158/1c048f18-415e-11e7-9082-e6405d3a616c.png)

```javascript
function preload(){
	alpha = loadImage('alpha.png');
}

function setup() {
	createCanvas(500,500, WEBGL);
}

function draw (){
	background(100);

	translate(-50,-50,1);
	texture(alpha);
	plane(300,300);

[...]

Issue #1638, opened 2016-10-17
line() function not working in WEBGL


jacobrake:
Running into error in p5.js when using line() with WEBGL starting on line 7790
code excerpt:

p5.prototype.line = function() {
  if (!this._renderer._doStroke) {
    return this;
  }
  var args = new Array(arguments.length);
  for (var i = 0; i < args.length; ++i) {
    args[i] = arguments[i];
  }
  //check whether we should draw a 3d line or 2d
  if(this._renderer.isP3D){
    **this._renderer.line(** //line7790 is the error
      args[0],
      args[1],
      args[2],
      args[3],
      args[4],
      args[5]);
  } else {
    this._renderer.line(
      args[0],
      args[1],
      args[2],
      args[3]);
  }
  return this;
};

I used basic line code used by [https://github.com/processing/p5.js/wiki/Getting-started-with-WebGL-in-p5](url) to see if my code was bad but running into same error with this code.

lmccart:
Could you please post a a snippet of the code that isn't working?

jacobrake:
this is the basic code I get an error with....

function setup(){
  createCanvas(1080,1080,WEBGL);
  background(0,10,200);
  for(var i = 0; i < 500; i+=100){
    push();
    fill(i * 0.1, 100, 100);
    //line
    translate(0, 100, 0);
    line(-100, 0, i, 100, 0, i);
    pop();
  }
}

indefinit:
hmm, I thought we implemented the line primitive but it looks like it's not in master anymore.  Here's a quick fix until that happens:

```
function setup(){
  createCanvas(1080,1080,WEBGL);
  background(0,10,200);
  for(var i = 0; i < 500; i+=100){
    push();
    beginShape(LINES);
    stroke(i * 0.1, 100, 100);
    //line
    translate(0, 100, 0);
    vertex(-100, 0, i);
    vertex(100, 0, i);
    pop();
    endShape();
  }
}
```

jacobrake:
Thank you that was pretty much what I had done in my actual code.

lmccart:
yeah I'm pretty sure it was in there too, but I'm searching through history and can't find it, weird.

tildebyte:
I'm probably completely off-base here but I found `p5.Renderer3D.prototype.line` (starting at line 565) commented out in a1d296a.

lmccart:
[...]

PR #1962, opened 2017-05-25
Set gl.BLEND in texture to allow transparent pngs && added line to webgl primitives

sixhat:
This deals with issue https://github.com/processing/p5.js/issues/1878 where users need to set pngs with transparency as textures. Until now the hack would be to call `fill` with a non-opaque color before calling `texture` or to manually enable the gl.BLEND mode with 

```javascript
gl = this._renderer.GL;
gl.enable(gl.BLEND);
gl.blendFunc(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA);
```

This solution activates gl.Blend when a texture call is made.


sixhat:
The latest commit also adds `line` to the webgl primitives as explained in https://github.com/processing/p5.js/issues/1638

l3ehfar:
> ```js
> gl = this._renderer.GL;
> gl.enable(gl.BLEND);
> gl.blendFunc(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA);
> ```

hi i can't figure out where should i put this? would you give me an example with code? should I put it in my setup function?

stalgiag:
This isn't necessary anymore. Transparent pngs work out of the box now. [Here is an example.](https://editor.p5js.org/stalgiag/sketches/eTcSSKLWd)

birkb85:
> This isn't necessary anymore. Transparent pngs work out of the box now. [Here is an example.](https://editor.p5js.org/stalgiag/sketches/eTcSSKLWd)

Hi.

I tested out the example and found that it does not behave as I expected..

If I modify the code to display two clouds, one rotating and the other one staying still.
The transparency only applies for the cloud to be transparent with the background, not the other cloud.
Sorry if my sentence above does not make enough sense, please try out this code:

```
let tex;

function preload() {
  tex = loadImage('cloud.png');
}

function setup() {
  noStroke();
}

function draw() {
  background(0,200,0);
  
  push();
  texture(tex);
  rotateY(frameCount * 0.02);
  rect(-width/2, -height/2 + 40, 400, 400);
  pop();
  
  push();
  texture(tex);
  rect(-width/2, -height/2 - 40, 400, 400);
  pop();
}
```

Is this the intended behavior?
Then the transparency is really not much use, or what??

I have also attached an image demonstrating the problem.
See that the transparence does not apply between the two clouds, it only works for the background.
[Screenshot 2020-10-09 110828
](url)

Issue #1954, opened 2017-05-22
reference items that need examples


lmccart:
[this wiki section](https://github.com/processing/p5.js/wiki/Inline-documentation#adding-example-code) contains info on how to add examples. anyone, please feel free to add to this list with reference entries that could use examples added.

- [ ] https://p5js.org/reference/#/p5.MediaElement/connect
- [ ] https://p5js.org/reference/#/p5.MediaElement/disconnect
- [x] https://p5js.org/reference/#/p5.Graphics/remove
- [x] https://p5js.org/reference/#/p5.PrintWriter/write
- [x] https://p5js.org/reference/#/p5.PrintWriter/print
- [x] https://p5js.org/reference/#/p5.PrintWriter/flush
- [x] https://p5js.org/reference/#/p5.PrintWriter/close
- [ ] https://p5js.org/reference/#/p5/setMoveThreshold
- [ ] https://p5js.org/reference/#/p5/setShakeThreshold
- [ ] https://p5js.org/reference/#/p5/save
- [x] https://p5js.org/reference/#/p5.Table.js:397: method p5.Table.matchRow
- [x] https://p5js.org/reference/#/p5.Table.js:637: method p5.Table.getColumnCount
- [x] https://p5js.org/reference/#/p5.Table.js:647: method p5.Table.getRowCount
- [x] https://p5js.org/reference/#/p5.Table.js:657: method p5.Table.removeTokens
- [x] https://p5js.org/reference/#/p5.Table.js:706: method p5.Table.trim
- [x] https://p5js.org/reference/#/p5.Table.js:907: method p5.Table.setString
- [x] https://p5js.org/reference/#/p5.Table.js/getArray
- [x] https://p5js.org/reference/#/p5.TableRow.js:43: method p5.TableRow.set
- [x] https://p5js.org/reference/#/p5.TableRow.js:74: method p5.TableRow.setNum
- [x] https://p5js.org/reference/#/p5.TableRow.js:89: method p5.TableRow.setString
- [x] https://p5js.org/reference/#/p5.TableRow.js:104: method p5.TableRow.get
- [x] https://p5js.org/reference/#/p5.TableRow/getNum
- [x] https://p5js.org/reference/#/p5.TableRow/getString
- [x] https://p5js.org/reference/#/p5.XML/addChild
- [ ] https://p5js.org/reference/#/p5/orbitControl
- [ ] https://p5js.org/reference/#/p5/loadShader
[...]

PR #1964, opened 2017-05-26
Fix for part of #1954: weather api -> earthquake api in loadJSON example

mikewesthad:
Fix for part of #1954. Since openweathermap doesn't support HTTPS, swap it out for USGS's earthquake api which does support HTTPS (and bonus - doesn't require an API key).

Codepen version of examples: https://codepen.io/mikewesthad/pen/qmMPjq


lmccart:
awesome, this is great!

PR #1969, opened 2017-05-26
Changing Select Option

### LINKED ISSUE #1554: Changing options in a select created by `createSelect`
I miss a method for changing existing options in a select. Detailed access to these would be a nice addition.

mlarghydracept:
First attempt. Doesn't seem like an ideal design as it requires that you to know both the name and value of the option you want to change.

Open to other suggestions!

Closes #1554

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -575,23 +575,29 @@
     }
     self.option = function(name, value) {
       var index;
+      //see if there is already an option with this name
       for (var i = 0; i < this.elt.length; i++) {
         if(this.elt[i].innerHTML == name) {
-            index = i;
-            break;
+          index = i;
+          break;
         }
       }
+      //if there is an option with this name we will modify it
       if(index !== undefined) {
+        //if the user passed in false then delete that option
         if(value === false) {
           this.elt.remove(index);
         } else {
+          //otherwise if the name and value are the same then change both
           if(this.elt[index].innerHTML == this.elt[index].value) {
             this.elt[index].innerHTML = this.elt[index].value = value;
+          //otherwise just change the value
           } else {
             this.elt[index].value = value;
           }
         }
       }
+      //if it doesn't exist make it
       else {
         var opt = document.createElement('option');
         opt.innerHTML = name;
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -573,22 +573,25 @@
       }
       self = addElement(elt, this);
     }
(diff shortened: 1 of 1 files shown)


lmccart:
@mlarghydracept I think something like this pseudcode is close-ish:

when I call: `mySelect.option(name, value)`

```javascript
if (option exists with name) {
  if (value === false) {
    remove option
  }
  else {
    if (option.name currently == option.value) {
      option.name = option.value = value;
    } else {
      option.value = value;
      // option.name stays the same
    }
  }
} else {
  add option as normal
}
```

mlarghydracept:
Thanks for the help! I implemented basically exactly what you wrote. The only small problem is that it is impossible for a user to have two options with the same name as trying to add a new one with the same name will just overwrite the value for the old one. This seems kind of unavoidable without adding an unwieldy boolean argument into the mix. For now this feels pretty user friendly!

ggorlen:
> The only small problem is that it is impossible for a user to have two options with the same name as trying to add a new one with the same name will just overwrite the value for the old one. This seems kind of unavoidable without adding an unwieldy boolean argument into the mix. For now this feels pretty user friendly!

This led to a confusing situation for a student I mentored that resulted in them spending a couple of days trying to understand. They were creating a `select` element from a large column pulled in from a CSV file that happened to contain duplicates. The existing p5 `.option()` logic resulted in `undefined` showing up many times in their dropdown on 0.8.0. After we upgraded to 1.4.2 while working to isolate the problem to p5, the `undefined`s disappeared from the option text contents but we noticed some elements were still duplicated and some values were inexplicably being set to `undefined`.

To make matters worse, some of the columns had unescaped HTML special characters like `&`. P5 uses

```js
      const opt = document.createElement('option');
      opt.innerHTML = name;
```
to make the element. `.innerHTML` causes `"foo & bar"` to become `"foo &amp; bar"`, so the subsequent `if (this.elt[i].innerHTML === name) {` fails to detect the option as already existing. Is there a reason to use `.innerHTML` rather than `.textContent`?

Maybe I missed it, but I don't see mention that options must be unique in the documentation. I'll open a PR for improving the documentation and open an issue for potentially changing this behavior and link things up, but I wanted to leave a note here in case anyone else runs into this.

I'm not sure if it could be considered, but I'd prefer if the behavior didn't allow changing options via the `select` object, but rather provided a hook on each `option` to change its own value and text. This would allow duplicate options and generally be less magical and error-prone. I'll raise this proposal as well.

Issue #1958, opened 2017-05-23
Word spacing is handled differently for default font compared to the same font loaded from file


cqx931:
Test case: http://chenqianxun.com/testcases/p5textTest.html
In default font:
Line 1: The whole string is written with text()
Line 2: Each word is written separately with an offset of the space width
Two lines are exactly the same.

In loaded Helvetica_light.ttf (which is the default sans-serif font on mac)
Line 3: The whole string is written with text(), the spacing is different compared with default font
Line 4: Same as line 2, Each word is written separately with an offset of the space width

mlarghydracept:
Hmm after an embarrassingly long time spent looking through the code for parsing glyphs, I decided to just look at the advanceWidth property for each of the glyphs. Strangely the advancedWidth property is the exact same (278) for the space in the default and the loaded font in the given example. So I assume that this means the mistake in width isn't happening when reading info from the font table. I will keep working on this but will post as I go in case someone with more experience coding for text display wants to jump in and offer advice.

dhowe:
A couple of things to note here. First, due to a bug in the opentype library when this was initially written, spaces are handled separately from other glyphs in loaded fonts. I requested an update to opentype in p5 once this was fixed, which I think was done [here](https://github.com/processing/p5.js/pull/1903). Since space-handling changed in opentype (and now should work correctly) you might just want to try removing this special casing ([here](https://github.com/processing/p5.js/blob/29fe10f5467308398d984d84f08ace5b60db3fc3/src/typography/p5.Font.js#L137)).

dhowe:
Glancing at the code, I also notice what appears to be another small bug in textBounds (called by textWidth), which I am fixing now (though I don't think it will affect this issue).

dhowe:
Actually, I'm not even sure this is a bug. There is an assumption here that the spacing between words is _only_ determined by the width of a space (and does not take into account things like kerning), but that may not be the case for all fonts

mlarghydracept:
Hm yeah I don't know enough about type to know whether this is expected behavior. Removing the special casing you mentioned doesn't give a different result on the extra wide line but it does break the calls to `textWidth(' ')` on the loaded font.

dhowe:
I don't see the 'break' you mention. Can you try the same test with [this PR]( https://github.com/processing/p5.js/pull/1970) ?

mlarghydracept:
[...]

Issue #1957, opened 2017-05-23
Monospace font doesn't rendered correctly when loaded from file


cqx931:
The monospace font is copy pasted from my default font folder on mac, so the file itself should be correct. However, the text width for each character rendered from p5.js are not the same.
Please see the test sketch here: http://chenqianxun.com/testcases/monoSpaceFontTest/
I tried many different monospace font, otf,ttf, but none of them can get same character width when rendered from p5.js.

If I use the default monospace font in browser  `textFont("courier",100);`, the text width for each character is rendered correctly.

dhowe:
I am not sure this is a bug. I don't think that we can assume any font is true monospace, just from the name. One way to check is to use the same font file in some other environment, then compare the displayed widths of the glyphs...

PR #1970, opened 2017-05-27
Fixes bug in bounds detection

dhowe:
- Fixes bug in bounds detection where offset to edge of leftmost glyph was not being deducted from width
- Removed special cases for spaces, now that opentype has been updated
- Added some additional tests
- Removed some old comments


dhowe:
@lmccart lets not merge this until we have fixes for #1957 and #1958

PR #1974, opened 2017-05-30
Dictionary Data Object

### LINKED ISSUE #422: Data objects
@benfry and I were just talking about the [data classes from Processing](http://processing.org/reference/javadoc/core/) like `IntDict`, `FloatDict`, `StringList` etc in JS and we came to a conclusion that it would be nice to implement some of these for p5.   I am thinking:
- `NumberList`
- `StringList`
- `NumberDict`
- `StringDict`

Obviously without types in JavaScript there can just be a Dict and List, however, I think it would be nice to separate out Numbers and Strings.  For example, there are enough API differences and reference-wise it would be confusing for it to be all in one.  If you want a more advanced list or dictionary with multiple data types in JS then you would implement it yourself with native JS arrays and/or objects.   But these would be simple ways to do quick operations on lists and dictionaries with basic data types.    This aligns with Processing: `IntDict` for simple dictionary, `HashMap` from Java for more advanced.

@therewasaguy am I correct that these have not been started yet?  I am very happy to take on working on these (likely after the ITP semester) and thought it might be nice to start the discussion here.

Thoughts?

mlarghydracept:
**Not ready to merge.** I just wanted to put this here _in progress_ to get feedback on my design. I also welcome any thoughts on methods that people think would be useful.

closes #422


src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,183 @@
+/**
+ * @module Data 
+ * @submodule Dictionary
+ * @for p5
+ *@requires core
+ */
+
+var p5 = require('../core/core');
+
+//TESTED
```
lmccart: can these //TESTED notes turn into [unit tests](https://github.com/processing/p5.js/wiki/Development#testing)? ;)

src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,183 @@
...
+   if(this._validate(obj[key])) {
+     this.data[key] = obj[key];
+     this.count++;
+   }
+  }
+}
+
+
+//TESTED
+p5.TypedDict.prototype.create = function(key, value) {
```
lmccart: with a variable number of params, our convention is to leave the parens empty

src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,183 @@
...
+ * @extends p5.Element
+ * @param {String} 
+ */
+
+p5.StringDict = function(key, value) {
+  this.data = {};
+  this.count = 0;
+
+  if(arguments.length === 1) {
+    if(keys instanceof Object) {
```
lmccart: hm is it keys or key? or are these two different things?

src/data/p5.TypedDict.js
lmccart: it this a public method? seems like it could be private in which case let's call it _create.

lmccart:
looks good so far, I made a few small notes

mlarghydracept:
Okay this is getting close. All of the included unit tests work.

I feel like what it mainly needs at this point is the documentation and methods to write out to a file. I was thinking about adding the ability to write to json and csv. Any opinions on any of the above or anything else about the code?

src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,371 @@
...
+ *
+ */
+p5.TypedDict.prototype.create = function() {
+  if(arguments.length === 1) {
+    if(arguments[0] instanceof Object) {
+      this._addObj(arguments[0]);
+    }
+  }
+  else if(arguments.length === 2) {
+    var key = arguments[0];
```
lmccart: this code below looks a little familiar... could we just call this._addObj() here with the arguments obj?

src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,371 @@
...
+//  *
+//  *
+//  * @class p5.StringDict
+//  * @constructor
+//  * @extends p5.Element
+//  * @param {String}
+//  */
+p5.StringDict = function(key, value) {
+  this.data = {};
+  this.count = 0;
```
lmccart: I'm wondering about the need to keep track of count vs just calling `Object.keys(this.data).length`?

src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,371 @@
...
+//  *
+//  * A simple Dictionary class for Numbers.
+//  *
+//  *
+//  * @class p5.NumberDict
+//  * @constructor
+//  * @extends p5.TypedDict
+//  * @param
+//  */
+p5.NumberDict = function(key, value) {
```
lmccart: are all these constructors generally the same? can we pass this work on to the p5.TypedDict() constructor instead to reduce redundancy?

src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,371 @@
...
+ *
+ */
+p5.TypedDict.prototype.print = function() {
+  for (var item in this.data) {
+    console.log('key:' + item + ' value:' + this.data[item]);
+  }
+};
+
+p5.TypedDict.prototype._validate = function(key, value) {
+  return true;
```
lmccart: maybe the default method could make use of the type of object.. we already have number or string in the name, instead of having to write a custom validation method for each.

for example, on construction we could store a type:
`this.type = m.constructor.name.replace('Dict', '').toLowerCase();`

then the test would be
`return (typeof value === this.type);`

something like that?

src/data/p5.TypedDict.js
mlarghydracept: hm I guess I am a bit confused, I was trying to handle if the user says `stringDict.create('string1', 'string2'` or they say `stringDict.create( {'string1': 'string2', 'string3' : 'string4'} )`. Sorry for my confusion!

src/data/p5.TypedDict.js
mlarghydracept: Totally!

src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,371 @@
...
+
+p5.NumberDict.prototype.maxKey = function() {
+  if(this.data.count === 0) {
+    throw 'Unable to use minValue on an empty NumberDict';
+  } else if(this.data.count === 1) {
+    return Object.keys(this.data)[0];
+  } else {
+    var result = Object.keys(this.data)[0];
+    for(var i=1; i>this.data.count; i++) {
+      if(Object.keys(this.data)[i] < result) {
```
lmccart: I think this should be >

src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,371 @@
...
+    console.log('The key - ' + key + ' does not exist in this dictionary.');
+  }
+};
+
+/**
+ * Return the lowest key
+ *
+ *
+ */
+p5.NumberDict.prototype.minValue = function() {
```
lmccart: i have the feeling these four blocks could be reduced into one function, since only one symbol or line changes between them. for example, if it's just a < that flips.. you can't pass operators as arguments but you could write a function that takes 1 or -1, and multiplies both sides of the comparison by that operator. multiplying by -1 would have the effect of flipping the < to a >.

lmccart:
looking good! I left a few comments. I love the idea of being able to export to json and csv.

mlarghydracept:
Thanks! Those comments are super helpful. RE the export to json and csv -- I noticed that there are already 'blob' writers in p5. Does it make sense to make use of one of these classes that already exist elsewhere in the library?

src/data/p5.TypedDict.js
lmccart: right so the first part of the if takes care of your second case: `stringDict.create( {'string1': 'string2', 'string3' : 'string4'} ). `

the else if part could read like:

```
else if (arguments.length == 2) {
  this.addObj(arguments);
}```

File: src/data/p5.TypedDict.js (hunk 7 above)
Comment (mlarghydracept) [inline src/data/p5.TypedDict.js]: I can't figure out how to do this now that both types are falling on the `TypedDict` constructor. I tried things like `p5.StringDict.constructor = p5.StringDict` to unexpected results.

File: src/data/p5.TypedDict.js (hunk 6 above)
Comment (mlarghydracept) [inline src/data/p5.TypedDict.js]: In the latest commit, I used apply to set call the `TypedDict` constructor which works but creates a certain degree of messiness with the passing of arguments. I tried other approaches like `Object.create` and setting their constructor to the TypedDict constructor but they weren't retaining the variables made by the constructor like `data`. Not sure if there is a more elegant way.

**Comment by mlarghydracept** [comment]
I made a save function but the way it currently works is that the user needs to pass in either `'csv'` or `'json'` as the first argument. Making people write it as a string might be a bit confusing, not sure. Other options I can think of are either making a constant or making separate `saveJSON` & `saveCSV`.

**Comment by lmccart** [comment]
we actually have saveJSON() and saveTable() methods in p5 already, so it may make sense to duplicate this syntax.

**Comment by mlarghydracept** [comment]
Don't know how I overlooked those, I should probably duplicate that syntax but I should also just use them to do the work. *slaps head*

**Comment by lmccart** [comment]
@mlarghydracept is this ready for final review?

**Comment by mlarghydracept** [comment]
@lmccart Thanks for checking in! I am going to do another once over, write comments/examples, and a few more unit tests. I will do that this week and let you know when it is finished

**Comment by lmccart** [comment]
@mlarghydracept great, I'm going to close this until it's ready for review. just reopen whenever it is. this way I'll get the email notification to review.

**Comment by mlarghydracept** [comment]
I think this is ready for review!

A few notes -- I said this elsewhere in you comments on my code but I couldn't put all of the construction work in the TypedDict constructor because of the way the initializing data (the first arguments) were being retained. I tried a bunch of different approaches but maybe I am missing something.

Also this is my first time writing a new file with new functions so I wasn't completely sure what I was doing with the commenting and documentation work. I tried to follow examples from other files.

File: src/data/p5.TypedDict.js [hunk 10]
```diff
@@ -0,0 +1,591 @@
...
+
+/**
+ * Returns value stored at supplied key.
+ *
+ * @method get
+ * @param {Number|String} key that you want to access
+ * @return {Number|String} the value stored at that key
+ *
+ * @example
+ * <div><code>
```
lmccart: if the examples aren't drawing anything, let's add class="norender" so we don't have blank gray squares in the ref.

src/data/p5.TypedDict.js
```diff
@@ -0,0 +1,591 @@
...
+};
+
+/**
+ * Empties Dictionary of all key-value pairs
+ * @method clear
+ *
+ *
+ */
+
+p5.TypedDict.prototype.clear = function(){
```
lmccart: this could be simplified as
`this.data = [];`
i think?

lmccart:
@mlarghydracept looks good! thanks for this big contribution. I'm going to go ahead and merge. if you can make a later pull request with the two small notes I made, that'd be great.

will you make a note of that question about constructors and we can look at it again when we're in the same place? I couldn't quite remember what I was thinking so maybe my point didn't even make sense, but I want to take one last look together at some point just in case there was some further refinement we could make.

PR #1986, opened 2017-06-05
fixing for createSelect

mlarghydracept:
Realized that the radio attribute change caused attribute changes to select elements to log an error. Fixed that and tested with all elements. Sorry about that! I will check everything more carefully before submitting PR from now on.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1461,12 +1461,14 @@
   p5.Element.prototype.attribute = function(attr, value) {
     //handling for checkboxes and radios to ensure options get
     //attributes not divs
-    if(this.elt.firstChild.type === 'checkbox' || this.elt.firstChild.type === 'radio') {
-      if(typeof value === 'undefined') {
-        return this.elt.firstChild.getAttribute(attr);
-      } else {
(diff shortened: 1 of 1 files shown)


lib/addons/p5.dom.js
```diff
@@ -1461,12 +1461,14 @@
...
     //handling for checkboxes and radios to ensure options get
     //attributes not divs
-    if(this.elt.firstChild.type === 'checkbox' || this.elt.firstChild.type === 'radio') {
-      if(typeof value === 'undefined') {
-        return this.elt.firstChild.getAttribute(attr);
-      } else {
-        for(var i=0; i<this.elt.childNodes.length; i++) {
-          this.elt.childNodes[i].setAttribute(attr, value);
+    if(this.elt.firstChild != null) {
+      if(this.elt.firstChild.type === 'checkbox' || this.elt.firstChild.type === 'radio') {
```
lmccart: we could put this on one line like:
```javascript
if (this.elt.firstChild != null && (this.elt.firstChild.type === 'checkbox' || this.elt.firstChild.type === 'radio')) {
```

PR #1994, opened 2017-06-09
Adding an Issue Template

sakshamsaxena:
Hi !

I've created an Issue Template which would greatly help in categorising and attending the future Issues of p5.js. Let me know about any suggestions which might pop up, would be glad to discuss and add!

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,5 +1,6 @@
 <!--
-Please use the below format to report the issue in a better way.
+Hi there! If you are here to report a bug, or to discuss a feature (new or existing), you can use the below template to get started quickly. Fill out all those parts which you're comfortable with, and delete the remaining ones.
+
 To check any option, replace the space between the "[" and "]" with a "x".
 -->
 
@@ -9,7 +10,7 @@ To check any option, replace the space between the "[" and "]" with a "x".
 - [ ] Existing Feature Enhancement
 - [ ] New Feature Request
 
-#### Most Appropriate Sub Area ?
+#### Most Appropriate sub area of p5.js where it fits ?
 
 - [ ] Color
 - [ ] Core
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -1 +1,47 @@
...
+- [ ] Math
+- [ ] Typography
+- [ ] Utilities
+- [ ] WebGL
+
+<!--
+	Fill the following sections as suitable, and delete the part which is not applicable.
+-->
+
+<!-- If you found a bug, then please consider the following points -->
```
lmccart: can we say "if you know them" in there somehow? I don't want to make beginners that don't know what an OS or how to find the version number feel dumb or like they can't post an issue

ISSUE_TEMPLATE.md
```diff
@@ -1 +1,47 @@
...
+#### Details about the Bug : 
+
+- p5.js Version : 
+- Web Browser and it's Version :
+- Operating System :
+- Platform / Device : 
+- How can we reproduce this ? 
+<!--
```
lmccart: let's ask for a code snippet rather than a github repo, so we lower the chance of people posting large projects with hard to isolate bugs. ideally, the snippet is small enough to fit in the issue itself. a gist or jsfiddle or similar works too.

i think these posts are nice references for how to describe what we're asking for.
https://github.com/d3/d3/issues/new
https://gist.github.com/mbostock/370b737ce71bed0749e103b01ce1bfaf

ISSUE_TEMPLATE.md
```diff
@@ -1 +1,47 @@
...
+- [ ] Image
+- [ ] IO
+- [ ] Math
+- [ ] Typography
+- [ ] Utilities
+- [ ] WebGL
+
+<!--
+	Fill the following sections as suitable, and delete the part which is not applicable.
+-->
```
lmccart: I think we can remove this comment, and add the note to delete the remaining parts to the comment at the top

ISSUE_TEMPLATE.md
```diff
@@ -1 +1,47 @@
...
+
+- [ ] Color
+- [ ] Core
+- [ ] Events
+- [ ] Image
+- [ ] IO
+- [ ] Math
+- [ ] Typography
+- [ ] Utilities
+- [ ] WebGL
```
limzykenneth: I suggest adding `Mobile` and `Others` here. `Mobile` because it can help speed up isolation of problem (touch based or mobile sensor based etc), `Others` is just a catch all in case something really weird comes along.

sakshamsaxena:
@lmccart @limzykenneth Made a few changes based on your feedback. Do tell if there's anything which I missed :-)

ISSUE_TEMPLATE.md
sakshamsaxena: I totally agree with your suggestion, and I did actually include a "Platform/Device" point under the "Details of Bug" part, but it wasn't so clear then. Added your suggestion as a little separate question below this to better help in understanding !

lmccart:
thanks @sakshamsaxena hoping to get to review and merge this in the next few days.

PR #1997, opened 2017-06-11
Added LogBase function

TheBrenny:
Not a necessity, but for a library that needs to be simple for non-programmers, this new function can come in handy.

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -360,6 +360,19 @@ p5.prototype.lerp = function(start, stop, amt) {
 p5.prototype.log = Math.log;
 
 /**
+ * Calculates the logarithm of a number [x] with base [base]. This is the equivalent
+ * to calling [log(x) / log(b)].
+ *
+ * @method logBase
+ * @param  {Number} base The base of the log
+ * @param  {Number} x    n > 0
+ * @return {Number}      The log of [x] using [base]
+ */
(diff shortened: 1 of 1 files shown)


limzykenneth:
I'm not sure how much this would help a user as opposed to just have them doing `log(a)/log(b)` but I'm not opposed to adding this, let's see what the others think. Also see line note.

src/math/calculation.js
```diff
@@ -360,6 +360,19 @@ p5.prototype.lerp = function(start, stop, amt) {
...
+ * Calculates the logarithm of a number [x] with base [base]. This is the equivalent
+ * to calling [log(x) / log(b)].
+ *
+ * @method logBase
+ * @param  {Number} base The base of the log
+ * @param  {Number} x    n > 0
+ * @return {Number}      The log of [x] using [base]
+ */
+p5.prototype.logBase = function(b,x) {
+  return log(x) / log(b);
```
limzykenneth: Potentially use `Math.log` just to be safe and it's not that much more complex anyway.

lmccart:
thanks @TheBrenny! I think this is beyond the scope of the p5.js library right now since the workaround @limzykenneth mentioned works ok for most cases. but this could be a great starting point for an addon library if you wanted to package it up as one.

PR #2002, opened 2017-06-13
Added example to loadXML taken from p5.xml.js

bomoko:
Hi all,

This is a partial (and I hope, temporary) fix for issue #1954. Specifically, it adds an example for loadXML.
This example is taken directly from io/p5.xml.js.

I've been working on this for a few days, and wanted to get _something_ submitted.

I've actually written a number of examples of loadXML calls (both invoked inside preload and with a callback), but I've had _real_ trouble getting them to work locally inside the p5.js-website framework. 
It's possible I'm building the documentation incorrectly, but every other example that _doesn't_ use loadXML seems to run just fine (I'm getting this weird error where the parseXML function isn't being found, but the examples run just fine if I run them outside of the docs).

It's my suspicion that this is why the only loadXML example we _do_ have is invoked with the "norender" class, and why the documentation for loadXML itself _mentions_ the existence of an example, even though there is no example to be had.

So I hope this will do for the moment, I'm going to try dig into the parseXML issue a little more, and if I can't work out what's going on I'll open a separate issue for that.

Thanks :)

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -710,8 +710,8 @@ p5.prototype.parseXML = function (two) {
  * line in your sketch is executed. Calling loadXML() inside preload()
  * guarantees to complete the operation before setup() and draw() are called.
  *
- * <p>Outside of preload(), you may supply a callback function to handle the
- * object:</p>
+ * Outside of preload(), you may supply a callback function to handle the
+ * object.
  *
  * @method loadXML
  * @param  {String}   filename   name of the file or URL to load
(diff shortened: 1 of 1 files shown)


lmccart:
thanks! I suspect the local documentation building hasn't been updated since we changed the website structure. when I test, I typically just build the reference, manually drop it into my website repo and build that and look at it all. not exactly simple and convenient :\
that's my recollection but I haven't looked closely recently so please feel free to open an issue with more specifics of what you notice.

Issue #2007, opened 2017-06-16
decrementPreload set to failureCallback, even though loadImages() called from setup() or later


mjvDEV:
A little bit of context: I'm using the preload function to load two batches of strings: one is an array of filenames for the images I'm going to be loading, the other is an array of timestamps corresponding to the images. I do not use callback functions with loadStrings() inside preload(). I mention this just so you know that I do have a preload() function, in case that helps.

I'm writing a program that can frequently load and display then dump images based on user interactions. To do this I use loadImage() wrapped inside some control functions that figure out which images to load and when. I noticed that my failureCallback is consistently called after every image is loaded, even though the image displays correctly.

I did a little digging in the debugger, and found that even though I'm calling loadImages() inside setup(), (then later from mousePressed(), keyPressed, etc), the logic inside the function behaves as though I'd called it from preload(). So it treats the last argument, my failureCallback, as though it were decrementPreload, even though decrementPreload has not been pushed into the argument list (I believe that's what happens when you call a load function from preload?). Therefore my failureCallback is called despite no error occurring.

This could be a bit of a problem, because looking at the logic inside loadImage(), I don't think my failureCallback would ever be called if there actually was an error. Worse, if I hadn't included a failureCallback, I suspect that it would call my successCallback regardless of whether the load succeeded.

For now, I'm suppressing the behaviour inside my failureCallback by telling it not to do anything if it doesn't receive an error event argument, but this means I literally just have the failureCallback there as a blank placeholder.

Has anyone else encountered something like this?

limzykenneth:
I think I know what is happening here, I'll have a look and fix it if I can. However, if you can provide a minimal example demonstrating the problem it would help greatly.

mjvDEV:
Hi, sorry for the delay I was away from my desk over the weekend.

This code replicates the error. The file given to loadStrings() contains an array of image file names, which as you can see I use in loadImages():

<pre>
var files;

function preload() {
    files = loadStrings( "./data/summer-day-frames.txt" );
}

function setup() {
    canvas = createCanvas( 640, 360 );
    loadImage( "./img/plot_1207/half/" + files[0], loadSuccess, loadFail );
}

function loadSuccess( img ) {
    console.log( "p5 reports that the load succeeded: " + img );
    if (img) {
        image( img, 0, 0, 640, 360 );
    }
}

function loadFail( e ) {
    console.log( "p5 reports that the load failed: " + e );
}

</pre>

[...]

PR #2008, opened 2017-06-18
Tweak arguments parsing to fix decrementPrelaod bug

### LINKED ISSUE #2005: update documentation for loadXX() methods to note that callbacks should not be combined with preload
Trying to load a sound file in chrome, the file loads successfully but the loading element never gets removed and the canvas is never drawn.

```
var song;

function preload(){
   console.log('Executing preload();');
   song = loadSound('test.mp3', soundready, sounderror, soundloading);
}

function soundready() {
  console.log('Sound file ready.');

  // hack for chrome to play songs
  this._decrementPreload();
}

function sounderror() {
  console.log('Error loading sound file.');
}

function soundloading() {
  console.log('Loading sound file...');
}
```

I managed to remove it by calling `this._decrementPreload();` from within the onload callback method. This removes the loading element and executes the expected methods to initiate the canvas.

I'm not sure what is the cause of this perhaps the preload handling method isn't recognising a successful load?, nor do I have the time to investigate further but this is the hacky fix that works for me right now.

__p5.js version:__ v0.5.11
__p5.dom.js version:__ v0.3.3
__p5.sound.js version:__ v0.3.3

limzykenneth:
Should fix #2005 and #2007, please check to make sure I didn't miss anything obvious or break anything.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -271,8 +271,10 @@ var p5 = function(sketch, node, sync) {
 
   this._decrementPreload = function(){
     var context = this._isGlobal ? window : this;
-    context._setProperty('_preloadCount', context._preloadCount - 1);
-    context._runIfPreloadsAreDone();
+    if(typeof context.preload === 'function'){
+      context._setProperty('_preloadCount', context._preloadCount - 1);
+      context._runIfPreloadsAreDone();
+    }
   };
 
   this._wrapPreload = function(obj, fnName){
@@ -284,7 +286,7 @@ var p5 = function(sketch, node, sync) {
       for (var i = 0; i < args.length; ++i) {
         args[i] = arguments[i];
       }
-      args.push(this._decrementPreload.bind(this));
+      // args.push(this._decrementPreload.bind(this));
       return this._registeredPreloadMethods[fnName].apply(obj, args);
     }.bind(this);
   };
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -66,12 +66,6 @@ require('../core/error_helpers');
 p5.prototype.loadImage = function(path, successCallback, failureCallback) {
   var img = new Image();
   var pImg = new p5.Image(1, 1, this);
-  // var decrementPreload = p5._getDecrementPreload.apply(this, arguments);
-  // decrementPreload should never be confused with callbacks
-  // if(decrementPreload === successCallback ||
-  //    decrementPreload === failureCallback){
-  //   decrementPreload = null;
-  // }
 
   var self = this;
   img.onload = function() {
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -67,6 +67,11 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
   var img = new Image();
   var pImg = new p5.Image(1, 1, this);
   var decrementPreload = p5._getDecrementPreload.apply(this, arguments);
+  // decrementPreload should never be confused with callbacks
+  if(decrementPreload === successCallback ||
+     decrementPreload === failureCallback){
+    decrementPreload = null;
+  }
 
   img.onload = function() {
     pImg.width = pImg.canvas.width = img.width;
@@ -78,15 +83,13 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
     if (typeof successCallback === 'function') {
       successCallback(pImg);
     }
-    if (decrementPreload && (successCallback !== decrementPreload)) {
+    if (decrementPreload) {
       decrementPreload();
     }
   };
   img.onerror = function(e) {
     p5._friendlyFileLoadError(0,img.src);
-    // don't get failure callback mixed up with decrementPreload
-    if ((typeof failureCallback === 'function') &&
-      (failureCallback !== decrementPreload)) {
+    if (typeof failureCallback === 'function') {
       failureCallback(e);
     }
   };
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -66,13 +66,14 @@ require('../core/error_helpers');
 p5.prototype.loadImage = function(path, successCallback, failureCallback) {
   var img = new Image();
   var pImg = new p5.Image(1, 1, this);
-  var decrementPreload = p5._getDecrementPreload.apply(this, arguments);
+  // var decrementPreload = p5._getDecrementPreload.apply(this, arguments);
   // decrementPreload should never be confused with callbacks
-  if(decrementPreload === successCallback ||
-     decrementPreload === failureCallback){
-    decrementPreload = null;
-  }
+  // if(decrementPreload === successCallback ||
+  //    decrementPreload === failureCallback){
+  //   decrementPreload = null;
+  // }
 
+  var self = this;
   img.onload = function() {
     pImg.width = pImg.canvas.width = img.width;
     pImg.height = pImg.canvas.height = img.height;
@@ -83,9 +84,8 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
     if (typeof successCallback === 'function') {
       successCallback(pImg);
     }
-    if (decrementPreload) {
-      decrementPreload();
-    }
+
+    self._decrementPreload();
   };
   img.onerror = function(e) {
     p5._friendlyFileLoadError(0,img.src);
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -18,25 +18,6 @@ var fetchJsonp = require('fetch-jsonp');
 require('../core/error_helpers');
 
 /**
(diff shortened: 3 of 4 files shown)


bomoko:
So I'm not 100% sure, but I think that there is still a path for failure here - I came across it when applying your suggestions to my p5.js-sound PR.

```
p5.prototype.loadImage = function(path, successCallback, failureCallback) {
  var img = new Image();
  var pImg = new p5.Image(1, 1, this);
  var decrementPreload = p5._getDecrementPreload.apply(this, arguments);
  // decrementPreload should never be confused with callbacks
  if(decrementPreload === successCallback ||
     decrementPreload === failureCallback){
    decrementPreload = null;
  }
```

Okay - so assume the following. We run this from `preload()` - we _only_ supply a successCallback.
Now, when this runs, `_decrementPreload()` is now in `failureCallback()`'s position (arguments[2]).

Now, we have two problems.
First, `decrementPreload === failureCallback` returns true, so we're going to kill whatever's in decrementPreload.
Second, `decrementPreload` is now only ever going to run if there's an error (in the example above).

I hope I'm making sense here?

limzykenneth:
I understand the first and I think the implementation style in `loadJSON` may be able to take care of this (deal with argument list directly without using arguments names). The second I don't quite get, did you mean error callback instead of error?

I'll do some tests and make the necessary changes.

bomoko:
Ah, sorry, I see how I wrote that is misleading! So sorry, @limzykenneth.

So, in my example above, because we didn't supply a callback for failureCallback, the automatically injected function "_decrementPreload" is now assigned to "failureCallback", and so the only time it ("_decrementPreload()", the injected argument) will ever be run is if "failureCallback" is called.

Please let me know if that's any clearer - if not, I'll figure out how to put up some example code.
Again, sorry I'm not 100% clear.

limzykenneth:
My initial thought is that there is no way to determine whether the final argument is the failure callback or decrementPreload. This is turning out to be rather tricky, I'll try to figure this out but any help would be great.

bomoko:
@limzykenneth I've tried to annotate the code I quoted above, I hope this helps a little to explain what I was trying to say ...

```
//assume we call loadImage with only two arguments
//loadImage(s1, f1)
//so, wrt the vars inside loadImage
// path = s1 - a string
// successCallback == f1 - some user provided function
// failureCallback == undefined - since we didn't call anything
// so from the calling code's perspective the arguments array looks like [s1, f1]
// now, wrap preload adds _decrementPreload (_dP)
// so when loadImage is called, arguments actually looks like [s1, f1, _dP]

p5.prototype.loadImage = function(path, successCallback, failureCallback) {
  //path == s1, successCallback == f1, failureCallback == _dP, decrementPreload == undefined

  var img = new Image();
  var pImg = new p5.Image(1, 1, this);
  
  var decrementPreload = p5._getDecrementPreload.apply(this, arguments);
  //path == s1, successCallback == f1, failureCallback == _dP, decrementPreload == _dP  
  //ie., decrementPreload will == _dP == failureCallback **(1)**

  // decrementPreload should never be confused with callbacks
  if(decrementPreload === successCallback ||
     decrementPreload === failureCallback){ //will be true because of **(1)** above
    decrementPreload = null; //decrementPreload is now == null
  }
  //path == s1, successCallback == f1, failureCallback == _dP, decrementPreload == null
  //so at this point, the only time that _dP will be called is when failureCallback
  //invoked, since it's the only thing that points to it.
  
```

bomoko:
@limzykenneth 

> My initial thought is that there is no way to determine whether the final argument is the failure callback or decrementPreload

This is exactly the problem I identified!
So as it stands, this will only work if you pass _no_ optional callbacks, or you pass _all_ of them.
We need to guarantee that `_getDecrementPreload()` only ever returns a valid decrementPreload function, not just whatever's on the end of `arguments`

limzykenneth:
Hmm...actually, is there any reason for `decrementPreload` to be attached to the argument list? Can it just call `this._decrementPreload` directly? Looking at the implementation, there doesn't seem to be a need for it to be attached to the function like that: 

```
this._decrementPreload = function(){
  var context = this._isGlobal ? window : this;
  context._setProperty('_preloadCount', context._preloadCount - 1);
  context._runIfPreloadsAreDone();
};
```

bomoko:
I don't think so there is, actually. 
I suppose you could argue that it introduces some coupling, but no more than what's already there. And it makes the implementation a lot more transparent.

bomoko:
An alternative approach would be to have _wrapPreload wrap the first function in the arguments list in another function that will call both the original function _and_ decrementPreload.
This requires that the first callback be the "success" callback. But something like that would let us move all this decrementing logic out of the loadX functions.

limzykenneth:
Still work in progress but the latest commit https://github.com/processing/p5.js/pull/2008/commits/b359f85b4908ad3e109ab1ab0ef455a5f4800c77 is a proof of concept: I made `loadImage` call `_decrementPreload` of its context directly and it seemed to work for all my tests (in preload and setup, with and without one or both callbacks, in instance mode).

@lmccart I'm not sure who first implemented `preload` but if you can do have a look in case I missed some obvious reasons why this is not a good idea.

limzykenneth:
I've moved everything over to calling `_decrementPreload` directly, making sure they are calling their instance's `_decrementPreload` so in instance mode they don't pollute each other. All my test sketch seems to work as intended but running the automated test throws a warning preventing it from completing:

```
Warning: TypeError: undefined is not an object (evaluating 'self._decrementPreload')
    at undefined (http://localhost:9001/lib/p5.js:30366)
    at undefined (http://localhost:9001/lib/p5.js:15200)
    at onload (http://localhost:9001/lib/p5.js:14889) Use --force to continue.
```

I think it probably is because it test the code in individual units and as such cannot find references to `_decrementPreload` which is in other parts of the code base. Not sure which test threw the warning though.

If the test can be fixed and everything else checks out (this is not a terrible idea or anything else I missed) this can probably be merged.

Jared-Sprague:
@limzykenneth @lmccart @bomoko  Ok I figured out which unit test was breaking and how to fix it.  For some reason, the font_loading.js test was breaking when calling it directly from `p5.prototype.loadFont`  The way I fixed it was by calling it from a p5 instance.  See patch below.  What I can't figure out is why this is only affecting loadFont()  as far as I can see the other load* unit tests use p5.prototype.load*  and work fine, and I don't see any major difference about why loadFont is calling `self.decrementPreload()`  

So with this fix all the tests run and my [manual example](https://github.com/processing/p5.js/blob/master/test/manual-test-examples/loadingscreen/preload_success_callbacks.js) of using call backs in load* functions in preload() is also successful.

So I think this PR looks good, but I'd still like to know why the loadFont* unit test failed while the other ones are fine with no changes.


```
Index: test/unit/typography/font_loading.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- test/unit/typography/font_loading.js	(revision 75c1029db032d37116737c31d185f50c37a0a6a2)
+++ test/unit/typography/font_loading.js	(revision )
@@ -46,6 +46,7 @@
 new p5(mySketch, null, true);
 
 suite('Fonts outside preload()', function () {
+  var p5Inst = new p5(function () {}, true);
 
   var loadFont = p5.prototype.loadFont;
 
@@ -58,7 +59,7 @@
 
     test('should call success-callback with object when font loads',
       function (done) {
-        loadFont('manual-test-examples/p5.Font/acmesa.ttf',
+        p5Inst.loadFont('manual-test-examples/p5.Font/acmesa.ttf',
           function (p5Font) {
             assert.isObject(p5Font);
             assert.isTrue(p5Font instanceof p5.Font);
@@ -73,7 +74,7 @@
     );
 
     test('should call error-callback when font fails to load', function (done) {
-      loadFont('invalid-path',
+      p5Inst.loadFont('invalid-path',
         function () {
           assert.ok(false);
           done();
```

limzykenneth:
@Jared-Sprague Awesome, thanks!

If it is only affecting loadFont and not the other loading functions then I think it probably have more to do with p5.Font itself (which is in a separate module) or with opentype itself.

Jared-Sprague:
@limzykenneth I wasn't sure how to add a commit to this PR that fixes the loadFount unit test.  I can't commit to your fork.  It's probably easiest if you just apply the patch in my comment above to your fork and resync the PR.

limzykenneth:
@Jared-Sprague No worries, pushed your fixed. I've had a closer look at the tests and my initial guess is more accurate, the tests for `loadFont` and other `loadX` are different in that `loadFont` tries to create an instance of p5 without other parts that was needed while `loadX` is just doing some pretty basic sanity tests and not really using `preload` properly or callbacks extensively.

In either case, tests pass now and maybe we can revisit the tests after #2023 is done.

lmccart:
this all looks good to me, thanks everyone for helping with this fix!

PR #2009, opened 2017-06-20
Added randomPoisson for generating discrete natural numbers following the poisson distribution

sixhat:
Although many random events can be generated with uniform and gaussian distributions, there are classes of events that are better simulated with a Poisson distribution, like the life expectancy of a light bulb in hours or the number of patients arriving in an emergency room during a certain period. 

Because of this need I've implemented the randomPoisson number generator in random.js as it holds the random functions. It implements the 1969 Donald Knuth algorithm modified to have numeric stability for large lambdas, that is pretty straightforward , but doesn't include optimisations of most recent algorithms. In any case I'll be happy to improve upon this version later if users find this version slow.

### DIFF
--- a/src/math/random.js
+++ b/src/math/random.js
@@ -243,15 +243,14 @@ p5.prototype.randomGaussian = function(mean, sd)  {
 };
 
 /**
- *
  * Returns a random number (or an array of numbers) fitting a Poisson
- * distribution using Donald Knuth algorithm.
+ * distribution using Donald Knuth algorithm adapted by Junhao.
  * <br><br>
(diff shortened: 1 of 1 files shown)


sixhat:
Please hold this PR as currently this only works for small values of `mean`. I'll implement a better version soon and I'll get back to this.

sixhat:
This latest version https://github.com/processing/p5.js/pull/2009/commits/962af5a3370da2055540b6e4ee6cbd8fc7da519b solves the number stability problem of the initial version and now works with large numbers also.

lmccart:
this is great, but I think it's beyond the scope of p5.js library right now.. it could be a great starting point for an addon library though.

PR #2018, opened 2017-06-22
added validateNumParameter & test-implemented to 2d_primitives

almchung:
This is the first version of `validateNumParameters`.

This version needs comments on code design:  As the first part of designing friendly error system, I decided to start with the functions that only take `Number` type as their required parameters. The logic behind this: In p5.js (excluding `dom` and `sound`), there are ~80 functions whose parameters should all be `Number` types. If we handle parameter-checking for those (and validate the implemention with unit test), that will cover ~1/3 of all functions in p5.js.

Here is the current flow for using validateNumParameters:

`validateNumParameters` takes the caller's function name (`func`) as `string`, a list of arguments (`input`) to be validated, and the number of numeric parameters (`length`) that must be present in the arguments.

`validateNumParameters` checks the first `length` parameters of `input`, and then checks for `EMPTY_VAR` or `WRONG_TYPE` errors.

_The current version throws an error and leaves a friendly-error-msg-log when it detects a missing parameter or a wrong type parameter, bypassing message function in error_helper.js (mainly for unit test purpose)._

`validateNumParameters` is currently test-implemented for 2d primitives functions. I added unit tests for `arc()`,`ellipse()`,`line()`,`rect()`, and `triangle().` `line()`, `quad()`, `point()` still need unit tests for 3D (webgl) cases.

After this pull request, I plan to work on the remaining number-only functions next.
Then, I will expand to other cases based on comments.


lmccart:
hey @almchung, looking good. two main questions:
1. what about functions that take varying numbers of arguments? right now it looks like you pass in a single number as the third parameter, but with color() for example, it could take 1, 2, or 4 arguments.
2. maybe this is the next step, but is there a way to parse this directly from the documentation prior to the function definition, so that the documentation and functionality always stay sync'd, rather than introducing the possibility of a bug with the two of them not matching up?

src/core/2d_primitives.js
```diff
@@ -75,6 +75,12 @@ p5.prototype.arc = function(x, y, w, h, start, stop, mode) {
   for (var i = 0; i < args.length; ++i) {
     args[i] = arguments[i];
   }
+  // check with FES:validateNumericParameters
+  var validatePar = (this._validateNumParameters('arc', args, 6));
+  if (validatePar[0] === false) {
+    console.log(validatePar[3]);
+    throw new Error(validatePar[1]);
```
lmccart: I think we just want to log to console, rather than throw errors, so the program could still potentially run. or do you feel differently?

src/core/error_helpers.js
```diff
@@ -105,6 +105,59 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+    } else if (argType !== 'number') {
+      message = 'FES: ' + func + ' was expecting a number' +
+        ' for parameter #' + p + ' (zero-based index), received ';
+      // Wrap strings in quotes
+      message += 'string' === argType ? '"' + args[p] + '"' : args[p];
+      message += ' instead.';
+      return [false, 'WRONG_TYPE', p, message];
+    }
+  }
+  return [true];
```
lmccart: could it just return true, rather than an array containing true? this just speeds things up ever so slightly if cut out the array creation and array access step.

src/core/error_helpers.js
```diff
@@ -105,6 +105,59 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+*  "ellipse was expecting a number for parameter #1,
+*           received "foo" instead."
+*/
+function validateNumParameters(func, input, length) {
+  var message;
+  if (arguments.length < 3) {
+    message = 'Missing input values for validating numeric parameters';
+    return [false, 'INIT_VALNUMPAR_FAIL', 0, message];
+  }
+  var args = new Array(length);
```
lmccart: is it necessary to copy over all the args into a new array?

src/core/2d_primitives.js
almchung: This was mainly for simplifying the unit test but easily swapped out. I will remove comment-out for the log.console() in error_helpers.js and introduce back the function.

src/core/error_helpers.js
almchung: Oops! Will remove this.

src/core/error_helpers.js
almchung: Abandoning "throwing error" and re-introducing console.log message will eventually get rid of this array. I will work on this right away!

almchung:
Answering the main two questions, I will try to find a more general solution like you suggested.
One idea I currently have is to read data.json generated through jsdoc, but it has its own disadvantages, so i'm considering other options as well.

src/color/p5.Color.js
```diff
@@ -47,7 +47,7 @@ p5.Color = function(renderer, vals) {
   this.levels = this._array.map(function(level) {
     return Math.round(level * 255);
   });
-
+  this.name = 'p5.Color';   // for friendly debugger system
```
almchung: @lmccart 
Because the way we declared our classes, typical ways to check class name would not work ([ref1](https://stackoverflow.com/questions/1249531/how-to-get-a-javascript-objects-class), [ref2](https://stackoverflow.com/questions/12944719/how-does-devtools-determine-an-objects-constructors-name)). 
I tried to resolve the issue by adding `name` property into p5 classes.

Currently affected classes are: 
`color/p5.Color`
`core/p5.Element`
`core/p5.Graphics`
`core/p5.Renderer`
`core/p5.Renderer2D`
`image/p5.Image`
`io/p5.Table`
`io/p5.TableRow`
`io/p5.XML`
`math/p5.Vector`
`typography/p5.Font`
`webgl/p5.Geometry`
`webgl/p5.Matrix`
`webgl/p5.RendererGL`

src/core/error_helpers.js
```diff
@@ -105,6 +110,110 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+        ' for parameter #' + p + ' (zero-based index), received ';
+      // Wrap strings in quotes
+      message += 'string' === argType ? '"' + args[p] + '"' : args[p];
+      message += ' instead.';
+      return [false, 'WRONG_TYPE', p, message];
+    }
+  }
+  return [true];
+}
+function validateParameters(func, args) {
```
almchung: @lmccart 
this new function, `validateParameters()` is more generalized parameter validation function which looks up `doc/references/data.json` to check parameters.

src/core/error_helpers.js
```diff
@@ -105,6 +110,110 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+      message = 'FES: ' + func + ' was expecting a number' +
+        ' for parameter #' + p + ' (zero-based index), received ';
+      // Wrap strings in quotes
+      message += 'string' === argType ? '"' + args[p] + '"' : args[p];
+      message += ' instead.';
+      return [false, 'WRONG_TYPE', p, message];
+    }
+  }
+  return [true];
+}
```
almchung: @lmccart 
this new function, `validateParameters()` is more generalized parameter validation function which looks up `doc/references/data.json` to check parameters.

src/core/error_helpers.js
almchung: Uncommented LN 61-77 to enable friendly error msg.

test/unit/core/error_helpers.js
```diff
@@ -1,4 +1,94 @@
...
+        Error, 'got unwanted exception');
+    });
+    test('documentation look up - arc()', function() {
+      assert.doesNotThrow(function() {
+          p5.prototype._validateParameters('arc',
+            ['1', 1, 10.5, 10, 0, Math.PI, 'pie']);
+        },
+        Error, 'got unwanted exception');
+    });
+    test('documentation look up - ambientLight()', function() {
```
almchung: This generates no err msg.

test/unit/core/error_helpers.js
```diff
@@ -1,4 +1,94 @@
...
+      assert.doesNotThrow(function() {
+          var c = myp5.color(255, 204, 0);
+          p5.prototype._validateParameters('ambientLight', [c]);
+        },
+        Error, 'got unwanted exception');
+    });
+  });
```
almchung: This generates no err msg.

test/unit/core/error_helpers.js
```diff
@@ -1,4 +1,94 @@
...
+        },
+        Error, 'got unwanted exception');
+    });
+    test('documentation look up - arc()', function() {
+      assert.doesNotThrow(function() {
+          p5.prototype._validateParameters('arc',
+            ['1', 1, 10.5, 10, 0, Math.PI, 'pie']);
+        },
+        Error, 'got unwanted exception');
+    });
```
almchung: This generates err msg:
`> p5.js says: FES: arc was expecting Number for parameter #0 (zero-based index), received "1" instead. [http://p5js.org/reference/#p5/arc]`

test/unit/core/error_helpers.js
```diff
@@ -1,4 +1,94 @@
...
+    test('documentation look up - arc()', function() {
+      assert.doesNotThrow(function() {
+          p5.prototype._validateParameters('arc',
+            [1, 1, 10.5, 10]);
+        },
+        Error, 'got unwanted exception');
+    });
```
almchung: This generates console.log err msg:
`> p5.js says: FES: It looks like arc received an empty variable in spot #4 (zero-based index). If not intentional, this is often a problem with scope: [link to scope]. [http://p5js.org/reference/#p5/arc]
p5.js:42683 > p5.js says: FES: It looks like arc received an empty variable in spot #5 (zero-based index). If not intentional, this is often a problem with scope: [link to scope]. [http://p5js.org/reference/#p5/arc]`

test/unit/core/error_helpers.js
```diff
@@ -1,4 +1,94 @@
...
+
+  // unit tests for validateParameters
+  suite('validateParameters', function(){
+    test('documentation look up - arc()', function() {
+      assert.doesNotThrow(function() {
+          p5.prototype._validateParameters('arc',
+            [1, 1, 10.5, 10, 0, Math.PI, 'pie']);
+        },
+        Error, 'got unwanted exception');
+    });
```
almchung: This generates no `console.log` err msg.

lmccart:
@almchung what do you think about creating a new branch to develop this in? I am thinking this might let us review and merge things in faster as they come even if the full version isn't quite ready for release. we can then choose certain points at which to pull the changes into master and make releases along the way. if this makes sense to you, feel free to create a new branch on the p5.js repo and PR against that one. if not, we can discuss :)

almchung:
@lmccart Sounds good! I'll create a new branch today.

almchung:
branch moved to `processing:friendly-err-sys`

almchung:
branch moved to `processing:friendly-err-sys`

PR #2021, opened 2017-06-23
Cache strokeStyle and fillStyle for performance

### LINKED ISSUE #2011: Accessing strokeStyle and fillStyle of drawing context slow
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [x] Core
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.5.11
- Web browser and it's version: Chrome Version 61.0.3128.0 (Official Build) dev (64-bit)
- Operating System: OS X 10.11.6
- Steps to reproduce this: Repeatedly call "vertex"

In profiling a sketch I'm working on that makes heavy use of `vertex`, I found that calls were quite slow, and specifically due to the calls internally to `_getStroke` and `_getFill` (see screenshot).

screen shot 2017-06-20 at 13 33 54

It seems that its slow for it to access the `fillStyle` and `strokeStyle` of the drawing context. Although I don't know why this would be the case, I found that by locally caching these values, and then updating the cache in every place where they are set, performance improves significantly:

```
p5.Renderer2D.prototype._getFill = function(){
  return this.cachedFillStyle;
};

p5.Renderer2D.prototype._setFill = function(fillStyle){
    this.drawingContext.fillStyle = fillStyle
    this.cachedFillStyle = this.drawingContext.fillStyle;
};

p5.Renderer2D.prototype._getStroke = function(){
  return this.cachedStrokeStyle
};

p5.Renderer2D.prototype._setStroke = function(strokeStyle){
    this.drawingContext.strokeStyle = strokeStyle
    this.cachedStrokeStyle = this.drawingContext.strokeStyle;
}
```

this will, of course, fail if someone tries to modify the stroke or fill style outside of these helper functions. There may be another solution by modifying the part of the `vertex` that calls this.

```
    vert[5] = this._renderer._getFill();
    vert[6] = this._renderer._getStroke();
```

If the helper function-based approach looks good, I can probably package up my changes as a PR. Let me know!

wxs:
Fixes #2011 

Many code paths repeatedly reset or access the stroke or the fill, for
instance when drawing shapes every call to vertex() queries the fill and
stroke styles. This seems to be quite an expensive operation, and for
almost every use-case is not changing between calls to vertex.

Here I replace all direct accesses to the canvas' drawingContext's
strokeStyle and fillStyle to go through accessors that cache the result.
This shows significant performance improvements, at least on Chrome on
Mac.

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -956,7 +956,7 @@ p5.Renderer2D.prototype._getFill = function(){
 p5.Renderer2D.prototype._setFill = function(fillStyle){
   if (fillStyle !== this._cachedFillStyle) {
     this.drawingContext.fillStyle = fillStyle;
-    this._cachedFillStyle = this.drawingContext.fillStyle;
+    this._cachedFillStyle = fillStyle;
   }
 };
 
@@ -967,7 +967,7 @@ p5.Renderer2D.prototype._getStroke = function(){
 p5.Renderer2D.prototype._setStroke = function(strokeStyle){
   if (strokeStyle !== this._cachedStrokeStyle) {
     this.drawingContext.strokeStyle = strokeStyle;
-    this._cachedStrokeStyle = this.drawingContext.strokeStyle;
+    this._cachedStrokeStyle = strokeStyle;
   }
 };
 
@@ -1301,8 +1301,8 @@ p5.Renderer2D.prototype.push = function() {
 p5.Renderer2D.prototype.pop = function() {
   this.drawingContext.restore();
   // Re-cache the fill / stroke state
-  this.cachedFillStyle = this.drawingContext.fillStyle;
-  this.cachedStrokeStyle = this.drawingContext.strokeStyle;
+  this._cachedFillStyle = this.drawingContext.fillStyle;
+  this._cachedStrokeStyle = this.drawingContext.strokeStyle;
 };
 
 module.exports = p5.Renderer2D;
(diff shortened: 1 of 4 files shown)


src/core/p5.Renderer2D.js
```diff
@@ -953,13 +946,29 @@ p5.Renderer2D.prototype.strokeWeight = function(w) {
 };
 
 p5.Renderer2D.prototype._getFill = function(){
-  return this.drawingContext.fillStyle;
+  return this.cachedFillStyle;
```
lmccart: can we call this this._cachedFillStyle? similar with this._cachedStrokeStyle?

lmccart:
this is great! see the one comment i left inline.

wxs:
Hey @lmccart I've addressed your comment!

I've also made a couple more changes.
 - The fillStyle was being set directly in `p5.Font.js`, I've fixed that
 - I found a further performance improvement, due to the fact that setting a fillStyle or a strokeStyle doesn't necessarily return the same string back if you access it later. This slightly complicates the tests.
 -  Some tests fail intermittently for me due to a timeout. This makes development really frustrating so I increased the timeout time

lmccart:
this is a great improvement, thank you!

kadamwhite:
This is excellent!

PR #2032, opened 2017-06-28
Add example for httpGet

rbwendt:
### Changes

* Add an example for the `httpGet` method.

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -782,6 +782,37 @@ p5.prototype.loadXML = function() {
  * @param  {Function}      [errorCallback] function to be executed if
  *                                    there is an error, response is passed
  *                                    in as first argument
+ * @example
+ * <div class='norender'><code>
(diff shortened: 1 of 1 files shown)


lmccart:
thanks!!

limzykenneth:
@lmccart This may not work as expected. `preload` will not wait for `httpGet`, because it is based entirely on callbacks and not return anything, and thus may exit earlier than the response coming back.

rbwendt:
@limzykenneth That's why draw waits for `earthquakes` to be present. Is that not sufficient?

limzykenneth:
@rbwendt ah, right. I see, I missed that part. In that case it is fine. Thanks!

lmccart:
@rbwendt I wonder if it could be confusing that this is in preload. generally we say to only place loadX() methods there, since these are the only ones that preload waits for. could we make a small edit to just rename what is currently preload to setup?

PR #2033, opened 2017-06-28
validateParameter() & p5._friendlyParamError()

almchung:
Migrating from `almchung:validateNumPar`
This is the first version of validateParameters() and p5._friendlyParamError().

Here is the current flow for using validateNumParameters:

`validateParameters()` takes an array of input arguments to be validated, looks up `docs/reference/data.json` and checks for `EMPTY_VAR`, `WRONG_CLASS` or `WRONG_TYPE` errors. This design supports both multi-type and multi-format cases. 

list of helper functions:
`lookupParamDoc(func)` returns `queryResult` which is an array of correct formats
`testParamFormat(args, format)` returns `errorArray` for each format
`testParamClass(param, types)` returns `true`/`false` for individual class test
`testParamType(param, types)` returns  `true`/`false` for individual type test

Friendly-error-msg-logs are produced through `p5._friendlyParamError()`.

`validateParameters()` is currently test-implemented for some of 2d primitives functions:`arc()`,`ellipse()`,`line()`,`rect()`, and `triangle()`.


almchung:
copied from the old branch:

> Because the way we declared our classes, typical ways to check class name would not work ([ref1](https://stackoverflow.com/questions/1249531/how-to-get-a-javascript-objects-class), [ref2](https://stackoverflow.com/questions/12944719/how-does-devtools-determine-an-objects-constructors-name)). 
> I tried to resolve the issue by adding `name` property into p5 classes.
> 
> Currently affected classes are: 
> `color/p5.Color`
> `core/p5.Element`
> `core/p5.Graphics`
> `core/p5.Renderer`
> `core/p5.Renderer2D`
> `image/p5.Image`
> `io/p5.Table`
> `io/p5.TableRow`
> `io/p5.XML`
> `math/p5.Vector`
> `typography/p5.Font`
> `webgl/p5.Geometry`
> `webgl/p5.Matrix`
> `webgl/p5.RendererGL`

src/core/error_helpers.js
```diff
@@ -105,6 +110,159 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+  }
+  return errorArray;
+}
+// testClass() for object type parameter validation
+// Returns true if PASS, false if FAIL
+function testParamClass(param, types){
+  var result = false;
+  for (var i = 0; i < types.length; i++) {
+    if (param.name === types[i]) {
+      result = true;    // class name match, pass
```
lmccart: we could change these lines to `return true`, then outside of the for loop have a `return false` line. this would add two small performance improvements (1) no need to create variable called result each time, and (2) we can break out of the loop as soon as a true value is found.
this is the same for testParamType() below.
```javascript
function testParamClass(param, types){
  for (var i = 0; i < types.length; i++) {
    if (param.name === types[i]) {
      return true;
    }
    else if (param.name === 'Constant') {
      return true;
    }
  }
  return false;
}
```

src/core/error_helpers.js
```diff
@@ -105,6 +110,159 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+    }
+    if (minErrInd > -1){
+      // generate error msg with a smaller number of errors
+      for(var n = 0; n < errorListArray[minErrInd].length; n++) {
+        p5._friendlyParamError(errorListArray[minErrInd][n], func);
+      }
+    }
+  } else {
+    var errorArray = testParamFormat(args, arrDoc[0]);
+    if (errorArray.length > 0){
```
lmccart: we can remove this if statement. if the length of the array is 0, the following for loop will just skip to the end without running

src/core/error_helpers.js
```diff
@@ -105,6 +110,159 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+      if (format[p].optional !== true) {
+        error = {type:'EMPTY_VAR', position: p};
+        errorArray.push(error);
+      }
+    } else {
+      var types = format[p].type.split('|'); // case accepting multi-types
+      var pass;
+      if (argType === 'object'){             // if obejct, test for class
+        pass = testParamClass(args[p], types);
+        if (!pass) {  // if fails to pass
```
lmccart: we could remove the creation of the variable pass, and just rewrite this line as
`if (!testParamClass(args[p], types)) {`
same with the other one in the else statement below

src/core/error_helpers.js
```diff
@@ -105,6 +110,159 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+    var minErrCount = 0;
+    for (var j = 0; j < errorListArray.length; j++) {
+      var numErr = errorListArray[j].length;
+      if (numErr >= arrDoc.length){   // exclude clean cases
+        if(j===0){  // the first case
+          minErrCount = numErr;
+          minErrInd = j;
+        }else{
+          if(minErrCount > numErr){
+            minErrInd = j;
```
lmccart: does minErrCount also need to get updated here?
`minErrCount = numErr`

src/core/error_helpers.js
```diff
@@ -105,6 +110,159 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+    for (var i = 0; i < arrDoc.length; i++) {
+      errorListArray.push(testParamFormat(args, arrDoc[i]));
+    }
+    // compare errors from all formats
+    var minErrInd = -1;
+    var minErrCount = 0;
+    for (var j = 0; j < errorListArray.length; j++) {
+      var numErr = errorListArray[j].length;
+      if (numErr >= arrDoc.length){   // exclude clean cases
+        if(j===0){  // the first case
```
lmccart: the typical way to eliminate the need for this case is just to set minErrCount to a very high number like 999999, and setting minErrInd to 0 already. then the first object in the array will definitely meet the condition that numErr < minErrCount. does that make sense?

src/core/error_helpers.js
```diff
@@ -105,6 +110,159 @@ p5._friendlyFileLoadError = function (errorType, filePath) {
...
+    var errorListArray = [];
+    for (var i = 0; i < arrDoc.length; i++) {
+      errorListArray.push(testParamFormat(args, arrDoc[i]));
+    }
+    // compare errors from all formats
+    var minErrInd = -1;
+    var minErrCount = 0;
+    for (var j = 0; j < errorListArray.length; j++) {
+      var numErr = errorListArray[j].length;
+      if (numErr >= arrDoc.length){   // exclude clean cases
```
lmccart: a larger picture question here -- it seems to me that rather than checking the args against every possible param combo and then checking that at least one of them didn't fail, we could just invert this and just look for one that matches, and stop checking as soon as we find it. this would cut down checking time for functions that have many different possible arg combos. this doesn't need to be addressed before we merge, but it may be something to revisit later?

lmccart:
hey @almchung looking good. this pull request is still against master though. is there a way to submit the pull request directly to the validateParam branch of the p5.js repo so we can use that as the experimental friendly error development area?

src/core/error_helpers.js
lmccart: I'm happy to sketch out ideas for the logic of this with you, when I'm back in LA!

src/core/error_helpers.js
almchung: Yes! Fixed.

src/core/error_helpers.js
almchung: That's great way to do it. I will still have the default value for `minErrInd = -1;` for the case there's no error.

src/core/error_helpers.js
almchung: I agree that would optimize the validation process a lot. 
The reason I wrote this way was: 
when a function allows multiple-format cases with the same number of parameters, we want to pick the format that gives the min # of errors for validation. If we look for the first error match and stop, the validation function can 1) flag false positive (ex: user used the second format) or 2) give biased-suggestions to use the first format instead of other options.

src/core/error_helpers.js
almchung: Certainly. Removed! :)

src/core/error_helpers.js
almchung: This one's changed too.

almchung:
Created `friendlyErrSys` branch and moved the base branch of the pull request to it. Hopefully I did it correctly this time....

PR #2034, opened 2017-06-29
added example for dragOver and dragLeave

bomoko:
Partially addresses #1954 - specifically, it adds examples for  [dragOver](https://p5js.org/reference/#/p5.Element/dragOver) and[dragLeave]( https://p5js.org/reference/#/p5.Element/dragLeave). 

I've used the same example for both, since in the example code I use both functions.
I'm happy to simplify and make each example only use the appropriate function, but I _think_ the example makes sense as it stands.

### DIFF
--- a/src/core/p5.Element.js
+++ b/src/core/p5.Element.js
@@ -743,6 +743,35 @@ p5.Element.prototype.touchEnded = function (fxn) {
  * @param  {Function} fxn function to be fired when mouse is
  *                    dragged over the element.
  * @return {p5.Element}
+ * @example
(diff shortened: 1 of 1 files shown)


lmccart:
hey @bomoko this is great thanks! i think it would be good to separate the examples just so it's less overwhelming for someone that's not so familiar with the code. i've merged this in for now, but if you have a chance to make an update, that'd be great!

bomoko:
@lmccart fantastic! That's what I was a little worried about. Will update as soon as I'm able.

PR #2035, opened 2017-06-30
simplified the dragOver and dragLeave examples

bomoko:
Partial fix for #1954 . 

As requested by @lmccart, I've simplified the two examples added last time.
I also realised there was some spacing issue in the last commit, this sorts that out too.


lmccart:
great!

PR #2037, opened 2017-07-02
fix a few yuidoc type errors

Spongman:
hey, i'm working on a tool that uses the reference/data.json schema and i noticed that there are a few missing things or inconsistencies. this should fix most of those...

except:
- some of the sound stuff is still wrong
- directionalLight()/pointLight(), wow ;-)


lmccart:
this is great, thank you @Spongman!! 

one note -- the p5.sound file is actually generated from the p5.js-sound repo here: https://github.com/processing/p5.js-sound. would you be willing to make the changes over there as well so we can make sure it gets incorporated into future version. I'm so sorry for the duplication of work. if you don't have time, perhaps @jvntf who is working p5.sound this summer could work on this? (cc @therewasaguy)

Spongman:
Oh, sure. It looks like it should be pretty easy to merge it in.

Issue #2041, opened 2017-07-05
clarify API reference for color(), fill(), stroke() to specify CSS string + alpha value is not supported


pehagg:
#### Nature of issue?

- [X] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [X] Color
- [ ] Core
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 0.5.11
- Web browser and it's version: Chrome 59.0.3071.115
- Operating System: macOS
- Steps to reproduce this:

Both

```javascript
var color = color("#010203", 200);
```

and 

```javascript
fill("#010203", 200);
```

result in:

```
p5.js:9499 Uncaught Error: [object Arguments]is not a valid color representation.
    at p5.Renderer2D.p5.Color._parseInputs (p5.js:9499)
    at new p5.Color (p5.js:8937)
    at p5.color (p5.js:8620)
    at <anonymous>:1:1
```

with-a-k:
I'd like to take a look at this one.

PR #2043, opened 2017-07-05
issue #2041 - allow colors to be initialized with a CSS string and an…

with-a-k:
… alpha value

Addresses issue #2041 
Feedback very welcome, please comment.

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -460,7 +460,7 @@ p5.Color._parseInputs = function() {
       return results;
     }
 
-  } else if (numArgs === 1 && typeof arguments[0] === 'string') {
+  } else if (typeof arguments[0] === 'string') {
 
     var str = arguments[0].trim().toLowerCase();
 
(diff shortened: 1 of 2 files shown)


sixhat:
Please note that W3 colour Level 4 spec already specifies that colours specified by hexadecimal strings can include alpha information in the form of '#rrggbbaa' or '#rgba'. See https://developer.mozilla.org/en/docs/Web/CSS/color_value for example. 

Therefore it is nonsense to call `color('#rgb', alpha)` as the user is saying "I want a solid colour, I want transparency" Which one does the user want? Well, I assume that most people calling the function like this want to define the "solid colour with a certain transparency".

In any case, it is necessary to check that the user doesn't call the function with two incompatible alphas like `color('#rgba', alpha2).

limzykenneth:
I tend to agree with @sixhat here, the more we can avoid non-standard implementation the better.

However, drawing from that, is it right to assume the docs is wrong for providing an option for `color(value, [alpha])`? Because although `color("#000000", 50)` is nonsensical, would `color("black", 50)` make sense? If so it will need to be taken cared of because I don't think that works out of the box.

In either case, the docs will need to be updated to reflect what can and cannot be done more accurately.

sixhat:
I don't want to make a big issue of this, and really I just want people to be aware of the issues. Standards are a good thing, I think. So, for me `color("black", 50)` is also problematic as black is one of the named solid colours (opaque) in the standard. See https://www.w3.org/TR/css-color-4/#named-colors where it clearly states that _named colours_ are opaque. But I understand that in P5 `color(color , alpha)` is perceived by the users as taking the solid colour and _adding_ transparency to it.

For the potential problem with gray colours the standard decided on the use of the `gray` keyword in the form `gray(level, [alpha])`, very similar how `color` currently works. 

I don't know how to bring all these things together, but at least is important that the community is aware of them.

with-a-k:
@sixhat I didn't know this, and my assumption is that the user that submitted the issue didn't either. In any case, `#rgba` and `#rrggbbaa` don't look like they're supported by p5 currently, so I might go ahead and add that support instead of allowing the user to also supply an integer alpha; that seems like something that would go on a different branch than this one to me.

@limzykenneth Based on all this, I think updating the docs and the API reference is the right choice here.

lmccart:
thanks all, I agree with the decision to support #rgba and #rrggbbaa and to update the docs to be more clear that a CSS string with alpha value won't work. I'm going to close this PR, and rename issue #2041. @with-a-k FYI the documentation is built from inline docs in this repo, more info here https://github.com/processing/p5.js/wiki/Inline-documentation. feel free to take a pass at clarifying the docs if you'd like.

PR #2051, opened 2017-07-09
[WIP] Modularising p5

sakshamsaxena:
Hello!

This is still Work in Progress due to unsuccessful bundling of Typography, IO, and WebGL modules. The rest ones of Core, Color, Math, Utility, Images are perfect and pass all tests as well.

I specifically need help in identifying why those 3 are blatantly failing tests. Preliminary investigation using logs showed that the p5 Prototype attached to window had barely 10-12 function attached, which is strange because the other bundles are being built using the same config, and they pass the tests too.

If someone could take a look or provide any ideas, it'd be awesome! 

@limzykenneth @kadamwhite @shiffman @lmccart Your reviews would be greatly appreciated :smile:


limzykenneth:
I'm not completely up to date with your work so I'll just comment on the IO part which I was looking into before. 

I think the tests for IO may need to be rewritten as the current test structure is a bit weird, it needs to test the `preload` function but it's using a hack by making `preload = true` and run the tests directly without creating a sketch or p5 instance. That may cause some problems especially when it is trying to reference `this`. If possible, it may be better to create instances of p5 before running the tests on them instead of running off the prototype object.

I've also looked into using something like [SinonJS](http://sinonjs.org) to create fake server and spies which for a previous project I found to be quite useful for testing APIs/AJAX calls.

Another thing regarding modularization which you may have come across as well is that `loadFont` is bundled with `io/files.js` whereas the test for it is in `typography/font_loading.js`. I feel that it may be more at home to be bundled with `typography/loading_displaying.js` since to use external fonts you will have to load it in first.

lmccart:
I think @limzykenneth is correct about typography and IO. I will work on moving this around.
I'm not totally sure about webgl, but can we move forward without it for now since it is in very active development? Once it is more stable, I think we will be able to see the issue more clearly.

sakshamsaxena:
@lmccart I think this is good to go. Indeed, renaming those export names did the trick (even though there is p5.Color etc classes, which somehow don't really hinder). Would be great to somehow roll this out as a beta release, for users to try it out! It does work theoretically, let's see how it works in the real situations.

lmccart:
great! two more questions:

I think the core module should really include more than just the core folder. I think the breakdown should be like this:

core/
color/
events/
utilities/
image/ ?
math/  ?

the other independent modules would be:

webgl/
typography/
io/

second, right now it's not possible to include both io and typography modules linked in one project, because they both include the core code inside them. so we may want to think about an interface for people to select the modules they want a la modernizr.com that builds the module with only one copy of core inside.

sakshamsaxena:
@lmccart Didn't really get you. Could you elaborate a teensy bit ? Thanks! :sweat_smile:

lmccart:
the p5.Core.js module should include the files from all of those folders.
this will allow the core build to have the basics of setting fill, stroke, background (which are useful for things like drawing text), as well as mouse/keyboard events.

then, there we could have three additional modules:
p5.Webgl.js
p5.Typography.js
p5.IO.js

lmccart:
secondly.. right now if you try to link in both p5.Typography.js and p5.IO.js, there are errors because both modules include copies of the code from p5.Core.js. it would be ideal if these modules did not include that core code but instead could be included along with p5.Core.js. alternatively, the modules could be build dynamically after the user decides which modules they'd to include in the build (similar to what modernizr.js does).

sakshamsaxena:
Added a fresh commit. Now you can do a call to grunt like `grunt combineModules:color:events:io` and it'll create a bundle containing core, color, events, and io.

sakshamsaxena:
### Summary 

> A proposed implementation of a special build task of p5.js, where user select components of p5.js are considered for the build. 

### Implementation Details :

A new build task is added to the `build/tasks` folder, which is similar to the existing `browserify` task (many many thanks to @kadamwhite for his contribution there!). It is invoked manually by `grunt combineModules:module_1:module_2`, where `module_n` is essentially the component name which you want to have, except from the Core itself, which is by default included in all such combinations.

__NOTE__: The `module_n` can only take the names of the respective folders found in `/src`, viz. `color, events, image, io, math, typography, utilities`.

The output file generated can be found under `lib/modules` directory. 

### Testing and Reliability

Tests were initially run on individual module builds, and the failed cases were resolved in a foolproof way.  As those build saw predefined combinations of modules in the build, which passed their respective tests eventually, it is supposed to maintain the same reliability.

### Pending Tasks

* Uglifying the generated bundle, and serving for production use.

src/core/transform.js
```diff
@@ -225,10 +225,13 @@ p5.prototype.scale = function() {
   for(var i = 0; i < args.length; i++) {
     args[i] = arguments[i];
   }
-  if(args[0] instanceof p5.Vector){
-    x = args[0].x;
-    y = args[0].y;
-    z = args[0].z;
+  // Only check for Vector argument type if Vector is available
+  if (typeof p5.Vector !== 'undefined') {
+    if(args[0] instanceof p5.Vector){
```
lmccart: let's put this together into one if:
`if (typeof p5.Vector !== 'undefined' && args[0] instanceof p5.Vector) {`

test/complete-build-tests/test-reference.html
```diff
@@ -11,18 +11,24 @@
...
+  <script src="../js/chai.js" type="text/javascript" ></script>
+  <script src="../js/bind.js" type="text/javascript" ></script>
+  <script src="../js/sinon.js" type="text/javascript" ></script>
+  <script src="../js/modernizr.js"></script>
+
   <script src="js/chai.js" type="text/javascript" ></script>
   <script src="js/bind.js" type="text/javascript" ></script>
   <script src="js/sinon.js" type="text/javascript" ></script>
   <script src="js/modernizr.js"></script>
+
```
lmccart: looks like there are duplicate links here? I think the second set should be removed.

test/unit/image/loading.js
```diff
@@ -7,7 +7,7 @@ var testImageRender = function(file, sketch, callback) {
   var p = sketch.pixels;
   var ctx = sketch;
 
-  sketch.clear();
+  sketch.drawingContext.clearRect(0,0,this.width, this.height);
```
lmccart: this shouldn't be necessary because the color/ files will always be packaged with core

test/unit/typography/font_loading.js
```diff
@@ -59,7 +59,7 @@ suite('Fonts outside preload()', function () {
 
     test('should call success-callback with object when font loads',
       function (done) {
-        p5Inst.loadFont('manual-test-examples/p5.Font/acmesa.ttf',
+        p5Inst.loadFont('../manual-test-examples/p5.Font/acmesa.ttf',
```
lmccart: I feel a little bit weird about all these links that go up one level `../`. I wonder if there is a way to set a working directory to run the tests from that is the `test/` folder? if this can't be done easily, this isn't a big point, just wanted to ask.

lmccart: sorry, I wasn't clear. I was asking about the idea of setting a working directory for all unit tests to be run from. the solution in the changes you pushed recently only sets the directory for this particular file, but leaves in the "../" in other files still.

sakshamsaxena: I think we can put these back on where they were before, since the modules' test folder is removed now. What say?

lmccart: can you explain a bit more about what you mean?

sakshamsaxena: we can do away with the `complete-tests-build` directory, and keep the HTML in the root of test/ like it was before

sakshamsaxena: now that the individual module tests are removed especially

lmccart: ohh I see, yes sounds good!

sakshamsaxena: Have a look now @lmccart

lmccart:
!!

PR #2053, opened 2017-07-09
noFill() for WebGL

mlarghydracept:
This is a different approach to `noFill()` in WebGL. At first I was using barycentric coords but the results were less than ideal. You can see here with [this PR](https://github.com/processing/p5.js/pull/2050) that I closed. This just draws with `gl.LINES` rather than `gl.TRIANGLES` when in `noFill()` state. It also does some of the improvements from the other PR - applying stroke color to wireframe, only writing to attributes that are available in the current shader, and some fixes to a few of the manual tests for webgl.


mlarghydracept:
Also note that the large line addition stat is because of the teapot.obj file included with the manual test. The file is relatively small but github is reading it as text. I didn't actually add 5,000 lines to p5 😅

kadamwhite:
Shouldn't be strictly necessary but we could use [a `.gitattributes` file](https://git-scm.com/docs/gitattributes) to make Git/GitHub treat the .obj as a binary file, if desired

PR #2054, opened 2017-07-10
Fix 1678

kjhollen:
Fix for missing default stroke and fill color, also requiring some updates to the manual tests.

I have not yet checked the examples. I suspect there is more work to do there! But wanted to get some eyes on this to see if it makes sense to call these methods in the constructor and to get an opinion on the updated behavior.

Before, it looked like normalMaterial was the default rendering for retain mode shapes. As implemented, this is now overridden, but consistent with Processing. I'm not sure what the right way is for our context, so curious to hear what others think.

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -53,8 +53,8 @@ p5.RendererGL = function(elt, pInst, isMainCanvas) {
   //default drawing is done in Retained Mode
   this.isImmediateDrawing = false;
   this.immediateMode = {};
-  this.curFillColor = [0.5,0.5,0.5,1.0];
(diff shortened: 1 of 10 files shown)


mlarghydracept:
This looks good to me!

Though it maybe does highlight how we should probably be proactive about working from the same branch and merging asap because I did similar changes to the examples in the PR I have open right now for the WebGL_GSOC2017 branch and we may get conflicts when we try to merge both of these.  Also I think I used the value for RendererGL.curStrokeColor at one point. We can figure it out though!

kjhollen:
Great. I just cherrypicked these commits onto the gsoc branch. Will close this an send an updated pull request.

Issue #2042, opened 2017-07-05
Remove parentheses from YUIDoc params?


kaganjd:
#### Nature of issue?

- [X] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Events
- [X] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

Not sure if this is technically a bug but something about having parentheses inside the YUIDoc params isn't working with the script I'm using to generate the Tern.js autocomplete/code hints file. Would it break anything if we change the following things?

1.  **src/webgl/loading.js**
[Here](https://github.com/processing/p5.js/blob/5ad15eecc37fec2ca2fb6559acc4309f70b32363/src/webgl/loading.js#L28-L31) and [here](https://github.com/processing/p5.js/blob/5ad15eecc37fec2ca2fb6559acc4309f70b32363/src/webgl/loading.js#L38-L39), either make both params`{Function}` **or** `{p5.Geometry}` and `{Event}` **instead of** `{Function(p5. Geometry)}` and `{Function(Event)}`

2.  **src/image/loading_displaying.js**
[Here](https://github.com/processing/p5.js/blob/6ff044d351d662c90a103da9c93d7b5e4037917e/src/image/loading_displaying.js#L32-L35), either make both params `{Function}` **or** `{p5.Image}` and `{Event}` **instead of** `{Function(p5.Image)}` and `{Function(Event)}` 

3.  **lib/addons/p5.sound.js**
[Here](https://github.com/processing/p5.js/blob/004ba93e2ecfe4f7c5c7323eac6c07b8a801cef1/lib/addons/p5.sound.js#L8432), remove parentheses from param description (`part(s)`)

lmccart:
yes that would be totally fine, and an improvement

kaganjd:
I hadn't pulled the latest in a while, and I believe 1 and 3 above were fixed with #2037. 

2 is still causing a problem though. I'm wondering if it's a YUIDoc type problem like the others, not a punctuation problem. I'll look at this today and update.

kaganjd:
[...]

PR #2056, opened 2017-07-10
more yuidoc fixes

Spongman:
here's some more yuidoc fixes.

there's still some weirdness in the docs when different overloads of a method have different values for the @static, @chainable and @return attributes. currently the docs just render as if the first overload was the only one, which looks a little confusing when they're different.

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -403,7 +403,7 @@ var colorPatterns = {
  * For a number of different inputs, returns a color formatted as [r, g, b, a]
  * arrays, with each component normalized between 0 and 1.
  *
- * @param {Array-like} args An 'array-like' object that represents a list of
+ * @param {Array|Object} args An 'array-like' object that represents a list of
  *                          arguments
  * @return {Array}          a color formatted as [r, g, b, a]
  *                          Example:
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -409,9 +409,9 @@ var colorPatterns = {
  * For a number of different inputs, returns a color formatted as [r, g, b, a]
  * arrays, with each component normalized between 0 and 1.
  *
- * @param {Array|Object} args An 'array-like' object that represents a list of
+ * @param {Array} [...args] An 'array-like' object that represents a list of
  *                          arguments
- * @return {Array}          a color formatted as [r, g, b, a]
+ * @return {Number[]}       a color formatted as [r, g, b, a]
  *                          Example:
  *                          input        ==> output
  *                          g            ==> [g, g, g, 255]
--- a/src/core/attributes.js
+++ b/src/core/attributes.js
@@ -36,7 +36,7 @@ var constants = require('./constants');
  *
  * @method ellipseMode
  * @param  {Constant} mode either CENTER, RADIUS, CORNER, or CORNERS
- * @return {p5}                   the p5 object
+ * @chainable
  * @example
  * <div>
  * <code>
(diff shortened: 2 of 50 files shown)


Spongman:
also, i created [a PR](https://github.com/processing/p5.js-sound/pull/199) for the sound stuff.

src/color/p5.Color.js
```diff
@@ -403,7 +403,7 @@ var colorPatterns = {
  * For a number of different inputs, returns a color formatted as [r, g, b, a]
  * arrays, with each component normalized between 0 and 1.
  *
- * @param {Array-like} args An 'array-like' object that represents a list of
+ * @param {Array|Object} args An 'array-like' object that represents a list of
```
lmccart: maybe this should just say {Array}, as it is now it gives the idea it might give one or the other

src/core/attributes.js
```diff
@@ -242,7 +242,7 @@ p5.prototype.strokeCap = function(cap) {
  *
  * @method strokeJoin
  * @param  {Constant} join either MITER, BEVEL, ROUND
- * @return {p5}                   the p5 object
+ * @chainable
```
lmccart: this change is great, much more clear

src/image/filters.js
```diff
@@ -140,10 +140,9 @@ Filters._createImageData = function (width, height) {
  * use case of creating a filter chain that executes without actually updating
  * the canvas in between everystep.
  *
- * @param  {[type]} func   [description]
- * @param  {[type]} canvas [description]
- * @param  {[type]} level  [description]
- * @return {[type]}        [description]
+ * @param  {HTMLCanvasElement} canvas [description]
+ * @param  {function(ImageData,Object)} func   [description]
```
lmccart: can we simplify and just say {function}? see issue #2042

src/color/p5.Color.js
lmccart: oh I see further down it says Number[] for some other color methods, maybe we go with that then?

src/core/attributes.js
Spongman: indeed, and it gives the opportunity later to customize the render specifically for this case.

src/image/filters.js
Spongman: i would recommend keeping the metadata for as long as possible, some tools may be able to use it, others may need to be fixed...

Spongman:
I used `Number[]` where it's expected that the array be an array of numbers, and i kept `Array` where there's no constraints on the element types. i think it's worth keeping the types in there if possible so it's clear that's what the method expects.

src/color/p5.Color.js
Spongman: yes, this one confused me for a while, and i admit i just gave up and said 'hey, object'. but i think you're right, it always returns a `Number[]` ;-)

Spongman:
oh. it looks like i was mistaken. i didn't realize there was a constructor for color that takes an array containing a single string: `color(["#ff0000"])`

lmccart:
@Spongman hmm I don't think that this constructor should exist.. can we remove documentation for it if it does? if the user is entering a css string it should just be a single string, not an array of strings.

Spongman:
yeah, there isn't any documentation for it. i think it's just a consequence of the way `p5.prototype.color()` works:

````
} else if (arguments[0] instanceof Array) {
    if (this instanceof p5.Renderer) {
      return new p5.Color(this, arguments[0]);
    } else {
      return new p5.Color(this._renderer, arguments[0]);
    }
  } else {
````

i'm guessing the intention was to allow `color([r,g,b])` to be equivalent to `color(r,g,b)`, but it also means that _any_ of the other overrides can be invoked by wrapping the arguments in an array.

lmccart:
ah ok well if it just works, I think that's ok. and we can keep it undocumented to remove confusion.

Spongman:
ok, sounds good. if we're ignoring that special case then i think it's valid for `alpha()`, `hue()` et al. to just take `{p5.Color|Number[]}`

Issue #1424, opened 2016-05-16
Texture UV coordinates?


harveymoon:
Processing has the ability to choose UV coords for vertex Pshapes.
`vertex(x, y, u, v)`

How can this work with p5js? Is this possible?

indefinit:
Right now vertex() doesn't support u,v's but this is a good feature request, especially for the webgl renderer.  I'm doing some work on beginShape and endShape now, so this should fit nicely into that upcoming PR.  Use case would look like the example here:  https://processing.org/reference/vertex_.html

We'd also need to implement a textureMode() function for mapping u,v's.

JeroenDerks:
This feature would be so great.
Does anyone have any updates on the matter?

dhowe:
big +1 for this

cqx931:
+1 Need this feature in p5.js for a project.

dhowe:
any progress here?

kjhollen:
added in #2166!

PR #2060, opened 2017-07-12
texture coordinates for immediate mode (implements #1424)

kjhollen:
basic implementation but no error handling yet.

- add test for using texture() in immediateMode
- add performance test for mixing and matching fill and texture modes
- add test for mixing between vertex, shapes, and retain mode.

brings us almost up to parity with Processing, except that we don't have a one-argument version that takes an array. using "apply" is new to me, I hope I've done this right.

I think this is done enough that it makes sense to merge to our working branch.

still needs documentation update and to check examples.


mlarghydracept:
This looks great!

re "apply" I don't know that it is necessary. I think the `this._renderer(x,y,moveTo,u,v)` will work the same. I think "apply" is useful when you need to specify which "this" is to be used on the other end but the RendererGL is already the correct "this" and there isn't any switching into a new renderer that I can see. The arguments array will work on the other end without the apply too I think. Not sure that this matters or that I am even correct.

kjhollen:
What I noticed using the raw call to `this._renderer(x,y,moveTo,u,v)` was that it wasn't possible in the 3D renderer to determine if the user had passed in an argument that was undefined vs. didn't supply the argument in the first place (there were always 5; any left out would just be undefined). I think it's nice to know if the user is calling the 2-arg version and leaving out the z coordinate vs. trying to use the 3-arg version and accidentally supplying a bogus z value. Maybe a good thing to do is to check with Alice about what's useful, error message-wise?

After I submitted this, I also got to thinking about where this error handling should even happen. Something I noticed in the Processing code base was that a lot of functions had something like _vertexImpl where the arguments have already been pre-processed elsewhere, so the actual execution of graphics stuff can just happen (many of these functions have like 8-10 arguments for various gl or graphics params). I'm not sure yet if I think the checking belongs in the specific renderer or at a higher level—it would be nice not to duplicate code for parameter validation across renderers!

Issue #2069, opened 2017-07-16
p5.Font.textToPoints not working for space


Niels-NTG:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [x] Typography
- [ ] Utilities
- [ ] WebGL

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 0.5.11
- Web browser and it's version: Google Chrome 59.0.3071.115 and Apple Safari 10.1.1
- Operating System: Mac OS 10.12.5
- Steps to reproduce this: Load a `.ttf` file with the `loadFont` method. Then call the `textToPoints` method with this font using a string that contains at least 1 whitespace character.

```javascript
var font;

function preload() {
  font = loadFont('NotoSans-Bold.ttf');
}

function setup() {
  createCanvas(windowWidth, windowHeight);

  font.textToPoints('Hello world!', 0, 0);
}
```

This will generate the following error in Chrome:
```
Uncaught TypeError: Cannot read property '0' of undefined
```
And Safari, the same error, worded more clearly:
```
TypeError: undefined is not an object (evaluating 'pathArray[0]')
```

After some simple debugging it appears that variable `pathArray` in the function `pathToAbsolute` is `undefined` when the character being passed is a whitespace. This only happens for the a handful of `.ttf` font files I tested. Not for any of the `.oft` files I've used.

I've written a fix that prevents the error while still rendering the whitespace. Specifically, it returns the array `res` straight away if `pathArray` is `undefined`, not executing the rest of the function.

Niels-NTG:
This issue seem to have some similarities with #1788. Perhaps that solution can be reused?

lmccart:
fixed with https://github.com/processing/p5.js/pull/2108

PR #2070, opened 2017-07-16
Fix for textToPoints for text with whitespaces (#2069)

Niels-NTG:
Fix for issue #2069 

Return res straight away if pathArray is undefined.

This solution fixes the issue, although I could see that the underlying cause is found somewhere in a place I couldn't find in my stack traces. So people with an extensive knowledge of `p5.Font.js`, please weigh in if I'm the right track.


lmccart:
oops, sorry, I missed this pull request! I think we've gotten to the root cause now, thanks.

PR #2074, opened 2017-07-17
Load pixels webGL

### LINKED ISSUE #1712: loadPixels() does not work with PGraphics in WEBGL mode
When I call a pg.loadPixels() function, on a PGraphics object created on WEBGL mode, such as this:

function setup() {
  createCanvas(500,500)
  pg = createGraphics(500,500,WEBGL)
}

function draw() {
  pg.loadPixels()
}

I get this error:

20090: Uncaught TypeError: undefined is not a function

When using the P2D renderer the loadPixels() / updatePixels() functions work smoothly

mlarghydracept:
closes #1712 

This PR adds loadPixels and the pixels[] for WebGL. In order to get the pixel data from gl preserveDrawingBuffer has to be true. It shouldn't be true by default for performance reasons on mobile. At first I had loadPixels check to see if preserveDrawingBuffer was already true and if it wasn't reset the context. There were potentially solvable but difficult timing issues with this approach though. Now instead I implemented `setAttributes()` (which was already another todo).

So with this PR, the user is required to put `setAttributes('preserveDrawingBuffer', true)` or `setAttributes({preserveDrawingBuffer: true})` in `setup()`. You can see an example in the manual test folder. I know this may be less than ideal thus the WIP (also I need to add commenting/documentation). So if anyone has opinions or suggestions about how this can work I would be super grateful for the feedback.


mlarghydracept:
I adjusted everything so that loadPixels works by default. Now preserveDrawingBuffer defaults to true. There currently isn't an updatePixels or set so changes can only happen through direct manipulation of the pixels[]. Still even as is, get and loadPixels provide a lot of uses.

test/manual-test-examples/webgl/pixels/index.html
```diff
@@ -0,0 +1,21 @@
...
+  <title></title>
+  <link rel="stylesheet" href="">
+  <script language="javascript" type="text/javascript" src="../../../../lib/p5.js"></script>
+  <script language="javascript" type="text/javascript" src="sketch.js"></script>
+  <style>
+    html, body {margin:0; padding:0;}
+  </style> 
+</head>
+<body>
+<h2 style='margin-left:100px'>If you click the canvas, the box should get textured with an image of the canvas</h2>
```
sansumbrella: The textured cube is pretty subtle when I view it. It becomes just a tiny speck. Maybe changing the background color at the same time would help. Or using the copied pixels as screengrabs that tile the page.

screen shot 2017-07-20 at 12 04 57 pm
screen shot 2017-07-20 at 12 05 02 pm

src/webgl/p5.RendererGL.js
```diff
@@ -83,6 +81,93 @@ p5.RendererGL.prototype._initContext = function() {
...
+    console.log('setAttributes() only works in WebGL');
+    return;
+  }
+  var attr = {};
+  if(arguments.length === 2) {
+    attr[arguments[0]] = arguments[1];
+  }
+  else if (arguments.length === 1) {
+    attr = arguments[0];
+  } else {
```
lmccart: this case should eventually be handled by the friendly errors, so we don't need to build it into the logic. @almchung can we decide on a comment tag that is easily searchable to mark places that should be revisited when friendly error system is working? maybe something like "@TODO FES"?

src/webgl/p5.RendererGL.js
```diff
@@ -83,6 +81,93 @@ p5.RendererGL.prototype._initContext = function() {
...
+    setTimeout(function() {
+      callback.apply(window._renderer, options);
+    }, 0);
+  }
+};
+
+
+/**
+ *
+ * Set attributes for the WebGL Drawing context.
```
lmccart: let's add one more line to give a little more sense for what this is. something like "This is a way of adjusting ways that the WebGL renderer works to fine-tune the display and performance."

src/webgl/p5.RendererGL.js
```diff
@@ -83,6 +81,93 @@ p5.RendererGL.prototype._initContext = function() {
...
+ *
+ * preserveDrawingBuffer - if true the buffers will not be cleared and
+ * and will preserve their values until cleared or overwritten by author
+ * (note that p5 clears automatically on draw loop)
+ * default is true
+ *
+ * @method setAttributes
+ * @param  {String|Object}  String name of attribute or object with key-value pairs
+ * @param  {Boolean}        New value of named attribute
+ *
```
lmccart: can we add an example or two? you can use <div class='norender'> if the visual result doesn't display anything useful. i've found it can be helpful for people just to see the code written in context. we could add a line of comment after the call in the example explaining what difference should be noticeable in a running sketch.

src/webgl/p5.RendererGL.js
```diff
@@ -375,6 +460,63 @@ p5.RendererGL.prototype._setNoFillStroke = function() {
...
+ * is returned.
+ *
+ * Returns undefined if the region is outside the bounds of the image
+ *
+ * @method get
+ * @param  {Number}               [x] x-coordinate of the pixel
+ * @param  {Number}               [y] y-coordinate of the pixel
+ * @param  {Number}               [w] width
+ * @param  {Number}               [h] height
+ * @return {Array|Color|p5.Image}     color of pixel at x,y in array format
```
lmccart: agree, we recently discussed this with the documentation for get() in 2D mode. you could take a look at that entry for guidance.

test/manual-test-examples/webgl/pixels/sketch.js
```diff
@@ -0,0 +1,30 @@
...
+  {
+    texture(c);
+  }
+  box(75);
+  pop();
+}
+
+function mousePressed() {
+  c = get();
+}
```
lmccart: can we add a few unit tests that aren't manual, too? these automated unit tests are the ones that get run with every grunt build so it's easier to catch things that break with them.

test/manual-test-examples/webgl/pixels/index.html
mlarghydracept: Oh strange! When I saw this I was a bit confused as the `pointLight()` call alone should keep the box from looking like this and in Chrome the example looks like this.

screen shot 2017-07-20 at 9 31 34 am
screen shot 2017-07-20 at 9 35 29 am

When I saw yours I realized I forgot to check firefox. Not sure if this has something to do with pointLight. I'll investigate more.

mlarghydracept: Okay after more investigation. I tried several different approaches and this behavior seems to be the same across this branch and master in all scenarios. I got pixels from a 2d canvas and made them the texture in master and got the same barely present effect. This seems specific to pointLight and has something to do with how it is interacting with a p5.Image returned from get(). I will keep investigating but seems unrelated to this PR. I took the pointLight out of the example so that expected output is the same across browsers.

src/webgl/p5.RendererGL.js
almchung: @lmccart Sounds great. I guess `@todo_FES` works?

mlarghydracept:
Thanks for the help! I handled almost everything you mentioned and any exceptions I explained in comments on your comments. I think this should be ready to merge.

src/webgl/p5.RendererGL.js
```diff
@@ -83,6 +81,93 @@ p5.RendererGL.prototype._initContext = function() {
...
+  var c = document.getElementById(defaultId);
+  if(c){
+    c.parentNode.removeChild(c);
+  }
+  c = document.createElement('canvas');
+  c.id = defaultId;
+  if (this._pInst._userNode) {
+    this._pInst._userNode.appendChild(c);
+  } else {
+    document.body.appendChild(c);
```
mlarghydracept: That is my understanding but maybe I could be missing some edge case. This is how the rest of p5 handles it so I made assumptions.

src/webgl/p5.RendererGL.js
```diff
@@ -375,6 +460,63 @@ p5.RendererGL.prototype._setNoFillStroke = function() {
...
+  }
+  var pd = this._pInst._pixelDensity;
+  var x = arguments[0] || 0;
+  var y = arguments[1] || 0;
+  var w = arguments[2] || this.width;
+  var h = arguments[3] || this.height;
+  w *= pd;
+  h *= pd;
+  var gl = this.GL;
+  var pixels = new Uint8Array(gl.drawingBufferWidth * gl.drawingBufferHeight * 4);
```
mlarghydracept: I didn't because the 2D loadPixels and get doesn't as far as I can tell.

src/webgl/p5.RendererGL.js
mlarghydracept: I added two small examples as a sort of before and after thing. Hopefully this is okay.

src/webgl/p5.RendererGL.js
```diff
@@ -26,10 +15,20 @@ var attributes = {
...
  */
-p5.RendererGL = function(elt, pInst, isMainCanvas) {
+p5.RendererGL = function(elt, pInst, isMainCanvas, attr) {
   p5.Renderer.call(this, elt, pInst, isMainCanvas);
+  this.attributes = {};
+  attr = attr || {};
+  this.attributes.alpha = attr.alpha === undefined ? true : attr.alpha;
+  this.attributes.depth = attr.depth === undefined ? true : attr.depth;
+  this.attributes.stencil = attr.stencil === undefined ? true : attr.stencil;
+  this.attributes.antialias = attr.antialias ===
```
mlarghydracept: I agree but p5 has been strict line length rules for building and even if it is split at the `?` the last couple of attributes are still too long to build. I am still splitting at the `===` but I changed it so even the shorter attributes do this for consistency.

PR #2078, opened 2017-07-18
validateParameters() logic update and 2d_primitives test units

almchung:
Updated flow in `validateParameters()` so we can exit the function whenever we find a matching parameter format. 

`validateParameters()` is also inserted into all 2d_primitives functions. All their unit tests are successfully running, but there are some minor false negative cases escaping the validation logic (only occurs for functions with multiple parameter formats).

Some of the remaining questions are:

1. Right now  `validateParameters()` is called inside the individual functions. Should we move it around so it can be turned on/off by setting a global parameter? Such as : https://github.com/processing/p5.js/issues/1329.

2. Not sure how our error msgs will be shown within the web editor.

Thank you!

### DIFF
--- a/test/unit/core/2d_primitives.js
+++ b/test/unit/core/2d_primitives.js
@@ -16,6 +16,24 @@ suite('2D Primitives', function() {
         assert.ok(arc);
         assert.typeOf(arc, 'function');
       });
+      test('arc(): no friendly-err-msg', function() {
+        assert.doesNotThrow(function() {
+            myp5.arc(1, 1, 10.5, 10, 0, Math.PI, 'pie');
+          },
+          Error, 'got unwanted exception');
+      });
+      test('arc(): missing param #4, #5', function() {
+        assert.doesNotThrow(function() {
+            myp5.arc(1, 1, 10.5, 10);
+          },
+          Error, 'got unwanted exception');
+      });
+      test('arc(): wrong param type at #0', function() {
+        assert.doesNotThrow(function() {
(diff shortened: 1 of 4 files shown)


lmccart:
cc @catarak see question 2 above. two thoughts:
* By default the messages would just print to console, which I think is fine for now, but we may want to consider a special handling or treatment in the future.
* The alpha editor links to p5.min.js. our plan has been to use a grunt setup that only includes the friendly error system in the p5.js version, and deactivates it in p5.min.js for performance reasons. I wonder then if we want some kind of settings option that would actually let you switch which version of the library the index links to.

catarak:
If the messages just print to the JS console then they would appear in the web editor console as well.

Maybe there could be a preference option in the web editor that turns the friendly errors on or off? And they could be on by default?

lmccart:
oh! we just realized for the settings... you don't need to change which version is linked to, you just need to set `p5.disableFriendlyErrors = true;`

test/unit/core/2d_primitives.js
```diff
@@ -56,6 +99,106 @@ suite('2D Primitives', function() {
...
+          },
+          Error, 'got unwanted exception');
+      });
+      test('line(): missing param #3', function() {
+        assert.doesNotThrow(function() {
+            myp5.line(0, 0, Math.PI);
+          },
+          Error, 'got unwanted exception');
+      });
+      test('line(): missing param #4', function() { // this err case escapes
```
almchung: false negative case

lmccart:
per our conversation, I think this is good to merge

PR #2087, opened 2017-07-21
fix #1678 by setting default stroke and fill colors in 3d renderer.

### LINKED ISSUE #1678: beginShape should not require call to fill() in order to render
I have a minimal example set up that uses the webgl renderer 
``` 
function setup(){
}

function draw(){
    beginShape(POINTS);
    vertex(0,0);
    endShape();
} 
```

Running this produces the following errors:
```
p5.js:31516 WebGL: INVALID_VALUE: enableVertexAttribArray: index out of rangep5.RendererGL._bindImmediateBuffers @ p5.js:31516p5.RendererGL.endShape @ p5.js:31447p5.endShape @ p5.js:15190draw @ sketch.js:8p5.redraw @ p5.js:14256(anonymous function) @ p5.js:9143
p5.js:31520 WebGL: INVALID_VALUE: vertexAttribPointer: index out of range
```
I'm using the latest version of p5.js (p5.js v0.5.4 October 01, 2016)

kjhollen:
updated this request to reflect lots of new changes. Some of this is formatting; let me know if you want me to save that for another commit!

uses this.stroke and this.fill methods when the renderer is initialized,
so that all necessary gl context side effects also happen.

clean up gl tests:

- made the camera tests use normalMaterial(), instead of default
  stroke and fill colors.
- some tests were using basicMaterial(), which no longer exists. Changed
  to fill() or normalMaterial() as appropriate.
- remove trailing whitespaces in tests.

update conditional in _getCurShader to reflect that interleaving
immediate mode and retain mode calls requires changing shaders,
even if no change is made to fill color.

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -53,8 +53,8 @@ p5.RendererGL = function(elt, pInst, isMainCanvas) {
   //default drawing is done in Retained Mode
   this.isImmediateDrawing = false;
   this.immediateMode = {};
-  this.curFillColor = [0.5,0.5,0.5,1.0];
-  this.curStrokeColor = [0.5,0.5,0.5,1.0];
+  this.fill(255, 255, 255, 255);
+  this.stroke(0, 0, 0, 255);
   this.pointSize = 5.0;//default point/stroke
 
   this.emptyTexture = null;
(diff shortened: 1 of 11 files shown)


kjhollen:
Just kidding, I broke all the examples. 🙃🙃🙃🙃  Hold please...

PR #2090, opened 2017-07-22
Shader refactor

kjhollen:
I took a pass at refactoring shader handling a bit to create a new p5.Shader class. The goal here would be to narrow down the possible paths that the code might take when setting up the gl context to render: everything is shader-centric, so it feels right to me to have a Shader class that manages uniforms and attributes, preparing the gl context to render, and other shader related tasks. Ideally, this would also make it more obvious where to do matrix handling so that setting up a more user-friendly camera API (an eventual goal) would be easier to understand for contributors.

this is built on the fill/stroke patch I have open.

I feel like this is maybe 60% complete (the commit log notes todo items), but because I moved so much stuff I wanted to share this as soon as it seemed workable. all of the examples are still working. Looking forward to your feedback!


src/webgl/shader.js
```diff
@@ -16,5 +19,286 @@ module.exports = {
...
+    gl.linkProgram(this.glProgram);
+    if (!gl.getProgramParameter(this.glProgram, gl.LINK_STATUS)) {
+      alert('Snap! Error linking shader program');
+    }
+
+    this.loadAttributes();
+    this.loadUniforms();
+
+    // TODO move elsewhere?
+    this._renderer._createEmptyTexture();
```
mlarghydracept: I agree that it is a little weird to have here but it has to happen once the first time a shader is initialized. The logic inside of the function itself should ensure that it only does anything once. There might be a better place for it but I am not sure where.

DoubleJump:
I think the createEmptyTexture call could go in the init function for the gl renderer as you always need it and multiple shaders potentially would reference it? 

On shaders, and this is just personal opinion, but I always write my vert and frag shaders together and rely on preprocessors for code reuse. This means you can load shaders as single units and store them in an object using the file name as the key which is easy to remember.

Agreed that the shader code does look a lot cleaner now :)

src/webgl/shader.js
kjhollen: yeah! it wasn't immediately obvious to me where I might move it to, but it felt weird to put it here in the shader class when it's dealing with a private method of the renderer. just left this note here to come back to it later.

kjhollen:
@mlarghydracept yeah, I think the constructor for this still isn't quite right—like it should take in the path for loading the shader files instead of the constants for our pre-defined ones (one of the goals I had in mind for this was to not necessarily to implement custom shaders but to leave room for someone to do it easily in the future). I think the code in the renderer could also be more legible if it said things like shader = p5.Shader.getWireframeShader();  or p5.Shader.getNormalShader(); (and the method can then make the shader if it hasn't been used yet). if it's not expensive to make all of the shaders at start-up, I suppose we could always do that instead, and then the built-in shaders could read more like constants, e.g. p5.Shader.WIREFRAME_SHADER. Maybe a good thing for me to do next is to pick one of these directions & pursue it.

@DoubleJump thanks :) yes I think there will be a little bit more mix and match with the various shader types as we get into building out more of the functionality. I think we'll learn more as Stalgia continues to work on shaders for handle stroke + fill, too.

src/webgl/shader.js
```diff
@@ -16,5 +19,286 @@ module.exports = {
...
+  gl.useProgram(this.glProgram);
+};
+
+/**
+ * Wrapper around gl.uniform functions.
+ * As we store uniform info in the shader we can use that
+ * to do type checking on the supplied data and call
+ * the appropriate function.
+ */
+p5.Shader.prototype.setUniform = function(uniformName, data)
```
lmccart: is this a public or private method (and the others)?
maybe we should come up with some clear conventions about using _ for private things or not, it seems to be a total mix throughout the library right now :)

src/webgl/shader.js
lmccart: maybe for now a comment mentioning this only happens once, so a developer doesn't have to look up the _createEmptyTexture() definition to confirm that.

kjhollen: good suggestion. I am comfortable with that for now!

src/webgl/shader.js
kjhollen: I think this one is public (thinking ahead to when someone may write their own shaders, this is how they'd supply a value for a custom uniform or attribute)—but yes, in general, there needs to be some clean-up for what's public and private!

after talking to Stalgia a little bit today, I think my next step is to define a fixed set of default shaders so we can get rid of the material hash. I think that will clear up a lot of the remaining ambiguity about public/private too—and then I'll write some unit tests for each of the default shaders. :)

lmccart: cool and yeah, if they're public we should add the inline docs and @method property before we merge so they show up in the reference.

kjhollen: cool. working on this and some other clean up now. from what I can tell, the gl tests aren't currently being run because window.Modernizr.webgl is missing. I'm out of my js depth here, any ideas?

mlarghydracept: Yeah I hit that when making the tests for loadPixels(). From what I can tell it isn't possible to run those WebGL unit tests from the cmd line build process (I could very well be wrong!) but if you want to run those tests you can run the browser version of the build tests with the included html file. That is how I ran my tests.

kjhollen: Ohhhhhhhhh I get it! I'd assumed the test.html file was for running things behind the scenes for some reason and hadn't even tried to open it. Got it working now, thanks @mlarghydracept (and @lmccart you can disregard the email I sent earlier ;))

lmccart: oh I just saw this, @mlarghydracept answered it even better than my email, I forgot about the possibility to open the browser version. coool!

kjhollen:
Ok @lmccart I think this is ready if you do! I just pushed the unit tests (and have some thoughts on public/private in the last commit message there).

src/webgl/p5.Shader.js
```diff
@@ -0,0 +1,330 @@
...
+
+  return this;
+};
+
+/**
+ * Creates, compiles, and links the shader based on its
+ * sources for the vertex and fragment shaders (provided
+ * to the constructor). Populates known attributes and
+ * uniforms from the shader.
+ * @method init
```
lmccart: can you add @chainable for the ones that are?
we've recently added support for it in yuidocs build

src/webgl/p5.Shader.js
```diff
@@ -0,0 +1,330 @@
...
+    // 1. creating and getting a gl id for the shader program,
+    // 2. compliling its vertex & fragment sources,
+    // 3. linking the vertex and fragment shaders
+    this._vertShader = gl.createShader(gl.VERTEX_SHADER);
+    //load in our default vertex shader
+    gl.shaderSource(this._vertShader, this._vertSrc);
+    gl.compileShader(this._vertShader);
+    // if our vertex shader failed compilation?
+    if (!gl.getShaderParameter(this._vertShader, gl.COMPILE_STATUS)) {
+      alert('Yikes! An error occurred compiling the vertex shader:' +
```
lmccart: wondering about this use of alert() here. I don't think we're using it elsewhere in the library for errors. would a friendly error suffice here? or is this one critical enough that we need to break that pattern?

src/webgl/p5.Shader.js
```diff
@@ -0,0 +1,330 @@
...
+    gl.attachShader(this.glProgram, this._fragShader);
+    gl.linkProgram(this.glProgram);
+    if (!gl.getProgramParameter(this.glProgram, gl.LINK_STATUS)) {
+      alert('Snap! Error linking shader program');
+    }
+
+    this._loadAttributes();
+    this._loadUniforms();
+
+    // TODO move elsewhere?
```
lmccart: can you add that comment explaining this only happens the first time

src/webgl/p5.Shader.js
```diff
@@ -0,0 +1,330 @@
...
+
+  switch(uniform.type){
+    case gl.BOOL:{
+      if(data === true) {
+        gl.uniform1i(location, 1);
+      }
+      else {
+        gl.uniform1i(location, 0);
+      }
+      return;
```
lmccart: should this be break; ?

src/webgl/p5.Shader.js
```diff
@@ -0,0 +1,330 @@
...
+  // TODO BIND?
+  var uniform = this.uniforms[uniformName];
+  if(!uniform) {
+    //@todo warning?
+    return;
+  }
+  var location = uniform.location;
+
+  switch(uniform.type){
+    case gl.BOOL:{
```
lmccart: these brackets for the cases can be omitted

src/webgl/p5.Shader.js
```diff
@@ -0,0 +1,330 @@
...
+  this._vertSrc = vertSrc;
+  this._fragSrc = fragSrc;
+  this._vertShader = -1;
+  this._fragShader = -1;
+  this.glProgram = 0;
+  this.loadedAttributes = false;
+  this.attributes = {};
+  this.loadedUniforms = false;
+  this.uniforms = {};
+  this.bound = false;
```
lmccart: if these are public they should probably be documented with @property tags. maybe this is coming later? or should we discuss having two types of public methods/properties -- documented and undocumented? (I know this is currently the case in the rest of the library, but not necessarily by design heh, so it would be useful at this point I think to clarify and try to be more strict moving forward. what are your thoughts?)

src/webgl/p5.Shader.js
kjhollen: I think a friendly error is definitely more appropriate.

src/webgl/p5.Shader.js
kjhollen: Yes. My thinking is that some of these are in-between: that they are something more akin to what Java would call protected (where the renderer, and components within the p5 GL pipeline, have access to them and are expected to use them but other modules within the project would not). I don't know if there are JS conventions for this...

kjhollen: just a note here that we went with @private as the tag!

PR #2098, opened 2017-07-28
yuidoc templates hide private stuff, link to extended classes

therewasaguy:
Some changes so that we can utilize yuidoc tags in p5.sound reference pages (and I think they're helpful everywhere?):
- `@private` methods and properties don't show up on the library pages or class page
- child class that `@extends` another class automatically links to that class

This PR updates the Backbone.js templates that are used by the requirejs task to generate a reference.js file for the reference pages. Let me know if I also generate the new reference.js file and send a PR to the website repo?

cc @Spongman @jvntf


lmccart:
awesome

PR #2108, opened 2017-08-06
proper fix to #2069

dhowe:
This should address the underlying issue(s) in #2069 

We may see other strange cases in specific fonts, but we'll have to address these as they arise (this patch should handle the common cases, and at least fail gracefully otherwise)


lmccart:
great, thanks!

PR #2112, opened 2017-08-08
create p5.Texture for all of our texturing needs

kjhollen:
moves most texture related code into p5.Texture.js

shader handles looking up and binding the textures as needed.

note: this change will let us use the same p5.Image in multiple
gl contexts! before this wasn't possible, because the texture
ID was attached directly to the image, and there could only be one ctx.

now the textures are managed by the renderer, albeit by searching
a list, which may be slow in the long run. let's see...

this is still a work in progress. I can't tell if it makes the code easier to understand or not, but I like that it puts most (all?) texturing concerns into a single file with a little bit of context management in the shader and renderer. I think my next thing to do is to test this out with another feature that might require texturing (maybe implementing text? or custom shaders, which might supply multiple images as samplers?). but I thought it might be good to get another set of eyes on this while i figure out more details.


kjhollen:
Thanks! Yeah, detecting updates is still the tough part. `p5.Renderer.getTexture()` will create a new texture if it needs to, then also call `set` on the texture (new or old) with the new image data. Right now `set` does a very crude check to see if the size of the texture has changed for some reason otherwise will just use `texSubImage2D`.

I haven't yet done a real performance comparison to see if using `texSubImage2D` is faster, but I looked at one of the manual tests that swaps between textures & fill and didn't see an obvious improvement (but it may also be the nature of the test itself). I think it would be better if we didn't push the texture at all if no change occurred. For about 2 minutes earlier today I thought I could just add a field to p5.Image but it's more than that because there are also DOM elements to consider. 🙃 I'll keep researching!

DoubleJump:
As far as I know calling `texSubImage2D` won't be faster if you are re-uploading all the pixels. You'll only see a speed up for updating portions of a texture (say an individual glyph in font atlas). Then you would also need a way of passing in the offsets which I imagine would be tricky. 

The sampler/filter stuff you have is pretty much what I had too. I had some preset options that you could pass in to the texture() function but that's basically it.

The common use case of getting an image via `loadImage` then calling `texture()` should definitely be faster by checking if it already exists. I know some openGL libraries (Cinder) cache their entire GL state to avoid calling as few `gl.xx` functions as possible.

I'm drawing a blank on how you would check for an updated texture without looking at all the pixels too. Checking width/height might the best that can be done. Will keep thinking though :)

src/webgl/p5.RendererGL.js
```diff
@@ -631,7 +635,22 @@ p5.RendererGL.prototype._getColorShader = function () {
   return this._defaultColorShader;
 };
 
+p5.RendererGL.prototype.getTexture = function (img) {
+  // efficiency?
+  var f = function(element) {
```
sansumbrella: Curious why you don't use a map here for simplicity.

```js
if (this.textures[img] === undefined) {
  this.textures[img] = new p5.Texture(this, img);
}
return this.textures[img];
```

Performance of finding a single element in the array is good, but will probably degrade as the list grows (though there is a hardware limit to how many textures you could have). Map lookup times for the single element are comparable.

src/webgl/p5.RendererGL.js
```diff
@@ -631,7 +635,22 @@ p5.RendererGL.prototype._getColorShader = function () {
...
+    return element.src === img;
+  };
+
+  var tex = this.textures.find(f);
+  if (tex === undefined) {
+    tex = new p5.Texture(this, img);
+    this.textures.push(tex);
+  }
 
+  tex.set(img);
```
sansumbrella: Do we need to set the image if we are just fetching it?

I think that updating the texture with new pixels should probably be a separate concern from just getting the texture for rendering.

sansumbrella: To me, it makes sense to use `createTexture` or some analogue to it to do any texture creation or updating. That way, we don't concern ourselves with whether the data has changed when setting uniforms for rendering. The uniform setting just points to data that's already there.

src/webgl/p5.RendererGL.js
kjhollen: Ohhh this is just hitting the limits of my experience with Javascript. I'd rather have it behave this way, but thought I couldn't use an object like p5.Image as the map key.

src/webgl/p5.RendererGL.js
kjhollen: Hey, I'm not sure I'm understanding these two comments together.

at present, `getTexture` is only used when setting the uniform, so it will create the texture if it can't find an existing one or or just update it if it finds one. Would it be clearer if the method was called `createOrFindTexture` and `set` was called in the shader?

`p5.Shader.bindTextures` is where the texture is actually fetched/bound for rendering (though not updated) but not with `getTexture`—instead it uses the cached texture object associated with shader's sampler uniforms.

sansumbrella: When logging in the console, I can see that `getTexture()` is called every frame. Since it's a relatively expensive operation to always copy the pixels when we do that, I think we can do something that binds the texture without also copying the pixels.

As a further optimization, once a texture is bound to a given location, the uniform and texture don't need to be rebound unless you want the uniform to point to a different texture or for the texture to change.

src/webgl/p5.RendererGL.js
sansumbrella: Yeah, for Objects I believe it ends up making a hash from the address (which we don't have access to), so it's still a good lookup key.

src/webgl/p5.RendererGL.js
sansumbrella: So I'm just concerned that we're doing this repeatedly in a tight loop. Pretty sure we can limit what happens at draw time to just what's necessary.

kjhollen: Got it. This was already happening in the code before; but a goal for this refactor was to make it easier to understand where an optimization might go. My plan is that `p5.Texture.set` can check the source image to see if any updates are needed, and then only do the texture upload if necessary.

The data itself may come from a p5.Image but can also be from a DOM video or img element (possibly containing an animated gif). I have the optimization working for p5.Image and think I can use the timestamp or check if the video is playing for the video, but I'm stumped on what to do for a DOM img with an animated gif. Any thoughts?

kjhollen:
Ok, I think this is ready for a final review. I've added some documentation and some light unit tests, which helped me uncover a bug (hooray unit tests). @sansumbrella I reverted to searching the array with find instead of using the p5.Images as keys. Javascript matches object keys as strings, which means they're all '[object Object]'!

lmccart:
sorry I was slow to this one, thanks for all the earlier reviewing everyone. looks good to me!

PR #2115, opened 2017-08-09
FES implementation for color, curves, string util functions

almchung:
1. Implemented FES & wrote new unit tests
* `src/color/creating_reading.js`
* `src/core/curves.js`
* `src/utilities/string_functions.js`

2. Reformatted style for these functions:
* `bezierPoint()`, `bezierTangent()`, `curvePoint()`, `curveTangent()`

3. Updated inline documentations
* `src/color/creating_reading.js`
* `src/utilities/string_functions.js`

Remaining issues:

1. We get false positive FES msgs from `nf()`. -> need to update inline doc.
2. `this` sometimes breaks down during unit tests. Temporarily replaced with `p5.prototype`.

### DIFF
--- a/src/color/creating_reading.js
+++ b/src/color/creating_reading.js
@@ -295,9 +295,14 @@ p5.prototype.brightness = function(c) {
  *                                 and alpha components of the color
  * @return {p5.Color}
  */
+/**
+ * @method color
+ * @param  {p5.Color}     color
+ * @return {p5.Color}
+ */
 
 p5.prototype.color = function() {
-  //p5.prototype._validateParameters('color', arguments);
+  //this._validateParameters('color', arguments);
   if (arguments[0] instanceof p5.Color) {
     return arguments[0];  // Do nothing if argument is already a color object.
   } else if (arguments[0] instanceof Array) {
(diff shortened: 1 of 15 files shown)


src/color/creating_reading.js
```diff
@@ -277,19 +272,32 @@ p5.prototype.brightness = function(c) {
...
+ */
+/**
+ * @method color
+ * @param  {Array}      values  an array containing the red,green,blue &
+ *                                 and alpha components of the color
+ * @return {p5.Color}
  */
 
 p5.prototype.color = function() {
+  //p5.prototype._validateParameters('color', arguments);
```
lmccart: should this be here?

src/color/creating_reading.js
```diff
@@ -420,35 +424,36 @@ p5.prototype.hue = function(c) {
  *
  */
 
-p5.prototype.lerpColor = function(c1, c2, amt) {
+p5.prototype.lerpColor = function() {
```
lmccart: if there's just one combination of arguments, we can leave the named arguments in the signature. if there are multiple combos, then we normally leave the parens empty. or were they removed here for another reason?

almchung: Sorry, I think this was mistake while I was cleaning up some stuff. I will double-check for other ones too!

PR #2131, opened 2017-08-20
refactor color unit test to use a simple version of shared examples

outofambit:
Hi! I was looking at the color unit test file in preparation for another task and I did a quick stab at a refactor to shorten (and DRY up) the test some. I took some inspiration from this [mocha wiki page](https://github.com/mochajs/mocha/wiki/Shared-Behaviours) and created a simple version of shared examples that could be used throughout the color test suites.

What do people think about this approach?

This could definitely be more robust (I hardcode the RGBA values in the shared example). It also relies on the use of `this` which I have mixed feelings about. But it does help make the tests a little easier to read and modify (in my opinion). I'm happy to push this farther if people think its valuable!

### DIFF
--- a/test/unit/color/p5.Color.js
+++ b/test/unit/color/p5.Color.js
@@ -5,35 +5,36 @@ suite('p5.Color', function() {
   });
   var c;
 
+  function subject() {
+    return myp5.color(255, 0, 102);
+  }
+
   // shared tests for p5.Color
   // assign something to `this.c` for this to work in a suite
-  function isAP5Color() {
+  function isAP5Color(subject) {
     test('should create instance of p5.Color', function() {
-      assert.instanceOf(this.c, p5.Color);
+      assert.instanceOf(subject(), p5.Color);
     });
   }
-  function behavesLikeABasicColor() {
+  function behavesLikeABasicColor(subject) {
 
-    isAP5Color();
+    isAP5Color(subject);
 
     test('should correctly set RGBA property', function() {
-      assert.deepEqual(this.c.levels, [255, 0, 102, 255]);
+      assert.deepEqual(subject().levels, [255, 0, 102, 255]);
     });
   }
 
   suite('p5.prototype.color(r,g,b)', function() {
-    setup(function() {
-      this.c = c = myp5.color(255, 0, 102);
-    });
 
-    behavesLikeABasicColor();
+    behavesLikeABasicColor(subject);
 
     test('shouldn\'t set HSBA property before hsb access func is called', function() {
-      assert.equal(c.hsba, undefined);
+      assert.equal(subject().hsba, undefined);
     });
 
     test('shouldn\'t set HSLA property before hsb access func is called', function() {
-      assert.equal(c.hsla, undefined);
+      assert.equal(subject().hsla, undefined);
     });
 
     test('color(): missing param #0 + throws error', function() {
@@ -44,11 +45,11 @@ suite('p5.Color', function() {
   });
 
   suite('p5.prototype.color("#rgb")', function() {
-    setup(function() {
-      this.c = c = myp5.color('#f06');
-    });
+    function subject() {
+      return myp5.color('#f06');
+    }
 
-    behavesLikeABasicColor();
+    behavesLikeABasicColor(subject);
 
     suite('spot check', function() {
       test('numeric hex values', function() {
@@ -76,13 +77,15 @@ suite('p5.Color', function() {
   });
 
   suite('p5.prototype.color("#rgba")', function() {
-    setup(function() {
-      this.c = c = myp5.color('#f016');
-    });
-    isAP5Color();
+
+    function subject() {
+      return myp5.color('#f016');
+    }
+
+    isAP5Color(subject);
 
     test('should correctly set RGBA property', function() {
-      assert.deepEqual(c.levels, [255, 0, 17, 102]);
+      assert.deepEqual(subject().levels, [255, 0, 17, 102]);
     });
 
     suite('spot check', function() {
@@ -112,10 +115,10 @@ suite('p5.Color', function() {
 
   suite('p5.prototype.color("#rrggbb")', function() {
     setup(function() {
-      this.c = c = myp5.color('#ff0066');
+      return myp5.color('#ff0066');
     });
 
-    behavesLikeABasicColor();
+    behavesLikeABasicColor(subject);
 
     suite('spot check', function() {
       test('numeric hex values', function() {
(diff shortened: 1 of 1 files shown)


limzykenneth:
This looks like a great start to me. I've always feel the tests can be DRYer.

I also don't much like using `this` so much and would much prefer `behavesLikeABasicColor()` to be more functional, ie `behavesLikeABasicColor(c) // return true or false`, same goes for `isAP5Color()`. That way it should be more descriptive even without seeing the function definition. The test can just assert the value returned to be true or false.

The hardcoded colour value didn't bother me as much and I don't think it would become some special edge case in the future.

outofambit:
Thanks for the feedback @limzykenneth! I'll keep pushing it forward.

I did some digging on removing `this` but it looks like that is the only way to pass information between test contexts in mocha. [Here's the mocha issue requesting that](https://github.com/mochajs/mocha/issues/1441) (the owner of the project refuses to add the feature). Even [packages that offer mocha shared tests](https://github.com/jpstevens/mocha-shared) use the `this` approach. 😞 As long as we're using mocha, `this` may be the way to go. (pun intended)

limzykenneth:
I may be missing something but is it necessary to pass any information between the tests using the test instance itself? Wouldn't something like the following work?

```
function isAP5Color(c) {
  test('should create instance of p5.Color', function() {
    assert.instanceOf(c, p5.Color);
  });
}

function behavesLikeABasicColor(c) {

  isAP5Color(c);

  test('should correctly set RGBA property', function() {
    assert.deepEqual(c.levels, [255, 0, 102, 255]);
  });
}

suite('p5.prototype.color(r,g,b)', function() {
  setup(function() {
    c = myp5.color(255, 0, 102);
  });

  behavesLikeABasicColor(c);

  ...
});

**Comment by limzykenneth** [comment]
Hmm....I had a try by cloning your repo and the best I can come up with is 
```
setup(function() {
  c = myp5.color(255, 0, 102);
  behavesLikeABasicColor(c);
});
```

Not sure if that's a good solution or we are better off using `this`. 

The main problem is that `setup` is only run right before a `test` block in the same block. Any `test` block outside of the immediate block including those called by a function will not have `setup` called and thus the variable not set. I think mocha also runs everything that is not in `setup` or `test` regardless of position so I can't find any way to run `setup` before our function call.

Why assigning to `this` works? Beats me... 😵 

I also tried assigning value of `c` outside of a `setup` block but then I got a couple of unexpected renderer related warnings and the FES's message. That can be another option but we need to figure out why the messages are there but not in the `setup` block.

**Comment by outofambit** [comment]
@limzykenneth thank you so much for taking a look at this! i really appreciate it. i was able to get something similar in a `test` block. 😒 totally agree with your `setup` sequencing explanation.

`this` is [used liberally](https://github.com/mochajs/mocha/blob/2bb2b9fa35818db7a02e5068364b0c417436b1af/lib/suite.js#L45) when a `describe` (or `suite`) block is defined and i believe refers to the block itself. so i guess its relatively safe given mocha's source? it looks like it won't be available to anything outside of the `describe`/`suite` block in which its assigned.

**Comment by outofambit** [comment]
I found a better solution!, I think, though this definitely needs plenty of cleanup. If we use a `subject` function we can define it at the top of the outermost `suite` block and pass it around to functions at the places we want. plus! with a little more work we can reuse subjects between tests really shorten this file. 😸

File: test/unit/color/p5.Color.js [hunk 1]
```diff
@@ -5,24 +5,36 @@ suite('p5.Color', function() {
   });
   var c;
 
-  suite('p5.prototype.color(r,g,b)', function() {
-    setup(function() {
-      c = myp5.color(255, 0, 102);
-    });
+  function subject() {
```
outofambit: subject function returns the color to be used for many tests

kadamwhite:
In general I really like the approach you're taking, but I agree that trusting so deeply in the references and execution order here feels a little brittle and constantly re-computing the color by calling `subject()` for multiple assertions in the same test technically means each assertion tests a different object.

I think a compromise might be to define a `factory()` method (renamed from `subject` to indicate it returns an object; naming's endlessly debatable!) for each suite, and to pass that factory to the helpers, but to still assign `c` using that method and testing against `c` with the suite-specific tests. This will mean many fewer lines need to be changed since `c` will still be valid:
```js
suite('p5.Color', function() {
  var myp5 = new p5(function( sketch ) {
    sketch.setup = function() {};
    sketch.draw = function() {};
  });
  var c;

  function isAP5Color(factory) {
    test('should create instance of p5.Color', function() {
      assert.instanceOf(factory(), p5.Color);
    });
  }

  suite('p5.prototype.color(r,g,b)', function() {
    function factory() {
      return myp5.color(255, 0, 102);
    }

    setup(function() {
      c = factory();
    });

    // Pass the method to the helper...
    isAP5Color(factory);

    test('shouldn\'t set HSBA property before hsb access func is called', function() {
      // ...but `c` still works here
      assert.equal(c.hsba, undefined);
    });
  });
});
```
From my point of view re-using subjects between tests just to shorten the file feels excessive, though, at least to me; I like being able to see all of the moving parts in a test without having to scroll too far. </personal-opinion> :)

limzykenneth:
I think @kadamwhite's solution is a good idea. Although as mentioned at the end as well:

> I like being able to see all of the moving parts in a test without having to scroll too far.

That is my feeling as well, especially with test like p5.Color being thousands of lines long, which is why I wanted it to be somewhat functional and not do too much opaque setups, ideally function calls are descriptive enough that I don't really need to look up its definition.

An idea, what about a factory function like so:

```javascript
function factory(){
  return myp5.color.apply(this, arguments);
}
```

That way it can be defined once and used multiple times. Although at this point it feels like just a fancy wrapper around `p5.color` function...

lmccart:
I like where this is going, it definitely feels cleaner. My one concern is about confusion it may cause for beginners. The oddities around the way contexts and variables are handled in mocha would not be obvious, and so it might not be clear why we're using this structure. What I want to avoid is someone thinking it's overly complicated and trying to simplify and getting lost, or undoing the work.

My suggestion would be to add some clear comment in this file explaining what's going on, and for other files we apply this approach to, adding a comment at the top linking back to the explanation here.

outofambit:
@lmccart that makes so much sense! let me take some time to think about what we can do to make this cleaner while maintaining its beginner friendliness. subjects, factories, and shared examples may not be the way to go. i want to be careful about increasing learning curves and complexity.

wherever we land i can also definitely do inline comments, as well as a wiki page.

lmccart:
@outofambit what's the status on this one -- my understanding is that it's in progress, is this correct? for organizational purposes, can we close this and reopen when it's ready for review?

outofambit:
@lmccart yes, absolutely, we can close it for the time being. its in progress, but i want to get more familiar with the codebase before i propose changes like this.

PR #2137, opened 2017-08-25
Initial implementation of custom shaders.

kjhollen:
- loadShader implementation via preload
- shader must be called in setup or draw, once gl context is
  established (can't be called in preload).
- call shader before setting uniforms
- test multi-texturing, with video texture as one of them
- toon shader test
- updates lighting method to test shader type before
  switching to a new shader; allows lighting to be
  used with custom shader.
- adds a few methods to p5.Shader for checking the shader
  type (does it have lighting? textures? color?)
- optimizes shader code by avoiding complication of prototype
  chain with local variables within drawing related methods.

A few custom examples are provided in
test/manual-test-samples/customShader:
- toonShader, a classic toon shader applied to a sphere
- videoMultiTextureShader, a weird shader that combines
  two images (one from a p5.Image and another from a video)
  and changes aspects of the color with movement of the
  mouse.

There's still more to do, like:

- Linking to detailed tutorials about how to write one's own shader
  (which is beyond the scope of the inline docs).
- More robust tests for all of the possible built-in uniforms for
  all shaders.
- Allowing the user to only specify a fragment shader, using a default
  vertex shader (or: choosing an appropriate vertex shader based
  on the user's supplied fragment shader).
- Letting the user specify shader source inline in the code?
- Unit tests.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -571,7 +571,8 @@ p5.prototype._preloadMethods = {
   loadShape: p5.prototype,
   loadTable: p5.prototype,
   loadFont: p5.prototype,
-  loadModel: p5.prototype
+  loadModel: p5.prototype,
+  loadShader: p5.prototype
 };
 
 p5.prototype._registeredMethods = { init: [], pre: [], post: [], remove: [] };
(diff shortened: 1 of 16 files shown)


kjhollen:
fixed some conflicts with the bind buffer changes, now going to merge this per conversation with @mlarghydracept!

Issue #1696, opened 2016-12-03
Rotation in WEBGL is not clockwise, but Counterclockwise 


technosalon:
Rotations in WEBGL mode should be clockwise, since P5 is on the left-hand system. However, it's been COUNTERCLOCKWISE, I think. You can see it easily by rotating a box along Z axis in a top view. The source code of "p5.Matrix.prototype.rotate" refers to "Toji 's glMatrix library" http://glmatrix.net, and both code seems to be identical, but I guess Toji's library has been on the RIGHT-HAND system, since the normal OpenGL depends on it. So, I guess it might be necessary to invert the direction of angles somewhere in p5... I hope I am wrong... Thank you in advance!

lmccart:
p5 uses the right-hand system, so the rotation is correct. more about webgl in p5 here: https://github.com/processing/p5.js/wiki/Getting-started-with-WebGL-in-p5

technosalon:
Very disappointed answer after 4 months. You should check your reference by yourself. 

![img_3419](https://cloud.githubusercontent.com/assets/18400681/24319212/9608891c-1158-11e7-8dd8-d64572fef37d.PNG)

If p5 is on the "left-hand" sistem, we must change the directions of the x and y axis everywhere in our source codes. You should check what happens now by yourself, since this is the basis of our codes. Therefore, I think that this issue should not be closed.

technosalon:
Dear @lmccart This is the very sad reply after 4 months. The p5 is on "right-handed" system. You should have read your reference by yourself, before answering.

![img_3419](https://cloud.githubusercontent.com/assets/18400681/24319279/230fbc30-115a-11e7-8985-822b560d745c.PNG)

If the p5 is on the "left-handed" system, then, you must change your directions of the x and y axis. I think you should check what happens now on p5's 3D. Therefore, for the future of this lovely p5 to be extended in 3d world, you  should not close this issue, and should ask some 3d experts for what we should fix, I sincerely recommend.  I am sorry that I cannot help this anymore. All the best.

technosalon:
Dear @lmccart this is the sad answer after 4 months. You should have read your reference by yourself, before answering.

![img_3419](https://cloud.githubusercontent.com/assets/18400681/24319335/572df62e-115c-11e7-86f0-38e42399e6d5.PNG)

[...]

PR #2139, opened 2017-08-26
webgl: use right lefthand coordinate system; add more camera api.

kjhollen:
this should fix issue #1696 with coordinate system rotation.

CAMERA:
can now supply a eye coord, center coord, and up vector.

COORDINATE SYSTEM:
when applying the perspective and orthographic transformations for
the camera, we now negate the y-component of the projection transform
in webgl mode. this results in a left-handed coordinate system through
and through: whereas before, the y-coordinates for positions were
inverted in the vertex shaders and in translate(). the result was
something that behaved like 2d-mode in mapping the positive y-direction
downward, but handled rotation differently. now everything should
match up and more closely match behavior from Processing, too.

so with that, changes include:
- update default vertex shader .glsl files to not negate y-coordinate
- implement custom perspective and orthographic handling of camera
  outside of the matrix class, so as to apply our little y-coordinate
  trick in the same way Processing does.
- implement a camera matrix that includes the lookAt information,
  which is persistent between frames. that matrix is then the
  first thing applied to the modelview matrix in subsequent frames;
  it will clobber any existing things in the modelview matrix if
  called by a user mid-draw.

there is more documentation and clean-up to do here, but I want to
get this in so that Stalgia can use it!

### DIFF
--- a/src/webgl/camera.js
+++ b/src/webgl/camera.js
@@ -34,9 +34,113 @@ var p5 = require('../core/core');
  * blue square shrinks in size grows to fill canvas. disappears then loops.
  *
  */
-p5.prototype.camera = function(x, y, z){
-  //what it manipulates is the model view matrix
-  this._renderer.translate(-x, -y, -z);
+p5.prototype.camera = function(){
+  this._renderer.camera.apply(this._renderer, arguments);
+};
+
+p5.RendererGL.prototype.camera = function () {
+  var eyeX, eyeY, eyeZ;
+  var centerX, centerY, centerZ;
+  var upX, upY, upZ;
+  if (arguments.length === 0) {
+    eyeX = this.defaultCameraX;
+    eyeY = this.defaultCameraY;
+    eyeZ = this.defaultCameraZ;
+    centerX = eyeX;
+    centerY = eyeY;
+    centerZ = 0;
+    upX = 0;
+    upY = 1;
+    upZ = 0;
+  } else {
+    eyeX = arguments[0];
+    eyeY = arguments[1];
+    eyeZ = arguments[2];
+    centerX = arguments[3];
+    centerY = arguments[4];
+    centerZ = arguments[5];
+    upX = arguments[6];
+    upY = arguments[7];
+    upZ = arguments[8];
+  }
+
+  this.cameraX = eyeX;
+  this.cameraY = eyeY;
+  this.cameraZ = eyeZ;
+
+  // calculate camera Z vector
+  var z0 = eyeX - centerX;
+  var z1 = eyeY - centerY;
+  var z2 = eyeZ - centerZ;
+
+  this.eyeDist = Math.sqrt(z0 * z0 + z1 * z1 + z2 * z2);
+  if (this.eyeDist !== 0) {
+    z0 /= this.eyeDist;
+    z1 /= this.eyeDist;
+    z2 /= this.eyeDist;
+  }
+
+  // calculate camera Y vector
+  var y0 = upX;
+  var y1 = upY;
+  var y2 = upZ;
+
+  // computer x vector as y cross z
+  var x0 =  y1 * z2 - y2 * z1;
+  var x1 = -y0 * z2 + y2 * z0;
+  var x2 =  y0 * z1 - y1 * z0;
+
+  // recomputer y = z cross x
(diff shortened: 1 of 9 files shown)


mlarghydracept:
the PR title should maybe be "use correct left-handed coordinate system ..." :stuck_out_tongue_closed_eyes:
No but actually you are my hero!

kjhollen:
ahhh so the camera() function in this patch is a lookAt function in the old sense of the term, where you have to give the eye point, an aim vector, and an up vector. I think we could have a lookAt function that takes only the eye point or aim though, since that seems more in keeping with modern GL toolkits?

mlarghydracept:
Yeah I mean it really isn't super necessary but if I am following then a new camera object is made and the old is disposed when you want to change the eye point/aim. It might be nice to be able to throw a `lookAt(myVector)` in the draw loop without performance hits if I want to follow a location that is animating or something.

src/webgl/camera.js
```diff
@@ -34,9 +34,113 @@ var p5 = require('../core/core');
...
+    eyeX = this.defaultCameraX;
+    eyeY = this.defaultCameraY;
+    eyeZ = this.defaultCameraZ;
+    centerX = eyeX;
+    centerY = eyeY;
+    centerZ = 0;
+    upX = 0;
+    upY = 1;
+    upZ = 0;
+  } else {
```
lmccart: looks like the documentation requests three numbers, but this method is expecting 0 or 9. can we update the docs to reflect these two options?

lmccart: does it make most sense to pass in the arguments at 9 numbers, or as three vectors of some kind? maybe plain numbers matches better with the rest of the API?

src/webgl/camera.js
```diff
@@ -78,14 +182,33 @@ p5.prototype.camera = function(x, y, z){
...
-  this._renderer._curCamera = 'custom';
+p5.prototype.perspective = function() {
+  this._renderer.perspective.apply(this._renderer, arguments);
+};
+
+p5.RendererGL.prototype.perspective = function() {
+  var fovy = arguments[0] || this.defaultCameraFOV;
+  var aspect = arguments[1] || this.defaultCameraAspect;
+  var near = arguments[2] || this.defaultCameraNear;
+  var far = arguments[3] || this.defaultCameraFar;
```
lmccart: are all four of these optional, or is this just a safeguard? if optional, let's indicate this in the docs with [] around the params

src/webgl/p5.Matrix.js
```diff
@@ -82,10 +82,26 @@ p5.Matrix.prototype.set = function (inMatrix) {
   if (inMatrix instanceof p5.Matrix) {
     this.mat4 = inMatrix.mat4;
     return this;
-  }
-  else if (inMatrix instanceof GLMAT_ARRAY_TYPE) {
+  } else if (inMatrix instanceof GLMAT_ARRAY_TYPE) {
     this.mat4 = inMatrix;
     return this;
+  } else if (arguments.length === 16) {
```
lmccart: looks like this case is not indicated as an option in the docs, can we add it?

src/webgl/camera.js
kjhollen: Yes. I've also been wondering if this 9-argument (or 3-vector) version is outdated in 2017 (it didn't seem that three.js still defines the camera in this way, for example, but this is what feels comfortable to me from gluLookAt). It probably makes sense for their to be a 3-argument version, too, or another method called lookAt that would only affect the aim of the camera.

src/webgl/p5.Matrix.js
kjhollen: yes. thanks for catching this and other incongruities! I can get to this tomorrow & resubmit.

kjhollen: question for this one: the reason to do 16 args is to get around array creation, because this now gets used every frame for camera work and creating arrays slows down the computation—but it feels a bit much to list all 16 arguments here. I went with 

```
 * @param {Number}          [n00..n33] 16 numbers passed by value to avoid
 *                                     array copying.
```
Does that seem ok? From what I can tell, the Matrix class isn't included in the docs but other devs would want to see this for reference.

lmccart: ah I see. this makes sense to me.

Issue #1517, opened 2016-07-12
Color property setters


Jackbennett:
I had a look around and it seems like I shouldn't go in the the _array object but this creates the effect I was after. It might be nice to open up the color properties to be modified by other code a bit easier. Or perhaps I could have set an `alpha()` value before fill to be more inline with the p5 api?

```
var pallet = {}
var pallet.focus = color(159,39,55)
items.forEach(function(current, index, array){
  pallet.focus._array[4] = map(index, 0, array.length, 0, 1)
  fill(pallet.focus)
  ellipse(current.x, current.y ,15, 15)
})
```

nok:
In your sketch the red, green and blue value are fix. Have a look at the following version:

```javascript
var items = [];

function setup() {
  createCanvas(120, 80);
  background(210);
  stroke(255);
  
  items.push(createVector(30, 40));
  items.push(createVector(50, 40));
  items.push(createVector(70, 40));
  items.push(createVector(90, 40));
  
  items.forEach(function(current, index) {
    var alpha = map(index, 0, items.length-1, 0, 255);  // <--- alpha
    fill(color(159, 39, 55, alpha));                    // <--- r,g,b
    ellipse(current.x, current.y ,15, 15);
  });
}
```
http://codepen.io/nok/pen/qNVPqb

But if you want to store the color values separately, you can create an own class, e.g.:

```javascript
function MyColor (r, g, b, a) {
    this.r = r;
    this.g = g;
    this.b = b;
    this.alpha = a;
}
var c = new MyColor(159, 39, 55, 100);
c.alpha = 200;
color(c.r, c.g, c.b, c.alpha);  // 159, 39, 55, 200
```

lmccart:
I think this is a good idea, I can give it a go next chance I get.

adding this code snipped proposal from @indefinit (https://github.com/processing/p5.js/issues/1566):
```javascript
var myCol = color(100,200,12); //create an opaque color
...
myCol.alpha(40); //set the alpha property to 40
// or
myCol.alpha =  40;
```

lmccart:
closed with https://github.com/processing/p5.js/pull/2143

PR #2143, opened 2017-08-28
color: red, green, blue, and alpha setters

outofambit:
Hi there! I've implemented setters for red, green, blue, and alpha values for `p5.Color`, similar to what was requested in #1517. There was a little refactoring in the process, mostly to make certain parts of the code reusable within the class (see `p5.Color._calculateLevels()`).

A few key choices I made:
* syntax is `p5.Color.setAlpha()` (instead of `p5.Color.alpha()`), etc so use for method is clearer
* the input for each setter is interpreted in the color mode and ranges defined in the color instance itself (we don't change that based on the current color mode and ranges of the renderer)
* red, green, and blue setters can be used on colors that have a HSB or HSL mode stored (the color's ranges for RGB mode that are stored at the time of construction are used)
* `p5.Color.levels` is automatically recalculated immediately after any setting using these methods

Let me know what you think!

PS—
I have another branch for hue, saturation, lightness, and brightness setters but that's a little more complicated so I'd like to submit it as a separate PR later.

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -31,8 +31,7 @@ var color_conversion = require('./color_conversion');
 p5.Color = function(renderer, vals) {
 
   // Record color mode and maxes at time of construction.
-  this.mode = renderer._colorMode;
-  this.maxes = renderer._colorMaxes;
+  this._storeModeAndMaxes(renderer._colorMode, renderer._colorMaxes);
 
   // Calculate normalized RGBA values.
   if (this.mode !== constants.RGB &&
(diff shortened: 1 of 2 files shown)


lmccart:
This looks great. I paused for a while thinking about .setRed() vs .red(). We have a general rule (or Processing did at least) of using less words when possible. However, we have such a mix and exceptions in the API currently that I think it's best to proceed with what feels most intuitive. In this case, and especially because it's not a top level API function, it feels good to me.

PR #2145, opened 2017-08-28
Stroke and Fill

### LINKED ISSUE #1680: strokeWeight() not working in WebGL mode.
Hi, I'm trying to build some 3D objects in the WebGL mode and I'm having some trouble with the strokes and fills.

When I use the pre-fabd objects, such as Box, Cylinder, Sphere I cannot see the lines that form the skeleton. That is, if I set the fill() to an alpha value of 0 the object disappears completely, even if the stroke() is set to 255 for instance.

Inversely, when I create my own 3D objects using the beginShape() / vertex() / endShape() technique, I can see the edges of the objetcs, colored by the stroke() function, but I cannot fill the sides of the object with the fill() function as I did in Processing.
Moreover, the strokeWeight() function seems to have no effect at all, with a static value of 1 for all lines.

Thank you!

### LINKED ISSUE #1703: beginShape() not taking stroke() values in WebGL
Hello, I've found that the beginShape() / endShape function does not take into account the stroke() value when drawing in the WebGL mode. It seems to me as if the fill() value overrid the stroke() value, since the lines are colored with this value. However this is problematic when we need to color the lines with a different color than the sides, and specially since the line(), point() and ellipse() functions do not work in WebGL, what forces the use of beginShape() / endShape(), even for simple shapes.
Also, the strokeWeight() is not respected by the beginShape() / endShape() objects....
Any help will be appreciated! Thanks!

mlarghydracept:
This PR implements GPU stroking for retained-mode primitives and loaded models bringing p5 more in line with default 3D behavior of Processing. Strokes accept colors and weight. They also adjust size relative to distance being displayed from viewport.

**NOTE** This PR includes code from @kjhollen's PR #2139. Any changes in camera.js or p5.Matrix.js are part of those changes not this one. That PR should be merged before this one**NOTE**

**NOTE** This PR allows two shaders to be loaded simultaneously. This is done through the use of `curFillShader` and `curStrokeShader` in `RendererGL`. This is a fairly large system-wide change and will likely cause conflict with @kjhollen's Custom Shader PR #2137. That PR should also be merged before this one so that I can fix all of the conflicts (which I anticipate many) **NOTE**

This PR still has a few major todos but wanted to put this up so it can be reviewed while I write documentation.

Todos before merge:
-  ~~Website examples~~
- ~~Unit tests~~

Todos for later (will open issue after this is merged):
- ~~Allow the line primitive to use the lineShader. This should be straightforward.~~
- make joins and caps for lines (still a fairly big task)
- ~~Stroke immediate-mode objects. This too should be straightforward.~~

closes #1680 
closes #1703


src/webgl/material.js
```diff
@@ -126,12 +129,15 @@ p5.prototype.texture = function(){
   gl.enable(gl.BLEND);
   gl.blendFunc(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA);
 
-  renderer.drawMode = 'texture';
-  var shader = renderer.setShader(renderer._getLightShader());
+  renderer.drawMode = constants.TEXTURE;
+  var shader = renderer.setFillShader(renderer._getLightShader());
+  if(this._renderer.curFillShader.active === false) {
```
lmccart: is this check necessary?

src/webgl/p5.Geometry.js
```diff
@@ -147,6 +156,69 @@ p5.Geometry.prototype.averagePoleNormals = function() {
...
+ * @return {p5.Geometry}
+ */
+p5.Geometry.prototype._makeTriangleEdges = function() {
+  if (Array.isArray(this.strokeIndices)) {
+    for (var i=0, max=this.strokeIndices.length; i<max; i++) {
+      this.edges.push(this.strokeIndices[i]);
+    }
+  }
+  else {
+    for (var j=0, amt=this.faces.length; j<amt; j++) {
```
lmccart: this for loop instantiation is hard to read, can we just write:
`for (var j=0; j<this.faces.length; j++) {`

src/webgl/p5.Geometry.js
```diff
@@ -147,6 +156,69 @@ p5.Geometry.prototype.averagePoleNormals = function() {
...
+      this._addEdge(this.faces[j][1], this.faces[j][2]);
+      this._addEdge(this.faces[j][2], this.faces[j][0]);
+    }
+  }
+  return this;
+};
+
+p5.Geometry.prototype._addEdge = function(i, j) {
+  var edge = [i, j];
+  this.edges.push(edge);
```
lmccart: you can collapse these two lines into one:
`this.edges.push([i,j]);`

src/webgl/p5.Geometry.js
```diff
@@ -147,6 +156,69 @@ p5.Geometry.prototype.averagePoleNormals = function() {
...
+
+/**
+ * Create 4 vertices for each stroke line, two at the beginning position
+ * and two at the end position. These vertices are displaced relative to
+ * that line's normal on the GPU
+ * @return {p5.Geometry}
+ */
+p5.Geometry.prototype._edgesToVertices = function() {
+  this.lineVertices = [];
+  var vertices = this.lineVertices;
```
lmccart: what's happening on this line? could we just write this.lineVertices anywhere vertices is written below?

src/webgl/p5.Geometry.js
```diff
@@ -147,6 +156,69 @@ p5.Geometry.prototype.averagePoleNormals = function() {
...
+
+  for(var i = 0, max = this.edges.length; i < max; i++)
```
lmccart: `for(var i = 0; i < this.edges.length; i++) {`

src/webgl/p5.Geometry.js
```diff
@@ -147,6 +156,69 @@ p5.Geometry.prototype.averagePoleNormals = function() {
...
+        c = end,
+        d = end;
+    var dirAdd = dir.array();
+    var dirSub = dir.array();
+    // below is used to displace the pair of vertices at beginning and end
+    // in opposite directions
+    dirAdd.push(1);
+    dirSub.push(-1);
+    this.lineNormals.push(dirAdd,dirSub,dirAdd,dirAdd,dirSub,dirSub);
+    _store([a, b, c, c, b, d]);
```
lmccart: do we need a, b, c, d instantiated above, or can we just write begin and end? or maybe the a, b, c, d adds some clarity that I'm not realizing?

src/webgl/p5.Geometry.js
```diff
@@ -147,6 +156,69 @@ p5.Geometry.prototype.averagePoleNormals = function() {
...
+  }
+
+  function _store(verts) {
+    for (var i = 0, max = verts.length; i < max; i += 1) {
```
lmccart: `for (var i = 0; i < verts.length; i++) {`

lmccart: actually, rather than a for loop, I think this is a good place to use the array.map function:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -98,36 +106,95 @@ p5.RendererGL.prototype.vertex = function(){
...
-
+    this.setFillShader(this._getImmediateModeShader());
     // note that if we're using the texture shader...
     // this shouldn't change. :)
   }
-  shader.bindShader();
+
+  if(this.curStrokeShader.active === true) {
+    for(var i=0; i<this.immediateMode.vertexPositions.length; i++) {
+      if(i+1 < this.immediateMode.vertexPositions.length) {
```
lmccart: `if (i < this.immediateMode.vertexPositions.length - 1) {` is a little more conventional way to write this

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -98,36 +106,95 @@ p5.RendererGL.prototype.vertex = function(){
...
+
+/**
+ * Create 4 vertices for each stroke line, two at the beginning position
+ * and two at the end position. These vertices are displaced relative to
+ * that line's normal on the GPU
+ * @return {p5.Geometry}
+ */
+p5.RendererGL.prototype._edgesToVerticesImmediateMode = function() {
+  var vertices = this.immediateMode.lineVertices;
+  for(var i = 0, max = this.immediateMode.edges.length; i < max; i++)
```
lmccart: this can be simplified as above

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -98,36 +106,95 @@ p5.RendererGL.prototype.vertex = function(){
...
+  return this;
+};
+
+/**
+ * Create 4 vertices for each stroke line, two at the beginning position
+ * and two at the end position. These vertices are displaced relative to
+ * that line's normal on the GPU
+ * @return {p5.Geometry}
+ */
+p5.RendererGL.prototype._edgesToVerticesImmediateMode = function() {
```
lmccart: this code looks very similar to _edgesToVertices() which makes me wonder if it can be abstracted somehow rather than duplicating the code?

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -165,19 +232,43 @@ function(mode, isCurve, isBezier,isQuadratic, isContour, shapeKind){
...
+  } else {
+    return [];
+  }
+}
+
+function vToNArray(arr){
+  return flatten(arr.map(function(item){
+    return [item.x, item.y, item.z];
+  }));
+}
```
lmccart: Is the arr here a p5.Vector? Can we use normal JS arrays instead? 
then I think you could just call flatten on that array and skip this step.

src/webgl/p5.RendererGL.Retained.js
```diff
@@ -83,48 +92,60 @@ p5.RendererGL.prototype.createBuffers = function(gId, obj) {
...
-  shader.enableAttrib(
-    shader.attributes.aTexCoord.location, 2, gl.FLOAT, false, 0, 0);
-
-  if(this.drawMode === constants.STROKE) {
-    this._drawElements(gl.LINES, gId);
-  } else {
+  var fillShader = this.curFillShader;
+  var strokeShader = this.curStrokeShader;
+
+  if(strokeShader.active !== false) {
```
lmccart: `if (strokeShader.active === true) {`

src/webgl/p5.RendererGL.Retained.js
```diff
@@ -83,48 +92,60 @@ p5.RendererGL.prototype.createBuffers = function(gId, obj) {
...
+    this._bindBuffer(this.gHash[gId].lineVertexBuffer, gl.ARRAY_BUFFER);
+    strokeShader.enableAttrib(strokeShader.attributes.aPosition.location,
+      3, gl.FLOAT, false, 0, 0);
+    this._bindBuffer(this.gHash[gId].lineNormalBuffer, gl.ARRAY_BUFFER);
+    strokeShader.enableAttrib(strokeShader.attributes.aDirection.location,
+      4, gl.FLOAT, false, 0, 0);
+    this._drawArrays(gl.TRIANGLES, gId);
+    strokeShader.unbindShader();
+  }
+  if(fillShader.active !== false) {
```
lmccart: same as above

src/webgl/primitives.js
```diff
@@ -74,10 +74,17 @@ p5.prototype.plane = function(){
...
-      .computeNormals();
+       .computeFaces()
+       .computeNormals();
+    if(detailX <= 1 && detailY <= 1) {
+      planeGeom
+        ._makeTriangleEdges()
+        ._edgesToVertices();
+    } else {
+      console.log('Cannot draw stroke on plane objects with more'+
+        'than 1 detailX or 1 detailY');
```
lmccart: missing a space

src/webgl/primitives.js
```diff
@@ -134,6 +141,23 @@ p5.prototype.box = function(){
...
+        [8,9],
+        [9,11],
+        [14,15],
+        [16,17],
+        [17,19],
+        [18,19],
+        [20,21],
+        [22,23]
+      ];
+      this.strokeIndices = strokeIndices;
```
lmccart: we can cut this and have line 146 read:
`this.strokeVertices = [`

mlarghydracept:
Thanks for the feedback! Yeah I agree with all of the changes 100%. The code definitely needs a cleanliness and clarity pass. I will probably wait until @kjhollen's custom shader PR is merged to make those changes. That merge will require a bit of rewriting for me and it probably makes sense to know the final layout of everything before doing final pass stuff.

src/webgl/p5.Geometry.js
sansumbrella: IIRC, we need to capture a reference to the vertices variable so it is properly scoped in the closure that pushes data to it (as the meaning of `this` changes).

mlarghydracept: Yeah that is the reason for doing that before but I am changing it quite a bit right now and it looks more like `this.lineVertices.push(a, b, c, c, b, d);`. I convert them to arrays earlier in the code and that way I can get rid of the `store` function altogether.

lmccart: ah got it

src/webgl/p5.RendererGL.Immediate.js
mlarghydracept: Not sure how this can be done cleanly. A number of the prototypes that come along with the p5.Vector class are used. The problem mainly comes with abstracting `edgesToVertices` such that both immediate and retained mode use the same function. In doing so I have to assume that the Immediate Mode `vertexPositions` are also p5.Vector object because the retained mode positions need to be because there is extensive use of p5.Vector functions in calculating normals (among other things)

src/webgl/p5.Geometry.js
mlarghydracept: The use of a,b,c,d is helpful in clarifying that these are four points that are displaced in opposite directions on each side of a rectangle. Obviously this is just an opinion and I can get rid of it!

src/webgl/p5.Geometry.js
mlarghydracept: I changed this for-loop as per your request but just wanted to clarify why I was doing that. These loops are sometimes going over arrays that are many 10,000's of items long, and if a variable isn't established with the length of the array then the length of the array is checked 10,000's of times. It is a minor performance boost in most of the manual tests though and I agree that clarity and ease of entry for new contributors is more important than performance here!

mlarghydracept:
Okay I figured out how to merge other work in and changed the way everything works a bit. Also added the changes that you requested @lmccart. All manual and unit tests are tested and working! Let me know if there is anything else that should happen!

lmccart:
🙌

PR #2151, opened 2017-09-03
Minor correction to pixels documentation example

caitelatte:
The documented example for changing a pixel's color used `var d = pixelDensity` which wouldn't work because pixelDensity is a function. I would like to add the `()` afterwards. This line is already correct in other examples in the same file.


outofambit:
@caitelatte great catch!

lmccart:
thank you!

caitelatte:
Thanks all! 😃

PR #2163, opened 2017-09-12
docs/httpPost

austince:
Working on documentation from this issue: https://github.com/processing/p5.js/issues/1954

Let me know if this format looks good and I'll keep doin' it!

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -721,47 +721,40 @@ p5.prototype.httpGet = function () {
  * // Examples use jsonplaceholder.typicode.com for a Mock Data API
  *
  * var url = 'https://jsonplaceholder.typicode.com/posts';
- * var post = { userId: 1, title: 'p5 Clicked!', body: 'p5.js is way cool.' };
+ * var postData = { userId: 1, title: 'p5 Clicked!', body: 'p5.js is way cool.' };
  *
  * function setup() {
  *   createCanvas(800, 800);
  * }
  *
  * function mousePressed() {
- *
  *  // Pick new random color values
  *  var r = random(255);
  *  var g = random(255);
  *  var b = random(255);
  *
  *  httpPost(url, 'json',
- *      post,
+ *      postData,
  *      function (result) {
  *        strokeWeight(2);
  *        stroke(r, g, b);
  *        fill(r, g, b, 127);
  *        ellipse(mouseX, mouseY, 200, 200);
  *        text(result.body, mouseX, mouseY);
- *      },
- *      function (error) {
- *        console.error(error);
  *      });
  *
  *  // you can also omit the datatype and httpPost will guess
  *  httpPost(url,
- *      post,
+ *      postData,
  *      function (result) {
  *        strokeWeight(2);
  *        stroke(r, g, b);
  *        fill(r, g, b, 127);
  *        ellipse(mouseX, mouseY, 200, 200);
  *        text(result.body, mouseX, mouseY);
- *      },
- *      function (error) {
- *        console.error(error);
  *      });
- *      }
  *
+ * }
  * </code>
  * </div>
  *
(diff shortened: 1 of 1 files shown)


src/io/files.js
```diff
@@ -714,6 +714,92 @@ p5.prototype.httpGet = function () {
...
+ * <div><code>
+ *
+ *  var url = 'https://invalidURL'; // A bad URL that will cause errors
+ *  var post = { title: 'p5 Clicked!', body: 'p5.js is way cool.' };
+ *
+ *  function setup() {
+ *     createCanvas(800, 800);
+ *  }
+ *
+ *  function mousePressed() {
```
austince: Should I add a note about p5.dom ?

src/io/files.js
```diff
@@ -714,6 +714,88 @@ p5.prototype.httpGet = function () {
  * @param  {function}      [errorCallback] function to be executed if
  *                                    there is an error, response is passed
  *                                    in as first argument
+ *
+ * @example
+ * <div><code>
+ * // Examples use jsonplaceholder.typicode.com for a Mock Data API
+ *
+ * var url = 'https://jsonplaceholder.typicode.com/posts';
```
limzykenneth: I assume this service doesn't have rate limiting?

src/io/files.js
```diff
@@ -714,6 +714,88 @@ p5.prototype.httpGet = function () {
...
+ *  httpPost(url, 'json',
+ *      post,
+ *      function (result) {
+ *        strokeWeight(2);
+ *        stroke(r, g, b);
+ *        fill(r, g, b, 127);
+ *        ellipse(mouseX, mouseY, 200, 200);
+ *        text(result.body, mouseX, mouseY);
+ *      },
+ *      function (error) {
```
limzykenneth: I think the error callback can be omitted to show it's optional and I think it may only be used in some cases.

src/io/files.js
```diff
@@ -714,6 +714,88 @@ p5.prototype.httpGet = function () {
  * @param  {function}      [errorCallback] function to be executed if
  *                                    there is an error, response is passed
  *                                    in as first argument
+ *
+ * @example
+ * <div><code>
+ * // Examples use jsonplaceholder.typicode.com for a Mock Data API
+ *
+ * var url = 'https://jsonplaceholder.typicode.com/posts';
+ * var post = { userId: 1, title: 'p5 Clicked!', body: 'p5.js is way cool.' };
```
limzykenneth: I would prefer another name for the variable just so to not be confused with `httpDo`'s "method" argument

src/io/files.js
limzykenneth: This form of `mousePressed` is bundled with p5.js, only `p5.Element.mousePressed` is bundled with p5.dom.js so the user probably won't need p5.dom.js when running this example on their own.

src/io/files.js
```diff
@@ -714,6 +714,88 @@ p5.prototype.httpGet = function () {
...
+ *       var g = random(255);
+ *       var b = random(255);
+ *
+ *       httpPost(url, 'json',
+ *           post,
+ *           function (result) {
+ *             // ... won't be called
+ *           },
+ *           function (error) {
+ *             console.error(error);
```
limzykenneth: Maybe remove this `console.error`, the error callback should be a mechanism to catch error and not throw them. (Avoiding exception that will be thrown by default if no error callback is provided)

limzykenneth:
Thanks @austince! Overall just have a look at the line comments I left. Other than that, it would be nicer to have something that do more than just printing the returned data on screen but this serves as a good start.

src/io/files.js
austince: Cool - new here. Thanks!

src/io/files.js
austince: Sure, will do!

austince:
Sounds good, I can get more creative with it, just wanted to get general feedback on the format before I got too deep.

src/io/files.js
austince: I don't think so. Can't find any documentation and my own testing seems to agree

src/io/files.js
```diff
@@ -709,11 +709,87 @@ p5.prototype.httpGet = function () {
...
+ *      postData,
+ *      function (result) {
+ *        strokeWeight(2);
+ *        stroke(r, g, b);
+ *        fill(r, g, b, 127);
+ *        ellipse(mouseX, mouseY, 200, 200);
+ *        text(result.body, mouseX, mouseY);
+ *      });
+ *
+ *  // you can also omit the datatype and httpPost will guess
```
lmccart: I think we can omit this block.. it's useful to know but makes the example look longer than it actually is. Maybe instead, we just add your note to the [datatype] param definition line above

lmccart:
looks great! one small note then it's ready to merge.

src/io/files.js
austince: 👍

austince:
Cool, thanks! 💯

lmccart:
awesome thank you! 🎉

PR #2164, opened 2017-09-13
docs/httpDo

austince:
Working on documentation from this issue: https://github.com/processing/p5.js/issues/1954

Only added one example, using the `Request` options, but did something more creative with the data. Could do an example of `PUT` but I feel like it's better to let users get nifty with it. 

Using the same earthquake data api as used in the `httpGet` example;

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -774,6 +774,7 @@ p5.prototype.httpPost = function () {
  *    var feature = earthquakes.features[eqFeatureIndex];
  *    var mag = feature.properties.mag;
  *    var rad = mag / 11 * ((width + height) / 2);
+ *    fill(255, 0, 0, 100);
  *    ellipse(
  *      width / 2 + random(-2, 2),
  *      height / 2 + random(-2, 2),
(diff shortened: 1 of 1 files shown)


src/io/files.js
```diff
@@ -740,6 +740,56 @@ p5.prototype.httpPost = function () {
...
+ *
+ * // displays an animation of all USGS earthquakes
+ * var earthquakes;
+ * var eqFeatureIndex = 0;
+ *
+ * function preload() {
+ *    var url = 'https://earthquake.usgs.gov/fdsnws/event/1/query?format=geojson';
+ *    httpDo(url,
+ *      {
+ *        method: 'GET',  // http method - "GET", "PUT", "POST", etc.
```
lmccart: let's omit this comment for readability, since this info is mentioned in the parameters already

austince: Just the method comment or the Request options one as well?

lmccart: I think the other one is fine to leave, since those options may be less clear

lmccart:
thanks!

PR #2165, opened 2017-09-13
Dust off old Vector testcases

quinox:
Converted the disabled `p5.Vector` Jasmine testcases to Mocha. 

Based on the output of the test runner all of the new tests are being executed, and successfully. If I randomly change numbers in the testcases they start failing so it seems they do check things correctly. 

I tried to change as little as possible. Besides whatever was needed to move from Jasmine to Mocha:
* I updated two testcases to change `v.get()` to `v.copy()`; `v.get` seems no longer available
* A bunch of testcases used `p5.Vector.angleBetween(v1, v2)`; this no longer works, I changed them into `v1.angleBetween(v2)`
* I fixed a typo, `differnce` -> `difference`


lmccart:
awesome, thank you!

PR #2166, opened 2017-09-15
Web gl gsoc17 for merge

kjhollen:
Updated against the latest master branch and ready to go!


kjhollen:
I assigned Stalgia as a reviewer, but we updated this in person together!

lmccart:
there are some conflicts due to the recently merged unit test updates. I can help resolve these when I pull in the changes.

lib/addons/p5.dom.js
```diff
@@ -1745,6 +1745,7 @@
     this._cueIDCounter = 0;
     this._cues = [];
     this._pixelDensity = 1;
+    this.modified = false;
```
sansumbrella: Is `modified` a property that p5 users should be aware of/might use?

Consider `_modified` if not.

src/image/loading_displaying.js
```diff
@@ -74,6 +74,7 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
 
     // Draw the image into the backing canvas of the p5.Image
     pImg.drawingContext.drawImage(img, 0, 0);
+    pImg.modified = true;
```
sansumbrella: Should we prefer the `setModified()` method here instead?

src/webgl/camera.js
```diff
@@ -34,20 +46,129 @@ var p5 = require('../core/core');
  * blue square shrinks in size grows to fill canvas. disappears then loops.
  *
  */
-p5.prototype.camera = function(x, y, z){
-  //what it manipulates is the model view matrix
-  this._renderer.translate(-x, -y, -z);
+p5.prototype.camera = function(){
```
sansumbrella: Can this long section of camera math be simplified using the p5.Vector class? Or other math primitives?

I realize speed might be slightly compromised by not having everything inlined...

sansumbrella:
Thanks for putting this together. I'm excited to see these changes in master!

kjhollen:
addressed David's concerns re: modified field being private and pushed an update.

for the camera, let's leave the optimizations for now. there are some new slow-downs from other features that have been added to the repo over the summer—I'll file issues for those. but in the mean time, I'd like to keep from cutting other optimizations especially if they involve array/vector creation. I had to add the 16-arg method signature for matrix copy because the performance hit was substantial.

kjhollen:
@lmccart I think I got the test updates in another commit so we should just be ready to go!

lmccart:
💻 🎉 🎉 🎉 🎉

kjhollen:
woooooooooooo!

PR #2168, opened 2017-09-16
Update calculation.js

akashraj9828:
Renamed variables in dist( ) function for easy understanding

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -171,17 +171,14 @@ p5.prototype.constrain = function(n, low, high) {
  * 2 ellipses joined by line. 1 ellipse moves with mouse X&Y. Distance displayed.
  *
  */
-p5.prototype.dist = function(x1, y1, z1, x2, y2, z2) {
+p5.prototype.dist = function() {
  
-  if (arguments.length === 4) {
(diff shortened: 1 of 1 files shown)


limzykenneth:
For me it seems more confusing because it is fitting a 2D four arguments signature with a six argument signature, then for 3D, everything kinda breaks down...

To better illustrate, with the original implementation, only two arguments are using the wrong name (in the 2D version), while with the updated one, four arguments are using the wrong name (in the 3D version).

akashraj9828:
I thought that mostly everyone uses dist() to calculate 2D distances and not so often 3D so I made it readable for 2D, but your point is completely fair here more variables are using wrong name. But this doesn't affects/breaks the dist( )  function so I guess it's okay.

lmccart:
I think the clearest way to do this is actually to remove the listed arguments altogether and then access them depending on case using the arguments array, for example:
```javascript
return hypot(arguments[1]-arguments[0], arguments[3]-arguments[2]);
```

akashraj9828:
i did as u suggested @lmccart 
now you can merge

lmccart:
thanks!

Issue #2169, opened 2017-09-16
TravisCI has not been running for ~4 Months


meiamsome:
Is this intentional?

There are 7 failing tests at the current commit (fd750bf), that were introduced in fcd82ab - a fix for #2156  by my estimates.

I will pull request fixes for these in a bit, but the TravisCI not running

lmccart:
no definitely not intentional, will look into it..

sakshamsaxena:
is #1881 related?

sakshamsaxena:
I logged in to Travis, only to find out that the building options were disabled. I did switch em on, let's see ..

lmccart:
wow I have no idea how they got switched off. 
it looks like it's running now as far as I can tell. 
<img width="1044" alt="screen shot 2017-09-28 at 5 27 25 pm" src="https://user-images.githubusercontent.com/191056/30995973-509cc4a8-a472-11e7-8ea5-6cc0dfb31df3.png">

PR #2170, opened 2017-09-16
Bugfix for color issues from fcd82ab

### LINKED ISSUE #2156: "default" fill color used to be white, is now black (0.5.14)
The following code produces two solid `rgb(0, 0, 0)` shapes (note that I have not called `fill()` before drawing the shape):

```
function setup() {
  createCanvas(400, 400);
}

function draw() {
  background(100, 149, 237);
  ellipse(50, 50, 60, 60);
  rect(40, 40, 35, 45);
}
```

Previously I *believe* that identical code would have produced `rgb(255, 255, 255)` shapes (with a `rgb(0, 0, 0)` stroke), and checking with an older version of the library loaded confirms this. You can see the difference in behavior between [the current tutorial page](https://p5js.org/reference/#/p5/rect) and [the same tutorial page from earlier this year](https://web.archive.org/web/20170613033453/https://p5js.org/reference/#/p5/rect).

Interestingly, the same code called without `createCanvas()` produces the `rgb(255, 255, 255)` fill:

```
function setup() {

}

function draw() {
  background(100, 149, 237);
  ellipse(50, 50, 60, 60);
  rect(40, 40, 35, 45);
}
```

Ultimately the difference isn't that important, and I'm willing to live with a "just call fill() first if you don't want undefined behavior" answer. (I originally came across this when revising tutorial materials for ICM this year, in which I introduce `createCanvas()` before introducing `fill()`.) But if it's not already defined somewhere easily accessible, it might be nice to know what behavior *should* be expected in this circumstance? Thanks!

meiamsome:
Re-fixes #2156 in a way that doesn't break some other tests (Boolean leaking into `color()`). Related to #2169.

Adds a test for the #2156 to prevent regression.

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -25,8 +25,9 @@ p5.Renderer2D = function(elt, pInst, isMainCanvas){
 p5.Renderer2D.prototype = Object.create(p5.Renderer.prototype);
 
 p5.Renderer2D.prototype._applyDefaults = function() {
-  this._setFill(constants._DEFAULT_FILL, true);
-  this._setStroke(constants._DEFAULT_STROKE, true);
+  this._cachedFillStyle = this._cachedStrokeStyle = undefined;
+  this._setFill(constants._DEFAULT_FILL);
+  this._setStroke(constants._DEFAULT_STROKE);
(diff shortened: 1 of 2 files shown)


lmccart:
great, thank you!

PR #2172, opened 2017-09-16
Fix applyMatrix naming scheme, add examples for applyMatrix and resetMatrix

meiamsome:
Unifies naming scheme with standard matrix notation, and updates documentation on the function. Added example usage of both `applyMatrix` and `resetMatrix`


src/core/transform.js
```diff
@@ -14,30 +14,97 @@ var constants = require('./constants');
...
- * @param  {Number} n02 numbers which define the 3x2 matrix to be multiplied
- * @param  {Number} n10 numbers which define the 3x2 matrix to be multiplied
- * @param  {Number} n11 numbers which define the 3x2 matrix to be multiplied
- * @param  {Number} n12 numbers which define the 3x2 matrix to be multiplied
+ * @param  {Number} n11 numbers which define the 2x3 matrix to be multiplied
+ * @param  {Number} n12 numbers which define the 2x3 matrix to be multiplied
+ * @param  {Number} n21 numbers which define the 2x3 matrix to be multiplied
+ * @param  {Number} n22 numbers which define the 2x3 matrix to be multiplied
+ * @param  {Number} n31 numbers which define the 2x3 matrix to be multiplied
+ * @param  {Number} n32 numbers which define the 2x3 matrix to be multiplied
```
limzykenneth: The original uses 0 index which is usually kore consistent with native APIs, do you think this can be changed to be 0 index as well?

src/core/transform.js
```diff
@@ -61,7 +128,16 @@ p5.prototype.pushMatrix = function() {
  * @example
  * <div>
  * <code>
- * // Example in the works.
+ * translate(50, 50);
+ * applyMatrix(1, 0, 0, 0, 0, 0);
+ * rect(0, 0, 20, 20);
+ * // The matrix here is degenerate, it is
```
limzykenneth: This is rather confusing, I don't know what degenerate here means nor why it is impossible to draw anything with a y-dimension and I doubt users with less experience with transformation matrix will understand this either. 

I think simply doing a `translate` -> `rect` -> `resetMatrix` -> `rect` can be enough to illustrate the functionality of `resetMatrix()`

limzykenneth:
Added a couple of comments, overall good work! One thing that would be nice though is to add is a link to a resource to either generate transformation matrix or explain how to prepare one yourself, extra points for good explanations of how they work.

src/core/transform.js
meiamsome: My reasoning behind this was the underlying web api uses notation that is closer to this: https://developer.mozilla.org/en-US/docs/Web/CSS/transform?v=a, 
in addition to most mathematical references using this notation.
Nethertheless, I'm certainly open to changing back to zero indexing - my main issue is that the original function arguments were incorrectly named for the zero-indexed case.

meiamsome:
@limzykenneth 
Totally, I would probably link to this wikipedia article: https://en.wikipedia.org/wiki/Transformation_matrix#Examples_in_2D_computer_graphics but I'm not sure if there's any specific markup for references in this documentation system?

limzykenneth:
@meiamsome You can find some explanation for writing docs [here](https://github.com/processing/p5.js/wiki/Inline-documentation)

src/core/transform.js
limzykenneth: I don't have a math or cs background so I couldn't really say which is better, I'm just more familiar with code stuff being zero-indexed. Let's have others pitch in what they think.

lmccart: the numbering does more closely align here, as counterintuitive as it may seem. however, I would suggest that we match the naming exactly as the mdn spec: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/transform, or use a-f which may be less confusion to someone that doesn't understand what m11 vs dx mean.

lmccart:
@meiamsome you should be able to include hyperlinks by using an html a tag, for example: https://github.com/processing/p5.js/blob/master/lib/addons/p5.dom.js#L919

src/core/transform.js
meiamsome: The base standard by W3C uses a-f: https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-transform, so perhaps it's best to change to that? It also warns against using matrix numbering because the transposition is used in some applications where the numbers would be reversed. I don't know what would be best for 'beginners', though whether any beginner would actually use this function is questionable. Perhaps I should add a link to MDN/W3C to the documentation as well?

lmccart: yeah let's go with a-f. I think adding a link to one of those refs is a good idea, too.

meiamsome:
Okay, @limzykenneth & @lmccart I think I've fixed those things now, let me know what you think (I included the matrix form in the docs for quicker reference but let me know if that's not a good idea or if I should switch to an embedded LaTeX rendered image or something else)

limzykenneth:
Overall looks good to me. I like the inclusion of the matrix directly in the docs, only thing I can pick on is as you mentioned, the presentation of it. I didn't build this PR to have a look on the site so I can't be sure if it works or not but I think maybe a LaTeX image wouldn't hurt things. 😉

meiamsome:
Here's a screenshot for you, but I think a latex image would be nicer... I'll work on that

image

meiamsome:
And this is with LaTeX:
image

limzykenneth:
Yeah, I personally prefer LaTeX especially with the brackets

meiamsome:
I'm just wondering: is there any issue with using an SVG instead of a PNG file for this?

limzykenneth:
Hmm...none that I know of, I think, since you can embed SVG in a  tag if you want.

meiamsome:
I was thinking it might be a problem with a font or something... I'll err on the side of caution and use PNG here.

limzykenneth:
Don't think it would matter but PNG should be just as fine

PR #2176, opened 2017-09-18
Test Standardization

meiamsome:
This is a continuation of the changes in #2175, so you don't need both.

Primary features:

- Fixed issues loading Modernizr for doc tests.
- Unified test specification for test.html and test-minified.html, in addition to adding a couple of missing files (1136 vs 1226 tests, so about 90 missing)
- Update all tests to have more appropriate suites so that names aren't duplicated unnecessarily.
- Updated asynchronous tests to make use of the Promise API. This means that they will no longer function on some platforms (Not sure which), but now all tests succeed on Firefox & Chrome where previously ~7 failed. This also means that assertions in asynchronous parts of these tests are now correctly handled, instead of being hidden. It shouldn't be a problem because this only matters during testing.


lmccart:
amazing, thank you!

Issue #2167, opened 2017-09-15
Bad lerp behaviour with string numbers


meiamsome:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Details about the bug: 

- p5.js version: 0.5.14
- Steps to reproduce this:

Lerp behaves very badly when passed numbers in strings:
```javascript
lerp("0", 2, 0.5) // Results in "10"
```
My proposed fix would be to replace the current
`return amt*(stop-start)+start;`
with:
`return amt * (stop - start) + +start;`
Which appears to fix all test cases I've tried...
Otherwise, an overly precautions take would be to do:
`return (+amt)*(+stop - +start) + +start;`

lmccart:
lerp is designed to take numbers rather than strings. perhaps this is a case for the friendly error system to help check this, but I worry about going down the rabbit hole of trying to make every function support any type of parameter.

meiamsome:
That does sound sensible. 

I wonder if there is a way to make the FES automatically apply itself to every publicly documented method. That seems like it would be simpler than relying on it being added to some by choice.

PR #2177, opened 2017-09-19
Automatically Inject FES

meiamsome:
I vaguely suggested this in #2167, but then I decided to see if it was possible and now this is the result.

**This is probably not ready for merging, this is just to see comments regarding this**

So, I removed all of the manual FES calls and instead injected them based on the documentation.

This of course means that we're relying on the documentation to be absolutely correct, and that's a problem. For example, `createCanvas` triggers a message by default simply because there's a hidden private 4th parameter. I think the way to fix that would be to mark it up as private, but the system doesn't seem to support that yet.

I would be interested in anyone trying this out with sketched to see if there are significant performance issues with using this compared to without.

For anyone who wants to try out a built version of this, I have it available at https://meiamso.me/p5.js

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -62,6 +62,7 @@ require('./webgl/camera');
 require('./webgl/interaction');
 
 require('./fes/error_helpers');
+require('./fes/inject');
 
 require('./core/init.js');
 
--- a/src/app.js
+++ b/src/app.js
@@ -61,6 +61,8 @@ require('./webgl/p5.Shader');
 require('./webgl/camera');
 require('./webgl/interaction');
 
+require('./fes/error_helpers');
+
 require('./core/init.js');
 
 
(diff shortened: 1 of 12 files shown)


lmccart:
I think this is exciting.. and it forces the documentation to be correct which is nice. It would be good to support some way of disabling it on a per function basis, I could imagine cases where it doesn't make sense or slows things down to be doing this checking. Either a tag in the documentation or a function call in the fxn definition. Or maybe there are groups of functions where we wouldn't want this, though I'm not sure what those are. @almchung you probably have thoughts :)

almchung:
@meiamsome Hi Oliver, this looks great and I also saw your comment on #2190. I'm unfortunately mostly unavailable to work on FES until November, so would you like take the lead on FES from this point? I would love to see the outcome of this approach.

PR #2179, opened 2017-09-19
More Testing Updates & some small doc fixes

### LINKED ISSUE #2178: Is resizeCanvas's third parameter optional?
The documentation suggests it is not, but the example suggests it is.

meiamsome:
All tests now operate on an instance of p5 to be closer to reality.

Fixed some instances where the wrong `this` was being passed in apply.

Updated two docs that were wrong (`resizeCanvas` and `texture`)

Promisified doctests to catch more async errors 

Fixes #2178

### DIFF
--- a/src/core/rendering.js
+++ b/src/core/rendering.js
@@ -123,7 +123,7 @@ p5.prototype.createCanvas = function(w, h, renderer) {
  * @method resizeCanvas
  * @param  {Number} w width of the canvas
  * @param  {Number} h height of the canvas
- * @param  {Boolean} noRedraw don't redraw the canvas immediately
+ * @param  {Boolean} [noRedraw] don't redraw the canvas immediately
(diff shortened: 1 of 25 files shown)


meiamsome:
Hmm, okay. Side effect of add `_pInst` to images is that https://github.com/processing/p5.js/pull/2179/files#diff-b224eabb1710349c5e23fcf38de6c7d9R213 no longer seems to correctly pick the image. I don't know the best way to deal with this problem - perhaps something like

```javascript
var ctx = (this.loadPixels || this._pInst.loadPixels)();
```

meiamsome:
Okay, I pulled out the commit that modified the code mentioned above because I had misunderstood what the this was representing. This should now be okay for a merge again.

lmccart:
wonderful!

PR #2197, opened 2017-09-23
Revised map functionality

marcusparsons:
Added in a `withinBounds` optional parameter that constrains the output to `stop2` and `start2` if set.

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -418,6 +418,7 @@ p5.prototype.mag = function(x, y) {
  * @param  {Number} stop1  upper bound of the value's current range
  * @param  {Number} start2 lower bound of the value's target range
  * @param  {Number} stop2  upper bound of the value's target range
+ * @param  {Boolean} withinBounds constrain the value to the newly mapped range
  * @return {Number}        remapped number
  * @example
  *   <div><code>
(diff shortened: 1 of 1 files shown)


marcusparsons:
Absolutely! How does that example look in the inline docs now?

src/math/calculation.js
```diff
@@ -418,6 +418,7 @@ p5.prototype.mag = function(x, y) {
  * @param  {Number} stop1  upper bound of the value's current range
  * @param  {Number} start2 lower bound of the value's target range
  * @param  {Number} stop2  upper bound of the value's target range
+ * @param  {Boolean} withinBounds constrain the value to the newly mapped range
```
lmccart: you can write it [withinBounds] to indicate that it's optional

src/math/calculation.js
```diff
@@ -435,7 +436,9 @@ p5.prototype.mag = function(x, y) {
  *       background(204);
  *       var x1 = map(mouseX, 0, width, 25, 75);
  *       ellipse(x1, 25, 25, 25);
- *       var x2 = map(mouseX, 0, width, 0, 100);
+ *       //This ellipse is constrained to the 0-100 range
+ *       //after setting withinBounds to true
+ *       var x2 = map(mouseX, 0, width, 0, 100, true);
```
lmccart: this is great!

marcusparsons:
Awesome blossom! I added in those brackets for the optional parameter. :)

lmccart:
🎉

Issue #2010, opened 2017-06-20
Detaching events


limzykenneth:
#### Nature of issue?

- [ ] Found a bug
- [x] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [x] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

I wonder if there is some way to detach event listeners from either the global scope or an element that are attached via p5.js or p5.dom.js? I've skim through the reference and found nothing about removing event listeners. If there is we might need to document it in the events themselves.

What I want to achieve is to swap out event handlers (like mouseClicked or mousePressed, etc) for a specific element. For obvious reasons I don't want a handler to be attached more than one time when I'm attaching event handlers at runtime (say clicking a DOM button) so I want to first remove the event listener before attaching a new one.

limzykenneth:
Found a workaround by directly removing the event listener from the DOM element
```
// From source of p5.dom.js
p5Element.elt.removeEventListener(ev, p5Element._events[ev]);
```

I still think it would be best to have something that handles this in the library itself.

lmccart:
yes, I think this would be useful... what about passing false or null as an argument? something like
```javascript
myElement.mousePressed(false);
```

limzykenneth:
That would be a start and I think can be implemented. However, I think if it can have finer grain control it would be even better, ie: when binding two different `mousePressed` listeners, I can choose one of them to be removed by referencing the callback passed, more or less like the native implementation.

lmccart:
so something like this?
```javascript
var handler = myElement.mousePressed(doStuff);
...
myElement.mousePressed(handler, false);
```
I wonder how far we go though? If you are at the point of using multiple event handlers and storing references to them, perhaps it's easier to just native JS? I'm not sure.

lmccart:
or maybe it's more intuitive to say:
```javascript
myElement.mousePressed(doStuff);
...
myElement.mousePressed(doStuff, false); // removes specific listener
myElement.mousePressed(false); // removes all listeners
```

for reference, jQuery has the [unbind](http://api.jquery.com/unbind/) method

limzykenneth:
The second one is more what I had in mind. jQuery also has namespacing in the form of `click:namespace` which is really nice.

[...]

PR #2204, opened 2017-09-24
Detach events

outofambit:
fix for #2010 

I've added a `detachListener` function that gets called when someone passes `false` to the user facing event registration methods or if they pass a new handler to be attached in place of another. Thus, users are limited to a single handler per user facing event registration method.

Also, added a test to make sure the basic machinery does what it says it will. 😸


lmccart:
the code all looks great. however, the three tests are failing for me, I'm not sure why.. are they passing on your machine?

then there's the question of documentation. my proposal would be to add an example to mousePressed() documentation (in events/mouse.js) that shows detaching a listener. then updating the docs for every method that calls adjustListener() with a second signature that takes one boolean argument. does this sound reasonable?

outofambit:
@lmccart oof sorry about the tests! (i swear i had them passing but it wasn't true.) they should be good now.

that documentation plan sounds totally reasonable. i'll have that up soon!

lmccart:
🎉

lyle:
It looks like this change, specifically "limit to attaching one listener per event at a time" caused #3591

lyle:
I don't think this implementation should be changed at all - I think .drop needs to figure out how to safely add it's two events. So ignore this new issue in this context.

PR #2209, opened 2017-09-27
Made regexp for reference test easier for Windows

sakshamsaxena:
On Windows, the acceleration examples were failing constantly, as the regexp to check for those files included forward slashes which are used only in Unix/BSD. 

(loadJSON still gives error randomly, working on that separately)


lmccart:
great!

Issue #2182, opened 2017-09-20
Implement image() for webgl mode


kjhollen:
#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Feature enhancement details:

image() only works in 2D mode currently, but it would be great for WEBGL mode, too. I think it could be implemented in a pretty straightforward way using other parts of the API: begin/endShape with vertex, and the p5.Texture cache in the renderer. 

this is a nice self-contained issue and would be great for a new contributor who wants to write some webgl code. see: [WebGL Module Architecture](https://github.com/processing/p5.js/blob/master/developer_docs/webgl_mode_architecture.md) for relevant documentation.

bomoko:
***raises hand***
I'd love to take a swing at this, if nobody else is working on it yet?

lmccart:
go for it!

bomoko:
@kjhollen I've managed to get a sense of what's involved here, and have even managed to bang up a working version (totally throwaway, pasted below) - I was hoping you could point me in the right direction with the texture though.

```
p5.RendererGL.prototype.image = function(image, sx, sy, sw, sh, dx, dy, w, h) {
  texture(image); //  <-- presumably I should be doing this some other way ...
  var iWidth = image.width;
  var iHeight = image.height;

  var sxPerc = sx / iWidth; //TODO: rework var names
  var syPerc = sy / iHeight;
  var sx1Perc = (sx + sw) / iWidth;
  var sy1Perc = (sy + sh) / iHeight;

  this.beginShape();
  this.vertex(dx, dy, 0, sxPerc, syPerc);
  this.vertex(dx + w, dy, 0, sx1Perc, syPerc);
  this.vertex(dx + w, dy + h,0, sx1Perc, sy1Perc);
  this.vertex(dx, dy + h,0, sxPerc, sy1Perc);
  this.endShape(CLOSE);
  return this;
}
```
You'll see that I'm calling `texture()` in the first line of my function to load up the image that gets passed through from the `image()` function in  `src/image/loading_displaying.js`.

[...]

Issue #2160, opened 2017-09-10
Update documentation for image() to clarify what arguments mean


greenfIame:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [x] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [x] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: v0.5.4
- Web browser and version: tested on Crome & Safari
- Operating System: tested on windows and MacOSX
- Steps to reproduce this:
Its a documentation issue, when using p5's Image function to draw an image the code listed on the reference is:
image(img,dx,dy,dWidth,dHeight,sx,sy,sWidth,sHeight)
(d for destination, s for source)
When in reality it is:
image(img,sx,sy,sWidth,sHeight,dx,dy,dWidth,dHeight)
(d for destination, s for source in both examples).

kaganjd:
@mathuramg and I are taking this so we can use it next Thurs in a 'How to contribute to p5' workshop at ITP.

austince:
@kaganjd is that workshop open to the public by any chance?

kaganjd:
@austince yeah, if you're in NYC, you should come by! details:

9/21, 6-7:30pm (ish)
721 Broadway, 4th floor, Meeting Room
my email is jen.kagan@nyu.edu if you have trouble getting into the building.

if you're not in NYC, we'll publish the notes here: https://github.com/MathuraMG/p5-contr-workshop

austince:
@kaganjd I am! I'll try to come by, thanks for the deets 🥇

kaganjd:
Please ignore that commit above-- was just a demo in the workshop.

But I did dig into this issue and I think the documentation for this function is confusing. I struggled to make sense of the "source" "destination" and "sub-rectangle" language.

What if we use "main canvas" "image canvas" "original image" and "drawn image" to describe how these dimensions relate to each other? See the comments in this sketch, where I draw different parts of the same photo using the different parameters: https://alpha.editor.p5js.org/kaganjd/sketches/SkLtyxzo-

[...]

PR #2211, opened 2017-09-27
Added image function to webGL renderer

bomoko:
***Edit - no longer WIP - see discussion below***

This addresses #2182 

I was hoping to get some feedback on this - particularly the fact that I'm passing the p5 instance through to my function as an argument in order to invoke its texture() function.

I've tried my best to not impact the logic of the `src/image/loading_displaying.js` image() function.

As it stands, this seems to give the expected results in both P2D and WEBGL modes.


kjhollen:
Thank you for doing this! Looks about right to me. For the p5 instance, the GL renderer has a variable this.pInst which would be appropriate for this use case—then you won't need to pass the p5 object as an argument!

bomoko:
ah, wonderful! Thanks for that, @kjhollen -- I'll make the change! Exactly the kind of advice I was hoping for.

bomoko:
Made changes as suggested.

limzykenneth:
The parameter order for 2D `image()` function in p5.js is `image(src, dx, dy, dw, dh, sx, sy, sw, sh)` (which does not follow HTML specs but perhaps easier to understand for beginners). I noticed this implementation follows the HTML specs of putting source before destination. I think it would be better to have them both have similar signature, by either switching the one here or the 2D one.

See also #2160, #2217 and #2232 for some discussion around the docs for `image()`

bomoko:
hey @limzykenneth

If you check out /src/image/loading_displaying.js there's what's essentially the "master" image() function there. That actually calls _either_ the 2D renderer or the WEBGL renderer's image() function.
Both of these latter functions share the same signature - see this, from `src/core/p5.Renderer2D.js` 

```
p5.Renderer2D.prototype.image =
  function (img, sx, sy, sWidth, sHeight, dx, dy, dWidth, dHeight) {
```
There image() that gets called is the same in both cases - it just passes along the values to the renderer.

We could certainly change the arguments, but we'd need to change both renderers and the "core" image function.
Happy to though.

limzykenneth:
Ah I see, you will be calling the "master" `image()` anyway if I understand correctly. I guess I'm a bit confused because there's a documentation entry in there but I assume they are not meant to be rendered on the site itself?

bomoko:
@limzykenneth  - understood 100%!
Should I remove the docs?
Sorry, I added them as an afterthought - I didn't consider that they might be rendered.
Silly, when I consider now that the other renderer's image doesn't have them.

limzykenneth:
I don't know what's best. On one hand it may help with maintaining the codebase so it may be nice to have an inline docs that doesn't get rendered meant for maintainers only.

bomoko:
Okay - I'm going remove them for now, just to keep it consistent with the existing 2D implementation, and since this seems like it might move us into "broader discussion" territory.

bomoko:
@kjhollen I think this is ready for you to review if you'd like to take a look (absolutely no rush, I just realized that you might not have seen this since the discussion above). 
I'm not quite sure how/if I'm supposed to set the PR into review status (sorry about that).

src/webgl/primitives.js
```diff
@@ -901,4 +902,52 @@ p5.RendererGL.prototype.line = function(x0,y0,z0,x1,y1,z1) {
...
+    this.beginShape();
+    this.vertex(dx, dy, 0, u0, v0);
+    this.vertex(dx + dw, dy, 0, u1, v0);
+    this.vertex(dx + dw, dy + dh,0, u1, v1);
+    this.vertex(dx, dy + dh,0, u0, v1);
+    this.endShape(constants.CLOSE);
+  }
+  return this;
+};
+
```
kjhollen: this is looking pretty good to me. Could we add an example sketch to manual-test-examples/webgl to see it in action? It would be great if that sketch can try the various imageModes, and also different argument versions of the function. Thank you for taking this on! :)

bomoko: Absolutely! I'll get on that today.

bomoko:
Hi @kjhollen -

My vote would be to "make it restore the previous render conditions after it finishes drawing the image" - I think this makes the most sense. Shall I go ahead and get that working?

kjhollen:
@bomoko yes, that sounds great!

bomoko:
@kjhollen - sorry this has taken some time (real life stuff came up) - I was wondering if you wouldn't mind pointing me in the right direction with regards to saving and then restoring the rendering conditions?
I don't want to hack an ugly solution to this if I don't need to :)
Do we do this anywhere else that you're aware of? Or do you have any suggestions on implementation that would be obvious to someone with more experience in the WebGL guts of the library?

kjhollen:
no problem! this is timely, reading this I realized we'd have the same issue with text rendering in WEBGL—I'm planning to implement it with textures, so I think we will need to save the previous the shader and texture info there, too. ideally we'd come up with something that works in both cases!

the most obvious thing I can think of is to put this in the push/pop methods for RendererGL—in fact, looking at the code, I think as implemented the push/pop methods probably don't work with styles in WEBGL mode (the properties get reset, but side effects setting up shaders etc. do not happen).

So what I'd like to propose is that we accept this patch as-is and open a new issue for fixing push/pop that I can take on as part of the text rendering work. What do you think @bomoko @mlarghydracept @lmccart?

bomoko:
@kjhollen I'm more than happy with that from my side (and thanks for the response!).
I'll watch your work on the new issue and when it's at a point, I'll rework the WebGL image to use what you implement.

bomoko:
Hi @kjhollen 
Just following up on this. I've tried to follow some of the issues around the push/pop implementations, and it looks like there has been some work done on it, but I couldn't find anything about textures specifically (if I compare what I've found with what you've done on #2370). 

I can't quite make out if it's yet possible to push/pop textures - is there anywhere else that this has been sorted out (like the text rendering you mentioned) that I might be able to take a look at?
Thanks!

Spongman:
i would recommend on holding off on this until after #2670 is merged. after that, this additional texture saving should be trivial to implement as that PR moves all of the rendering state out of the shaders and into the renderer and removes the dependency on stateful shader selection.

kjhollen:
This is still on my mind! An update: while #2670 did not get merged, I'll work on extracting the changes needed to make push/pop work from there and will leave this PR open until that's ready.

bomoko:
Fantastic, thanks for the update!

bomoko:
I'm going to close up this PR, I haven't been following the progress and I don't reckon this is going to address the issue as is.

wanbinkimoon:
is the pr merged and in production?

Issue #2212, opened 2017-09-28
Potential logic error in Line() function's parameter tests


bomoko:
#### Nature of issue?

- [X] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: v0.5.14

I think I've come across a small logic error in the WebGL line code. 

Currently, the function has a check of its parameters (just pasting the entire function here, it's teensy):

```
p5.RendererGL.prototype.line = function(x0,y0,z0,x1,y1,z1) {
  if (typeof x0 !== 'undefined' ||
      typeof y0 !== 'undefined' ||
      typeof z0 !== 'undefined' ||
      typeof x1 !== 'undefined' ||
      typeof y1 !== 'undefined' ||
      typeof z1 !== 'undefined')
  {
    this.beginShape();
    this.vertex(x0, y0, z0);
    this.vertex(x1, y1, z1);
    this.endShape();
  }
  return this;
};
```
The check will pass if _any_ of the parameters are _not_ undefined. I _think_ that what we actually want is for the line to be drawn only if we have all parameters -- that is, I think that what makes more sense here are ANDs and not ORs.

The problem with not checking that _every_ parameter is defined is that if someone calls this like one might call the 2D version, i.e. `line(x, y, x1, y1)` then the first vertex is going to be loaded with x1 as its z value.
Step debugging a call to `line()` in this manner confirms that this is what happens.

bomoko:
Closing - have a better solution ...

bomoko:
Apologies for closing then reopening the issue - I had an idea for an enhancement, then I thought it was probably better to deal with this directly, first.

lmccart:
fixed with https://github.com/processing/p5.js/pull/2213

PR #2213, opened 2017-09-28
Changed WEBGL line() parameter check logic

bomoko:
addresses #2212


bomoko:
Changing this logic ... closing till then

bomoko:
This fixes #2212 - reopened (I had some enhancements I wanted to suggest, but I thought it appropriate to discuss it after this was fixed)

lmccart:
good catch and fix. I think we could simplify this by leaving the args unnamed:

```javascript
p5.RendererGL.prototype.line = function() {
  if (arguments.length === 6) {
  }
  ...
```

lmccart:
on second thought I think this check should be removed completely and replaced with a validateParameters call if we want. as is, this will fail silently and be hard to track down the cause.

bomoko:
Hi @lmccart - So this function is called by the `line()` in `/src/2d_primitives.js` (which dispatches the call to either the 2d or 3d version of line(), depending on the renderer) - and there's actually already a `_validateParameters()` call in there - but because there any number of parameters greater than 4 is "valid", it's not going to complain (the documentation for the FES uses line() itself to illustrate false negatives).

I thought that we might move the _validateParameters() call out to the versions of line() in their respective renderers, but I don't think that makes sense, since they never call those functions explicitly.

lmccart:
I'm not sure I understand. Is the case we're missing just one where undefined is passed in as an argument instead of a number? At this point, I think we can call this on the side of asking the user to input reasonable variables, and if we feel this is still a common issue, work on refining the FES more to catch it.

bomoko:
Hi @lmccart 

I'm so sorry I wasn't clear - it's sometimes confusing to talk about functions like `line() `of which there are three. There's the "main" `line()` function that gets called from the sketches (and which lives in `src/core/2d_primitives.js`), and then there are the two "renderer" versions which live in `src/webgl/primitives.js` and `src/core/p5.Renderer2D.js`.
The latter two are called _from_ the first one. And it is the first line() function that already _has_ a `_validateParameters()` call in it.

If you take a look at my latest commits - what I've done is checked the number of arguments passed to the WEBGL version of line(). If the arguments length is 6, I treat the 6 values as two 3D points.

I also offer a fallback in case the person calls it like they would a 2D version of line. That is, if there are four arguments, I assume that they've passed two 2D coordinates, and I fill in the z-value for both with a 0.

Finally, I've followed some of the examples in the rest of the webgl/primitives.js file and added a console.log with a friendly error, in case someone passes anything other than 6 or 4 arguments.

I hope this is getting closer to a useful fix!

src/webgl/primitives.js
```diff
@@ -885,18 +885,20 @@ p5.RendererGL.prototype.curve=function
...
-    this.vertex(x1, y1, z1);
+    this.vertex(arguments[0], arguments[1], arguments[2]);
+    this.vertex(arguments[3], arguments[4], arguments[5]);
     this.endShape();
+  } else if(arguments.length === 4) {
+    this.beginShape();
+    this.vertex(arguments[0], arguments[1], 0);
+    this.vertex(arguments[2], arguments[3], 0);
+    this.endShape();
+  } else {
```
lmccart: sorry I'm still not understanding why this last part is necessary. shouldn't the validateParameters check in the top level line() function catch the case where line is called without either 4 or 6 arguments?

bomoko: Hi @lmccart - no worries. I'll try do a better job of explaining myself in future!

There are actually two separate issues here.

First - the issue of the validateParameters() call. Why is there a need for a parameter check if validateParameters() is _already_ being called?

The problem is that the validateParameters check in the top level line _won't_ catch the cases where more than 4, but not 6, parameters are passed to line() in 3D.
*If* the first 4 arguments passed in are numbers, *then* the validation passes.
This is a known issue with the FES - it's the second point in [this section in the docs](https://github.com/processing/p5.js/wiki/Friendly-Error-System#known-limitations).

I'm guessing that this is what the original series of checks that I flagged were trying to address - they just got the logic wrong (|| instead of &&).

Second - there's this extra condition I've added.
The reason I've added this condition for a 4 parameter call is that I'm assuming that if someone calls line() with 4 parameters, they're expecting the behavior that they'd get if they were in the 2D drawing mode. Namely, that the four parameters represent two 2d points.
In the case of 4 parameters passed to the line() function in 3D, I'm just popping in a 0 for the z-coordinate.

So if we drop this "else if (arguments.length === 4)" condition I've added, then line() in 3D will only ever work with 6 parameters, which is 100%.
It's only there to support the usual 2D case while drawing in 3D - more than happy to remove it though.

meiamsome:
Just as a note, I fixed FES' issue with additional arguments when I was messing around in #2177. I could pull that commit out as a separate PR quickly if you want?

bomoko:
@meiamsome - for my part I think that'd be awesome - then I'd be able to pull out some of the cruft in my PR too.

bomoko:
Hi @lmccart 

On this latest version I've reverted to trying to just address the logic error, but I've retained the fallback in the case that 4 parameters are passed (which is a valid set of inputs for line in 2D).

Happy to pick this up the FES question once @meiamsome's changes are in.

lmccart:
👍  sorry for all the back and forth, this looks good. the fallback makes complete sense. thanks!!

PR #2217, opened 2017-09-28
Update docs and add example for image()

kaganjd:
~~WIP~~ for #2160


lmccart:
is this ready for merge? or are you looking for feedback? if not, can we close this for organizational purposes and reopen when it needs review?

kaganjd:
Ah, sorry! Feedback, please!

lmccart:
This is looking good. My first thought is that a beginner coming to this just wanting to draw a single image at an x and y position could get lost or overwhelmed really fast. What if we try to split the two signatures more. This means:

0. We break the signatures into two, and don't reuse x, y, width and height. For the second signature we use dx, dy, dwidth and dheight.
1. The description text has two blocks. The first block describes things in terms of image, x, y, width, and height. It makes no mention of destination or source canvases. The second block could be more like your current description but using the second signature param names.
2. The example could be broken into two (or more) as well. The first few examples are only using x, y, width, height and the comments refer to them as such. The later examples could use the second signature params and the comments refer to them in terms of destination and source params.

Does this make sense?

kaganjd:
Yes, that all sounds good! Thanks. I agree that it would be a lot to stumble into.

Feel free to close for now and I'll reopen when I've incorporated those changes.

lmccart:
great!

bomoko:
Hi @kaganjd - wonderful stuff!
The only feedback I have has to do with lines 138 to 143. I have two points.

1. (not so important) I think that it might be useful to just add a short note about why you would want to pass through the source image parameters. Namely, if you want to display only a section of the source image (or a "chunk", as you say in the example). That'll give some context as to why we're interested in the source image at all.
2. I was a little confused by the line 

> Note that "destination canvas" in this context is different from the main sketch canvas you create with createCanvas().

I wasn't sure how this was the case. This is probably showing my own misunderstandings here, but I thought the destination canvas _was_ the main sketch canvas. And if it isn't, then maybe that line could use a little expanding (if I'm not getting it, then someone brand new probably won't too).

But, yeah, great stuff!

bomoko:
Hi @kaganjd, wonderful stuff.
Just two points.

First - I think it might be useful to have some context as to why we'd worry about the source image dimensions at all. That is, let them know that we use those parameters to grab a subsection (or "chunk" as you say in your example) of the source image, rather than drawing the whole thing to the canvas.

Second, I was a little confused by the remark that the "destination canvas" isn't the same canvas as created by `createCanvas()` (confession - I thought it was until reading this :/). Reading the link confused me a little more (they also speak of a destination _rectangle_). If it's confusing me, then there's a chance it might confuse other newbies.

Those are my only notes, hope you find them helpful! (and feel free to ignore them if I'm just being silly :) )

lmccart:
@kaganjd hey sorry, didn't mean to close your PR before others got to comment. I was just on an issue/pr cleanup rampage the other night and wasn't thinking :)

limzykenneth:
Thanks @kaganjd 

A few notes from me:
- `The last two parameters in this case—x and y—are the coordinates of the top-left corner of the first parameter, img.` This feels a bit unclear, "coordinates of the top-left corner" when replaced with something like "the position to place the image" feels easier to understand. I'm don't think much people will find the image position refer to the top left corner of the image very unintuitive but I don't have much experience to draw on myself so you can make the judgement.
- I agree with @bomoko about `destination canvas` being a bit confusing. I think it may be better that when we say "canvas" we mean the p5 main drawing canvas and not have another terminology that uses canvas as a name as well.
- If I'm not wrong, the signature we use differs from the HTML specs in that we do `image(img, dx, dy, dw, dh, sx, sy, sw, sh)` while the specs do `drawImage(img, sx, sy, sw, sh, dx, dy, dw, dh)`. I personally prefer to keep the current signature and so saying `p5.js follows the HTML specification` although referring to the naming of the parameters, can be misleading.
- Speaking of the specs, linking to the HTML living standard page is rather too verbose and not beginner friendly, as it sometimes contains implementation details meant for JS engine vendors. And since you just want to link to the diagram, you can simply add it directly to the documentation by using a HTML `` tag, you don't even need to navigate away for the diagram this way. 😃 Whether to link directly to the living standard's diagram or upload it to this repo depends on the website's deployment preference.

kaganjd:
Sorry, I did a bad Github thing.  Closing this while I fix it...

kaganjd:
I ruined this PR ;( 
Thanks for your feedback, everyone. The latest is now at #2232.

PR #2225, opened 2017-09-30
Added p5.MediaElement.src example

bomoko:
Partially addresses #1954 - specifically, it adds an example for p5.MediaElement.src


lmccart:
this is great, thank you!

Issue #2210, opened 2017-09-27
WebGL - Ellipse should not take detailY as an arg


Siphonophora:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: Current master branch
- Steps to reproduce this:
The ellipse primitive takes a detailX and detailY args. detailX determines the number of verticies in the ellipse. If detailY == 1 then the primitive is correctly rendered. For each value of detailY above 1, an entire set of redundant verticies/faces is added. 

I can fix that by fixing detailY = 1, but wanted to check that I am not missing some reason why the redundant geometry would be desired. 

Also, should the vertex locations be rounded at all (or is webgl going to automatically weld these together)? I noticed that the the vertices at the beginning and end of the ellipse didn't exactly match, similar to below
```
vector(200, 0, 0)  
vector(200,-4.2e-10, 0);

```

mlarghydracept:
Hi! Thanks for looking into this. I don't think any of the code that constructs the ellipse primitive has been looked at in a long time. I agree that the detailY sounds unnecessary and rounding would be nice even if only just for a cleaner understanding of the ellipse construction.

Spongman:
this was fixed by #2473

Issue #2201, opened 2017-09-23
Tests take a very long time in Chrome


meiamsome:
For some reason, the tests in p5.js currently take about 200s to run in Chrome on my PC, whereas the same tests take 6s on Firefox.

kaininc:
That's weird

sakshamsaxena:
A little more info here would be great. Specifically, OS, and the number of total tests passed and failed in both browsers. You think some were skipped due to some incompatibility/uncertainty in FF ?

meiamsome:
This seems to be fine for me now - Chrome is  61.0.3163.100.
This was at the same time I posted another issue stating Chrome as 59.0.3071.86
So therefore, I can only assume it was solved in Chrome 61?

Issue #2200, opened 2017-09-23
Three WEBGL tests are failing


meiamsome:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

On Chrome 59.0.3071.86 and Firefox 55.0.2 (And presumably others) there are three failing WEBGL tests that also break some doc tests.

meiamsome:
So, this just got a whole lot weirder for me.

I am currently working on getting headless chrome setup to run tests as well as phantom, and almost of of the tests that are failing in my browser are not failing in headless chrome. Maybe the particular webgl used with a real chrome is different?

meiamsome:
Given that this is broken on both FF and Chrome running on my computer in the same way, but is fine on a headless chrome with software-backed WEBGL I'm going to suggest that this may actually be an issue with the Linux NVIDIA Driver.

Issue #2174, opened 2017-09-17
PhantomJS is now no longer maintained


meiamsome:
For information, see https://github.com/ariya/phantomjs/issues/15105

This is just an issue to note that testing should probably be moved over to headless chrome (Probably puppeteer: https://github.com/GoogleChrome/puppeteer ) when the ecosystem is a bit better (I can't see a way to integrate it in the grunt file easily yet)

meiamsome:
I was working on a potential implementation of headless chrome testing with chrome-mocha here:

https://github.com/meiamsome/p5.js/tree/testing/headless-chrome

However I currently have an issue where on Travis it will always crash at the same place:

https://api.travis-ci.org/jobs/283092554/log.txt?deansi=true

Any ideas?

lmccart:
is webgl supported in headless-chrome?
https://bugs.chromium.org/p/chromium/issues/detail?id=617551

lmccart:
oh I just saw your comments on https://github.com/processing/p5.js/issues/2200 so I guess that isn't the problem

meiamsome:
So, I managed to get it to not crash by using the beta version of chrome, instead it just reports `Error: Error: Error creating webgl context`. Interestingly, I can only get that behaviour locally (Even when running inside one of Travis' docker containers) by using the `--disable-webgl` flag on chrome. Maybe it's disabled automatically on Travis. I might raise an issue over there at some point.

meiamsome:
@lmccart This is not fixed by 74019da, as @Spongman commented on there, we need to remove grunt-mocha and switch to another test runner (grunt-karma, perhaps?). Please re-open this.

PR #2229, opened 2017-10-01
Web GL - Arc added, Ellipse refactored

### LINKED ISSUE #2181: Implement arc() for webgl mode
#### Nature of issue?

- [ ] Found a bug
- [x] New feature request

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

arc() only works in 2D mode currently, but it would be great for WEBGL mode, too.

This would make a great first contribution for someone who wants to get to know the project and write a little webgl code! the implementation for ellipse() would be a good place to look as a reference.

Siphonophora:
Being my first pull, I would absolutely appreciate some feedback. 

Pull to close #2181 and #2210. 

1. Adds Arc to src/webgl/primitives.js and turned Ellipse into a wrapper for Arc to simplify the code and fix issues with Ellipse geometry
2. Updated documentation for above.
3. Moved the adjustment for ellipseMode() used by Arc from src/core/p5.Renderer2D.js to src/core/2d_primitives.js so the adjusment is shared by the 2d and webgl versions of Arc.
4. Updated test cases with the additional detailX parameter for Arc and Ellipse

Questions:
- Even though test-reference passes for everything I added, the Arcs won't render for WebGL within the reference. I have no idea why, given those tests pass, and the same code runs just fine as its own sketch. 
- There are tests failing, but they look like they are from other areas of the code. is there an easy way to diff the test failures in master against those in my branch?
- The reference for arc isn't useful without stroke rendering. I gathered that's because of an open issue and I shouldn't add it directly to the documentation, but let me know if that is wrong. 

Not sure how you guys do this review, but for what its worth, I was using this sketch to look at the output. you can easily switch it back and forth to webgl. mouse-wheel will adjust the detail in webgl mode. 

``` javascript
var img;
var wheelPos = 100020;


function preload(){
    img = loadImage("assets/ColorGrid.png");
}
// true false
function setup(){
    if(true){
        createCanvas(900,900,WEBGL)     
    } else {
        createCanvas(900,900)       
    }

}

function draw(){
    background(230)
    strokeWeight(10);
    fill(300,100,0)
    //texture(img);
     var grid = 210

    if(_renderer.name != 'p5.Renderer2D'){
        rotateY(frameCount * 0.01)
        translate(-width/2, -height/2)
    } 
    translate(-grid/2,- grid/2)

    //Config to vary the tests
    var d1     = 200
    var d2     = 200
    var start  = 0
    var eA     = 0
    var detail = shapeScroll();
    ellipseMode(CENTER);

    ellipse(grid*1, grid*1, d1, d2                     , detail);
    ellipse(grid*1, grid*2, d1, d2                     , detail);
    ellipse(grid*1, grid*3, d1, d2                     , detail);
    ellipse(grid*1, grid*4, d1, d2                     , detail);
    arc(    grid*2, grid*1, d1, d2, start, eA + PI/2  ,CHORD, detail);
    arc(    grid*2, grid*2, d1, d2, start, eA + PI    ,CHORD, detail);
    arc(    grid*2, grid*3, d1, d2, start, eA + PI*3/2,CHORD, detail);
    arc(    grid*2, grid*4, d1, d2, start, eA + TAU   ,CHORD, detail);
    arc(    grid*3, grid*1, d1, d2, start, eA + PI/2  , OPEN, detail);
    arc(    grid*3, grid*2, d1, d2, start, eA + PI    , OPEN, detail);
    arc(    grid*3, grid*3, d1, d2, start, eA + PI*3/2, OPEN, detail);
    arc(    grid*3, grid*4, d1, d2, start, eA + TAU   , OPEN, detail);
    arc(    grid*4, grid*1, d1, d2, start, eA + PI/2  , PIE , detail);
    arc(    grid*4, grid*2, d1, d2, start, eA + PI    , PIE , detail);
    arc(    grid*4, grid*3, d1, d2, start, eA + PI*3/2, PIE , detail); 
    arc(    grid*4, grid*4, d1, d2, start, eA + TAU   , PIE , detail);
}


function mouseWheel(event) {
  wheelPos += event.delta > 0 ? -1 : 1;
}


function shapeScroll(){
  return wheelPos % 40 + 3 ;
}
```


Siphonophora:
So, I guess the two webgl examples I added to the are taking longer than expect (50-60ms). Could that be the issue? I can't tell from the build log. (all of these tools are new to me).

meiamsome:
It's best to run `grunt test` before you commit to check everything is good - two examples are failing, yes. To check this it's best to run `grunt test --keepalive` and open the link it gives you in a browser, then browse to `test/test-reference.html` (Or `test/test.html` if it's normal tests that are failing). Your browser will run all the tests - note that there are some WEBGL failures, because these aren't tested in Travis correctly and so aren't included there.

The base WEBGL issues you're having are filed as an issue: #2200 
I don't know enough about WEBGL, so perhaps you can help fix these?

You need to not include your examples in Travis somehow - I'm not sure the best way to do that, perhaps we can add a `data-feature="WEBGL"` that will not run it to the code element in the docs - some input on that would be great, I am happy to implement that into the testing infrastructure if it's deemed appropriate

Siphonophora:
@meiamsome thanks for the input. I figured out a few things. 

1. If I remove webgl (and rotateY) they do not trigger the message below from grunt. When I look in test-reference, these are only flagged with a time in ms. I don't get a red error message and they aren't on the failures list. So whatever is happening, there is some internal inconsistency there. I see four other tests AccelerationX, Y, Z. and LoadJSON all are behaving the same. I would tend to agree that the item 1 here is some issue with the configuration of the test. To deal with that I will remove the examples from the documentation to get the build to pass. 
```
1) src/core/2d_primitives.js arc documentation example #6 works:
_initContext@http://localhost:9001/lib/p5.js:68893:24
  RendererGL@http://localhost:9001/lib/p5.js:68808:20
  createCanvas@http://localhost:9001/lib/p5.js:49470:53 (truncated)
```
2. The ellipse webgl example runs in the reference. The arc doesn't run and on the console says arc isn't a function. Identical code to the test works runs on the browser, but **does not rotate unless I include call background()** which is really confusing. This was true with other primitives, like box which I didn't alter.

meiamsome:
@Siphonophora

To be clear here, running `grunt test` will cause the tests to be run inside PhantomJS which (Apart from being discontinued - #2174) has no support for WEBGL. In order for the WEBGL tests to not show up as failures, they are disabled based on browser compatibility: https://github.com/processing/p5.js/blob/cc4a489f3d974fc4ad78d1443c69b448de3d3f31/test/test-reference.html#L51-L55

This doesn't take account of WEBGL examples in the normal documentation - which is the issue you're having here. I would suggest either marking up your examples to not run or removing them for now while this is 'fixed'. Fixing this is a complicated problem, however, hence my suggestion above to add a `data-feature` tag to the example.

I will expand on the specific broken WEBGL tests in #2200 because as far as I can tell these are definitely bugs that need fixing, but for now here is what I have for your arc documentation in browser:

```
TypeError: this.curFillShader.attributes.aNormal is undefined
[71]</p5.RendererGL.prototype.createBuffers@http://0.0.0.0:9001/lib/p5.js:68663:3
[75]</p5.RendererGL.prototype.arc@http://0.0.0.0:9001/lib/p5.js:71019:5
[17]</p5.prototype.arc@http://0.0.0.0:9001/lib/p5.js:43133:3
draw@test-reference.html line 77 > eval:12:1
```

Are you using chrome to test? Just showing timeouts sounds similar to the problem in #2201 which I will look at soon. If so, I suggest testing with Firefox for the moment - it gives you tracebacks and takes less time.

meiamsome:
It appears chrome is behaving a bit better for me now (Weird) and gives me the following error:

```
TypeError: Cannot read property 'location' of undefined
    at p5.RendererGL.createBuffers (http://0.0.0.0:9001/lib/p5.js:68663:73)
    at p5.RendererGL.arc (http://0.0.0.0:9001/lib/p5.js:71019:10)
    at p5.arc (http://0.0.0.0:9001/lib/p5.js:43133:18)
    at p5.draw (eval at <anonymous> (test-reference.html:28:11), <anonymous>:12:1)
    at p.draw (test-reference.html:166:27)
    at p5.redraw (http://0.0.0.0:9001/lib/p5.js:50108:7)
    at p5.<anonymous> (http://0.0.0.0:9001/lib/p5.js:44778:12)
    at p5.<anonymous> (http://0.0.0.0:9001/lib/p5.js:44673:12)
    at new p5 (http://0.0.0.0:9001/lib/p5.js:44964:12)
    at test-reference.html:151:20
```

Both seem to imply this line is the problem: https://github.com/processing/p5.js/blob/cc4a489f3d974fc4ad78d1443c69b448de3d3f31/src/webgl/p5.RendererGL.Retained.js#L74-L75
But as stated, I don't know how to fix it.

kjhollen:
ahhh I thought I had gone through all the webgl pull requests a while ago but somehow missed this one! let me start a review now—thanks for submitting this and being patient :)

src/core/p5.Renderer2D.js
```diff
@@ -389,7 +388,7 @@ p5.Renderer2D.prototype._acuteArcToBezier =
 p5.Renderer2D.prototype.arc =
   function(x, y, w, h, start, stop, mode) {
   var ctx = this.drawingContext;
-  var vals = canvas.arcModeAdjust(x, y, w, h, this._ellipseMode);
+  var vals = {x: x, y: y, w: w, h: h};
```
kjhollen: Maybe we can lose vals here? I know it requires updating more lines of code, but I don't think anything is gained by putting these parameters into an object now that the core p5 module does the adjustment of the coordinates.

src/webgl/primitives.js
```diff
@@ -632,9 +630,103 @@ p5.prototype.torus = function(){
...
+        this.strokeIndices.push([this.vertices.length - 1, 1]);
+      } else if (mode === 'open'){
+        this.faces.push([0, this.vertices.length - 1, 1]);
+      } else if (mode === 'chord'){
+        this.faces.push([0, this.vertices.length - 1, 1]);
+        this.strokeIndices.push([this.vertices.length - 1, 1]);
+      } else if (mode === 'pie'){
+        this.strokeIndices.push([this.vertices.length - 1, 0]);
+        this.strokeIndices.push([0, 1]);
+      } else {console.log('Error, unknown mode');}
```
kjhollen: I see this error (unknown mode) when I run an example like this:

```
function setup(){
}

function draw(){
  background(220);
  arc(width/4, -height/4, 50, 50, PI/3, 2 * PI/3);
}
```
Probably should be a default type for arc?

src/webgl/primitives.js
```diff
@@ -667,64 +759,21 @@ p5.RendererGL.prototype.triangle = function
...
-      console.log('Cannot stroke ellipse with more'+
-        ' than 24 detailX or 16 detailY');
-    }
-
-    this.createBuffers(gId, ellipseGeom);
-  }
-  this.drawBuffers(gId);
+p5.RendererGL.prototype.ellipse = function(args){
+  //Note, the x and y adjustments here are because ellipses and
+  //arc get different treatment in core/canvas.js
```
kjhollen: Could this explanation be elaborated upon here for future developers?

test/unit/core/2d_primitives.js
```diff
@@ -38,16 +50,28 @@ suite('2D Primitives', function() {
...
+    test('no friendly-err-msg, 2D', function() {
+      assert.doesNotThrow(function() {
+          myp5.ellipse(0, 0, 100, 100);
+        },
+        Error, 'got unwanted exception');
+    });
+    test('no friendly-err-msg, 3D', function() {
       assert.doesNotThrow(function() {
-          myp5.ellipse(0, 0, 100);
+          myp5.ellipse(0, 0, 100, 100, 24);
```
kjhollen: I know some earlier comments were about GL test issues, but everything is passing for me. :)

Siphonophora:
@kjhollen Thanks for the feedback. I should have time next week to address your feedback. 

I think a few of these come back to the way core/canvas.js works. Arc coordinates are handled differently from Ellipse mode and Rect mode (it must come from the original 2d implementation of Arc, which I mimicked in webgl). Probably its easier to bring the arc coordinates into alignment with the other two. I wouldn't need the change in src/core/p5.Renderer2D.js. I can also make the ellipse in webgl less confusing.

test/unit/core/2d_primitives.js
Siphonophora: Sounds good. I see the same failures as before, but don't think they are related to my code.

src/core/p5.Renderer2D.js
Siphonophora: Fixed as part of removing arcModeAdjust as its own function

src/webgl/primitives.js
Siphonophora: I made 'OPEN' the default.

src/webgl/primitives.js
Siphonophora: Fixed as part of removing arcModeAdjust as its own function. This adjustment doesn't happen any longer.

Siphonophora:
@kjhollen Please take a look at the changes. I believe this addresses everything from your review. I included a new manual test which I found very helpful in testing the edits. I can drop it from this PR if that makes sense. 

It does seem that I can't resolve the conflicts myself (a permission issue I guess). In renderer 2d and webgl/primitives I would take my code. In the unit tests, I was definitely guessing what to add and am not sure I did that right in the first place. I was trying to make sure the new parameters in Arc for 3d mode were covered in the test. 

Let me know if you need anything else from me.

src/core/2d_primitives.js
```diff
@@ -47,30 +45,46 @@ require('./error_helpers');
...
+ * arc(45, 50, 70, 70, HALF_PI, PI + HALF_PI, CHORD);
+ * arc(55, 50, 70, 35, PI + HALF_PI, HALF_PI, CHORD);
+ * </code>
+ * </div>
  * @alt
- *shattered outline of an ellipse with a quarter of a white circle bottom-right.
- *white ellipse with black outline with top right missing.
- *white ellipse with top right missing with black outline around shape.
- *white ellipse with top right quarter missing with black outline around the shape.
+ * shattered outline of an ellipse with a quarter of a white circle bottom-right.
```
kjhollen: do these accessibility descriptions need to be updated to reflect the changes in the examples?

src/core/2d_primitives.js
```diff
@@ -137,19 +155,33 @@ p5.prototype.arc = function(x, y, w, h, start, stop, mode) {
...
  * </div>
  *
+ * <div>
+ * <code>
+ * //An ellipse does not need to be circular
+ * ellipse(50, 50, 55, 25);
+ * </code>
+ * </div>
  * @alt
  *white ellipse with black outline in middle-right of canvas that is 55x55.
```
kjhollen: accessibility description needed for new example

src/core/p5.Renderer2D.js
```diff
@@ -389,13 +388,15 @@ p5.Renderer2D.prototype._acuteArcToBezier =
...
-  var vals = canvas.arcModeAdjust(x, y, w, h, this._ellipseMode);
-  var rx = vals.w / 2.0;
-  var ry = vals.h / 2.0;
+  var rx = w / 2.0;
+  var ry = h / 2.0;
   var epsilon = 0.00001;  // Smallest visible angle on displays up to 4K.
   var arcToDraw = 0;
   var curves = [];
 
+  x += rx; //Adjustments to match 2d ellipse.
```
kjhollen: comment says we're matching 2d ellipse but this is in the 2d renderer?

src/webgl/primitives.js
```diff
@@ -667,64 +766,21 @@ p5.RendererGL.prototype.triangle = function
...
-    }
-
-    this.createBuffers(gId, ellipseGeom);
-  }
-  this.drawBuffers(gId);
+p5.RendererGL.prototype.ellipse = function(args){
+  //Pass arguments X, Y, Width, Height, and detail to Arc
+  //Supply additional parameters directing Arc to draw a full ellipse
+  this.arc(args[0], args[1] ,args[2], args[3],
+           0, Math.PI * 2, 'ellipse', args[4]);
```
kjhollen: great. can 'ellipse' be a constant? ok to add this in a future commit, but let's open an issue so we don't forget!

test/manual-test-examples/webgl/geometry/2DPrimitives/framesketch.js
```diff
@@ -0,0 +1,265 @@
...
+//-Include an informative title in testinfo.html(''). If any differences are 
+// expected between the 2D and 3D tests, include a note here. 
+//
+//-Unless you are explictly testing stroke or fill, do not set them in the test
+
+
+
+var params
+var n
+var testinfo
```
kjhollen: missing ; on 3 lines above

test/manual-test-examples/webgl/geometry/2DPrimitives/framesketch.js
```diff
@@ -0,0 +1,265 @@
...
+var params
+var n
+var testinfo 
+
+
+function setup(){
+  params = getURLParams();
+  if(params.mode == '3d'){
+    createCanvas(400,400,WEBGL);
+    console.log('3d loaded')
```
kjhollen: missing ; here and in a number of places below

test/manual-test-examples/webgl/geometry/2DPrimitives/framesketch.js
```diff
@@ -0,0 +1,265 @@
+//Framesketch.js expects to be called with parameters specifying whether
+//to run the page in 2d or 3d, the test number, and whether or not
+//to use fill or stroke.
```
kjhollen: this looks like a great addition. thank you!

test/manual-test-examples/webgl/geometry/2DPrimitives/framesketch.js
```diff
@@ -0,0 +1,265 @@
...
+    smooth();
+    ellipse(-100, -100, 200, 200);
+    noSmooth();
+    ellipse( 100,  100, 200, 200);
+  }
+}
+
+function showThisTest(){
+  //Helper function so the tests can be reordered easily.
+  n++
```
kjhollen: this is clever but I think it would be good if you could provide a deeper comment at the beginning of draw, that showThisTest will increment the number n and test if that matches the current test parameter, resulting in only one test being run but checking them all. it's a little confusing that all the conditionals are the same...

kjhollen:
@Siphonophora are you still interested in working on this? we've made some updates to the gl core that are in conflict here, but I'd still like to find a way to use what you've done! If you're not interested, is it ok if I use your branch as a starting point for updates to get it merged in? thanks!

lmccart:
hi @Siphonophora I'm going to close this for organizational purposes. please feel free to reopen if you make the requested changes or have further thoughts, and we can review. thanks! @kjhollen since there's been no activity for a while on this, I think it would be ok to use this branch as a starting point to merge the updates you're working on.

PR #2235, opened 2017-10-03
Aligned manual tests for webgl/origin

Siphonophora:
Aligned tests meant to be direct comparisons. Origin tests were inconsistent. Apparently written expecting quad to take 12 arguments, not 8. Corrected ordering of vertices in quad in topleft_origin


kjhollen:
Sorry I missed this one somehow! This looks good to me—thanks so much for doing this!

PR #2237, opened 2017-10-03
Added p5.MediaElement.stop example

bomoko:
Partially addresses #1954 - adds example for p5.MediaElement.stop.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1911,7 +1911,7 @@
    *     if(sampleIsPlaying) {
    *       //if the sample is currently playing
    *       //calling the stop() function on
-   *       //out p5.MediaElement will stop
+   *       //our p5.MediaElement will stop
    *       //it and reset its current
    *       //time to 0 (i.e. it will start
    *       //at the beginning the next time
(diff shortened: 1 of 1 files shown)


lib/addons/p5.dom.js
```diff
@@ -1877,6 +1877,66 @@
...
+   *     }
+   *   }
+   * }
+   *
+   * //simply set the message and background shade of grey
+   * //on the canvas. Included to make the example
+   * //above more readable.
+   * function setMessageAndBackground(message, backgroundColor) {
+   *   background(backgroundColor);
+   *   textAlign(CENTER);
```
lmccart: couldn't these two lines get set in setup() then the text() line could replace the setMessageAndBackground() call in the mousClicked() function

bomoko: Hi @lmccart 

I had the setMessageAndBackground() function in order to try simplify the repeated setting of the background and text. To be honest, I was worried about it (which is why it had that comment preamble about readability).

What I didn't consider was just not changing the background from light to dark gray. I've done that now and have pulled out the setMessageAndBackground() function.
I think it's better as an example now.

Thanks so much for the feedback!

PR #2245, opened 2017-10-06
a couple of docs fixes

Spongman:
specifically:
- img() & copy() can take a p5.Graphics as the 1st parameter.
- the docs for some of the new WEBGL stuff is backwards.


lmccart:
thanks!

PR #2255, opened 2017-10-10
 fix pointLIght & directionLight docs

### LINKED ISSUE #2246: p5.prototype.directionalLight takes unused parameter
p5.prototype.directionalLight = function(v1, v2, v3, a, x, y, z);

but the parameter `a` is never used.

also, the docs for this method should be split into separate overrides to indicate the valid permutations of its parameters. for example, parameter `x` is required, but `v2` and `v3` are optional... there's no indication (without looking at the code) how it resolves this.

Spongman:
fixes #2246

i created the unambiguous overrides for these two functions, and removed the unused 'a' parameter.


sakshamsaxena:
The diff looks way too large and it somehow violated Travis build. @Spongman, could you try this again ? Keep these points in mind : 
1. Take a rebase/pull from upstream and then start your changes.
2. Do run grunt before committing 
Thanks! 😄

Spongman:
ok, i fixed the whitespace bugs.

Spongman:
i'm going to try again. it seems my magical `git` incantations were the wrong shade of green.

sakshamsaxena:
That's okay! Thanks for considering ! :-) Good luck!

PR #2257, opened 2017-10-11
fix directionalLight & pointLIght

Spongman:
ok, 2nd attempt. this one fixes directionalLight/pointLight with appropriate overloads. also removes unused `a` parameter.


Spongman:
ugh, yes, sorry about that. i didn't realize github would automatically include that push in this PR. I'm loathe to do anything about it, though. git & I really don't get along these days. (actually, ever)

kjhollen:
ah, thanks for clarifying! yes, git can be a beast—it took me over a year to get the hang of it. are you comfortable using git on the command line? one thing that might work well is to start a new branch based on master, then try:

`git cherry-pick ddf9188a9916f3d4d161af7092d0c6c714416081`

which should pick up only the first commit.

regarding the lighting code, the only suggestion I have is to keep v1, v2, v3 as the variable names for the first 3 args since they can be either RGB or HSB (other parts of the code do this as well, for example, in fill())

Spongman:
please see the new PR here: https://github.com/processing/p5.js/pull/2300

PR #2265, opened 2017-10-14
Add example for p5.Image

### LINKED ISSUE #1988: A better example for dealing with pixels array on different pixelDensity would be great!
Hi there,
I think the reference at p5js.org is lacking of sufficient info/examples for dealing with pixels array on different pixelDenisty displays. In fact I find the following example on the [p5.Image](https://p5js.org/reference/#/p5.Image) reference page quite obscure.

`var d = pixelDensity;
for (var i = 0; i < d; i++) {
  for (var j = 0; j < d; j++) {
    // loop over
    idx = 4 * ((y * d + j) * width * d + (x * d + i));
    pixels[idx] = r;
    pixels[idx+1] = g;
    pixels[idx+2] = b;
    pixels[idx+3] = a;
  }
}`

I would like to have a reference page at p5js.org that show an example on how I can deal with each pixel, like I normally do on "standard density displays" like this:

`function setup(){
  createCanvas(600, 600);
  pixelDensity(1);
  background(0);
  loadPixels();
  for (var y = 0; y < height; y++) {
    for (var x = 0; x < width; x++) {
      // loop over
      var index = (x + (y * width)) * 4;
      pixels[index+0] = random(255);
      pixels[index+1] = random(255);
      pixels[index+2] = random(255);
      pixels[index+3] = 255;
    }
  }
  updatePixels();
}`

I think this would help many not super expert programmers better understanding the topic.

Thanks!

Zalastax:
Closes #1988 

I don't think that the original example [1] is correct. I checked the array size on my phone with pixelDensity() === 3 and it's not larger than on my PC. If you can confirm it to be valid / invalid I can update the PR accordingly.

[1]:
```javascript
var d = pixelDensity();
for (var i = 0; i < d; i++) {
  for (var j = 0; j < d; j++) {
    // loop over
    idx = 4 * ((y * d + j) * width * d + (x * d + i));
    pixels[idx] = r;
    pixels[idx+1] = g;
    pixels[idx+2] = b;
    pixels[idx+3] = a;
  }
}
```

### DIFF
--- a/src/image/p5.Image.js
+++ b/src/image/p5.Image.js
@@ -35,6 +35,52 @@ var Filters = require('./filters');
  * <br><br>
  * Before using the pixels[] array, be sure to use the loadPixels() method on
  * the image to make sure that the pixel data is properly loaded.
+ * @example
+ * <div><code>
+ *
+ * function setup() {
+ *   var img = createImage(100, 100); // same as new p5.Image(100, 100);
+ *   img.loadPixels();
+ *   createCanvas(100, 100);
+ *   background(0);
+ *
+ *   // helper for writing color to array
+ *   function writeColor(image, x, y, red, green, blue, alpha) {
+ *     var index = (x + (y * width)) * 4;
+ *     image.pixels[index]   = red;
+ *     image.pixels[index+1] = green;
+ *     image.pixels[index+2] = blue;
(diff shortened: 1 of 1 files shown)


lmccart:
thanks!

Issue #105, opened 2014-02-02
code style guide


lmccart:

lmccart:
Reopening this issue after discussion about `x /= y` vs `x = x / y`. We have a link to [idiomatic.js](https://github.com/rwaldron/idiomatic.js) in the developer doc, but I think we could use more customization or maybe just a shorter, clearer cheat sheet that accompanies it. The p5.js project has a lot of contributors that may be newer to JS or developing for this sort of project, so anything we can do to be clear about the style will be helpful. At the very least it'd be nice to expand the section in the dev doc where we include the link and explain a little what it is and why it matters.

shiffman:
Obviously things are drastically different given JS vs. Java, but Processing just finally published a code style guidelines in case this is a helpful reference!

https://github.com/processing/processing/wiki/Style-Guidelines

workergnome:
Oh, that's very helpful, bot for content and tone.  Thanks, @shiffman!

@lmccart, do you know if travis is running jshint?  codifying some of this there is also helpful.

lmccart:
yes, travis runs `grunt` which runs jshint, I believe.

workergnome:
Exemplary.

dhowe:
Want to +1 @taseenb's comment re: [editorconfig](http://editorconfig.org/). One difficult thing contributing to a new oss project is setting up an editor according to the required coding style, and otherwise getting bunches of linter errors on each change. With something like editorconfig, we can reduce this significantly (it works with sublime, atom, vim, etc)

kadamwhite:
I'd like to try to codify some of the existing stylistic practices via [JSCS](http://jscs.info), and integrate that into the Gruntfile—JSCS has emerged as the most-customizable tool for JS style checking, and is designed to complement JSHint (which has deprecated all stylistic options). Are there any objections to my giving this a swing over the holiday weekend?

workergnome:
I think that's a good idea, and could (should?) be done in tandem with the **editorconfig** file?  Also, it might be helpful to have it in prose, as well.

Another thought is that someone else (me, perhaps, or @lmccart, or anyone else?) could take a run through the pull request and try and understand what the JSCS is doing and say it in words, just so there's another set of eyes looking hard at the rules and trying to understand them...

lmccart:
addressed with https://github.com/processing/p5.js/pull/2269

dhowe:
[...]

PR #2269, opened 2017-10-16
Replace jscs and jshint with eslint

Zalastax:
jscs has been deprecated in favor of eslint which can handle both the tasks of jscs and jshint.
Rules were migrated using polyjuice and then slightly hand-tuned to cause less errors.
The style problems that were left after that has been fixed by hand.

I hope the style changes are not a problem. They are mainly cleaning up weird styles (e.g. line-break at column 80) and I've tried to keep the changes minimal. In the future I'd recommend enabling more style rules and adding an auto-formatter like [prettier](https://github.com/prettier/prettier). You can ping me if that sounds interesting.


lmccart:
this is great thanks @Zalastax. I think more style rules and auto-formatting would help a lot. if you want to look into this, that'd be awesome.

dhowe:
> is there any readable documentation on coding style for the project? i was unable to quickly pull it up when asked, and would be pretty useful for new contributors (imho)... [from #105]

PR #2282, opened 2017-10-21
Add parenthesis to pixelDensity call (docs change)

DonKarlssonSan:
Otherwise d will become undefined and the code will fail silently.


Zalastax:
Just for reference, I also fix this in #2265 😃

lmccart:
thank you!

PR #2284, opened 2017-10-22
Adjust stroke weight in ortho example

### LINKED ISSUE #2283: strokeWeight too thick in ortho camera
The [`ortho()`](https://p5js.org/reference/#/p5/ortho) example page shows the following:

ortho

Zalastax:
It was impossible to understand the example when the stroke weight was at the default setting (1).
Closes #2283

### DIFF
--- a/src/webgl/camera.js
+++ b/src/webgl/camera.js
@@ -251,6 +251,7 @@ p5.RendererGL.prototype.perspective = function() {
  * function draw(){
  *  background(200);
  *  orbitControl();
+ *  strokeWeight(0.1);
  *  for(var i = -1; i < 2; i++){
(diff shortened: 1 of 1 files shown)


mlarghydracept:
I say this should be merged but the issue shouldn't be closed. As I said over in that issue, this is a great short term solution but the issue should stay up until the root cause is addressed.

Zalastax:
@mlarghydracept I can rebase + change my comment so that the issue is not automatically closed if you want. Alternatively we can open a new issue about changing the orthographic rendering. A new issue is cleaner in my opinion (new title, the original issue is actually resolved with this fix).

mlarghydracept:
@Zalastax You are completely right! I'll just make a new issue now. Thanks!

PR #2301, opened 2017-10-24
Undocumented warning

Spongman:
this PR add an 'unnamed public member' warning to the yuidoc preprocessor.

it also marks existing 'documented' private members with an `@private` attribute to hide the warning for those members. those members are now excluded from the data.json file.

there is an outstanding PR (https://github.com/processing/p5.js/pull/2304) to fix the errors in the sound docs.


lmccart:
fixed with https://github.com/processing/p5.js/pull/2304

Spongman:
all test pass, and with the new p5.sound stuff, no undocumented public member warnings are shown.

lmccart:
this is super helpful!

Issue #785, opened 2015-07-03
Properly stub out methods in minified files


kadamwhite:
We attempt to suppress two methods, `_friendlyFileLoadError` and `_validateParameters`, when we build the minified file. The manner in which we're stubbing these methods doesn't actually have an effect in some circumstances, specifically when we require P5 via AMD or CommonJS require calls, vs depending on it as a global.

One option is to move these methods to their own module, then have browserify create two builds, one for the unminified source that works as-is and another that stubs out these modules, and only minify the latter build. This way they'll be properly disabled no matter how p5 is called, and may save us a byte or two in the process.

lmccart:
addressed with https://github.com/processing/p5.js/pull/2302

PR #2302, opened 2017-10-25
improved parameter validation

### LINKED ISSUE #1329: Use IS_MINIFIED to disable friendly debugger in minified builds

Spongman:
this PR improves the parameter validation a little:
- it now handles array types (eg. Number[]) and checks the types of the array's elements.
- it checks object types (eg. p5.Color)
- it checks Constant values (eg. HSB|RGB, etc...)
- it checks Integer values are integral. (and updated doc comments where appropriate)
- it pre-calculates more of the work needed to do the validation, and caches the `lookupParamDoc` results. performance is significantly improved in debug.
- validation has been added to more p5 methods.
- the validation code & data.json is excluded from IS_MINIFIED builds, reducing the un-deflated size from 1,100KB to 389KB.

fixes #1329, #785


Spongman:
it looks like everything changed in error_helpers.js but that because I had to split apart the parameter validation (which is conditional on IS_MINIFIED) and the top-level warning stuff (which is not). also, jslint is fussy about the ordering of functions inside `{}` blocks (but not elsewhere, for some reason).

meiamsome:
This looks like good improvements. It may be worth cherry picking 68de25a and 464c2f5 from #2177 into this PR & then that PR can be rebased onto this.

Spongman:
ok, this does the right thing with the argument count error. i fixed up some docs to accomodate.

i also added the location of the error in the message, which you can click on to navigate in chrome's devtools .

meiamsome:
Ah, that's good that 2bb7aa removed the need for that extra parameter to `createCanvas`! It's a shame that Travis has your account blacklisted and won't run the tests on this PR - perhaps you can give them a shout to un-blacklist you?

Spongman:
travis? blacklisted? i have no idea what that is.

meiamsome:
It's the thing that's supposed to automatically run tests on PRs to check they run fine.

Your PRs are marked as "abuse detected" here: https://travis-ci.org/processing/p5.js/requests
I'm not sure why, though, https://github.com/travis-ci/travis-ci/issues/6942 says you should just email them to get it cleared up!

Eitherway, I'm just confirming this passes tests and afaics is ready for merge!

Spongman:
I don't know what I have to do to get the abuse thing cleared. I have tried emailing them, filing an issue and contacting one of the devs directly, but all I'm getting is static. It seems like their process for handling false positives is broken. Oh well...

Spongman:
i'm sorry if I messed up the merge, it seems like i have a love/hate relationship with git: it just loves to hate me.

lmccart:
this is great, sorry it took so long to merge

PR #2307, opened 2017-10-26
docs: disambiguate overloads

Spongman:
if you look at the docs for `p5.Vector.add()` (https://p5js.org/reference/#/p5.Vector/add), you'll see that for the non-static overload, it's apparently ok to call it like this:

    var r = v.add([a,b,c], y, z);

because `x` can be a `{Number|p5.Vector|Array}`.

this PR splits those into separate overloads (all except `save()` !) so this ambiguity doesn't occur.


lmccart:
👍

Issue #2292, opened 2017-10-23
byte doesn't work for strings


ronraisch:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [x] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 
the function byte() should return the byte reprsention of a number, boolean or string but it doesnt for strings.

meiamsome:
Looks like this is caused by not checking if `nn` is NaN in the if condition on https://github.com/processing/p5.js/blob/master/src/utilities/conversion.js#L164 
The doc example doesn't seem to be showing up for me either for whatever reason - I'll have a look into that

Jared-Sprague:
Hi @ronraisch,  I think the confusion is that the documentation when the docs say byte representation of a string, it means string representation of a number e.g. "10".  A string like "Hello" will return NaN because byte() wraps the javascript function parseInt() Which will try to parse an int out of a string if it can,  but will return NaN for any strings that it can't parse.   Check out the MDN doc for [parseInt()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt).

I'll make a pull request for the docs to make this more explicit that it means a string representation of an number.

lmccart:
documentation confusion addressed with https://github.com/processing/p5.js/pull/2308

PR #2308, opened 2017-10-26
Documentation update for #2292

Jared-Sprague:
Make byte() documentation explicit that it can only convert a string representation of a number.  Based on feedback from issue #2292


lmccart:
thanks!

Issue #2243, opened 2017-10-05
Grunt yui's local reference site doesn't use local p5.min.js


Siphonophora:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: master branch 753680c6df5e579f49c4be3e2ac289466a319a07
- Web browser and version: Firefox/Chrome
- Operating System: Windows
- Steps to reproduce this:
If you go to p5.js\docs and delete reference. Then run grunt yui to rebuild the references. This creates the index.html file with an absolute reference to get p5.min.js off p5js.org instead of the local p5 which is what I assumed it would do since the text of inline docs are ending up in the reference when I view it on a local server. Either this is a bug, or the documentation on the inline documentation needs to be clarified. This was causing difficulties in verifying why examples for the arc function I added in #2229 couldn't run in the documentation.

```html
<script src='https://p5js.org/assets/js/p5.min.js'></script>
```
I assume it should be as shown below, but couldn't find where this is configured. 
```html
<script src='../../../lib/p5.min.js'></script>
```

meiamsome:
Yeah, I was confused by this as well. As far as I can see it's mostly because the docs are copied verbatim to the p5js-website repo where the assets system is differently structured and the link would have to be altered all the time. Note that this page also references scripts such as `assets/js/render.js` that aren't included in p5's repo currently.

Siphonophora:
If it is the desired behavior, then maybe getting the the local server to rewrite the URL would be better?

PR #2314, opened 2017-10-26
grunt/connect: rewrite requests to /lib/*.js

Spongman:
this add a redirect to the connect grunt task that allows the docs to use the locally-built p5.min.js (+ addons) when run locally on port 9001.

fixes https://github.com/processing/p5.js/issues/2243


lmccart:
awesome <3

Spongman:
Oh, I forgot to add a change to the handlebars template (making the p5*.js URLs relative) that actually makes this useful. However, after looking at it, I'm not sure whether or not this is same to do. Is the template used for something other than the docs/reference stuff?

Issue #1990, opened 2017-06-07
Feature request/discussion: Simpler for loop syntax


kristianpedersen:
I still remember the disappointment and confusion when I was introduced to for loops. 

Printing out a range of numbers, or repeating something n times, should be simpler. Ruby has `1000.times`, and lodash.js has `_.times(1000)`, for example.

# Someone somewhere meets the for loop

Student:
"Can I make a program that counts to 1000? ツ

Teacher:
"Yes, you can! Counting is one of the things computers do best! First you write "for", just because that's the word they use. Then you write parenthesis and declare a variable called i, which will be your counter program's starting point. Then you write a condition which will keep the loop running as long as it evaluates to true. In this case, the loop should run while i is less than or equal to your finishing number. Ah and don't forget to increment the variable. Yes, you have to do it in this order. You thought you knew addition? Well, now you can write two pluses! Or you also say i = i + 1, or i += 1.":

```
for (var i = 1; i <= 1000; i++) {
  print(i);
}
```

Student: 
ಠ_ಠ

Of course, this is a terrible explanation to give to someone who has never programmed before. Regardless, the code above is quite a handful to take in for a beginner, and I think it's kind of awkard to be honest.

# Simpler approaches

What I would like to see is something along these lines:

`count(1, 1000, print(i))` // 1..1000
or 
`count(1000, print(i))` // 0..999

I think it makes sense to have it default to 0, do you agree? Counting from 0 instead of 1 trips up beginners no matter what, so at least it would be nice with a more readable syntax.

Something very close to what I want can already be done in lodash.js, although it seems you have to use -functions- for it to work:

`_.times(1000, i => print(i+1))`

`_.times(1000, () => ellipse(random(width), random(height), random(100), random(100)))`

For arrays that already have contents, forEach and map (native JavaScript, not p5) are nice:

`blobArray.forEach(blob => blob.update())`
is much more readable than
`for(var i = 0; i < blobArray.length; i++) {blobArray[i].update()}`

Next up is the this dot, never forget the this dot.

# People who pick up p5.js just wanna have fun

Of course, every programmer should know about for loops, but not everyone is going to be a programmer, and I think it would be cool if beginners could focus more on high-level ideas instead of awkward syntax. 

P5 users come from a very diverse set of ages, ideas, backgrounds, goals, skillsets and personalities - and they all have something to share. It's already a super exciting initiative, and I'm looking forward to seeing it continue to grow. Cheers!

[...]

PR #2316, opened 2017-10-27
Implementation of range()

meiamsome:
This is my suggestion for a `range()` implementation as discussed in #2305 & #1990 

Potential points of thought:

You cannot do the following:
```javascript
var iter = range(10);
iter.next(); // 0
iter.next(); // 1
```
You can do the following:
```javascript
var my_range = range(10);
for(var i of my_range) {...}
for(var i of my_range) {...}
```
and both iterate [0, 10)

I think this is how JS iterators are intended to operate, that is dissimilar from python iterators. It would be possible to make it only iterate once, but I am unsure if that would be better. I can see uses for both cases really and the way iterators are designed made me lean this way. Discussion is very welcome!

Oh, and the documentation is excluded from automatic testing because PhantomJS doesn't support `for..of` (See #2174)

### DIFF
--- a/src/utilities/iterators.js
+++ b/src/utilities/iterators.js
@@ -51,7 +51,7 @@ var p5 = require('../core/core');
  */
 p5.prototype.range = function(start, stop, step) {
   if(typeof start === 'undefined') {
-    throw new Error('Not enough parameters for range');
+    throw new TypeError('Not enough parameters for range');
   }
   if(typeof stop === 'undefined') {
     stop = start;
@@ -67,7 +67,7 @@ p5.prototype.range = function(start, stop, step) {
       var i = start;
       return {
         next: function() {
-          if(dir * (i - stop) >= 0) {
+          if(stop !== null && dir * (i - stop) >= 0) {
             return {
               done: true
             };
--- a/src/utilities/iterators.js
+++ b/src/utilities/iterators.js
@@ -1,0 +1,88 @@
+/**
+ * @module Utilities
+ * @submodule Iterators
+ * @for p5
+ * @requires core
+ */
+
+'use strict';
+
+var p5 = require('../core/core');
+
+
+/**
+ * Produces an iterator over the numbers in the range. The result is given by
+ * starting at start, and adding step repeatedly until the result has gone past
+ * stop.
+ * In the single parameter version, start is set to zero and step is set to one.
+ *
+ * @method range
+ * @param  {Number}  stop      Number to stop at
+ * @return {Iterator}          An iterator that returns the values sequentially
+ * @example
+ * <div>
+ * <code>
+ * for(var i of range(width)) {
+ *   line(i, 0, i, i);
+ * }
+ * </code>
+ * </div>
+ * <div>
+ * <code>
+ * var start = width / 4;
+ * var stop = 3 / 4 * width;
+ * for(var i of range(start, stop, 2)) {
+ *   line(i, start, i, i);
+ *   line(start, i, i, i);
+ * }
+ * </code>
+ * </div>
+ *
+ * @alt
+ * Triangle filling the top right corner of the canvas
+ * Square in the center of the canvas, with a striped effect
+ */
+/**
+ * @method range
+ * @param  {Number}  start       Number to start at (default 0)
+ * @param  {Number}  stop
+ * @param  {Number}  [step]      Amount to step by (default 1)
+ * @return {Iterator}
+ */
+p5.prototype.range = function(start, stop, step) {
+  if(typeof start === 'undefined') {
+    throw new Error('Not enough parameters for range');
+  }
+  if(typeof stop === 'undefined') {
+    stop = start;
+    start = 0;
+  }
+  if(typeof step === 'undefined') {
+    step = 1;
+  }
+  var dir = step < 0 ? -1 : 1;
+  if(Symbol && Symbol.iterator) {
+    var res = {};
+    res[Symbol.iterator] = function() {
+      var i = start;
+      return {
+        next: function() {
+          if(dir * (i - stop) >= 0) {
+            return {
+              done: true
+            };
+          } else {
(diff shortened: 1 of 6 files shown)


src/utilities/iterators.js
```diff
@@ -0,0 +1,88 @@
...
+              value: i - step,
+              done: false
+            };
+          }
+        }
+      };
+    };
+    return res;
+  } else {
+    throw new Error('range() unsuported');
```
tawez: In this case we can return just an array of values.
It is not memory efficient, but it is a kind of acceptable polyfill

src/utilities/iterators.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+  if(typeof stop === 'undefined') {
+    stop = start;
+    start = 0;
+  }
+  if(typeof step === 'undefined') {
+    step = 1;
+  }
+  var dir = step < 0 ? -1 : 1;
+  if(Symbol && Symbol.iterator) {
```
tawez: It is quite unusual that the sketch is started in one environment and during its life the environment is changing.
This condition can be checked once
```javascript
p5.prototype.range = ((isIteratorSupported) => {
  if (isIteratorSupported) {
    return iteratorBasedImplementation;
  } else {
    return polyfill;
  }
})(Symbol && Symbol.iterator);
```

src/utilities/iterators.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+/**
+ * @method range
+ * @param  {Number}  start       Number to start at (default 0)
+ * @param  {Number}  stop
+ * @param  {Number}  [step]      Amount to step by (default 1)
+ * @return {Iterator}
+ */
+p5.prototype.range = function(start, stop, step) {
+  if(typeof start === 'undefined') {
```
tawez: I think that comparison
```javascript
if (start === undefined) { ... }
```
is good enough.
There is one operation vs. two

src/utilities/iterators.js
meiamsome: I'd say this is a breaking polyfill - it would be very easy to crash browsers in ways that work fine as a generator:

`range(0, 1, 0);`
`range(10000000);`

etc. I'm not sure if it's good to have a polyfill. Could partial polyfill up to n numbers, though?

src/utilities/iterators.js
```diff
@@ -0,0 +1,88 @@
...
+    throw new Error('Not enough parameters for range');
```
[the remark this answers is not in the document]
meiamsome: Yep, meant to fix it but promptly forgot!

test/unit/utilities/iterators.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  suite('range', function() {
+    test('passes examples', function() {
+      // These are the examples from python's range documentation.
+      assert.deepEqual(
+        Array.from(myp5.range(10)),
+        [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]);
+      assert.deepEqual(
+        Array.from(myp5.range(1, 11)),
```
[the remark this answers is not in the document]
meiamsome: These were just some tests I pulled from Python's docs so that it was equivalent, I'm happy to go add in more - especially edge cases

src/utilities/iterators.js
meiamsome: Yep, I guess the name is declared so this is overly protective!

src/utilities/iterators.js
tawez: I just wanted to be kind to the user,
but OK, You have convinced me ;)

src/utilities/iterators.js
Spongman: actually, this breaks if someone declares a variable called `undefined` in scope.

it's usually preferrable to do

    typeof value === 'undefined'

or, for newer browsers

    value === void 0

also works.

src/utilities/iterators.js
meiamsome: Frankly, I'm not that convinced myself...

test/unit/utilities/iterators.js
```diff
@@ -0,0 +1,56 @@
...
+      assert.deepEqual(
+        Array.from(myp5.range(0, -10, -1)),
+        [0, -1, -2, -3, -4, -5, -6, -7, -8, -9]);
+      assert.deepEqual(
+        Array.from(myp5.range(0)),
+        []);
+      assert.deepEqual(
+        Array.from(myp5.range(1, 0)),
+        []);
+    });
```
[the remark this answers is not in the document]
meiamsome: What is the expected behaviour of these? Should we expect there to be no result?

meiamsome:
So, as it turns out none of these tests work on Phantom because it doesn't support `Symbol` - they weren't showing up because I used a `for...of` loop in the tests and so the test file never loaded because it was a syntax error.

Looks like this is blocked on #2174 unless we just ignore iterators not loading on Phantom.js and test in browsers exclusively

meiamsome:
I am not sure why Travis failed that build.

test/unit/utilities/iterators.js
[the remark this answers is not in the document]
tawez: I we are basing on Python implementation, then both cases should return `[]`.

Zalastax:
I've fixed the conflicts of the PR in https://github.com/Zalastax/p5.js/tree/feature/iterators/range. Let me know if you need any more help with making this PR mergeable again. Thanks for your time!

My fork is just a `pull upstream master --rebase` with `npm run lint:fix` for each commit.

kjhollen:
I'm trying to resolve as many unmerged PRs as I can in the next couple of weeks. For this one, I'm noticing that the discussions in #2305 and #1990 seem to favor adding features like this in an external library building on p5, so I'm going to close this for now. Please re-open if you'd like to continue the discussion. Thanks!

PR #2320, opened 2017-10-29
getCanvasImage() : new image.js function for screenshots

ogoossens:
**Hi,** 

I've added a new function **getCanvasImage()** to image.js that **returns a P5.image object of the current canvas**. 

This can be considered as a **screenshot** of the current canvas and be **used in any way P5.image** would be used. I needed this functionality and was surprised this wasn't implemented _natively_ since I believe it's quite handy.

I didn't want the users to save the image locally - so saveCanvas() wasn't really helping and after longer search I found a single workaround from 2015 with saveFrames() and custom callback function that I found _difficult_ and _unintuitive_ to implement. So I went through the saveCanvas() function, found, extracted and edited a bit this one single line of code. 

Should you decide to merge it - I will prepare the documentation and working examples for the P5.js webpage.

**Thank you**

**Oliver**

### DIFF
--- a/src/image/image.js
+++ b/src/image/image.js
@@ -343,24 +343,40 @@ p5.prototype._makeFrame = function(filename, extension, _cnv) {
 
 /**
  * Returns the current canvas as a loaded p5.Image. This allows an assignment
- * or immediate use in the users sketch as a canvas screenshot.
+ * or immediate use in the sketch as a canvas screenshot.
  *
  * @method getCanvasImage
- * @param none
- *
  * @example
- * // This example shows how a user can directly create a screenshot of his
- * // canvas and have it saved for future use or use immediately
- * draw() {
- *   // .. other code
- *   var screenshot = getCanvasImage();
- *   image(screenshot, 10, 10, 50, 50);
+ * <div>
+ * <code>
+ * var screenshot;
+ *
+ * function draw() {
+ *   background(237, 34, 93);
+ *
+ *   // Sample circle
+ *   strokeWeight(10);
+ *   ellipse(50, 50, 80, 80);
+ *
+ *   // Draw the screenshot if the screenshot has been made
+ *   if(!(screenshot == null)) {
+ *     image(screenshot, 5, 5, 60, 60);
+ *   }
  * }
  *
+ * // When a mouse button is pressed, the screenshot is captured
+ * function mousePressed() {
+ *   screenshot = getCanvasImage();
+ * }
+ * </code>
+ * </div>
+ *
+ * @alt
+ * Screenshots are made and displayed on mouse click
  */
 p5.prototype.getCanvasImage = function() {
   if (this._curElement && this._curElement.elt) {
-    return loadImage(this._curElement.elt.toDataURL('image/png'));
+    return this.loadImage(this._curElement.elt.toDataURL('image/png'));
   }
 };
 
--- a/src/image/image.js
+++ b/src/image/image.js
@@ -359,7 +359,9 @@ p5.prototype._makeFrame = function(filename, extension, _cnv) {
  *
  */
 p5.prototype.getCanvasImage = function() {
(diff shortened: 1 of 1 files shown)


ogoossens:
I've noticed there was a check for **this._curElement.elt** and **this._curElement** before assigning/using this resource so I've added this check too.

**Was:**
```
return loadImage(this._curElement.elt.toDataURL('image/png'));`
```

**Proposed now:**
```
if (this._curElement && this._curElement.elt) {
  return loadImage(this._curElement.elt.toDataURL('image/png'));
}
```

ogoossens:
**Hi,**

Could someone feedback me the proposed changhes? 

And two questions, if I may, please: 

What is the difference between p5.image.js and image.js in the src/image directory? Is there a reason to have two files? 

And can anyone please tell me why the travis-ci check is failing? Im not sure why this keeos happening.

**Thank you**

**Oliver**

limzykenneth:
Hi @ogoossens, generally, I would advice opening an issue for a feature request first, even if you intend to create the PR yourself, before working on any changes. Just so that you won't have to put in effort on features that may or may not be accepted into the library itself. Also, it would help to follow the steps outlined [here](https://github.com/processing/p5.js/wiki/Development#setup) when making changes to the library, especially the tests since that's exactly what travis is running.

Having said that, I don't think this is a bad feature to have in the library. However, I don't really understand the changes you have made. I'll make some line comments in a bit.

Thanks.

src/image/image.js
```diff
@@ -341,4 +341,27 @@ p5.prototype._makeFrame = function(filename, extension, _cnv) {
...
+ * Returns the current canvas as a loaded p5.Image. This allows an assignment
+ * or immediate use in the users sketch as a canvas screenshot.
+ *
+ * @method getCanvasImage
+ * @param none
+ *
+ * @example
+ * // This example shows how a user can directly create a screenshot of his
+ * // canvas and have it saved for future use or use immediately
+ * draw() {
```
limzykenneth: it should have been `function draw()`? Also have a look at the other functions and how they format the documentation and examples, you are missing some HTML tags I believe

src/image/image.js
```diff
@@ -341,4 +341,27 @@ p5.prototype._makeFrame = function(filename, extension, _cnv) {
...
+ *   // .. other code
+ *   var screenshot = getCanvasImage();
+ *   image(screenshot, 10, 10, 50, 50);
+ * }
+ *
+ */
+p5.prototype.getCanvasImage = function() {
+  if (this._curElement && this._curElement.elt) {
+    return loadImage(this._curElement.elt.toDataURL('image/png'));
```
limzykenneth: This is the part Travis is complaining about. `loadImage` is not defined in this scope and you should be accessing it via `this.loadImage`. However, I feel this implementation is a bit clunky as you are converting the canvas buffer into a png image then loading it in via `loadImage`, I wonder if it is possible to do a more direct conversion from the canvas buffer to an image buffer without this convert to a png image bit.

ogoossens:
**Hello Kenneth,**

thank you very much for your input. Sadly I'm quite new - so thank you for the guidance.

**Latest commit:**

- I've fixed the **example** part with correct HTML formatting
- Also changed it to a better example
- fixed also the scope and Travis CI - passed

example

**Thoughts:**

- The issue/wiki approach, I'm sorry, I didn't know that a feature request also should be there - will do it next time for sure. (I've checked and all I saw were issues and problems)

- Yes, I also was looking for a more direct approach between canvas and P5.image but even the p5.js **saveFrames()** function - where I took inspiration from uses this approach so I though, this is the best we have. 

**Thank you**

**Oliver**

ogoossens:
Should you still feel a discussion about implementation should be opened in the issues section - please let me know and I will do so. **Thank you**

ogoossens:
Uhmm.. I just realized, that **get()** with **no parameters** already does what I propsed. My piece of code is obsolete. Thank you though. Oliver

PR #2322, opened 2017-10-30
Cross-browser Benchmarking

### LINKED ISSUE #1524: add performance testing and benchmarking

Jared-Sprague:
This is a pull request that fixes #1524


emalgamatedsoy:
Thank you.

PR #2325, opened 2017-10-31
#1524 Fix karma Safari timeout disconnect

Jared-Sprague:
Fixes an issue with Safari timing out and disconnecting when running the `grunt karma` on mac.  See: https://github.com/processing/p5.js/issues/1524#issuecomment-340819526

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,6 +1,6 @@
 /**
  *  This is the global shared configuration for karma.  Specific
- *  benchmark configuation is located in grunt-karma.js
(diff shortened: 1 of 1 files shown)


Jared-Sprague:
Hmm looks like one test didn't pass, but the tests pass locally for me:

```
  346 passing (35s)
  1 failing
  1) src/io/files.js loadJSON documentation example #1 works:
```
I think the build just needs to be re-run?

limzykenneth:
Yeah, that's still a problem that is being worked on.

Issue #1881, opened 2017-04-07
grunt tests occasionally fail


lmccart:
rerunning grunt works again. this happens both with travis ci and locally

<img width="1024" alt="screen shot 2017-04-07 at 1 30 13 pm" src="https://cloud.githubusercontent.com/assets/191056/24818724/77a5386a-1b96-11e7-9d54-0ce38ec696a2.png">

outofambit:
this seems to reliably come from `grunt mocha:yui`, in the first example for `loadJSON`. my guess is that its because this example loads JSON from an external web resource (the USGS Earthquake API) and the request fails for some external reason. (i'm able to reproduce the failure by turning my laptop's wi-fi off.)

kadamwhite:
Not sure how applicable they are here but we could consider using a library like VCR or Replay to save a snapshot of that response locally, to allow the tests to pass offline

outofambit:
i'm looking at https://github.com/seglo/grunt-connect-prism. seems like it wouldn't be too hard to integrate into our infrastructure

meiamsome:
If this is just an intermittent network issue we could add `retries(5)` to the test so it can try again?

outofambit:
@meiamsome that's a good short term solution, but in the longer term I think the tests should pass even when there is no internet access on the machine.

@sakshamsaxena let me know if there's any way i can help!

meiamsome:
@outofambit I totally agree. It looks to me like the real issue here is that `loadJSON` doesn't call `self._decrementPreload()` when the load fails - is there some defined intended behaviour for failures in preloads? If not then it's kind of undefined behaviour when this fails, so maybe the test should not be considered a failure?

sakshamsaxena:
Yep I'm looking into it now. I worked a lot with grunt this summer, so I believe I've developed a good understanding and likeability for it, and also the patience to deal with such random failures :laugh :sweat_smile I'll keep ya posted!

sakshamsaxena:
Found new errors while I was on Windows. See #2209 

Still working on isolating other errors meanwhile.

sakshamsaxena:
Here's what I've figured out till now :
On Windows, ```Files p5.prototype.loadTable CSV files should handle escaped quotes and returns within quoted fields ``` is a constant error due to the difference of ```\r\n``` and ```\n``` which fails the assertion.
[...]

PR #2335, opened 2017-11-03
Improve p5.Vector

### LINKED ISSUE #2318: A few thoughts about p5.Vector implementation
I can make PR, but first I would like to discuss the list below.

1.  Whenever Vector instance is created with `new p5.Vector(...)`, the default value
    for x, y and z is 0.

    Then in every Vector method that accepts Vector as argument value, default 0 is applied again and again, for example:
    ```javascript
    p5.Vector.prototype.set = function (x, y, z) {
      if (x instanceof p5.Vector) {
        this.x = x.x || 0;
        this.y = x.y || 0;
        this.z = x.z || 0;
        return this;
      }
      ...
    };
    ```
    I think it is not necessary, because the default value is already there.

2.  The default value of `n` in `p5.Vector.prototype.mult` is 0.
    When I call `mult` with no argument, my vector will be mutated.
    I think the default value of `n` should be 1.

3.  We don't check if we divide by 0 in `p5.Vector.prototype.div`.
    There is a lot of validation in other methods, but not in `div`.
    Is it done by design?

    Checking divide by 0 case in `div` method would make `p5.Vector.prototype.normalize` a bit simpler.

4.  Performance optimisations:

    - this.p5.random() => Math.random()
    - Math.PI*2 => constants.TWO_PI

    BTW. I can't get the fundamental difference between `if` and `else` parts,
    as the results seems to be the same, except the `if` part will be slower.

    ```javascript
    p5.Vector.random3D = function random3D() {
      ...
      if (this.p5) {
        angle = this.p5.random(0,constants.TWO_PI);
        vz = this.p5.random(-1,1);
      } else {
        angle = Math.random()*Math.PI*2;
        vz = Math.random()*2-1;
      }
      ...
    }
    ```

5.  Constructor could be much simpler if there were named arguments and different order of them.
    `function() => function(x, y, z, context)`

    ```javascript
    p5.Vector = function(x, y, z, context) {
      // This is how it comes in with createVector()
      if (context instanceof p5) {
        this.p5 = context;
      }
      this.x = x || 0;
      this.y = y || 0;
      this.z = z || 0;
      this.name = 'p5.Vector';
    };
    ```
    This of course should be propagated in other places, but there is only a few such a cases. 

6.  Functions should be named. This will help during debugging.

    ```javascript
    // Bad example
    p5.Vector.prototype.set = function (x, y, z) {...}

    // Good example
    p5.Vector.prototype.set = function set(x, y, z) {...}

    // Still good, but over engineered example
    p5.Vector.prototype.set = function p5VectorSet(x, y, z) {...}
    ```

tawez:
Improve debugging by naming all anonymous functions.
Optimize calculations performance.
Improve code readability and consistency.
Improve p5.Vector.prototype.mult behaviour:
- do nothing if n is undefined or not a valid number and throw a console warning.

Improve p5.Vector.prototype.div behaviour:
- do nothing if n is undefined or not a valid number and throw a console warning;
- do nothing if n is 0 and throw a console warning "divide by 0"

Resolves #2318


tawez:
It looks like it is a problem with Travis, not the with the code itself.
Weird... and it looks like this problem occurs in other PRs.

Is there a way to retrigger Travis?

limzykenneth:
@tawez No worries, that's "expected" see #1881 I'll restart the tests

Spongman:
i think you should start again, and disable you editor's blatant 'reformat everything' setting.

limzykenneth:
The current failure is related to #2338 which should be fixed in the source. If travis doesn't run the new tests on this PR then we need to manually review this then merge it in if everything seems fine and let travis do its job then.

tawez:
@Spongman What do you mean by

> disable your editor's 'reformat everything' setting

?

Spongman:
it seems that most of the changes in this PR are whitespace and/or formatting changes. you should probably split those out into a separate PR.

tawez:
Well, if it will help, I can split this PR into two, 1st - code, 2nd - formating

lmccart:
I think this one is taken care of with #2350

PR #2336, opened 2017-11-04
Add code auto-formatting using eslint fix and prettier

Zalastax:
Improves on #2269 by adding commands that formats the code automatically, thereby closing #105.
The precommit hook makes sure that lint and style guides are adhered to, but I opted to not have it run eslint with fix on since I don't expect my code to be modified when I make a commit. Style-errors can also be fixed while you write code by adding appropriate plugins (e.g. vscode-eslint or prettier-vscode).
I'll update the Development wiki page after merge, replacing the JS Style guide section and adding a part about running `npm run fix` when seeing `X errors, Y warnings potentially fixable with the `--fix` option` / commits stopped by the precommit hook.


package.json
```diff
@@ -9,8 +9,16 @@
     "build": "grunt build",
     "docs": "grunt yui",
     "test": "grunt",
+    "fix": "grunt eslint:fix",
```
tawez: To fix what?
I would prefer to be more descriptive.

`lint:fix` will be good.
And no `eslint:fix` please, because we have changed linting tool recently and we can do this in the future.
With `lint` script name we can do this whenever we want and the command will remain unchanged.

.eslintignore
```diff
@@ -0,0 +1 @@
+/src/external/**/*.js
```
tawez: Just general question: why is it needed?
I don't see such a folder in repository.

.eslintrc
```diff
@@ -11,17 +11,22 @@
...
   },
+  "extends": ["eslint:recommended", "prettier"],
+  "plugins": ["prettier"],
   "rules": {
+    "prettier/prettier": [
+      "error",
+      {
+        "printWidth": 100,
+        "singleQuote": true, 
+        "trailingComma": "es5"
```
tawez: Please, no. Don't force trailing commas.
1. It doesn't make any change in code quality, especialy when we have linter and precommit hook and we are abble to fix all the issues before the commit.
2. As far as I know we are pre ES5 compliant.
3. It's just the lazy programmer's excuse. "Hey, look. If I would like to add another array element, the comma is already there".

.eslintignore
Zalastax: It was there in the previous Gruntfile. Happy to remove it if it's not needed.

package.json
Zalastax: Absolutely!

.eslintrc
Zalastax: Looking at our browser compatibility table ES5 looks like a requirement. I prefer trailing comma since I can work faster with it and have to think less. It can also reduce noise in commits by not changing two lines when adding at the end. @lmccart what's your take on it? Also see if there are other style rules to change.

.eslintignore
tawez: I don't have this knowledge. We need a confirmation from the Elders ;)

.eslintrc
tawez: > I prefer trailing comma since I can work faster with it and have to think less.

Frankly speaking this can be stated about almost every code formatting rule ;)
All in all it's a matter of personall feelings.

All I want to say is, we should chose a subset od settings, that will be widely acceptable (it's hard).
Trailing commas is not so widely acceptable rule (basing on my experience with various open source and commercial projects).

limzykenneth: Not sure if constructive as I don't have any opinion on this, I normally omit trailing commas just out of habit. May be a bit obstructive to force it though but I don't expect it to be too big a problem for me.

.eslintignore
limzykenneth: If the folder doesn't exist, then there's no reason to keep it. I think it was for the reqwest we were using before or something like that

.eslintrc
Zalastax: > All in all it's a matter of personall feelings.
Yes, exactly. There is an argument to be made that we should stick to the defaults of prettier, meaning no trailing comma. Any other rules to remove / add?

Jared-Sprague: Just like you, I also prefer trailing commas for speed,  but p5 has been using idiomatic.js. which does not use them,  so when I'm workyon p5 I leave them out.   I think it's better to keep the eslint and prettier as close to idiomatic.js rules as possible since that's been the standard for p5.

.eslintignore
lmccart: yes we can remove this

.eslintrc
lmccart: it's a matter of personal preference of course. I would agree with @Jared-Sprague here, omitting the trailing commas also aligns with the documentation on the rest of the website so I think it would be more intuitive for beginner developers coming to the codebase which is a good guiding principle in terms of the rules in general I think.

Spongman: i vote for trailing commas. it makes code motion significantly simpler.

Zalastax:
I've addressed all the comments so please take a new look. I'm getting errors when I test locally, but they are the same as what I get on the current master. We should solve #2338 as fast as possible.

Zalastax:
Tests pass!

src/core/curves.js
```diff
@@ -340,15 +340,11 @@ p5.prototype.curve = function() {
...
     args.push(curveDetail);
     this._renderer.curve(args);
-  } else{
-    this._renderer.curve(
-      args[0], args[1],
-      args[2], args[3],
-      args[4], args[5],
-      args[6], args[7]);
+  } else {
+    this._renderer.curve(args[0], args[1], args[2], args[3], args[4], args[5], args[6], args[7]);
```
limzykenneth: There are many lines like this that are rather long. We used to enforce 80 character limit per line with jscs, I'm not sure if eslint does that right now but I think it is good to keep otherwise you may get some lines that are hard to read on a text editor because you need to scroll sideways and also hard to view on git commits.

Zalastax: Alright, I'll go back to 80 characters. To me 100 characters looked better in my editor, but I get your points.

limzykenneth:
I think this may still need more tweaking (manually even) and some more considerations. I had some time to have a look at the changes in the source themselves and while it cleaned up a lot of inconsistencies, it also "cleaned up" some desirable "mess" as well, some of them are around comments and some are the way functions arguments are passed. For example, [this](https://github.com/processing/p5.js/pull/2336/files#diff-f5034d55fd652b7016c8def87cb18042L457) function call to drawImage have a nicely formatted arguments order that we can easily see what is passed and how but after formatting, it's just a long list of arguments vertically.

I'm starting to think whether we should automate prettifying the code base as it's rather hard for algorithms to understand nuances like the above, and maybe we should just run this once and not include it as a step that "clean" everything...

Zalastax:
In general I think it's better to not have arguments that want to be grouped like that, and instead create new objects / arrays for grouping. Changing the whole code base to such style would be a waste of time and effort so using special formatting is the pragmatic approach. Prettier supports ignoring code from being formatted, using [special comments](https://prettier.io/docs/en/ignore.html), letting us keep auto-formatting where it's not a problem

Zalastax:
I don't understand why the build is failing. If I lint locally I have no errors and the branch is up to date https://github.com/processing/p5.js/compare/969b97926eef0c99b9ffb07a1bf0485a49134f98...afa6731c50efcd1d3263f1c7f48baa337c765868.

limzykenneth:
`drawImage` for example is a native JS function and there's not much we can do in terms of grouping arguments into array or objects (unless you want to call it with something like `.apply`). There's no need to change the whole code base to follow that style that's why I think a manual approach may be needed. Using special comments is a nice solution to this but I'm worried that we may end up with a bunch of them polluting the code base randomly.

Not sure why the tests are failing though..

Zalastax:
I don't think the comments will pollute the code base too much and the benefits of auto-formatting should outweigh the drawbacks by a large margin. Having to go through all the changes manually is a bit annoying but my quick glance didn't see that many places with weird formatting.

Zalastax:
@limzykenneth I used a regex to find every place I think could become a formatting problem and added comments there. Are there any more places you think needs to be excluded?

limzykenneth:
There are too much files to go through right now for me but I think you may have gotten most of them.

Zalastax:
@lmccart ready to merge if you are happy!

Zalastax:
Can I get a review on this? I don't want to keep rebasing the changes on top of master. To me it looks ready.
@lmccart can I steal your time?

lmccart:
@Zalastax sorry for the delay! this all looks good. I think the // prettier-ignore is a reasonable solution and the inconvenience of that is worth it to have the rest of the codebase cleaner. thank you for your work on this! will you update the developer wiki page with the relevant prettier info?

Zalastax:
@lmccart great!  
I've updated [the Development wiki page](https://github.com/processing/p5.js/wiki/Development). One thing hit me: Why don't we add all the interesting grunt commands as npm-scripts so the developer don't have to install grunt-cli globally?

Also, all of the open PRs need to have their Travis builds re-run to check for style problems. I can do that if you give me the permission in Travis, or you can just re-run them yourself. In the meantime I'll try to fix the PR's that already have merge conflicts to not add that burden on those with open PRs.

lmccart:
@Zalastax cool thanks! I'm not sure how to give you travis access, do you know how? I just added you to the p5.js team within the processing organization. I can also rerun them myself too if thats easier.

Zalastax:
Thanks @lmccart! Travis picked it up so I can now retrigger builds!

Issue #2203, opened 2017-09-23
reference doc for arc() doesn't explain modes OPEN, PIE, CHORD


makeyourownalgorithmicart:
The reference documentation for arc() at [https://p5js.org/reference/#/p5/arc](https://p5js.org/reference/#/p5/arc) doesn't explain the modes, OPEN, PIE, CHORD.

The examples show OPEN but the parameter list only lists PIE and CHORD. 

None are explained. The examples look the same, so not illustrating the modes

meiamsome:
The reference image colours are currently broken as a consequence of #2156, and should be fixed whenever the documentation site is updated. I do agree a description update here would be useful.

Siphonophora:
I can flesh out the documentation while working on #2181

eJohnstonePhd1991:
Hi all.
I'm looking to make my first contribution to the project. Perhaps I could update the documentation and then @Siphonophora can update it when the WEBGL extension of it is live?

lmccart:
fixed with https://github.com/processing/p5.js/pull/2346

PR #2346, opened 2017-11-07
Issue #2203: Updated arc() reference documentation

eJohnstonePhd1991:
I have slightly expanded and reworded the documentation for the arc() reference in order to better explain the modes OPEN, CHORD and PIE. I have also added an example showing the default mode to show how it compares with the other modes.


lmccart:
thanks!

Issue #2262, opened 2017-10-12
I can't see the examples in references


hjoaco:
Hi there! I consult the reference for 3D Primitives and I can't see the examples. The console of Chromium show this error:

```
Uncaught TypeError: Cannot read property 'location' of undefined
    at d.RendererGL.createBuffers (p5.min.js:37)
    at e.d.sphere (p5.min.js:38)
    at draw (eval at s (render.js:178), <anonymous>:8:3)
    at e.d.redraw (p5.min.js:33)
    at e.<anonymous> (p5.min.js:32)
    at e.<anonymous> (p5.min.js:32)
    at new e (p5.min.js:32)
    at render.js:227
```
And this is how I see the example for sphere:

![error-example](https://user-images.githubusercontent.com/2771752/31498917-b7322a08-af0f-11e7-89f1-c6a7b2d1254e.png)

#### Most appropriate sub-area of p5.js?

- [x] 3D Primitives References

#### Which platform were you using when you encountered this?

- [x] Desktop/Laptop

meiamsome:
What is your browser's version, OS and Graphics Card (If any)?

I am experience the same issue on Ubuntu 17.04 with Chrome & Firefox using the native rendering (nvidia graphics card). This does not happen with a software-based WEBGL.

hjmg:
Chromium and Firerox both update are installed in Manjaro, my laptop is a Lenovo Thinkpad T410i with Intel and Nvidia graphis card.

meiamsome:
Okay, and I presume then that if you run chromium with the GPU disabled (`--disable-gpu` after chromium on the command line) the reference renders correctly?

hjmg:
Thank you!
I have downloaded the reference and works perfectly offline, but the problem is in the website

vincenzoml:
Wait, exactly what did you do? I have the same error only if I enable webgl.

copy:
@meiamsome I think this should be reopened, as disabling webgl isn't a proper fix.

achimkoh:
Agree to reopen! My mobile browser (Android 7.0.0, Chrome 63.0.3239.111 32-bit) gives me the same error for all 3d primitives examples.

Spongman:
this is fixed by #2349. some implementations of the webgl shader compilers optimize out attributes that aren't used by the program.

tendau:
Hope this is related but I also cannot open any webgl examples getting the error
[...]

Issue #2344, opened 2017-11-06
TRIANGLE_STRIP with WEBGL not working properly?


eivindml:
#### Nature of issue?

- [X] Found a bug

#### Most appropriate sub-area of p5.js?

- [X] WebGL

#### Which platform were you using when you encountered this?

- [X] Desktop/Laptop

#### Details about the bug: 

- p5.js version: p5.js v0.5.16 October 11, 2017
- Web browser and version: Safari 11.0.2 
- Operating System: macOS High Sierra
- Steps to reproduce this:

Using this code:

```javascript
var cols, rows, w, h;
var scl = 100;

function setup() {
    createCanvas(windowWidth, windowHeight, WEBGL);

    w = windowWidth;
    h = windowHeight;
    cols = w / scl;
    rows = h / scl;
}

function draw() {
    background(245, 244, 236);
    stroke(17, 18, 16);
    noFill();

    translate(0, 600);
    rotateX(PI / 3);
    translate(-w / 2, -h / 2);

    for (var y = 0; y < 1; y++) {
        beginShape(TRIANGLE_STRIP);
        for (var x = 0; x < cols; x++) {
            vertex(x * scl, y * scl, 1);
            vertex(x * scl, (y + 1) * scl, 2);

        }
        endShape();
    }
}
```

I get this result:
<img width="1542" alt="screen shot 2017-11-06 at 23 05 51" src="https://user-images.githubusercontent.com/202450/32466376-1081fe8a-c347-11e7-9531-c67d6abd4c0a.png">

I would expect all vertices  to be connected like this:

<img width="693" alt="screen shot 2017-11-06 at 23 07 32" src="https://user-images.githubusercontent.com/202450/32466445-4ca44940-c347-11e7-9534-b7f80a8e4167.png">

It works like it should in 2D environment, but not when I set it to WEBGL.

eivindml:
Anyone else experiencing this?

sahilkhanna:
Facing the same issue on P5.js "0.5.16"
Running it on chrome on windows with WEBGL.

eivindml:
Yes. It definitely behaves as a `LINE_LOOP` rather than `TRIANGLE_STRIP`. It works as expected in 2D, but not in 3D.

mlarghydracept:
Good catch! The strokes themselves are actually drawn with gl.TRIANGLES because they are polygons with faces but you are right that there should be a way to fake the other drawing modes. I'll look into how Processing does it.

[...]

PR #2349, opened 2017-11-09
Custom shaders

Spongman:
i was playing around and discovered that it's impossible to use a custom shader that doesn't make use of all the vertex attributes that are used in the built-in shaders. this PR loosens this restriction, and adds an example using such a custom shader to the p5.Shader constructor docs.


kjhollen:
yay! yes, this was on my list of shortcomings for the current shader implementation—thank you for taking this on. I've messed up my notifications for this project somehow and am sorry I didn't see this sooner—I'll try to review soon!

Spongman:
thanks! i just learned that you can add '?w=1' to the end of a diff's URL in github and it'll ignore whitespace. might be helpful in this case...

Spongman:
@kjhollen did you ever manage to review this? I have a few other webgl-related PRs that might help close some issues.

Spongman:
fixes #2435

Spongman:
also fixes #2262

Spongman:
Mostly it removes the assumption that all those buffers are used by the shader. Without this fix you get an exception thrown if you try to use a shader that doesn't define all those attributes (eg, try to run the sample in the 2nd commit).

This problem also manifests in some browsers where the attributes are declared in the shader, but are optimized away by the shader compiler. This is what is causing the issues on Firefox/Linux and chrome/android.

I'll take a look at the retained-mode code...

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -161,25 +161,31 @@ p5.RendererGL.prototype._drawFillImmediateMode = function(
...
+      this.curFillShader.attributes.aPosition.location,
+      3,
+      gl.FLOAT,
+      false,
+      0,
+      0
+    );
+  }
+
+  if (
```
kjhollen: @lmccart any style guide suggestions here for multi-line if statement? I'm not sure we've used this convention elsewhere.

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -197,7 +203,11 @@ p5.RendererGL.prototype._drawFillImmediateMode = function(
       0
     );
   }
-  if (this.drawMode === constants.TEXTURE) {
+
+  if (
+    this.drawMode === constants.TEXTURE &&
+    this.curFillShader.attributes.aTexCoord
+  ) {
```
kjhollen: hmm, if the drawMode is TEXTURE but there are no texture coordinates available, this feels like maybe an error worth noting to the user? (under what circumstances is this edge case possible—is this true if we're drawing lines?)

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -221,20 +231,14 @@ p5.RendererGL.prototype._drawFillImmediateMode = function(
   if (this.drawMode === constants.FILL || this.drawMode === constants.TEXTURE) {
     switch (this.immediateMode.shapeMode) {
       case constants.LINE_STRIP:
```
kjhollen: @mlarghydracept is the math for calculating the shapes done elsewhere (are they computed to be triangle fans always? or should we mark this part of the code as TODO for future implementation / given issue #2344 ?)

src/webgl/p5.RendererGL.Retained.js
```diff
@@ -43,127 +56,24 @@ p5.RendererGL.prototype.createBuffers = function(gId, obj) {
...
-    this.curFillShader.attributes.aTexCoord.location,
-    2,
-    gl.FLOAT,
-    false,
-    0,
-    0
-  );
-  //}
-};
+  if (this.curStrokeShader.attributes.aPosition) {
```
kjhollen: this looks great! it occurs to me that now it's probably worth adding a comment about what each if-block is doing here to make the code a little more scannable. we probably should have had this already. :)

src/webgl/p5.Shader.js
```diff
@@ -16,6 +16,47 @@ var p5 = require('../core/core');
...
+ *   '  float n = 0.0;' +
+ *   '  for (int i = I; i > 0; i --) {' +
+ *   '    if(z.x*z.x+z.y*z.y > 4.0) {' +
+ *   '      n = float(i)/float(I);' +
+ *   '      break;' +
+ *   '    }' +
+ *   '    z = vec2(z.x*z.x-z.y*z.y, 2.0*z.x*z.y) + c;' +
+ *   '  }' +
+ *   '  gl_FragColor = vec4(0.5-cos(n*17.0)/2.0,0.5-cos(n*13.0)/2.0,0.5-cos(n*23.0)/2.0,1.0);' +
+ *   '}';
```
kjhollen: example needs a short description for accessibility

src/webgl/p5.Shader.js
```diff
@@ -16,6 +16,47 @@ var p5 = require('../core/core');
...
+ *
+ * var mandel;
+ * function setup() {
+ *   var renderer = createCanvas(100, 100, WEBGL);
+ *   mandel = new p5.Shader(renderer, vs, fs);
```
kjhollen: hmm, other docs say that the return type from createCanvas is an HTMLCanvasElement, not an instance of p5.RendererGL?

src/webgl/p5.RendererGL.Immediate.js
Spongman: #2336 added auto-formatting for the whole library. this is how it's configured, apparently.

src/webgl/p5.RendererGL.Immediate.js
Spongman: i don't think that's an error. you could certainly have a shader that uses a texture, but does not use the aTexCoord vertex attributes.

src/webgl/p5.RendererGL.Immediate.js
Spongman: I'm not sure about that. i don't think i changed any logic here...

src/webgl/p5.Shader.js
Spongman: yup. looks like the docs are wrong ;-)

Spongman: maybe it might be nice to add a `createShader()` method that wraps this functionality?

kjhollen: oops :)

there's `loadShader()` but it takes file names! https://p5js.org/reference/#/p5/loadShader, maybe this can be overridden to do what you describe?

documentation for this method needs a good bit of work, but there are some working examples in the manual tests.

src/webgl/p5.RendererGL.Immediate.js
kjhollen: ok, we will go with what the auto-formatter wants :)

src/webgl/p5.Shader.js
Spongman: add a comment here is easier said than done ;-)

Spongman: oh, i see you meant an @alt tag, i'll add one...

Spongman:
i added a `createShader()` method that takes the shader sources as strings. and i moved the mandelbrot sample to that and removed the now-unnecessary fetching of the renderer.

i also added some comments to the sample.

src/webgl/p5.RendererGL.Immediate.js
kjhollen: Right. Something about not having the individual outcomes explicitly (even though they were all the same) made me wonder if this was done on purpose originally as a stop-gap and maybe doesn't indicate that as clearly, updated in this way. It's functionally the same, though!

src/webgl/p5.Shader.js
```diff
@@ -44,38 +77,28 @@ var p5 = require('../core/core');
...
-  this._glProgram = 0;
-  this._loadedAttributes = false;
-  this.attributes = {};
-  this._loadedUniforms = false;
-  this.uniforms = {};
-  this._bound = false;
-  this.samplers = [];
-
-  return this;
+p5.prototype.createShader = function(vs, fs) {
```
kjhollen: this looks great. the rest of this file only defines the Shader class, but no p5.prototype methods. I think it feels ok here but wonder if it belongs in material.js alongside loadShader? this file has more of the p5 public methods for working with shaders.

kjhollen:
just built this on my machine and everything seems great! automated tests are passing, manual tests work (actually one is broken, but is broken in the mainline, too) :) can we just get the createShader method moved over to material.js? then we'll be ready to go. lots of great work in here—THANKS!!!

kjhollen:
awesome, THANK YOU!! 🎉

src/webgl/p5.RendererGL.Immediate.js
mlarghydracept: Hey sorry - my notifications are whacky so I sometimes miss my mentions. No changes are made here. The reason things are happening like they do over in #2344 isn't because of gl modes because gl isn't actually rendering those lines. So, counter-intuitively, the constants.LINE_STRIP has nothing to do with that problem.

saitogroup:
I looked at another thread and it said that error was solved here... but I couldn't see/understand that. I think it has something to do with when you're texturing an object, rather than texturing a primitive, but I'm not entirely sure (some kind of reference error).

PR #2362, opened 2017-11-13
Docs duplicate routes fix

limzykenneth:
Should fix https://github.com/processing/p5.js-website/issues/146 while preserving existing behaviours.


Zalastax:
When reviewing this please note that it adds a [package-lock.json](https://docs.npmjs.com/files/package-lock.json). Adding a lock-file is a good idea since it prevents people from accidentally running different versions of our dependencies. npm5 and yarn both support lock-files but they use different names and formats so before accepting the commit make a conscious decision on whether to officially support npm5 or yarn. Hopefully they will play along better in the future.

limzykenneth:
Yeah, let me know if I should move that out of this PR though in case things are not confused with each other.

Spongman:
yeah, you should definitely remove the package-lock.json commit.

Issue #2351, opened 2017-11-09
rotate() map to rotateZ() in WEBGL?


shiffman:
In Processing,  if `rotate()` is called in `P3D` it is the same as `rotateZ()`. This allows code to move back and forth between 2D and 3D more easily when testing. Advisable to do the same here?

At the moment, if you call `rotate()` in `WEBGL`, the following error appears:

```
p5.min.js:8 Uncaught TypeError: Cannot read property 'transpose3x3' of null
    at d.Matrix.inverseTranspose (VM48 p5.min.js:8)
    at d.RendererGL._setMatrixUniforms (VM48 p5.min.js:9)
    at d.RendererGL.f.RendererGL.drawBuffers (VM48 p5.min.js:9)
    at d.RendererGL.rect (VM48 p5.min.js:9)
    at e.d.rect (VM48 p5.min.js:4)
    at draw (VM116 about:srcdoc:54)
    at e.d.redraw (VM48 p5.min.js:6)
    at e.<anonymous> (VM48 p5.min.js:4)
    at e.<anonymous> (VM48 p5.min.js:4)
    at new e (VM48 p5.min.js:5)
```

Spongman:
i fixed this in #2366, however it doesn't address the issue that the origins are different between P2D & WEBGL modes. obviously it would be a rather large breaking change to move the WEBGL origin to the top-left...

lmccart:
fixed with #2366

PR #2366, opened 2017-11-16
make 'rotate' default to Z-axis

Spongman:
prevents 'rotate()' from crashing when not passed an axis in WEBGL mode.


Spongman:
@shiffman ran into this bug during his live stream.

mlarghydracept:
This is good!  While you are at it could you add a check to see if the user passed no arguments and log a more informative error to the effect of "The function rotate() requires at least 1 argument, the number of radians to rotate"?

Just want to avoid people getting the cryptic error that @shiffman got [over here](https://github.com/processing/p5.js/issues/2351).

Spongman:
i already have a PR pending that improves the parameter validation stuff. i don't want to add a whole bunch of ad-hoc validation that would just be obsoleted when that's merged...

shiffman:
Thank you for working on this!

Issue #2378, opened 2017-11-26
textLeading(0) has surprising result


aparrish:
According to the documentation, a call to `textLeading(0)` should result in no pixels between each line (i.e., all lines of text should be drawn on top of one another). What's happening now is that `textLeading(0)` has no effect. The following code demonstrates the issue; [here's the code running in the web editor](https://alpha.editor.p5js.org/allison.parrish/sketches/SyS5_cdlM).

```
function setup() { 
  createCanvas(400, 400);
} 

function draw() { 
  background(220);
  textAlign(LEFT, TOP);
  textLeading(0);
  text("This\nis\na\ntest", 20, 20);
}
```

I suspect this is happening because JavaScript counts `0` as `false`, which means that [this expression in the `if` statement fails](https://github.com/processing/p5.js/blob/a5751a8573ae5e198fcb92392db6330565ddf57b/src/core/p5.Renderer.js#L88) and the function behaves as though it has received no parameters at all. (Actually, looking at that file just now, it seems like the same problem would apply to a call to `textSize(0)`.)

Not a high-priority bug for sure, but it did lead to surprise and consternation when I was doing a tutorial about text and type the other day!

dhowe:
thanks @aparish -- i guess this is more of an API decision: do we want to allow zero or even negative leadings? I suppose there is no harm, but if not, we should probably have a warning message for such cases

aparrish:
whoops, I'm @aparrish, not @aparish! I guess it is an API decision, though personally I don't really see any reason other than olde-tyme-type-drawer verisimilitude to not support (or even to warn about) zero or negative leading. (negative leading can produce some fun in-class examples, [like this one](http://alpha.editor.p5js.org/allison.parrish/sketches/rktvmGckG).)

dhowe:
ah right, sorry about that... anyhow, I was suggesting _either_ to support or warn

kadamwhite:
I’d be in favor of supporting; the result should be fairly visible and it yields interesting results. It doesn’t feel like a warning is appropriate if this is something that somebody might want to do on purpose and I’ve seen a couple of nice projects that use 0 leading elsewhere on the web, so there’s definitely a use case

lmccart:
I agree this makes sense, thanks for the patch @dhowe.
fixed with https://github.com/processing/p5.js/pull/2379

PR #2379, opened 2017-11-27
Fix to #2378

dhowe:


dhowe:
Now checking for number rather than truthy arg in textSize(n) and textLeading(n)

PR #2383, opened 2017-11-28
Commit package-lock.json

limzykenneth:
Commit package-lock.json for npm 5. I don't expect problem with the dependency tree but it may be best to try out this fork first before merging.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,13318 @@
+{
+  "name": "p5",
+  "version": "0.5.16",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "JSONPath": {
+      "version": "0.10.0",
+      "resolved": "https://registry.npmjs.org/JSONPath/-/JSONPath-0.10.0.tgz",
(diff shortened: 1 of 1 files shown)


limzykenneth:
I may have messed something up as the test is not passing but I'm not sure exactly what... 😞

Zalastax:
Great idea!
Before merging read my comment on the choice your making then: https://github.com/processing/p5.js/pull/2362#issuecomment-344681398. I'm fine with both choices as long as you know that you're making it.

I guess you could rebase the commit and push it again if you want to trigger a rebuild. Or someone with Travis rights could do it. I think it needs to be updated for the latest package.json anyways.

limzykenneth:
@Zalastax i have travis rights but it is trying to test a branch that doesn't exist so it won't pass either way, I think it's ok to ignore it. 😜 

And yeah, do decide if npm or yarn lock file are preferred. I personally prefer npm as that feels more like a standard.

lmccart:
I agree, I'd like to stick with npm. I've just pushed the latest copy of package-lock.json so I will close this pr. thanks!

PR #2389, opened 2017-11-29
add Blob saving support to downloadFile()

### LINKED ISSUE #2270: Bug with saving complicated/large drawings
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Core
- [x] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: v0.5.6
- Web browser and version: Chrome Version 61.0.3163.100
- Operating System: Mac OS Sierra V10.12.2
- Steps to reproduce this:
```
function setup() {
	createCanvas(2048, 2048);
}

function draw() {
        fill(random(255), random(255), random(255));
        noStroke();
        var x = Math.random() * width;
        var y = Math.random() * height;
        blendMode(OVERLAY);
        rect(x, y, 10, 10);
    }
	noLoop();
	save();
}
```

The current feature to save an image stops working when the image is fairly complicated (such as the one above), giving the message "failed - network error"
screen shot 2017-10-16 at 2 32 45 am

A solution that my team and I encountered is to use the library [filesaver.js](https://github.com/eligrey/FileSaver.js/).

Spongman:
this one uses `canvas-toBlob` and `file-saver` to work around the file-size limit with data URLs (at least in browsers that support that mechanism).

fixes #2270


Spongman:
i don't know why it still thinks there are conflicts.

limzykenneth:
A couple things: 
1. Is the `canvas-toBlob` package needed? From what I see from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob) `ctx.toBlob()` is supported on all major browser and there even is a simple polyfill on the page.
2. Since the packages are supposed to be packaged with the library itself I think it makes more sense to save them as `dependencies` instead of `devDependencies`.

Zalastax:
`file-saver` has a nice table showing browser support: https://www.npmjs.com/package/file-saver#supported-browsers. Comparing that and the [MDN link](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob) to https://github.com/processing/p5.js/wiki/Supported-browsers I don't see any browsers we support that need `canvas-toBlob`.

Spongman:
I haven't tested it but I believe android requires the a polyfill. I use the package instead of the MDN polyfill since it looks like it handles some of the error cases better, and it's simpler to remove once it's not required.

Are you sure about the dependencies/devDependencies thing? The polyfills are bundled into the .min.js at build time, so they shouldn't need to be employed separately.

limzykenneth:
@Spongman Hmm...I don't have an Android device to test it so I'm not sure. But I see from MDN that Android browser indeed have no support. Then I guess the question will be whether we want to support Android browser or not. From what I can find it has a 3% market share in all mobile devices. I don't have strong opinion either ways.

Technically with the way we setup the build and deployment we shouldn't have any problem putting packages in dependencies or devDependencies, I just prefer it semantically and also future proof it in a sense. The other polyfills and libraries that are packaged with the library are also in dependencies like `opentype.js` and `whatwg-fetch`.

src/image/image.js
```diff
@@ -143,82 +143,35 @@ p5.prototype.createImage = function(width, height) {
...
-    aText += 'go to File --> Save As.\n';
-    alert(aText);
-    window.location.href = cnv.toDataURL();
-  } else {
-    var mimeType;
-    if (typeof extension === 'undefined') {
-      extension = 'png';
+  var mimeType;
+  switch (extension) {
+    default:
```
limzykenneth: Shouldn't `default` case be at the bottom of the `switch` block? Otherwise it will always trip on `default` and `break` here, making mime type always `image/png`. Or you can just remove the `break` statement as well if that feels more efficient.

Spongman: I'm not sure prcisely what you're saying, but i think you might be confused about how switch statements work: https://codepen.io/Spongman/pen/WXmMaP

Zalastax: `default` can be anywhere in the switch, but it's usually placed at the end because of fallthrough and it being easier to think of it like taking the first match. Move it to the end to reduce cognitive load when reading the code 😃

limzykenneth: Ah yes, @Zalastax is right, I'm not used to reading switch statement like that. Sorry

Spongman: yeah, the eslint recommended rules `no-fallthrough` and `no-duplicate-case` ensure that case fallthrough and ordering should no longer be an issue.

PR #2395, opened 2017-12-04
Add link from blend constant to blend() and blendMode()

jedahan:
How do we feel about this? Is there a better way to link to other parts of the documentation, like with `@link p5.Image/color` or something...

screen shot 2017-12-04 at 11 34 04 am

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -323,6 +323,12 @@ module.exports = {
 
   // RENDERING
   /**
+   * Defines the default image blending mode.
+   * For examples on how to use blend, see [blend()][] and [blendMode()][]
+   *
+   * [blend()]: #p5.Image/blend
+   * [blendMode()]: #p5/blendMode
+   *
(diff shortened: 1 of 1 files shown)


limzykenneth:
It would be nice to have a short description about what the blend mode does to differentiate it from another blend mode say `ADD` etc. Also is it possible to remove the work "final" at the bottom?

Spongman:
how about replacing `final` with `constant` (in docs/yuidoc-p5-theme-src/scripts/tpl/item.html)

limzykenneth:
@Spongman That sound good to me

Spongman:
actually, this can be done automatically. no need for additional markup. these constants are themselves generated from the yuidocs of the methods that use them and all that information is in the data.json.

Spongman:
i created a PR #2397 that does this.

jedahan:
awesome Spongman! closing in favor of the other PR

PR #2397, opened 2017-12-05
automatically add method references to constant docs pages

Spongman:
on the docs pages for enumerated constants, add a list of links to those methods that take those constants as parameters.

image

### DIFF
--- a/docs/preprocessor.js
+++ b/docs/preprocessor.js
@@ -24,11 +24,15 @@ function mergeOverloadedMethods(data) {
   var methodsByFullName = {};
   var paramsForOverloadedMethods = {};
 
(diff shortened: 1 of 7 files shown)


Spongman:
i have no idea why those are failing - this PR doesn't change those files. maybe some bad interaction between eslint and travis?

Spongman:
Oh thanks!! I'll fix these in a mo.

lmccart:
very cool

lmccart:
@Spongman this test is failing for me, any ideas?
I also noticed that the changes in this PR seem to go against the lint check preferences, am I missing something?
screen shot 2017-12-09 at 12 40 00 pm

PR #2401, opened 2017-12-06
webgl: cache viewport in renderer

### LINKED ISSUE #2399: WEBGL mode performs much worse after version 0.5.14
#### Nature of issue?

- [ x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x ] WebGL

- [x ] Desktop/Laptop

#### Details about the bug: 

_MacOS - Chrome_

Hey all,

I noticed some of my WEBGL mode sketches are performing much worse in p5 version `0.5.15` and later. After digging around in the chrome performance tab a bit I found a call to `getParameter` in `p5.Shader.prototype._setViewportUniform` taking > 90% of each frames time that isn't present in version `0.5.14`.

Issue can be reproduced by changing the p5 version in the `script` `src` from `0.5.15` to `0.5.14` and comparing performance stats.
  
```
<!DOCTYPE html>
<html lang="en">
<head>
  <script src='https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.15/p5.js'></script>
</head>
<body>
  <script type="text/javascript">
    function setup() {
      createCanvas(500, 500, WEBGL)
    }
    
function draw() {
  background(0)
  for(var i = 0; i < 100; i += 1) {
    push()
    rotateX(i * frameCount * 0.1)
    rotateZ(i * frameCount * 0.2)
    box(250)
    pop()
  }
}
  </script>
</body>
</html>

```
_0.5.14_
screen shot 2017-12-05 at 5 01 04 pm

_0.5.15_
screen shot 2017-12-05 at 4 59 37 pm

I don't know enough about the webgl internals of p5 to fix this issue by myself, but hopefully this ticket helps someone out. 

Thanks,

Dex

Spongman:
caches the viewport in the renderer in order to avoid expensive getParameter calls.

fixes #2399


Zalastax:
I would assume we need to invalidate the cache whenever `gl.viewport(x, y, width, height);` is called.
The cached object is immutable and won't get changed when viewport is called, right?

My search only gave `p5.RendererGL.js` as a place to fix.
It would be nice to have this cache management as a separate object, but since we're using ES5 syntax I don't know how nice that becomes.

Spongman:
i don't think we need anything more complicated than what's in d2a78fb6eeac7d507b6d225d91d95ff9295c06e6.

Zalastax:
Do you mean that the cache invalidation should be simple, or is not needed at all? I don't understand the code using the cached value, but I'm suspicious of using stale data. If the cache don't need to be invalidated I'd like a comment in the code explaining that.

Spongman:
there's no need for any invalidation. the renderer's _viewport object is updated immediately after the gl.viewport() call.

mlarghydracept:
Great work! I tested pretty thoroughly and this does get rid of unnecessary calls to `getParameter()`.

Thanks! 🎆

PR #2404, opened 2017-12-08
fix eslint:fix fix

Spongman:
i noticed that in a52950c90423b64c5cbc03526adb609f1511d9dc `Object.values` was changed to `Object.keys`.

i'm guessing this was due to compatibility with older versions of node?

anyway, unfortunately `grunt eslint:fix` now does nothing. this PR fixes that (in a non-es2017-requiring fashion).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -334,7 +334,7 @@ module.exports = function(grunt) {
 
   // eslint fixes everything it checks:
   gruntConfig.eslint.fix.src = Object.keys(gruntConfig.eslint)
-    .map(s => s.src)
+    .map(s => gruntConfig.eslint[s].src)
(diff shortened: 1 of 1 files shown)


lmccart:
ah, I didn't realize I just needed to update my node version. would it be preferable to stick with `Object.values`? I think that would be fine..

Spongman:
i think it's probably best to go with the backwards-compatible fix (this PR) just in case someone else with an older version of node runs into the same issue.

there's probably some really clever way to require a minimum version of node to build with, but i'm not sure if that's even a thing, or how one might do it.

lmccart:
@Spongman ok, yeah I had the same questions about node, let's stick with keys for now then.

PR #2406, opened 2017-12-10
Auto const docs

Spongman:
hmmm... looks like something got lost in the wash during d22d1c01cc85198df087636e1e10c6f76df65585. that's what was causing the mocha test failure.

here's a fix for that...


lmccart:
whoops sorry about that, looks all good with this, thanks!

PR #2408, opened 2017-12-11
make select() and createRadio() accept exisiting radio elements

### LINKED ISSUE #2299: Use existing dom elements with p5
#### Nature of issue?

- [ ] Found a bug
- [x] New feature request

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] DOM

#### New feature details:

It would be fine to use existing dom elements together with p5. 
Even though `select("#elem")` is working fine for div-Elements, it does not return a p5.Element.
When an existing element is selected, it does not provide any p5 features like listeners ( e.g. `mousePressed()` or `changed()`...)

Maybe it is possible to define a method like `registerElement(elem)`  or `new p5.Element(select(elem))` to equip existing elements with p5 features.

Here is my workaround for elements like radio-buttons:
```
function registerElement(id) {
  var elem = {};
  var tmp = createRadio();
  elem.elt = document.querySelector(id);
  elem.value = tmp.value;
  elem._events = tmp._events
  elem.changed = tmp.changed;
  tmp.remove();
  return elem;
}

function setup() {
  var radio = registerElement("#existing-radio-button");
  radio.changed(my_function);
}
```

outofambit:
Allow `Select()` to successfully `createRadio()` from existing elements.

Implementation
* add logic to `Select()` that inspects the given elements children to see if they are _only_ `input`s and `label`s
* modify `createRadio()` to take an element and use it instead of making new elements in the dom
* make `radio.selected()` and `radio.value()` more robust to different children configurations

Fixes #2299

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -683,7 +683,7 @@
    * }
    * </code></div>
    */
-  p5.prototype.createRadio = function() {
+  p5.prototype.createRadio = function(existing_radios) {
     p5._validateParameters('createRadio', arguments);
     // do some prep by counting number of radios on page
     var radios = document.querySelectorAll('input[type=radio]');
@@ -705,23 +705,28 @@
     }
     // see if we got an existing set of radios from callee
     var elt, self;
-    var arg = arguments[0];
-    if (typeof arg === 'object') {
+    if (typeof existing_radios === 'object') {
       // use existing elements
-      self = arg;
-      elt = this.elt = arg.elt;
+      self = existing_radios;
+      elt = this.elt = existing_radios.elt;
     } else {
       // create a set of radio buttons
       elt = document.createElement('div');
       self = addElement(elt, this);
     }
     // setup member functions
+    self._getInputChildrenArray = function() {
+      return Array.prototype.slice.call(this.elt.children).filter(function(c) {
+        return c.tagName === 'INPUT';
+      });
+    };
+
     var times = -1;
     self.option = function(name, value) {
       var opt = document.createElement('input');
       opt.type = 'radio';
       opt.innerHTML = name;
-      if (arguments.length > 1) opt.value = value;
+      if (value) opt.value = value;
       else opt.value = name;
       opt.setAttribute('name', 'defaultradio' + count);
       elt.appendChild(opt);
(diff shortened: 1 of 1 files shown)


lib/addons/p5.dom.js
```diff
@@ -190,7 +191,9 @@
       return new p5.MediaElement(elt);
     } else if ( elt.tagName === "SELECT" ){
       return createSelect( new p5.Element(elt) );
-    } 
+    } else if (children.length > 0 && children.every(function(c) { return c.tagName === "INPUT" || c.tagName === "LABEL" })) {
```
outofambit: does this seem like a safe assumption?

lib/addons/p5.dom.js
```diff
@@ -698,8 +702,19 @@
     else if (radios.length == 1){
       count = 1;
     }
-    var elt = document.createElement('div');
-    var self = addElement(elt, this);
+    // see if we got an existing set of radios from callee
+    var elt, self;
+    var arg = arguments[0];
```
lmccart: in the spirit of PR #2424, can we add a named argument to the signature and test it's type instead of extracting arguments[0]?

lib/addons/p5.dom.js
```diff
@@ -735,17 +740,12 @@
       }
       return opt;
     };
-    self.selected = function() {
+    self.selected = function(value) {
```
outofambit: @lmccart does this look right? i tried to add a `validateParameters` but ran into issues since they are not documented independently.

lib/addons/p5.dom.js
```diff
@@ -754,17 +754,12 @@
         }
       }
     };
-    self.value = function() {
+    self.value = function(value) {
```
outofambit: @lmccart also here

lib/addons/p5.dom.js
lmccart: ah yes i see the issue with validateParameters. i think it's fine to skip it in this case, at least for now.

lmccart:
looks great!

PR #2411, opened 2017-12-13
Eslint samples

Spongman:
this PR implements a custom `eslint` plugin in order to extract, lint (and in theory automatically fix) the samples in the doc comments. this is run during the `precommit` stage along with the other eslint stuff. most of the complexity in the plugin is due to the ` * ` prefix of the doc comments and how it messes with the column and character offsets in the linter's error messages and fix locations (see `utils/sample-linter.js` for more).

in order to commit this, of course, i had to fix all the lint errors.


lmccart:
awesome

PR #2423, opened 2017-12-18
fix bogus 'this' reference in createCanvas

Spongman:
there's an erroneous use of `this.` inside an anonymous function in `createCanvas()`.

this PR fixes this `this`.

also, fix a case where `resizeCanvas()` would crash trying to restore some read-only canvas properties.


lmccart:
good catch!

PR #2424, opened 2017-12-19
avoid use of 'arguments' object where possible

Spongman:
this PR redices the use of the `arguments` object wherever possible. it:
- uses declared parameters instead of indexing the `arguments` arrray
- checks parameter  using `typeof` instead of `arguments.length`
- avoids duplicating the `arguments` array

use of `arguments` is sometimes unavoidable, especially when handling overloads that include out-of-order optional parameters.

### DIFF
--- a/src/core/2d_primitives.js
+++ b/src/core/2d_primitives.js
@@ -78,12 +78,7 @@ require('./error_helpers');
  *
  */
 p5.prototype.arc = function(x, y, w, h, start, stop, mode) {
-  var args = new Array(arguments.length);
-  for (var i = 0; i < args.length; ++i) {
-    args[i] = arguments[i];
-  }
-
-  p5._validateParameters('arc', args);
+  p5._validateParameters('arc', arguments);
   if (!this._renderer._doStroke && !this._renderer._doFill) {
     return this;
   }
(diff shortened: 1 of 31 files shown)


src/core/2d_primitives.js
```diff
@@ -233,21 +215,12 @@ p5.prototype.ellipse = function() {
...
  */
 p5.prototype.line = function() {
-  if (!this._renderer._doStroke) {
-    return this;
-  }
-  var args = new Array(arguments.length);
-  for (var i = 0; i < args.length; ++i) {
-    args[i] = arguments[i];
-  }
+  p5._validateParameters('line', arguments);
```
lmccart: I think these call should generally only happen if the _doStroke and _doFill checks pass, to save time. What do you think?

lmccart:
this is great! just the one question about the placement of validateParameters()

src/core/2d_primitives.js
Spongman: hmmm... you're right, the parameters aren't used if the renderer's current state doesn't match certain requirements - and this is true for several other methods throughout p5. but... i think if I were using this api and i was passing the wrong stuff to it, i'd want to know as soon as possible that i was making an error. if the renderer's state depends on some non-deterministic process and only triggers this validation error rarely then I'd have a Heisenbug on my hands :(

i believe that the perf implications of doing this every time should be reasonably minimal. especially in the minimized build, it's essentially a no-op.

PR #2426, opened 2017-12-20
Added example for volume method

odm275:


Spongman:
Weird. The precommi hook should have to n. Try running `npm lint:fix`

lib/addons/p5.dom.js
```diff
@@ -1837,6 +1837,70 @@
...
+	 *     background(colorLow);
+	 *     text('volume(0.1)', width / 2, height / 2);
+	 *   }
+	 * }
+	 * 
+	 * function mouseClicked() {
+	 *   if (mouseX > 0 && mouseX < width && mouseY < height && mouseY > 0) {
+	 *     counter++;
+	 *     if (counter === 0) {
+	 *       high();
```
lmccart: rather than defining the functions above, you could just put the single lines of code right here in the if statements. this will simplify the example a bit.

lib/addons/p5.dom.js
```diff
@@ -1837,6 +1837,70 @@
...
+	 *     background(colorMid);
+	 *     text('volume(0.5)', width / 2, height / 2);
+	 *   } else if (counter === 2) {
+	 *     background(colorLow);
+	 *     text('volume(0.1)', width / 2, height / 2);
+	 *   }
+	 * }
+	 * 
+	 * function mouseClicked() {
+	 *   if (mouseX > 0 && mouseX < width && mouseY < height && mouseY > 0) {
```
lmccart: we can eliminate this check for simplicity

lib/addons/p5.dom.js
```diff
@@ -1837,6 +1837,70 @@
...
+	 *   } else if (counter === 1) {
+	 *     background(colorMid);
+	 *     text('volume(0.5)', width / 2, height / 2);
+	 *   } else if (counter === 2) {
+	 *     background(colorLow);
+	 *     text('volume(0.1)', width / 2, height / 2);
+	 *   }
+	 * }
+	 * 
+	 * function mouseClicked() {
```
lmccart: let's use mousePressed to be consistent with other examples

lib/addons/p5.dom.js
```diff
@@ -1837,6 +1837,70 @@
    * @param {Number}            [val] volume between 0.0 and 1.0
    * @return {Number|p5.MediaElement} current volume or p5.MediaElement
    * @method volume
+   * @example
+	 * <div>
+	 * <code>
+	 * var audio;
+	 * var colorLow;
```
lmccart: to simplify, we can remove these color variables and just call background directly with the rgb values in draw

odm275:
Added the changes. Ref: #2498. How do I re-open a pull request >GitHub novice here.

lmccart:
got it thanks!

PR #2427, opened 2017-12-20
webgl: normalize normals

### LINKED ISSUE #2420: webgl: scaled normal calculation bug
i think something's up with how the webgl bit handles geometry normals and scaling.

this example illustrates this: https://codepen.io/Spongman/pen/KZVwRy?editors=0010.

image

the squares are rotated 45degrees around the Y axis, and lit from the left with a `directionalLight`.

the top square is being resized by passing a different parameter into the `plane()` method.
the bottom square is being scaled using the `scale()` method.

the results should be identical. but as is clear from watching the animation, the resultant normals of the scaled square are incorrect, causing the light to vary as the scale changes.

Spongman:
the face normals, after they have been transformed by the view matrix, need to be normalized. otherwise scaling effects in the view matrix will throw off the lighting calculation.

fixes #2420.

### DIFF
--- a/src/webgl/shaders/light.vert
+++ b/src/webgl/shaders/light.vert
@@ -30,7 +30,7 @@ void main(void){
   vec4 positionVec4 = vec4(aPosition, 1.0);
   gl_Position = uProjectionMatrix * uModelViewMatrix * positionVec4;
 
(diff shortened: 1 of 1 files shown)


mlarghydracept:
Nice fix!

PR #2428, opened 2017-12-21
webgl: lighting fixes

### LINKED ISSUE #2414: only the final light source (pointLight, directionalLight) is rendered
- Found a bug
- WebGL
- Desktop
- p5.js version: 0.5.16
- Web browser and version: chrome Version 63.0.3239.84 (Official Build) (64-bit)
- Operating System: Mac OSX 10.11.6

In the following simple example which positions two point lights (one on the left, one on the right) the second call to `pointLight()` overrides the first, rather than adding a new light.

```javascript
function setup() {
  createCanvas(400, 300, WEBGL);
}

function draw() {
  background(175);
  pointLight(0, 0, 255, -200, 0, 0);
  pointLight(255, 0, 0, 200, 0, 0);
  ambientMaterial(255);
  noStroke();
  sphere(50);
}
```

With the red light second, i.e.

```javascript
  pointLight(0, 0, 255, -200, 0, 0);
  pointLight(255, 0, 0, 200, 0, 0);
```

screen shot 2017-12-14 at 5 58 30 pm

With the blue light second:
```javascript
  pointLight(255, 0, 0, 200, 0, 0);
  pointLight(0, 0, 255, -200, 0, 0);
```
screen shot 2017-12-14 at 5 58 40 pm

This bug appeared somewhere between 0.5.14 and 0.5.16. Using 0.5.14, both lights are active:

screen shot 2017-12-14 at 6 00 23 pm

The same bug exists for `directionalLight()`.

This may be related to: #2251 and #1622 

(this issue was filed by the viewers of The Coding Train live!)

Spongman:
this PR adds back support for multiple lights of different types.

fixes #2414

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -187,7 +187,12 @@ module.exports = function(grunt) {
       // Watch the codebase for doc updates
       // launch with 'grunt requirejs connect watch:yui'
       yui: {
-        files: ['src/**/*.js', 'lib/addons/*.js'],
+        files: [
+          'src/**/*.js',
+          'lib/addons/*.js',
+          'src/**/*.frag',
+          'src/**/*.vert'
+        ],
         tasks: ['browserify', 'yuidoc:prod', 'minjson', 'uglify'],
         options: {
           livereload: true
--- a/src/webgl/light.js
+++ b/src/webgl/light.js
@@ -62,32 +62,27 @@ var p5 = require('../core/core');
  * @chainable
  */
 p5.prototype.ambientLight = function(v1, v2, v3, a) {
-  if (!this._renderer.curFillShader.isLightShader()) {
-    this._renderer.setFillShader(this._renderer._getLightShader());
-  }
 
-  var color = this._renderer._pInst.color.apply(
-    this._renderer._pInst,
-    arguments
-  );
+  var color = this.color.apply(this, arguments);
 
+  var shader = this._renderer._useLightShader();
+  
   //@todo this is a bit icky. array uniforms have
   //to be multiples of the type 3(rgb) in this case.
   //a preallocated Float32Array(24) that we copy into
   //would be better
-  var colors = new Float32Array(color._array.slice(0, 3));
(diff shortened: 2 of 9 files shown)


Spongman:
also added a fix for the pointLight location calculation in the lighting shader.

fixes #1622

mlarghydracept:
Cool thanks for the work! I was testing and I ran into a couple of issues. I would check out the manual tests for WebGL on this one. It breaks a couple of those that are using immediate mode with fill. The function `_drawFillImmediateMode()` doesn't seem to have access to the vertex color shader object. Something maybe isn't being reset when you are switching the fill shader to a light shader?

Also I would check out the tests in the lights section. These tests are little outdated and could benefit from updating, but that aside once I checkout your branch I don't see a point light in the aptly named point light example.

Spongman:
ok, i'll take a look, thanks!

Spongman:
I don't think the pointLight is not related to this PR. the scaling of the pointLight coordinate space changed in #2139.

Also, i don't think the other issue is related to this PR either.

Can you check master to see if you see those issues?

Spongman:
BTW: #2435 is fixed by one of my other pending PRs, #2349.

mlarghydracept:
Those issues aren't present in master. I checked again just now with the very newest commits and the two problems I mentioned aren't present. When I switch from master to your lighting-fixes branch I get sketch breaking errors on `camera/box/` test and `immediateMode/minTextureAndFill` test. I also get different lighting in `lights/multipleLights/` and `lights/pointLight`. Hopefully this helps!

Spongman:
ok, that should fix the immediate mode issue.

the other lighting differences you are seeing are cased by one of:
1) the pointLight coordinate system change (#2139), or
2) the fact that lighting was broken before (#1622) and is now fixed.

Spongman:
I added another PR to fix the coordinate systems in the other two tests that use pointLights.

mlarghydracept:
Okay tested pretty thoroughly. Thanks for these smart changes!

shiffman:
Thanks for this wonderful fix!

PR #2431, opened 2017-12-21
Fix sample-linter on Windows

Zalastax:
Without this, sample-linter will use `\r\n` on Windows, clashing with the [linebreak style rule](https://eslint.org/docs/rules/linebreak-style). To try it out yourself, set `var EOL = '\r\n'`.

@Spongman before this change I also got a crash in the linting rule from here:
```
var startLine = msg.line + sampleLine;
msg.column += this.lines[startLine].prefixLength;
```
This is because prettier was complaining about the `\r`, which is beyond the lines of the original file, causing `this.lines[startLine]` to be `undefined`. Do you think there is any point to checking that `startLine` is not out of bounds?

### DIFF
--- a/utils/sample-linter.js
+++ b/utils/sample-linter.js
@@ -57,7 +57,9 @@ function splitLines(text) {
   return lines;
 }
 
-var EOL = require('os').EOL;
+// We're enforcing Unix line endings
+// See https://eslint.org/docs/rules/linebreak-style
+var EOL = '\n';
 
 var userFunctions = [
   'draw',


Spongman:
no. this is bad. the eslint `linebreak` rule is not recommended. the default git installation on windows translates to/from CRLF correctly. if you're using windows and your p5.js files are not CRLF terminated then you've messed something up. please do not merge this.

Spongman:
it may be worth configuring autocrlf to true in the repository just to ensure that this isn't a problem for people whose global configs are bad.

Zalastax:
I was wrong about what rule is causing this. It's actually prettier and it's reading the [.editorconfig](https://github.com/processing/p5.js/blob/6511d628967307add82ff10d68f9a13d0ca0a6bf/.editorconfig#L15). Personally, I never use CRLF. Reading [dealing with line endings](https://help.github.com/articles/dealing-with-line-endings/), I see that we can commit a special `.gitattributes` file, overriding an individual's `core.autocrlf`. I'll add it to this PR if you agree to that solution.

Spongman:
yes, indeed. it _used_ to be that autocrlf was recommended to be `false` due to bugs in git's crlf handling. but those have since been fixed and it's now recommended to use `autocrlf=true` on windows in _most_ cases and explicitly protect binary files if necessary - as noted in those github recommendations.

that's weird about the prettifier and .editorconfig, though. i have `autocrlf=true` set globally on my windows machines, all the files in my p5.js repos on those machines have CRLF accordingly, yet the prettier does not attempt to change them to LF and I have never seen any EOL-related issues. If you do not have `autocrlf=true`, that may be the cause of your issues.

yes, we should probably create a .gitattributes class to override the user's local settings, otherwise we'd need to somehow correctly determine the user's ambient EOL setting in the sample linter plugin which sounds easier said than done.

Zalastax:
I do have `autocrlf=false`, and we *should* add a `.gitattributes` file. I tried it out, and as you said git handles it perfectly. Prettier does not handle it at all though, which might be related to https://github.com/prettier/prettier/issues/537. These are my changes: https://github.com/Zalastax/p5.js/tree/crlf-test. I'd prefer not even looking into this and just setting `autocrlf=false`.

Spongman:
the problem is that `autocrlf=false` on windows is not a good idea unless you really know what you're doing, because many windows users' tools/editors will automatically convert to CRLF (since that's what's native to the platform) -- the prettier bug is a good example of this. you then need to have a bunch of verification tools run to make sure that these changes don't make it into the repo. very messy.

or you could just set `autocrlf=true`.

just curious, what version of git are you using?

Zalastax:
I don't have any problems with git, it's doing its job perfectly. I'm actually quite impressed how fast and problem free it was. Prettier is unfortunately going bananas if it finds carriage returns.

I've never had a tool convert files from LF to CRLF and the linter will make sure it never gets into the repo. If prettier didn't warn on CRLF I'd be fine with your suggestion, even though I really prefer that my VCS gives me my files without any conversions. Setting `autocrlf=false` means we don't need to fix prettier and we don't have to identify what file endings are text files.

Spongman:
really. trust me. `autocrlf=true` is the way to go.

Zalastax:
I think I've got my head sorted out now. Prettier works as long as you are consistently using LF or CRLF. As far as I can see, there is no way for us to force autocrlf to be either false or true from the repository level.

Most people are probably using `autocrlf=true` so just documenting that they might need to override their globals if they have `autocrlf=false` seems fine.
Alternatively, since the problem is just in sample-linter, we could normalize the line endings we emit from there. That way people don't need to change anything at all. I propose that I make a new PR where:
* .gitatributes is added
* sample-linter normalizes its emitted line endings

pflannery:
I just ran in to this issue whilst trying to the generate the latest build files. 
I see the following error when running `grunt` on Windows 10
`Warning: Cannot read property 'prefixLength' of undefined Use --force to continue.`

I'm currently managing to get around this by skipping the lint tasks and just running `grunt yui build` to just generate the files..

I've always setup git to use `autocrlf=input` on all my Windows machines. (This isn't "bad" it's a choice I prefer)

- All my code and text editors support LF very well.
- I use LF in all my projects on windows (last few years now)
- I find the majority of other peoples work are built with LF and therefore potentially fail when auto converting to CRLF
- autocrlf=input allows me to still work with CRLF projects without any issues

Either "adding gitattributes with autocrlf=true" **OR** "sample-linter normalizes its emitted line endings" works for me :P

Zalastax:
@pflannery, there's unfortunately no way to set autocrlf=true in the gitattributes. My suggestion in https://github.com/processing/p5.js/pull/2484 is to guide Windows users to set autocrlf when they clone the repo. I'll let @Spongman decide on what solution is most appropriate.

pflannery:
@Zalastax oops, your right.

I did a test with `*        text eol=native` in the .gitattributes file and seems to work

My steps were:
- Add `*        text eol=native` to the .gitattibutes file
- Commit the .gitattibutes file
- Run `rm .git/index` (at the root of the project. note: this will lose any pending changes)
- Run `git reset`
- I now see all the files with CRLF on windows

IMO it's probably best to set the project to have one expected line ending for everyone regardless of OS. I'm more in favour of ensuring all the files and the linting config use LF.

pflannery:
I'm not seeing this issue on my machine 😆 

image

Zalastax:
I'm closing this since it's not the right solution. Instead merge #2484 or/and #2505

Issue #2422, opened 2017-12-18
Inconsistent Angle Mode in WEBGL for rotateX, rotateY, rotateZ


PixelBaer:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: p5.js v0.5.16 October 11, 2017
- Web browser and version: Chrome 63.0.3239.84
- Operating System: Windows 10 Pro
- Steps to reproduce this:

The documentation says rotateX/Y/Z accepts only radiants as arguments. But if you change angleMode to DEGREES it seems to work with the Y and Z axis but not with the X axis.

Here's a simple demo with sliders for X/Y/Z rotation. Notice how  y and z works differently than x.

```javascript
var div, divX, divY, divZ, sliderX, sliderY, sliderZ;

function setup() {
  createCanvas(600, 400, WEBGL);
  background(51);
  angleMode(DEGREES);
  
  div = createDiv('Sliders:');
  divX = createDiv('X: ');
  divY = createDiv('Y: ');
  divZ = createDiv('Z: ');
  divX.parent(div);
  divY.parent(div);
  divZ.parent(div);
  sliderX = createSlider(0, 180, 30, 0.1);
  sliderX.parent(div);
  sliderY = createSlider(0, 360, 45, 0.1);
  sliderY.parent(div);
  sliderZ = createSlider(0, 360, 0, 0.1);
  sliderZ.parent(div);
}

function draw() {
  background(51);
  
  divX.html('X: '+sliderX.value());
  divY.html('Y: '+sliderY.value());
  divZ.html('Z: '+sliderZ.value());
  
  rotateX(sliderX.value());
  rotateY(sliderY.value());
  rotateZ(sliderZ.value());
  
  noStroke();
  normalMaterial();
  box(50, 100, 120); 

}
```

mlarghydracept:
[...]

PR #2439, opened 2017-12-23
fix matrix angle modes

Spongman:
in order to fix matrix rotation use of angleMode, this PR ensures that the `p5` instance is threaded through matrix creation/manipulation.

it also moves the `p5` parameter to the `p5.Matrix` constructor to the last parameter in order to simplify the calling & handling of that parameter.


Spongman:
fixes #2422

mlarghydracept:
Hi! I was just running through and testing with the manual tests first and we are losing a reference to the matrix object somewhere. I am not sure if it is the only culprit but I would look at the beginning of `p5.Matrix.rotate()`. The simplest way to recreate the problem - 
```
function setup() {
}
function draw(){
rotateX(frameCount * 0.02);
sphere(100);
}
```

Spongman:
oh. hold on. i messed up.

Spongman:
noooooooooooooo........

mlarghydracept:
Nice fix!

PR #2444, opened 2017-12-23
Added example for p5.MediaElement.clearCues()

bansalnitish:
Addresses #1954


lmccart:
This is great, but I might suggest moving the clearCues() and the addCue() calls after it to a mousePressed() function, otherwise the use feels a little confusing. That is, why addCue() just to clearCues() immediately after.

bansalnitish:
@lmccart I have done as you requested.

PR #2446, opened 2017-12-23
fromAngle length parameter, 3d fromAngles

Spongman:
- add optional `length` parameter to `p5.Vector.fromAngle()`
- update `fromAngle()` sample to use it,
- add 3d  `p5.Vector.fromAngles()` + sample

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -907,7 +907,7 @@ p5.Vector.fromAngle = function fromAngle(angle, length) {
  * @param {Number}     [length] the length of the new vector (defaults to 1)
  * @return {p5.Vector}       the new p5.Vector object
  * @example
- * <div>
+ * <div modernizr='webgl'>
  * <code>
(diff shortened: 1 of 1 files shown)


lmccart:
@Spongman can you explain the decision to add the optional length param here? is it needed for the 3D version of this method?

Spongman:
No, it's not related to the 3D version. It just extends fromAngle() to cover the entire plane with polar coordinates instead of just the unit circle - making it useful in many more situations.

lmccart:
yup makes sense to me. thanks!

lmccart:
@Spongman I forgot about the setMag() function as mentioned by @iridule here https://github.com/processing/p5.js/pull/2447#issuecomment-354376498. i am feeling that it makes sense to keep the api more simple and keep setMag() and fromAngle() separate, rather than duplicating the functionality of setMag() in fromAngle(). are you ok this?

PR #2447, opened 2017-12-23
new method for creating p5.Vector from polar coordinates

iridule:
Hey,

I thought it would be incredible useful to have a method to create vectors from polar coordinates
to be built into the p5 library.
I have added the method into the vector src file.
let me know if this is something that is worth adding or not, perhaps it could be improved...

`let myVector = p5.Vector.fromPolar(100, PI);`

best

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -1157,4 +1157,17 @@ p5.Vector.mag = function mag(vecT) {
   return Math.sqrt(magSq);
 };
 
+/**
+ * @method fromPolar
+ * @param {Number} radius
+ * @param {Number} angle
+ * @return {p5.Vector}
+ */
+p5.Vector.fromPolar = function(radius, angle) {
+  let x = radius * cos(angle);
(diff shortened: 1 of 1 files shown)


Spongman:
;-) there is already a `p5.Vector.fromAngle` method that does this. nearly...

i added a PR (#2447) earlier that adds a `length` parameter to the `fromAngle` method.

Spongman:
;-) there is already a `p5.Vector.fromAngle` method that does this. nearly...

i added a PR (#2447) earlier that adds a `length` parameter to the `fromAngle` method.

austince:
#2449 I believe you meant!

Spongman:
actually, #2446 ;-)

iridule:
oh no I realised after posting :stuck_out_tongue: ! it would be useful to have the length feature implemented... thought this works really well anyway:

`p5.Vector.fromAngle(0).setMag(10)`

lmccart:
thanks! this should be covered with the new length feather and fromAngle

PR #2449, opened 2017-12-24
webgl: Per pixel lighting

Spongman:
`setAttributes('perPixelLighting', true)` enables use of a Phong specular lighting model.

before & after:
image

this PR includes and relies on some other PRs:
- #2428 lighting-fixes
- #2433 implicit normals (for primitive geometry)
- #2446 fromAngles (for the getAttributes sample)

this PR should simplify significantly if those are merged first.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -187,7 +187,12 @@ module.exports = function(grunt) {
       // Watch the codebase for doc updates
       // launch with 'grunt requirejs connect watch:yui'
       yui: {
-        files: ['src/**/*.js', 'lib/addons/*.js'],
+        files: [
+          'src/**/*.js',
+          'lib/addons/*.js',
+          'src/**/*.frag',
+          'src/**/*.vert'
+        ],
         tasks: ['browserify', 'yuidoc:prod', 'minjson', 'uglify'],
         options: {
           livereload: true
--- a/src/webgl/light.js
+++ b/src/webgl/light.js
@@ -137,8 +137,8 @@ p5.prototype.ambientLight = function(v1, v2, v3, a) {
  *   background(0);
  *   //move your mouse to change light direction
  *   var dirX = (mouseX / width - 0.5) * 2;
- *   var dirY = (mouseY / height - 0.5) * -2;
- *   directionalLight(250, 250, 250, dirX, dirY, 0.25);
+ *   var dirY = (mouseY / height - 0.5) * 2;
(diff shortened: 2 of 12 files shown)


mlarghydracept:
Cool! A few thoughts as I am testing everything and preparing to merge your various awesome pull requests over the next few days - 

Can you make sure to label something if it is a work-in-progress?
Also do you think I should address your pending GL pull requests in a specific order? Should I assume the three you mentioned above are first in line? Just want to make sure there are as few conflicts as possible with merge order.

Spongman:
awesome! yes, all these PRs are ready to commit (aside from all the bugs i missed, of course ;-)

if you merge those three PRs first (they're all independent), this one should be much simpler. i'll try to fix any conflicts that come up if i catch them in time.

kjhollen:
😻 😻 😻 wow these results look awesome! @mlarghydracept do you want to review this one?

mlarghydracept:
Yeah I agree. That shader has fantastic results! I would say it might be a good idea to write a manual test for this one. Something that switches in and out of the phong specular. Just to have something to test with down the line. Otherwise I would say this looks good to merge but I would like to merge #2446 first. I don't feel qualified to evaluate that one so I pinged @lmccart for a review of it. Will merge this as soon as that is merged.

lmccart:
@mlarghydracept #2446 has been merged

mlarghydracept:
Okay cool just tested a bunch and this is fantastic! Ready for a merge except two really small requests:

1: Could you put a short instruction in the perPixelLighting manual test? Just for future contributors to know that they should click to test. You can put it in the HTML if that is easier.

2. Also in the manual-test, can you load the teapot model in [preload()](https://p5js.org/reference/#/p5/preload)? I know that several other manual tests don't this but it is something I realize we should fix as we go because sometimes it is hard to test on slower machines because you can't tell if it is broken while loading.

Thanks again!

mlarghydracept:
Thank you!

sanketsingh24:
@Spongman  I just wanted to know your logic behind using [500 in line 89 of phong.frag](https://github.com/processing/p5.js/pull/2449/files#diff-38ed3d0faa51831a518d47b0717d49bbR89). I am implementing a `falloff` function which will set custom attenuation, so just wanted to know about this.

Spongman:
oh you can ignore my code there... it's not correct because it's not scale-invariant. just replace it with whatever you have. i had a PR a while back that included an implementation of Processing's `falloff`, but it didn't make it in, unfortunately.

PR #2454, opened 2017-12-24
Corrected if condition and added example for removeCue()

bansalnitish:
The if condition in the removeCue() function was checking (this_cue[i] == id) instead of (this_cue[i].id == id). I had correctly replaced the if condtion and also  added the example for the same. 
Addresses #1954


lmccart:
great!

PR #2463, opened 2017-12-28
ortho() implementation

diwi:
Finished and fixed ortho().

1) The former args check failed when numbers are 0 (= not undefined, or null, etc ...), and as a result a wrong matrix was created.

2) ortho() prototype for the gl renderer


src/webgl/camera.js
```diff
@@ -275,36 +275,43 @@ p5.RendererGL.prototype.perspective = function(fovy, aspect, near, far) {
...
+  this._renderer.ortho.apply(this._renderer, arguments);
+  return this;
+};
 
-  var x = 2.0 / w;
-  var y = 2.0 / h;
-  var z = -2.0 / d;
+p5.RendererGL.prototype.ortho = function(l, r, b, t, n, f) {
+  // ortho() ... top/left = 0/0
+  if(l === undefined) l = 0;
```
kjhollen: thank you for this fix! I'm noticing that these values were also changed from the former defaults (-this.width/2, for example, for left)—can you expand on the reasoning for this? in GL mode 0,0 is in the center of the canvas.

I would also prefer if we can keep the expanded variable names for left, right, bottom, top, etc., which I think makes the code more readable.

Zalastax:
The errors come from the source not being [formatted correctly](https://github.com/processing/p5.js/wiki/Development#code-style). You can fix this automatically by running `npm run lint:fix`.

Just to make sure our setup is working, have you run `npm install` on your machine before committing? Doing so should install a pre-commit hook that stops ill-formatted code from being committed 😃

PR #2464, opened 2017-12-28
docs: added examples for p5.Element.class

austince:
Adding two examples for `p5.Element.class` as listed here: #1954 🎊

Perhaps note in `pressed` doesn't belong here but I thought it was a nice way to show a way the zero-args overload was useful. Could make it more useful with `add/removeClass`

### DIFF
--- a/src/core/p5.Element.js
+++ b/src/core/p5.Element.js
@@ -167,16 +167,16 @@ p5.Element.prototype.id = function(id) {
  * function setup() {
  *   createP('No matter how small.')
  *      .class('seuss')
- *      .mousePressed(clicked);
+ *      .mousePressed(pressed);
  *   createP("But someone, I ain't sayin' who, Has got a guilty face.")
  *      .class('shel')
- *      .mousePressed(clicked);
+ *      .mousePressed(pressed);
  * }
  *
- * function clicked() {
+ * function pressed() {
  *  // `this` refers to the clicked element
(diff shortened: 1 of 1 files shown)


austince:
Having some trouble with the linter. Can't find task `eslint-samples:source` using the grunt-cli

src/core/p5.Element.js
```diff
@@ -145,7 +145,42 @@ p5.Element.prototype.id = function(id) {
...
+ * function setup() {
+ *   const personText = createP("A person's a person");
+ *   personText.class('seuss');
+ *   createP('No matter how small.').class('seuss');
+ * }
+ * </code>
+ * </div>
+ *
+ * @alt
+ * no display.
```
lmccart: can you please add a text description of what is visible in the example?

src/core/p5.Element.js
```diff
@@ -145,7 +145,42 @@ p5.Element.prototype.id = function(id) {
...
+ * function pressed() {
+ *  // `this` refers to the clicked element.
+ *  // Sets the class name to the pressed element's name plus '-pressed'
+ *  this.class(this.class() + '-pressed');
+ * }
+ * </code>
+ * </div>
+ *
+ * @alt
+ * no display.
```
lmccart: can you please add a text description of what is visible in the example, and what changes with mouse interaction? this makes the example accessible for seeing impaired users.

lmccart:
we're all set for an example for this one now, thanks!

PR #2471, opened 2017-12-29
universal parameter validation

Spongman:
this PR:
- removes all ad-hoc calls to _validateParameters => zero overhead in minified builds!
- for non-minified builds, at startup, hooks all(*) public documented methods with a call to validateParameters.

(*) except `httpDo`, `createCapture` and the `p5.sound` stuff.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -75,7 +75,6 @@
    *
    */
   p5.prototype.select = function(e, p) {
-    p5._validateParameters('select', arguments);
     var res = null;
     var container = getContainer(p);
     if (e[0] === '.') {
@@ -146,7 +145,6 @@
    *
    */
   p5.prototype.selectAll = function(e, p) {
-    p5._validateParameters('selectAll', arguments);
     var arr = [];
     var res;
     var container = getContainer(p);
@@ -234,7 +232,6 @@
    *
    */
   p5.prototype.removeElements = function(e) {
-    p5._validateParameters('removeElements', arguments);
     for (var i = 0; i < this._elements.length; i++) {
       if (!(this._elements[i].elt instanceof HTMLCanvasElement)) {
         this._elements[i].remove();
@@ -322,7 +319,6 @@
    * </code></div>
    */
   p5.prototype.createImg = function() {
-    p5._validateParameters('createImg', arguments);
     var elt = document.createElement('img');
     var args = arguments;
     var self;
@@ -365,7 +361,6 @@
    * </code></div>
    */
   p5.prototype.createA = function(href, html, target) {
-    p5._validateParameters('createA', arguments);
     var elt = document.createElement('a');
     elt.href = href;
     elt.innerHTML = html;
@@ -418,7 +413,6 @@
    * </code></div>
    */
   p5.prototype.createSlider = function(min, max, value, step) {
-    p5._validateParameters('createSlider', arguments);
     var elt = document.createElement('input');
     elt.type = 'range';
     elt.min = min;
@@ -461,7 +455,6 @@
    * </code></div>
    */
   p5.prototype.createButton = function(label, value) {
-    p5._validateParameters('createButton', arguments);
     var elt = document.createElement('button');
     elt.innerHTML = label;
     if (value) elt.value = value;
@@ -495,7 +488,6 @@
    * </code></div>
    */
   p5.prototype.createCheckbox = function() {
-    p5._validateParameters('createCheckbox', arguments);
     var elt = document.createElement('div');
     var checkbox = document.createElement('input');
     checkbox.type = 'checkbox';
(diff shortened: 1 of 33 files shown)


lmccart:
when @almchung was developing this, we made the decision not to include validateParams in every function and instead to strategically place it in ones where we felt errors were likely. the reason being that the validateParams causes a real performance hit, and we don't necessarily want to do that for every function call when some are pretty straightforward.

Spongman:
you might be surprised at how little a performance hit this now incurs. the `validateParameters` call is significantly faster than it once was since most of the parsing & extraction of the information in `data.json` is cached. also, we can explicitly exclude any performance-sensitive methods from parameter validation just by adding them to the `skipValidation` array.

i have another change coming for the `validateParameters` code that fixes some issues with `undefined` parameters, allows parameter validation error to cause errors in the mocha tests, and further reduces the overhead of the validateParameters call. maybe we can revisit this after that.

lmccart:
ok ill take a look at that one first, thanks

kjhollen:
Thank you for putting together this proposal. I looked at this for the first time today and decided to close it because it hides some important code from less experienced contributors. Please open an issue to discuss the scope of the friendly error system if you'd like to resubmit.

Spongman:
huh?

kjhollen:
It's valuable for newcomers to the project to see where the parameter validation happens with an explicit call in functions where it is needed; it makes the codebase of the project itself more legible. It means we lose some automation in the interest of transparency, but that seems okay as a trade-off given this library's emphasis on being an educational tool with a lower barrier to entry for both users and contributors.

The original idea for the friendly error system was to provide meaningful suggestions to users working with p5.js to learn to program for the first time; it was not intended to wrap every function in the library. At present, the automation of the process already has lost some specificity in the error messages generated. (For example, a call to `ellipse(100, 100)` could tell you that you have to add the diameter as a third parameter rather than `ellipse() was expecting Number for parameter #2 (zero-based index), received an empty variable instead`, which is pretty tough to understand for a new programmer). I'd love to see some effort on making that better, rather than covering the entire library with these generic messages.

I suggest opening an issue to discuss further and collect more requirements from contributors and library users.

Spongman:
a few thoughts: apologies for the random nature...

i'm not sure i understand the statement that everything in the library has to be explicit. there are plenty of things that are done that use an almost identical mechanism to what has been used here. for example, most `p5` methods are copied into the `window` prototype in global mode, but there isn't a line of code in every library method to do this. there are a bunch of methods that are wrapped specifically to handle `preload` callbacks, yet there isn't specific code in each of those methods to do that. `p5.Graphics` shares a whole bunch of methods with `p5`, however there aren't explicit definitions of those methods in `p5.Graphics.prototype`, it's just done with a for loop and a function proxy (`bind`). maybe there's a difference here i'm not seeing.

those existing `_validateParameters` calls really don't add anything of value to those methods, but they are one more error-prone thing that needs to be maintained, and are a source of unnecessary/confusing validations that occur when library methods call other library methods (see ca5429a above for a commit which addresses this performance/usability issue). another benefit of this PR (and #2782) is that almost all library methods have their parameters validated during the unit test runs - and validation errors result in test failures. several PRs containing fixes to bugs highlighted during such test runs have already been merged (i have an addition to this PR that also validates the return values of functions during test runs which also resulted in a slew of bug fixes).


> The original idea for the friendly error system was to provide meaningful suggestions to users working with p5.js to learn to program for the first time

yeah, i agree this is a very important

> it was not intended to wrap every function in the library

i understand that, and this PR does not attempt to do that. however, i'm not sure why validating more functions would necessarily impinge on the original FES goal you stated above. regardless, due to  improvements in the docs and in the FES system itself, the number of functions that are _not_ currently checked by FES is vanishingly small - in fact it's not much larger than the set of functions this PR excludes from validation (maybe for performance, incompatibility with FES (eg `httpDo`), or for those methods that do their own custom parameter validation). i _do_ however think there is value in validating as many library methods as possible (barring performance/compatibility issues) since this leads to a better path to success for the library user, helps them to write code that works now and will continue to work with future releases of the library, and maybe in the future take advantage of a tighter integration with the web editor.


> the automation of the process already has lost some specificity

i'm not quite sure what you're saying here. are you saying that the original FES system eliminated a more comprehensive system that existed before it? i wasn't aware anything existed before FES. or are you saying that changes to the FES system have made it less effective? i hope this isn't the case.

---

I agree that the FES error messages could definitely use some work. please see my PR https://github.com/processing/p5.js/pull/2780 that tries to address issues like your `ellipse` example above (along with those in #2740) by significantly improving the performance (objectively) and applicability (arguably) of these error messages in many circumstances, (eg: `ellipse() was expecting at least 3 arguments, but received only 2`).

lmccart:
Hi @Spongman could we talk further off-thread? Please drop me an email at hello@p5js.org. Thanks!

Issue #2252, opened 2017-10-10
Bug with rect in WEBGL mode.


a-y-u-s-h:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 
Not sure if this is a legit bug. I was just checking out behavior of rect in WEBGL and I ran this code first to oscillate rectangles : 

![1](https://user-images.githubusercontent.com/19586719/31378898-4f481c54-adca-11e7-9a71-409e83e66159.png)

..but the output was too slow

But when I used translate : 

![4](https://user-images.githubusercontent.com/19586719/31378727-d7a5c318-adc9-11e7-95c7-cac72a549841.png)

It was all good.

- p5.js version: 0.5.14
- Web browser and version:Chrome 61.0.3163.91 (Official Build) (64-bit)
- Operating System: Linux - Ubuntu 17.04
- Steps to reproduce this: As done in the images above.

Siphonophora:
Hi, 

Whats happening here is that the first option is generating every rectangle from scratch (because they each have a unique set of arguments), while the second sketch is reusing a single rectangle. Probably not exactly a bug, but also not ideal. 

It is something I think I hadn't mentioned in #2226 yet, but this is probably part of that discussion. It doesn't make sense for the X and Y coordinates to be in the geometry ID, except that we need them to properly place the geometry. Thoughts on whether we can include only H and W in the geometry ID and then take care of the translation for the user? Similar issues would apply for someone drawing a bunch of triangles or quads.

kjhollen:
@Siphonophora this is spot on. it might be worth doing a test to see if immediate mode (using beginShape + vertex + endShape) is now fast enough that we can use that for some of the 2D primitives. previously there had been a bug that made it very, very slow, but we resolved it over the summer. It might make some basic uses like this one faster!

Spongman:
[...]

PR #2473, opened 2017-12-29
webgl:  cache un-scaled primitive geometries

Spongman:
this PR attempts to reduce geometry creation when drawing many 3d primitives of different sizes. spheres/ellipsoids, cylinders & boxes are all scale-invariant, toruses only differ by the ratio between their two radii. cones are troublesome (as usual).


mlarghydracept:
Cool idea! A couple of quick things:

My instinct is to say that you probably don't need to make a `drawBuffersScaled` if this works solidly across the board it should just replace `drawBuffers` with scaling defaulting to 1.

This appears to break the normal material shader, or at least force it to return 0 (pure black) for all of the fragment colors.

Spongman:
unfortunately it doesn't work for everything that uses `drawBuffers`. cones, and the 2d primitives are still drawn un-scaled. there may be ways to optimize those in the future, but i didn't want to go too far with this one.

i'll check the normal shader. it may just need a simliar fix to #2427...

Spongman:
fixes #2252

Spongman:
Was there anything else you wanted me
to change with this one?

src/webgl/p5.Matrix.js
```diff
@@ -458,42 +477,32 @@ p5.Matrix.prototype.mult = function(multMatrix) {
...
  * @chainable
  */
-p5.Matrix.prototype.scale = function() {
-  var x, y, z;
-  var args = new Array(arguments.length);
-  for (var i = 0; i < args.length; ++i) {
-    args[i] = arguments[i];
-  }
+p5.Matrix.prototype.scale = function(x, y, z) {
   //if our 1st arg is a type p5.Vector
```
kjhollen: this part reads a little funny (x = x[0] or x = x.x), but I get what you're doing here. I think it's worth a quick note inline that this is done for optimization purposes, so someone doesn't try to clean this up later.

src/webgl/p5.Matrix.js
```diff
@@ -505,22 +514,22 @@ p5.Matrix.prototype.scale = function() {
...
   }
-  if (axis instanceof p5.Vector) {
-    x = axis.x;
-    y = axis.y;
-    z = axis.z;
-  } else if (axis instanceof Array) {
-    x = axis[0];
-    y = axis[1];
-    z = axis[2];
+  if (x instanceof p5.Vector) {
```
kjhollen: same suggestion as above; add a note here that arguments array is not used for optimization purposes

src/webgl/p5.RendererGL.Retained.js
```diff
@@ -288,6 +288,21 @@ p5.RendererGL.prototype.drawBuffers = function(gId) {
   return this;
 };
 
+p5.RendererGL.prototype.drawBuffersScaled = function(
```
kjhollen: I suggest a detailed private comment for library developers here, explaining how the geometries for built-in primitives are scaled in this function (and citing a few examples that operate differently... like box + torus, what assumptions are made about their dimensions before scaling) and how it might be expanded for future functionality and the missing primitive shapes we have.

src/webgl/primitives.js
```diff
@@ -757,7 +694,22 @@ p5.RendererGL.prototype.triangle = function(args) {
     this.createBuffers(gId, triGeom);
   }
 
-  this.drawBuffers(gId);
+  var uMVMatrix = this.uMVMatrix;
```
kjhollen: add a comment explaining why drawBuffersScaled is not used here.

src/webgl/primitives.js
```diff
@@ -769,66 +721,65 @@ p5.RendererGL.prototype.ellipse = function(args) {
...
-        'Cannot stroke ellipse with more' + ' than 24 detailX or 16 detailY'
-      );
+      console.log('Cannot stroke ellipse with more than 50 detailX');
     }
 
     this.createBuffers(gId, ellipseGeom);
   }
-  this.drawBuffers(gId);
+
+  var uMVMatrix = this.uMVMatrix;
```
kjhollen: as above, a comment explaining the need to manipulate the matrix here & why drawBuffersScaled isn't used would be helpful to future devs

src/webgl/primitives.js
```diff
@@ -842,7 +793,17 @@ p5.RendererGL.prototype.rect = function(args) {
     this._edgesToVertices(rectGeom);
     this.createBuffers(gId, rectGeom);
   }
-  this.drawBuffers(gId);
+
```
kjhollen: add a comment here, too

Spongman:
is that close to what you had in mind?

kjhollen:
yes this is great, thank you! I'll test & merge this shortly...

Spongman:
fantastic, thanks!

Issue #2469, opened 2017-12-28
NaN error in vector mult() during teapot load


Spongman:
run one of the doc samples that loads the teapot.

notice that there's a `NaN` error in `p5.Vector.mult()` - the bug happens in released 0.5.16, but the warning is only visible in the latest github code.

the bug is caused by trying to calculate the normal of a face with two duplicate vertices. it looks like something is wrong with either `parseObj` or the model itself. needs more investigation.

mlarghydracept:
Are you referring to the `p5.Vector.prototype.mult: n is undefined or not a finite number` console warning that has been popping up in a few of the manual tests or something different?

Spongman:
that's the one!

mlarghydracept:
Cool yeah I thought maybe. It doesn't seem to be exclusive to the teapot model. If you load the `webgl/camera/box` test it has the same warning in the console. That is a pretty new warning and appears to come from [this commit](https://github.com/processing/p5.js/commit/3683a72ae0443683d73bdaaf8912917c518b07bb#diff-579922bca7cfcf5ecbc32456f0310b2f). I imagine that `isFinite()` is returning false on some of the numbers. My instinct is to say that that check should use `isNaN()` instead.

Spongman:
i believe the warning is valid. if you set a breakpoint on it and look at the `face` variable in the `computeNormals()` method you'll see that two of the vertices of the erroneous are the same - a singularity.

mlarghydracept:
Hm yeah I see now. It appears a seemingly disparate selection of sketches. Curious.

Spongman:
it seems like the `docs/reference/assets/teapot.obj` file is broken. there are many faces in that file that have duplicate vertices, (eg line 1458: `f  211  212  212`). i think just ignoring such faces should be sufficient.

mlarghydracept:
I see what you mean. That makes sense. I was looking at the blender teapot this entire time.

Spongman:
LOL! not only is the `teapot.obj` file in the docs broken wrt. duplicated vertices. it turns out _all_ the normals are inverted! I suggest we replace it with the good blender one.

mlarghydracept:
fixed by #2474

PR #2474, opened 2017-12-29
Fix broken teapot

Spongman:
- when loading models, ignore faces that have duplicate faces (prevents singularities in the normal calculations)
- use the Blender-generated teapot model in the samples. the old one has incorrect winding order resulting in inverted normals.


Spongman:
fixes #2469

mlarghydracept:
Great thanks!

PR #2483, opened 2018-01-01
Typo in create canvas

### LINKED ISSUE #2482: Typo in CreateCanvas() Documentation
In the P5-js documentation for the createCanvas() function there is a typo:

"Creates a canvas element in the document, and sets the dimensions of it in pixels. This method should be called only once at the start of setup. Calling createCanvas more than once in a sketch will result in very **unpredicable** behavior. If you want more than one drawing canvas you could use createGraphics (hidden by default but it can be shown). "

The word you are looking for is "Unpredictable".

It's a small thing, but love making p5.js a little better :)

bansalnitish:
Fixes #2482


lmccart:
thanks!

PR #2500, opened 2018-01-06
fix createShader example

Spongman:
sorry, i thought this was working, i must have broken this along the way...

this fixes the `createShader()` example, which was missing a call to `createCanvas()`.


mlarghydracept:
Good catch!

PR #2501, opened 2018-01-06
ensure correct use of `preload()` in examples

Spongman:
moved some `load*()` calls to `preload()` in some examples & tests.

### DIFF
--- a/src/image/p5.Image.js
+++ b/src/image/p5.Image.js
@@ -409,7 +409,7 @@ p5.Image.prototype.set = function(x, y, imgOrCol) {
  * <div><code>
  * var img;
  *
- * function setup() {
+ * function preload() {
  *   img = loadImage('assets/rockies.jpg');
  * }
 
--- a/src/webgl/loading.js
+++ b/src/webgl/loading.js
@@ -38,10 +38,12 @@ require('./p5.Geometry');
  * //draw a spinning teapot
  * var teapot;
  *
+ * function preload() {
+ *   teapot = loadModel('assets/teapot.obj');
+ * }
+ *
  * function setup() {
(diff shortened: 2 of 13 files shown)


mlarghydracept:
I can say for sure that switching that in the manual tests is welcome. Not sure about the website examples though. It would make sense to me but not sure if there is a preferred approach with mixing different ideas in single examples. @lmccart thoughts?

Spongman:
yeah, i cansee that. it does make those examples a little more complex. i guess its a payoff between complexity and displaying recommended behavior.

i wonder if it might be worth considering extending the `preload()` delaying behavior to work with calls made in `setup()`. that would actually make using `preload()` unnecessary in most cases. it wouldn't cover cases where you need to consume loaded data in the `setup()` call, but those cases could be handled with `preload()`, or by using the callback feature of most of those loading methods.

can't wait for ES2017 ;-)

limzykenneth:
I actually think it is a good idea to include `preload` in examples just so that it is clear the loading functions should be in the `preload` function to prevent problems with async loading, could prevent potential problems and confusions down the line.

I've been thinking about implementing the loading functions so that they return a promise, at least when outside the preload block or some other ways. If they are promised based, we can even go as far as recommending async/await as a way of using these functions in `setup`.

lmccart:
I think this makes sense to demonstrate the use of preload. thanks! the reason we didn't want to make setup blocking was to preserve the asynchronous behavior should users want that. some sketches may make use of assets later and then it wouldn't make sense to hang the whole page load waiting for them.

Spongman:
Oh, that makes sense, thanks for explaining hat!

PR #2502, opened 2018-01-06
fix createGraphics(WEBGL) crash

### LINKED ISSUE #2478: createGraphics throws error with WebGL
- [x] Bug
- [x] WebGL
- [x] Desktop/Laptop
- [x] 0.5.16

The Following codesnippet contains only the really necessary parts to demonstrate a couple of issues/bugs I come across during experimenting with multiple P2D and WEBGL combinations.
Bug description is in the comment.

```javascript
//
// p5.bug, WEBGL
//
// config:
//   primary P2D
//   offscreen WEBGL
//
// throws:
//   Uncaught TypeError: Cannot read property 'apply' of undefined
//   at p5.RendererGL._applyColorBlend (p5.js:67579)
//
//
// related: 
//   https://github.com/processing/p5.js/issues/2342
//


var renderer;
var graphics;


function setup() { 

  createCanvas(800, 600, P2D);
  // createCanvas(800, 600, WEBGL);

  renderer = this._renderer;
  
  if(renderer instanceof p5.RendererGL){
  }

  graphics = createGraphics(800, 600, WEBGL)
} 


function draw() {
  
  // offscreen
  {
    graphics.push();
    graphics.resetMatrix();
    graphics.ortho();
    graphics.background(64);
    graphics.noStroke(); // buggy stroke 
    graphics.fill(255,0,0);
    graphics.rectMode(CENTER);
    graphics.rect(0, 0, 300, 300);
    graphics.pop();
  }
  
  
  // onscreen
  {
    push();
    background(32);
    
    if(renderer instanceof p5.Renderer2D){
      image(graphics, 0, 0);
    }
    
    if(renderer instanceof p5.RendererGL){
      resetMatrix();
      ortho();
      texture(graphics);
      plane(width, height);
      // 0/0 ... top left corner
      translate(-width/2,-height/2);
    }
    

    // rect attached to mouse
    translate(mouseX, mouseY);
    
    // tilt to see Aliasing
    if(renderer instanceof p5.Renderer2D) rotate (3 * PI/180);
    if(renderer instanceof p5.RendererGL) rotateZ(3 * PI/180);
    
    rectMode(CENTER);
    noStroke();
    fill(0,255,0);
    rect(0, 0, 50, 300);
    pop();
  }
  
}




p5.RendererGL.prototype._applyColorBlend = function(v1,v2,v3,a){
  var gl = this.GL;
  
  // < fix, ... not really, but at least i got the demo going with this
  var inst = this._pInst;
  if(inst instanceof p5.Graphics){
    var inst = inst._pInst;
  }
  // fix >
  

  var color = inst.color.apply(inst, arguments);
  var colors = color._array;
  if(colors[colors.length-1] < 1.0){
    gl.depthMask(false);
    gl.enable(gl.BLEND);
    gl.blendEquation( gl.FUNC_ADD );
    gl.blendFunc( gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA );
  } else {
    gl.depthMask(true);
    gl.disable(gl.BLEND);
  }
  return colors;
};


```

Spongman:
- move _colorMode/_colorMaxes to instance
- fixup push/pop to handle separate property locations
- initialize instance variables in function `p5.prototype._initializeInstanceVariables`, shared with p5.Graphics
- rearrange graphics constructor

closes #2478

### DIFF
--- a/src/color/creating_reading.js
+++ b/src/color/creating_reading.js
@@ -311,19 +311,10 @@ p5.prototype.color = function() {
   p5._validateParameters('color', arguments);
   if (arguments[0] instanceof p5.Color) {
     return arguments[0]; // Do nothing if argument is already a color object.
-  } else if (arguments[0] instanceof Array) {
-    if (this instanceof p5.Renderer) {
-      return new p5.Color(this, arguments[0]);
-    } else {
-      return new p5.Color(this._renderer, arguments[0]);
-    }
-  } else {
-    if (this instanceof p5.Renderer) {
-      return new p5.Color(this, arguments);
-    } else {
-      return new p5.Color(this._renderer, arguments);
-    }
   }
+
+  var args = arguments[0] instanceof Array ? arguments[0] : arguments;
+  return new p5.Color(this, args);
 };
 
 /**
@@ -439,8 +430,8 @@ p5.prototype.hue = function(c) {
 
 p5.prototype.lerpColor = function(c1, c2, amt) {
   p5._validateParameters('lerpColor', arguments);
-  var mode = this._renderer._colorMode;
-  var maxes = this._renderer._colorMaxes;
+  var mode = this._colorMode;
+  var maxes = this._colorMaxes;
   var l0, l1, l2, l3;
   var fromArray, toArray;
 
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -30,9 +30,9 @@ var color_conversion = require('./color_conversion');
  * @class p5.Color
  * @constructor
  */
-p5.Color = function(renderer, vals) {
+p5.Color = function(pInst, vals) {
   // Record color mode and maxes at time of construction.
-  this._storeModeAndMaxes(renderer._colorMode, renderer._colorMaxes);
+  this._storeModeAndMaxes(pInst._colorMode, pInst._colorMaxes);
 
   // Calculate normalized RGBA values.
   if (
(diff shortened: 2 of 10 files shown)


mlarghydracept:
The idea seems sound enough but this touches a lot of different parts of the library so I will have to defer to @lmccart on this one.

kjhollen:
Hi @Spongman can you elaborate on how the architecture changes here fix the related bug? that would help us review a little faster. thanks! :)

Spongman:
oh, sure. the problem is that the `Graphics` constructor calls the `Renderer` constructor which needs to construct colors according to the current colorMode. the colorMode is fetched by looking at _pInst's renderer. this "works" currently when creating a new canvas because there already is a 2d renderer attached to the p5 instance. the reason this bug occurs is that the GL renderer is created with the graphics as its _pInst and that _pInst doesn't yet have a renderer - so it crashes.

this PR fixes this by moving the colorMode & colorMaxes style properties out of the renderer and into the instance/graphics.

the `push`/`pull` methods needed to be changed because now not all style properties are stored in the renderer.

there's a new `p5.prototype._initializeInstanceVariables` function that provides a common place for `p5` and `p5.Graphics` instances to initialize their common members. `p5.Graphics` extends `p5.Element` but also receives `p5`s methods and these common state variables. this monkey-patching is moved earlier in `p5.Graphics`'s constructor because it needs to occur before the renderer is created.

kjhollen:
Awesome, thanks for the details! I've read this over a couple times and think it is good to go as-is. I rebased this on the latest in master and the tests are passing, a random sampling of color-related reference examples look ok, manual tests seem good, too. because it touches color code for both 2D and 3D mode, do you want to have a look before merging this @lmccart?

lmccart:
thanks for the review @kjhollen, looks good to me too.

PR #2505, opened 2018-01-06
Normalize line endings in sample-linter

Zalastax:
Prevents accidental mixing of line endings if the repo is checked out with autocrlf=false on Windows.
Replaces #2431 and solves the same problem as #2484. This can optionally be combined with #2484.

I've tested this with both `autocrlf=false` and `autocrlf=true`.
@Spongman what do you think?

### DIFF
--- a/utils/sample-linter.js
+++ b/utils/sample-linter.js
@@ -57,7 +57,7 @@ function splitLines(text) {
   return lines;
 }
 
-var EOL = require('os').EOL;
+var EOL = '\n';
 
 var userFunctions = [
   'draw',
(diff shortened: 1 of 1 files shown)


Zalastax:
I think that  #2484 must be included even if this is merged. The reason being that  .editorconfig has to be changed to not enforce '\n'. Without that change, any editor that supports editorconfig will keep writing '\n' even though the repo uses autocrlf.

lmccart:
@Spongman I'm not very familiar with what this is doing but I see that you and @Zalastax were discussing previously on [another thread](https://github.com/processing/p5.js/pull/2431). can you offer a recommendation on how best to proceed with this PR and also #2484?

lmccart:
I guess my general approach would be to do the minimal structuring needed to make things work ok for people, rather than going further toward enforcing particular ways of working. but I don't know how this plays out in this case exactly.

lmccart:
i don't have enough information about this one to see why this is necessary or how it fits with #2484. i'd prefer to leave as is now and rely on developers to set their settings manually.

PR #2511, opened 2018-01-07
fix fill/stroke order bug

### LINKED ISSUE #2509: webgl: results dependend on order of stroke/fill calls
you get different results if you call `fill()` before `stroke()`, or visa-versa.

image

````
function setup() {
  createCanvas(windowWidth, windowHeight-4, WEBGL);
  strokeWeight(2);
}
function draw() {
  var t = millis();
  background(0);
  
  push();
  fill(255, 0, 0, 50);
  stroke(0, 255, 0, 50);
  
  translate(-100, 0, 0);
  rotateY(t/1000);
  rotateZ(t/1234);
  box();
  pop();
  
  push();
  stroke(0, 255, 0, 50);
  fill(255, 0, 0, 50);
  
  translate(100, 0, 0);
  rotateY(t/1000);
  rotateZ(t/1234);
  box();
  pop();
}
````

https://codepen.io/Spongman/pen/WdXabo?editors=0010

Spongman:
this PR:
- removed the `shader.active` flag which was getting out of sync, and just uses the existing `_doFill`/`_doStroke` style properties instead.
- removed the redundant GL-specific `noFill`/`noStroke` methods.
- delays setting the blend stuff until just before the draw call (since it's a global setting shared between fill & edge shaders).
- removes some of the unnecessary `useProgram`s
- fixes #2509

### DIFF
--- a/src/color/setting.js
+++ b/src/color/setting.js
@@ -507,7 +507,7 @@ p5.prototype.fill = function() {
  * </code>
  * </div>
  *
- * <div>
+ * <div modernizr='webgl'>
  * <code>
  * function setup() {
  *   createCanvas(100, 100, WEBGL);
@@ -547,7 +547,7 @@ p5.prototype.noFill = function() {
  * </code>
  * </div>
  *
- * <div>
+ * <div modernizr='webgl'>
  * <code>
  * function setup() {
  *   createCanvas(100, 100, WEBGL);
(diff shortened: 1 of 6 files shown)


mlarghydracept:
Great idea! The organization of that section of the code is a lot better now. I don't have an opportunity to investigate deeply. Can you just flag me in some way when you feel that this is ready for a merge? I can run through it then.

Spongman:
oops, i forgot that i had to exclude the webgl examples from the tests, now that they're actually in the docs.  should be good to go now, thx!

mlarghydracept:
Cool so I just ran through a bunch of use scenarios and tests. I really like that this now falls in line with using _doFill/_doStroke along with the 2D renderer. Thank you!

PR #2515, opened 2018-01-07
webgl: use Processing's line vertex shader

### LINKED ISSUE #2488: webgl: incorrect billboarding of thick lines
When the view is rotated, the segments of thick lines are not being billboarded correctly.

(also, the joins are not pretty, but that's a different bug)

image

````
function setup() {
}

function draw() {
  background(0);
  rotateX(millis()/1000);
  noFill();
  stroke(255);
  strokeWeight(100);
  
  beginShape();
  vertex(-100, -100);
  vertex( 100, -100);
  vertex( 100,  100);
  vertex(-100,  100);
  endShape(CLOSE);
}
````

https://codepen.io/Spongman/pen/ppwQEj?editors=0010

### LINKED ISSUE #2510: webgl: artifact caused by edge z adjustment
here, the red face is behind the green face, but its edge appears in front (neither faces are transparent):

image

````
function setup() {
  createCanvas(windowWidth, windowHeight-4, WEBGL);
  stroke(255, 255, 255);
}

function draw() {
  background(0);
  
  fill(255, 0, 0);
  rect(-100, -100, 200, 200, 1, 1);

  translate(0, 0, 0.5);
  fill(0, 255, 0);
  rect(0, 0, 200, 200, 1, 1);
}
````
https://codepen.io/Spongman/pen/YYEJjy?editors=0010

### PR DESCRIPTION
**Pull request by Spongman**
ok, this one uses (a tweaked version of) Processing's line vertex shader.

(partially) fixes #2488, fixes #2510.

it doesn't display the same wireframe scaling behavior that Processing does, but it's the same as p5.js currently is.

### DIFF
--- a/src/webgl/shaders/line.vert
+++ b/src/webgl/shaders/line.vert
@@ -70,7 +70,7 @@ void main() {
   vec2 normal = vec2(-tangent.y, tangent.x);
 
   float thickness = aDirection.w * uStrokeWeight;
-  vec2 offset = normal * thickness;
+  vec2 offset = normal * thickness / 2.0;
 
   // Perspective ---
   // convert from world to clip by multiplying with projection scaling factor
--- a/src/webgl/shaders/line.vert
+++ b/src/webgl/shaders/line.vert
@@ -1,42 +1,89 @@
+/*
+  Part of the Processing project - http://processing.org
+  Copyright (c) 2012-15 The Processing Foundation
+  Copyright (c) 2004-12 Ben Fry and Casey Reas
+  Copyright (c) 2001-04 Massachusetts Institute of Technology
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation, version 2.1.
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+  You should have received a copy of the GNU Lesser General
+  Public License along with this library; if not, write to the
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mlarghydracept** [comment]
also fixes #2324, which is a major bonus

So I still feel like that snapping that I mentioned in #2488 is too pronounced. You are right that Processing has it, but it is significantly less severe in Processing. Over here it feels like a too big of a problem to ignore. You can see this by rotating any primitive with a large stroke. For example this shows the effect pretty strongly:

```
  background(255);
  rotateX(frameCount * 0.02);
  rotateY(frameCount * 0.02);
  strokeWeight(50);
  box(300);
```
This shader does look better otherwise but this is definitely something we should fix before merging. We could probably move the edge toward the camera but then we will probably hit similar problems to #2510 again. I see that the problem is one half of the edge appearing suddenly but I don't understand why the billboarding is sometimes gradual as expected and sometimes it goes from full invisibility, perpendicular to camera, to being fully parallel to the camera in what seems like a single frame.

mlarghydracept:
Looking at this some more it might be something worth pinging Processing folks about too because it seems shader specific and it definitely is noticeable on their end too now that I am remembering to account for the difference in strokeWeight

Spongman:
yeah, the snapping is definitely bad. but it's there in master, also..

mlarghydracept:
😅 You are very right! I was so caught up in this branch that I forgot to go back to try out the huge strokes on master. I will do a final run through and merge this tomorrow. As for the snapping, it can become a separate issue and we can reach out to the folks that work on P3D and see what they think. The problem seems just as present on Processing even if slightly mitigated by the better joins.

Spongman:
maybe one way to do this is to project the camera-facing edge quads back onto each face?

mlarghydracept:
That's a thought. Though now that I have been playing around with Processing more it feels like less of a priority. It might make sense to figure out our joins first. This PR is super straightforward and everything seems to be working nicely with the swap out. Thanks for this!

PR #2519, opened 2018-01-08
p5.Color.toString() now returns multiple string formats

stellartux:
Edited p5.Color.toString() to be able to format the string as RGBA, HSBA, HSLA, percentage values or hex color code.

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -52,9 +52,18 @@ p5.Color = function(renderer, vals) {
 };
 
 /**
+ * This function returns the color formatted as a string. This can be useful
+ * for debugging, or for using p5.js with other libraries.
  * @method toString
- * @param {String} [format]
- * @return {String}
+ * @param {String} [format] How the color string will be formatted.
+ * Leaving this empty formats the string as rgba(r, g, b, a).
+ * '#rgb' '#rgba' '#rrggbb' and '#rrggbbaa' format as hexadecimal color codes.
+ * 'hex3' 'hex4' 'hex6' and 'hex8' are aliases for the above formats respectively.
+ * 'hex' and 'web-color' are aliases of 'hex6'.
+ * 'rgb' 'hsb' and 'hsl' return the color formatted as each color mode.
+ * 'rgba' 'hsba' and 'hsla' are the same as above but with alpha channels.
+ * 'rgb%' 'hsb%' 'hsl%' 'rgba%' 'hsba%' and 'hsla%' format as percentages.
+ * @return {String} the formatted string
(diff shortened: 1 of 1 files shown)


Spongman:
Maybe the 'hexN' versions should omit the '#' prefix? Then it would be easier to use this with other prefies (like '0x', '&', etc...)

amoetodi:
Good idea, I'll make those changes.

amoetodi:
This should be good to go now, unless you can think of anything else.

src/color/p5.Color.js
```diff
@@ -52,8 +52,18 @@ p5.Color = function(renderer, vals) {
...
 /**
+ * This function returns the color formatted as a string. This can be useful
+ * for debugging, or for using p5.js with other libraries.
  * @method toString
- * @return {String}
+ * @param {Number|String} [format] How the color string will be formatted.
+ * Leaving this empty formats the string as rgba(r, g, b, a).
+ * '#rgb' '#rgba' '#rrggbb' and '#rrggbbaa' format as hexadecimal color codes.
+ * 'hex3' 'hex4' 'hex6' and 'hex8' return the same, without the '#''
+ * 16 'hex' and 'web-color' are aliases of 'hex6'.
```
lmccart: are these ones necessary? I am thinking of a beginner that reads this, it feels a bit overwhelming in the options and text to read through right now.

src/color/p5.Color.js
```diff
@@ -52,8 +52,18 @@ p5.Color = function(renderer, vals) {
...
 
 /**
+ * This function returns the color formatted as a string. This can be useful
+ * for debugging, or for using p5.js with other libraries.
  * @method toString
- * @return {String}
+ * @param {Number|String} [format] How the color string will be formatted.
+ * Leaving this empty formats the string as rgba(r, g, b, a).
+ * '#rgb' '#rgba' '#rrggbb' and '#rrggbbaa' format as hexadecimal color codes.
+ * 'hex3' 'hex4' 'hex6' and 'hex8' return the same, without the '#''
```
lmccart: is this one necessary? it's adding in new language that we haven't really used elsewhere in the library. the other formats mostly align with inputs for the color functions and feel more intuitive. couldn't the user just use substring to remove the '#' if desired?

stellartux: I get where you're coming from, I was torn between trying to be easy to understand and trying to cover all the options someone might guess. Do you think it'd be better to remove the 'hex' options altogether, just have '#rrggbb' etc for the different hex color codes?

lmccart:
thanks!

PR #2520, opened 2018-01-08
more yui docs tweaks

Spongman:
- more `@chainable` fixes
- fix curve/bezier parameters & examples
- fix example whitespace


lmccart:
👍

PR #2524, opened 2018-01-09
p5.Matrix - rework/updated/fixed/...

diwi:
This PR is around quite a bit of re-implemented/new-implemented matrix functions.

It takes care to get same results/behaviour/usage, as with the original Processing PMatrix3D. Also, it doesnt break any existing functionality.

Altough I labeled some of the 3x3-methods deprecated since the new updated methods handle both mat3 and mat4. In general, It's probably worth discussing a clearer separation between mat3 and mat4, e.g. not merging both into just one class.

Here is a quick Overview:

```
identity()
copy()
set()
setColumn(col, vsrc)
setRow(row, vsrc)
getArray()
isMat3()
isMat4 ()

multVec(vsrc, vdst)
mult(mright, mdst)
premult(mleft)

rotateX(angle)
rotateY (angle)
rotateZ(angle)
rotate(angle, axis)
scale(vsrc)
shearX(angle)
shearY(angle)
translate(vsrc)

determinant()
invert(msrc, mdst)
transpose(msrc, mdst)
inverseTranspose (msrc)

invert3x3(msrc) // deprecated
transpose3x3(msrc) // deprecated

frustum(left, right, bottom, top, near, far)
ortho(left, right, bottom, top, near, far)
perspective(fovy, aspect, near, far)

toString()
print()

assertAngleMode(angle) // angle degree or radians
```


src/webgl/p5.Matrix.js
```diff
@@ -1,717 +1,1240 @@
...
  *
- * @method set
- * @param {p5.Matrix|Float32Array|Number[]} [inMatrix] the input p5.Matrix or
- *                                     an Array of length 16
+ * @method copy
+ * @return {p5.Matrix} matrix
  * @chainable
  */
+p5.Matrix.prototype.copy = function() {
+  return new p5.Matrix(this);
```
Spongman: you're losing the p5 instance here.

diwi: The p5.Matrix is copied in the constructor, this includes the p5 reference, if there was one.

However, this is not very well designed anyways. 
A Matrix should do matrix stuff and not care about any deg-to-rad-etc-conversion - which is the only reason for the p5-ref in the first place. 
IMO, in a next step, the p5 member should be completely removed from the matrix or any other specialized kind for that matter.
Apart from that, it is valid to create a matrix without any p5-reference and then apply/set it as one of the core-matrices (modelview, projection, etc...). In such a case the reference is lost by all means. As a result, code and execution and will be just a mess.

The deg/rad-mode option (if useful at all) should be handled by a layer (p5 itself) that is responsible for that option.

Spongman: Oh yes sorry about that.

Another thing: can you specify the parameter arguments fully (i.e. No varargs). You can add an overload if there's an ambiguity.

diwi: Updated. But I'm not sure whats the preferred way of documenting such overloaded functions.

Spongman:
Oh, I see its private. It may not actually need its parameters documenting after all...

mlarghydracept:
This looks really nice! Thanks for the elaborate and thoughtful updates. There does some seem to be one major issue currently. Some calls to `translate()` when drawing with immediateMode seem to break the matrix or at least give it unexpected behavior. You can see this in a number of the immediateMode manual tests.

diwi:
I see, the example call translate with 2 args ... translate(x,y). 

Atm I assumed translate(v3) gets a 3d vector passed. Should there really be done a whole lot of arguments parsing in the matrix-class? because if so, this probably gets inconsistent very soon. Or should the p5.RendererGL takes over that part.
Which actually is the same question for the p5-reference above.

The renderer has some transformation-implementation (rotate,scale and translate) and should probably take care of the deg-to-rad conversion as well as the scale/translate arguments parsing for only 2 arguments.

```
immediate.scale(x,y),  // calls matrix.scale(x,y,1);
immediate.translate(x,y); // calls matrix.translate(x,y,0);
immediate.rotate(a) // which calls matrix.rotateZ(a), which btw. is missing and would be useful to have, to be compatible to the non GL renderer
```

It seems a bit of an overkill to do all this in the matrix class, because the more work is spent on parsing and making assumptions, the less work is spent on the actual matrix-related stuff.

However, I can quickly fix the translate(x,y) in the matrix too, to not break existing sketches. However, i think this should really be done in the renderer.

diwi:
Updated.
The Matrix class now (carefully) parses the arguments for translate and scale. The renderer's just passing things through for scale/translate/rotate.


```
the following is valid. 
scale(number)
scale(x,y)
scale([x,y])
scale(x,y,z)
scale([x,y,z])
scale(p5.Vector)
```

Spongman:
please see `color.prototype.background` for an example of how to document the parameters of overloaded methods. it's not hugely important as this is a private class and so doesn't appear in the docs, generate typescript definitions or have its parameters validated, but it's probably worth getting it right just in case that changes later on...

diwi:
Apparently there is already a bunch of inconsistency going on in the core-files for handling the angle-mode-responsibility.

e.g. in transform.js:
p5.prototype.rotate - handles the anglemode, while rotateX,Y and Z dont, same for shearX and shearY.
Same issue applies probably for all the current renderer implementations.

This should probably be taken care of in a separate PR as atm this is quite a mess across a couple of source files.
(Probably get rid of it at all).


```javascript
function setup () {
}

function draw () {

  var angle = map(mouseX, 0, width, 0, 180);

  background(0);
  
  push();
  translate(-200,0,0);
  angleMode(RADIANS);
  rotate(angle * PI/180);
  displayScene();
  pop();
  
  push();
  translate(+200,0,0);
  angleMode(DEGREES);
  rotate(angle);
  displayScene();
  pop();
}


function displayGizmo(size){
  strokeWeight(1);
  stroke(255,0,0); line(0,0,0,size,0,0);
  stroke(0,255,0); line(0,0,0,0,size,0);
  stroke(0,0,255); line(0,0,0,0,0,size);
}

function displayScene(){
  displayGizmo(150);
  
  noStroke();
  pointLight(255,255,255, 100, 100, 100);
  
  ambientMaterial(255,255,255);
  sphere(20);
  plane(300);
  
  push();
  ambientMaterial(255,0,0);
  translate(100,0,0);
  box(20);
  pop();
  
  push();
  ambientMaterial(0,255,0);
  translate(0,100,0);
  box(20);
  pop();
}
```

Also notice, that the pointlight is not transformed as expected.




anglemode_bug

kjhollen:
Agreed that there is more work to do with angleMode... and ok to leave that for a future PR!

At a glance, it seems like some unit tests should be added for new matrix functionality (shear, frustum, perspective, etc.). It would also be helpful if you could break this PR into two: one for refactor and another for new features & functionality—it looks like the first commit could be its own PR, because it's relatively contained.

Just to make sure I'm understanding, the motivation for these changes is to optimize performance of matrix operations by reducing the number of array creations that happen in each function—is that right?

src/webgl/p5.Matrix.js
```diff
@@ -1,717 +1,1324 @@
...
 /**
- * Inverts a 3x3 matrix
- * @method invert3x3
+ * Multiplies a vector with this matrix:
+ *
+ * operation: vdst = mat * vsrc
+ *
+ * note: src can be equal to vdst.
+ *
+ * @method multVec
```
kjhollen: can we add a unit test for this or remove it?

src/webgl/camera.js
```diff
@@ -287,28 +278,72 @@ p5.RendererGL.prototype.ortho = function(left, right, bottom, top, near, far) {
...
+ * </div>
+ *
+ * @alt
+ * 3 3d boxes, reveal several more boxes on 3d plane when mouse used to toggle
+ *
+ */
+p5.prototype.frustum = function() {
+  this._renderer.frustum.apply(this._renderer, arguments);
+  return this;
+};
```
kjhollen: Cool! I'll merge this, but in the future please separate PRs for a new feature like this one. It doesn't seem to be an optimization, from what I can tell.

src/webgl/p5.Matrix.js
```diff
@@ -1,717 +1,1324 @@
...
+ * If this matrix owns a p5-reference it will respect the p5._angleMode flag.
+ *
+ * TODO: find a better way for doing this
+ *       or, get rid of anglemode and the p5 instance
+ *
+ * @method assertAngleMode
+ * @param {Number} angle
+ * @returns {Number} the angle (deg or rad)
+ */
+p5.Matrix.prototype.assertAngleMode = function(angle) {
```
kjhollen: I think this has been superseded by new changes. I can address this when I handle merge conflicts (I spent some time on it while testing this afternoon).

src/webgl/p5.Matrix.js
```diff
@@ -1,717 +1,1324 @@
...
-    d08 = this.mat4[8] * this.mat4[15] - this.mat4[11] * this.mat4[12],
-    d09 = this.mat4[9] * this.mat4[14] - this.mat4[10] * this.mat4[13],
-    d10 = this.mat4[9] * this.mat4[15] - this.mat4[11] * this.mat4[13],
-    d11 = this.mat4[10] * this.mat4[15] - this.mat4[11] * this.mat4[14];
+p5.Matrix.prototype.rotateX = function(angle) {
+  angle = this.assertAngleMode(angle);
+  var c = Math.cos(angle);
+  var s = Math.sin(angle);
+  // prettier-ignore
+  this.mult([1, 0, 0, 0,
```
kjhollen: I'm not sure I understand why these calculations are done here instead of deferring to `rotate(angle, [0, 0, 1])`? and likewise for rotateY, rotateZ.

src/webgl/p5.RendererGL.js
```diff
@@ -679,58 +679,43 @@ p5.RendererGL.prototype.clear = function() {
...
-  }
-  this.uMVMatrix.rotate(rad, axis);
+p5.RendererGL.prototype.rotate = function(angle, axis) {
+  this.uMVMatrix.rotate(angle, axis || [0, 0, 1]);
+  return this;
+};
+
+p5.RendererGL.prototype.rotateX = function(angle) {
+  this.uMVMatrix.rotateX(angle);
+  return this;
```
kjhollen: what is the advantage to setting the Matrix directly in rotateX/Y/Z instead of deferring to rotate here? is there a performance improvement?

lmccart:
hi @diwi I'm going to close this for organizational purposes. but please feel free to reopen when you make the requested changes and we can review. thanks!

PR #2531, opened 2018-01-11
Update Gruntfile.js

Spongman:
`yui:dev` needs to minify p5.js in order for the local docs to run correctly.


outofambit:
thank you!

PR #2534, opened 2018-01-11
resize 3d primitive examples

Spongman:
this just resizes some of the 3d primitives in the docs examples so they fit better in the 100x100 square.


mlarghydracept:
Great! Thanks!

PR #2538, opened 2018-01-13
add example for p5.Element.class

sajoy:
👋 Hi! First PR and try at writing an example. I based it off the id method documentation. Is it preferred to write similar examples like such, or have a variety?


lmccart:
this is great! thank you!

PR #2543, opened 2018-01-15
loadPixels cleanup

### LINKED ISSUE #2539: loadPixels(x,y,w,h) ignores arguments
- [x] Found a bug
- p5.js version: 0.5.16

arguments are not passed to the subsequent renderer.loadPixels().

https://github.com/processing/p5.js/blob/master/src/image/pixels.js#L521

I feel like I have seen similar things a before but didnt keep track of it, so this is just a hint for a general check.
Here is the updated version.

```javascript
p5.prototype.loadPixels = function() {
  this._renderer.loadPixels.apply(this._renderer, arguments);
};
```

mlarghydracept:
Okay actually just had a second to glance over this and I remembered that the reason that `gl.readPixels` doesn't deposit data directly into the p5 instance's `pixels` is because it needs to write into a Uint8Array. So instead of making a renderer-level `pixels` that would need to be resized with calls to `loadPixels` I just opted for temporary array creation but now I realize that GC would be slow for an array of that size if it is being created and discarded rapidly.

So this PR adds a renderer-level array buffer for grabbing pixel data from the GL drawing context before pushing it up to `_pInst`'s pixels array. This also gets rid of the x, y, width, height parameters for `loadPixels`. It would be possible to check if the array needs to be resized based on the arguments but just grabbing the whole canvas at once in the same way as the 2D renderer seems better for simplicity.

closes #2539

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -624,7 +624,7 @@ p5.RendererGL.prototype.loadPixels = function() {
   h *= pd;
   //if there isn't a renderer-level temporary pixels buffer
   //make a new one
-  if (this.pixels === undefined) {
+  if (typeof this.pixels === 'undefined') {
     this.pixels = new Uint8Array(
       this.GL.drawingBufferWidth * this.GL.drawingBufferHeight * 4
     );
@@ -672,7 +672,7 @@ p5.RendererGL.prototype.resize = function(w, h) {
     this._setDefaultCamera();
   }
   //resize pixels buffer
-  if (this.pixels !== undefined) {
+  if (typeof this.pixels !== 'undefined') {
     this.pixels = new Uint8Array(
       this.GL.drawingBufferWidth * this.GL.drawingBufferHeight * 4
     );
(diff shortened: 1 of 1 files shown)


src/webgl/p5.RendererGL.js
```diff
@@ -620,17 +616,29 @@ p5.RendererGL.prototype.loadPixels = function() {
...
+  var y = 0;
+  var w = this.width;
+  var h = this.height;
   w *= pd;
   h *= pd;
-  var pixels = new Uint8Array(
-    this.GL.drawingBufferWidth * this.GL.drawingBufferHeight * 4
+  //if there isn't a renderer-level temporary pixels buffer
+  //make a new one
+  if (this.pixels === undefined) {
```
lmccart: can you make this
`if (typeof this.pixels === 'undefined') {`
this is nit-picking, but we've generally stuck to checking typeof across the library. this post (second answer) is old and isn't particularly convincing, but since we've been going with this anyway, may as well stick with it?
https://stackoverflow.com/questions/4725603/variable-undefined-vs-typeof-variable-undefined

mlarghydracept: Oh! I knew this at one point but probably forgot along the way. Thank you for catching this and reminding me of the convention!

PR #2545, opened 2018-01-15
#1954: added an example for createVideo(), with video assets

eJohnstonePhd1991:
issue #1954. I noticed this function lacked an inline example so I added a simple one. The video assets are not mine but are from here (http://techslides.com/sample-webm-ogg-and-mp4-video-files-for-html5), where the author has said they are open to use. Please let me know if there are any problems or if you think the example can be improved.


lmccart:
super, thanks!

mavr92:
Meaby

Issue #2528, opened 2018-01-10
smooth/noSmooth not defined in WebGL


micuat:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [x] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 0.5.16
- Web browser and version: Chrome 63.0.3239.132 (Official Build) (64-bit)
- Operating System: Windows 10
- Steps to reproduce this:

```
function setup() { 
  createCanvas(400, 400, WEBGL);
} 

function draw() { 
  background(220)
  smooth();
  rectMode(CENTER);
  rect(100, 100, 100, 100);
}
```

https://alpha.editor.p5js.org/micuat/sketches/rk_KIxV4f

WebGL Primitives are discussed [here](https://github.com/processing/p5.js/issues/2226#issuecomment-356707969). As described `smooth` is not defined and terminates the script - but it should instead give a friendly error. There's a workaround for smoothing in Webgl but never implemented as `smooth` function https://github.com/processing/p5.js/issues/2250

kjhollen:
I want to propose that smooth/noSmooth in the WebGL renderer can be functions that just defer to `this.setAttributes('antialias', true);` or `this.setAttributes('antialias', true);`. a call to push/pop before & after may also be required. this can probably go right into src/webgl/p5.RendererGL.js

Issue #2370, opened 2017-11-18
style push/pop for webgl mode


kjhollen:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Details about the bug: 

- p5.js version: 0.5.16
- Web browser and version: any :)
- Operating System: any
- Steps to reproduce this:
call push(), change the stroke and/or fill, call pop()

expected results: stroke and fill revert to color last used before push()
actual results: color does not revert

in 2D mode, the styles are cached wholesale as part of the canvas API—but we need to do a little bit more work to reset the shaders in 3D mode.

for reasons outlined in pull request #2211, it would be good to also cache which texture is in use. so that we can use push/pop internally to change textures in gl mode for image() and future calls to text() when implemented—and maybe other uses we haven't anticipated yet.

I have a first pass at this semi-working that is not quite ready for review. needs unit tests!

kjhollen:
work in progress here: https://github.com/kjhollen/p5.js/commit/bd40111984c78bdae0349075ac6a6c845dc67e61

@mlarghydracept this needs a lot more testing, though it seems to be working in simple cases. I'm worried that the code is a little messy or too complicated—if you've got time, can you have a look and let me know if you have any ideas for resolving that? I used the main push/pop as a model for saving styles.

Spongman:
i think it might simplify things a _little_ if, instead of handling multiple stacks in the renderer, you just allow the renderer to add its state into the objects pushed onto the `_styles` array. well, in the future if there are more renderers it would reduce the amount of duplicated stack-handling code.

for example, if the top-level `push` function looked something like:

    p5.prototype.push = function () {
      var rendererState = this._renderer.push();
      this._styles.push({
        rendererState: rendererState,
        properties: {
          _doStroke: this._renderer._doStroke,
          _strokeSet: this._renderer._strokeSet,
          ...
        }

and `pop` looked like:

    p5.prototype.pop = function () {
      var lastS = this._styles.pop();
      for(var prop in lastS.properties) {
        this._renderer[prop] = lastS.properties[prop];
      }
      this._renderer.pop(lastS.rendererState);
    };


[...]

Issue #2369, opened 2017-11-17
Manipulating p5.Geometry Vertices (WebGL)


BillDNA:
#### Nature of issue?

- [ ] Found a bug
- [x] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

- p5.js version:  v0.5.16
- Web browser and version: Chrome
- Operating System: MacOSX

#### New feature details:

When using a p5 Geometry object to draw 3D models, it would be nice since you have access to the vertices if you change the value of one.  If you can recompute the shape and it will deform the model appropriately.  Looking at the object it seems to have functions like computeNormals and computeFaces but these don't seem to be what I'm looking for.   Maybe the feature exists but there doesn't seem to be and documentation on it.

Siphonophora:
Hi, could you give an example of the type of geometry you are trying to update? 

For most geometry, the compute faces and compute normal functions are only called once at the time the geometry is first drawn. That computed geometry is then put into the webgl buffer. We then use a reference to that geometry to redraw the geometry in subsequent frames. 

I believe once the geometry is defined we don't ever modify it. Possibly we can't, but I am not sure.

BillDNA:
So the idea is by moving the specific vertices of a geometry the shape will
deform in ways that scale, translate or rotate can't.

After digging into it I found the buffer based on the Geometries based on
the 'gid' property.  So in a supper messy way I'm just changing the gid so
the 'morphed' shape get added into buffer.  but it would be nice if the
geometry object had a method that would re add it to the buffer.

Siphonophora:
What type of geometry are you trying to deform (2d, 3d) and how often? 

My understanding of what happens once we hand geometry off to the buffer is pretty limited, but I think if you generate new geometry you would want a new gid and not to overwrite the buffer. I am not sure what will happen if you try to make changes.

BillDNA:
3d and the only way i've found is to change the gid after adjusting the vertices so it is forced to recalculate and insert into buffer.

Spongman:
try this for a hack:

````
var renderer;
[...]

PR #2547, opened 2018-01-17
Render pipeline changes

Spongman:
(Apologies for the size of this PR. It kind of came out like to is and I didn't have much luck splitting it apart)

This PR:
- Delays setting shader uniforms until just before the drawX call. This facilitates both switching shaders without losing state, and allows including this state in push/pop.
- Expands push/pop to include render state.
- Delays creating render buffers until just before the drawX call. This facilitates reusing geometries with different shaders (previously only buffers used by the current shader was created during geometry creation), optimizes away creation of unnecessary buffers and allows buffers to be refreshed as needed.
- Changes the meaning of 'curXXXshader’ to only hold the custom shader, if set. The shader used to render is determined by the render state prior to rendering.
- For rendering geometries, the association between source array, render buffer and vertex attribute is contained in a BufferDef object, arrays of which describe inputs for stroke and fill shaders. Previously repetitious attribute & buffer handling code now just involves looping through these arrays (see prepare buffers).
- Unifies the immediate & retained rendering code. Now the immediateMode object is just a p5.Geometry and is rendered like all the other geometries. Flags are added to the the geometry to track whether or not a source array is dirty and requires its render buffer to be refreshed. The immediate-mode renderers handle per-vertex attributes (color, ambient, shininess, etc…) instead of per-geometry uniforms.
- Expands the set of material properties to include those from processing: emissiveMaterial, specularMaterial, shininess. Plus additional lights, noLights, lightSpecular, lightFalloff methods for setting light properties. In addition, the rendered output more closely matches that of processing, for example materials/vertices can now have distinct ambient, diffuse & specular colors, each lit by their corresponding light components, and directional lights have specular reflections.
- Unifies the lighting code between all lighting shaders (immediate, per-vertex & per-pixel). Uses a simple include hack to 'include’ the lighting code in order to avoid duplication in the final .js file. This significantly reduces the visual difference between the per-pixel and lit per-vertex renderers.
- Fixes most of the immediate-mode draw modes (though still no quads or points)
- ~~Thunks render methods. This one is pretty simple but causes the most noise. Previously a lot of 3d-specific methods were declared directly on p5, or stub methods were used to thunk down to p5’s _renderer. Now there's a method 'thunkRendererMethods’ that runs during startup and adds these thunks automatically and additionally adds warning messages for unimplemented methods. This removes a bunch of boiler-plate code, and ensures that, for example, calling 'sphere()’ on the canvas renderer just outputs an error instead of crashing.~~

I think that about covers it. Sorry again for the mess!

### DIFF
--- a/src/core/2d_primitives.js
+++ b/src/core/2d_primitives.js
@@ -231,15 +231,7 @@ p5.prototype.ellipse = function(x, y, w, h, detailX) {
  * @param  {Number} z2 the z-coordinate of the second point
  * @chainable
  */
-p5.prototype.line = function() {
-  p5._validateParameters('line', arguments);
-
-  if (this._renderer._doStroke) {
-    this._renderer.line.apply(this._renderer, arguments);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * Draws a point, a coordinate in space at the dimension of one pixel.
@@ -266,15 +258,7 @@ p5.prototype.line = function() {
  *4 points centered in the middle-right of the canvas.
  *
  */
-p5.prototype.point = function() {
-  p5._validateParameters('point', arguments);
-
-  if (this._renderer._doStroke) {
-    this._renderer.point.apply(this._renderer, arguments);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * Draw a quad. A quad is a quadrilateral, a four sided polygon. It is
@@ -320,15 +304,7 @@ p5.prototype.point = function() {
  * @param {Number} z4
  * @chainable
  */
-p5.prototype.quad = function() {
-  p5._validateParameters('quad', arguments);
-
-  if (this._renderer._doStroke || this._renderer._doFill) {
-    this._renderer.quad.apply(this._renderer, arguments);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * Draws a rectangle to the screen. A rectangle is a four-sided shape with
@@ -390,16 +366,7 @@ p5.prototype.quad = function() {
  * @param  {Integer} [detailY] number of segments in the y-direction
  * @chainable
  */
-p5.prototype.rect = function(x, y, w, h, detailX, detailY) {
-  p5._validateParameters('rect', arguments);
-
-  if (this._renderer._doStroke || this._renderer._doFill) {
-    var vals = canvas.modeAdjust(x, y, w, h, this._renderer._rectMode);
-    this._renderer.rect([vals.x, vals.y, vals.w, vals.h, detailX, detailY]);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * A triangle is a plane created by connecting three points. The first two
@@ -425,14 +392,6 @@ p5.prototype.rect = function(x, y, w, h, detailX, detailY) {
  * white triangle with black outline in mid-right of canvas.
  *
  */
-p5.prototype.triangle = function() {
-  p5._validateParameters('triangle', arguments);
-
-  if (this._renderer._doStroke || this._renderer._doFill) {
-    this._renderer.triangle(arguments);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 module.exports = p5;
--- a/src/core/attributes.js
+++ b/src/core/attributes.js
@@ -103,10 +103,6 @@ p5.prototype.ellipseMode = function(m) {
  * 2 pixelated 36x36 white ellipses to left & right of center, black background
  *
  */
-p5.prototype.noSmooth = function() {
-  this._renderer.noSmooth();
-  return this;
-};
 
 /**
  * Modifies the location from which rectangles are drawn by changing the way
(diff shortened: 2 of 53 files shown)


Spongman:
fixes #2528, #2370, #2344

partially addresses #2369,

kjhollen:
Hey, thanks for submitting this! WOW, this is an immense amount of great work! This patch is also huge, and you’re an experienced contributor now, so let’s talk a little about how we can streamline the PR process. :)

If you include “closes #xxxx” in the commit message summary, it’ll close the corresponding issue automatically when the PR is accepted. The details you’ve started adding for the PRs are awesome and should be included in the commit messages, too—then they can be visible in git history & command line tools, which will be helpful for other developers. Can you amend the commits to include those details?

It’s good to aim for one PR per GitHub issue, when possible. Keeping the PR succinct not only helps us review faster, but also makes it easier to browse the git history if another developer needs to pinpoint the source of a change later. I suggest breaking this PR up like this:

- one PR for GL architecture changes: push/pop to include render state (please include existing issue # in commit msg), buffer refactoring, caching of shader uniforms, making p5.Geometry object with immediateMode
- one or more PRs for new features: lighting, new immediate mode drawing modes
- one PR for thunkRenderMethods—I also suggest opening an issue for this one, so that more contributors will see & comment. this feels like a significant philosophical shift in the architecture library-wide, so I think it'll be good to get more input, and I wouldn't want these great GL changes to get stalled while that's discussed.

Does this seem reasonable? I'm happy to help with dividing things up, if that's useful—I'm eager to get the GL updates merged so I can use them for text rendering!

Spongman:
yeah, i agree the size is definitely a problem. i did try splitting some of the pieces apart before submitting, but i ran into a complexity brick wall. the main problem is that so many of the changes mentioned in the summary above are dependent on each other. i'll have another go at splitting out the thunking stuff, i think that might be the simplest.

i didn't want to add 'closes' because i didn't want to assume i was correct in that assumption. i can in the future, but i think i'd rather have someone else double-check that it's actually true ;-)

Spongman:
ok, i split out the thunk stuff into a separate PR: #2555

Spongman:
ok, i have pulled out all the parts i think i can. i don't think there's anything else that can be separated, not without a significant rewrite in order to produce a throwaway intermediate.

Spongman:
ugh. please stand by. it looks as though all these git gymnastics have turned my perfectly good PR into a jumbled mess of nonsense. ugh. i cannot understate my disdain for this terrible tool.

lmccart:
@Spongman it's really difficult to review and merge a PR this large. though it takes longer, it's necessary to break out the code in pieces. if one change is dependent on the next, we will just need to review the first before you build the next PR on top of it. otherwise it puts the reviewer in a position of having too many different issues and changes to be able to accurately review, making the process much longer than it needs to be.

in general, if you are adding features, making significant changes, or refactoring, it's important to first discuss on a github issue and iron out the details before submitting the code. thanks!

Spongman:
yeah, i split out most of the pieces into separate commit as per Kate's request. i can move those to separate PRs if you'd like, but they don't really make sense on their own - they're just a prelude to the main course (72f5dd8d241c928e95913b05b26e6a5c35270b4b) which is really hard to split any further.

lmccart:
Could you please split these ones into their own PR.

* Expands push/pop to include render state.
* Expands the set of material properties to include those from processing: emissiveMaterial, specularMaterial, shininess.
* Plus additional lights, noLights, lightSpecular, lightFalloff methods for setting light properties. In addition, the rendered output more closely matches that of processing, for example materials/vertices can now have distinct ambient, diffuse & specular colors, each lit by their corresponding light components, and directional lights have specular reflections.

Also, the thunk code is still in there, please remove this from this PR.

Thanks!

Spongman:
instead of the thunking code, i'm going to submit a PR that:
- moves renderer specific methods that are currently in `p5.prototype` into their own renderers,
- adds 'not implemented' methods where appropriate, ie. for _every_ webgl call (like `pointLight`, etc...) there will be a corresponding method in `p5.Renderer2D` just to show the not implemented error for that method.
- add boiler-plate code to `p5.prototype` passing its arguments into the renderer-specific version of that method.

this is going to be a pretty big change. i just want to make absolutely sure this is what you want?

kjhollen:
@Spongman can you open a new issue outlining why you want to make the changes suggested in your most recent post on this thread?

and are you willing to update this PR to remove the thunkRendererMethods commit? I can also review this patch much faster, and with less back and forth on further changes, if you move the lighting feature additions into a separate PR.

lmccart:
hi @Spongman I'm going to close this for organizational purposes. please feel free to reopen if you make the requested changes and we can review. thanks!

Spongman:
i _did_ start to break this out into separate PRs a while back -- i did the 'push/pop' thing first, but that PR just got hung up on an out-of-scope issue regarding browser compatibility. i closed the PR i didn't want my PR to be responsible for fewer people being able to run p5 on their machines. i'm honestly a little reticent to continue splitting this out since to be honest i feel like it will be a waste of time in the end. i'm going to continue to make changes and improvements in my branch, but those will not continue to be reflected here since github doesn't track changes to branches in closed PRs.

kjhollen:
Please find a way to break this PR into smaller pieces if you'd like it to be reviewed again in the future, as requested. Further changes to the same branch—producing a larger PR—are not likely to be accepted.

PR #2552, opened 2018-01-17
angleMode helper functions

Spongman:
this PR adds 3 internal helper functions to the `p5` instance that help tidy up some of the angleMode handling code:
- `_toRadians(angle)`, converts an angle from the current angleMode to RADIANS. (for example, for passing into `Math.sin()` )
- `_fromRadians(angle)`, converts an angle from RADIANS to the current angleMode. (for examples, for converting the output of `Math.atan2()` )
- `_toDegrees(angle)`, converts an angle from the current angleMode to DEGREES (this one isn't actually used currently).


src/math/trigonometry.js
```diff
@@ -353,4 +325,49 @@ p5.prototype.angleMode = function(mode) {
...
+ * converts angles from the current angleMode to RADIANS
+ *
+ * @method _toRadians
+ * @private
+ * @param {Number} angle
+ * @returns {Number}
+ */
+p5.prototype._toRadians = function(angle) {
+  if (this._angleMode === constants.DEGREES) {
+    return angle * 2 * Math.PI / 360;
```
kjhollen: use constants.DEG_TO_RAD here

src/math/trigonometry.js
```diff
@@ -353,4 +325,49 @@ p5.prototype.angleMode = function(mode) {
...
+ * converts angles from the current angleMode to DEGREES
+ *
+ * @method _toDegrees
+ * @private
+ * @param {Number} angle
+ * @returns {Number}
+ */
+p5.prototype._toDegrees = function(angle) {
+  if (this._angleMode === constants.RADIANS) {
+    return angle * 360 / (2 * Math.PI);
```
kjhollen: likewise, use constants.RAD_TO_DEG here

src/math/trigonometry.js
```diff
@@ -353,4 +325,49 @@ p5.prototype.angleMode = function(mode) {
...
+ * converts angles from RADIANS into the current angleMode
+ *
+ * @method _fromRadians
+ * @private
+ * @param {Number} angle
+ * @returns {Number}
+ */
+p5.prototype._fromRadians = function(angle) {
+  if (this._angleMode === constants.DEGREES) {
+    return angle * 360 / (2 * Math.PI);
```
kjhollen: use RAD_TO_DEG here

src/core/p5.Renderer2D.js
```diff
@@ -1128,20 +1128,12 @@ p5.Renderer2D.prototype.scale = function(x, y) {
 };
 
 p5.Renderer2D.prototype.shearX = function(angle) {
```
kjhollen: hmm, maybe worth a comment that unlike it's p5.prototype counterpart, this method expects radians only? same for shearY. I see that this is already true for rotate, so it makes sense to handle this the same way—maybe there's room to add a note about this general pattern somewhere?

Spongman: yeah, i think the assumption now is that the renderers _only_ deal in radians, and the angleMode stuff is all contained within the p5 instance. i'll put a comment to that affect near the `_angleMode` declaration.

src/webgl/camera.js
```diff
@@ -217,12 +217,12 @@ p5.RendererGL.prototype.perspective = function(fovy, aspect, near, far) {
...
   }
 
-  this.cameraFOV = fovy;
+  this.cameraFOV = this._pInst._toRadians(fovy);
   this.cameraAspect = aspect;
   this.cameraNear = near;
   this.cameraFar = far;
 
-  this.uPMatrix = p5.Matrix.identity(this.pInst);
+  this.uPMatrix = p5.Matrix.identity();
```
kjhollen: given other changes in this patch, if rotate is called on this later, won't the angle mode be lost?

kjhollen:
Tests are passing & everything looks good. I'll merge this after the review question is resolved!

src/webgl/camera.js
Spongman: yes indeed!! but this is not a problem now since the _all_ the renderer's methods are _only_ passed radians, uPMatrix is never manipulated with degrees. so, in fact, we _do_ want the matrix to ignore `angleMode`.

kjhollen: Ok, thank you. I need to run the tests again because additional changes were added in the newest commit.

Spongman: yeah, sorry about that. i had to merge again and fix the conflicts.

kjhollen:
this seems fine. noticed issue #2594 while testing, but not related to this patch (present in master branch, too). thanks!

PR #2555, opened 2018-01-17
thunk renderer methods

Spongman:
(this is part of the 'Render pipeline changes' PR)

Previously a lot of 3d-specific methods were declared directly on p5, or stub methods were used to thunk down to p5’s _renderer. Now there's a method 'thunkRendererMethods’ that runs during startup and adds these thunks automatically and additionally adds warning messages for unimplemented methods. This removes a bunch of boiler-plate code, and ensures that, for example, calling 'sphere()’ on the canvas renderer just outputs an error instead of crashing.

### DIFF
--- a/src/core/2d_primitives.js
+++ b/src/core/2d_primitives.js
@@ -233,15 +233,7 @@ p5.prototype.ellipse = function(x, y, w, h, detailX, detailY) {
  * @param  {Number} z2 the z-coordinate of the second point
  * @chainable
  */
-p5.prototype.line = function() {
-  p5._validateParameters('line', arguments);
-
-  if (this._renderer._doStroke) {
-    this._renderer.line.apply(this._renderer, arguments);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * Draws a point, a coordinate in space at the dimension of one pixel.
@@ -268,15 +260,7 @@ p5.prototype.line = function() {
  *4 points centered in the middle-right of the canvas.
  *
  */
-p5.prototype.point = function() {
-  p5._validateParameters('point', arguments);
-
-  if (this._renderer._doStroke) {
-    this._renderer.point.apply(this._renderer, arguments);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * Draw a quad. A quad is a quadrilateral, a four sided polygon. It is
@@ -322,15 +306,7 @@ p5.prototype.point = function() {
  * @param {Number} z4
  * @chainable
  */
-p5.prototype.quad = function() {
-  p5._validateParameters('quad', arguments);
-
-  if (this._renderer._doStroke || this._renderer._doFill) {
-    this._renderer.quad.apply(this._renderer, arguments);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * Draws a rectangle to the screen. A rectangle is a four-sided shape with
@@ -392,16 +368,7 @@ p5.prototype.quad = function() {
  * @param  {Integer} [detailY] number of segments in the y-direction
  * @chainable
  */
-p5.prototype.rect = function(x, y, w, h, detailX, detailY) {
-  p5._validateParameters('rect', arguments);
-
-  if (this._renderer._doStroke || this._renderer._doFill) {
-    var vals = canvas.modeAdjust(x, y, w, h, this._renderer._rectMode);
-    this._renderer.rect([vals.x, vals.y, vals.w, vals.h, detailX, detailY]);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * A triangle is a plane created by connecting three points. The first two
@@ -427,14 +394,6 @@ p5.prototype.rect = function(x, y, w, h, detailX, detailY) {
  * white triangle with black outline in mid-right of canvas.
  *
  */
-p5.prototype.triangle = function() {
-  p5._validateParameters('triangle', arguments);
-
-  if (this._renderer._doStroke || this._renderer._doFill) {
-    this._renderer.triangle(arguments);
-  }
-
-  return this;
-};
+// see thunkRendererMethods
 
 module.exports = p5;
--- a/src/core/attributes.js
+++ b/src/core/attributes.js
@@ -103,10 +103,6 @@ p5.prototype.ellipseMode = function(m) {
  * 2 pixelated 36x36 white ellipses to left & right of center, black background
  *
  */
-p5.prototype.noSmooth = function() {
-  this._renderer.noSmooth();
-  return this;
-};
 
 /**
  * Modifies the location from which rectangles are drawn by changing the way
@@ -201,10 +197,7 @@ p5.prototype.rectMode = function(m) {
  * 2 pixelated 36x36 white ellipses one left one right of center. On black.
  *
  */
-p5.prototype.smooth = function() {
-  this._renderer.smooth();
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * Sets the style for rendering line endings. These ends are either squared,
@@ -231,17 +224,7 @@ p5.prototype.smooth = function() {
  * 3 lines. Top line: rounded ends, mid: squared, bottom:longer squared ends.
  *
  */
-p5.prototype.strokeCap = function(cap) {
-  p5._validateParameters('strokeCap', arguments);
-  if (
-    cap === constants.ROUND ||
-    cap === constants.SQUARE ||
-    cap === constants.PROJECT
-  ) {
-    this._renderer.strokeCap(cap);
-  }
-  return this;
-};
+// see thunkRendererMethods
 
 /**
  * Sets the style of the joints which connect line segments. These joints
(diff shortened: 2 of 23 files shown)


kjhollen:
reopening for consideration separate from other render pipeline changes & requesting a review from @lmccart because the changes cover so much of the library.

kjhollen:
@Spongman can you please also amend this commit message so that it has a thorough explanation of what it entails? the current message is only 3 words long. :) (if you haven't amended a commit before, there are some helpful instructions here: https://help.github.com/articles/changing-a-commit-message/) thanks!

Spongman:
do you want me to include the PR's 1st comment in the commit message?

Spongman:
closes #2528

kjhollen:
yes, please! and any other info that you think might be relevant: if you can elaborate on how parameter validation works here, that would be helpful, too. another thing that I can think of is that if the list of methods isn't complete, which ones were left out, and why?

Spongman:
apparently you can't do multi-line rewords in an interactive rebase :( maybe you can paste it in there if you know the magic git incantation to do it?

i don't have anything to add wrt. the parameter validation - it just uses the same method as before. and i'm not quite sure what you mean about the completeness of the list. it's not meant to be complete in any way, it just enumerates the methods that are thunked.

kjhollen:
hmm what I mean is that it's helpful to have a note in the commit log about whether or not you think there's any remaining work to do to make this refactor/reorganization complete. should all external API methods eventually be handled in this way? I think that might be your intent, but I'm not sure because some methods (for example, beginContour) are missing and the commit messages + code itself don't explain why. even if you plan to submit another PR to do this soon, it's nice to have a clear record for other devs! and we can open an issue for remaining work so that others know you're working on it.

Spongman:
oh, no. this isn't intended to be done for all methods, it's really just a way to remove most of the `this._renderer.<method>.apply()` boilerplate code that's often duplicated (sometimes erroneously), and to unify/implement most of the 'not implement in this renderer' warnings. the set of methods i considered are those that just look like thunks into the renderer. i didn't change methods like `beginContour` simply because they don't match that criteria. they _might_ in the future (in which case they could easily be added to init.js as thunks), but i don't intend to change that list further as part of this PR.

i usually consider a pull request to be ready to pull unless specifically stated otherwise in the PR comment. although, usually i'll prefer to just link to a forked branch in a discussion issue rather than submit an incomplete PR. +/- bugs, of course ;-)

Spongman:
image didn't make the cut because it's not a simple thunk into the renderer. i really didn't want to rewriting anything complex like that. i feel that when/if someone implements the 3D `image()` method, _then_ would be a good time to work out if any of what's currently in `p5.prototype.image` can be shared or not, and whether or not what's left constistutes a simple thunk into the renderer. i would think it a good thing to share as much code as possible between renderers, so i didn't want to blindly split everything apart completely.

lmccart:
thanks for this @Spongman. after looking at this closer, I appreciate that this is more concise, but I worry about making the codebase more confusing for beginner developers. in general, we've tended toward making some trades in terms of elegance and performance, in order to keep the code really readable and easy to follow for someone new to the project. in this case, the repetition of lines like the validate parameters I think can be helpful for people learning the patterns of the library. toward that goal, my preference would be to leave this out for now. I'm really grateful for all your work here and don't want you to feel like I'm just shutting this down, so let me know what your thoughts are about this.

Spongman:
i can see your point, but unfortunately this area specifically -- the delegating of p5 instance methods to renderer implementations -- is so error-prone that it's probably the one area left where something like this makes sense. this isn't really about removing repetition of the `validateParameters` calls (although that _is_ a necessary side-effect), it's more about correctly handling the renderer method delegation and ensuring that the unimplemented method calls are handled correctly.

there are at least three other places in the library that do something similar:
- the delegation of `p5` methods and properties in the `window` object for global-mode and. again, the alternative to doing this would be a whole bunch of similar thunk methods.
- the patching and un-patching of methods for `preload` handling.
- the delegation of `p5` methods into `Graphics` objects. should we get rid of this and just explicitly implement all the boiler-plate methods in the p5.Graphics.js file?

if anything, i think this PR actually makes it simpler to develop error-free renderer code in the future. there's no more requirement on picking which buggy thunk implementation to copy - you just add your function name to an array.

limzykenneth:
I only had a quick look and don't have a strong opinion on this, but one thing that jumps out at me about this is it will break the link between the reference and the source code. For example on the [line documentation](https://p5js.org/reference/#/p5/line) at the bottom of the page you see this line
> Find any typos or bugs? line() is documented and defined in src/core/2d_primitives.js. Please feel free to edit the file and issue a pull request!

Which nicely links to the documentation source code and the function's source code itself. 

Making this change will move the function source code out of that link which can maybe make some functions harder to debug and we miss the opportunity for a first time contributor to be able to look at the source code for the documentation they are fixing. In a way, it feels like separating the source code and the documentation.

Spongman:
The problem is that now there are two renderers that comment doesn't really make sense any more, the implementation of many of these methods exists in two places - eg where is ellipse implemented?

Spongman:
Besides, the 2d line is _still_ implemented in 2d_primitives.js in `p5.Renderer2D.prototype.line` - that hasn't changed. What's changed is that now the `p5.prototype.line` method, which doesn't really do anything and probably shouldn't be changed (without fully testing both 2D & 3D versions of `line`), is gone.

kjhollen:
I agree that with two renderers it's worth spending some time thinking about (and discussing) how we might make the delegation more consistent and easier to understand—but I don't think that this patch solves the consistency issue.

"thunk" was a new technical word for me, so I had to look it up. my understanding, now, is that a thunk is an argument-less method with internal state that allows it to be executed at some later date. that's not what I see here for the methods that were chosen to be refactored in this patch, so it's still unclear to me why some methods were included and others were not. 

Is there a delegation method we could adopt that would more cleanly apply to all cases (and still be beginner friendly enough to point a contributor to the right place in the docs)? I'm inclined to move this discussion to an open issue to solicit more ideas and close the PR for now (for housekeeping purposes), does that seem okay?

Spongman:
actually thunks are functions used for several purposes including virtual method dispatch, overlay handling, and wrapping additional functionality around existing functions. all of these are what's going on  here. i agree the term might be a little obscure - i first learned of it coding 16-bit DOS drivers. but, as far as names go, it's the right tool for the job.

Spongman:
> Is there a delegation method we could adopt that would more cleanly apply to all cases

i'm not sure we _want_ to delegate all methods in the same way, though. i think it's best to keep as much functionality in the p5 instance as possible and only resorting to an automatic delegation mechanism (thunks) when the differences between the two implementations are so different that it's more confusing to have them in the same method than not.

first, though, i think it's worth discussing, if we don't change anything what needs to be done to fix the existing bugs. for example, methods like `pointLight` currently crash if you call them when in `P2D` mode. so do we:
- add extra boilerplate code to all these methods to check the correct renderer is set. in my experience the most problematic code of all is boilerplate code. it's often misunderstood by newcomers, neglected, copy/pasted erroneously and ends up being a huge source of bugs. much of the work of modern language design is centered around meta-programming designed precisely to reduce these kinds of bugs.
- and/or add delegating methods to `p5.prototype` where they don't exist now, and add 'not implemented' methods for each renderer for each method that it doesn't implement. again, this is a whole bunch of dulpicate code that needs to be manually maintained. every time a new method is added in one renderer, the delegate needs to be added and the 'not implemented' method needs to be added to the other renderer. imagine the management mess if someone adds another renderer!

kjhollen:
I don't think pulling in obscure computing terms is a good idea for this project, when our primary goal is supporting new programmers who want to create visual and interactive works (or experienced programmers that don't want a lot of overhead for doing the same). Moreover, the comment "see thunkRendererMethods" is unlikely to be helpful to a novice level contributor and it's not obvious where they would go to learn more. It makes the placement of the documentation in various files seem arbitrary and distances new contributors editing documentation from the code. 

I like the second approach you've mentioned here. It doesn't seem so bad to me that the renderer has to implement the method to declare its non-existence. I think it's clearer and more direct for new contributors—and it gives us a chance to distinguish between "not yet implemented" or "never will be implemented, change your render mode."

I totally get that there are more efficient ways to handle this, and the initial solution you've presented is elegant. p5 is pretty different from other libraries in that it aims to be accessible to new coders—not just in its public facing API, but in the core source code as well. It means we lose on performance sometimes, or that we don't always get to use some of the fancier features of Javascript, but it means we can have a different kind of community that is more welcoming to beginners.

We also don't have to rush to implement this right now, either way. Maybe there's a good middle-ground solution that we haven't thought of yet? Opening an issue for this can be a good way to outline what kind of problem you're trying to solve and getting more feedback from other users and contributors.

Spongman:
ok, so if the goal is to make the actions of the library more explicit in the code and reduce the amount of 'tricky' code, then should we also remove the patching of the 'preload' method and add corresponding boilerplate code to them, for that code is very similar to what i have submitted here - it's dynamically patching thunks on top of library methods in order to add additional functionality. we could easily remove that switching code and add checks to each of the preload methods and make it more obvious to newcomers what is going on. or how about the `p5.prototype` methods that are dynamically copied into the `p5.Graphics` instance in its constructor? those methods could quite easily be explicitly stubbed out in `p5.Graphics.js' to make it clearer what's going on - for those, again are doing precisely what i'm suggesting here.
maybe i'm being a little obsequious here, but it's only because i don't really understand the reasoning behind why one part of the library can do this while another cannot.

kjhollen:
This isn't a constructive argument for accepting the PR. I think part of trouble we're having is that the goals of this patch weren't discussed before it was implemented, but it touches/destabilizes a significant portion of the library, which makes me appropriately anxious about merging it. The minor issues that it fixes could have been resolved with much less code and less risk to the library overall.

Let's open an issue for further discussion, but I'm closing this PR for now. I would love to merge some of the other things you've been working on that depend on this patch—I can be available to help with any git logistics to make that happen.

PR #2564, opened 2018-01-19
remove unused `detailY` parameter from `ellipse` method

### LINKED ISSUE #2210: WebGL - Ellipse should not take detailY as an arg

Spongman:
this parameter isn't used any more.

closes #2210

### DIFF
--- a/src/core/2d_primitives.js
+++ b/src/core/2d_primitives.js
@@ -162,10 +162,9 @@ p5.prototype.arc = function(x, y, w, h, start, stop, mode) {
  * @param  {Number} y
  * @param  {Number} w
  * @param  {Number} h
- * @param  {Integer} detailX number of segments in the x-direction
- * @param  {Integer} [detailY] number of segments in the y-direction
+ * @param  {Integer} detail number of radial sectors to draw
  */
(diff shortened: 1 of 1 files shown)


kjhollen:
thank you!

PR #2566, opened 2018-01-20
fix createGraphics(WEBGL) (3rd attempt)

Spongman:
sorry, this should have made it into the 'createGraphics()' fix PR. i'm having a bit of a git episode again today :(

- remove `fill()` call from p5.RendererGL constructor

without it, the createGraphics bug isn't actually fixed :(


kjhollen:
ok! let's get rid of the now-stale comment about fill() in the two lines above this edit, then I think we'll be ready.

kjhollen:
thanks! sorry was just double checking this and realized that I should have asked if the same changes are necessary for stroke? right now the stroke method is still being called, but the properties are not set directly.

Spongman:
i cannot for the life of me get this bug to repro now on master. so i'm going to close on the assumption that something else was going on at the time i fixed this that was either a local issue, or has been fixed since.

PR #2567, opened 2018-01-20
Icosphere

Spongman:
it's like the long/lat sphere, but all the sides are essentially the same size & shape, and the angles between faces are more consistent.

added bonus: detail 0 is another platonic solid! (the best one)

image

### DIFF
--- a/src/webgl/p5.Texture.js
+++ b/src/webgl/p5.Texture.js
@@ -31,8 +31,8 @@ p5.Texture = function(renderer, obj) {
   this.mipmaps = false;
   this.glMinFilter = gl.LINEAR;
   this.glMagFilter = gl.LINEAR;
-  this.glWrapS = gl.CLAMP_TO_EDGE;
-  this.glWrapT = gl.CLAMP_TO_EDGE;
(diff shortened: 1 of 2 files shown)


lmccart:
hi @Spongman, as described in the [developer docs](https://github.com/processing/p5.js/wiki/Development):

> Implement a new feature :
> 
> If you have an original idea about a feature, open an issue describing your idea preferably in the format provided there, and let us know you're interested! Initiate a discussion, wait for the green signal, and then assign it to yourself!

new features need to first be discussed as issues before submitting as PR. if you're interested in adding this please open an issue and we can determine as a group if it makes sense.

Spongman:
Oh, I missed that part. I was hoping this PR might serve as a forum for such a discussion?

lmccart:
yeah, it seems like we're running into some issues lately where the reviewing of code and merging gets mixed up with discussing the addition of new features or code changes, making it hard to move things forward. so I'd like to be more strict about enforcing the protocol of first a github issue and discussion, then if generally approved, code changes and a PR. hope that makes sense.

PR #2570, opened 2018-01-21
Added example for flush()

hackertron:
added example for Printwriter flush() method .  
refering to this issue : https://github.com/processing/p5.js/issues/1954


limzykenneth:
Hi, could you add a brief description for the method, explaining what `flush` means and what it does? Thanks

src/io/files.js
```diff
@@ -1125,6 +1125,7 @@ p5.PrintWriter = function(filename, extension) {
     this.content += data + '\n';
   };
   /**
+   * Flushes the PrintWriter object that calls it
```
limzykenneth: I would actually go for something more descriptive, as "flush" means nothing to someone unfamiliar with a PrintWriter. Something like "Clears the data already written to the PrintWriter"

hackertron:
Description : Clears the data already written to the PrintWriter object  👍

Spongman:
the only issue i have with this is that it's not actually necessary to call `flush()` immediately before a call to `close()` since `close` calls flush internally.

also, another weird thing, and this has nothing to do with your example, is that usually when you have a writer API with a `flush()` method, that method's action is usually to force writing the current contents of any buffers in the writer to its output. in p5's `PrintWriter` the `flush()` method just clears the internal buffer. i realize that this is because the write (usually) has to be done in one step since the `downloadFile` operation is on a Blob, not a stream. it sesms to me that a better name for this method would be `clear()` or `reset()`.

hackertron:
@Spongman  understood your first point but I just wanted to show flush example in it's simplest form .

lmccart:
@Spongman I see your point, but I think this is nice as a simple example. we can always add a second more complex example that demonstrates a clear use case.

Issue #2575, opened 2018-01-22
TypeScript stroke definition missing


rdelpeso:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Other (specify if possible)

TypeScript Typings Generation.

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: NPM p5.js v0.6.0 January 19, 2018
- Web browser and version: Not Applicable
- Operating System: (Not really applicable) Linux Centos7 + xfce4 + Visual Studio Code 1.20.0-insider
- Steps to reproduce this:

While using p5.js in a typescript project, attempting to call the method stroke with a single parameter (Grayscale integer value) typescript complaints that there is no definition for the shape of method calling.

    stroke(0)
    // [ts] Argument of type '0' is not assignable to parameter of type 'Color'.

After tracing through the p5.d.ts I noticed that the following definition is missing:

    declare class p5 {
    ...
      /**
      *   Sets the color used to draw lines and borders 
      *   around shapes. This color is either specified in 
      *   terms of the RGB or HSB color depending on the 
      *   current colorMode() (the default color space is 
      *   RGB, with each value in the range from 0 to 255). 
      *   The alpha range by default is also 0 to 255.  If a 
      *   single string argument is provided, RGB, RGBA and 
      *   Hex CSS color strings and all named color strings 
      *   are supported. In this case, an alpha number value 
      *   as a second argument is not supported, the RGBA 
      *   form should be used. 
      * 
      *  
      *   A p5 Color object can also be provided to set the 
      *   stroke color.
      *
      *   @param gray specifies a value between white and
      *   black
      *   @param [a] opacity of the background relative to
      *   current color range (default is 0-255)
      *   @chainable
      */
      stroke(gray: number, a?: number): p5;
    ...
    }

I cobbled together the above by merging the existing background(gray: number, a?: number): p5; definition together with the documentation from stroke.

The same issue occurs in p5.global-mode.d.ts, the function definition for stroke(number, number?) is also missing.

[...]

PR #2577, opened 2018-01-22
fix & unify color parameter docs

Spongman:
this fixes up the parameter docs for some methods that take color inputs. there were some errors and missing overloads.

### DIFF
--- a/src/color/creating_reading.js
+++ b/src/color/creating_reading.js
@@ -288,25 +288,12 @@ p5.prototype.brightness = function(c) {
  * @param  {Number}        [alpha]
  * @return {p5.Color}
  */
-
 /**
  * @method color
- * @param  {String}        value   a color string
- * @param  {Number}        [alpha]
+ * @param {p5.Color|Number[]|String} color p5.Color object, color components,
+ *                                         or CSS color
  * @return {p5.Color}
  */
-/**
- * @method color
- * @param  {Number[]}      values  an array containing the red,green,blue &
- *                                 and alpha components of the color
- * @return {p5.Color}
- */
-/**
- * @method color
- * @param  {p5.Color}     color
- * @return {p5.Color}
- */
-
 p5.prototype.color = function() {
   p5._validateParameters('color', arguments);
   if (arguments[0] instanceof p5.Color) {
--- a/src/color/setting.js
+++ b/src/color/setting.js
@@ -32,7 +32,13 @@ require('./p5.Color');
  * A p5.Image can also be provided to set the background iamge.
  *
  * @method background
- * @param {p5.Color} color     any value created by the color() function
+ * @param {Number} v1     red or hue value (depending on the current color
+ *                        mode)
+ * @param {Number} v2     green or saturation value (depending on the current
+ *                        color mode)
+ * @param {Number} v3     blue or brightness value (depending on the current
+ *                        color mode)
+ * @param  {Number} [alpha] alpha value
  * @chainable
  *
  * @example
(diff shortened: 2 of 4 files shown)


lmccart:
we split these out so that the different signatures show up separately to help clarify possibilities.
screen shot 2018-01-24 at 6 40 11 pm

this is documented in the inline documentation wiki notes:
https://github.com/processing/p5.js/wiki/Inline-documentation#additional-signatures

Spongman:
err. please take another look at this. when i said 'unify' i meant 'make all the methods that take color parameters look the same. the overloads are still there.

the currect docs are broken wrt. validation.

src/color/creating_reading.js
```diff
@@ -288,25 +288,12 @@ p5.prototype.brightness = function(c) {
...
-/**
- * @method color
- * @param  {Number[]}      values  an array containing the red,green,blue &
- *                                 and alpha components of the color
- * @return {p5.Color}
- */
-/**
- * @method color
- * @param  {p5.Color}     color
+ * @param {p5.Color|Number[]|String} color p5.Color object, color components,
```
lmccart: for example this one would remove the three independent bottom signatures in the image, replacing it with one single line siganture. unless I am missing something?

src/color/setting.js
```diff
@@ -127,52 +133,26 @@ require('./p5.Color');
...
+ *                                 and black.
+ * @param  {Number}        [alpha]
  * @chainable
  */
-
 /**
  * @method background
- * @param  {Number[]}      values  an array containing the red,green,blue &
- *                                 and alpha components of the color
+ * @param {p5.Color|Number[]|String} color p5.Color object, color components,
```
lmccart: this one too

src/color/creating_reading.js
Spongman: indeed, but those are all just different types for representing the same thing - a color. and this is not new, it's just borrowed from most of the other color-taking methods, for example, see the existing `brightness`: https://p5js.org/reference/#/p5/brightness. there are a bunch that are like this. this PR just makes all the color-taking methods look the same, and fixes a bunch of the errors.

there are many instances of methods whose parameters can be of different types. for examples: https://p5js.org/reference/#/p5/image the `img` parameter can be either a `p5.Image` or a `p5.Element`. the docs have always been like this and the validation code that uses them has handled these type alternations since when i first looked at it.

lmccart: in this case we felt it was important because color takes so many different formats that it can become very confusing. the priority here is making things clear for the beginning programmer. i'm pretty set on this, i'm sorry.

Spongman: ok, are you saying that you want `color` to have different docs than all the other methods that take the same parameters?

lmccart: if there are other functions that are as confusing as color, i am open to breaking some of those out as well

Spongman: there's a bunch. basically anything that calls `p5.prototype.color()`, `this.color()` or `this._pInst.color.apply()`.

i think pointLight will require 24 overloads.

lmccart: I feel like you are trying to pick a fight and I'm not interested in that. I'm leaving this thread now.

Spongman: oh no not at all, i'm sorry you got that impression. i'm just trying to work out which methods you want to have these overload separated, and which ones not. i'd like to get the errors in the docs fixed because they are causing errors in the mocha tests that i'm trying to fix.

Spongman:
closes #2575

PR #2578, opened 2018-01-22
add type column to doc parameters table

Spongman:
this PR adds a column to the 'parameters' table in the reference docs. it separates the type(s) from the description and makes both easier to read. adds a zebra stripe to the table.

eg:
image


lmccart:
thanks. I'd like to leave this one as is. I feel that the three columns makes it harder to identify the parameter, and the zebra stripe adds noise to the design where there's already a lot going on.

Spongman:
oh, i copied the zebra stripe stying directly from the example code:

image

lmccart:
yes, the example has zebra stripes because it is multi-line code example.

PR #2579, opened 2018-01-23
Update primitives.js

Spongman:
fix scaled cylinder dimensions.
this is pretty embarrasing :-/


mlarghydracept:
Hm wonder why I didn't catch this in testing? I guess the values passed to cylinder in the tests that use it aren't different enough to notice something like this just by looking. Should figure out how to avoid these kind of easy slip-ups in the future. Maybe a unit test?

PR #2587, opened 2018-01-25
per-renderer push/pop

Spongman:
this moves the handling of pushing & poping renderer-common and per-renderer  style properties out of the p5 class and into the renderer classes themselves. this makes it easier to expand the set of renderer-specific properties stored in the pushed style (for example, the webgl camera & matrix properties are now stored as part of this style object instead of on their own stacks).

### DIFF
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -60,6 +60,46 @@ p5.Renderer = function(elt, pInst, isMainCanvas) {
 
 p5.Renderer.prototype = Object.create(p5.Element.prototype);
 
+p5.Renderer.prototype.push = function() {
+  return {
+    properties: {
+      _doStroke: this._doStroke,
+      _strokeSet: this._strokeSet,
+      _doFill: this._doFill,
+      _fillSet: this._fillSet,
+      _tint: this._tint,
+      _imageMode: this._imageMode,
+      _rectMode: this._rectMode,
+      _ellipseMode: this._ellipseMode,
+      _textFont: this._textFont,
+      _textLeading: this._textLeading,
+      _textSize: this._textSize,
+      _textStyle: this._textStyle
+    }
+  };
+};
+
+function assign(to, firstSource) {
+  for (var i = 1; i < arguments.length; i++) {
+    var nextSource = arguments[i];
+    if (nextSource === undefined || nextSource === null) {
+      continue;
+    }
+
+    for (var nextKey in nextSource)
+      if (nextSource.hasOwnProperty(nextKey)) {
+        to[nextKey] = nextSource[nextKey];
+      }
+  }
+  return to;
+}
+
+p5.Renderer.prototype.pop = function(style) {
+  if (style.properties) {
+    assign(this, style.properties);
+  }
+};
+
 /**
  * Resize our canvas element.
  */
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -82,7 +82,7 @@ p5.Renderer.prototype.push = function() {
 };
 
 // this is implementation of Object.assign() which is unavailable in
-// IE11 and android browsers.
+// IE11 and (non-Chrome) Android browsers.
 // The assign() method is used to copy the values of all enumerable
 // own properties from one or more source objects to a target object.
 // It will return the target object.
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -60,6 +60,8 @@ p5.Renderer = function(elt, pInst, isMainCanvas) {
 
 p5.Renderer.prototype = Object.create(p5.Element.prototype);
 
+// the renderer should return a 'style' object that it wishes to
+// store on the push stack.
 p5.Renderer.prototype.push = function() {
   return {
     properties: {
@@ -98,6 +100,9 @@ function assign(to, firstSource) {
   return to;
 }
 
+// a pop() operation is in progress
+// the renderer is passed the 'style' object that it returned
+// from its push() method.
 p5.Renderer.prototype.pop = function(style) {
   if (style.properties) {
     // copy the style properties back into the renderer
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -81,7 +81,8 @@ p5.Renderer.prototype.push = function() {
   };
 };
 
-// this is implementation of Object.assign()
+// this is implementation of Object.assign() which is unavailable in
+// IE11 and android browsers.
 // The assign() method is used to copy the values of all enumerable
 // own properties from one or more source objects to a target object.
 // It will return the target object.
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -79,6 +79,10 @@ p5.Renderer.prototype.push = function() {
   };
 };
 
+// this is implementation of Object.assign()
+// The assign() method is used to copy the values of all enumerable
+// own properties from one or more source objects to a target object.
+// It will return the target object.
 function assign(to, firstSource) {
   for (var i = 1; i < arguments.length; i++) {
     var nextSource = arguments[i];
@@ -96,6 +100,7 @@ function assign(to, firstSource) {
 
 p5.Renderer.prototype.pop = function(style) {
   if (style.properties) {
+    // copy the style properties back into the renderer
     assign(this, style.properties);
   }
 };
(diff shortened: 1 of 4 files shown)


src/core/p5.Renderer.js
```diff
@@ -60,6 +60,46 @@ p5.Renderer = function(elt, pInst, isMainCanvas) {
...
+      _ellipseMode: this._ellipseMode,
+      _textFont: this._textFont,
+      _textLeading: this._textLeading,
+      _textSize: this._textSize,
+      _textStyle: this._textStyle
+    }
+  };
+};
+
+function assign(to, firstSource) {
```
kjhollen: could this be _assign so we don't clutter this name in the global scope? a comment about it's expected use (here and in renderer subclasses) would also be helpful—or maybe that can be encapsulated in the function name.

src/core/p5.Renderer2D.js
```diff
@@ -1385,13 +1385,18 @@ p5.Renderer2D.prototype._applyTextProperties = function() {
 
 p5.Renderer2D.prototype.push = function() {
   this.drawingContext.save();
+
+  // get the base renderer style
+  return p5.Renderer.prototype.push.apply(this);
 };
 
-p5.Renderer2D.prototype.pop = function() {
+p5.Renderer2D.prototype.pop = function(style) {
```
kjhollen: I understand what's happening here, but a comment about what style is would be helpful for future devs.

src/webgl/p5.RendererGL.js
```diff
@@ -178,11 +175,16 @@ p5.RendererGL.prototype._resetContext = function(attr, options, callback) {
     document.body.appendChild(c);
   }
   this._pInst.canvas = c;
+
+  this._pInst.push();
```
kjhollen: _resetContext sounds to me like it would re-establish renderer defaults, but they look like they're being preserved here. it's worth a comment here to clarify the method's intent.

src/core/p5.Renderer.js
Spongman: actually, this isn't in global scope, this is local to the 'p5.Renderer.js' module. in the minified library it'll probably end up renamed to `a()` ;-)

src/webgl/p5.RendererGL.js
Spongman: yeah, i agree. the reason i did this was that calling `setAttributes` calls `_resetContext` (actually it's the _only_ thing that calls _resetContext) which means that

    fill(150);

results in the effects of the `fill` call being lost.

maybe we can move the push/pop into `setAttributes()` ?

kjhollen: that seems good!

src/core/p5.Renderer2D.js
```diff
@@ -1383,15 +1383,30 @@ p5.Renderer2D.prototype._applyTextProperties = function() {
 // STRUCTURE
 //////////////////////////////////////////////
 
+// a push() operation is in progress.
+// the renderer should return a 'style' object that it wishes to
+// store on the push stack.
+// derived renderers should call the base class' push() method
+// to fetch the base style object.
```
kjhollen: this comment seems like it should go with p5.Renderer rather than p5.Renderer.2D?

src/core/p5.Renderer.js
kjhollen: still, a comment about its intended use, or more descriptive name, would be helpful.

src/core/p5.Renderer2D.js
Spongman: makes sense. i added it to the base renderer also...

src/core/p5.Renderer.js
```diff
@@ -60,6 +60,56 @@ p5.Renderer = function(elt, pInst, isMainCanvas) {
...
+      _textStyle: this._textStyle
+    }
+  };
+};
+
+// this is implementation of Object.assign()
+// The assign() method is used to copy the values of all enumerable
+// own properties from one or more source objects to a target object.
+// It will return the target object.
+function assign(to, firstSource) {
```
kjhollen: hmm, what's the reason to re-implement here instead of just using Object.assign directly?

Spongman: because Object.assign isn't widely supported. i initially used a polyfill from npm, but it had issues with global/instance modes

kjhollen: Seems like only IE 11 doesn't support? http://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.assign

Spongman: on the desktop yeah, but many mobile devices don't support it either. i'm not sure this is really an issue. if i had just included the for loop in the pop() method, would this have even come up?

kjhollen: while working on the GL code over last summer, I came across a number of stale + unused methods that had been left hanging in the code because of a refactor. I just want it to to be obvious where + why methods in the global scope are being used so that they don't get left around in future changes. using a built in method would also meet this goal; having a name of the method that makes its intended use/reach would also meet that goal.

Spongman: oh i absolutely agree. i just think it's a little too soon to add a dependency on Object.assign. i named the function the same name and implemented it in a compatible way so it'll be a more obvious replacement when, down the line, down-level browser compatibility for this method becomes less of an issue, or the browser compatibility requirements change.

i think this is a bigger issue, though. maybe it's worth having a document somewhere that specifically calls out what browsers/platforms the library is intended to support and then decisions like this can be made against that. i think a few extra bytes of zipped .js is a reasonable price to pay for improved comaptability.

kjhollen: we do have a list of supported browsers, https://github.com/processing/p5.js/wiki/Supported-browsers @limzykenneth should we update this for 2018? do you think it's safe to use Object.assign?

limzykenneth: Yes that list is pretty outdated and can use some updates. As far as I can find, only Android web browser and Opera for Android don't support `Object.assign` which accounts for maybe ~2% of browser market (most of which are probably really outdated Androids), for me it's not a big enough share to worry about and unlike other less popular browsers (say Edge), there doesn't seem to be any sign that they will be updated to follow new standards. I'm also considering removing some of the existing polyfills like the `promise` polyfill from the library as we move to support newer browsers.

However, if it is decided that we will still keep the polyfill for now, I would suggest making some modification or just use the one on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) which throws similar errors to native implementation and also attaches directly to the `Object`...object. That can make our future transition from polyfill to no polyfill smoother.

Spongman: can we delay that to another issue/PR to address in the future?

kjhollen: My initial comments (now rendered outdated in the PR) asked for clarity on where this method is expected to be used and to declutter the global/module namespace, which makes using Object.assign a perfectly reasonable solution. Reducing the amount of custom code we have to maintain in favor of a built-in method seems like a win.

src/webgl/p5.RendererGL.js
kjhollen: just noting that it seems this hasn't been addressed yet.

src/core/p5.Renderer.js
Spongman: i'd rather close this PR than kill compatibilty with IE & android devices.

the 25 bytes just aren't worth it.

kjhollen: I might have misunderstood something, but this wouldn't kill compatibility with Android devices and it is also supported by Edge, so I think we're just talking about IE 11. The unsupported Android browser that Kenneth mentioned being <2% of the market is only on older devices and is longer the default browser (it is now Chrome). Are you sure you still want to withdraw on this point?

I'd also be okay with updating the comment to say that this is a polyfill & why it's used instead of Object.assign. I just want the maintenance to be easy to track.

src/webgl/p5.RendererGL.js
Spongman: oops, i missed this one.

limzykenneth:
I personally prefer the MDN polyfill which directly attaches `assign` to the native `Object` object but I don't have too strong of an opinion on this matter.

limzykenneth:
@Spongman Yeah if the idea is just to make it a private utility function just for this part of the library then it certainly made more sense to implement it like this. If no other problem arise should be ok to merge @kjhollen?

kjhollen:
yes! testing now—just wanted to make sure we closed the loop on everything discussed here. 

my primary concern was making sure that the method would have enough context that someone who encounters this in the future could make a good decision about a refactor. we encountered a few things like this in the GL code over the summer that had been left behind, and it was hard to tell where/if they were needed.

PR #2588, opened 2018-01-25
pre-flatten uvs array

Spongman:
this just builds the `uvs` array pre-flattened so it doesn't require flattening before rendering. this is an optimization that will be important when the uvs array is used during immediate-mode rendering.


Spongman:
thanks!

PR #2589, opened 2018-01-25
setUniform: don't crah when passed empty arrays

Spongman:
this fixes crashes that occur when calling `setUniform()` with empty arrays.


kjhollen:
Cool! Can you add some unit tests to ensure this doesn't break again in the future?

kjhollen:
thanks!

PR #2592, opened 2018-01-25
Parameter validation tests

Spongman:
this PR allows the mocha/precommit tests to catch validation errors in the examples, and bugs in the doc comments.

this stemmed from the spew of FES errors that occur during the test runs. the problem is that most of these are intentional FES errors in the tests that the tests weren't previously able to detect. the error messages would just get spewed to the output masking any _actual_ FES errors that were _not_ expected.

- allows parameter validation errors to trigger mocha test failures. the tests set a global flag which causes custom Errors to be throw when an FES error occurs. some of the tests _expect_ to catch these errors, and any that aren't expected cause the test run to fail.
- fixes some erroneous doc comment parameter definitions. in order to get the tests to actually pass now some of the previously masked, unexpected FES errors need to be fixed.
- fixes the parameterValidation handling of undefined/null parameters.
- uses a scoring mechanism to handle 'undefined' trailing parameters, improve error applicability and avoid extraneous error array allocation. 

the first three commits hold the bulk of the changes and can each be considered separately, but they do require each other (and the additional requested changes) in order to work correctly.

### DIFF
--- a/src/color/setting.js
+++ b/src/color/setting.js
@@ -469,6 +469,12 @@ p5.prototype.colorMode = function(mode, max1, max2, max3, maxA) {
 /**
  * @method fill
  * @param  {String}        value   a color string
+ * @chainable
+ */
+
+/**
+ * @method fill
+ * @param  {Number}        gray   a gray value
  * @param  {Number}        [alpha]
  * @chainable
  */
@@ -715,6 +721,12 @@ p5.prototype.noStroke = function() {
 /**
  * @method stroke
  * @param  {String}        value   a color string
+ * @chainable
+ */
+
+/**
+ * @method stroke
+ * @param  {Number}        gray   a gray value
  * @param  {Number}        [alpha]
  * @chainable
  */
--- a/src/core/error_helpers.js
+++ b/src/core/error_helpers.js
@@ -147,6 +147,7 @@ if (typeof IS_MINIFIED !== 'undefined') {
 
   var docCache = {};
   var builtinTypes = [
+    'null',
     'number',
     'string',
     'boolean',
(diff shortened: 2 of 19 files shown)


Spongman:
re-opened since the shader test fixes are in.

Spongman:
closing/reopening due to npmjs 'losing' packages! https://twitter.com/npmstatus/status/959150881676541952

Spongman:
fixes #2740, adds test to include these cases.

meiamsome:
I have a feeling that we should drop most of the 'validation' tests anyway - for example the following:
```javascript
    test('wrong param type at #0', function() {
      assert.validationError(function() {
        c = 20;
        val = myp5.alpha(c);
      });
    });
```
This doesn't really test anything related to the alpha function, and is more a test of FES/the docs than of the functions and so IMO it should not be in the unit tests for the alpha function. I think it would make sense to have a small set of functions directly testing FES (With perhaps dummy functions added to the p5 instance with specific structure so it's not bound to the API at all.

If we drop all such tests, then I don't think we lose any real coverage - any tests that are not updated when the API changes or unintended API changes will start throwing ValidationErrors and flag themselves up anyway. Ideally, we can replace them with more thorough testing of the function's intended functionality, but I would just delete them for now.

test/unit/spec.js
```diff
@@ -37,3 +37,15 @@ Object.keys(spec).map(function(folder) {
...
+
+// a custom assertion for validation errors that correctly handles
+// minified p5 libraries.
+assert.validationError = function(fn) {
+  if (p5.ValidationError) {
+    assert.throws(fn, p5.ValidationError);
+  } else {
+    assert.doesNotThrow(fn, Error, 'got unwanted exception');
+  }
+};
```
meiamsome: I don't really think this should be in the `spec.js` file. I was looking at adding some helpers in too and I think the best course of action is probably to add a `p5_helpers.js` file to the `test/js/` folder and then add it in the test html files, similar to how `chai_helpers.js` is added.

Spongman: Ok... I know very little about mocha or how it's files are supposed to be structured. This is the first place I found that worked for both the in-browser and command-line execution of the tests. Requiring a per-suite setting of this global flag seems a little fragile.

Spongman: Ok... I know very little about mocha or how it's files are supposed to be structured. This is the first place I found that worked for both the in-browser and command-line execution of the tests. Requiring a per-suite setting of this global flag seems a little fragile.

Spongman:
The aim of this PR wasn't to restructure the tests - just to correctly manage the errors of the existing tests. I agree that what you suggest sound like a good idea, but I think it might be beyond the scope of this PR.

test/unit/spec.js
meiamsome: Yeah, it is a bit fragile - really the code in the html files should be factored out into a js file at least, but it hasn't been done yet. I see there's remnants of you attempting to set the `p5._throwValidtionErrors` in those files in the diffs. I think it would be best to have a short `<script>p5._throwValidationErrors = true;</script>` in the HTML immediately after the p5 loading script?

meiamsome:
Yes, that makes sense. I must have missed your edits to `tests/unit/core/error_helper.js` when I was looking through before because these look very much like the tests I meant would be good to make sure were checking all of FES.

Spongman:
ah. I think I remember why I did it this way (it's been awhile). I used to need to set the _throwValidationErrors at the beginning of each test because some tests would set it to false due to legacy behavior. I have since changed those tests and added the assertions so I think I should be able to clean this up along the lines you mentioned.

Spongman:
ok, i think that's a bit cleaner. hopefully that's close to what you had in mind.

Spongman:
fixes #2779

src/core/transform.js
```diff
@@ -222,8 +222,8 @@ p5.prototype.rotate = function(angle, axis) {
  * 3d box rotating around the x axis.
  */
 p5.prototype.rotateX = function(angle) {
-  p5._validateParameters('rotateX', arguments);
   this._assert3d('rotateX');
+  p5._validateParameters('rotateX', arguments);
```
lmccart: for example, I think these additions of validateParameters calls could be in a separate PR

src/webgl/light.js
```diff
@@ -266,7 +277,12 @@ p5.prototype.directionalLight = function(v1, v2, v3, x, y, z) {
 p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
   this._assert3d('pointLight');
   //@TODO: check parameters number
-  var color = this._renderer._pInst.color.apply(this, [v1, v2, v3]);
+  var color;
+  if (v1 instanceof p5.Color) {
+    color = v1;
+  } else {
+    color = this.color(v1, v2, v3);
+  }
```
lmccart: these changes for color param handling could be in a separate PR

src/color/setting.js
```diff
@@ -715,6 +721,12 @@ p5.prototype.noStroke = function() {
 /**
  * @method stroke
  * @param  {String}        value   a color string
+ * @chainable
+ */
+
+/**
+ * @method stroke
+ * @param  {Number}        gray   a gray value
```
lmccart: these signature changes could be in a separate PR

src/core/error_helpers.js
```diff
@@ -159,26 +160,53 @@ if (typeof IS_MINIFIED !== 'undefined') {
   // validateParameters() helper functions:
   // lookupParamDoc() for querying data.json
   var lookupParamDoc = function(func) {
```
lmccart: it seems like the rewrites to the lookupParamDoc and testParamType methods could be there own PR, separate from the validation error addition.

lmccart:
thanks @Spongman. this is good stuff. however, I can't review it when it's so much together like this. I'm going to close this PR, please reopen individual ones with pieces broken out. thank you.

Spongman:
i'm afraid i cannot break anything out as any of the pieces i remove will fail to work on their own. as i explained in the explanation at the top the (first 3) separate commits can be considered on their own, but they do depend on each other in order to work correctly. i could break them out into separate PRs but each of those PRs would not be able to pass the tests without the others.

PR #2595, opened 2018-01-26
fixes #2539, quiets missing example output for WebGL

### LINKED ISSUE #2539: loadPixels(x,y,w,h) ignores arguments

kjhollen:
removes "missing example" warnings for WebGL methods that do not need
to be part of the public API, and the renderer itself (which wasn't generating
a page).

in this process I noticed that the base p5.Renderer class isn't private like the p5.RendererGL one is, but that it doesn't appear in the docs anyway. I couldn't track down why that was—if there's another way that I should have approached this, let me know!

### DIFF
--- a/src/webgl/p5.RendererGL.Immediate.js
+++ b/src/webgl/p5.RendererGL.Immediate.js
@@ -22,6 +22,7 @@ var constants = require('../core/constants');
  * performance will likely drop as a result of too many calls to
  * beginShape() / endShape().  As a high performance alternative,
  * please use p5.js geometry primitives.
+ * @private
  * @method beginShape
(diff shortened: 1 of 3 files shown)


kjhollen:
some technical difficulties... hold on.

PR #2598, opened 2018-01-26
Improve construction of cylinder and cone primitives and geometry validation

### LINKED ISSUE #2594: p5.Geometry._getFaceNormal generates many warnings in console via p5.Vector.mult
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 
- Operating System: macOS Sierra
- Steps to reproduce this:

Run example in manual-test-samples/webgl/geometry. It produces hundreds of errors saying,

`p5.Vector.prototype.mult: n is undefined or not a finite number`

which seems to trace up to trying to take the angle between two vectors `[0, 0, 0]` and a divide-by-zero NaN in p5.Geometry._getFaceNormal. happens for `cylinder` and `cone` on initialization, doesn't repeat again after first frame. other shapes in this example are fine.

tau-:
This resolves the issue of spurious warnings in p5.Geometry._getFaceNormal as described in #2594. Closes #2594

### DIFF
--- a/src/webgl/p5.Geometry.js
+++ b/src/webgl/p5.Geometry.js
@@ -83,7 +83,9 @@ p5.Geometry.prototype._getFaceNormal = function(faceId) {
   var ln = p5.Vector.mag(n);
   var sinAlpha = ln / (p5.Vector.mag(ab) * p5.Vector.mag(ac));
   if (sinAlpha === 0 || isNaN(sinAlpha)) {
-    return new p5.Vector(0, 0, 0);
+    throw new Error(
+      'Degenerate geometry: This face either has a colinear sides or a repeated vertex.'
+    );
   }
   return n.mult(Math.asin(sinAlpha) / ln);
 };
--- a/src/webgl/p5.Geometry.js
+++ b/src/webgl/p5.Geometry.js
@@ -83,9 +83,11 @@ p5.Geometry.prototype._getFaceNormal = function(faceId) {
   var ln = p5.Vector.mag(n);
   var sinAlpha = ln / (p5.Vector.mag(ab) * p5.Vector.mag(ac));
   if (sinAlpha === 0 || isNaN(sinAlpha)) {
-    throw new Error(
-      'Degenerate geometry: This face either has colinear sides or a repeated vertex.'
+    console.warn(
+      'p5.Geometry.prototype._getFaceNormal:',
+      'face has colinear sides or a repeated vertex'
     );
+    return n;
   }
   return n.mult(Math.asin(sinAlpha) / ln);
 };
--- a/src/webgl/p5.Geometry.js
+++ b/src/webgl/p5.Geometry.js
@@ -82,6 +82,9 @@ p5.Geometry.prototype._getFaceNormal = function(faceId) {
   var n = p5.Vector.cross(ab, ac);
   var ln = p5.Vector.mag(n);
   var sinAlpha = ln / (p5.Vector.mag(ab) * p5.Vector.mag(ac));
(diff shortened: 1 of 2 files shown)


Spongman:
while this does avoid the division by zero, i think it might be worth taking a look to see why these degenerate faces are being created in the first place, and possibly preventing that. i think the issue is that some of the faces are created with two identical vertices (maybe different vertices in the same 3d location, or just duplicate vertex indices). it might be worth putting a check in the geometry function to help track this down...

tau-:
I agree, I will take a look at tracking down the degenerate faces. A sensible warning during geometry creation would be helpful. Worth keeping this check in or should the function rely on the correctness of the passed parameters?

Spongman:
i think since this is an internal library method it would probably be more useful to see the error rather than to hide it, since this error is really just a symptom of another error in the caller. also, adding a zero-length normal merely pushes the error down into the glsl shader's lighting code and onto the graphics card which doesn't have any mechanism for reporting such errors.

Spongman:
the reason this only occurs the 1st time through the draw loop is that the geometry is cached and the `computeNormals()` call is not repeated. if you were to resize the cone/cylinder every frame you would see this error each `draw()`.

tau-:
Yes, each time a new cone or cylinder is added through `cone()` or `cylinder()` or their geometry is adjusted a new vertex set is computed with `_truncatedCone()` and this repeats many vertices in the array. This looks to be so that the code for creating the faces as a triple of vertex indices can be simplified. I am looking at removing that issue from `_truncatedCone()` and adding a descriptive error, either in `computeNormals()` or when a geometry is added to the buffer in `_renderer.createBuffers()`.

tau-:
While the geometry issues are fixed, there is some adjustments needed to the vertexNormals.

tau-:
Vertex and face normals are fixed now.

Spongman:
this all looks good to me. i was wondering if it might make sense to do the drawScaledBuffers thing in this PR in order to simplify things even futher?

similar to some of the other primitives, the idea is to just create a single version of each geometry with unit sides/radius and just have the vertex shader scale everything appropriately. sometimes things get much simpler if all the lengths/angles are constant. just an idea...

tau-:
Currently `drawScaledBuffers` is used for the cylinder but not the cone. Does that function handle the normals on the side of the cone correctly? Maybe things could be simplified for cones as well, which would give some speed ups if there happen to be a large number of distinct cone geometries in the scene. I will do some testing and maybe make a little "forest" sketch to test the performance gains of the drawScaledBuffer approach.

Also, I am considering moving the error to a warning with a more detailed message. Since the rendering is still successful in the cases where zero normals were generated, providing a helpful message may be enough instead of killing the sketch.

Spongman:
yes, i believe the normals are handled correctly during the stretch for the geometries that do it currently. as for the cone, i'm not 100% certain, but i _think_ it should work. the trees will tell ;-)

tau-:
The normals are correct, but no performance improvements in the quick test I ran. Regardless I think it is a nice simplification.

kjhollen:
Cool, yes, I'll merge this once the geometry ID is updated! Thanks!

tau-:
Totally right! I will remove the unneeded values from the cone geometry ID.

kjhollen:
thanks! 🎆 & welcome!

PR #2599, opened 2018-01-26
fix shader tests

### LINKED ISSUE #2594: p5.Geometry._getFaceNormal generates many warnings in console via p5.Vector.mult

### LINKED ISSUE #2597: gl unit test failure: check activate and deactivating fill and stroke
#### Nature of issue?

- [x] Found a bug

- [x] WebGL

One of the WebGL unit tests doesn't pass. Here's what it says:

```
AssertionError: fill shader still active after noFill(): expected false to be true
    at Function.assert.isTrue (http://localhost:8000/node_modules/chai/chai.js:2483:31)
    at Context.<anonymous> (unit/webgl/stroke.js:50:14)
```

Spongman:
this one fixes 3 failures in webgl test that only fail when run in the browser (webgl tests dont run in PhantomJS).

closes #2597

the shader uniforms/attributes need to be removed from the checks because they're not actually used in the shader and on some browsers they are optimized away, causing the checks to fail. this is basically the same bug as #2262, but now that the code has checks to prevent crashe on those browsers, (#2349) the webgl tests are now able to run better.

the `_doFill` change is because previously the webgl renderer didn't affect the `_doFill` property at all (it used `curFillShader.active` and other state) so this property wasn't actually changed. now that both renderers use the same flag and it _is_ updates in the `fill()` call the bug in the test is exposed.

### DIFF
--- a/test/unit/webgl/p5.Shader.js
+++ b/test/unit/webgl/p5.Shader.js
@@ -128,12 +128,11 @@ suite('p5.Shader', function() {
       );
     });
     test('Color Shader definition', function() {
-      var expectedAttributes = ['aPosition', 'aNormal', 'aTexCoord'];
+      var expectedAttributes = ['aPosition'];
 
       var expectedUniforms = [
         'uModelViewMatrix',
         'uProjectionMatrix',
-        'uNormalMatrix',
         'uMaterialColor'
       ];
 
@@ -162,7 +161,7 @@ suite('p5.Shader', function() {
       );
     });
     test('Normal Shader definition', function() {
-      var expectedAttributes = ['aPosition', 'aNormal', 'aTexCoord'];
+      var expectedAttributes = ['aPosition', 'aNormal'];
 
       var expectedUniforms = [
         'uModelViewMatrix',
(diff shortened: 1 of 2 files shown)


kjhollen:
thanks @Spongman! next time, can you put the description you wrote into the commit itself, instead of the PR? then it can show up in `git log` and other git history tools. I copied it into the merge commit this time, but it's better if it's tied directly to the commit.

Spongman:
oh thanks. i'll try to remember the commit message in the future.

Issue #426, opened 2014-11-19
Elements created with createDiv contain text 'undefined' if no inner html is specified


lizstarin:
For instance, this js: 
`var sidebar = createDiv();  
sidebar.id("sidebar");`
produces this html:
`<div id="sidebar">undefined</div>`

austince:
Heyo, I think this is still a problem in 0.6.0. [Here's a Codepen example.](https://codepen.io/austince/pen/OQPqov/)

I'll put in a PR shortly

PR #2615, opened 2018-01-31
[Bug fix] createDiv (and other create[Node]) contain text 'undefined' if no innerHTML is specified

austince:
Addressing this issue https://github.com/processing/p5.js/issues/426 and creating a unit test for it.
[Demo](https://codepen.io/austince/pen/OQPqov/)
Let me know if the merge commits are a problem!


lmccart:
good catch!

PR #2620, opened 2018-01-31
move _edgedToVertices

Spongman:
this just moves _edgedToVertices from p5.Renderer (which it doesn't use at all) to p5.Geometry so it can be used as a fluent method like the other geometry methods.

geometry initialization which previously looked like:

    boxGeom._makeTriangleEdges();
    this._renderer._edgesToVertices(boxGeom);

now becomes

    boxGeom._makeTriangleEdges()._edgesToVertices();


kjhollen:
cool, thanks!

PR #2621, opened 2018-02-01
Lint manual tests

Spongman:
this adds the scripts in test/manual-test-examples to the eslint tasks in Gruntfile.js so they use the same styling and receive the same precommit validation as the library and doc samples.


Spongman:
ugh. cannot reopen after a rebase. thanks github :(

PR #2622, opened 2018-02-01
lint manual test examples

Spongman:
This adds the scripts in test/manual-test-examples to the eslint tasks in Gruntfile.js so they use the same styling and receive the same precommit validation as the library and doc samples.


lmccart:
does this add time to the run of grunt? the manual test examples are really meant to be scratch examples used for quick testing. they're not meant to serve as example code, and sometimes developers might want to put some slightly unorthodox code to test something particular, so i think it's ok to be laxer with the linting?

Spongman:
it does increase the precommit time a little. i think at the very least it's worth checking in the results of the prettification as it fixes some issues that were hilighted in some of the tests that were just plain broken - eg calling non-existent methods & syntax errors. (eg, https://github.com/processing/p5.js/pull/2622/files#diff-466d5205577fd00f94b4734e8b615adc)

how about we comment out the grunt rule that causes it to run during the tests?

lmccart:
@Spongman this makes sense to me, can you make the update?

lmccart:
👍

Issue #1430, opened 2016-05-20
windowWitdth


gregor:
Mapping the window.innerWidth value should be replaced by document.body.clientWidth as this will return the actual avaible space. innerWidth does not substract to the space needed for the scrollbar. Setting the width of your canvas to windowWidth therefore results in horizontal scrollbars.

limzykenneth:
I tend to agree, just that it should be `document.documentElement.clientWidth` instead of `document.body.clientWidth` and also ignoring the scroll bar means that if content on the page is generated dynamically and at some point the content move from no scroll bar to having a scroll bar, value of `document.documentElement.clientWidth` would change where `window.innerWidth` won't.

lmccart:
works for me!

Issue #1001, opened 2015-10-13
Auto Resize Canvas with Browser


zettam:
I've been trying to make this work:
http://p5js.org/reference/#p5/resizeCanvas

The resize works, owevr, scrollbars are always there (canvas is always slightly bigger)

What should I to make it fit?

lmccart:
many browsers have default padding or margin on a page, making the canvas overflow past the window if it's the size of the window.

you can add this line to the head of your html:
`<style> body{padding:0; margin:0;} </style>`
see more explanation in the p5.js overview tutorial: https://github.com/processing/p5.js/wiki/p5.js-overview#createcanvas

in general, this type of question is better for the forum, the issues are mostly for reporting bugs or discussing new features.
http://forum.processing.org/two/

PR #2624, opened 2018-02-01
hide scrollbars for full-window sketches

### LINKED ISSUE #2619: scrollbars with full-window canvas
demo: https://codepen.io/Spongman/pen/ZrGoQz

a simple way to do a full-screen canvas is adding 

in your setup, (eg, https://p5js.org/reference/#/p5/windowWidth, https://p5js.org/reference/#/p5/windowResized) and

    html, body { padding: 0; margin: 0; }

to your css styles.

however this doesn't work properly and results in unwanted scrollbars.

the problem is that the `canvas` element is inline, and every browser/OS combo i have tried adds scrollbars to the page even though the canvas element itself doesn't extend outside the window.

in the demo above, pressing the mouse button on the canvas changes the `display` css style of the canvas to `block`: thus removing the scrollbars.

some possible solutions:
- add `block` to the inline style for all createCanvas-created canvas elements. could be a breaking change for sketches containing non-canvas elements.
- update the docs to indicate that in order to remove the scrollbars you _need_ to add a css style to ensure those canvas elements are `block`. maybe we should add a `.p5canvas` class to those elements to make it easy to distinguish them from other user-created canvas elements on the page?

Spongman:
add p5jscanvas classes to canvas, set display to block if it fills the window to avoid showing scrollbars

closes #2619 
also #1430, #1001 (which aren't actually fixed)

### DIFF
--- a/src/core/rendering.js
+++ b/src/core/rendering.js
@@ -55,6 +55,8 @@ p5.prototype.createCanvas = function(w, h, renderer) {
   var r = renderer || constants.P2D;
   var c;
 
+  var fillsWindow = w >= this.windowWidth && h >= this.windowHeight;
+
   if (r === constants.WEBGL) {
     c = document.getElementById(defaultId);
     if (c) {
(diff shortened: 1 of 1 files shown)


RGFTheCoder:
just use 
````css
canvas {
    overflow: hidden;
}

PR #2626, opened 2018-02-02
Custom geometry

Spongman:
if you take this example: https://codepen.io/Spongman/pen/rJOoJa and try to run it on 0.6.0 it'll crash the browser window due to memory overflow and spending too long between draw calls.

these changes together fix the issues that are causing this:
- geometry arrays are not cleared before filling causing memory overflow
- `translate()` docs don't include vector version, causing false-positive FES errors
- `computeNormals()` is doing too much work, and can be easily simplified to a O(N) algorithm
- vector `normalize()` method does too many `sqrt()`s and other checks
- `_getFaceNormal()` sometimes gets a float overflow error during angle calculation causing NaN errors.

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -550,6 +550,8 @@ p5.Vector.prototype.dist = function dist(v) {
  */
 p5.Vector.prototype.normalize = function normalize() {
   var len = this.mag();
+  // here we multiply by the reciprocal instead of calling 'div()'
+  // since div duplicates this zero check.
   if (len !== 0) this.mult(1 / len);
   return this;
 };
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -549,7 +549,9 @@ p5.Vector.prototype.dist = function dist(v) {
  * </div>
  */
 p5.Vector.prototype.normalize = function normalize() {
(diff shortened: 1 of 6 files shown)


src/math/p5.Vector.js
```diff
@@ -549,7 +549,9 @@ p5.Vector.prototype.dist = function dist(v) {
  * </div>
  */
 p5.Vector.prototype.normalize = function normalize() {
-  return this.mag() === 0 ? this : this.div(this.mag());
+  var len = this.mag();
+  if (len !== 0) this.mult(1 / len);
```
kjhollen: great—worth a note here about the optimization, so that someone doesn't come along later and decide to replace this with div for easier reading :)

(also just noticing something weird adjacent here: `= function normalize ()`? should be just `= function ()`?)

Spongman: yeah, i seem to remember a while back someone went through and changed a bunch of methods from anonymous to named methods. something to do with not being able to see function names in the debugger.

limzykenneth: The discussion for named anonymous function is at #2318. Main reason is so that if that function throws an error in the console it will have a name and not just `(anonymous function)`

PR #2628, opened 2018-02-03
hide 'private' constructors in reference docs

### LINKED ISSUE #2601: reference docs contain syntax for constructors that should probably be hidden
in the reference docs there are pages for the constructors of classes in the library whose class page needs to exist (because it contains a description of the class and lists its members), but whose constructor probably shouldn't be advertised since there are other correct methods of instantiating objects of those classes, and the constructor definition exposes hidden implementation details of the library innards.

for example, the `p5.Font` page describes a constructor that takes a `pInst` parameter, however it's probably _never_ correct for a user of the library to call this constructor, they should instead be pointed to the `p5.prototype.loadFont()` function which hooks everything up correctly.

the following docs pages have the similar issues:
- https://p5js.org/reference/#/p5.Color (should use `color()` to ensure correct `colorMode` handling)
- https://p5js.org/reference/#/p5.Graphics (should use `createGraphics()`)
- https://p5js.org/reference/#/p5.TypedDict (shouldn't be directly constructed)
- https://p5js.org/reference/#/p5.NumberDict (should use `createNumberDict()`)
- https://p5js.org/reference/#/p5.Image (should use `createImage()`)
- https://p5js.org/reference/#/p5.PrintWriter (should use `createWriter()`)
- https://p5js.org/reference/#/p5.Vector (should use `createVector()` for correct `angleMode` handling)
- https://p5js.org/reference/#/p5.Font (should use `loadFont()`)
- https://p5js.org/reference/#/p5.Texture (maybe should be hidden entirely)
- https://p5js.org/reference/#/p5.Element (should use `createElement()` or other tag-specific methods)

------

my proposed solution involves removing the `@constructor` attribute from the constructor's doc comments, and adjusting the docs preprocessor and routing code to handle showing pages for public, constructorless classes correctly.

Spongman:
closes #2601

this removes various constructors from the reference docs that shouldn't be visible to the user.

### DIFF
--- a/docs/yuidoc-p5-theme-src/scripts/main.js
+++ b/docs/yuidoc-p5-theme-src/scripts/main.js
@@ -51,7 +51,7 @@ require([
 
     // Get classes
     _.each(classes, function(c, idx, array) {
-      if (c.is_constructor) {
+      if (!c.private) {
         App.classes.push(c);
       }
     });
--- a/docs/yuidoc-p5-theme-src/scripts/tpl/class.html
+++ b/docs/yuidoc-p5-theme-src/scripts/tpl/class.html
@@ -1,5 +1,5 @@
 
-<% if (is_constructor) { %>
+<% if (typeof constructor !== 'undefined') { %>
 <div class="constructor">
   <!--<h2>Constructor</h2>--> 
   <%=constructor%>
--- a/docs/yuidoc-p5-theme-src/scripts/tpl/item.html
+++ b/docs/yuidoc-p5-theme-src/scripts/tpl/item.html
@@ -57,6 +57,8 @@
 
 </div>
 
+<% if (isConstructor || !isClass) { %>
+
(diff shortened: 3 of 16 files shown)


lmccart:
is there a way to do this where we keep the `@constructor` and just check whether the entry is marked @private and ignore if so? it feels a little hacky to use `@ constructor` and likely to be misunderstood or confused in the future.

Spongman:
Actually the '@ constructor' is meaningless - it has no effect - I just left it in as an indication that this _used_ to be marked with a constructor attribute. They can safely be removed if you'd prefer.

Unfortunately we can't use @private for this as that would remove the class altogether, and there are other classes that rely on that behavior.

lmccart:
ah ok I see. yes let's remove altogether then to reduce confusion.

lmccart:
👍

PR #2632, opened 2018-02-05
Reference Added

faithwyu:
Added reference examples for p5Table getRowCount() and getColumnCount() functions.

### DIFF
--- a/src/io/p5.Table.js
+++ b/src/io/p5.Table.js
@@ -639,6 +639,35 @@ p5.Table.prototype.addColumn = function(title) {
  *
  *  @method  getColumnCount
  *  @return {Integer} Number of columns in this table
+ * @example
+ * <div>
+ * <code>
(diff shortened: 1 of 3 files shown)


src/io/p5.Table.js
```diff
@@ -639,6 +639,35 @@ p5.Table.prototype.addColumn = function(title) {
...
+ *   createCanvas(200, 100);
+ *   textAlign(CENTER);
+ *   background(255);
+ * }
+ *
+ * function draw() {
+ *   text(
+    'There are ' + table.getColumnCount() + ' columns in the table.',
+    100,
+    50
```
lmccart: this is a little hard to read, can we do it in one line like line 703?
if that's too long, I'd suggest pulling out the string like:
```javascript
var str = 'There are ' + table.getColumnCount() + ' columns in the table.';
text(str, 100, 50);
```

lmccart:
👍

Issue #2638, opened 2018-02-09
saveCanvas('myfile', 'png') downloads files as "myfile.myfile"


Ucodia:
`saveCanvas('myfile', 'png')` is supposed to download the canvas as a PNG file named `myfile.png` but instead downloads as `myfile.myfile`.

p5.js version: 0.6.0

#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [x] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

lmccart:
fixed with #2651

PR #2639, opened 2018-02-09
Fixed a bug in saveCanvas

Ucodia:
Fixed issue #2638 
- File was downloaded as "filename.filename"
- Fixed filename and extension assignment when canvas is absent from parameters

### DIFF
--- a/src/image/image.js
+++ b/src/image/image.js
@@ -148,8 +148,8 @@ p5.prototype.saveCanvas = function(cnv, filename, extension) {
   if (cnv instanceof p5.Element) {
     cnv = cnv.elt;
   } else if (!(cnv instanceof HTMLCanvasElement)) {
-    filename = cnv;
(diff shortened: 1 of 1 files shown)


lmccart:
thanks! this should fix it but I realize the mistake probably occurred in the first place from confusion about # of arguments passed in different cases. it still looks a little confusing from the naming. could we switch this around, instead of using named arguments, leave the parens empty and then in the if statements, check `arguments.length` and ` arguments[0] instanceof` to determine which case we're dealing with?

Ucodia:
Agreed, that would be much higher standard. Switching the arguments in the first place is probably what caused the confusion. Will send a revision.

lmccart:
hi @Ucodia I'm going to close this PR for organizational purposes, please feel free to reopen when you've made the update discussed, and I will get the notification to review. thank you!

PR #2644, opened 2018-02-13
Fix accidental italics in blendMode reference

cdfuller:
The asterisks in the formulas rendered all the text in between in italics.

image


lmccart:
thanks!

PR #2651, opened 2018-02-17
Fixed a bug in saveCanvas, simplified argument parsing

Ucodia:
Fixed issue #2638  
- File was downloaded as "filename.filename"
- Fixed filename and extension assignment when canvas is absent from parameters
- Simplified the saveCanvas function argument parsing


lmccart:
thanks!

PR #2657, opened 2018-02-20
Fix 3d stroke

Spongman:
this fixes the webgl stroke popping & disappearing issues seen here:

https://codepen.io/Spongman/pen/JpLdao?editors=0010

it moves the strokes a tiny distance toward the camera, and halves the default stroke width (since previously half the stroke was usually hidden behind the face).

pi

### DIFF
--- a/src/webgl/shaders/line.vert
+++ b/src/webgl/shaders/line.vert
@@ -23,7 +23,7 @@ uniform mat4 uProjectionMatrix;
 uniform float uStrokeWeight;
 
 uniform vec4 uViewport;
-vec3 scale = vec3(1.0);
+vec3 scale = vec3(0.9995);
 
 attribute vec4 aPosition;
 attribute vec4 aDirection;
--- a/src/webgl/shaders/line.vert
+++ b/src/webgl/shaders/line.vert
@@ -23,6 +23,10 @@ uniform mat4 uProjectionMatrix;
 uniform float uStrokeWeight;
 
 uniform vec4 uViewport;
+
+// using a scale <1 moves the lines towards the camera
+// in order to prevent popping effects due to half of
+// the line disappearing behind the geometry faces.
 vec3 scale = vec3(0.9995);
 
 attribute vec4 aPosition;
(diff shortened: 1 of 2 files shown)


src/webgl/shaders/line.vert
```diff
@@ -23,7 +23,7 @@ uniform mat4 uProjectionMatrix;
 uniform float uStrokeWeight;
 
 uniform vec4 uViewport;
-vec3 scale = vec3(1.0);
+vec3 scale = vec3(0.9995);
```
kjhollen: great! worth a comment here about what this is doing (moving closer to camera), but otherwise this looks good to me.

Spongman:
ah, yes, well spotted!

this was actually the previous behavior (in 0.5.16, demo here: https://codepen.io/Spongman/pen/YYEJjy?editors=0010). this was reported in #2105 & #2510, and 'fixed' in #2511 and #2515.

i don't think there's any way (without huge architectural changes, scene graphs, etc...) to fix both problems at the same time. i think fixing the edge popping issue is probably preferrable here since it occurs in almost all cases and there's no workaround, whereas the edge z-fighting issue only occurs in the co-planar case and the workaround is pretty simple.

FWIW: processing 3.3.6 also shows this behavior:
image

kjhollen:
just a note to say I'm thinking about this one! I'm not sure which mode to preference, here—need just a little more time to figure out what to prioritize, so we don't end up pushing bugs back and forth.

Spongman:
merging this would still fix the stroke popping bug. i believe the fix in #2510 was actually in error, since the processing shader actually also exhibits the overlapping stroke issue. the main reason for using it is that it fixes the stroke issue in #2488. i think reverting to the behavior that existed since 0.5.15 is fine.

mlarghydracept:
I agree with @Spongman that it makes sense to prioritize the stroke-popping. It is pretty common to have coplanar z-fighting issues with any 3D graphics framework and learning how to quickly remedy it is maybe a useful part of learning how to work with 3D graphics. I agree that pushing the bugs back in forth isn't ideal but hopefully this thread could help deter that from happening in the future?

kjhollen:
thanks!

fwiw, I encountered this issue in Processing for a project earlier this month and it was confusing. It doesn't really feel like a lesson in stacking and ordering objects in 3D graphics because it subverts the expectations of an end user. The fill ordering follows expectations (draw order) but the lines behave differently. It forces the end user to know that the lines are drawn in a separate render pass, at a different depth, and work around that. I think it's probably worth tracking this in another issue, in case someone has an idea for another render method that might fix it in the future!

Issue #2489, opened 2018-01-03
p5 Website Env.setADSR() issue


Siggi005:
#### Nature of issue?

- [x] Found a bug
- [x] Existing feature enhancement

#### Most appropriate sub-area of p5.js?

- [x] Website of p5

#### Which platform were you using when you encountered this?

- [x] Desktop/Laptop

#### Details about the bug: 
css styles on the website https://p5js.org/reference/#/p5.Env/set cause troubles: The example and the description are overlapping.

```
.example-content .edit_space,
.example-content pre {
    position: absolute !important;
    top: 0.5em !important;
    left: 120px !important;
    padding-top: 0 !important;
    border: none !important;
}
```

#### Feature enhancement details:
Delete `position: absolute !important;`

lmccart:
This issue was moved to processing/p5.js-website#155

PR #2660, opened 2018-02-22
Fix error on set() reference page

emclaren:
Resolves issue #2489, and the linked issue in processing/p5.js-website#155

`var l3 = dL` causes an Uncaught ReferenceError because dL is not defined. This causes the description text on the reference page to overlap with the example code: [https://p5js.org/reference/#/p5.Env/set.](https://p5js.org/reference/#/p5.Env/set)

The issue is fixed by replacing this value with a number.

### DIFF
--- a/lib/addons/p5.sound.js
+++ b/lib/addons/p5.sound.js
@@ -6048,7 +6048,7 @@ env = function () {
    *  var t2 = 0.3; // decay time in seconds
    *  var l2 = 0.1; // decay level  0.0 to 1.0
    *  var t3 = 0.2; // sustain time in seconds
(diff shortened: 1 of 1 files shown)


limzykenneth:
Hi @emclaren, can you make this change to the [p5-sound.js](https://github.com/processing/p5.js-sound) repo? The p5-sound.js file here is just the copy of the final build from that repo that we package up when there's a new release. Thanks.

Spongman:
this is already fixed in that repo.

emclaren:
Thanks for the heads up @limzykenneth and @Spongman - just started with contributing and still trying to figure out how all the pieces fit together.

PR #2665, opened 2018-02-24
Closes #2648 Unifies trigonometric angle conversion.

### LINKED ISSUE #2648: Precision problem with angle conversion on webkit
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [x] Math

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.6.0
- Web browser and version: Safari 11.0.3
- Operating System: macOS 10.13.3
- Steps to reproduce this:

This is a bit of a spaghetti of a bug but I'll try my best to show my findings. Sorry in advance for the long report.

First, the tests are failing on my computer, giving only `Fatal error: test.titlePath is not a function` as message, with no info about what that means exactly. I figured it wasn't a p5 related error so I searched online to find it's a problem with mocha and a possible problem is that some test is failing and mocha somehow isn't able to handle the thrown error (someone can look more into this).

That means some test is failing and I cannot tell which one so I switch the reporter from Nyan to spec which I managed to trace it to a test for `atan` (as much as I love Nyan, maybe it's not fit for purpose anymore 😢). Specifically the test for `atan` in degree mode. The test is testing for equality between `atan` in degree mode and `Math.atan` converted to degree with `degree()`. The log result from the test show that they are not equal, `atan` gives 26.565051177077994 and `Math.atan` gives 26.56505117707799

I tried to run these two in a standalone sketch and in Firefox and Chrome they are both equal to 26.56505117707799, while Safari gave the same result as above. I'm not sure what is common between Safari and node.js on my computer and different with Firefox and Chrome so I have no idea why the precision problem arise.

However, looking into the source I found where the problem stem from: `angleMode(DEGREES)` and `degrees()` uses different function to convert from radians to degrees. `angleMode` uses [`p5.prototype._fromRadians`](https://github.com/processing/p5.js/blob/master/src/math/trigonometry.js#L370) while `degrees` uses [polargeometry.js `radiansToDegrees`](https://github.com/processing/p5.js/blob/master/src/math/polargeometry.js#L8). One uses 180 and PI while the other uses 360 and 2*PI.

For a start we need to decide which one to use consistently as they are doing the same thing just [one uses Pi the other uses Tau](https://www.youtube.com/watch?v=ZPv1UV0rD8U). As for the mocha issue, we'll need to look for more info.

tau-:
Removes math/polargeometry.js and replaces its use with the DEG_TO_RAD and RAD_TO_DEG constants instead.


limzykenneth:
I've tested the branch, test now pass. LGTM.

kjhollen:
awesome, thanks!

PR #2669, opened 2018-02-27
Add initial support for other draw modes in beginShape().

### LINKED ISSUE #2662: beginShape(POINTS) not working in WebGL
It will display line instead of points in under WebGL renderer. Tested in both Safari and Chrome.

tau-:
Pass through shapeMode in _drawStrokeImmediateMode and update the immediate vertex and fragment shaders to support setting gl_PointSize as a uniform. Set new uniforms in RendererGL.strokeWeight. Stroke color and material shader can conflict (z-fight). The current line.vert shader attempts to render lines with thickness, but this can cause strange behavior if gl_PointSize is set so that is removed. Ideally this shader would nicely handle all point and line drawing options.

Definitely a WIP. An immediate mode vertex shader with a better implementation of the line options we want to support is ideal. For some examples see: https://mattdesl.svbtle.com/drawing-lines-is-hard

Fix #2662

### DIFF
--- a/src/webgl/p5.RendererGL.Immediate.js
+++ b/src/webgl/p5.RendererGL.Immediate.js
@@ -316,7 +316,11 @@ p5.RendererGL.prototype._drawStrokeImmediateMode = function() {
   }
 
   this._applyColorBlend(this.curStrokeColor);
-  gl.drawArrays(gl.TRIANGLES, 0, this.immediateMode.lineVertices.length);
+  gl.drawArrays(
+    this.immediateMode.shapeMode,
+    0,
+    this.immediateMode.lineVertices.length
+  );
 
   // todo / optimizations? leave bound until another shader is set?
   this.curStrokeShader.unbindShader();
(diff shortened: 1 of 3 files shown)


Spongman:
btw: check here for more discussion of thick lines, a z-fighting: #2488, #2657

tau-:
Thank you for the context. What is the state on accepting the changes to the rendering pipeline? It looks very much like the direction I was heading as I am understanding the p5 renderer more.

kjhollen:
hey @tau-, thank you for sharing this work in progress. I was testing this and it seems that stroke rendering no longer works at all, which means I can't merge this to the mainline—it breaks too prominent of a feature. What would you like to do as a next step? (Or were you just looking for some initial feedback?)

kjhollen:
hi @tau-, I'm going to close this for administrative purposes, but if you make further changes and would like a second review, feel free to re-open. thanks!

Issue #2676, opened 2018-03-01
clear() doesn't work if translate() is used


PhilipRL:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 0.6.0
- Web browser and version: Chrome 64.0.3282.167
- Operating System: MacOSX 10.13.3 (17D47)
- Steps to reproduce this:
```javascript
function setup() {
  createCanvas(400, 300);
  document.body.style.backgroundColor = '#fff'; // Make sure that the background is white
}

function draw() {
//push(); // Using push() & pop() actually fixes it
  translate(width/2, height/2);
  strokeWeight(8); // Just to be clear... ;)
  line(-width/2, -height/2, width/2, height/2); // Draw a line across the canvas
//pop();
  clear(); // Clear the canvas - or at least try
}
````

What happens in this example is that only one quarter of the screen is being cleared. This is because the canvas is translated while the `clear()` function always assumes that the top left corner is x=0, y=0 when in fact it should be the negative values (of the totals of all instances) of `translate()`. *(x = -200, y = -150 in this case)*

One workaround for this problem is to always use `push()` and `pop()` around `translate()`.

lmccart:
fixed with #2682

PR #2682, opened 2018-03-03
Modify #clear() to take translation into consideration when clearing …

mislavmilicevic:
…the canvas

Fixes issue #2676 by resetting the transformation matrix prior to clearing


lmccart:
thanks!

PR #2683, opened 2018-03-05
add error message for smooth function

danarmulder:


lmccart:
can you change this to console.log, so the program can continue to run even if smooth() is called?

PR #2689, opened 2018-03-08
webgl assertions

Spongman:
if you call method of `p5.prototype` that are webgl-only when you are not using the webgl renderer, you get undefined behavior - often crashes.

this PR adds assertions to all such webgl-only p5.prototype methods.

### DIFF
--- a/src/webgl/material.js
+++ b/src/webgl/material.js
@@ -29,7 +29,6 @@ require('./p5.Texture');
  * vertex and fragment shader files.
  */
 p5.prototype.loadShader = function(vertFilename, fragFilename) {
-  this._assert3d('loadShader');
   var loadedShader = new p5.Shader();
 
   var self = this;
--- a/src/webgl/material.js
+++ b/src/webgl/material.js
@@ -29,6 +29,7 @@ require('./p5.Texture');
  * vertex and fragment shader files.
  */
 p5.prototype.loadShader = function(vertFilename, fragFilename) {
+  this._assert3d();
   var loadedShader = new p5.Shader();
 
   var self = this;
@@ -116,6 +117,7 @@ p5.prototype.loadShader = function(vertFilename, fragFilename) {
  * zooming Mandelbrot set. a colorful, infinitely detailed fractal.
  */
 p5.prototype.createShader = function(vertSrc, fragSrc) {
+  this._assert3d();
   return new p5.Shader(this._renderer, vertSrc, fragSrc);
 };
 
@@ -131,6 +133,7 @@ p5.prototype.createShader = function(vertSrc, fragSrc) {
  * shapes.
  */
 p5.prototype.shader = function(s) {
+  this._assert3d();
   if (s._renderer === undefined) {
     s._renderer = this._renderer;
   }
@@ -168,6 +171,7 @@ p5.prototype.shader = function(s) {
  *
  */
 p5.prototype.normalMaterial = function() {
+  this._assert3d();
   this._renderer.drawMode = constants.FILL;
   this._renderer.setFillShader(this._renderer._getNormalShader());
   this._renderer.curFillColor = [1, 1, 1, 1];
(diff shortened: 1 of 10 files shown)


src/webgl/p5.RendererGL.js
```diff
@@ -1048,4 +1049,12 @@ p5.RendererGL.prototype._vToNArray = function(arr) {
   );
 };
 
+/**
+ * ensures that p5 is using a 3d renderer. throws an error if not.
+ */
+p5.prototype._assert3d = function() {
+  if (!this._renderer.isP3D)
+    throw 'Not supported in P2D mode. Please use webgl mode';
```
kjhollen: Maybe we can take this as an opportunity to update this message to offer new users a hint for how to fix the issue. Here are some suggestions, trying to build on what I see in the friendly error system:

* (method name) is only supported in WEBGL mode. If you'd like to use 3D graphics and Web GL, see https://p5js.org/examples/form-3d-primitives.html for more information.
* (method name) is not supported in 2D mode. Please create the canvas using WEBGL mode, see https://p5js.org/reference/#/p5/createCanvas for more information.

I think we could probably update the createCanvas docs to have a 3D example or we could have a more appropriate/simpler 3D sketch in the examples page, but that can wait until a future patch.

test/manual-test-examples/webgl/customGeometry/sketch.js
```diff
@@ -63,6 +66,6 @@ function draw() {
   pointLight(150, 150, 150, moonPos);
 
   geometry.computeFaces().computeNormals();
-  renderer.createBuffers("!", geometry);
-  renderer.drawBuffersScaled("!", 1000, 1000, 1);
+  renderer.createBuffers('!', geometry);
+  renderer.drawBuffersScaled('!', 1000, 1000, 1);
```
kjhollen: oops, forgot to comment on this initially: this is fine, but for future PRs please only update the test examples if they're relevant to the core of the patch.

Spongman: oh sorry, i don't know how that got in there!

src/webgl/material.js
```diff
@@ -29,6 +29,7 @@ require('./p5.Texture');
  * vertex and fragment shader files.
  */
 p5.prototype.loadShader = function(vertFilename, fragFilename) {
+  this._assert3d('loadShader');
```
kjhollen: Looking through the manual test examples, I got this when looking at the toon shader example...

```
Uncaught TypeError: Cannot read property 'isP3D' of undefined
    at p5._assert3d (p5.js:72953)
    at p5.loadShader (p5.js:69876)
    at p5.<anonymous> (p5.js:46111)
    at preload (sketch.js:4)
    at p5.<anonymous> (p5.js:46076)
    at new p5 (p5.js:46363)
    at _globalInit (p5.js:48341)
```

During preload, this._renderer isn't yet defined on the p5 instance. If it's not obvious how to fix this, my suggestion is to remove this line from the commit so that custom shaders continue to work with preload and we can merge this patch promptly. Then we can open a new issue to resolve this for loadShader as a unique case.

test/unit/core/renderer.js
```diff
@@ -89,4 +89,27 @@ suite('Renderer', function() {
...
+      expect(function() {
+        myp5.rotateX(100);
+      }).to.throw(Error);
+    });
+    test('normalMaterial() should throw an Error', function() {
+      expect(function() {
+        myp5.normalMaterial(100);
+      }).to.throw(Error);
+    });
+  });
```
kjhollen: Great! Could all of the methods expected to throw an error be added here?

src/webgl/material.js
Spongman: you're right. this assertion is completely invalid. i don't think it's every necessary here.

kjhollen:
awesome, thanks!

PR #2692, opened 2018-03-12
Add example for p5.Table.getArray()

seagalputra:
Hi, all
This is my first contribution in here.
Recently, i read an issue here #1954. So, i add a simple code for method getArray()
Thank you

### DIFF
--- a/src/io/p5.Table.js
+++ b/src/io/p5.Table.js
@@ -1213,6 +1213,7 @@ p5.Table.prototype.getObject = function(headerColumn) {
  *
  * @method  getArray
  * @return {Array}
+ *
  * @example
  * <div class="no render">
  * <code>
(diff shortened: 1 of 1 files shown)


src/io/p5.Table.js
```diff
@@ -1213,6 +1213,32 @@ p5.Table.prototype.getObject = function(headerColumn) {
  *
  * @method  getArray
  * @return {Array}
+ * @example
+ * <div class="no render">
```
limzykenneth: There shouldn't be a space between "no" and "render" otherwise they are actually two classes `.no` and `.render` instead of one class `.norender`.

src/io/p5.Table.js
```diff
@@ -1107,7 +1107,7 @@ p5.Table.prototype.getNum = function(row, column) {
  * @return {String}
  *
  * @example
- * <div class="norender">
+ * <div>
```
limzykenneth: If the `norender` class is not present, the sketch will still create a canvas but it would be empty in this particular case since nothing will be drawn on screen. Can be a bit confusing I think since most people won't have the console opened all the time.

limzykenneth:
Thanks! Left a couple of comments, do have a look.

seagalputra:
@limzykenneth thanks for your correction. I will fix it as soon as possible

lmccart:
thanks @seagalputra and @limzykenneth!

Issue #2674, opened 2018-02-28
Implementation of loadBytes


shiffman:
I am opening a new issue since #40 about file i/o is a long thread and this relates specifically to `loadBytes()`. I am working on an example that loads binary data and thought I would take a crack at adding `loadBytes()` to p5. I have something working and am happy to pull request an implementation, however, I have a couple questions I'd love thoughts or advice on:

The working new method is towards the bottom.

1. As discussed in [Example of async function for callback and preload()](https://github.com/processing/p5.js/wiki/Libraries#example-of-async-function-for-callback-and-preload) on the libraries wiki, it's required that an original object pointer is retained. I've done this by placing the `Uint8Array` into a `values` property of a blank object. Thoughts on how this could be improved? Can I set the values of a `Uint8Array` if I don't know the size in advance?

2. I think I am running into #2568 as I can only get this to work in 0.5.11 (no later versions).

```javascript
p5.prototype.registerPreloadMethod('loadBytes');

function loadBytes(file, callback) {
  let data = {};
  let oReq = new XMLHttpRequest();
  oReq.open("GET", file, true);
  oReq.responseType = "arraybuffer";
  oReq.onload = function(oEvent) {
    let arrayBuffer = oReq.response;
    if (arrayBuffer) {
      data.values = new Uint8Array(arrayBuffer);
      if (callback) {
        callback(data);
      }
    }
  }
  oReq.send(null);
  return data;
}
```

Spongman:
you'll only run into #2568 if you're using p5 in instance mode. probably the reason you're not setting `preload()` complete is you're not calling `_decrementPreload()` in your `loadBytes` callback.

i was working on something similar a while back using an augmented version of `httpDo()` that handles `Blob`s and `ArrayBuffer`s:

````
    p5.prototype.loadBytes(file, callback, errorCallback) {
      let data = {};

      var self = this;
      this.httpDo(
        file,
        'GET',
        'arrayBuffer',
        function(arrayBuffer) {
          data.bytes = new Uint8Array(arrayBuffer);

          if (typeof callback !== 'undefined') {
            callback(data);
          }

          self._decrementPreload();
        },
        errorCallback
      );
      return data;
    };
````

note the call to `self._decrementPreload();` at the end of the callback.

Dhakshith:
@Spongman
It Is Working, But Error:
`Uncaught TypeError: Failed to construct 'Request': Request with GET/HEAD method cannot have body.`

[...]

PR #2693, opened 2018-03-12
Implementing loadBytes() #2674

shiffman:
Dear diary, 

This commit adds the `loadBytes()` function to p5.js as demonstrated during a livestream on The Coding Train. There are some remaining issues here. For example there are no tests. Also, example code is missing. Finally, I am also not using `httpDo()` which would be more of the convention for p5 as discussed in #2674. I am also handling errors like 404 or CORS in perhaps an odd way and would be happy to have any feedback.

A thanks to @Spongman for this branch which was a nice reference for me: https://github.com/processing/p5.js/compare/master...Spongman:loadBytes

It's ok if this doesn't get merged, hopefully doing this on a livestream was helpful to somebody!

Sincerely yours,
The Coding Train

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -684,6 +684,68 @@ p5.prototype.loadXML = function() {
 };
 
 /**
+ * Loads a binary file from a file or a URL, and returns an Object
+ * with a bytes property containing a Uint8Array.
+ *
+ * This method is asynchronous, meaning it may not finish before the next
+ * line in your sketch is executed.
+ *
+ * @method loadBytes
+ * @param  {String}        path       name of the file or url to load
+ * @param  {function}      [callback] function to be executed after
+ *                                    loadBytes() completes, data is passed
+ *                                    in as first argument
+ * @param  {function}      [errorCallback] function to be executed if
+ *                                    there is an error, response is passed
+ *                                    in as first argument
+ * @return {Object}             contains a Uint8Array bytes property
+ * @example
+ *
+ * <p>Calling loadBytes() inside preload() guarantees to complete the
+ * operation before setup() and draw() are called.</p>
+ *
+ * <div><code>
+ * // TODO: Add Example with preload
+ * </code></div>
+ *
+ * <p>Outside of preload(), you may supply a callback function to handle the
+ * data:</p>
+ * <div><code>
+ * // TODO: Add Example with callback
+ * </code></div>
+ *
+ * @alt
+ * 50x50 ellipse that changes from black to white depending on the current humidity
+ * 50x50 ellipse that changes from black to white depending on the current humidity
+ *
+ */
+p5.prototype.loadBytes = function(path, callback, errorCallback) {
+  var self = this;
+  var ret = {};
+  var oReq = new XMLHttpRequest();
+  oReq.open('GET', path, true);
+  oReq.responseType = 'arraybuffer';
+  oReq.onload = function(oEvent) {
+    if (oReq.status >= 400) {
+      return;
+    }
+    var arrayBuffer = oReq.response;
+    if (arrayBuffer) {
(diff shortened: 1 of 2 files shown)


src/io/files.js
```diff
@@ -683,6 +683,68 @@ p5.prototype.loadXML = function() {
...
+ *
+ */
+p5.prototype.loadBytes = function(path, callback, errorCallback) {
+  var self = this;
+  var ret = {};
+  var oReq = new XMLHttpRequest();
+  oReq.open('GET', path, true);
+  oReq.responseType = 'arraybuffer';
+  oReq.onload = function(oEvent) {
+    if (oReq.status >= 400) {
```
meiamsome: In httpDo, it checks on fetch's `res.ok` which is only valid if the status code is 2xx. So `status >= 200 && status <300` would be the same behaviour as the rest of p5 but I'm not sure what the feeling for this is.

Link2Twenty: Incase people come from the video and don't understand why `status >= 200 && status < 300` makes sense.

Here is some further reading:
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/status

src/io/files.js
```diff
@@ -683,6 +683,68 @@ p5.prototype.loadXML = function() {
...
+ * </code></div>
+ *
+ * <p>Outside of preload(), you may supply a callback function to handle the
+ * data:</p>
+ * <div><code>
+ * // TODO: Add Example with callback
+ * </code></div>
+ *
+ * @alt
+ * 50x50 ellipse that changes from black to white depending on the current humidity
```
shiffman: A note that this needs to be fixed as its leftover from the `loadJSON()` example.

limzykenneth: Probably need to be fixed, not sure when it was added but I think someone is currently working on adding alt text to canvases. I would say maybe edit it so it matches the example sketch or remove it for now.

src/io/files.js
limzykenneth: A couple quick comments before I test the code locally: it would probably be best to check for `status >= 200 && status < 300` only as `>=300` will not return a response body and so it won't populate the intended variable. The client is supposed to take care of `>=300` but I'm guessing without knowledge of the endpoint we can't handle it properly.

As @meiamsome mentioned, if response is not ok an error should be thrown so that it won't fail silently, I'm not sure if in this case simply throwing the error will make the error be passed onto `oReq.onerror` otherwise the user won't really be able to handle the error themselves.

Spongman:
I was looking at the sound library recently and I noticed there were a few places here it was loading array buffers. I wonder if there's a way for it to reuse some if this? The httpGet addition I posted above allows for loading of arraybuffers and Blobs, which might prove more reusable in the future.

shiffman:
@tresjasper The code still needs a few improvements based on some comments. This pull request is also somewhat speculative for a tutorial about contributing. 

@Spongman if you want to take a crack at your idea please go ahead. 

This should be ready to merge once I fix the error handling, but I'm very glad for it to be replaced by a better solution.

limzykenneth:
I think I would prefer to merge @Spongman's version since that is more consistent and clearer, there's also some improvements on `httpDo` itself which I'm eager to test out. However, @shiffman you can still fix the current code for tutorials and if you managed to fix the error handling we can maybe merge this before merging @Spongman's version since it will still need to be checked out as well.

@Spongman if you can make a PR for your version I'll have a closer look and test some time this week.

shiffman:
Thanks! This sounds like a good plan. I'll try to do as you suggest @limzykenneth but if @Spongman is faster than me no need to wait. This can be closed.

limzykenneth:
Just merged #2771 with @Spongman's implementation

shiffman:
Yay, thanks for your work @limzykenneth and @Spongman! Closing!

PR #2695, opened 2018-03-13
Add function to get hex value of color

TanviKumar:
- Add function hexstring to get hex representation of a color
	- Add example for this function
	- Add unit test for same


lmccart:
thanks @TanviKumar we already have a toString() method which does this:
https://github.com/processing/p5.js/blob/master/src/color/p5.Color.js#L56

in general, new features should first be discussed as an issue before a PR is submitted, to avoid unneccessary work

TanviKumar:
I missed that, will do for the future! Thanks!

PR #2697, opened 2018-03-13
Preload bugs

### LINKED ISSUE #2568: registerPreloadMethod mechanism is broken
0.6.0: https://p5js.org/reference/#/p5/loadModel

the sketch never starts because `_incrementPreload()` is called twice, once for the wrapped `loadModel` call, and once more for `loadStrings`. and `_decrementPreload()` is only called once.

it has something to do with the reference docs site, because if you copy that sketch to its own .html file it works fine - the `loadStrings` doesn't result in `_incrementPreload()` getting called.

meiamsome:
Attaches preload functions to the global instance as well as to window when creating a global object. Fixes `loadModel` preload. Fixes #2568. Incompatible with #2696

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -256,7 +256,10 @@ var p5 = function(sketch, node, sync) {
         var obj = this._preloadMethods[method];
         //it's p5, check if it's global or instance
         if (obj === p5.prototype || obj === p5) {
-          obj = this._isGlobal ? window : this;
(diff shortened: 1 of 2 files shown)


Spongman:
it looks like this was already merged into master as part of #2700 (which I rebased on top of your changes here).

lmccart:
thanks! yes this was merged with @Spongman's rebased PR.

PR #2700, opened 2018-03-14
Test (almost) all examples

Spongman:
this PR adds ~200 more existing reference examples to the mocha test run that were previously excluded.

- make `text(void 0)` not crash.
- add a `notest` class to examples that shouldn't be tested, and change associated filter in test-reference.html.
- fixes some XML methods to use `childNodes` instead of the non-standard `children`, not supported in PhantomJS.
- removes 1 dependency on `[].find`, also not supported in PhantomJS.
- fix `XML._setAttributes` to handle undefined attributes.
- fix bogus reference to `window.millis` in PrintWriter

most of the remaining examples are impossible to test. for example, they invoke the save dialog, interact with expected elements on the page, or media controls.

(it is rebased on @meiamsome's fix to the preload stuff, i'll rebase again onto master once that is merged.)


lmccart:
thanks! can you add a note about `notest` here: https://github.com/processing/p5.js/wiki/Inline-documentation#adding-example-code?

PR #2717, opened 2018-03-21
fix loadShader preload, add example

Spongman:
this PR fixes the `loadShader()` function's preload callback refcounting and adds an example for that method.

### DIFF
--- a/docs/yuidoc-p5-theme/assets/swirl.frag
+++ b/docs/yuidoc-p5-theme/assets/swirl.frag
@@ -1,0 +1,18 @@
+precision highp float;
+varying vec2 vPos;
+uniform float time;
+
+float r;
+
+float Q(float a, float b, float c, float d, float e) {
+  float v = 0.5 + 0.5 * sin(time * a + atan(vPos.y, vPos.x) * b + sin(log(r * c) * d * sin(time * e)));
+  return v*v*v;
+}
+void main() {
+  r = length(vPos);
+
+  gl_FragColor.a = 1.0;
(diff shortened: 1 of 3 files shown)


Spongman:
oh yeah, @meiamsome fixed an issue with the preload stuff in #2697. the result of this fix is that the preload count actually works as expected now wrt. nested calls, so `loadShader()` only needs to call `_decrementPreload()` once when it's done.

meiamsome:
Whoops, sorry that this change broke this here - I'm working on adding more tests for the preload functions because the current ones don't check it works correctly. Hopefully nothing else got broken that hasn't been flagged up!

kjhollen:
Awesome, got it!

For the example—could we go with something simpler as the primary example? Maybe something that just translates mouseX and mouseY to one of the shader uniforms, so the reader doesn't have to also parse all of the trig & rotation in the example?

Spongman:
actually, the trig doesn't appear in the sample - that's all hidden away in the shader code which the user can't actually see.

image

lmccart:
the loadShader() fix has been mered in #2821. @kjhollen if this example is not appropriate let's close this, or if you think it's ok, we can merge.

lmccart:
another possibility is to put this example on the examples page

Spongman:
I'm really confused as to why this was rejected, but #2881 was merged. without a related issue, no less. :confused:

kjhollen:
I've just been swamped; I'm sorry I missed this (and the other example). I would still like something simpler as the first example in the docs—where maybe just the mouse coordinate is used to control a color. the mandelbrot set is also quite complex.

PR #2724, opened 2018-03-23
Running remove prior to setup removes the sketch.

### LINKED ISSUE #2723: remove() does not function correctly prior to setup() running
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [x] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug: 

The `remove()` function in p5 does not work correctly prior to the execution of `setup()` - it leaves the sketch there and doesn't do anything to it. This is particularly an issue when the instance is stuck because of a failed load it cannot be removed simply:

```javascript
function preload() {
  loadStrings('nosuchfile');
  setTimeout(remove, 1000);
}
```

meiamsome:
- Fixes #2723 
- Correctly removes globals in this case
- Also removes `p5.instance` from referencing any removed sketch

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -427,6 +427,12 @@ var p5 = function(sketch, node, sync) {
    *
    */
   this.remove = function() {
+    var loadingScreen = document.getElementById(this._loadingScreenId);
+    if (loadingScreen) {
+      loadingScreen.parentNode.removeChild(loadingScreen);
+      // Add 1 to preload counter to prevent the sketch ever executing setup()
+      this._incrementPreload();
+    }
     if (this._curElement) {
       // stop draw
       this._loop = false;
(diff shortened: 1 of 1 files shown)


src/core/core.js
```diff
@@ -457,28 +463,27 @@ var p5 = function(sketch, node, sync) {
...
-            try {
-              delete window[p2];
-            } catch (x) {
-              window[p2] = undefined;
-            }
+            window[p2] = undefined;
           }
         }
       }
+      p5.instance = null;
```
lmccart: can you explain this line here, I'm not sure I understand what this does?

meiamsome: This removes the `p5.instance` reference to sketch from hanging around in global scope after removal. I guess it isn't necessary addition but it lets the GC cleanup more than otherwise.

lmccart: ah got it, i missed the place where it was assigned. makes sense.

PR #2731, opened 2018-03-25
Added some provision for testing Release script out

sakshamsaxena:
Now you can simply edit up a few config vars with your own creds to point the releases at them instead of processing. 

In tasks/release/release-p5.js
> Kept URLs as config vars, so that anyone can change them to add their own, to test if release works or not.
> Uncomment L#55 to set dry run as true for testing


lmccart:
this is great, thanks!

sakshamsaxena:
@lmccart Done! [Here](https://github.com/processing/p5.js/wiki/Release-steps)

PR #2737, opened 2018-03-26
Updated reference bezier,curve,quadrativeVertex

AdilRabbani:
Updated reference examples for bezierVertex(), curveVertex() and quadraticVertex(). Making them more self-explanatory.


lmccart:
very helpful, thanks!

Issue #2732, opened 2018-03-25
Using weights when selecting elements from array in random()


BenjaminHabert:
The `random()` function can be used to select an element from an array. It could be useful to provide weights so that some elements from the array are returned more often than others.

I would be interested in implementing this feature.

#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### New feature details:

```javascript
// Existing behaviour
var animal = random(["cat", "dog", "platypus"]);

// Proposed behaviour
var weights = [2, 1, 0.1];  // platypus is less liekely than other animals
var animal = random(["cat", "dog", "platypus"], weights);
```

The second array parameter `weights` would be optional and have the following constraints:

- same length as the first array parameter
- contains numbers
- (not required but might speed up the computation: sum of elements is equal to 1. In this case the array would be called `probabilities`.)

#### Current Status

I went ahead and implemented it in a fork. Following the instructions from the Development guide, here is the current status:

- [x] Implementing feature
- [x] Inline doc for website
- [x] unit tests
- [ ] benchmark. Not really sure how to proceed here.

Also: I am not sure if some checks should be made inside the function itself to help the user (eg: check that arrays are the same length, check that weights are non-negative, check that the sum is greater than 0).

I created a pull request, sorry if it was too soon.

lmccart:
I think this function is beyond the scope of the p5.js api (we try to keep as minimal as possible) but it could be a great starting point for an addon library. see the docs here for how to create an addon: https://github.com/processing/p5.js/wiki/Libraries#creating-a-new-library

BenjaminHabert:
I understand the necessity to keep the API simple. However I believe that the proposed feature is a common use case (I tend to use this a lot). 

- pick a particle direction / speed out of a given probability distribution
- pick a random pixel in an image where darker pixels are more likely
- pick a random word in a dictionary where words with many 'a's are more likely
- pick random colors out of a finite palette while making sure that some colors are used more than others (useful to have a base color + a highlight color)

But obviously I am a bit biased on this topic :)

[...]

PR #2739, opened 2018-03-26
Using weights when picking elements in random

BenjaminHabert:
The random() function can now take two array parameters.
choice = random(choices, weights). The second parameter
is treated as weights (same length as choices, contains
numbers). Elements with higher weights are more likely
to be returned by the function.

Closes: #2732

### DIFF
--- a/src/math/random.js
+++ b/src/math/random.js
@@ -87,11 +87,17 @@ p5.prototype.randomSeed = function(seed) {
  * If one argument is given and it is a number, returns a random number from 0
  * up to (but not including) the number.
  *
+ * If two number arguments are given, returns a random number from the
+ * first argument up to (but not including) the second argument.
+ *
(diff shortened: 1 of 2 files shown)


src/math/random.js
```diff
@@ -144,22 +160,37 @@ p5.prototype.random = function(min, max) {
   } else {
     rand = Math.random();
   }
+
   if (typeof min === 'undefined') {
     return rand;
-  } else if (typeof max === 'undefined') {
-    if (min instanceof Array) {
-      return min[Math.floor(rand * min.length)];
+  } else if (min instanceof Array) {
```
limzykenneth: Can we change this check to use `Array.isArray`, `instanceof` checks may cause problems in rare cases.

BenjaminHabert: Sure; changed it.

PR #2742, opened 2018-03-27
Add default class to p5.js created canvases

limzykenneth:
As discussed in #2619 this PR adds a `p5Canvas` class to all p5.js created canvases. Do check to make sure I got nothing wrong.

### DIFF
--- a/src/core/rendering.js
+++ b/src/core/rendering.js
@@ -12,6 +12,7 @@ require('./p5.Graphics');
 require('./p5.Renderer2D');
 require('../webgl/p5.RendererGL');
 var defaultId = 'defaultCanvas0'; // this gets set again in createCanvas
+var defaultClass = 'p5Canvas';
 
 /**
  * Creates a canvas element in the document, and sets the dimensions of it
@@ -67,6 +68,7 @@ p5.prototype.createCanvas = function(w, h, renderer) {
     }
     c = document.createElement('canvas');
     c.id = defaultId;
+    c.classList.add(defaultClass);
   } else {
     if (!this._defaultGraphicsCreated) {
       c = document.createElement('canvas');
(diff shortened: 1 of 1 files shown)


src/core/rendering.js
```diff
@@ -12,6 +12,7 @@ require('./p5.Graphics');
 require('./p5.Renderer2D');
 require('../webgl/p5.RendererGL');
 var defaultId = 'defaultCanvas0'; // this gets set again in createCanvas
+var defaultClass = 'p5Canvas';
```
BenjaminHabert: maybe `'p5-canvas'` would be me idiomatic for a css class. Plus this avoid any [confusion](https://stackoverflow.com/questions/12533926/are-class-names-in-css-selectors-case-sensitive)
 about case sensitive / insensitive.

BenjaminHabert:
A canvas is also created in [p5.Graphics.js](https://github.com/processing/p5.js/blob/master/src/core/p5.Graphics.js#L29)
and in [p5.Render2D.js](https://github.com/processing/p5.js/blob/master/src/core/p5.Renderer2D.js#L136). 

Maybe they do not need to get a class as they are not the main canvas, I dont't know.

src/core/rendering.js
limzykenneth: Yeah, I personally prefer `p5-canvas` as well, but as the default ID uses camel case I just thought maybe keep the casing consistent here. What do you think?

limzykenneth:
I think if they are not displayed on screen and not meant to be manipulated by CSS it should be fine to not include the class names on those.

src/core/rendering.js
BenjaminHabert: I am not super confident about this.

I think it is better to align with the standard used elsewhere (ex [leaflet](https://www.phpclasses.org/browse/file/142068.html), [bootstrap](https://www.w3schools.com/bootstrap/bootstrap_ref_all_classes.asp), ...)

However it seems like the default ID is already [being used quite a lot](https://github.com/search?q=defaultCanvas0&type=Code&utf8=%E2%9C%93)  as a css selector. In that case you are probably right about keeping it consistent.

Issue #2077, opened 2017-07-18
Issues with Retina-screen, createGraphics and pixeldensity


rockfarmor:
Hello! I've been looking around the forum for similar issues, and found some that partly handles my problem. It's known that when using createGraphics() on a screen with pixeldensity > 1, the graphicsObject becomes distorted. A known fix for this is to manually set the pixeldensity of the graphicsObject to 1. But now to the problem -> When trying to render text to the object, the text-quality is becomes really low, and when rendering the text with the graphicsObject's pixeldensity=2 the text becomes distorted and to large.

#### Nature of issue?

- [x] Found a bug
- [x] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [x] Color
- [x] Core
- [ ] Events
- [x] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [x] Macbook with retina-screen
- [ ] Others (specify if possible)

#### Details about the bug: 
- p5.js version: 0.5.11
- Web browser and it's version: Chrome
- Operating System: Mac OSX
- Steps to reproduce this: Rendering text to a graphics-object with pixelDensity 1, on a screen with pixelDensity 2

Here are two printscreens displaying the problem. The text "Numbers:" is rendered to the GraphicsObject with very low resolution, whilst the text "windowtitle" is high-res and rendered to the canvas directly.

GraphcisObject with density 1 (On screen with density2)
<img width="368" alt="2" src="https://user-images.githubusercontent.com/10855270/28295164-2af21268-6b5f-11e7-9de1-3db7a6c4faae.png">

GraphcisObject with density 2 (On screen with density2)
<img width="358" alt="1" src="https://user-images.githubusercontent.com/10855270/28295160-25eea72c-6b5f-11e7-81ab-eaf5f234065e.png">

Tldr; Is there any work-around to render high-res text to the p5.Graphics object on high-pixelDensity screens

limzykenneth:
May not be helpful but is it possible to use DOM element instead? By using DOM element to render text, the browser and in turn the OS will take care of text on different pixel density display so you don't need to worry about this and let the canvas do graphics drawing instead of text rendering.

quinton-ashley:
[...]

Issue #2064, opened 2017-07-14
width and height of p5.Graphics object not adjusted when resized


mjvDEV:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [x] Core
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

As best I can tell, when `resizeCanvas(w,h);` is called on a graphics buffer created with `createGraphics(w,h);` the width and height of the p5.Graphics object are not adjusted, only the lower level width and height values. This causes the graphics to be displayed at the wrong size.

The code I've provided below demonstrates this. After 5 seconds, the graphics buffer is supposed to be resized to half the size, but instead displays at one quarter of its original size. 5 seconds later I adjust the width and height of the p5.Graphics object to the correct values, and the graphics buffer is then displayed at the correct size.

Apologies if this has already been brought up.

- p5.js version: /*! p5.js v0.5.11 June 01, 2017 */
- Web browser and it's version: Chromium Version 59.0.3071.109 (Developer Build) Built on Ubuntu , running on Ubuntu 16.04 (64-bit)
- Operating System: Ubuntu 16.04 LTS
- Steps to reproduce this: 

```
var graphics; 
var gWidth = 400;
var gHeight = 200;

function setup() {
    createCanvas(600,400);
    graphics = createGraphics(gWidth, gHeight);
    graphics.background(0,255,0);

    // Resize to half size in 5 seconds.
    setTimeout( function() {
        gWidth /= 2;
        gHeight /= 2;
        graphics.resizeCanvas(gWidth, gHeight);
        graphics.background(0,255,0);

        // 5 seconds later, correct the width and height.
        setTimeout( function() {
            graphics.width = gWidth;
            graphics.height = gHeight;
        }, 5000 );
    }, 5000 );
}

function draw() {
    background(155);
    image(graphics, 100, 100, gWidth, gHeight);
}
```

limzykenneth:
[...]

PR #2746, opened 2018-03-28
Fix scaling issue when using Graphics on retina

BenjaminHabert:
Using `image(p5.Graphics)` on a screen with `pixelDensity()`
higher than 1 results in distorted results.

This commit fixes this issue by using the technique implemented
in [`_copyHelper()`](https://github.com/processing/p5.js/blob/master/src/core/p5.Renderer2D.js#L218). I am however not sure if this commit breaks anything
else as  the rendering tests are sparse and I am only discovering
the codebase.

I do not know how to add unit tests for this part of the code; I didn't write any.

Closes: #2077

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -110,12 +110,16 @@ p5.Renderer2D.prototype.image = function(
     if (!cnv) {
       cnv = img.canvas || img.elt;
     }
+    var s = 1;
+    if (img.width && img.width > 0) {
+      s = cnv.width / img.width;
+    }
     this.drawingContext.drawImage(
       cnv,
-      sx,
-      sy,
-      sWidth,
-      sHeight,
+      s * sx,
+      s * sy,
+      s * sWidth,
+      s * sHeight,
       dx,
       dy,
       dWidth,
--- a/src/core/rendering.js
+++ b/src/core/rendering.js
@@ -152,6 +152,8 @@ p5.prototype.resizeCanvas = function(w, h, noRedraw) {
       }
     }
     this._renderer.resize(w, h);
+    this.width = w;
+    this.height = h;
     // reset canvas properties
     for (var savedKey in props) {
       try {
--- a/test/unit/core/graphics.js
+++ b/test/unit/core/graphics.js
@@ -120,7 +120,6 @@ suite('Graphics', function() {
       var graph = myp5.createGraphics(10, 17);
       graph.pixelDensity(1);
       graph.resizeCanvas(19, 16);
-      // this fails:
       assertValidGraphSizes(graph, 19, 16, 1);
     });
 
@@ -142,7 +141,6 @@ suite('Graphics', function() {
       var graph = myp5.createGraphics(10, 17);
       graph.pixelDensity(2);
       graph.resizeCanvas(19, 16);
-      // this fails:
       assertValidGraphSizes(graph, 19, 16, 2);
     });
 
--- a/test/unit/core/graphics.js
+++ b/test/unit/core/graphics.js
@@ -1,0 +1,163 @@
+suite('Graphics', function() {
+  var myp5;
+
+  setup(function(done) {
+    new p5(function(p) {
+      p.setup = function() {
+        myp5 = p;
+        done();
+      };
+    });
+  });
+
+  teardown(function() {
+    myp5.remove();
+  });
+
+  function assertValidGraphSizes(graph, w, h, density) {
+    assert.strictEqual(graph.pixelDensity(), density, 'invalid pixel density');
+    assert.strictEqual(graph.width, w, 'invalid width');
+    assert.strictEqual(graph.height, h, 'invalid height');
+  }
+
+  function assertValidCanvasSizes(canvas, w, h, density) {
+    assert.strictEqual(canvas.width, w * density, 'canvas.width');
+    assert.strictEqual(canvas.height, h * density, 'canvas.height');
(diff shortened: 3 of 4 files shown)


Spongman:
i wonder if this shouldn't be using the same pixelDensity state instead of relying on the size of the canvas element?

BenjaminHabert:
doing something like this ?

```javascript
var s = 1;
if (img.hasOwnProperty('pixelDensity') {
  s = img.pixelDensity();
}
```

I have no strong opinion either way but I feel that the proposed way is more robust: someone might change the size of the `img.canvas` at some point ?

Spongman:
Yeah that's what I was thinking. If the canvas changes size, are subsequent drawing operations also going to be scaled?

BenjaminHabert:
Nicely spotted, there is indeed an issue with resizing!

However I am a bit confused in my tests. What would be the standard way of resizing a graphics object ? I find that `graph.resizeCanvas()` changes the buffer size but does not update the exposed properties `graph.width` and `graph.height` (related to #2064 ?). Furthermore if `graph.resizeCanvas()` is called inside a `draw()` loop this raises a stackOverflowErrror. 

It appears that the only consistent way of resizing a graphics is to do (outside of draw):

```javascript
graph.resizeCanvas(200, 200);
graph.width = 200;
graph.height = 200;
```

Am I missing something ?

BenjaminHabert:
After further investigation here are my conclusions:

- resizing a graphics may indeed cause some problems
- both of the proposed implementations solve the problem (either scaling using `canvas.width / img.width` as in `p5.Render2D.copy()` or scaling using `pixelDensity()` as suggested above)
- what actually makes it work is to resize the canvas using the following (**in that order!** as `resizeCanvas()` triggers a redraw where the properties `width` and `height` will be used):


```javascript
graph.width = 200;
graph.height = 200;
graph.resizeCanvas(graph.width, graph.height);
```

- using `.size()` (from `p5.Element`) on the graphics seems to mess up everything: it scales the main canvas (see [these lines](https://github.com/processing/p5.js/blob/master/lib/addons/p5.dom.js#L1765) in `p5.dom.js`)


Here is what I suggest:

- keep proposed fix in the pull request (`canvas.width / img.width`) to be consistent with the implementation of `.copy()`
- decide on a way to improve the resizing of graphics. Here are some options:
  1. disable the function `resizeCanvas()` for a p5.Graphics (not sure how to distinguish it from the main canvas)
  2. fix it to make sure it updates the `width` and `height` properties
  3. improve `.size()` of `p5.Element` so that  1: the hidden graphics doesn't become visible as mentioned in #2064 and 2: it doesn't resize / rescale the main canvas. This probably needs to be another issue.

Thoughts ?

Spongman:
ok this is great! another thing that might be worth considering (i'm not sure if it's an issue with/without your fix or if it's even possible to fix) is whether or not the post-resized width/height line up with what's returned from `loadPixels()` and whether or not `get(x,y)` with contiguous integer values uniquely addresses all the pixels in the image.

Spongman:
another thing which is loosely related: i believe currently that a browser scaling factor of 101% results in a `pixelDensity` of 2. i think this is probably overkill and it may make sense to round the pixel density in a different way.

maybe something like `pow(2, round(log(scaling)/log(2)))` ?

BenjaminHabert:
I will try to write tests that ensure that resizing and `load_pixels()` (and maybe `get()` ?) work properly for a createGraphics() object. 

I will disregard your second point though. As I am not touching `pixelDensity` at the moment I think this is out of scope

BenjaminHabert:
I have implemented a test that reproduce the resizing problem we were talking about. I am not sure how to implement a fix though; I could use some help.

```javascript
// in core.rendering.js
p5.prototype.resizeCanvas = function(w, h, noRedraw) {
  p5._validateParameters('resizeCanvas', arguments);
  if (this._renderer) {
    // save canvas properties
    // ...

    this._renderer.resize(w, h);
    // The previous call does indeed update the size of the inner canvas 
    // and the width and height properties of the inner _renderer.
    // however those are not the actual properties exposed by the Graphics object.
    // the following would solve it but this is probably not the right place for it: 
    this.width = w;
    this.height = h;

    // reset canvas properties
   //  ...
  }
};
```

BenjaminHabert:
By the way here are the sizes of the different object as expected by the tests I implemented, let me know if this is what is actually intended:

- Graphics has size w, h (*this is the only thing that doesn't work*)
- graphics.canvas has size (w * density), (h * density)
- graphics.canvas.style properties are "wpx", "hpx"
- number of pixels is w * density * h * density * 4

limzykenneth:
`Fatal error: test.titlePath is not a function` is a grunt mocha problem, not sure if they pushed to npm yet, I had that issue last time as well. https://github.com/disqus/grunt-mocha/pull/7

kjhollen:
hi @BenjaminHabert, checking in to see what the remaining work items are here. I've run the tests and they seem to be working great, thank you! it makes sense to me that both copy and this new version of image would use pixelDensity, as an additional change.

is the resizing issue that changing the size of the p5Graphics causes a conflict of pixel densities? can you elaborate, and was this fixed by your most recent commit—and if not, is there a code sample that reproduces the issue? I'm not sure I'm completely following on the part about resizing. Once I understand, i'll decide if we can merge this as-is or if we should make additional fixes first. thank you!!

kjhollen:
on second thought, I went ahead and merged this as-is because it also alleviates #2838, which seems like good progress. thanks!! can you open a new issue for any remaining resizing issues?

Issue #2709, opened 2018-03-19
remove array methods?


lmccart:
At a recent contributor meetup, the question was raised whether we really needed the array methods (append, concat, splice, etc), since they are such close aliases of [existing JS array methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array). There is the added complication with native JS methods of requiring the dot operator to use them, but by the time someone is using arrays, is it safe to assume they would be able to comprehend the dot? Is there another reason for keeping them in?

cc @shiffman

shiffman:
I am in favor of removing them! I think we could consider leaving our own "documentation" in the p5.js reference for the most common methods like `push()`, `splice()`, . . .

This is similar to what we did with `String` in Processing.

https://processing.org/reference/String.html
https://processing.org/reference/String_toLowerCase_.html

The array methods made sense in the Processing API since users would otherwise have to learn about ArrayLists, but here since the methods are simply duplicates of native JS I don't think it makes sense for us to re-implement them. When it's time for one of these functions, it's a good time for a user to learn about how an array is an object.

limzykenneth:
If we decided to remove our implementation in p5.js, I would prefer as @shiffman suggest, to leave the "documentation" (albeit edit to fit native implementation) behind as even I'm also sometimes confused as to which of the array functions mutates the array while which of them return a new array.

meiamsome:
Would this be a deprecation initially? Perhaps add an error to them from FES when they are first used in sketches.

Spongman:
i vote we keep the legacy method for backwards compatibility, but just mark them with `@deprecated` attributes in their docs and have those comments rendered prominently in the reference docs.

shiffman:
Would it make sense to do as @Spongman suggests but when there is a major release (`0.7.0`? `1.0.0`?) remove them permanently?

PaintedBlck:
@shiffman I think it makes perfect sense, but only when changing version. Of course, the choice between changing version, release or level becomes very important at that moment.

I'm in favour of the idea of @Spongman since backward compatibility is very important, especially for beginners and non-programmers/artists/... who don't like to redo the programming they already did.

Below, I'm using the terminology `version.release.level` I'm in favour of the following:

- A new `version` is like a new library or program - no supposed backward compatibility
- A new `release` comes with substantial new functionality and backward compatibility.
- A new `level` comes with a substantial number of bug fixes and backward compatibility.

[...]

PR #2755, opened 2018-04-02
Deprecated docs

Spongman:
as discussed in #2709, adds a section to the reference docs for the `@deprecated` attribute.

### DIFF
--- a/docs/yuidoc-p5-theme-src/scripts/tpl/item.html
+++ b/docs/yuidoc-p5-theme-src/scripts/tpl/item.html
@@ -13,6 +13,18 @@
 
 
 <div class="description">
+
+  <% if (item.deprecated) { %>
+    <p>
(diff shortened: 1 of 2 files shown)


docs/yuidoc-p5-theme-src/scripts/tpl/item.html
```diff
@@ -75,9 +87,11 @@ <h4><span id="reference-syntax">Syntax</span></h4>
 </span>
 <% } %>
 
+<!--
 <% if (item.deprecated) { %>
 <span class="flag deprecated"<% if (item.deprecationMessage) { %> title="<%=item.deprecationMessage%>"<% } %>>deprecated</span>
 <% } %>
+-->
```
lmccart: can we delete this block completely?

Issue #2766, opened 2018-04-06
WebGL clear fills with black pixels instead of transparent background


CrazyInfin8:
#### Nature of issue?

- [x] Found a bug

#### Most appropriate sub-area of p5.js?

- [X] WebGL

#### Which platform were you using when you encountered this?

- [X] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.6.0
- Web browser and version: Opera 52.0.2871.40
- Operating System: MacOSX
- Steps to reproduce this:

When clearing canvas' set to WEBGL mode, the canvas' pixels are replaced with black pixels instead of transparent ones

Following this example:
```javascript
let pg;
function setup() {
  createCanvas(500, 500);
  background(0, 255, 255)
  pg = createGraphics(250, 500, WEBGL)
  pg.clear();
  image(pg, 0, 0);
}
```
![image](https://user-images.githubusercontent.com/14926524/38447423-038b8156-39b2-11e8-9c2c-648d2c0e251e.png)
The background of the main canvas is cyan as set previously however the color of the WebGL graphics is set to black instead of trasparent thus the cyan background is hidden.

There is a simple workaround to clear the canvas by calling background on WebGL and setting it to transparent
```javascript
let pg;
function setup() {
  createCanvas(500, 500);
  background(0, 255, 255)
  pg = createGraphics(250, 500, WEBGL)
  pg.background(0, 0);
  image(pg, 0, 0);
}
```
![image](https://user-images.githubusercontent.com/14926524/38447479-611a3556-39b2-11e8-8263-66bbda3e258a.png)

This is my first time writing bug reports on GitHub so go easy on me ☺
I don't know if it's better to write an issue and then commit or if it is okay to just commit

but I found the issue in `p5.js/src/webgl/p5.RendererGL.js` function `p5.RendererGL.prototype.clear`
where the arguments passed to the native function `clearColor` of the canvas' `WebGLRenderingContext` is  null.
I attempted to fix the issue by setting default values of 0

again still my first time submitting issues ^u^

Qoutroy:
That's a pretty good first issue!! 👍👍

CrazyInfin8:
@Qoutroy heh… thanks!

lmccart:
fixed with https://github.com/processing/p5.js/pull/2767

mkmori:
This issue is back in p5.js 0.8.0--also, the workaround _background(0,0)_ no longer works.

PR #2767, opened 2018-04-06
Fix WebGL clear to make background transparent instead of black

CrazyInfin8:
Fixes the `clear` function for canvas' running on `WEBGL` rendering mode where pixels would turn black instead of transparent #2766

Also following a tutorial by The Coding Train on contributing to open-source projects.
This being my first commit to another project so go easy on me ☺.
(Thank you @shiffman for the tutorial)


lmccart:
thanks!

mkmori:
This issue is back in p5.js 0.8.0--also, the workaround _background(0,0)_ no longer works....

PR #2771, opened 2018-04-09
add arrayBuffer/Blob support to httpDo. implement loadBytes

### LINKED ISSUE #2674: Implementation of loadBytes

Spongman:
re: #2693, closes #2674.

this PR adds `arrayBuffer` and `Blob` support to the `httpDo` method, and adds a convenience method `loadBytes` for loading a `Uint8Array` specifically.

adds some tests for `loadBytes`, including one for ensuring that the error callback is called when a 404 occurs.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -593,6 +593,7 @@ p5.prototype._preloadMethods = {
   loadImage: p5.prototype,
   loadStrings: p5.prototype,
   loadXML: p5.prototype,
+  loadBytes: p5.prototype,
   loadShape: p5.prototype,
   loadTable: p5.prototype,
   loadFont: p5.prototype,
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -1065,13 +1065,8 @@ p5.prototype.httpDo = function() {
           return res.text();
       }
     })
-    .then(callback || defaultCallback)
-    .catch(errorCallback || defaultErrorCallback);
-
-  function defaultCallback() {}
-  function defaultErrorCallback(err) {
-    throw err;
-  }
+    .then(callback || function() {})
+    .catch(errorCallback || console.error);
 };
 
 /**
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -684,13 +684,66 @@ p5.prototype.loadXML = function() {
 };
 
 /**
+ * @method loadBytes
+ * @param {string}   file            name of the file or URL to load
+ * @param {function} [callback]      function to be executed after loadBytes()
+ *                                    completes
+ * @param {function} [errorCallback] function to be executed if there
+ *                                    is an error
+ * @returns {Object} an object whose 'bytes' property will be the loaded buffer
+ *
+ * @example
+ * <div class='norender'><code>
+ * var data;
+ *
+ * function preload() {
+ *   data = loadBytes('assets/mammals.xml');
+ * }
+ *
+ * function setup() {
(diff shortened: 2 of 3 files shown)


Spongman:
ugh. this test only fails on travis. i can't tell why...

Spongman:
there seems to be an issue with the mocha test framework where an Error thrown in an inner Promise causes the test to fail rather than be caught in the outer Promise. disabling this 404 test for now...

limzykenneth:
Ok, I've tested all that I can think of and here are a few points:

Not sure why I didn't caught this before (maybe because I switched to Firefox recently) but if an error callback isn't provided, simply throwing the returned error will in Firefox cause it to fail silently while in Chrome and Safari cause it to throw an "Unhandled Promise Rejection" error. For some context of why see this question https://stackoverflow.com/questions/30715367/why-can-i-not-throw-inside-a-promise-catch-handler#30741722 The suggested method in the question using `setTimeout` gave a very unhelpful message so I wouldn't suggest that. I think maybe substituting `throw err` with `console.error(err)` could work but if you have a better solution do go for it.

---

Next is about the mocha problem which I think I figured out. It seems that webkit haven't implement [`finally` for Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally) yet which is why it failed in PhantomJS and not most browsers. If the test for that can be rewritten it would be great. Normally I use [sinon.js](http://sinonjs.org/) to test that but it may not be worth to bring the whole thing in just for this test. You can decide how to proceed.

---

The rest may not be best for here as they are more general feature request, I can move it to a separate issue once this is merge.

- It would be nice to be able to specify type for the `httpDo` overload that accept `Request` object otherwise anything we cannot auto detect its type will be parsed as text.

- We probably need a new example for `httpDo` documentation that demonstrate its use without using the `Request` object as that may be a bit too high level for most not familiar with these standard.

Once the first two points are checked off this can be merged.

Spongman:
Fantastic!  I'll switch to ’console.error’ and workaround the missing finally.

limzykenneth:
Excellent, thanks!

PR #2772, opened 2018-04-09
updated preload() reference documentation

faithwyu:
updated preload() reference documentation to contain explanation of its blocking nature and usage.
let me know if there is more that you want to add - i think the original already contained a lot of useful information already.


lmccart:
this is great! my one suggestion is to add a sentence:

"If asynchronous loading is preferred, the load methods can instead be called in setup() or anywhere else with the use of a callback parameter."

faithwyu:
@lmccart ahh I see what you meant by updating the reference now. working on it!

lmccart:
super thanks!

PR #2775, opened 2018-04-12
patch grunt-mocha to fix 'test.titlePath' errors on test failure

Spongman:
re: #2648, #2746, https://github.com/disqus/grunt-mocha/issues/6, https://github.com/disqus/grunt-mocha/pull/7, 

this PR patches the grunt-mocha package with fixes so that test failures will now display the actual failure information instead of some internal `test.titlePath is not a function` error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,8 @@
     "lint-staged": "^4.3.0",
     "marked": "^0.3.9",
     "opentype.js": "^0.7.3",
+    "patch-package": "^5.1.1",
+    "postinstall-prepare": "^1.0.1",
     "prettier": "^1.7.4",
     "request": "^2.81.0",
     "whatwg-fetch": "^2.0.3",
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "lint:fix": "grunt lint-fix",
     "add": "all-contributors add",
     "generate": "all-contributors generate",
+    "prepare": "patch-package",
     "precommit": "lint-staged"
   },
   "lint-staged": {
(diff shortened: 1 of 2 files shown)


limzykenneth:
Seems like a good solution until we can maybe migrate to something that is either better maintained or something that doesn't rely on plugins so much.

Spongman:
indeed. somewhere i have a branch that uses headless-chrome for the test runner, but i had a bunch of issues with is not detecting properly when the process had ended.

meiamsome:
Alternate option: use the git url in the package.json (Preferably pinned to the particular commit) until this is published

Spongman:
That is possible. I'd have to create a fork and another branch since there's also the growl fix that isn't part of that grunt-mocha PR.

meiamsome:
@Spongman can you explain what the growl fix achieves? Surely the previous code goes through the `catch` section and sets it to an empty function anyway?

meiamsome:
Hmm, that's a bit odd considering `growl` isn't in the deps. Do you have it on global install? Could be a version mismatch if so.
I'd never even consider that there'd be such issues when not including an optional dependency...

Spongman:
i get this:
```
> yarn why growl
yarn why v1.5.1
[1/4] Why do we have the module "growl"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "growl@1.10.3"
info Reasons this module exists
   - "grunt-mocha#mocha" depends on it
   - Hoisted from "grunt-mocha#mocha#growl"
```

Spongman:
any chance we can get this merged so we can actually see build errors in travis?

lmccart:
thank you. as a reminder, issues need to be opened before pull requests are opened and tagged with the issue. this is necessary for tracking development and keeping discussion clear. thanks!

Spongman:
am I missing something here? I listed two issues above that this fix addresses.

Spongman:
Here's another: #2812

lmccart:
please open an issue that documents the specific problem you're solving with this pr. if it's solving a related point that comes up in a completely separate issue thread it is difficult to track. thanks.

Spongman:
closes #2820

Issue #2740, opened 2018-03-26
FES misses error cases


Spongman:
p5 0.6.0

these are some more errors that the current FES system fails to catch.

FES doesn't currently issue _any_ warning about any of these errors:
```javascript
color(0,0,0,'A'); // wrong type for optional parameter
color([0,0,0],0); // superfluous parameter 
color(['A','B','C']); // wrong element types
rect(0,0,0,0,null,0,0,0); // null, non-trailing, optional parameter
```

and it warns about a type error in this case:
```javascript
color('A','A',0,0,0,0,0,0,0,0);
```
where an argument count error would probably be more appropriate.

himanshuc3:
Hi. I want to start contributing to p5js. Would you be kind enough to tell me things i need to learn. 
P.S. I know javascript.

limzykenneth:
Hi @himanshuc3, your question is a bit unrelated to this issue but do see [this page](https://github.com/processing/p5.js/wiki/Development) to get started contributing to p5.js, you can also checkout issues marked with "good first issue" to start with. If you have further question do open another issue. Thanks!

ankur54:
Hi. Is this issue still open? I would like to contribute to this issue.

Spongman:
I think this was fixed a while back in the FES ‘scoring’ PR.

lmccart:
thanks @Spongman. I've tested and confirmed this is fixed in 0.8.0, all four cases mentioned are now caught.

PR #2780, opened 2018-04-13
parameter validation tests part 1

Spongman:
part 1 of #2592


Spongman:
closes #2740

src/core/error_helpers.js
```diff
@@ -159,26 +160,53 @@ if (typeof IS_MINIFIED !== 'undefined') {
...
+    var queryResult;
+    var classitems = arrDoc.classitems;
+    for (var ici = 0; ici < classitems.length; ici++) {
+      if (classitems[ici].name === func) {
+        queryResult = classitems[ici];
+        break;
+      }
+    }
+    // bail if we couldn't find it (should never happen)
+    if (!queryResult) throw new Error('missing docs for method ' + func);
```
lmccart: I don't think this is the appropriate place to throw this error, since it's not something the user can be reasonably expected to fix. If there are no docs found, let's give validateParams() a pass somehow. If we want to check for documentation of a function we can do that elsewhere, but that's a separate issue.

src/core/error_helpers.js
```diff
@@ -268,68 +296,138 @@ if (typeof IS_MINIFIED !== 'undefined') {
   };
 
   var testParamType = function(param, type) {
+    function isNumber() {
```
lmccart: does this need to be defined inside the testParamType definition?

src/webgl/light.js
```diff
@@ -162,7 +162,12 @@ p5.prototype.directionalLight = function(v1, v2, v3, x, y, z) {
   var shader = this._renderer._useLightShader();
 
   //@TODO: check parameters number
-  var color = this.color.apply(this, [v1, v2, v3]);
+  var color;
```
lmccart: not sure how these changes relate to above. please separate into separate PR and tag with related issue.

tasks/typescript/generate-typescript-annotations.js
```diff
@@ -55,7 +55,7 @@ function mod(yuidocs, localFileame, globalFilename, sourcePath) {
     String: 'string',
     Constant: 'any',
     undefined: 'undefined',
-    Null: 'null',
+    null: 'null',
```
lmccart: not sure how these changes relate to above. please separate into separate PR and tag with related issue.

src/core/error_helpers.js
```diff
@@ -268,68 +296,138 @@ if (typeof IS_MINIFIED !== 'undefined') {
...
+    }
+
     var errorArray = [];
     for (var p = 0; p < formats.length; p++) {
       var arg = args[p];
       var format = formats[p];
-      var argType = typeof arg;
-      if ('undefined' === argType || null === arg) {
-        if (format.optional !== true) {
+      if (arg == null) {
```
lmccart: is this supposed to be three equal signs or two? i think do a == test like this can be easily misread or misinterpreted , we could just check `if(!arg)` instead to be more claer

src/core/error_helpers.js
```diff
@@ -268,68 +296,138 @@ if (typeof IS_MINIFIED !== 'undefined') {
...
+        // handle non-optional and non-trailing undefined args
+        if (
+          !format.optional ||
+          (p < overload.minParams ? true : p < argCount)
```
lmccart: this line is very difficult to parse, can you write it more verbosely to add clarity?

src/core/error_helpers.js
Spongman: ok, i just put this in here to debug issues with missing docs. removing it won't change anything.

src/core/error_helpers.js
Spongman: yes, the `== null` is intentional and returns a different result than just the falsey check (which would be wrong here).

(sorry if this is duplicate, i wrote this before, but it looks like i deleted it by mistake)

lmccart: can you give me an example of a case where it returns a different result?

Spongman: 0, false, ''

src/webgl/light.js
Spongman: this change is necessary because the previous code is invalid when a `color` is passed to `directionalLight`/`pointLight` as the 1st parameter.

for example, if `c` is a `p5.Color` instance and you call `directionalLight(c, x, y, z)`, then the old code instantiates a color with (effectively) `color([c, x, y])` expecting `c,x,y` to be r,g,b values - but they're not.

this newer validation code catches this case and displays a warning, but during the unit test this causes a test failure. so, this change is necessary in order to get the tests to pass:

image

kjhollen:
I would like to merge the optimizations done here, but feel that the goals for the scoring algorithm are not clear, so it's hard to evaluate this patch. @Spongman would you be willing to move the optimizations into another PR? And to open an issue to discuss the scoring algorithm and its goals in service of presenting the user with direct, concise, beginner-friendly messages? Thanks!

Spongman:
i'm not quite sure what you mean by separating it out. the scoring algorithm _is_ how the optimization works.

i think it's pretty easy to understand. it's all pretty clearly commented. if yo have any questions about it i'd be happy to answer them.

kjhollen:
What is the expected behavior for each of the examples in the issue report for #2740? How do these new error messages better support beginner programmers using p5?

Spongman:
each of the (first 4) examples in #2740 are cases where the parameters passed do not match any of the documented overloads. there are many cases where FES warns the user that they're passing the wrong types/number of parameters, but these cases are not currently covered correctly. the last case is one where currently FES returns a type error which is correct, but an argument count error would probably be more helpful to the user since there is no `color` overload that takes 10 arguments.

i think i addressed all the questions. there was the typescirpt bit that snuck in there by mistake, the complex expression that i simplified. the `== null` is intentional, and the parameter handling in light.js _is_ actually wrong.

kjhollen:
Can you please list (explicitly) the expected error message for each case in #2740? Without that, I am looking to the code to determine the expected results, which is not ideal for a user-facing feature. There should be a clear feature specification, so that validating the specification can be part of the reviewer's job. Please clearly state the expected outcome and spell out each error message.

Please also make the changes Lauren requested, which I think are all reasonable—removing extraneous bug fixes helps keep the intent of the PR clear. Thanks!

Spongman:
oh sure. this is what i get for the first four errors:
```
> p5.js says: color() was expecting Number for parameter #3 (zero-based index), received string instead at pen.js:21:3. [http://p5js.org/reference/#p5/color]
> p5.js says: color() was expecting Number for parameter #0 (zero-based index), received array instead at pen.js:22:3. [http://p5js.org/reference/#p5/color]
> p5.js says: color() was expecting Number[] for parameter #0 (zero-based index), received array instead at pen.js:23:3. [http://p5js.org/reference/#p5/color]
> p5.js says: rect() was expecting Number for parameter #4 (zero-based index), received an empty variable instead. If not intentional, this is often a problem with scope: [https://p5js.org/examples/data-variable-scope.html] at pen.js:24:3. [http://p5js.org/reference/#p5/rect]
```

and the last one:
```
> p5.js says: color() was expecting no more than 4 arguments, but received 10 at pen.js:26:1. [http://p5js.org/reference/#p5/color]
```


please see my comment here: https://github.com/processing/p5.js/pull/2780#discussion_r184862436 for why the fix in light.js is necessary.

PR #2784, opened 2018-04-16
Replicating PR #188 from p5.js-website repo - add reference topic to page title

awarua:
Use javascript to update the document title on reference pages to reflect the current topic being browsed. 

image

This was originally submitted as a PR to the p5.js-website repo [PR 188](https://github.com/processing/p5.js-website/pull/188), but I realise that it would be better to submit it here, because these are the ultimate source of the reference page files.

The original pull request has more info, as does the original issue from the p5.js-website repo [issue 38](https://github.com/processing/p5.js-website/issues/38)

### DIFF
--- a/docs/yuidoc-p5-theme-src/scripts/views/itemView.js
+++ b/docs/yuidoc-p5-theme-src/scripts/views/itemView.js
@@ -106,6 +106,15 @@ define([
 
         renderCode();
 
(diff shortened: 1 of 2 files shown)


lmccart:
this is great, so happy to have this in there at last! thanks @awarua

PR #2788, opened 2018-04-16
Remove hanging references to p5.Graphics

### LINKED ISSUE #2785: createGraphics() objects not being garbage collected.
#### Nature of issue?

- [x] Found a bug ?

- [x] Memory usage

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: v0.6.0
- Web browser and version: 

Google Chrome : 65.0.3325.181 (Official Build) (64-bit)(cohort: Stable)
Mozilla Firefox : version 58.0.2

- Operating System: Windows 8 version 6.3
- Steps to reproduce this:

While trying to create a drawing app I found that the memory usage was eventually making the browser crash. After some troubleshooting I found that the objects created with _**createGraphics()**_ were not being eliminated from memory even after calling _**remove()**_ and setting them to _**null**_. The memory usage reaches 1GB after a few clicks while running the script I wrote to illustrate the issue which, by the way, sounds very similar to #1785. The browser takes longer to crash when the size of the objects created is smaller, but it still does.

```javascript
function setup(){
    c = createCanvas(windowWidth, windowHeight);
    c.mouseClicked(mo);
}

function mo(){
    ppp = createGraphics(5000, 5000); //oversized to make memory usage obvious
    ppp.background(color(random(255)));
    image(ppp, 0, 0);
    ppp.remove();
    ppp = null;
}
```

Given that I'm not well acquainted with JS, it's very likely that it's not a bug rather than an error on my part; but being that the processing.org forum is not allowing people to register I decided to try and get help here.

meiamsome:
Fixes #2785

### DIFF
--- a/src/core/p5.Graphics.js
+++ b/src/core/p5.Graphics.js
@@ -59,6 +59,7 @@ p5.Graphics = function(w, h, renderer, pInst) {
   this._renderer._applyDefaults();
 
   this.name = 'p5.Graphics'; // for friendly debugger system
+  this._pInst = pInst;
   return this;
 };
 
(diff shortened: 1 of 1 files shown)


src/core/p5.Graphics.js
```diff
@@ -59,6 +59,7 @@ p5.Graphics = function(w, h, renderer, pInst) {
   this._renderer._applyDefaults();
 
   this.name = 'p5.Graphics'; // for friendly debugger system
+  this._pInst = pInst;
```
JulianCRA: Isn't this attribute inherited from p5.Element? https://github.com/processing/p5.js/blob/970f6c14c6410205d10b82636b513e7eefd55cc2/src/core/p5.Element.js#L45

PR #2793, opened 2018-04-18
Two new examples for loadPixels

AdilRabbani:
Added two new examples for loadPixels(). One generates a simple texture using pixels and the other averages image pixels.

### DIFF
--- a/src/image/pixels.js
+++ b/src/image/pixels.js
@@ -550,12 +550,10 @@ p5.prototype.get = function(x, y, w, h) {
  *   for (var y = 0; y < height; y++) {
  *     for (var x = 0; x < width; x++) {
  *       var index = (x + y * width) * 4;
- *
- *       pixels[index] = (pixels[index] + pixels[index + 1] + pixels[index + 2]) / 3;
- *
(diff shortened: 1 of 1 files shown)


limzykenneth:
In #1954 the `loadPixels()` function that needed example is for the version of `loadPixels()` for the [WebGL renderer](https://github.com/processing/p5.js/blob/master/src/webgl/p5.RendererGL.js#L614) (currently not showing up in the reference because it is marked as a private function, probably still work in progress). I feel we don't need extra example in this instance of `loadPixes()` as it doesn't demonstrate additional functionality of the method that isn't already covered by the original example.

P.S. There seems to be a lot of noise in in the commits that are just adding and removing whitespace.

limzykenneth:
You can run `grunt test` locally to run the same tests that are ran on Travis and you can also run `npm run lint:fix` to try to automatically fix any code style issues.

Spongman:
I think that actually the #1954 issue is in error. The renderer methods are not exposed as part of the API, and _shouldn't_ appear in the reference docs.

The correct list of methods missing examples can be found by running `grunt yui doc:prod`.

limzykenneth:
Yeah, I'm thinking maybe they are outdated since we've only added the `@private` tag recently

PR #2801, opened 2018-04-21
Changed createCapture() documentation and example

acasmith:
The docs for createCapture() referenced the non-existent method video(). Corrected the description and updated the example.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1046,8 +1046,10 @@
   }
 
   /**
-   * <p>Creates a new &lt;video&gt; element that contains the audio/video feed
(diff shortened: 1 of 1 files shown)


lib/addons/p5.dom.js
```diff
@@ -1070,8 +1072,8 @@
    * var capture;
    *
    * function setup() {
-   *   createCanvas(480, 120);
-   *   capture = createCapture(VIDEO);
+   *   createCanvas(480, 480);
+   *   capture = createCapture(VIDEO).hide();
```
lmccart: can you break this into two lines so it's super clear?

acasmith: Sure thing.

lmccart:
thanks!

PR #2807, opened 2018-04-23
add more parameter validation

Spongman:
adds parameter validation to a bunch more (mostly webgl) methods.

### DIFF
--- a/src/core/attributes.js
+++ b/src/core/attributes.js
@@ -104,6 +104,7 @@ p5.prototype.ellipseMode = function(m) {
  *
  */
 p5.prototype.noSmooth = function() {
+  p5._validateParameters('noSmooth', arguments);
   this._renderer.noSmooth();
   return this;
 };
@@ -202,6 +203,7 @@ p5.prototype.rectMode = function(m) {
  *
  */
 p5.prototype.smooth = function() {
+  p5._validateParameters('smooth', arguments);
   this._renderer.smooth();
   return this;
 };
--- a/src/core/attributes.js
+++ b/src/core/attributes.js
@@ -104,7 +104,6 @@ p5.prototype.ellipseMode = function(m) {
  *
  */
 p5.prototype.noSmooth = function() {
-  p5._validateParameters('noSmooth', arguments);
   this._renderer.noSmooth();
   return this;
 };
--- a/src/core/vertex.js
+++ b/src/core/vertex.js
@@ -62,6 +62,7 @@ var isFirstContour = true;
  *
  */
 p5.prototype.beginContour = function() {
+  p5._validateParameters('beginContour', arguments);
   contourVertices = [];
   isContour = true;
   return this;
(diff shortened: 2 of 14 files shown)


src/core/attributes.js
```diff
@@ -104,6 +104,7 @@ p5.prototype.ellipseMode = function(m) {
  *
  */
 p5.prototype.noSmooth = function() {
+  p5._validateParameters('noSmooth', arguments);
```
lmccart: are these ones necessary?

src/core/vertex.js
```diff
@@ -446,6 +448,7 @@ p5.prototype.curveVertex = function(x, y) {
  *
  */
 p5.prototype.endContour = function() {
+  p5._validateParameters('endContour', arguments);
```
lmccart: are these ones with no parameters necessary?

src/events/acceleration.js
```diff
@@ -355,6 +355,7 @@ var shake_threshold = 30;
  * @param {number} value The threshold value
  */
 p5.prototype.setMoveThreshold = function(val) {
+  p5._validateParameters('setMoveThreshold', arguments);
```
lmccart: this seems like a double check... the if statement below checks if the param is a number

src/core/vertex.js
Spongman: i think so. isn't `endContour('foo')` an error?

src/events/acceleration.js
Spongman: this is true, however, without the `_validateParameters` call, there's no warning given to the user if the type is wrong, the argument is missing (or null), or if extra parameters are passed.

Spongman:
> validateParameters() significantly hits performance

i don't think this has been the case for quite some time. here's a related issue @shiffman's ran into recently: https://github.com/CodingTrain/website/issues/542. The perf hit when using 0.5.16 was >40% CPU usage - which caused him to switch to the `.min.js` version. With the improvements that were made to 0.6.0 (mostly in #2302) the CPU usage for that example is now down to 0.7%. PR #2780 provides a further improvement, with a pathological test case in that PR showing a 1200% improvement.

as for parameter-less methods, i think it's still worth doing the parameter validation since there are still error cases for those methods for which there is currently no feedback, and if anything the performance impact of validating these methods is less (because there is only one overload, and there are no parameters, so nothing to check).

i would be very interested to hear if you have had issues with parameter validation performance recently. Especially if it's not something (like excess memory allocation) that's addressed in #2780. i think tools like this can be invaluable especially for those learning the library for the first time.

src/core/vertex.js
lmccart: i don't see anything that would cause this code to break if 'foo' were passed in?

src/events/acceleration.js
lmccart: ok, then let's remove the if check below..

lmccart:
@Spongman i think we also want to balance with not spewing a bunch of console messages at people when their code works. adding a param to endContour() is harmless. i think small errors like these when someone is learning are fine and leaving them be lets them focus on more critical console messages and coding errors.

src/core/vertex.js
Spongman: that's true. but if someone passed in an argument presumably expecting it to do something, and it didn't, wouldn't it be useful to let them know?

lmccart: please see my comment here https://github.com/processing/p5.js/pull/2807#issuecomment-384143965

Spongman:
thanks!

PR #2816, opened 2018-04-25
fix yui task

Spongman:
when running `grunt docs:dev`, the watch task doesn't rebuild the minified library so the 'test-minified' unit test always runs with stale code. this just add the `browserify:min` task to the `watch:yui` task, so the minified library gets built.


lmccart:
thank you. as a reminder, issues need to be opened before pull requests are opened and tagged with the issue. this is necessary for tracking development and keeping discussion clear. future prs without issues will be closed. thanks!

PR #2819, opened 2018-04-25
fix open:yui path

### LINKED ISSUE #2818: grunt docs:dev uses invalid host
`grunt docs:dev` opens a browser window to `http://0.0.0.0:9001/`, but 0.0.0.0 is not a valid host address and (for security reasons it's for binding only) Windows does not have a hidden route from 0.0.0.0 to localhost so it results in

image

Spongman:
change `0.0.0.0` in yui docs url to `127.0.0.1`

closes #2818

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -360,7 +360,7 @@ module.exports = function(grunt) {
     },
     open: {
       yui: {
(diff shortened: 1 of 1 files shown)


lmccart:
thanks!

Issue #2817, opened 2018-04-25
createCanvas() doesn't return an HTMLCanvasElement


lorenzo-lipparini:
#### Nature of issue?

- [x] Found a bug
#### Most appropriate sub-area of p5.js?

- [x] Core/Environment/Rendering

#### Details about the bug: 

The documentation states that the `createCanvas()` function should return an `HTMLCanvasElement` corresponding to the created canvas, but this is clearly not the case since a `p5.Renderer` is returned istead.

The thing happens in [this](https://github.com/processing/p5.js/blob/master/src/core/rendering.js#L116) line of code of the p5 source.

However, I'm not really sure about which one is wrong between the documentation and the implementation.

limzykenneth:
`createCanvas` is indeed supposed to return a `p5.Renderer` instance and not a `HTMLCanvasElement`. The documentation probably should be corrected in this case, if anyone wants to make a PR for this change please feel free. 😃

lad475:
I can take care of this; I'll make the PR in a bit :)

lmccart:
fixed with https://github.com/processing/p5.js/pull/2823

Spongman:
What's even weirder is that createCanvas doesn't really create a new canvas at all. It just replaces the existing canvas with a canvas of the right size (and renderer).

PixsaOJ:
What is weirder is, if I use `canvas = createCanvas(...)` In typescript I should be getting `HTMLCanvasElement` so I can set width height and other properties to it, but when I get `Renderer` I have to give my variable type of p5.Renderer | HTMLCanvasElement

PR #2823, opened 2018-04-26
Fixed Documentation for createCanvas()

lad475:
Fix to the issue referenced in #2817 

Changed text to read:
@return {p5.Renderer} object 
for the createCanvas() function


lmccart:
thanks!

PR #2826, opened 2018-04-26
Disallow es6 in library/sample code

### LINKED ISSUE #2825: build: linter allows ES6-isms in library & sample code
re: #2812

the linter currently allows ES6-isms (like `let` & `const`) in the library and doc sample code.

this can cause some confusing travis errors since the version of webkit used in PhantomJS doesn't support ES6. so, while the precommit check passes, the build may later fail.

Spongman:
closes #2825 

changes the linter settings to disallow ES6-isms for now.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -135,6 +135,11 @@ module.exports = function(grunt) {
         }
       },
       source: {
+        options: {
+          parserOptions: {
+            ecmaVersion: 5
+          }
+        },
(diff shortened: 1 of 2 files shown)


lmccart:
thanks! at some point we should i guess move over to es6, but that day is not today :)

Spongman:
https://www.youtube.com/watch?v=iMfgcA1zlGY

Issue #2828, opened 2018-04-28
Improve camera() example.


awarua:
#### Nature of issue?

- [ ] Found a bug
- [X] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Feature enhancement details:

The example for the camera() function does not really display _well_ - it just shows as completely white, so it's not clear from the example what the function is doing. 

https://p5js.org/reference/#/p5/camera

![image](https://user-images.githubusercontent.com/92529/39394068-ad02dcce-4b10-11e8-8085-b94d44c89a6d.png)

I would propose that the example should 
1. Draw the plane smaller, so it fills less of the canvas.
2. Don't move through the plane, causing it to disappear (this is harder to understand for a newbie).
3. Set the background to a different colour (e.g. 204) so it's easier to see the plane against it. 

I have made some adjustments to the example as per the above and will suggest this as a pull request.

PR #2829, opened 2018-04-28
Improve the camera example code

awarua:
Improves the example code for the `camera()` function as shown on the reference page as described in issue #2828

The previous example drew a plane that completely filled the canvas, so even though the camera was moving back and forth, this wasn’t visible. All you could see was white. This change:
1. Draws the plane smaller (10x10)
2. Moves the camera less (20 +- 10) so the plane always stays visible
3. Sets the background colour to 204 so the plane is more visible.
4. Adjust the alt text to be more accurate.


kjhollen:
awesome, thanks!

PR #2831, opened 2018-04-29
optimize `color()` by using loops instead of map()

### LINKED ISSUE #2830: `color()` is unnecessarily slow
0.6.1

the implementation of `color()` uses `map()` (inside `_parseInputs` and `_calculateLevels`) to loop through the color components for scaling/clamping operations.

in a simple test case:
```javascript
function setup() {
  createCanvas(100, 100);
}
function draw() {
  for (var i = 0; i < 100000; i++)
    color([255,0,100,255]);
}
```

these `map` call result in >90% of total CPU usage:
image

Spongman:
switches some usages of `map()` to using loops instead.

closes #2830.

with the example in that issue, CPU usage drops to <40% which results in a 3x FPS increase.
image

this can make a big difference to sketches that switch colors (like stroke/fill) many times per frame.

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -366,6 +366,7 @@ p5.Color.prototype.setAlpha = function(new_alpha) {
 // calculates and stores the closest screen levels
 p5.Color.prototype._calculateLevels = function() {
   var array = this._array;
+  // (loop backwards for performance)
   var levels = (this.levels = new Array(array.length));
   for (var i = array.length - 1; i >= 0; --i) {
     levels[i] = Math.round(array[i] * 255);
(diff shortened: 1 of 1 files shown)


limzykenneth:
All looks good to me. Just curious about one thing: why iterate the array backwards?

Spongman:
iterating backwards eliminates all but one of the length calculations. sometimes the JIT can eliminate those if it can guarantee constness, but that's not universal. also the comparison with zero is cheaper.

it took serious self-control not to use the 'down-to' (wink) operator:
```javascript
for (var i = array.length; i --> 0; ) {
}
```

limzykenneth:
`-->` is exotic even for me 😨

lmccart:
@Spongman can you add a comment in the first place you're doing this in this function so people understand it is for performance reasons?

Spongman:
ah. it turns out the extra parens are required by the linter.

kjhollen:
great, thanks!

PR #2833, opened 2018-04-29
Optimize get/loadPixels

### LINKED ISSUE #2832: get() is slow
0.6.1

re: #2822, #2065, #1485, #1235

currently there are two ways of getting pixel values from p5.js canvas/image/video elements:
1) call the `get(x,y)` method.
2) call `loadPixels()` and access the `pixels[]` array directly.

the problem with 1) is that it's _really_ slow (calling `getImageData()` for every value) and the problem with 2) is that it's really complicated for the user to get right (you need to handle non-integer inputs, check for out-of-bounds conditions, scale correctly for pixelDensity, etc...).

Spongman:
closes #2832 

this adds a `_pixelsDirty` flag to the p5 instance/Image/MediaElement that tracks if `loadPixels` needs to actually do anything. `get()` is updated to use `pixels[]` if it's not dirty. the upshot of this is that you can call `loadPixels()` once followed by many fast calls to `get()`. Existing behavior is preserved wrt. calling `get()` without `loadPixels()` (for large images on mobile browsers) and calling `get()` to retrieve sub-images.

adds a manual test example demonstrating calling `get()` on a video element. (running this on 0.6.1 will hang the browser).

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1883,6 +1883,8 @@
     this._cues = [];
     this._pixelDensity = 1;
     this._modified = false;
+    this._pixelsDirty = true;
+    this._pixelsTime = -1; // the time at which we last updated 'pixels'
 
     /**
      * Path to the media element source.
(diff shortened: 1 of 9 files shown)


Spongman:
unfortunately not. i originally thought the same thing, but it turns out that there's really no functional overlap. for example `updatePixels()` can be called on an image, and the `_pixelsDirty` flag cleared, but that image would still need to be uploaded to the GPU the next time it's used in a uniform.

kjhollen:
okay, we'll take it as-is, then! thanks.

Issue #2848, opened 2018-05-04
unable to validate methods of helper classes


Spongman:
v0.6.1

in part due to naming conflicts, it's not possible to add parameter validation to public methods of classes like `p5.Table`, `p5.Color` & `p5.Element`,

consequently, there is no parameter validation for methods of these classes.

PR #2849, opened 2018-05-04
add support for validating parameters of methods of non-p5 classes

Spongman:
re: #2848

this makes it possible to specify the class name when validating parameters, eg: `p5._validateParameters('p5.Element.size', arguments);`

if no class name is specified, then `p5` is assumed.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -21,7 +21,7 @@
  *
  * @module p5.dom
  * @submodule p5.dom
- * @for p5.dom
+ * @for p5
  * @main
  */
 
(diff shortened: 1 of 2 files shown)


lib/addons/p5.dom.js
```diff
@@ -21,7 +21,7 @@
  *
  * @module p5.dom
  * @submodule p5.dom
- * @for p5.dom
+ * @for p5
```
kjhollen: I'm not sure I understand what this does, can you explain where else this is applied and why it was changed?

Spongman: sure, the `@for` attribute specifies the class that the following methods are members of. if you look in the `data.json` all these dom methods are listed as being in the `p5.dom` class. it didn't matter before that this was wrong because it wasn't used anywhere, but now that the `lookupParamDoc` method is looking up docs my class name, this needs to be fixed.

kjhollen:
thanks!

PR #2850, opened 2018-05-04
add validation for non-p5 classes

Spongman:
re: #2848 

adds parameter validation to exposed methods of the following classes:
- `p5.Element`, `p5.MediaElement`
- `p5.Color`
- `p5.Graphics`, `p5.Image`
- `p5.TypedDict`, `p5.NumberDict`
- `p5.Table`, `p5.TableRow`
- `p5.XML`
- `p5.Font`
- `p5.Shader`

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1211,6 +1211,7 @@
    * </code></div>
    */
   p5.Element.prototype.addClass = function(c) {
+    p5._validateParameters('p5.Element.addClass', arguments);
     if (this.elt.className) {
       // PEND don't add class more than once
       //var regex = new RegExp('[^a-zA-Z\d:]?'+c+'[^a-zA-Z\d:]?');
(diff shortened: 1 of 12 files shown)


kjhollen:
I'm worried that parameter validation for testing & maintenance purposes has been conflated with the friendly error system. I think a user working with functions on these objects probably has more experience than the intended audience for the friendly error system, and can work with the unmodified errors.

There is a lot of value in running unit tests that validate the entire library's functions against the documentation, but I'm not sure it makes sense to run the friendly error system for all functions. Could it make more sense to incorporate this into the testing infrastructure in some way?

Spongman:
the reason i made this PR in the first place is that I ran into an issue where I was calling one of these methods wrong in a sketch and it took me a long time to track down the issue. had there been FES on that method I would have seen the issue immediately.

i figured it can't hurt (especially after the #2780 optimization), and if it only ever saves one person a few minutes figuring out what's wrong then i'll be happy.

kjhollen:
@lmccart added some documentation for the [friendly error system in the developer docs](https://github.com/processing/p5.js/blob/master/developer_docs/friendly_error_system.md) to clarify the scope a little. rather than having these suggestions always show for every class and function, here's another suggestion. @Spongman what if we took some of the ideas in your original patch and injected validateParameters into all of the functions as a testing task during the build process? then we'd get the benefit of using this as a testing feature, but clarify the scope of the user-facing errors.

I'm going to close this for now because we'd like to keep the scope of the friendly error system focused and manageable. however, i'd love to see a patch that made use of validateParameters to test the entire lib as a grunt task.

PR #2864, opened 2018-05-08
fix scaling of `plane` geometry

### LINKED ISSUE #2863: webgl: plane primitive is not lit correctly
v0.6.1

the 3d `plane` primitive is not lit correctly. this is because the normals aren't scaled correcty.

Spongman:
closes #2863 

fix scaling of `plane` geometry. This causes the plane's normals not to have zero-length, and allows the plane to be lit.


kjhollen:
thanks!

Issue #2865, opened 2018-05-09
documentation help wanted!


lmccart:
Hi  👋👋👋

We are using this issue to focus on documentation! Documentation is essential for new learners and experienced programmers alike. It helps make our community inclusive by extending a friendly hand to those that are less familiar with p5.js. It also helps us find the bugs and issues with the code itself, because we test and try things out as we document.

[There are several ways to contribute, you can read more about them on this page.](https://github.com/processing/p5.js/blob/main/contributor_docs/contributing_documentation.md)

If you are adding some documentation, please comment below. And don't forget to add yourself to the [contributors list](https://github.com/processing/p5.js#contributors)!

Welcome! We're so glad you're here!
❤️ the p5.js community

m90:
For anyone who wants to work on adding missing examples and doesn't know where to start, this seems to be where there are still examples missing right now:

```
src/io/files.js:1329: method p5.save missing example
src/webgl/material.js:14: method p5.loadShader missing example
src/webgl/material.js:125: method p5.shader missing example
src/webgl/p5.Geometry.js:52: method p5.Geometry.computeFaces missing example
src/webgl/p5.Geometry.js:94: method p5.Geometry.computeNormals missing example
src/webgl/p5.Geometry.js:133: method p5.Geometry.averageNormals missing example
src/webgl/p5.Geometry.js:154: method p5.Geometry.averagePoleNormals missing example
src/webgl/p5.Geometry.js:247: method p5.Geometry.normalize missing example
src/webgl/p5.Shader.js:270: method p5.Shader.setUniform missing example
```

faithwyu:
I would love to help out porting the tutorials from processing (https://processing.org/tutorials/) to p5!

lmccart:
@faithwyu that would be great!

raulgrell:
I'm going through the processing.org examples - some of the stuff that is missing seems to address the language itself as opposed to the library. For example, processing.org talks about "Integers and Floats" and "Datatype Conversions". Should I make Javascript equivalent entries?

lmccart:
@raulgrell good question. not yet. we're not set up with the right website/reference infrastructure yet to be able to display those correctly.

cecilebertin:
Hi, I'd like to work on porting the examples from the processing example page to the p5.js example page.

lmccart:
[...]

PR #2866, opened 2018-05-09
Clean up and normalize reference for TypedDict classes

m90:
Starting with https://github.com/processing/p5.js/issues/2865 I cleaned up the reference for this.

It's mostly about normalizing and fixing a few tiny issues. At times I was unsure if I was using too "docs-y" terminology, but I am sure this might be adressed in the review.


lmccart:
🎉

PR #2518, opened 2018-01-08
add `npm run docs:dev`

outofambit:
adds a grunt task and npm script for contributors to easily see the reference website locally (with watching)

`grunt yui:dev` or `npm run docs:dev`

based on @Spongman 's  advice in #1954 

all feedback welcome!

PR #2867, opened 2018-05-09
Add missing examples for acceleration functions

m90:
Also for #2865 - add examples for functions that didn't have example yet.

That being said: I am currently unable to get the reference up and running locally. After running `grunt yui:dev` I will see a layout and everything, but it won't be populated with any docs.

Do we have docs on how to get the docs running?

### DIFF
--- a/src/events/acceleration.js
+++ b/src/events/acceleration.js
@@ -353,7 +353,38 @@ var shake_threshold = 30;
  *
  * @method setMoveThreshold
  * @param {number} value The threshold value
+ * @example
+ * <div class="norender">
+ * <code>
+ * // Run this example on a mobile device
+ * // You will need to move the device incrementally further
+ * // the closer the square's color gets to white in order to change the value.
+ *
+ * var value = 0;
+ * var threshold = 0.5;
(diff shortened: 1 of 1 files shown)


lmccart:
@Spongman is this the correct task? we should update the gruntfile if so, so it's more clear

Spongman:
this is the one i have been using since @outofambit added it in #2518. it builds the library & docs then starts the web serve and watches for change-triggered rebuilds.

m90:
@Spongman `docs:dev` does not seem to exist in the repo that I forked / cloned yesterday:

```
frederik@fr-xps:~/projects/p5.js$ grunt docs:dev
Warning: Task "docs:dev" not found. Use --force to continue.

Aborted due to warnings.
```

Running `yui:dev` does what you describe (setting up server and watching) this very moment, when yesterday it only rendered an empty skeleton, so there might be a bug hidden in the tasks. Anyways, it looks like this: https://github.com/processing/p5.js/blob/master/developer_docs/inline_documentation.md#generating-documentation is up to date and correct.

PR #2870, opened 2018-05-10
Use proper module group for color conversion

m90:
There's currently a dead entry in the reference TOC because of this.

image


lmccart:
good catch!

Issue #2857, opened 2018-05-07
Removing opentype.js comments


limzykenneth:
I'm currently working on refining the documentation rendering on the website and I want to parse YUIDoc documentation from the packaged p5.js file directly on the website repo. However I believe comments of opentype.js is preventing me from doing so, as they also have comment block that starts with `/**` but it doesn't follow YUIDoc completely. In any case I don't want YUIDoc to parse opentype.js comments anyway.

Is there a way to require opentype.js in the source without including all of its comments as well when we build the whole p5.js package? That way we can remove a lot of extra comments that isn't related to p5.js also. Note I still do want the comments in p5.js itself.

lmccart:
opentype.js does offer a minified version but I don't think there's any way to get this from npm on command. I guess we would need to minify this file or all of the node_modules dependencies maybe, using the uglify task, prior to building the library.

Spongman:
Doesn't the p5.min.js version already exclude all these comments?

Spongman:
how about this? https://github.com/Spongman/p5.js/commit/7e602b95e8a90a6cf7a6ee8ea0787c27e39cdbba

limzykenneth:
Pretty much exactly what I'm looking for haha. If you think said package is reliable enough for us to use we can try it out to make sure it builds without introducing any bug.

Spongman:
yeah, i think it's fine. it does need that fix to work around the options bug, though.

lmccart:
@limzykenneth has this been resolved with #209?

limzykenneth:
@lmccart It's been awhile, I assume it is solved or not relevant anymore. Closing.

PR #2875, opened 2018-05-10
Enable local development of non-english reference docs

m90:
Right now localized origins like `/es/reference/#group-DOM` 404 in development. We can rewrite them to serve the docs content as all localization is happening on the client.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -352,6 +352,7 @@ module.exports = function(grunt) {
           middleware: function(connect, options, middlewares) {
             middlewares.unshift(
               require('connect-modrewrite')([
+                '^/(es|zh-Hans)/(.*) /docs/$2 [L]',
                 '^/assets/js/p5(\\.min)?\\.js(.*) /lib/p5$1.js$2 [L]',
                 '^/assets/js/p5\\.(dom|sound)(\\.min)?\\.js(.*) /lib/addons/p5.$1$2.js$3 [L]'
               ]),


limzykenneth:
The actual translations for the reference are hosted on p5.js-website repo and not here so without moving the generated reference folder over to that repo the translation will still not show up (they will still be in English). 

Eventually I would like to move reference generation over to the website repo to help make translation easier.

m90:
Yeah I noticed that too, I got quite confused when looking at how the entire setup works.

I also think that moving the reference into `p5.js-website` would make a lot of sense. LMK when you need help or input when doing so, I'd be interested in looking into that.

Spongman:
Wouldn't removing the reference generation kill the `yui:doc` task in p5? I use that _all the time_.

Spongman:
would it make sense to make the p5.js library repo a submodule of the website? that way duplicate could be avoided.

m90:
I'd have two questions / remarks:

- if the idea is to generate the actual reference in `p5.js-website`, would it make sense / be possible to remove all of the custom code in this repository and just use non-modified YUI Doc for local development?
- I like the idea of using a submodule, but maybe it'd be even nicer to pull in the data by installing the npm module of `p5.js` in `p5.js-website` and then generate the reference from the code in that package? That would also ensure that versions are always aligned w/ what people are actually working against.

limzykenneth:
@Spongman Hmm...interesting idea tbh. I'm not very familiar with using submodules with git but I will give it a bit of thought. Even though it is the website I don't want to make it too daunting for beginners to play around with it.

@m90 We need the custom code mainly to display the examples p5.js sketch so if someone edited the reference example, they will be able to preview it. As for pulling data from npm, I thought about it but find it too hacky and a bit too dependent on npm behaving as we expected (including in the future) in terms of the node_modules folder.

Spongman:
I think also an issue with using non, is that the 05 package wouldn't include the templates and other stuff from the p5 repo. That would still require those things to be duplicated between repos. The issue with duplicating these things is that invariably changes get made in both places, there's no direct support for merging those changes and hence bugs creep in. The submodule thing does add an extra step, but I believe this can be simplified with npm & grunt. The main benefit is it would eliminate the duplication.

m90:
@Spongman Yeah, I wasn't aware that that the docs in this repo cannot be as bare bones as I thought they could be. In that case, your proposal sounds like the most viable one.

Spongman:
i don't have much experience with submodules either, but from what i have gleaned, it should be possible to do something like this in the website's `package.json` file:

```javascript
"scripts" : { "preinstall" : "git submodule update --init --recursive" }
```

the `grunt-submodule` task might be useful here.

Spongman:
I would imagine it be easiest just to build the submodule and then import the generated data.json file and read the docs from there...

limzykenneth:
@m90 Yes sorry, got caught up with other things. Maybe for now let's put #2857 on hold since the git submodule idea seems promising. But I will only be trying that out in a week or two probably.

m90:
@limzykenneth No worries, just wanted to make sure you are not waiting for input on that topic from my side.

PR #2885, opened 2018-05-12
fix textLeading return value

### LINKED ISSUE #2884: textLeading returns the wrong thing
v0.6.1

the `@chainable` version of `textLeading` should return the `p5` instance (in accordance with the regular p5 get/set pattern). it currently returns the renderer.

Spongman:
closes #2884 

fix `textLeading`'s return when called in its chainable form.


Spongman:
Looks like this could be invalid when 'this' is a Graphics

PR #2894, opened 2018-05-12
add check for navigator.mediaDevices.getUserMedia in createCapture

aferriss:
On firefox 60, `navigator.getUserMedia` returns undefined. I added an additional check for `navigator.mediaDevices.getUserMedia` so that createCapture() didn't error out.

Looks like firefox is deprecating [getUserMedia](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getUserMedia) in favor of [mediaDevices](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia)

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1118,7 +1118,7 @@
         cb = arguments[i];
       }
     }
-    if (navigator.getUserMedia) {
+    if (navigator.getUserMedia || navigator.mediaDevices.getUserMedia) {
       var elt = document.createElement('video');
 
       if (!constraints) {
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1118,7 +1118,7 @@
         cb = arguments[i];
       }
     }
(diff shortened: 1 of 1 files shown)


lib/addons/p5.dom.js
```diff
@@ -1118,7 +1118,7 @@
         cb = arguments[i];
       }
     }
-    if (navigator.getUserMedia) {
+    if (navigator.getUserMedia || navigator.mediaDevices.getUserMedia) {
```
m90: Since the code below is using `navigator.mediaDevices.getUserMedia` exclusively, it might make sense to skip the currently used check entirely and check for `navigator.mediaDevices && navigator.mediaDevices.getUserMedia` only?

aferriss: That makes sense to me! I left the old check in there in case of older browsers. But if it's just going to use mediaDevices anyway, then there's no point in doing the check for navigator.getUserMedia

kjhollen:
verified fix in Firefox for mac. thanks @aferriss!

PR #2897, opened 2018-05-13
Ensure errors are always thrown in the same fashion

m90:
The way p5.js currently throws errors is not entirely consitent, which makes it harder for other libraries to interact with it.

This PR normalizes all throwing of errors to use the currently prevailing style of

```js
throw new Error('this is the error message');
```

which also [seems to be the most user friendly one](https://www.nczonline.net/blog/2009/03/10/the-art-of-throwing-javascript-errors-part-2/) as it will always print the message.

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -1093,7 +1093,12 @@ p5.prototype.httpDo = function() {
 
   (type === 'jsonp' ? fetchJsonp(path, jsonpOptions) : fetch(request))
     .then(function(res) {
-      if (!res.ok) throw res;
+      if (!res.ok) {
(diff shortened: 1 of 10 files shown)


src/io/files.js
```diff
@@ -1093,7 +1093,7 @@ p5.prototype.httpDo = function() {
 
   (type === 'jsonp' ? fetchJsonp(path, jsonpOptions) : fetch(request))
     .then(function(res) {
-      if (!res.ok) throw res;
+      if (!res.ok) throw new Error(res.body);
```
m90: I wonder if this should also include `res.statusText`?

Issue #2893, opened 2018-05-12
In keyPressed() some ascii values of key are incorrect.


devonsoft:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [x] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 0.6.1
- Web browser and version: Google Chrome (64-bit) | 67.0.3396.40 Firefox (64-bit) 59.0.2

- Operating System: Windows 8 (Macbook Pro 2013)
- Steps to reproduce this:

In keyPressed() some ascii values of key are incorrect. These values are correct in processing.js in the same environment.

function keyPressed()
{
print(key);
}

',' comes in as '¼'
'.' comes in as '¾'
'/' comes in as '¿'
' comes in as 'Þ'
'[' comes in as 'Û'
']' comes in as 'Ý'
'\' comes in as 'Ü'
'~' comes in as 'À'
'-' and SHIFT+'=' are also wrong

The other ascii characters seem ok but I may have missed some.

m90:
Do you specifically need to use `keyPressed` or could you also use `keyTyped`? Due to browsers doing very strange things when handling keyboard events, the output is very hard to handle.

If it's an option for you, using [`keyTyped`](https://p5js.org/reference/#/p5/keyTyped) will probably give you what you expect for most use cases.

kjhollen:
Fixed in #2898. thanks @m90!

PR #2898, opened 2018-05-13
Try to use event.key for reading character from keydown and keyup events

m90:
This would improve the situation for #2893 

`KeyboardEvent.key` seems too have [ok support](https://caniuse.com/#feat=keyboardevent-key) and its output is probably much closer to what a user would expect.

I might be missing something here though.


kjhollen:
Sorry for the delayed response here—verified this fix! Thanks for leaving in the fallbacks so that we can support IE11 too.

PR #2913, opened 2018-05-17
WebGL Arc added

AdilRabbani:
First patch for webgl arc(). Needs reviews and suggestions.

### DIFF
--- a/src/webgl/primitives.js
+++ b/src/webgl/primitives.js
@@ -800,59 +800,110 @@ p5.RendererGL.prototype.triangle = function(args) {
 };
 
 p5.RendererGL.prototype.ellipse = function(args) {
-  var x = args[0];
-  var y = args[1];
-  var width = args[2];
-  var height = args[3];
-  //detailX and Y are optional 6th & 7th
-  //arguments
-  var detailX = args[4] || 24;
-  var gId = 'ellipse|' + detailX;
-  if (!this.geometryInHash(gId)) {
-    var _ellipse = function() {
-      this.vertices.push(new p5.Vector(0.5, 0.5, 0));
-      this.uvs.push([0.5, 0.5]);
+  this.arc(args[0], args[1], args[2], args[3], 0 , PI * 2, 'ellipse' , args[4]);
+};
 
-      for (var i = 0; i < this.detailX; i++) {
-        var u = i / this.detailX;
-        var theta = 2 * Math.PI * u;
+p5.RendererGL.prototype.arc = function(args){
 
-        var _x = 0.5 + Math.cos(theta) / 2;
-        var _y = 0.5 + Math.sin(theta) / 2;
+   var x       = arguments[0];
+   var y       = arguments[1];
+   var width   = arguments[2];
+   var height  = arguments[3];
+   var start   = arguments[4];
+   var stop    = arguments[5];
+   var mode    = arguments[6];
+   var detailX = arguments[7] || 25;
+   var detailY = 1;
 
-        this.vertices.push(new p5.Vector(_x, _y, 0));
-        this.uvs.push(_x, _y);
+   var gId = 'arc|' + x + "|" + y + "|" + width + "|" + 
+   height + "|" + start + "|" + stop + "|" + mode + "|" + detailX + "|";
 
-        this.faces.push([0, (i + 1) % this.detailX + 1, i + 1]);
-      }
+   if (!this.geometryInHash(gId))
+   {
+
+    var _arc = function()
+    {
+        this.strokeIndices = [];
+
+        for (var i = 0; i <= this.detailX; i++)
+        {
+            var u = i / this.detailX;
+            var theta = (stop - start) * u + start;
+
+            var _x = 0.5 + Math.cos(theta)/2;
+            var _y = 0.5 + Math.sin(theta)/2;
+
+            this.vertices.push(new p5.Vector(_x, _y, 0));
+            this.uvs.push([_x, _y]);
+
+           if (i < this.detailX && mode == PIE)
+           {                                               
+                this.faces.push([0, i + 1, i + 2]);
+                this.strokeIndices.push([i + 1, i + 2]);
+           }
+
+           if (i < this.detailX - 1 && (mode == CHORD || mode == OPEN || mode == 'ellipse'))
+           {
+                this.faces.push([0, i + 1, i + 2]);
+                this.strokeIndices.push([i + 1, i + 2]);
+           }
+
+        }
+
+        if (mode == PIE || mode == null){
+            this.vertices.unshift(new p5.Vector(0.5,0.5,0));
+            this.uvs.unshift([0.5,0.5,0]);
+            this.faces.push([0, this.vertices.length - 2, this.vertices.length - 1]);
+            this.strokeIndices.push([0 , 1]);
+            this.strokeIndices.push([0, this.vertices.length - 1]);
+        }
+
+        else if (mode == CHORD){
+            this.strokeIndices.push([0 , 1]);
+            this.strokeIndices.push([0, this.vertices.length - 1]);
+        }
+
+        else if (mode == OPEN || mode == 'ellipse') {
+            this.strokeIndices.push([0 , 1]);
+        }
+
     };
-    var ellipseGeom = new p5.Geometry(detailX, 1, _ellipse);
-    ellipseGeom.computeNormals();
+
+    var arcGeom = new p5.Geometry(detailX, 1, _arc);
+    arcGeom.computeNormals();
+
     if (detailX <= 50) {
-      ellipseGeom._makeTriangleEdges()._edgesToVertices();
-    } else {
-      console.log('Cannot stroke ellipse with more than 50 detailX');
+      arcGeom._makeTriangleEdges()._edgesToVertices(arcGeom);
+    } 
+    else {
+      console.log('Cannot stroke arc with more than 50 detailX');
     }
 
-    this.createBuffers(gId, ellipseGeom);
-  }
+    this.createBuffers(gId, arcGeom);
 
-  // only a single ellipse (of a given detail) is cached: a circle of
-  // _diameter_ 1 (radius 0.5).
-  //
-  // before rendering, this circle is squished (technical term ;)
-  // appropriately and moved to the required location.
+   }
+
   var uMVMatrix = this.uMVMatrix.copy();
+  
   try {
-    this.uMVMatrix.translate([x, y, 0]);
+    if (mode == 'ellipse'){
+        this.uMVMatrix.translate([x, y, 0]);
+    }
+    else {
+        this.uMVMatrix.translate([x - 15, y - 15, 0]);
+    }
     this.uMVMatrix.scale(width, height, 1);
 
     this.drawBuffers(gId);
-  } finally {
+  } 
+
+  finally {
     this.uMVMatrix = uMVMatrix;
   }
+
   return this;
-};
+
+}
 
 p5.RendererGL.prototype.rect = function(args) {
   var perPixelLighting = this.attributes.perPixelLighting;


AdilRabbani:
I just noticed I didn't change the gId according to the shape i.e if an ellipse is used it would still be creating 'arc' gId for that. Would change it soon.
Also one weird issue I had was that the center for ellipse would be perfectly aligned with the origin (0,0) but for arc it would have a slight shift of (-15,-15). I ended up translating the arc back to 0 using uMVMatrix.translate(). Please do suggest if there is a better work around :)
I will fix all the lint errors once you are satisfied with the code.

Spongman:
it may not be necessary to have different gIds for ellipses & arcs. presumably presumably any cache overlap would look identical anyway. my question was more about the necessity of including the x, y, width & height properties in the cache key.

kjhollen:
Hi @AdilRabbani, thanks for submitting this! Stalgia and I get notifications when a review is requested, so it's not necessary to tag us in in your GSoC post in addition to requesting the review—we'll see it. :) I can review this tomorrow morning LA time!

src/webgl/primitives.js
```diff
@@ -800,59 +800,110 @@ p5.RendererGL.prototype.triangle = function(args) {
...
+   if (!this.geometryInHash(gId))
+   {
+
+    var _arc = function()
+    {
+        this.strokeIndices = [];
+
+        for (var i = 0; i <= this.detailX; i++)
+        {
+            var u = i / this.detailX;
```
kjhollen: this loop does not push any values to faces or strokeIndices when mode is undefined. mode can be undefined in the default usage, for example:

`arc(0, 0, 100, 100, 0, PI);`

I wrote some test code following the format given in the reference examples, and the results were unexpected. I think this is because later in this loop, there's no conditional for what to do when the mode is undefined.

src/webgl/primitives.js
```diff
@@ -800,59 +800,110 @@ p5.RendererGL.prototype.triangle = function(args) {
...
-  // appropriately and moved to the required location.
   var uMVMatrix = this.uMVMatrix.copy();
+  
   try {
-    this.uMVMatrix.translate([x, y, 0]);
+    if (mode == 'ellipse'){
+        this.uMVMatrix.translate([x, y, 0]);
+    }
+    else {
+        this.uMVMatrix.translate([x - 15, y - 15, 0]);
```
kjhollen: I think there's a problem with the placement of the center point for arcs. in the ellipse mode, the p5.prototype.ellipse adjusts the center point based on the current ellipse mode, while for arc this is done in p5.Renderer2D.prototype. this should probably be moved to p5.prototype in core/2d_primitives.js instead, so both 2d and 3d mode can use it. then I think we can lose this adjustment, which I don't think will work for all arc sizes.

src/webgl/primitives.js
```diff
@@ -800,59 +800,110 @@ p5.RendererGL.prototype.triangle = function(args) {
...
+    arcGeom.computeNormals();
+
     if (detailX <= 50) {
-      ellipseGeom._makeTriangleEdges()._edgesToVertices();
-    } else {
-      console.log('Cannot stroke ellipse with more than 50 detailX');
+      arcGeom._makeTriangleEdges()._edgesToVertices(arcGeom);
+    } 
+    else {
+      console.log('Cannot stroke arc with more than 50 detailX');
```
kjhollen: a nice detail here might be to say "ellipse" instead when appropriate; so that user will see the name of the method they called

src/webgl/primitives.js
```diff
@@ -800,59 +800,110 @@ p5.RendererGL.prototype.triangle = function(args) {
...
+
+           if (i < this.detailX - 1 && (mode == CHORD || mode == OPEN || mode == 'ellipse'))
+           {
+                this.faces.push([0, i + 1, i + 2]);
+                this.strokeIndices.push([i + 1, i + 2]);
+           }
+
+        }
+
+        if (mode == PIE || mode == null){
```
kjhollen: I think the mode == null should not be here, or at least the stroke indices shouldn't be added in that case. The shape shouldn't be closed by default, so that it's possible for the user to draw just curved lines (see examples 2 & 5 here: https://p5js.org/reference/#/p5/arc — it's subtle, but the 2nd has no outline on non-curved edges)

src/webgl/primitives.js
```diff
@@ -800,59 +800,110 @@ p5.RendererGL.prototype.triangle = function(args) {
...
+   var start   = arguments[4];
+   var stop    = arguments[5];
+   var mode    = arguments[6];
+   var detailX = arguments[7] || 25;
+   var detailY = 1;
 
-        this.vertices.push(new p5.Vector(_x, _y, 0));
-        this.uvs.push(_x, _y);
+   var gId = 'arc|' + x + "|" + y + "|" + width + "|" + 
+   height + "|" + start + "|" + stop + "|" + mode + "|" + detailX + "|";
```
kjhollen: as Spongman notes, you can remove any parameters here that aren't factors in the geometry and instead happen through scale/translate later (x, y, width, height)

Spongman:
i'd recommend running a `grunt lint:fix` to fix most of the linting issues that travis is flagging. your git should have run the `precommit` task before allowing you to commit this. you might want to investigate why that's not happening.

src/webgl/primitives.js
AdilRabbani: Fixed. Tested on reference examples.

src/webgl/primitives.js
AdilRabbani: I moved the function arcModeAdjust from p5.Renderer2D.prototype to p5.prototype.arc. The problem still persists :( arc is not rendered on the expected position.

src/webgl/primitives.js
AdilRabbani: Fixed.

src/webgl/primitives.js
AdilRabbani: Fixed

src/webgl/primitives.js
AdilRabbani: Fixed. Removed x, y, width and height.

src/webgl/primitives.js
AdilRabbani: Fixed with latest commit.

test/manual-test-examples/webgl/arc/sketch.js
```diff
@@ -0,0 +1,27 @@
...
+
+function setup() {
+  createCanvas(800, 600, WEBGL);
+}
+
+function draw() {
+  background(230);
+  strokeWeight(3);
+  stroke(237, 34, 93);
+  fill(255);
```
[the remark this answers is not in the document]
kjhollen: also: just a note that in the "open" + fill case, it seems like maybe the stroke is still being closed? there are some curious rendering artifacts when the shape is very small, but it's hard to tell what's causing them.

package-lock.json
```diff
@@ -103,6 +103,7 @@
       "version": "5.5.2",
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-5.5.2.tgz",
       "integrity": "sha1-c7Xuyj+rZT49P5Qis0GtQiBdyWU=",
+      "dev": true,
```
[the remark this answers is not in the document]
kjhollen: (if you haven't rebased or amended a commit before, we can talk about how to do this—just let me know!)

Spongman:
looking really good. here's another example with 2d & 3d side-by-side:

two things:
- the line widths are wrong. but i think this is due to the issues addressed by #2657
- i think the normals may be flipping in some cases which may have implications when the ellipse is lit.

```javascript
var r3d, modes, a1 = 0, a2 = 0;

function setup() {
  modes = [OPEN, CHORD, PIE, (void 0)];

  createCanvas(windowWidth, windowHeight/2);
  angleMode(DEGREES);
  fill(255, 0, 0);
  stroke(255);
  strokeWeight(5);

  r3d = createGraphics(windowWidth, windowHeight/2, WEBGL);
  r3d.canvas.style.display = 'block';
  r3d.angleMode(DEGREES);
  r3d.fill(255, 0, 0);
  r3d.stroke(255);
  r3d.strokeWeight(5);
}

function draw() {
  a1 -= 1;
  a2 += Math.PI/2;

  background(0, 10, 0);
  r3d.background(10, 0, 0);

  translate(width/2, height/2);
  var s = min(width/5, height) * 2/3;

  for (var i = 0; i < modes.length; i ++) {
    var x = (i - (modes.length-1)/2) * width / 5;
    arc(x, 0, s, s, a1, a2, modes[i]);
    r3d.arc(x, 0, s, s, a1, a2, modes[i]);
  }
}
```

package-lock.json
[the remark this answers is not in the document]
AdilRabbani: Sorry for creating this mess. Don't know how this was added in the commit. And yes, I haven't rebased a commit before. Would need help in this one.

src/core/2d_primitives.js
```diff
@@ -77,7 +77,7 @@ require('./error_helpers');
  *white ellipse with top right quarter missing with black outline around the shape.
  *
  */
-p5.prototype.arc = function(x, y, w, h, start, stop, mode) {
+p5.prototype.arc = function(x, y, w, h, start, stop, mode, xDetail) {
```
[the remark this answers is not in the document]
AdilRabbani: Done!

src/webgl/primitives.js
```diff
@@ -800,49 +800,97 @@ p5.RendererGL.prototype.triangle = function(args) {
...
+  var width = arguments[2];
+  var height = arguments[3];
+  var start = arguments[4];
+  var stop = arguments[5];
+  var mode = arguments[6];
+  var detailX = arguments[7] || 25;
+
+  var shape;
+
+  if (Math.abs(stop - start) === Math.PI * 2) {
```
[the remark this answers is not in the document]
AdilRabbani: Fixed.

src/webgl/primitives.js
```diff
@@ -800,49 +800,97 @@ p5.RendererGL.prototype.triangle = function(args) {
...
+    shape = 'ellipse';
+  } else {
+    shape = 'arc';
+  }
+
+  var gId = shape + '|' + start + '|' + stop + '|' + mode + '|' + detailX + '|';
```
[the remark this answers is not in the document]
AdilRabbani: Fixed.

src/webgl/primitives.js
```diff
@@ -800,49 +800,97 @@ p5.RendererGL.prototype.triangle = function(args) {
...
+        this.uvs.push([_x, _y]);
+
+        if (i < this.detailX - 1) {
+          this.faces.push([0, i + 1, i + 2]);
+          this.strokeIndices.push([i + 1, i + 2]);
+        }
+      }
 
-        this.faces.push([0, (i + 1) % this.detailX + 1, i + 1]);
+      if (mode === 'pie') {
```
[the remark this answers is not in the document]
AdilRabbani: Fixed. Using constants now.

test/manual-test-examples/webgl/arc/sketch.js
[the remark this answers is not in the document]
AdilRabbani: Updated example.

src/webgl/primitives.js
```diff
@@ -800,49 +801,115 @@ p5.RendererGL.prototype.triangle = function(args) {
...
+  var stop = arguments[5];
+  var mode = arguments[6];
+  var detail = arguments[7] || 25;
+
+  var shape;
+  var gId;
+
+  if (Math.abs(stop - start) >= constants.PI * 2) {
+    shape = 'ellipse';
+    gId = shape + '|' + x + '|' + y + '|' + width + '|' + height + '|';
```
kjhollen: you can remove x,y, width and height here because they are part of the model scaling on line 915.

src/core/2d_primitives.js
```diff
@@ -97,36 +101,36 @@ p5.prototype.arc = function(x, y, w, h, start, stop, mode) {
   start %= constants.TWO_PI;
   stop %= constants.TWO_PI;
 
-  // account for full circle
-  if (stop === start) {
-    stop += constants.TWO_PI;
+  // account for full circle, don't display anything
+  if (stop.toFixed(10) === start.toFixed(10)) {
+    start += constants.TWO_PI;
```
kjhollen: let's update this to use the behavior Adam described in the bug.

PR #2928, opened 2018-05-21
In scale doc, replace rotate example with basic scale example.

kgolid:
The `scale()` documentation contains two code examples where one is a `rotate()` example.  Replaced it with the basic `scale()` example used on processing.org.


lmccart:
thanks!

PR #2944, opened 2018-05-23
Added clipping plane value checks to perspective.

### LINKED ISSUE #2936: Odd behavior when perspective() near plane set to 0
#### Nature of issue?

- [ ] Found a bug
- [X] Existing feature enhancement
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL

- [X] Desktop/Laptop

#### Details about the bug: 

- p5.js version:  0.6.1
- Web browser and version:  Chrome 66.0.3359.181
- Operating System:  MacOSX
- Steps to reproduce this:

```
function setup() {
  let fov = 60 / 180 * PI;
  perspective(fov, width / height, 0, 500);
}
function draw() {
  background(200, 100, 200);
  orbitControl();

  normalMaterial();
  rotateX(0.3);
  rotateY(frameCount * 0.01);
  box(100);
}
```

In perspective(), when the near clipping plane is set to 0, certain faces become transparent as their angle to the camera changes.  Any value above zero (even 0.001) stops this behavior.  Also, if the near clipping plane is set below 0, nothing is shown.

I am unclear whether this is desired behavior or not.  I would expect that when the near clipping plane is at or behind the camera (0 or below), everything within the frustum in front of the camera would be visible.  

Would it be possible to add a friendly warning OR a check for zero/below-zero values on the input?  I am happy to take a look at this if either solution sounds ok.

AidanNelson:
Fixes #2936.

Adds two checks to perspective() arguments: 

1. Checks whether near plane is at or below 0.0001.  If so, resets this value to 0.01 and logs a warning to console. This value (0.0001) seems to be about where strange artifacts become very noticeable.  See below screenshot of perspective near plane set to 0.00011:

screen shot 2018-05-23 at 7 07 16 pm

2. Checks whether far plane value is less than near plane and logs a warning to the console.

Is there any reason these checks should happen elsewhere in the function (after this.cameraNear and this.cameraFar are set, for instance)?

### DIFF
--- a/src/webgl/camera.js
+++ b/src/webgl/camera.js
@@ -221,6 +221,21 @@ p5.RendererGL.prototype.perspective = function(fovy, aspect, near, far) {
     far = this.defaultCameraFar;
   }
 
+  if (near <= 0.0001) {
+    near = 0.01;
+    console.log(
+      'Avoid perspective near plane values close to or below 0. ' +
+        'Setting value to 0.01.'
+    );
+  }
+
+  if (far < near) {
+    console.log(
+      'Perspective far plane value is less than near plane value. ' +
+        'Nothing will be shown.'
+    );
+  }
+
   this.cameraFOV = this._pInst._toRadians(fovy);
   this.cameraAspect = aspect;
   this.cameraNear = near;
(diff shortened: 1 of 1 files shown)


Spongman:
i think you might want to check that it's actually the absolute value of `near` that is causing the issue. it's my understanding that these kinds of issues are more due to a round error caused by a logarithmic difference between near and far as opposed to any absolute size. for example, having a near/far of 0.0001,0.0002 shouldn't be a problem (or even 1e-50,2e-50). but having a near/far of 0.0001,10000 will be problematic as there are 8 orders of magnitude difference there which is going to lose accuracy quickly in 32-bit float operations.

i think the check should probably be something like `assert(Math.log(far) - Math.log(near) < 8 * Math.log(10))`

kjhollen:
Hi Aidan! I think I need to refresh my camera matrix math a little bit before I review this one. I'm going to read over the article that Stalgia linked in the issue, then review this early next week.

AidanNelson:
Ok, I have some reading to do myself!  Spongman, the article you linked (https://www.khronos.org/opengl/wiki/Depth_Buffer_Precision) does complicate the problem a bit. 

Is there a simple way to advise users to keep their clipping planes as close together as possible?

Issue #2698, opened 2018-03-13
Promisify the preload functions


meiamsome:
If no one else minds, I would be interested in taking this on and converting the preloads to all use promises under the hood. I suggest this would involve adding `loadXPromise(..)` versions of each of the current loading functions and having the older `loadX(...)` functions be mapped onto them, perhaps along the lines of the following snippet, that should be able to keep backwards compatibility viable. 
```javascript
p5.prototype.loadX = function() {
  var ret = {};
  ret.promise = p5.loadXPromise.apply(this, arguments);
  ret.promise.then(function() {
    // Set properties on ret like before
  });
}
```

In the future, we could then deprecate the use of the return directly and prefer to use `loadX().promise.then(ret => variable = ret)` and after a long deprecation time we could convert to something like the following:

```javascript
p5.prototype.loadX = function() {
  var ret =  p5.loadXPromise.apply(this, arguments);
  ret.promise = ret;
}
```

And then deprecate the use of `.promise`. I think this makes sense from a future standpoint where we get rid of the current `loadX` behaviour but that is a somewhat desirable feature. Perhaps it may even be a good idea to let `preload` return a promise. This would let users do

```javascript
let file1, file2;
async function preload() {
  file1 = await loadJSON(...);
  file2 = await loadString(...);
}
```

Which would be relatively new user friendly I think. [Browser support](https://caniuse.com/#search=await) is getting there for async/await so I think this is actually viable in the near future. 

It is notable, however that using `await` like this would actually cause the loads to be called one-by-one, so perhaps this is not the best solution.

limzykenneth:
I was thinking about promisifying the loading functions since they already are promises underneath (using fetch) but I didn't get the time to try anything out. What you propose seem to me a good path to start exploring. A few notes:

The promisified function by convention (as per Bluebird) is to suffix `Async`, whether to follow this or not I don't have a strong opinion. Node.js also has a promisify function but instead of creating a new promisified function, it replaces the current one, with callback, with a promisified one using the same name.

We need to consider the case where the data that was loaded in would have the property `promise` thereby replaceing our own `promise` property.

The current way to deal with parallel `await` is to do `await Promise.all(...)` which I'm not sure is good enough usability-wise.

meiamsome:
@limzykenneth 
[...]

PR #2948, opened 2018-05-26
Preload Function Testing

meiamsome:
Updates/adds more testing for all preload functions in the core library.

Things of note:
- Some of the preload functions where throwing errors asynchronously, some of them were running `console.error` with the error. I have standardised them all to call `console.error`
- Are loadShader and loadModel webgl-specific? should they be moved into the webgl test folder?

`loadShader`:
- I had to add callback support
- The current implementation takes vertFilename, fragFilename in that order (So only the frag is optional, however the Processing documentation for [loadShader](https://processing.org/reference/loadShader_.html) there takes them in the opposite order, so the API between them is a bit inconsistent here, is that intentional?

(Base work for #2698)


meiamsome:
Split into #2985, #2987, #2988, #2989, #2991, #2992, #2993, #2994, #2995, #2996, #2997

kjhollen:
AWESOME! Thank you!!!! 🌟

PR #2949, opened 2018-05-26
Add support for a Promise being returned from httpDo, httpGet and httpPost

meiamsome:
Forwards the Promise from the underlying fetch so that it is exposed to users of the http APIs.

- Note that I have specified Promise as the return type, but this is showing a warning about typescript definitions. I do not know enough about these to fix this warning, however, so these need to be fixed by someone who knows that area of code.

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -769,6 +769,9 @@ p5.prototype.loadBytes = function(file, callback, errorCallback) {
  * @param  {function}      [errorCallback] function to be executed if
  *                                    there is an error, response is passed
  *                                    in as first argument
+ * @return {Promise} A promise that resolves with the data when the operation
+ *                   completes successfully or rejects with the error after
+ *                   one occurs.
  * @example
  * <div class='norender'><code>
  * // Examples use USGS Earthquake API:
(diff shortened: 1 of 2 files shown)


Spongman:
I trunk those errors can be safely ignored. There's a list of external types at the top of the generator script that you can add to if you want to shut it up.

src/io/files.js
```diff
@@ -1091,35 +1104,41 @@ p5.prototype.httpDo = function() {
...
+  } else {
+    promise = fetch(request);
+  }
+  promise = promise.then(function(res) {
+    if (!res.ok) {
+      var err = new Error(res.body);
+      err.status = res.status;
+      err.ok = false;
+      throw err;
+    }
```
limzykenneth: Can you explain a bit about why the style change here? Not that I don't like it or something like that but rather I believe @Spongman recently changed it to the one line style and I don't want this to keep switching around as different people work on the codebase. Thanks.

limzykenneth:
Except for the line comment, this looks great! It's been something I meant to do for some time now but never got the motivation to do it.

src/io/files.js
meiamsome:
```javascript
var promise = (type === 'jsonp' ? fetchJsonp(path, jsonpOptions) : fetch(request))
  .then(...)
```
fell fowl of the linter for having too long a line - I could either break the ternary operator across multiple lines or split it into an if with a variable. As a variable was needed later anyway to return the correct promise (Before the `.then` and `.catch`, so that the errors are propagated forward correctly), this seemed like the cleanest way to do it to me.

meiamsome:
@Spongman I wasn't sure what the correct name for a Promise is in typescript, or if it needs to be external or whatever. As you say, these appear to only be warnings, so that should be non-blocking on this PR and can then be fixed later by someone else.

mebble:
Hey there. The p5 website reference doesn't mention anything about these methods returning promises. Is this feature available in the current p5 version? If so I'd like to add to the documentation

limzykenneth:
@mebble The reference is currently out of date because of the bug on the deployment process. It has since been fixed but an updated version has not been pushed to the site yet.

mebble:
Great! Around when could we expect to see the updated docs?

limzykenneth:
The latest probably when the next version is released

PR #2950, opened 2018-05-26
halfImage example in image.js was doing full image

hamoid:
I'm not sure, but I think the example is missing `img.` to calculate how many pixels is half image.


kjhollen:
thanks!

Issue #2930, opened 2018-05-23
orbitControl z-axis is inverted


Spongman:
v0.6.1 webgl bug

https://p5js.org/reference/#/p5/orbitControl

if you click in the middle and drag the cursor left/right (with no vertical motion) then the model follows the cursor intuitively.

if you click in the middle and drag the cursor up/down the motion of the model seems inverted.

kjhollen:
fixed in webgl-gsoc-2018

PR #2956, opened 2018-05-28
Fix issue with orbitControl rotating inversely on the X axis

dekmm:
Addressing issue explained in #2930

### DIFF
--- a/src/webgl/interaction.js
+++ b/src/webgl/interaction.js
@@ -33,7 +33,7 @@ p5.prototype.orbitControl = function() {
   p5._validateParameters('orbitControl', arguments);
   if (this.mouseIsPressed) {
     this.rotateY((this.mouseX - this.width / 2) / (this.width / 2));
-    this.rotateX((this.mouseY - this.height / 2) / (this.width / 2));
+    this.rotateX((this.height / 2 - this.mouseY) / (this.height / 2));
   }
   return this;
 };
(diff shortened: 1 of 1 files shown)


dekmm:
Not sure if the inverted rotation was intented, but I would like to add a flag so the user can choose between normal and inverted rotation.

dekmm:
I guess I never noticed the different rotation speeds since I tested the change on a canvas of equal width and height. Must've slipped my mind that I made the change.

I'm assuming the change isn't that drastic unless a user uses some weird resolution for the canvas such as 1000x100 or has OCD and these little inconsistencies drive him crazy :P but I'll test it out and change it back if necessary.

kjhollen:
@dekmm one context I was thinking of was using the browser's developer console, which usually results in a screen much wider than tall because the dev panel takes up the bottom half of the screen. I'm often doing this to run tests or debug.

Spongman:
I wonder if it behaves differently if the canvas is large vs if it is small?  I think users might expect the turning rate to be related to the 'size' of the canvas, but also in a way that the horizontal & vertical turning rates are the same.

(I have a feeling that getting this exactly right would depend on inspecting the camera matrix, but on the other hand, I think that assuming a square pixel aspect ratio would probably be sufficient.)

kjhollen:
thanks for updating this. I just merged #2962 which included a similar fix + more features, so I'm going to close this, but the update was right on. :)

PR #2958, opened 2018-05-29
Remove redundant calls in p5.Shader.js

xiatianyu:
Calls to [`_loadAttributes()`](/processing/p5.js/blob/29fdfa36b8204ea1301270bb577d461377f282ea/src/webgl/p5.Shader.js#L197-L198) and [`_loadUniforms()`](/processing/p5.js/blob/29fdfa36b8204ea1301270bb577d461377f282ea/src/webgl/p5.Shader.js#L197-L198) inside `bindShader()` are deemed redundant as they would have been called by `init()`. Note that `getProgramParameter()`, `getActiveAttrib()`, `getAttribLocation()`, `getActiveUniform()` and  `getUniformLocation()` can hinder performance as they need to interrupt the GPU in order to query the WebGL states concerned, so we want to make sure they are not called for no reason. Also removed is the [return statement](https://github.com/processing/p5.js/blob/29fdfa36b8204ea1301270bb577d461377f282ea/src/webgl/p5.Shader.js#L38) in the constructor, which serves no real purpose.

### DIFF
--- a/src/webgl/p5.Shader.js
+++ b/src/webgl/p5.Shader.js
@@ -191,7 +191,7 @@ p5.Shader.prototype.bindShader = function() {
     this.useProgram();
     this._bound = true;
     this.bindTextures();
-    
+
     this._renderer._setDefaultCamera();
     this._setMatrixUniforms();
     if (this === this._renderer.curStrokeShader) {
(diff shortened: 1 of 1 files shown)


kjhollen:
thanks! yes, it makes sense to remove these as they are covered in `init` earlier in `bindShader`, but the performance impact here should be minimal because both `_loadAttributes()` and `_loadUniforms()` only make the GL calls if the attributes or uniforms haven't been loaded before.

Issue #2189, opened 2017-09-21
textureWrap() or similar for webgl mode


kjhollen:
#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Feature enhancement details:

Processing has a [textureWrap](https://processing.org/reference/textureWrap_.html) feature for changing the texture wrapping behavior.

this is a small size project for someone with a little experience working with open gl. see the [WebGL Module Architecture](https://github.com/processing/p5.js/wiki/WebGL-Module-Architecture) page on p5 conventions for webgl.

Siphonophora:
Thought I would add a useful example I found when looking up how these options should behave. http://www.cs.uregina.ca/Links/class-info/315/WWW/Lab5/#Repeat

Siphonophora:
Clamp mode does seem to be working fine (applied to a box, where I changed the UVs). Would this feature normally apply only to geometry made with createShape() as suggested in the processing documentation?

![image](https://user-images.githubusercontent.com/32316111/31364802-47b054d8-ad35-11e7-8092-63e0fb3208d5.png)

Benjamin-Davies:
If you have the texture bound then you can use
``` js
gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.REPEAT);
gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.REPEAT);
```
to tell webgl to repeat the texture when you go over the borders and
``` js
gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
```
to tell WebGL to clamp to the edge of the texture when you go over the borders.
However, this will only set the option for the currently bound texture.
To do it for all textures you would need to either iterate through all of the textures, or set a flag to do it when any texture is bound/used.

kartik-25:
Hi, I would like to have a go at this issue but since i am new to webgl I might require a bit of guidance.

[...]

PR #2960, opened 2018-05-31
Texture filtering and wrapping

aferriss:
I was wanting some more control over how textures were filtered and wrapped so I implemented these two functions on top of the Texture class. 

````
tex.setInterpolation( downscale, upscale);
// downscale & upscale can be constants NEAREST or LINEAR

tex.setWrapMode( wrapX, wrapY);
// wrapX and wrapY can be CLAMP, REPEAT, or MIRROR
// if the texture is not power of two it will fallback to clamp
````

WebGL2 does support non power of two texture for REPEAT and MIRROR but since there wasn't an obvious way to get at version information I just implemented it as a webGL1 function. I also left out mipmapping for now.

You can use them like so...
````
let canvas, img;
function preload(){
   img = loadImage('img.jpg');
}

function setup(){
  canvas = createCanvas(400,400,WEBGL);
  let tex = canvas.getTexture(img);
  tex.setTextureWrap(REPEAT, REPEAT);
  tex.setInterpolation(NEAREST, NEAREST);
}

````

### DIFF
--- a/src/webgl/p5.Texture.js
+++ b/src/webgl/p5.Texture.js
@@ -9,7 +9,7 @@
 'use strict';
 
 var p5 = require('../core/core');
-var constants = require('./constants');
+var constants = require('../core/constants');
 
 /**
  * Texture class for WEBGL Mode
@@ -254,24 +254,31 @@ p5.Texture.prototype.unbindTexture = function() {
 };
 
 /**
- * Sets texture min / mag filters
- * @method setTextureFilters
+ * Sets how a texture is be interpolated when upscaled or downscaled.
+ * Nearest filtering uses nearest neighbor scaling when interpolating
+ * Linear filtering uses WebGL's linear scaling when interpolating
+ * @method setInterpolation
+ * @param {String} downScale Specifies the texture filtering when
+ *                           textures are shrunk. Options are LINEAR or NEAREST
+ * @param {String} upScale Specifies the texture filtering when
+ *                         textures are magnified. Options are LINEAR or NEAREST
+ * @todo implement mipmapping filters
  */
-p5.Texture.prototype.setTextureFilters = function(minFilter, magFilter) {
+p5.Texture.prototype.setInterpolation = function(downScale, upScale) {
   var gl = this._renderer.GL;
 
-  if (minFilter === constants.NEAREST) {
+  if (downScale === constants.NEAREST) {
     this.glMinFilter = gl.NEAREST;
-  } else if (minFilter === constants.LINEAR) {
+  } else if (downScale === constants.LINEAR) {
     this.glMinFilter = gl.LINEAR;
   } else {
     // falling back to default
     this.glMinFilter = gl.LINEAR;
   }
 
-  if (magFilter === constants.NEAREST) {
+  if (upScale === constants.NEAREST) {
     this.glMagFilter = gl.NEAREST;
-  } else if (magFilter === constants.LINEAR) {
+  } else if (upScale === constants.LINEAR) {
     this.glMagFilter = gl.LINEAR;
   } else {
     // falling back to default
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/webgl/p5.Texture.js [hunk 1]
```diff
@@ -252,4 +253,112 @@ p5.Texture.prototype.unbindTexture = function() {
...
+p5.Texture.prototype.setInterpolation = function(downScale, upScale) {
+  var gl = this._renderer.GL;
+
+  if (downScale === constants.NEAREST) {
+    this.glMinFilter = gl.NEAREST;
+  } else if (downScale === constants.LINEAR) {
+    this.glMinFilter = gl.LINEAR;
+  } else {
+    // falling back to default
+    this.glMinFilter = gl.LINEAR;
```
kjhollen: Is the format of this block in anticipation of adding more possible scaling constants in the future? both of the last two conditions result in gl.LINEAR (and same is done below).

aferriss: I think I did this on the off chance that someone mistypes one of the constants so that it would still just choose something that worked. Is it worth throwing an error there instead or at least a warning that the constant was not recognized and LINEAR was chosen instead?

kjhollen: ah oops I think I wasn't clear! any reason not to just have the else part (and lose the else-if)?

aferriss: No reason at all except my oversight. Lets axe it!

kjhollen:
Thanks! This looks good to me. One thing I forgot to ask before, was if you can put an example in test/manual-test-examples/webgl ? There are a couple image assets in there already that you can use.

aferriss:
No problem, example added!

tcoppex:
Hi, 

So relative to issue #2189 I've been thinking a bit more about this and I'm not sure anymore if the two features could be easily merged.

As I see it, _p5.RendererGL.textureWrap_ is intended to change the value of the wrapping mode globally whereas _p5.Texture.setWrapMode_ change it locally, giving a more fine-grained control over the result, which is arguably better.

Ideally, here's what we want :


```javascript
function load_images() {
  A = loadImage('imageA.png');
  B = loadImage('imageB.png');
}

function draw() {
  // A wrapmode set to REPEAT.
  A.setTextureWrap(REPEAT, REPEAT);

  // Global wrapmode set to MIRROR.
  textureWrap(MIRROR);

  // A and B are MIRROR.
  textured_quad(A, u, v);
  textured_quad(B, u, v);

  translate(3, 0, 0);

  // Change A wrapmode to REPEAT.
  A.setTextureWrap(REPEAT, REPEAT);

  // A is REPEAT, B is MIRROR.
  textured_quad(A, u, v);
  textured_quad(B, u, v);
}
```

My first implementation matches the one from @aferriss  but the two proved incompatible.
I've tried somes angles, mainly storing states locally or globally, without much success.

Ideally _textureWrap()_ should broadcast its new state immediately to existing images,
which might be done by storing them in a buffer inside the p5 context when _createImage()_ is called
but this might broke the lifecycle intended by the user altogether.

Hopefully there is something simple I don't see, anyways any helps is welcome.

Thanks !

aferriss:
Hi @tcoppex ! Thanks for looking into this. Am I missing something or can't you just loop over the textures array stored in the canvas. Here is a little example I implemented on top of this PR that does exactly what you suggest (though I just implemented the global function in this sketch)

````
var A;
var B;
var Atex;

let canvas;
function preload() {
  A = loadImage('../../assets/UV_Grid_Sm.jpg');
  B = loadImage('../../assets/cat.jpg');
}

function setup() {
  canvas = createCanvas(windowWidth, windowHeight, WEBGL);
  Atex = canvas.getTexture(A);
}

function draw() {
  background(255);

  const dX = mouseX / windowWidth;
  const dY = mouseY / windowHeight;

  let u = lerp(1.0, 8.0, dX);
  let v = lerp(1.0, 8.0, dY);

  scale(width/9.0);

  // set just A to repeat
  Atex.setWrapMode(REPEAT, REPEAT);

  // set all textures to mirror
  textureWrap(MIRROR);

  push();
  translate(-1.5, -1, 0);
  textured_quad(A, u, v);
  pop();
  
  push();
  translate(-1.5, 1, 0);
  textured_quad(B, u, v);
  pop();

  // set just A to repeat
  Atex.setWrapMode(REPEAT, REPEAT);

  push();
  translate(1.5, -1, 0);
  textured_quad(A, u, v);
  pop();

  // b is still mirror
  push();
  translate(1.5, 1, 0);
  textured_quad(B, u, v);
  pop();
  
}

function textured_quad(tex, u, v) {
  texture(tex);
  beginShape(TRIANGLES);
  vertex(-1, -1,  0, 0, 0);
  vertex( 1, -1,  0, u, 0);
  vertex( 1,  1,  0, u, v);

  vertex( 1,  1,  0, u, v);
  vertex(-1,  1,  0, 0, v);
  vertex(-1, -1,  0, 0, 0);
  endShape();
}

function textureWrap(mode){
  for(let i =0; i<canvas.textures.length; i++){
    canvas.textures[i].setWrapMode(mode, mode);
  }
}
````

**Comment by tcoppex** [comment]
Hey @aferriss, 
I did not saw that array, that seems perfect indeed, thanks.
Maybe you can go ahead and implement the feature along your PR if you're feeling like it ? it might be simpler that way.

Cheers :+1:

**Comment by aferriss** [comment]
@tcoppex sure, I'll throw it in there too! Do you think it's good to add the method so that it can take either one or two arguments for maximum flexibility? For example...

````
// set clamp on both x and y axes
textureWrap(CLAMP); 

// set clamp on x axis and repeat on y axis
textureWrap(CLAMP, REPEAT);
```

tcoppex:
I think it won't hurt : it keeps the legacy with Processing and add flexibility :-)
(maybe @kjhollen can help though, I don't know if there is any philosophy behind multiple definitions in p5.js)

PR #2961, opened 2018-05-31
update gl pipeline to set shader uniforms at render time

### LINKED ISSUE #2370: style push/pop for webgl mode

kjhollen:
p5.RendererGL is updated to hold more state so that shader uniforms
can be set later, when it's time to render a shape. then, a shader
does not have to be bound at all time to change state. this enables
things like full support for push/pop.

in particular, p5.RendererGL state is added for tracking:

- whether or not the normal shader is in use
- use of lighting (boolean)
- use of specular component (boolean)

the curFillShader and curStrokeShader objects are removed from
p5.RendererGL in favor of selecting a default shader at render time,
based on current gl state (lighting conditions, texture settings,
and fill/stroke). additionally, _userFillShader and _userStrokeShader
are introduced to indicate that a custom shader has been set by the
library user using a call to shader(). in this case, the user shader
takes priority over any default shaders. this enables the user to mix
and match calls to fill, stroke, lights, etc. to set uniforms in
their own shader.

adds p5.RendererGL._setFillUniforms and
p5.RendererGL._setStrokeUniforms to set all of the known default
uniforms for the shader, to be used just before shape render time.

unit tests are updated to reflect this new structure; still heavily
dependent on internal state but better than nothing.

Fixes #2370.

### DIFF
--- a/src/webgl/light.js
+++ b/src/webgl/light.js
@@ -73,27 +73,13 @@ p5.prototype.ambientLight = function(v1, v2, v3, a) {
   p5._validateParameters('ambientLight', arguments);
   var color = this.color.apply(this, arguments);
 
-  var shader = this._renderer._useLightShader();
-
-  //@todo this is a bit icky. array uniforms have
-  //to be multiples of the type 3(rgb) in this case.
-  //a preallocated Float32Array(24) that we copy into
-  //would be better
-  shader.setUniform('uUseLighting', true);
-  //in case there's no material color for the geometry
-  shader.setUniform('uMaterialColor', this._renderer.curFillColor);
-
   this._renderer.ambientLightColors.push(
     color._array[0],
     color._array[1],
     color._array[2]
   );
-  shader.setUniform('uAmbientColor', this._renderer.ambientLightColors);
 
-  shader.setUniform(
-    'uAmbientLightCount',
-    this._renderer.ambientLightColors.length / 3
-  );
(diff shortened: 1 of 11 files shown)


kjhollen:
Yes, the timing is tough with summer of code. How about I wait until the point shader code is merged to the webgl-gsoc-2018 branch, then update my patch to account for that?

kjhollen:
Closing this / update with point shader coming shortly...

PR #2967, opened 2018-06-02
createFileInput example now inside code tags

TanviKumar:


kjhollen:
thanks @TanviKumar! 🎊

PR #2985, opened 2018-06-08
Add Testing helper

meiamsome:
Splitting #2948 - this is just helper functions that all the other PRs use.

### DIFF
--- a/test/js/p5_helpers.js
+++ b/test/js/p5_helpers.js
@@ -1,0 +1,23 @@
+function promisedSketch(sketch_fn) {
+  var myInstance;
+  var promise = new Promise(function(resolve, reject) {
+    myInstance = new p5(function(sketch) {
+      return sketch_fn(sketch, resolve, reject);
+    });
+  });
+
+  promise.catch(function() {}).then(function() {
+    myInstance.remove();
+  });
+  return promise;
+};
+
+function testSketchWithPromise(name, sketch_fn) {
+  var test_fn = function() {
+    return promisedSketch(sketch_fn);
+  };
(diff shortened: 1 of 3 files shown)


Spongman:
I can't help but think that all this would be so much easier & clearer if we could use async/await in the tests...

meiamsome:
@Spongman I'm not sure I agree with that statement - async/await is not as versatile as promises themselves. I guess it lets you limit the use of Promises to capturing small pieces of code. This PR mainly gives a new helper that can be used to minimise shared state between tests.

kjhollen:
hey @meiamsome thank you very much for this detailed write-up! I'm sorry that it's taken me so long to respond to this—I'm certainly a little out of my depth here, so it takes me a little longer to review and understand this part of the code base.

I think I might have been a little unclear about my questions with `testPreloadError` or `testPreloadSuccess`—does it make sense to use methods like these that would internally call `testSketchWithPromise`? I'm thinking that there still seems to be a lot of boiler plate or nearly copy pasted code in most of the preload tests, which feels ripe for some kind of abstraction. It would be nice to do something like `testPreloadError('loadFont', 'not-found.tff')` or `testPreloadSuccess('loadFont', 'known-font.ttf')` and have each know how to configure the promise call backs to just do the right thing. Does that (or something close to it) seem reasonable / possible?

kjhollen:
Lauren and I reviewed this in more detail today and are ready to merge it! future simplifications or refactors are welcome in another patch, but we're excited to get more stability and coverage in the load tests. thanks @meiamsome! additional comments other PRs is forthcoming.

PR #2993, opened 2018-06-08
Update loadBytes tests

meiamsome:


test/unit/io/loadBytes.js
```diff
@@ -0,0 +1,161 @@
...
+    reject
+  ) {
+    var object;
+    sketch.preload = function() {
+      object = sketch.loadBytes(validFile, function() {}, reject);
+    };
+
+    sketch.setup = function() {
+      resolve(
+        new Promise(function(resolve, reject) {
```
kjhollen: could this be written slightly more simply with an if statement (or series of if/else statements) checking the various conditions, resulting in calls to resolve() or reject()? it's a little tough to read with this additional Promise embedded here, because it doesn't seem that anything asynchronous is happening (unless I am misunderstanding what some of these functions do?)

meiamsome: Yeah, I agree this is sub-ideal. Could also do a try/catch around them all, like this, which might be more readable:
```javascript
  testSketchWithPromise('returns an object for object JSON.', function(
    sketch,
    resolve,
    reject
  ) {
    var object;
    sketch.preload = function() {
      object = sketch.loadBytes(validFile, function() {}, reject);
    };

    sketch.setup = function() {
      try {
        assert.isObject(object);
        // Check data format
        expect(object.bytes).to.satisfy(function(v) {
          return Array.isArray(v) || v instanceof Uint8Array;
        });
        // Validate data
        var str = 'GIF89a';
        // convert the string to a byte array
        var rgb = str.split('').map(function(e) {
          return e.charCodeAt(0);
        });
        // this will convert a Uint8Aray to [], if necessary:
        var loaded = Array.prototype.slice.call(object.bytes, 0, str.length);
        assert.deepEqual(loaded, rgb);
        resolve();
      } catch (e) {
        reject(e);
      }
    };
  });
```

Otherwise, now that we have chrome testing we could use async/await for better readability:
```javascript

  test('returns an object for object JSON.', async function() {
    const sketch = promisedSketch(function(sketch, resolve, reject) {
      var object;
      sketch.preload = function() {
        object = sketch.loadBytes(validFile, function() {}, reject);
      };

      sketch.setup = function() {
        resolve(object);
      };
    });
    const object = await sketch;
    assert.isObject(object);
    // Check data format
    expect(object.bytes).to.satisfy(function(v) {
      return Array.isArray(v) || v instanceof Uint8Array;
    });
    // Validate data
    var str = 'GIF89a';
    // convert the string to a byte array
    var rgb = str.split('').map(function(e) {
      return e.charCodeAt(0);
    });
    // this will convert a Uint8Aray to [], if necessary:
    var loaded = Array.prototype.slice.call(object.bytes, 0, str.length);
    assert.deepEqual(loaded, rgb);
  });
```

Let me know which of these is to your preference (I am very pro the last example there), and I'll update all of the test PRs

kjhollen: I agree the update is more legible! async/await is still relatively new to me, so I just want to make sure I'm understanding: the result is that when the promise resolve()s, it returns the object?

in any case, reading this version feels more linear to me, so it's a little easier to follow as an example: first the asynchronous stuff happens, then we validate the results. preload/setup are short and to the point. I like it. I saw that @Zalastax just reviewed the other patches. Do you want to update them or merge as-is?

Zalastax: I think it would be best to merge them as is. So far p5 is still developed as an ES5 project and before the discussion about ES6 has been resolved I think it's best to keep to that so the tests can be run in all browsers we support. If you want async/await please voice that opinion in the issue about rollup+babel!

kjhollen: hmm I thought using headless chrome was the only piece needed here in the tests to be ready to move on from ES5, at least in this capacity? we're isolated to the tests here, so is it necessary to support IE? these will mostly be run via the command line using the headless chrome module. or hmm, I guess it's true that people can open the test.html file in IE and then they'd have an issue. what do you think @meiamsome?

meiamsome: @kjhollen yep, you're right - `promisedSketch` takes a function that is an instance mode sketch with `resolve` and `reject`. When `await`-ing on that, if it resolves, the value passed to resolve is returned. If the promise is rejected, the error passed into reject is throw as it would be in a synchronous function. Because we are now inside the main testing function, errors thrown are reported correctly as failures by mocha, so we can use `assert` and `expect` freely. In either case, `promisedSketch` automatically cleans up the sketch for the test, so we will hopefully end up with less tests hanging around when they weren't supposed to.

In terms of these prs, this has got a bit complicated! I actually updated two of the PRs to async/await code already because they became stale, that's [loadJSON](https://github.com/processing/p5.js/pull/2988/files) and [loadStrings](https://github.com/processing/p5.js/pull/2989/files) - so you can check those for a fuller view on what the async/await implementations look like. [async/await support](https://caniuse.com/#feat=async-functions) is available in the three most recent versions of most common browsers that would be used in testing while progressing, so I don't think compatibility is that problematic, and if the user wants to run `npm run test` they will be using chrome which will have support anyway. I am massively in favour of the readability of the async/await format which should hopefully make it a bit easier for people to understand what's going on in the tests. In the worst case (i.e. using IE), the files will fail to parse correctly, so the tests will be ignored locally, but will still be flagged in PRs.

There is also a third option, which I guess would be like this:
```javascript
  test('returns an object for object JSON.', function() {
    return promisedSketch(function(sketch, resolve, reject) {
      var object;
      sketch.preload = function() {
        object = sketch.loadBytes(validFile, function() {}, reject);
      };

      sketch.setup = function() {
        resolve(object);
      };
    }).then(function(object) {
      assert.isObject(object);
      // Check data format
      expect(object.bytes).to.satisfy(function(v) {
        return Array.isArray(v) || v instanceof Uint8Array;
      });
      // Validate data
      var str = 'GIF89a';
      // convert the string to a byte array
      var rgb = str.split('').map(function(e) {
        return e.charCodeAt(0);
      });
      // this will convert a Uint8Aray to [], if necessary:
      var loaded = Array.prototype.slice.call(object.bytes, 0, str.length);
      assert.deepEqual(loaded, rgb);
    });
  });
```
Which gives you a less-clear but still linear flow. There are problems, though, including that it's very easy to accidentally not return the resulting Promise, which will make the test always get marked as a pass even if the code failed.

lmccart:
Hi all, to weigh in on a few things:

Yes, we are moving to ES6. The timeline is still being finalized on that. As of now, ES6 is not fully supported by IE, so a first step will be adding a polyfill for missing functionality. I would like to wait on this and other ES6 updating until we've officially kicked things off. I will post this in the rollup+babel thread as well.

However, since this is the test suite and the async/await adds significant legibility, I think we can consider an exception here. As long as the tests run from command line, I think that's sufficient. This means (if I'm understanding correctly) the tests cannot be run from an IE browser, but a developer working primarily on IE could still use command line to develop, build, and run all tests. If this is correct, I'm in favor of moving ahead with async/await here.

kjhollen:
@lmccart sounds good to me! Yes, I'll merge these once they're updated. Thank you @meiamsome for being patient while we figure this out. I'll review the updates whenever you're ready! and I expect to be online more in the coming weeks.

lmccart:
@meiamsome @kjhollen just checking in on these. are there plans to make the changes discussed?

lmccart:
Hey all, just trying to do some cleanup and organization here. I will close the loadX tests PRs unless I hear from @kjhollen or @meiamsome this week. If you need more time, just let me know.

meiamsome:
Hello all, sorry for the wait - I've been a bit busy lately. I'm gunna work on getting the PRs up to spec today. I'll tag you @kjhollen for review when each one is ready.

meiamsome:
As it turns out, I don't have the ability to request reviews, so I'll just mention here that I updated all the PRs, ready for your review @kjhollen

lmccart:
Thanks @meiamsome! Sorry this took so long to merge. They're all in there except the loadShader PR. I need to get someone from the webgl team to look at that. If you want to separate out the tests from the shader fix I can merge the test one now.

meiamsome:
Thanks @lmccart! - Re: shaders, I think all if not most of the tests will fail without that fix because they're mostly testing the standard loading function behaviours, so I think it's ok for those tests to wait.

lmccart:
@meiamsome got it, that makes sense. thanks

PR #2998, opened 2018-06-08
Fix error in keyboard events

meiamsome:
Changes event to e, fixing a bug introduced from #2898


m90:
Ah, I just did the same in #3001, sorry for the dupe - it might make sense to add the additional `eslint` rule though as it could have caught this.

kjhollen:
good catch! I tested in Chrome for Mac & didn't realize the event object wouldn't be available in other browsers. I'll merge this one to fix the current build, and will spend some more time looking at your patch @m90 to see what other impacts changing the linting rules have.

kjhollen:
thanks @meiamsome!

PR #3000, opened 2018-06-08
Allow Promises from lifecycle functions

meiamsome:
Permits a user to optionally return a Promise from `preload`, `setup` and/or `draw` and will use them to continue when appropriate.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -268,12 +268,10 @@ var p5 = function(sketch, node, sync) {
         obj[method] = this._wrapPreload(obj, method);
       }
 
-      userPreload();
-      this._runIfPreloadsAreDone();
+      this._incrementPreload();
+      Promise.resolve(userPreload()).then(this._decrementPreload.bind(this));
     } else {
-      this._setup();
-      this._runFrames();
-      this._draw();
+      this._runIfPreloadsAreDone();
     }
   }.bind(this);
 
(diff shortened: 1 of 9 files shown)


kjhollen:
thanks for trying this out, @meiamsome. Promises are a pretty advanced concept for new programmers, who will be reading the documentation for setup, draw, etc. I'd prefer to keep things as simple as possible here, but also understand how this functionality might be useful for testing on the dev side—I'm going to close this for now, but am open to more discussion if you have ideas for how we might get the testing ease without adding complexity to the docs. thanks!

meiamsome:
@kjhollen This isn't related to testing at all, this is just primarily for ease of use with asynchronous libraries such as tfjs. 

For example, I have this test example that uses tfjs to do a large convolution in the `draw` loop here: [source](https://github.com/meiamsome/Tensorflowjs-wave/blob/master/asynctest/sketch.js) [example](https://meiamsome.github.io/Tensorflowjs-wave/asynctest/) 
Which would be practically impossible from p5 as it currently stands without having to manually implementing the redraw logic inside the sketch.

I do totally accept the want (or even need) for simplicity, though, I just thought that I'd point that out in case it is worth reconsidering on its own merits? 

In addition, this only adds a documentation change to `redraw` that now returns a Promise, but the documentation there already discusses the fact that it is asynchronous (even though it wasn't before, but I digress)

meiamsome:
@lmccart Given the above - would you please consider re-opening this PR for me? I've been working on #2698 some more today and I really can't see it working in any reasonable way without at least `preload` being able to return a promise.

PR #3001, opened 2018-06-09
Fix typo when accessing keyboard events, add linting rule

m90:
There was a typo in #2898 that made it in due to `eslint` thinking that `event` might refer to the global.

I also added a rule to the `eslint` config that makes this an error.


m90:
Kind of a dupe of #2998

PR #3009, opened 2018-06-11
Typo: Repeated textSize(); command

gryffydd:


kjhollen:
nice catch—thanks @gryffydd!

PR #3010, opened 2018-06-11
Add YUIDoc for constants

Zalastax:
These constants are referenced in other documentation.
They therefore need to be documented to generate TypeScript types.

Is there a reason for some constants (not named _CONSTANT) not to have YUIDoc?

### DIFF
--- a/docs/yuidoc-p5-theme-src/scripts/tpl/library.html
+++ b/docs/yuidoc-p5-theme-src/scripts/tpl/library.html
@@ -11,7 +11,9 @@
     <div class="column_<%=col%>">
   <% } %>
   <% if (group.name !== module.name && group.name !== 'p5') { %>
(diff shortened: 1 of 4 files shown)


kjhollen:
I was testing this out today and think the changes show up in two places:

1) in the search results on the reference page (click search API and see autocomplete dropdown, which seems ok for constants within the core for p5, but the dom constants do not link to which functions they are associated with)
2) for the DOM lib, AUDIO and VIDEO also show up in the list of p5.dom functions, but should probably be removed (run `grunt yui:dev` and click on the p5.dom link at top of page to see)

kjhollen:
Yeah! Let's make another group on the p5.dom page for the constants.

Zalastax:
Does this look fine?
These changes seemed like the most reasonable to add support for a Constants group.

kjhollen:
yeah! awesome! built this and everything looks great.

PR #3025, opened 2018-06-15
filename changes in src/core

### LINKED ISSUE #2947: consider filename changes for src/core
#### Nature of issue?

Opening this issue as part of my research for my Processing Foundation Fellowship project working on internal developer-facing documentation and principles.

I think it's worth considering changing these filenames in `src/core` in the interest of clarity; I've articulated separate reasoning for each. I'm posting them all here in one issue, but it may be wise to break out into individual issues per filename for more detailed discussion.

- [ ] `2d_primitives.js` feels a bit wordy, especially because there is no file that defines 3d primitives, and no file that defines other non-drawing primitives. This is mostly about drawing shapes, and could be `shapes.js` or similar.
- [ ] `attributes.js` strongly suggests DOM attributes, not the miscellaneous drawing parameters it actually contains.
- [ ] `canvas.js` doesn't really contain functionality that is directly related to `<canvas>` elements.
- [ ] `core.js` is a bit redundant. This entire repository can be considered the p5 "core" (as distinct from addons), and it already has the `src/core` directory, and within that directory is `src/core/core.js`.
- [x] `structure.js` doesn't structure the whole library or the p5 *object*; it could be argued that it structures the drawing loop, but I think something that refers to that using a less generic term (maybe `loop.js`, if nothing else?) would be helpful.
- [ ] `p5.Graphics.js` creates optional graphics *buffers* for bespoke rendering scenarios, but certainly does not contain all graphics-related functionality.
- [x] `api_notifications.js`: "api" and "notifications" are both broad terms that apply to other concepts in this codebase — the p5 object provides a public API, Friendly Error Messages can be considered notifications, and so on. Maybe `legacy.js` or `processingCompatibility.js` or similar?

lmccart:
this pull request makes the following changes discussed in issue #2947 

* `canvas.js` > `helpers.js`
* `core.js` > `main.js`
* `webgl/primitives` > `webgl/3d_primitives.js`
* move `2d_primitive.js`, `attirbutes.js`, `curves.js`, and `vertex.js` to `shape` folder within `core` folder

closes #2947


lmccart:
hi @kjhollen hoping we can merge this one soonish.. I'm having a hard time understanding how to track and resolve conflicts and I don't want to lose anything.

kjhollen:
built this locally and merged it with master—there was only one minor conflict to fix in p5.Texture.js, I'm not sure why more are being reported. I did a double check on the README in src/core and things look good. maybe the best way to do this is to have you merge locally & push the results to the same branch @lmccart?

lmccart:
@kjhollen ok i think we should be good now..

kjhollen:
thank you @lmccart!

Issue #3017, opened 2018-06-13
Implementing point for WebGL mode


AdilRabbani:
So I was working on this one for quite a while and need some suggestions on what should be the most desirable approach in implementing this primitive?
WebGL by default draws square points. Making points circular requires some changes in the fragment shader but on the cost of performance i.e drawing default points is much faster. So is it necessary to go for a circular point or should the default one be more preferable?
Any suggestions/ideas would be helpful...

AidanNelson:
The 2D point implementation (https://p5js.org/reference/#/p5/point) draws a single pixel.  How does WebGL size points?  Will they be larger than a single pixel in screen-space?

AdilRabbani:
It all depends on the size given to gl_PointSize. I have these screenshots, 
![webgl](https://user-images.githubusercontent.com/15272015/41362358-9da7696c-6f4a-11e8-8aaf-631779291f61.PNG)
![p2d](https://user-images.githubusercontent.com/15272015/41362372-a3cda162-6f4a-11e8-99f7-4e832f8e5132.PNG)
I guess they look identical?

kjhollen:
2D mode has circles if you use something like `strokeWeight(10)` before calling `point()`

I think people will expect circular points. They could also get better performance using,

```
beginShape(POINTS);
// as many calls to vertex() as you want
endShape();
```

which would probably be faster than a bunch of calls to `point()` (and is also not currently implemented). I think @aferriss looked into this recently & might have some insight.

AdilRabbani:
One thing I would like to mention here is that, circular points are created by discarding pixels from a square point. So making gl_PointSize too small, would completely discard the points altogether, showing nothing...

aferriss:
@AdilRabbani I ran into the same thing too re: the circle points being really small. One solve I played around with was to just draw the square points when the point size is below a certain val, otherwise we discard to make the circle mask. Here's the shader I was using if it helps.


````
precision mediump float;
precision mediump int;

uniform vec4 uMaterialColor;
varying float vStrokeWeight;

void main() {

  float mask = 0.0;
  
  // make a circular mask using the gl_PointCoord (goes from 0 - 1 on a point)
  // might be able to get a nicer edge on big strokeweights with smoothstep but slightly less performant
  mask = step(0.98, length(gl_PointCoord * 2.0 - 1.0));

  // if strokeWeight is 1 or less lets just draw a square
  // this prevents weird artifacting from carving circles when our points are really small
[...]

Issue #2184, opened 2017-09-20
Implement point() for webgl mode


kjhollen:
#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Feature enhancement details:

point() only works in 2D mode currently, but it would be great for WEBGL mode, too.

this is a somewhat self-contained issue and would be great for a new contributor who wants to write some webgl code, but maybe a little more work than some of the other issues I've labeled with level:beginner. see: [WebGL Module Architecture](https://github.com/processing/p5.js/wiki/WebGL-Module-Architecture) for relevant documentation.

bomoko:
Hi, I'd really like to volunteer do this one.
I actually attempted this a little while ago (when working through some or other example processing sketch in Matt Pearson's book and found point wasn't implemented) but stopped when I heard the WebGL layer was being reworked.

kjhollen:
yay! go for it!

fractalfrenzy:
Any update on this?

bomoko:
Hi @fractalfrenzy (and @kjhollen, I suppose :) )
I've fallen a bit behind on this - wasn't as quick a patch as I'd imagined (required quite a bit more reading through the source, and reading about WebGL, than I'd anticipated - but that probably speaks to the lacunae in my own knowledge than the difficulty of the task itself).

I'm still planning on getting to this - but if anyone else wants to have a go, please feel free.

kjhollen:
added in webgl-gsoc-2018!

PR #3031, opened 2018-06-18
Webgl Point Added

### LINKED ISSUE #2662: beginShape(POINTS) not working in WebGL

AdilRabbani:
Added point for WebGL. This is based on the fragment shader provided by @aferriss #3017 . Uses default square points for small points and circular points for larger ones. Closes #2662 #2184


src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -326,4 +335,29 @@ p5.RendererGL.prototype._drawStrokeImmediateMode = function() {
   this._pInst._pixelsDirty = true;
 };
 
+p5.RendererGL.prototype.initPointVertexBufferImmediate = function() {
```
mlarghydracept: This function name feels a little bit long and misleading. It is doing more than simple initialization. One could argue that the line 51 with the call to `createBuffer()` is the actual immediate mode point vertex buffer initialization. This function is doing something a little closer to drawing. Could you change the name to something a little bit shorter and with an underscore to indicate its privacy like `_drawPoints(points, amt)`? Notice I also put arguments in that example. I think it would be fairly simple to collapse the immediate and retained mode versions of this into a single function at the renderer level.

src/webgl/primitives.js
```diff
@@ -744,7 +744,14 @@ p5.prototype.torus = function(radius, tubeRadius, detailX, detailY) {
 
 //@TODO
 p5.RendererGL.prototype.point = function(x, y, z) {
```
mlarghydracept: Make sure to write the comment text for generating an example in the docs.

src/webgl/shaders/point.vert
```diff
@@ -0,0 +1,11 @@
+attribute vec3 vPosition;
+uniform float vPointSize;
```
mlarghydracept: I would try to keep the naming convention from the other shaders. These should use that first letter to indicate the type of data. So this would be `uPointSize`. [The light vertex shader](https://github.com/processing/p5.js/blob/master/src/webgl/shaders/light.vert) has good naming conventions.

test/manual-test-examples/webgl/point/sketch.js
```diff
@@ -0,0 +1,27 @@
...
+    randStroke_g = Math.floor(Math.random() * 255 + 1);
+    randStroke_b = Math.floor(Math.random() * 255 + 1);
+    randStrokeWeight = Math.floor(Math.random() * 10 + 1);
+    stroke(randStroke_r, randStroke_g, randStroke_b);
+    strokeWeight(randStrokeWeight);
+    randX = Math.floor(Math.random() * (-800 - 800 + 1)) + 800;
+    randY = Math.floor(Math.random() * (-600 - 600 + 1)) + 600;
+    point(randX, randY);
+  }
+}
```
mlarghydracept: This is good. I would also add an example for immediate mode drawing that pushes the number of points being drawn to a much higher number.

kjhollen: Great suggestion. I'm also wondering—does this flash the dots quickly? If so, I suggest using noLoop() in setup or setting the frameRate to only update the drawing every ~5 seconds. The flashing/strobing effect may be harmful to some viewers.

src/webgl/shaders/point.vert
kjhollen: Some documentation of the existing shader uniforms [can be found here in webgl_mode_architecture.md](https://github.com/processing/p5.js/blob/ef774d56336eeb9781d340fc069ac2e1e85e26f0/developer_docs/webgl_mode_architecture.md)—and it would be great to add a column for the point shader at the end of this document!

src/webgl/p5.RendererGL.Retained.js
```diff
@@ -336,4 +336,31 @@ p5.RendererGL.prototype._drawElements = function(drawMode, gId) {
...
 
+p5.RendererGL.prototype.initPointVertexBuffer = function(x, y, z) {
+  var gl = this.GL;
+
+  var pointVertexBuffer = gl.createBuffer();
+
+  this._bindBuffer(
+    pointVertexBuffer,
+    gl.ARRAY_BUFFER,
+    [x, y, z],
```
kjhollen: Hmm, seeing this, I wonder if some slight performance gains can be made by making an array once and keeping it around.

src/webgl/shaders/point.frag
```diff
@@ -1,7 +1,7 @@
 precision mediump float;
 precision mediump int;
-uniform vec4 color;
-varying float v_strokeWeight;
+uniform vec4 uColor;
```
kjhollen: Could this be uMaterialColor to better match other shaders? Or is it a distinctly different property?

PR #3032, opened 2018-06-18
documentation

vijithassar:


src/README.md
```diff
@@ -10,10 +10,18 @@ The `app.js` file is simply an index of all the other modules which exports the
...
+Public APIs which are exposed to users of the p5.js library during the course of working on sketches and projects should use short, clear, declarative function names. `circle()` is preferable to `new Circle()`. If the name of a public API function is longer than one or two words, it may be worth carefully considering whether the API should be restructured into something more creative, expressive, or intuitive.
 
-The above are guidelines, and in practice there may be occasional exceptions.
+## Native
+
+Native APIs provided by the browser can sometimes be aliased in the style of public APIs to provide a more creative, expressive, or intuitive interface than the native implementation. For example, `print()` is easier to explain than `console.log()`. Idiomatic JavaScript is generally preferable, so aliasing in this manner needs to have tremendous creative or pedagogical benefits.
+
+Native browser functionality is also often aliased with properties on the `p5` object. This allows the library to consistently point toward the same reference, but in most cases this practice should not be considered part of the public API even though those properties are exposed to the user.
```
lmccart: I'm not quite sure what this is referring to, can you point me to an example?

vijithassar:
Here's a case where [device events are mirrored](https://github.com/processing/p5.js/blob/master/src/core/core.js#L226-L231) on the `p5` object and given slightly different names (camelcase, etc). This pattern is used repeatedly and it seemed to me to be worth explaining to potential new contributors why it's often preferable to use idiosyncratic proxies for the standard web APIs they may be more familiar with. Also don't mind removing this if you think that reasoning is self-evident and the explanation here is just additional baggage.

PR #3033, opened 2018-06-18
OrbitControl Update

### LINKED ISSUE #2497: WEBGL : make orbitControl() smoother and more continuous
#### Nature of issue?

- [ ] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

I think `orbitControl()` function in WEBGL mode should behave more like (or exactly like..) [this](https://threejs.org/examples/?q=text#webgl_buffergeometry_instancing2) sketch. As of right now, when user releases button after click, the camera position goes back to previous position instantly, which doesn't feel like desired behaviour.

AidanNelson:
Fixes #2497.

Adds persistence to `orbitControl()` and maintains y-axis up orientation. 

This change does the following:
- uses spherical coordinates to establish current camera position w/r/t center of sketch, adjust that position based on mouse movement and convert back to cartesian coordinates,
- makes adjustment persistent by applying those coordinates to rendererGL's `camera()` function which changes cameraMatrix,
- adds `cameraCenterX/Y/Z` properties to rendererGL to allow orbitControl about a center not at the origin.

### DIFF
--- a/src/webgl/camera.js
+++ b/src/webgl/camera.js
@@ -81,6 +81,10 @@ p5.RendererGL.prototype.camera = function(
   this.cameraY = eyeY;
   this.cameraZ = eyeZ;
 
+  this.cameraCenterX = centerX;
+  this.cameraCenterY = centerY;
+  this.cameraCenterZ = centerZ;
+
   // calculate camera Z vector
   var z0 = eyeX - centerX;
   var z1 = eyeY - centerY;
--- a/src/webgl/interaction.js
+++ b/src/webgl/interaction.js
@@ -49,8 +49,9 @@ p5.prototype.orbitControl = function(sensitivityX, sensitivityY) {
   }
 
   if (this.mouseIsPressed) {
-    var deltaTheta = 0.02 * sensitivityX * (this.mouseX - this.pmouseX);
-    var deltaPhi = 0.02 * sensitivityY * (this.mouseY - this.pmouseY);
+    var scaleFactor = this.height < this.width ? this.height : this.width;
+    var deltaTheta = sensitivityX * (this.mouseX - this.pmouseX) / scaleFactor;
+    var deltaPhi = sensitivityY * (this.mouseY - this.pmouseY) / scaleFactor;
 
     // camera position
     var camX = this._renderer.cameraX;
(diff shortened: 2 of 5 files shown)


kjhollen:
thanks, Aidan! the default sensitivity here feels a little high to me (I can move through the entire range of motion just by moving the mouse a few pixels)—is that the expected behavior?

AidanNelson:
I may have become used to it being a bit twitchy!  How does it feel now?  I lowered default sensitivity to 1/5th of what it was.

kjhollen:
It's better but I think the older behavior (and the relationship of the sensitivity factor) was easier to understand. I think you could still do something like:

```
var deltaPhi = sensitivityY * (this.mouseY - this.pmouseY) / (this.width / 2);
```

which would be closer to the old behavior, which tracks the mouse at nearly 1:1. If it's weird having both x & y depending on the width, we could opt for the smaller of the two, so that it's possible to get the full range of rotation in both directions, even if the aspect ratio is a little funky. What do you think?

AidanNelson:
Yes, it would be good to have a width/height dependent scale!  I'm not entirely sure why I changed that in the first place...  I added the smaller or width / height as the scale factor in this latest commit, which should allow at least 1 radian / ~60 degrees of rotation in each direction.  Does that feel alright? 

Also, while using the Chrome Device Toolbar on a touch device setting that orbitControl was jumping between positions, presumably because pmouseX/Y wasn't being updated except during touch events.   I will look into a workaround, unless resetting pmouseX/Y to current mouseX/Y at the beginning of each touchEvent might be more expected behavior with regard to touch events?   That way pmouseX/Y would either store the previous frame's mouseX/Y OR the current mouseX/Y, but never the mouseX/Y of 100, 200, etc. frames ago.

mlarghydracept:
@AidanNelson This is super nitpicky but since you are already changing the manual test - can you get rid of the HTML that tells me to "press mouse to toggle the world"? It bothers me an abnormal amount. 

In more helpful news, I just tried out this persistent orbit control on a few different sketches and got satisfying results in all of them. Excited for panning and zooming :-P

AidanNelson:
Yes, done!  And I will start a separate thread about whether/how pmouseX/Y should be updated with touch events.

kjhollen:
This looks good to me, Aidan! Feels much better. It looks like the update accidentally undoes the fix for #2930—if you want to patch that back in, I think then we'll be ready for merge.

AidanNelson:
Ok.  Neither axis should be inverted now!

kjhollen:
Awesome! Thank you!

PR #3042, opened 2018-06-25
Arc bug fix

### LINKED ISSUE #2919: arc renders unexpectedly when start and stop angle are the same
I noticed some weird behavior in the arc function when the start and stop angles are the same. It will render a full circle from 0 up to 90 degrees, beyond that it begins to render a 180degree~ arc that rotates, and then around degree 300 the arc starts to flash on and off.

What is the intended behavior when the values are the same? I was imagining that nothing would be drawn. I'm wondering if this line is causing some of the weirdness https://github.com/processing/p5.js/blob/5f6b4ab7e1c1a680186e52a9a1ac3f63450e8406/src/core/2d_primitives.js#L100-L103 

#### Nature of issue?

- [X] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

- [X] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.6.1
- Web browser and version: Firefox 60
- Operating System: Linux Ubuntu 16.04
- Steps to reproduce this: Set the same value for start and stop in arc(). Here's an example: https://alpha.editor.p5js.org/aferriss/sketches/SkpGgyaCf

AdilRabbani:
Arc seems to flash a rect on the screen when rotate() is called with intervals 270 (-90, 630, etc). This fixes the issue. Closes #2919


mlarghydracept:
Hi @AdilRabbani! Thanks for jumping on this. Could you explain why this number works? It feels natural to divide width by 2 to get radius, but 2.00000001 feels like a "magic number" here. You would also need to include a comment explaining the logic.

AdilRabbani:
Hey @mlarghydracept! Upon debugging I noticed that this only happens when both the translate and rotate are called at the same time in http://alpha.editor.p5js.org/aferriss/sketches/rJ9iyfi17 I'm still not confirm on this one, but I think that translate(width/2, height/2) and the rx, ry being calculated seemed to cause the problem. Then I started changing the value for rx,ry and it worked for values other than 2.0. I decided to make it close to 2 but still fixing the issue. But yes, you are right this feels more like a hack and not a proper fix.

Issue #3045, opened 2018-06-28
p5.Table.removeColumn() logs 'string' without obvious reason


oxxie:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [x] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 0.6.1
- Web browser and version: Safari 11.1.1
- Operating System: macOS 10.13.5
- Steps to reproduce this:
```
let data;

function preload() {
  data = loadTable(fileName, "header");
}

function setup() {
  data.removeColumn(columnId);
  console.log(data);
}
```
The code above will cause the string 'string' to be logged to the console, without obvious reason. Looks like an inadvertent debugging session left-over?

oxxie:
Issued #3046 to get rid of this nuisance :-)

oxxie:
#3046 has been merged. This issue closed.

PR #3046, opened 2018-06-28
Removed unnecessary `console.log()` from `removeColumn()`

oxxie:
Also see #3045.


kjhollen:
thanks @oxxie! :tada:

Issue #2188, opened 2017-09-21
textureMode() or similar for pixel-based texture coordinates


kjhollen:
#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Feature enhancement details:

Processing has a [textureMode](https://processing.org/reference/textureMode_.html) feature for using image pixel coordinates for u,v texture coordinates instead of the normal 0-1 convention. adding this feature to p5 would make some texturing operations easier for beginners.

this is a small to medium size project for someone with a little experience working with open gl. see the [WebGL Module Architecture](https://github.com/processing/p5.js/wiki/WebGL-Module-Architecture) page on p5 conventions for webgl.

Siphonophora:
A note from looking into this a bit, the documentation for vertex is missing u and v parameters which are supported in the code.

aferriss:
I'm interested in taking a look at this one if nobody else has started with it. Is this a webGL only feature? I know that texture only works in webGL, but the 2d renderer has uv options for vertex(). What purpose do the uv parameters of vertex() serve in non webGL mode?

Spongman:
is the support for vertex texture coordinates that i implemented in my rendering pipeline changes going to be merged?

PR #3056, opened 2018-07-04
Implement textureMode

aferriss:
This PR implements the textureMode() function and mirrors the functionality of the [original processing api](https://processing.org/reference/textureMode_.html).

The renderer is updated with a textureImage and a textureMode property. textureImage stores the most recently used image when texture() is called. 

Calls to vertex() will now check for textureMode and divide uv's by textureImage.width and height if textureMode(IMAGE) is set. 

This also adds inline examples, unit tests, and an example in the tests folder demonstrating the use of textureMode()

Hopefully this will close https://github.com/processing/p5.js/issues/2188 

Also I think this may introduce some duplicate changes as https://github.com/processing/p5.js/pull/2961 so it might make sense to wait until that is merged to add this one.


src/webgl/material.js
```diff
@@ -298,6 +299,80 @@ p5.prototype.texture = function(tex) {
...
+ *   textureMode(NORMAL);
+ *   beginShape();
+ *   vertex(-50, -50, 0, 0);
+ *   vertex(50, -50, img.width, 0);
+ *   vertex(50, 50, img.width, img.height);
+ *   vertex(-50, 50, 0, img.height);
+ *   endShape();
+ * }
+ * </code>
+ * </div>
```
kjhollen: these examples need an accessible description with the @alt tag after each one

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -98,6 +98,18 @@ p5.RendererGL.prototype.vertex = function(x, y) {
     vertexColor[3]
   );
 
+  if (this.textureMode === constants.IMAGE) {
+    if (this.textureImage !== undefined) {
+      u /= this.textureImage.width;
+      v /= this.textureImage.height;
```
kjhollen: do we need to be worried about divide by 0 here? or does something sensible enough happen / does some reasonable error appear in the console?

aferriss: I think divide by 0 returns the value Infinity. I can add in an additional check to make sure the width and height are > 0. Though I think if you have an image with a width and height of 0 you might have much bigger problems on your hands 😛

kjhollen: totally, haha. I'm not sure what the right behavior is here; in some other places in the GL pipeline we do check if the divisor is 0 before running the division, though there's no error message printed out directly if that happens (see `camera()` for example)—so I'm not sure that feels right either! I wonder if it makes sense to come up with some guidelines here to make sure there's consistent behavior across places...

lmccart:
@kjhollen looks like the alt text requested has been added, is this good to merge?

kjhollen:
@lmccart this one has some conflicts with another open PR (#3076) as both define a field for texture with a different name. I was hoping to merge the other first, but this one is probably closer to ready so we can take this one first and I'll update the other.

lmccart:
thanks @aferriss! and @kjhollen for reviewing.

PR #3074, opened 2018-07-09
Change model() example

TanviKumar:
The current `model()` example has the same error the previous `loadModel()` example had. This PR fixes that issue by using the octahedron example instead.


kjhollen:
@almchung I think it's ok to just have one example here, assuming this links to loadModel anyway!

PR #3075, opened 2018-07-09
Fixes wrong triangle stroke

### LINKED ISSUE #3065: Wrong stroke rendering with beginShape and large strokeWeight
#### Nature of issue?

- [x] Found a bug

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: http://alpha.editor.p5js.org/carmolim/sketches/HyfkoNCzX
- Web browser and version: Chrome  67.0.3396.99
- Operating System: MacOS 10.13.5 (17F77)
- Steps to reproduce this:
http://alpha.editor.p5js.org/carmolim/sketches/HyfkoNCzX
```
function setup() {
  
  createCanvas(400, 400);
  background(0);
  
  noFill();
  stroke(255);
  strokeWeight(10);
  
  scale(3);
  
  beginShape(TRIANGLES);
    vertex(30, 75);
    vertex(50, 20);
    vertex(80, 75);
	endShape();
}

function draw() {
}
```

Bug

AdilRabbani:
Fixes wrong triangle stroke with beginShape(). Closes #3065

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -867,7 +867,7 @@ p5.Renderer2D.prototype.endShape = function(
         this.drawingContext.moveTo(v[0], v[1]);
         this.drawingContext.lineTo(vertices[i + 1][0], vertices[i + 1][1]);
         this.drawingContext.lineTo(vertices[i + 2][0], vertices[i + 2][1]);
-        this.drawingContext.lineTo(v[0], v[1]);
+        this.drawingContext.closePath();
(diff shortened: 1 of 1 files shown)


mlarghydracept:
I tested this quite a bit already and it has worked well in all the situations that I have used it in. Pretty sure this can is ready to be merged.

Though while I was testing this I looked into the manual tests for p5.Shape and noticed that there is only one manual test which is not functioning. Not as a result of this patch but just because it is incomplete or empty.

PR #3076, opened 2018-07-09
update gl pipeline to set shader uniforms at render time

kjhollen:
p5.RendererGL is updated to hold more state so that shader uniforms
can be set later, when it's time to render a shape. then, a shader
does not have to be bound at all time to change state. this enables
things like full support for push/pop.

in particular, p5.RendererGL state is added for tracking:

- whether or not the normal shader is in use
- use of lighting (boolean)
- use of specular component (boolean)

the curFillShader, curStrokeShader, curPointShader objects are removed from
p5.RendererGL in favor of selecting a default shader at render time,
based on current gl state (lighting conditions, texture settings,
and fill/stroke). additionally, _userFillShader and _userStrokeShader
are introduced to indicate that a custom shader has been set by the
library user using a call to shader(). in this case, the user shader
takes priority over any default shaders. this enables the user to mix
and match calls to fill, stroke, lights, etc. to set uniforms in
their own shader.

adds p5.RendererGL._setFillUniforms and
p5.RendererGL._setStrokeUniforms to set all of the known default
uniforms for the shader, to be used just before shape render time.

unit tests are updated to reflect this new structure; still heavily
dependent on internal state but better than nothing.


kjhollen:
@mlarghydracept I think it's all good with the camera. there was a conflict in the p5.Shader class that I had to resolve, but @AidanNelson had already updated the properties in push() to include the camera object!

the main changes are for handling points, yes.

lmccart:
@mlarghydracept @kjhollen what's the status on this one?

kjhollen:
I need to update to resolve merge conflicts, but otherwise we should be ready. Sorry for lagging behind on this one. Moving/starting school got me way behind on outside projects. :( Should be able to do this after Christmas.

stalgiag:
Hi @kjhollen I am sure you are busy but I wanted to get your thoughts on the status of this PR. I think that it or something like it needs to be merged. It also has a certain weight on future contributions to the GL portion of the library so if we are going to make this update I think the timing is a bit crucial. I understand if you are too busy to handle the updating/resolving merge conflicts for this PR, and I can take a crack at taking it over. Let me know if you have any thoughts on timing, your capacity to update this or desire hand it off, potential problems, etc. Thanks!

Spongman:
i have rebased this branch onto master and fixed the merge conflicts: https://github.com/Spongman/p5.js/tree/updated-delay-uniforms

stalgiag:
Hi @Spongman great! It looks like updates were minimal. I think we can open a new PR that references this one and I can help look over all the changes one more time.

stalgiag:
Closing as this was continued in #3535

Issue #2533, opened 2018-01-11
weblg: rotate[XYZ] doesn't take angleMode into account


Spongman:
as noted in #2524, currently the 3D rotate functions don't take `angleMode` into account when calculating the angle.

````
angleMode(DEGREES);
rotateZ(45);
rect(-100, -100, 200, 200);
````

this leads to an inconsistency with the way the 2d `rotate` function works.

Spongman:
this was fixed with #2552

Spongman:
turns out this problem still exists.

lmccart:
should be fixed now with https://github.com/processing/p5.js/pull/3077

Issue #2965, opened 2018-06-01
Less rotation than expected with angleMode(DEGREES) in WEBGL mode


AidanNelson:
#### Nature of issue?

- [X] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version:  0.6.1
- Web browser and version:  Chrome 66.0.3359.181
- Operating System: Mac OSX 10.12.6
- Steps to reproduce this:

It seems that rotateX,Y,Z produce less rotation than expected when angleMode is set to DEGREES in WEBGL mode.  See the below sketch with 2D and 3D rotations.  2D mode exhibits expected behavior (degree and radian modes produce same rotation when scaled appropriately) whereas in WEBGL mode, the boxes rotate less in ```angleMode(DEGREES)```.  I briefly looked into this using the Chrome debugger, but didn't see anything super obvious.  I will keep looking!

```
function setup() {
  // createCanvas(400, 400);
	createCanvas(400, 400,WEBGL);
}

function draw() {
  background(200);
	
	let currentTime = millis();
	
// 	push();
// 	translate(width/2,height/4);
// 	angleMode(DEGREES);
// 	rotate((currentTime / 1000) * 360);
// 	fill(0,0,255);
// 	rectMode(CENTER);
// 	rect(0,0,100,50);
// 	pop();
	
// 	push();
// 	translate(width/2,height/4*3);
// 	angleMode(RADIANS);
// 	rotate((currentTime / 1000) * TWO_PI);
// 	fill(255,0,0);
// 	rectMode(CENTER);
// 	rect(0,0,50,100);
// 	pop();
	
	push();
	translate(0,-100,0);
	angleMode(DEGREES);
	rotateZ((currentTime / 1000) * 360);
	fill(0,0,255);
	box(100,50);
	pop();
	
	push();
	translate(0,50,0);
	angleMode(RADIANS);
	rotateZ((currentTime / 1000) * TWO_PI);
[...]

PR #3077, opened 2018-07-09
fix rotation in DEGREES mode

Spongman:
the (private) matrix methods are _only_ passed radians by p5, regardless of the `angleMode`. this PR fixes the double-conversion that occurs when using `DEGREES` mode.


Spongman:
closes #2533 - this particular issue was not correctly fixed.

kjhollen:
Hi @Spongman is this related to #2965? Do they fix the same or separate issues?

Spongman:
i guess so. it looks like @meiamsome brought up some issues regarding re-arranging the API. this PR is just the simple one-liner that fixes the issue without changing the API.

Issue #1967, opened 2017-05-26
p5.Element.mouseDragged not a function


limzykenneth:
I'm trying to attach `mouseDragged` event on a p5.Element, specifically an instance mode canvas. The reason I want to attach it to the element and not just use the global function or `p5.Element.mouseMoved` is because I have multiple canvases and I don't want event from one to leak over to the other, I will be dragging on the second canvas as well but I don't want the first canvas to fire its `mouseDragged` function.

First of all, is this actually a non-issue and I'm missing something obvious?

Second, I saw #752 asking about this but not much reason given why mouseDragged is not in p5.Element, so would it be possible at all to implement?

P.S. I thought I might mention this has nothing to do with the latest bug with mouseDragged in 0.5.10

lmccart:
hmm I vaguely recall there being some reason there's no mouseDragged method for p5.Element but it's hard to remember what it was. I'm open to adding this though, if possible.

limzykenneth:
I guess it is just chaining `mousePressed` and `mouseReleased`. There are several cases to consider: releasing the mouse outside the canvas; pressing outside then dragging in the canvas; pressing and releasing outside the canvas but still dragging in it. Not sure about the possibility of implementing them.

I have a project that may require this but at the moment I just used some hack to get around it. Currently still working on said project and couldn't work on this, it would be great if someone can have a closer look at this. 😜

lmccart:
thinking about it more... as I recall, the main problem was that you can't necessarily detect crossing into or out of an element while mouse is down. this confusion might cause the function to feel more broken than correct. that is, if you expect it to only fire when mouse is dragged over an element, then it would feel broken to not fire when you start outside and drag in, or you start inside and drag out.

limzykenneth:
@lmccart I'm hacking something together for my project with p5.element version of `mousePressed`, `mouseMoved` and `mouseReleased`, for the most part it works as I expected, mouse dragging is only handle when the mouse is pressed in said element; mouse moving (while pressed) will stop updating if the cursor went outside the element but will resume correctly when moving back which I think is reasonable behaviour.

The larger problem lies with releasing outside the element. The element, as you pointed out indirectly, will not know the mouse is released outside the element, as such I need to have a global mouse release listener to clean up the whole thing. The whole process will need to keep a state variable which simply keep track of whether it is in the "drag" state (mouse pressed in the element and not released anywhere yet).

[...]

PR #3082, opened 2018-07-13
Unlink touch/mouse events firing to fix #1875

### LINKED ISSUE #1875: <Select> Elements not responsive on mobile device browsers
After some investigation, and @JimishF pointing out that this is unrelated to #1864 and #1868, here's a whole new issue: select elements created using createSelect() are unresponsive on mobile browsers. I've tested this on:

* PhoneGap/Cordova
* Chrome on Android
* Safari on iOS 

To replicate this in Chrome or Safari:

1. Create the sketch below
2. Run a local server in the sketch directory: `$ python -m SimpleHTTPServer` works for me on OSX.
3. Open the page in either browser.
4. Tap the select element

You should get no response from the select, but you can add elements to it using the button (just proving that it really is unrelated to #1864 and #1868) 

The sketch:

```
var responseDiv, opts, addButton;
function setup() {
  console.log('starting');
  opts = createSelect();
  opts.position(10, 10);
  opts.size(150, 30)
  opts.option('Pick a device', random(34));
  opts.option('fee', random(34));
  opts.option('fie', random(34));
  opts.option('fo', random(34));
  opts.changed(fum);

  responseDiv = createDiv('tap the scan button to begin');
  responseDiv.position(10, 100);
  addButton = createButton('add an option');
  addButton.position(10, 40);
  addButton.touchEnded(addSomething);
}


function fum(event) {
  console.log(event.target);
  responseDiv.html(opts.value());
}

function addSomething() {
  var number = random(34);
  opts.option('another'+ number, number);
}

```

To test in PhoneGap:

1. Follow [Jiashan's instructions](https://github.com/OhJia/p5Mobile/wiki/Phonegap) to make a p5 PhoneGap project.
2. Add p5.dom.js to the list of script includes in the index.html as well.
3. run the app on your phone. Should see the same results.

To test in Chrome on OSX:

1. Open page locally
2. Open JavaScript Console
3. Click "Toggle Device Toolbar (shift-command-M) in JS console
4. Pick any of the Android or iOS devices
You should see the same results.

Strangely, I was able to get <select> items to work with an older version of p5, but it was a version where touch events were erroneously sent twice. I will attempt to replicate, haven't managed to so far. It was pre-0.5.3.

I'm cross-posting this on Jiashan's p5Mobile repo just in case it belongs there instead, but it seems general enough that it's worth solving for p5 in general.

kjhollen:
- core/mouse.js: removes touch events from firing in mouse.js events
- core/touch.js: removes mouse events from firing in touch.js events
- core/p5.Element.js: removes touch events from firing in mouse
  functions and mouse events from firing in touch-related functions

to alleviate double-firing of mouse & touch events when working
with p5.dom elements. now, library users should favor using
mouse events for p5.Elements instead of touch (modern mobile
devices will link these automatically -- but using the mouse
events ensures they will happen on a desktop machine for testing,
too). the exception is if the user would like to track touch
movement on a mobile or tablet device: then touchMoved should be
used.

touch events will no longer fire on desktop browsers, unless
using dev tools to simulate mobile devices.

relevant documentation for mouse and touch events is updated
to more accurately describe the new behavior.

the p5.Element case is a little strange because there is no passing of touchMoved to mouseDragged, because there is no mouseDragged in p5.Element (see #1967). moreover, whether or not a mouse event is triggered by touch seems a little inconsistent. I can make a mouseMoved event appear if I long-tap (contact with no movement) on the device, but only a single mouseMoved. a quick tap can trigger 4 different mouse events if they are all registered: mouseMoved, mousePressed, mouseReleased, and mouseClicked if all are registered. I feel like this information is valuable but couldn't figure out how to attach it to the p5.Element events; I am open to suggestions on this!

### DIFF
--- a/src/core/p5.Element.js
+++ b/src/core/p5.Element.js
@@ -231,7 +231,6 @@ p5.Element.prototype.class = function(c) {
  */
 p5.Element.prototype.mousePressed = function(fxn) {
   adjustListener('mousedown', fxn, this);
-  adjustListener('touchstart', fxn, this);
   return this;
 };
 
@@ -399,7 +398,6 @@ p5.Element.prototype.mouseWheel = function(fxn) {
  */
 p5.Element.prototype.mouseReleased = function(fxn) {
   adjustListener('mouseup', fxn, this);
-  adjustListener('touchend', fxn, this);
   return this;
 };
 
@@ -512,7 +510,6 @@ p5.Element.prototype.mouseClicked = function(fxn) {
  */
 p5.Element.prototype.mouseMoved = function(fxn) {
   adjustListener('mousemove', fxn, this);
-  adjustListener('touchmove', fxn, this);
   return this;
 };
 
@@ -748,7 +745,6 @@ p5.Element.prototype.mouseOut = function(fxn) {
  */
 p5.Element.prototype.touchStarted = function(fxn) {
   adjustListener('touchstart', fxn, this);
-  adjustListener('mousedown', fxn, this);
   return this;
 };
 
@@ -789,7 +785,6 @@ p5.Element.prototype.touchStarted = function(fxn) {
  */
 p5.Element.prototype.touchMoved = function(fxn) {
   adjustListener('touchmove', fxn, this);
-  adjustListener('mousemove', fxn, this);
   return this;
 };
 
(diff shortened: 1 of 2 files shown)


kjhollen:
hmm, something's not right in my git history. stand by...

PR #3080, opened 2018-07-10
p5.Camera Reorganization

AidanNelson:
Rebased p5.Camera Reorganization from #3063.  

Updates:
* several documentation updates,
* adds an `_isActive()` helper method to p5.Camera prototype, called before each update to the renderer in ortho(), perspective(), or camera() to support multiple cameras,
* fixes resize issue: current behavior has default cameras updating both eyeZ and aspectRatio during resize, whereas custom cameras only update aspectRatio,
* removes `return this;` from constructor.

PR #3083, opened 2018-07-13
p5.Camera unit tests

AidanNelson:
This adds unit tests for the p5.Camera object as added in #3080.

I wasn't quite sure how to create tests for projection matrices without simply recreating the ortho() and perspective() functions in the test, so any feedback on that would be appreciated!

### DIFF
--- a/test/unit/webgl/p5.Camera.js
+++ b/test/unit/webgl/p5.Camera.js
@@ -3,17 +3,6 @@ suite('p5.Camera', function() {
   var myCam;
   var delta = 0.001;
 
-  // copy of p5.Vector.prototype.angleBetween...
-  var angleBetween = function(a1, a2, a3, b1, b2, b3) {
-    var dotProd = a1 * b1 + a2 * b2 + a3 * b3;
-    var dotmagmag =
-      dotProd /
-      (Math.sqrt(a1 * a1 + a2 * a2 + a3 * a3) *
-        Math.sqrt(b1 * b1 + b2 * b2 + b3 * b3));
-    var angle = Math.acos(Math.min(1, Math.max(-1, dotmagmag)));
-    return angle;
-  };
-
   // returns values to test which have changed
   var getVals = function(cam) {
     return {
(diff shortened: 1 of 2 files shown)


kjhollen:
hey @AidanNelson, instead of replicating the ortho/perspective/etc functions in the unit tests, what you'll do instead is determine the expected values for each entry of the matrix and hard code them. in theory, the 'expected' values are generally known ahead of time and not computed—things that are easily unit-testable have predictable results.

this feels hard for me to explain, somehow—if this doesn't make sense, we can set up some time on google hangouts.

test/unit/webgl/p5.Camera.js
```diff
@@ -0,0 +1,208 @@
...
+  suite('Projection', function() {
+    test('ortho() sets renderer uPMatrix', function() {
+      myCam.ortho(-10, 10, -10, 10, 0, 100);
+
+      for (let i = 0; i < myCam.projMatrix.mat4.length; i++) {
+        assert.strictEqual(
+          myCam.projMatrix.mat4[i],
+          myp5._renderer.uPMatrix.mat4[i]
+        );
+      }
```
kjhollen: in the same test, we should check if the expected values are in the matrix with additional assertions. you can add an argument to the assert call to print a more helpful error so it's obvious if the error was that the matrices didn't match or that they didn't have the expected values (applies to next test, too)

test/unit/webgl/p5.Camera.js
```diff
@@ -0,0 +1,208 @@
...
+    });
+
+    test('perspective() sets projection matrix correctly', function() {
+      //prettier-ignore
+      var control = [1,  0,   0,  0,
+                     0, -1,   0,  0,
+                     0,  0,  -3, -1,
+                     0,  0, -40,  0];
+
```
kjhollen: cool. another good value to test is the default! you may want to write a helper function that iterates over two arrays or matrices to check their values—or maybe there's already a chai function that already handles arrays?

test/unit/webgl/p5.Camera.js
```diff
@@ -0,0 +1,208 @@
...
+        assert.strictEqual(myCam.projMatrix.mat4[i], control[i]);
+      }
+    });
+  });
+
+  suite('Helper Functions', function() {
+    test('copy() returns a new p5.Camera object', function() {
+      var newCam = myCam.copy();
+      assert.instanceOf(newCam, p5.Camera);
+      assert.notDeepEqual(newCam, myCam);
```
kjhollen: should check that the values for both are the same here, to ensure that the copy worked

test/unit/webgl/p5.Camera.js
```diff
@@ -0,0 +1,208 @@
...
+    });
+
+    test('_getLocalAxes() returns three vectors', function() {
+      var local = myCam._getLocalAxes();
+      for (let j = 0; j < 3; j++) {
+        assert.typeOf(local.x[j], 'number');
+        assert.typeOf(local.y[j], 'number');
+        assert.typeOf(local.z[j], 'number');
```
kjhollen: great. are there any other conditions that should be true here? should the axes be normalized? should they be orthogonal?

AidanNelson:
Ok, @kjhollen I added several unit tests and updated them throughout to ...
* set camera defaults in the test-suite setup, in case these default values ever change in the library,
* use a hardcoded camera or projection matrix as the expected value,
* use chai's `deepEqual` assertion to iterate over matrix values,
* also sometime also check eyeXYZ, or upXYZ so that error messages would be more helpful.

test/unit/webgl/p5.Camera.js
```diff
@@ -0,0 +1,208 @@
...
+      for (let i = 0; i < myCam.projMatrix.mat4.length; i++) {
+        assert.strictEqual(
+          myCam.projMatrix.mat4[i],
+          myp5._renderer.uPMatrix.mat4[i]
+        );
+      }
+    });
+
+    test('perspective() sets renderer uPMatrix', function() {
+      myCam.perspective(Math.PI / 3.0, 1, 1, 100);
```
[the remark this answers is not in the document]
AidanNelson: Yes, but the view will be reversed!  I wasn't sure whether / how to check for this (and a few other) edge cases which produce valid matrices but unexpected views.

test/unit/webgl/p5.Camera.js
AidanNelson: Yes, those checks are added.

test/unit/webgl/p5.Camera.js
AidanNelson: Added a test for this.

test/unit/webgl/p5.Camera.js
AidanNelson: Yes, switched to using chai's deepEqual assertion.

test/unit/webgl/p5.Camera.js
```diff
@@ -0,0 +1,540 @@
+suite('p5.Camera', function() {
+  var myp5;
+  var myCam;
+  var delta = 0.001;
+
+  // copy of p5.Vector.prototype.angleBetween...
+  var angleBetween = function(a1, a2, a3, b1, b2, b3) {
```
kjhollen: hmm, I think it ought to be ok to just use the vector version (and make some vectors) when you need this? unless I'm missing something?

AidanNelson: Ok.  I was trying to keep it isolated, but not for any particular reason!

AidanNelson:
Ok!  Once #3088 is merged, I could add a unit test for orbitControl using the `p5.Camera.prototype._orbit()` method directly (without needing mouse input).

kjhollen:
cool! a nice bonus is that using p5.Vector made the test more legible too :)

Issue #2187, opened 2017-09-20
Implement quadraticVertex for webgl mode


kjhollen:
#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Feature enhancement details:

curveVertex() only works in 2D mode currently, but it would be great for WEBGL mode, too.

this could be a bigger task because bezierVertex (see #2185) and curveVertex (see #2186) need to be implemented, too—so all of these could share some common structure. see: [WebGL Module Architecture](https://github.com/processing/p5.js/wiki/WebGL-Module-Architecture) for relevant documentation.

PR #3085, opened 2018-07-14
Webgl curves implementation

### LINKED ISSUE #2185: Implement bezierVertex for webgl mode
#### Nature of issue?

- [ ] Found a bug
- [x] New feature request

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

bezierVertex() only works in 2D mode currently, but it would be great for WEBGL mode, too.

this could be a bigger task because curveVertex and quadraticVertex need to be implemented, too—so all of these could share some common structure. see: [WebGL Module Architecture](https://github.com/processing/p5.js/wiki/WebGL-Module-Architecture) for relevant documentation.

AdilRabbani:
Implements bezierVertex(), quadraticVertex(), curveVertex() for WebGL mode. Uses [libtess](https://github.com/brendankenny/libtess.js/) to triangulate filled curves. Maintains a look up table to store coefficients required to draw curves. This helps in speeding up things (the coefficients aren't needed to be calculated the next time a specific curve is used). However, the coefficients are calculated again if the curveDetail() is changed.
Closes #2185, #2186, #2187.

### DIFF
--- a/src/core/vertex.js
+++ b/src/core/vertex.js
@@ -284,8 +284,8 @@ p5.prototype.beginShape = function(kind) {
  * @method bezierVertex
  * @param  {Number} x2 x-coordinate for the first control point
  * @param  {Number} y2 y-coordinate for the first control point
- * @param  {Number} x3 x-coordinate for the first control point
- * @param  {Number} y3 y-coordinate for the first control point
+ * @param  {Number} x3 x-coordinate for the second control point
+ * @param  {Number} y3 y-coordinate for the second control point
  * @param  {Number} x4 x-coordinate for the anchor point
  * @param  {Number} y4 y-coordinate for the anchor point
  * @chainable
@@ -327,6 +327,41 @@ p5.prototype.beginShape = function(kind) {
  * @param  {Number} y4
  * @param  {Number} [z4] z-coordinate for the anchor point (for WebGL mode)
  * @chainable
+ * @example
+ * <div>
+ * <code>
+ * function setup() {
+ *   createCanvas(100, 100, WEBGL);
+ *   setAttributes('antialias', true);
+ * }
+ * function draw() {
+ *   orbitControl();
+ *   background(50);
+ *   strokeWeight(4);
+ *   stroke(255);
+ *   point(-25, 30);
+ *   point(25, 30);
+ *   point(25, -30);
+ *   point(-25, -30);
+ *
+ *   strokeWeight(1);
+ *   noFill();
+ *
+ *   beginShape();
+ *   vertex(-25, 30);
+ *   bezierVertex(25, 30, 25, -30, -25, -30);
+ *   endShape();
+ *
+ *   beginShape();
+ *   vertex(-25, 30, 20);
+ *   bezierVertex(25, 30, 20, 25, -30, 20, -25, -30, 20);
+ *   endShape();
+ * }
+ * </code>
+ * </div>
+ *
+ * @alt
+ * crescent shape in middle of canvas with another crescent shape on positive z-axis.
  */
 
 p5.prototype.bezierVertex = function() {
--- a/src/core/vertex.js
+++ b/src/core/vertex.js
@@ -269,60 +269,42 @@ p5.prototype.beginShape = function(kind) {
 
 /**
  * Specifies vertex coordinates for Bezier curves. Each call to
- * <a href="#/p5/bezierVertex">bezierVertex()</a> defines the position of two control points and
+ * bezierVertex() defines the position of two control points and
  * one anchor point of a Bezier curve, adding a new segment to a
- * line or shape.
+ * line or shape. For WebGL mode bezierVertex() can be used in 2D
+ * as well as 3D mode. 2D mode expects 6 parameters, while 3D mode
+ * expects 9 parameters (including z coordinates).
  * <br><br>
- * The first time <a href="#/p5/bezierVertex">bezierVertex()</a> is used within a
- * <a href="#/p5/beginShape">beginShape()</a> call, it must be prefaced with a call to <a href="#/p5/vertex">vertex()</a>
- * to set the first anchor point. This function must be used between
- * <a href="#/p5/beginShape">beginShape()</a> and <a href="#/p5/endShape">endShape()</a> and only when there is no MODE
- * parameter specified to <a href="#/p5/beginShape">beginShape()</a>.
+ * The first time bezierVertex() is used within a beginShape() call,
+ * it must be prefaced with a call to vertex() to set the first anchor
+ * point. This function must be used between beginShape() and endShape()
+ * and only when there is no MODE or POINTS parameter specified to
+ * beginShape().
  *
  * @method bezierVertex
  * @param  {Number} x2 x-coordinate for the first control point
  * @param  {Number} y2 y-coordinate for the first control point
- * @param  {Number} x3 x-coordinate for the second control point
- * @param  {Number} y3 y-coordinate for the second control point
+ * @param  {Number} x3 x-coordinate for the first control point
+ * @param  {Number} y3 y-coordinate for the first control point
  * @param  {Number} x4 x-coordinate for the anchor point
  * @param  {Number} y4 y-coordinate for the anchor point
  * @chainable
  * @example
  * <div>
  * <code>
- * strokeWeight(5);
- * point(30, 20);
- * point(80, 20);
- * point(80, 75);
- * point(30, 75);
- *
- * strokeWeight(1);
  * noFill();
  * beginShape();
  * vertex(30, 20);
- * bezierVertex(80, 20, 80, 75, 30, 75);
+ * bezierVertex(80, 0, 80, 75, 30, 75);
  * endShape();
  * </code>
  * </div>
  *
  * <div>
  * <code>
- * strokeWeight(5);
- * point(30, 20);
- * point(80, 20);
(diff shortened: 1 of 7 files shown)


src/webgl/primitives.js
```diff
@@ -1203,4 +1203,323 @@ p5.RendererGL.prototype.line = function() {
...
+    var w_x = [];
+    var w_y = [];
+    var w_z = [];
+    var t, _x, _y, _z, i;
+
+    t = parseFloat(0.0);
+
+    if (
+      this._lookUpTableQuadratic.length === 0 ||
+      this._lookUpTableQuadratic[this._lookUpTableQuadratic.length - 1] !==
```
Spongman: will this ever be true?

AdilRabbani: True only when the first time quadraticVertex() is called. Also True when curveDetail() for which the lookUpTable was calculated the last time, isn't equal to the curveDetail() set afterwards.

So, in calls like these,

quadraticVertex()           // True
quadraticVertex()           // False
quadraticVertex()           // False

But for something like this,

quadraticVertex()            // True
quadraticVertex()            // False
curveDetail(5)                 
quadraticVertex()            // True
quadraticVertex()            // False

Spongman: oh, i see now. i was confused because it looked to me like _lookUpTableQuadratic/_lookUpTableBezier were arrays of coefficients. now i see that not all the elements of the array are the same type.

AdilRabbani: Just the last element of the array....to store the current curveDetail 😁. Oh....Why not insert it into the first element? 😲

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -49,6 +49,11 @@ p5.RendererGL.prototype.beginShape = function(mode) {
     this.immediateMode.lineVertexBuffer = this.GL.createBuffer();
     this.immediateMode.lineNormalBuffer = this.GL.createBuffer();
     this.immediateMode.pointVertexBuffer = this.GL.createBuffer();
+    this.immediateMode._bezierVertex = [];
+    this.immediateMode._quadraticVertex = [];
+    this.immediateMode._curveVertex_x = [];
+    this.immediateMode._curveVertex_y = [];
+    this.immediateMode._curveVertex_z = [];
```
mlarghydracept: Why does curveVertex store x ,y ,& z separately?

src/webgl/p5.RendererGL.js
```diff
@@ -1113,4 +1122,93 @@ p5.prototype._assert3d = function(name) {
...
+
+  var tessy = new libtess.GluTesselator();
+  tessy.gluTessCallback(libtess.gluEnum.GLU_TESS_VERTEX_DATA, vertexCallback);
+  tessy.gluTessCallback(libtess.gluEnum.GLU_TESS_BEGIN, begincallback);
+  tessy.gluTessCallback(libtess.gluEnum.GLU_TESS_ERROR, errorcallback);
+  tessy.gluTessCallback(libtess.gluEnum.GLU_TESS_COMBINE, combinecallback);
+  tessy.gluTessCallback(libtess.gluEnum.GLU_TESS_EDGE_FLAG, edgeCallback);
+
+  return tessy;
+})();
```
mlarghydracept: If you want to have a tesselator instance that is always initialized and available that seems fine but this is written with different conventions than the rest of the code. Maybe just have a renderer level `tessy` var that is initialized with a private prototype on renderer construction? Even utility functions are written this way in the rest of the code so probably best to stick with it.

src/webgl/p5.RendererGL.js
```diff
@@ -1113,4 +1122,93 @@ p5.prototype._assert3d = function(name) {
...
+// function to calculate QuadraticVertex Coefficients
+p5.RendererGL.prototype._quadraticCoefficients = function(t) {
+  var t2 = t * t;
+  var mt = 1 - t;
+  var mt2 = mt * mt;
+  return [mt2, 2 * mt * t, t2];
+};
+
+// function to convert Bezier coordinates to Catmull Rom Splines
+p5.RendererGL.prototype.Bezier2Catmull = function(w) {
```
mlarghydracept: I would keep with the underscore camelCase convention here as in `_bezierToCatmull`

src/webgl/primitives.js
```diff
@@ -1203,4 +1203,318 @@ p5.RendererGL.prototype.line = function() {
...
+
+    t = parseFloat(0.0);
+
+    if (
+      this._lookUpTableBezier.length === 0 ||
+      this._lookUpTableBezier[0] !== this._pInst._curveDetail
+    ) {
+      this._lookUpTableBezier = [];
+      this._lookUpTableBezier.push(this._pInst._curveDetail);
+      var step = 1 / this._pInst._curveDetail;
```
mlarghydracept: What happens to this while loop if the user sets [_curveDetail](https://github.com/processing/p5.js/blob/master/src/core/shape/curves.js#L400) to 0?

src/webgl/primitives.js
Spongman: does the current curveDetail _need_ to go in the array? wouldn't it be easier just to use a separate scalar?

src/webgl/primitives.js
AdilRabbani: @mlarghydracept should there be a minimum curveDetail (3 perhaps) or it should render nothing when curveDetail is set to 0?

mlarghydracept: That seems like a good solution to me.

src/webgl/p5.RendererGL.Immediate.js
AdilRabbani: Yes I had that in my mind as well. But there are some problems using a single array.
1) I would have to give proper indices when I'm converting x,y and z coordinates using Bezier2Catmull(). This would require a loop, or passing 4 vertices with their proper vertices.

2) I would have to use splice now when I have to shift the array. splice(0,2) for 2D curves and splice(0,3) for 3D curves. This is a lot slower than shift(). See [this](https://stackoverflow.com/questions/10742369/javascript-shift-versus-splice-are-these-statements-equal).

src/webgl/primitives.js
AdilRabbani: Yes, that can be done. But then we need a separate scalar for the look up table of bezierVertex() as well. @mlarghydracept what's your call?...

AdilRabbani:
Hey @mlarghydracept need your reviews on the commit. I have made the requested changes.
PS. I guess the developer docs should be changed as well? Updating the information about curves?

src/webgl/primitives.js
Spongman: i think that just at a conceptual level it's preferable to use arrays for holding homogeneous sets of data and use objects for structuring data of mixed types. i could go into a long diatribe about type theory, JIT optimizations, off-by-one bugs, eventual use of `for...of`, etc... but really it just seems cleaner to use the constructs that the language provides.

src/core/vertex.js
```diff
@@ -331,23 +313,75 @@ p5.prototype.beginShape = function(kind) {
...
-    isBezier = true;
-    var vert = [];
-    for (var i = 0; i < arguments.length; i++) {
-      vert[i] = arguments[i];
-    }
-    vert.isVert = false;
-    if (isContour) {
-      contourVertices.push(vert);
+    if (vertices.length === 0) {
+      throw 'vertex() must be used once before calling bezierVertex()';
```
mlarghydracept: Why not throw an Error object?

src/core/vertex.js
```diff
@@ -269,15 +269,17 @@ p5.prototype.beginShape = function(kind) {
```
mlarghydracept: This might have to be dealt with in the merge from "webgl-gsoc-2018" to "master" but worth noting now that this file is in the [shape folder](https://github.com/processing/p5.js/blob/master/src/core/shape/vertex.js) now in master

src/webgl/primitives.js
```diff
@@ -1203,4 +1203,352 @@ p5.RendererGL.prototype.line = function() {
...
+          w_y[2] * this._lookUpTableBezier[i][2] +
+          w_y[3] * this._lookUpTableBezier[i][3];
+        _z =
+          w_z[0] * this._lookUpTableBezier[i][0] +
+          w_z[1] * this._lookUpTableBezier[i][1] +
+          w_z[2] * this._lookUpTableBezier[i][2] +
+          w_z[3] * this._lookUpTableBezier[i][3];
+        this.vertex(_x, _y, _z);
+      }
+      this.immediateMode._curveVertex.shift();
```
mlarghydracept: I understand that such a short loop seems like overkill but this feels a bit odd.

src/core/vertex.js
kjhollen: I think this will show up as a conflict in github but will merge cleanly—that's what happened with other changes to files that had been moved.

src/webgl/primitives.js
mlarghydracept: I agree with @Spongman here. Arrays should probably hold homogenous sets of data. I feel like this will be more clear for future devs.

AdilRabbani:
🎉🎉

kjhollen:
nice work @AdilRabbani! 🎉

PR #3088, opened 2018-07-18
Update OrbitControl to use p5.Camera / add Panning

AidanNelson:
This updates orbitControl to use p5.Camera object and adds support for right mouse button panning.

### DIFF
--- a/src/webgl/interaction.js
+++ b/src/webgl/interaction.js
@@ -1,18 +1,26 @@
+/**
+ * @module Lights, Camera
+ * @submodule Interaction
+ */
+
 'use strict';
 
 var p5 = require('../core/core');
 
 /**
- * Allows rotation of a 3D sketch by dragging the mouse. As the mouse is dragged
- * away from the center of the canvas in the X or Y direction, the sketch is
- * rotated about the Y or X axis respectively. Note that this rotation only
- * affects objects drawn after orbitControl() has been called in the draw() loop.
- * To reverse movement in either axis, enter a negative number for sensitivity.
- * Calling this function without arguments is equivalent to calling orbitControl(1,1).
+ * Allows movement around a 3D sketch using a mouse or trackpad.  Left-clicking
+ * and dragging will rotate the camera position about the center of the sketch,
+ * right-clicking and dragging will pan the camera position without rotation,
+ * and using the mouse wheel (scrolling) will move the camera closer or further
+ * from the center of the sketch. This function can be called with parameters
+ * dictating sensitivity to mouse movement along the X and Y axes.  Calling
+ * this function without parameters is equivalent to calling orbitControl(1,1).
+ * To reverse direction of movement in either axis, enter a negative number
+ * for sensitivity.
  * @method orbitControl
  * @for p5
- * @param  {Number} [sensitivityX]        sensitivity to mouse movement along X axis
- * @param  {Number} [sensitivityY]        sensitivity to mouse movement along Y axis
+ * @param  {Number} [sensitivityX] sensitivity to mouse movement along X axis
+ * @param  {Number} [sensitivityY] sensitivity to mouse movement along Y axis
  * @chainable
  * @example
  * <div>
@@ -24,35 +32,42 @@ var p5 = require('../core/core');
  * function draw() {
  *   background(200);
  *   orbitControl();
- *   translate(-20, 15, -20);
- *   torus(20, 5);
- *   translate(15, 0, 15);
- *   torus(20, 5);
- *   translate(15, 0, 15);
- *   torus(20, 5);
- *   translate(15, 0, 15);
- *   torus(20, 5);
- *   translate(15, 0, 15);
- *   torus(20, 5);
- *   translate(15, 0, 15);
- *   torus(20, 5);
+ *   rotateY(0.5);
+ *   box(30, 50);
  * }
  * </code>
  * </div>
  *
  * @alt
- * Camera orbits around a series of floating shapes when mouse is
- * hold-clicked & then moved.
+ * Camera orbits around a box when mouse is hold-clicked & then moved.
  */
-//@TODO: implement full orbit controls including
-//pan, zoom, quaternion rotation, etc.
+
 // implementation based on three.js 'orbitControls':
 // https://github.com/mrdoob/three.js/blob/dev/examples/js/controls/OrbitControls.js
 p5.prototype.orbitControl = function(sensitivityX, sensitivityY) {
   this._assert3d('orbitControl');
   p5._validateParameters('orbitControl', arguments);
 
+  // If the mouse is not in bounds of the canvas, disable all behaviors:
+  var mouseInCanvas =
+    this.mouseX < this.width &&
+    this.mouseX > 0 &&
+    this.mouseY < this.height &&
+    this.mouseY > 0;
+  if (!mouseInCanvas) return;
+
   var cam = this._renderer._curCamera;
+
+  if (typeof sensitivityX === 'undefined') {
+    sensitivityX = 1;
+  }
+  if (typeof sensitivityY === 'undefined') {
+    sensitivityY = sensitivityX;
+  }
+
+  // default right-mouse and mouse-wheel behaviors (context menu and scrolling,
+  // respectively) are disabled here to allow use of those events for panning and
+  // zooming
 
   // disable context menu for canvas element and add 'contextMenuDisabled'
   // flag to p5 instance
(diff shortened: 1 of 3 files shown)


src/webgl/interaction.js
```diff
@@ -19,19 +19,30 @@ var p5 = require('../core/core');
...
  *   createCanvas(100, 100, WEBGL);
+ *   normalMaterial();
  * }
- *
  * function draw() {
  *   background(200);
- *   // Orbit control allows the camera to orbit around a target.
  *   orbitControl();
- *   box(30, 50);
+ *   translate(-20, 15, -20);
```
kjhollen: the original example here was a little simpler. did you think it didn't illustrate the concept clearly enough? could this more complex example be a second example?

src/webgl/interaction.js
```diff
@@ -19,19 +19,30 @@ var p5 = require('../core/core');
...
  * </code>
  * </div>
  *
  * @alt
- * Camera orbits around box when mouse is hold-clicked & then moved.
+ * Camera orbits around a series of floating shapes when mouse is
+ * hold-clicked & then moved.
  */
 //@TODO: implement full orbit controls including
 //pan, zoom, quaternion rotation, etc.
```
kjhollen: what's left from this TODO? anything? :)

src/webgl/interaction.js
```diff
@@ -41,6 +52,17 @@ p5.prototype.orbitControl = function(sensitivityX, sensitivityY) {
   this._assert3d('orbitControl');
   p5._validateParameters('orbitControl', arguments);
 
+  var cam = this._renderer._curCamera;
+
+  // disable context menu for canvas element and add 'contextMenuDisabled'
+  // flag to p5 instance
```
kjhollen: could we add an explanation here for why this is needed? e.g. right mouse button is used for which interactive component?

src/webgl/interaction.js
```diff
@@ -49,59 +71,81 @@ p5.prototype.orbitControl = function(sensitivityX, sensitivityY) {
...
+        _z + cam.centerZ,
+        cam.centerX,
+        cam.centerY,
+        cam.centerZ,
+        0,
+        1,
+        0
+      );
     }
+    if (this.mouseButton === this.RIGHT) {
```
kjhollen: this isn't if/else—does something strange happen if the user presses both buttons at once?

src/webgl/interaction.js
```diff
@@ -19,19 +19,30 @@ var p5 = require('../core/core');
  * <code>
```
kjhollen: description of interaction needs update to include pan behavior!

src/webgl/interaction.js
AidanNelson: I reverted to the original.  I was a bit concerned it would be difficult to see the panning behavior with only one object in the scene, but on second look, it doesn't seem to be a problem.

src/webgl/interaction.js
AidanNelson: Just added zooming, so unless anyone wants to implement quaternions, I think not.  Removed.

src/webgl/interaction.js
AidanNelson: Done!

src/webgl/interaction.js
AidanNelson: For whatever reason, it just uses the most recently pressed mouseButton behavior with my mouse, but I've switched to `else if` behavior just in case.

src/webgl/interaction.js
AidanNelson: Updated.

AidanNelson:
This most recent revision:
* adds zoom behavior (and corresponding `_mouseWheelDeltaY` and _pmouseWheelDeltaY` properties on the p5.prototype)
* refactors code into `p5.Camera._orbit()` function to be used by both orbit and zooming behaviors,
* updates documentation.

src/webgl/interaction.js
```diff
@@ -1,107 +1,141 @@
...
 p5.prototype.orbitControl = function(sensitivityX, sensitivityY) {
   this._assert3d('orbitControl');
   p5._validateParameters('orbitControl', arguments);
 
+  // If the mouse is not in bounds of the canvas, disable all behaviors:
+  var mouseInCanvas =
+    this.mouseX < this.width &&
+    this.mouseX > 0 &&
+    this.mouseY < this.height &&
+    this.mouseY > 0;
```
kjhollen: hmm, how does this feel in use in the web editor? It's very common to have the mouse in focus but of out frame there. I wonder if it will feel totally fine, or a little discontinuous?

src/webgl/interaction.js
```diff
@@ -1,107 +1,141 @@
...
+  }
+
+  if (this.mouseIsPressed) {
+    // ORBIT BEHAVIOR
+    if (this.mouseButton === this.LEFT) {
+      var deltaTheta =
+        -sensitivityX * (this.mouseX - this.pmouseX) / scaleFactor;
+      var deltaPhi = sensitivityY * (this.mouseY - this.pmouseY) / scaleFactor;
+      this._renderer._curCamera._orbit(deltaTheta, deltaPhi, 0);
+    } else if (this.mouseButton === this.RIGHT) {
```
kjhollen: I tried this out in the manual-test-examples and the default interaction is very well tuned :)

maybe my lingo is out of date, but I might have expected movement along the y-axis to tilt the camera, and nothing to happen with the z-axis?

src/webgl/interaction.js
AidanNelson: I added this check to avoid page-scrolling activating the zoom.  In that case, if a canvas were halfway down a page, the camera view would be too far from the origin to see the sketch by the time the user reached the canvas.  But perhaps it is not necessary for panning or zooming.  [Here](http://alpha.editor.p5js.org/full/BkqOcCrHm) is a web-editor example with two p5 library files: 
* "p5.js" is the current behavior (mouse position checked for orbit/pan/zoom)
* "p5-mousecheckonzoom.js" mouse position only checked for zoom

In the 2nd case, right-clicking outside of the canvas still activates the context menu (only deactivated on the canvas element itself), stopping the panning behavior anyway.  But a user can start a pan with the mouse inside of the canvas, and continue it with the mouse outside of the canvas, which seems correct.   

This case also introduces one odd behavior:  if you right-click and hold outside of the canvas, then release the mouse inside the canvas, the context-menu is disabled and panning is activated despite the mouse no longer being pressed.  This lasts until you click the mouse again.  I'm not sure that this is a common enough / weird enough behavior to constitute a bug, but I just wanted to point it out.

I think this 2nd behavior probably makes the most sense.  Thoughts?

kjhollen: Cool, thank you for putting this example together! I agree that it's fine-as-is, with the stops. And as for the weird bug, I think it's ok to file that as an edge case issue and see if we can develop a solution for it in the future? I think you're right that it's unlikely to occur, but good to have documented just in case.

src/webgl/interaction.js
kjhollen: oh sorry, I should have explained this better! I am looking at the manual-test-example for webgl/interaction. in the default orbitControl case, it looks like right click + x-axis movement pans but right click + y axis movement zooms in or out along the z-axis. is tilting not part of the default orbit control? I'm also not familiar with what the default expectation is for users navigating a 3D world—so feel free to ignore this if I'm out of my depth. ;)

lmccart:
@kjhollen go ahead and merge these webgl ones if you approve

src/webgl/interaction.js
AidanNelson: Ahh, ok!  Tilting is not a part of the default orbitControl as implemented here except that orbiting the center point tilts the camera to maintain the center point.  Panning behavior is also limited to the XZ plane.  This is how [three.js orbitControl](https://threejs.org/examples/?q=control#misc_controls_orbit) is implemented and I thought it was a good way to ensure that users don't 'get lost' by panning 'above' or 'below' their sketches.  

This arbitrary definition of a 'ground' as the XZ plane is definitely somewhat limiting if a user wants to move in the Y axis or has an especially 'tall' sketch, but I think is easier to use as a beginner.

src/webgl/interaction.js
eveevans: @kadamwhite @AidanNelson 
This is causing an annoying needs of first Touch on Mobile in order to activate orbitControl on a Phone (or touch device) 

I already create an issue for this. But not sure which will be the best approach
https://github.com/processing/p5.js/issues/5176

Issue #2504, opened 2018-01-06
Color picker <input> for DOM library


stellartux:
#### Nature of issue?

- [x] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [x] Color
- [x] Other (specify if possible) - p5.dom.js

#### New feature details:
I was thinking it'd be useful if there was a quick and easy way to add the color picker DOM element with p5.js by doing something like `createColor()` . It's possible to `createInput('#ffffff', 'color')` but calling `value()` on this input returns a hex color string, it'd be nice to be able to read and write the value as a `p5.Color`. For this I was thinking it'd be useful to parameterize the `Color.toString()` function so it could format the string in different ways, in this case as a 6 digit hex color code.

I'm going to attempt to implement this.

limzykenneth:
As you mentioned `createInput('#ffffff', 'color')` is already available to create a color picker element and so having an additional function `createColor()` may be confusing and unnecessary. 

As for returning a `p5.Color` object instead of a hex string, I'm also a bit skeptical of it. The main reason being having some sort of parity with the native HTML/JS API. The reason it returns a hex string is because with native API it returns a hex sting (ie. `docElement.value` returns a hex string). Not doing too much magic behind the scene helps with transitioning or working with other library/frameworks. It also aids the case when you need to send the color data off to a server instead of having to serialize it yourself. Getting the hex string and initializing a p5.Color object like so `color(myInput.value)` can easily create a `p5.Color` object as well.

Perhaps you have some use cases where `color(myInput.value)` is not a viable solution?

amoetodi:
I'm surprised you think its unnecessary having an easy way to input a color in a graphics library. If `createColor()` is unnecessary, by the same logic so are `createSlider()`, `createButton()`, `createCheckbox()`, `createSelect()` and `createRadio()`.

It's not that the current method isn't viable to use, it's just inconsistent with the rest of the dom library. Currently, if you want to create a color input and choose the color it starts with, you have to convert it to a hex color code manually, and you can't create one without choosing a default color.  Using `color(myInput.value)`  feels like an awkward workaround when none of the other dom element values need converted. I can see why you would want it to be an easy transition to work with other frameworks or native HTML/JS, but if that's the idea it'd be better not to use the dom library at all. Isn't the point of the dom library letting people use dom elements without having to think about what's going on behind the scenes? 

[...]

PR #3089, opened 2018-07-19
Add createColorPicker 

TanviKumar:
Base for the createColorPicker method.

Hoping someone can guide me on how to make the element return a p5.Color object.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -799,10 +799,8 @@
    * function setup() {
    *   createCanvas(100, 100);
    *   background('grey');
-   *   var c1 = '#ff0000';
-   *   inp1 = createColorPicker(c1);
-   *   var c2 = color('yellow');
-   *   inp2 = createColorPicker(c2);
+   *   inp1 = createColorPicker('#ff0000');
+   *   inp2 = createColorPicker(color('yellow'));
    *   inp1.input(setShade1);
    *   inp2.input(setShade2);
    *   setMidShade();
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -785,6 +785,81 @@
   };
 
   /**
+   * Creates a colorPicker element in the DOM for color input.
+   * The .value() method will return a hex string (#rrggbb) of the color
+   * The .color() method will return a p5.Color object with the current chosen color
+   *
+   * @method createColorPicker
+   * @param {String|p5.Color} [value] default color
+   * @return {p5.Element} pointer to <a href="#/p5.Element">p5.Element</a> holding created node
+   * @example
+   * <div>
+   * <code>
+   * var inp1, inp2;
+   * function setup() {
+   *   createCanvas(100, 100);
+   *   background('grey');
+   *   var c1 = '#ff0000';
+   *   inp1 = createColorPicker(c1);
+   *   var c2 = color('yellow');
+   *   inp2 = createColorPicker(c2);
+   *   inp1.input(setShade1);
+   *   inp2.input(setShade2);
+   * }
+   *
+   * function draw() {
+   *   // Finding a shade between the two
+   *   var commonShade = lerpColor(inp1.color(), inp2.color(), 0.5);
+   *   fill(commonShade);
+   *   rect(20, 20, 60, 60);
+   * }
+   *
+   * function setShade1() {
+   *   console.log('You are choosing shade 1 to be : ', this.value());
+   * }
+   * function setShade2() {
+   *   console.log('You are choosing shade 2 to be : ', this.value());
+   * }
+   * </code>
+   * </div>
+   */
+  p5.prototype.createColorPicker = function(value) {
+    p5._validateParameters('createColorPicker', arguments);
+    var elt = document.createElement('input');
+    var self;
+    elt.type = 'color';
+    if (value) {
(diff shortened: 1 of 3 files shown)


TanviKumar:
This PR aims to resolve #2504

limzykenneth:
The value passed in probably needs to be able to handle different value types (hex strings, p5.Color object, etc) so you will need to overload it by checking value type.

For the element to return a p5.Color object, you will need to create the p5.Color object in the function itself from its value using the `color` function then assign the created object to the relevant property on the returned object.

lmccart:
@almchung can you review this and help guide @TanviKumar in following up with @limzykenneth's suggestions?

TanviKumar:
@limzykenneth trying to add the `color` method or trying to use any of the p5.Color based methods (like `toString` to assign the right value to the created object produces an error as the methods aren't accessible. Would like some guidance on how to procees here

limzykenneth:
@TanviKumar How are you calling the `color()` method here? I would think `this.color()` should probably suffice but if you are already doing that, maybe try `p5.prototype.color()`

TanviKumar:
@almchung I'm getting error stating related to colorMode when I use `color`. Having issues with rectifying it

TanviKumar:
@almchung I've updated the code and the example, do take a look. Shall I also go ahead an add a sketch to manual-test-examples under the test folder?

almchung:
@TanviKumar Hey, it looks good!! One last small thing, if you have time can you check in [LN832](https://github.com/processing/p5.js/pull/3089/commits/ef00da5d0481d36c5a096b1b910ae126e1d69642#diff-84dfcaf5d56a24f4ee13a32c644817b0R832) if we can use `instanceof` to check if the input is an instance of p5.Color object? Thanks!

TanviKumar:
@almchung instanceof works just fine, added that to the code!

lib/addons/p5.dom.js
```diff
@@ -784,6 +784,81 @@
     return self;
   };
 
+  /**
+   * Creates a colorPicker element in the DOM for color input.
+   * The .value() method will return a hex string (#rrggbb) of the color.
+   * The .color() method will return a p5.Color object with the current chosen color.
+   *
```
lmccart: can we add an example here?

lmccart: oops nm i see it below!

lib/addons/p5.dom.js
```diff
@@ -784,6 +784,81 @@
...
+   *   inp1 = createColorPicker(c1);
+   *   var c2 = color('yellow');
+   *   inp2 = createColorPicker(c2);
+   *   inp1.input(setShade1);
+   *   inp2.input(setShade2);
+   * }
+   *
+   * function draw() {
+   *   // Finding a shade between the two
+   *   var commonShade = lerpColor(inp1.color(), inp2.color(), 0.5);
```
lmccart: to create a new color ever loop of draw will actually slow things down quite a bit when there's more going on in the sketch. is it possible to make an example using a `.changed()` listener instead? this would fit better with best JS practices, setting the user up for good programming patterns. see [this example](https://p5js.org/reference/#/p5.Element/changed) as a guide.

lib/addons/p5.dom.js
```diff
@@ -784,6 +784,81 @@
...
+      if (value instanceof p5.Color) {
+        elt.value = value.toString('#rrggbb');
+      } else {
+        p5.prototype._colorMode = 'rgb';
+        p5.prototype._colorMaxes = {
+          rgb: [255, 255, 255, 255],
+          hsb: [360, 100, 100, 1],
+          hsl: [360, 100, 100, 1]
+        };
+        var c = p5.prototype.color(value);
```
lmccart: is this intermediary variable required?

lib/addons/p5.dom.js
```diff
@@ -784,6 +784,81 @@
...
+    self = addElement(elt, this);
+    // Method to return a p5.Color object for the given color.
+    self.color = function() {
+      if (value.mode) {
+        p5.prototype._colorMode = value.mode;
+      }
+      if (value.maxes) {
+        p5.prototype._colorMaxes = value.maxes;
+      }
+      var c = p5.prototype.color(this.elt.value);
```
lmccart: is this intermediary variable required?

lib/addons/p5.dom.js
TanviKumar: Will shift that code to the Input Change callback methods, setShade1 and setShade2 here!

lib/addons/p5.dom.js
almchung: nope, we can remove them [LN841](https://github.com/processing/p5.js/pull/3089/files#diff-84dfcaf5d56a24f4ee13a32c644817b0R841), [LN856](https://github.com/processing/p5.js/pull/3089/files#diff-84dfcaf5d56a24f4ee13a32c644817b0R856) . Thank you, @TanviKumar !

TanviKumar:
@almchung @immcart I've made the required updations. Shall I add the manual test examples and push the code within this PR?

lmccart:
thanks @TanviKumar. one comment then this should be good to merge.

TanviKumar:
@lmccart Made the requested change!

lmccart:
thanks @TanviKumar!

Issue #2751, opened 2018-03-30
touchStarted() breaks slider


aatishb:
#### Nature of issue?

- [X] Found a bug

#### Most appropriate sub-area of p5.js?

- [X] Other - DOM

#### Which platform were you using when you encountered this?

- [X] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 0.6.0
- Web browser and version: Chrome 65.0, Safari 11.0.3
- Operating System: Mac os X
- Steps to reproduce this:

When I include the touchStarted() function in my code, the slider will no longer slide. But when I delete it, the slider works again.

[Example code](http://alpha.editor.p5js.org/aatish/sketches/BJYx2fhqf)

IcyTv:
I'm not sure on this, but as far as I see it, the slider only gets blocked if you return false. As soon as you return true the slider works...

aatishb:
That's interesting. Also noticed that touchMoved() also breaks the slider but touchEnded() does not.

LadySith:
Hi @aatishb ,

I'm looking into this issue and it seems to be working as expected since if you return true the slider works correctly. Would you be able to give the specific situation or link me to the code in which this is an issue?

aatishb:
Hi, Thanks for looking into this. It seems to work fine with return true. Perhaps then the documentation on the above pages (touchStarted, touchMoved, mouseDragged, mousePressed) should be updated to say 'return true' instead. Currently it says "To prevent any default behavior for this event, add "return false" to the end of the method." Other than that this seems ok to close.

PR #3092, opened 2018-07-22
Edited documentation for pointer events

LadySith:
Edited documentation for touchStarted(), touchPressed(), mouseMoved()
and mouseDragged() to specify inclusion of return true to enable default
behaviors when using methods.

### DIFF
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -466,6 +466,8 @@ p5.prototype._setMouseButton = function(e) {
  * Browsers may have different default
  * behaviors attached to various mouse events. To prevent any default
  * behavior for this event, add "return false" to the end of the method.
+ * Otherwise add "return true" if you wish for default
+ * mouse behavior to be recognized by browser.
  *
  * @method mouseDragged
  * @example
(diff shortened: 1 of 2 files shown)


src/events/mouse.js
```diff
@@ -466,6 +466,8 @@ p5.prototype._setMouseButton = function(e) {
  * Browsers may have different default
  * behaviors attached to various mouse events. To prevent any default
  * behavior for this event, add "return false" to the end of the method.
+ * Otherwise add "return true" if you wish for default
+ * mouse behavior to be recognized by browser.
```
lmccart: I thought that the default behavior would occur if there was no return line. Or do I have that wrong?

limzykenneth: Returning nothing should give the default behavior as the code only check for false with `===` so any other value (including `undefined` when there is no return) should give default behaviour

limzykenneth:
Hi @LadySith, can you explain a bit about this PR? The events functions doesn't need the user to return true to enable default behavior, only return false to disable them.

LadySith:
Hi @limzykenneth and @lmccart . Sorry I was trying to find a solution to this [Issue 2751](https://github.com/processing/p5.js/issues/2751) which turned out to be a non-issue. Since the default is to enable the behavior then this pull request can be closed.

PR #3095, opened 2018-07-24
refresh lock file

Spongman:
it looks like the `package-lock.json` file that's checked into master is out of sync with `package.json`, worse it looks like the lock file comes from the future - it includes many packages with higher versions than specified in `package.json`. maybe `package.json` was updated at some point, but only the lock file was checked in? this is currently breaking `yarn import` which is more sensitive to these kinds of errors than npm's `lock-verify`.

anyway... this PR is the result of running `npm install` with npm v6.1.0 on node v10.6.0. which should bring everything in sync.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,13 +4,37 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@types/core-js": {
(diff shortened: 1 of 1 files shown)


lmccart:
thanks!

PR #3097, opened 2018-07-27
Fixes arc stroke, removes arcModeAdjust

### LINKED ISSUE #2919: arc renders unexpectedly when start and stop angle are the same

AdilRabbani:
Fixes #2919 : 2D arc renders/flickers a rect on the screen. Also removes arcModeAdjust() as it is not being used with the new arc implementation in webgl-gsoc-2018.

### DIFF
--- a/src/core/canvas.js
+++ b/src/core/canvas.js
@@ -17,17 +17,5 @@ module.exports = {
     } else if (mode === constants.CENTER) {
       return { x: a - c * 0.5, y: b - d * 0.5, w: c, h: d };
     }
-  },
-
-  arcModeAdjust: function(a, b, c, d, mode) {
-    if (mode === constants.CORNER) {
-      return { x: a + c * 0.5, y: b + d * 0.5, w: c, h: d };
-    } else if (mode === constants.CORNERS) {
-      return { x: a, y: b, w: c + a, h: d + b };
-    } else if (mode === constants.RADIUS) {
-      return { x: a, y: b, w: 2 * c, h: 2 * d };
-    } else if (mode === constants.CENTER) {
-      return { x: a, y: b, w: c, h: d };
-    }
   }
 };
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -435,14 +435,14 @@ p5.Renderer2D.prototype._acuteArcToBezier = function _acuteArcToBezier(
 
   // Return rotated waypoints.
   return {
-    ax: Math.cos(start),
-    ay: Math.sin(start),
-    bx: lambda * cos_phi + mu * sin_phi,
-    by: lambda * sin_phi - mu * cos_phi,
-    cx: lambda * cos_phi - mu * sin_phi,
-    cy: lambda * sin_phi + mu * cos_phi,
-    dx: Math.cos(start + size),
-    dy: Math.sin(start + size)
+    ax: Math.cos(start).toFixed(7),
+    ay: Math.sin(start).toFixed(7),
+    bx: (lambda * cos_phi + mu * sin_phi).toFixed(7),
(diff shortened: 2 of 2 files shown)


Spongman:
i wonder if it might be better to try to track down the cause of this issue in `_acuteArcToBezier`? it seems a little heavy-handed to just render all the numbers as truncated strings and then re-reparse them.

does anyone know a reason why p5 doesn't just use canvas' `arc` method here (instead of approximating cubic beziers)?

AdilRabbani:
I agree with @Spongman. Shouldn't we use canvas' arc for this purpose? I have tested it on many examples and it works great. What do you think? @mlarghydracept

mlarghydracept:
I am not very familiar with this part of the library. There may be a good reason why [arc](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/arc) wasn't used. I dug through the history a bit but it doesn't go back far enough to see when or why this decision was first made. Maybe @lmccart or @joecridge have some insight?

AdilRabbani:
okay....I'm guessing this is because canvas' arc doesn't have an option for a width and a height....https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/arc

joecridge:
Hi all. @AdilRabbani is correct that the reason we roll our own arc implementation is so that we can support the `arc()` and `ellipse()` signatures of the Processing API, which (depending on `ellipseMode()`) allow you specify the width and height of the underlying ellipse. This should mean that you can port a Proccessing sketch to p5.js with those drawing calls meaning the same thing.

I haven’t had a chance to test yet, but I suspect that the flickering is caused by @AdilRabbani’s self-described hack (no disrespect intended 😛) in #3042 (specifically, [these two lines](
https://github.com/processing/p5.js/pull/3042/files#diff-b224eabb1710349c5e23fcf38de6c7d9R451)), and that the rounding induced by the stringification and re-parsing in this PR ‘resolves’ the issue (at least on the arc sizes being tested) by sweeping the impression under the carpet. I think that both of these changes mask the true issue in the original implementation (#2919 – unexpected behaviour when start and end angle are the same), and will make future bugs even more difficult to reason about (essentially, a hack fixing a hack).

I’ll dig up the previous discussions about exactly what start/end angle are supposed to do (the behaviour for p5.js  is well-defined, but deliberately _somewhat_ different from Processing, whose behaviour is not self-consistent) and I’ll post it over on #2919.

lmccart:
Thanks @AdilRabbani looks like this has been incorporated https://github.com/processing/p5.js/blob/master/src/core/p5.Renderer2D.js#L415 so I'll close this PR.

Issue #3087, opened 2018-07-16
mouseButton lags one click behind in canvas.mousePressed


j6k4m8:
#### Nature of issue?

- [x] Found a bug

#### Most appropriate sub-area of p5.js?
- [x] Core/Environment/Rendering

#### Which platform were you using when you encountered this?
- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.6.0
- Web browser and version: Chrome ~67 (Did not test on other browsers)
- Operating System: MacOS, Ubuntu, ChromeOS (Did not test on Windows)
- Steps to reproduce this:

It appears that the `mouseButton` lags one click behind when attaching the `mousePressed` listener to the canvas rather than the global p5 namespace (as per examples [here](https://github.com/processing/p5.js/blob/877a8dcdcc1b77d1d3d1dbd1d91150373b9ea033/test/manual-test-examples/instantiation-instance/case6/sketch.js) and explanation [here](https://github.com/processing/p5.js/issues/1437))

## Minimal reproduction:

```js
// Create canvas to populate later
let canvas;

function setup() {
  canvas = createCanvas(100, 100);

  // Prevent right-clicks from opening context menu.
  // Note: This is optional; bug persists either way.
  canvas.elt.oncontextmenu = (function() { return false });
  
  // Attach mousePressed to canvas, not global p5
  canvas.mousePressed(function() {
    console.log(mouseButton);
  })
}

function draw() {
  // Doesn't matter what goes here...could even be empty.
  background(0, 10, 10);
}
```

- Run the above code
- Click `left`, `right`, `right`
- The console will show:
  `0`, **`LEFT`**, `RIGHT`

At first, I thought this was a [race condition](https://github.com/processing/p5.js/issues/996) but it doesn't appear to be; I think `mouseButton` might just not be set yet when this function runs. Thoughts?

Spongman:
can you try this with 0.6.1?

j6k4m8:
Confirmed on 0.6.1: https://alpha.editor.p5js.org/j6m8/sketches/rJzZPF5QX

Spongman:
ok, it looks like it's because the `canvas.mousePressed` call is just setting the event listener functions on the canvas element, and this is bypassing (or rather just happening before) the `mousePressed` logic.

j6k4m8:
oh interesting. sounds like that's not an easy fix on this end then?

Spongman:
[...]

PR #3102, opened 2018-07-30
bugfix: Set mousepress details prior to Element#mousePressed

j6k4m8:
This resolves the issue in #3087, in which the `mouseButton` variable reports the incorrect value in `mousePressed` functions attached to `p5.Element`s rather than to the global scope.

Looking forward to hearing your thoughts; this does not change functionality; tests still pass as they do on latest mater, and I don't believe this affects or requires changes to any of the existing examples.

**[EDIT]** I'd be happy to add some inline docs if anyone feels that it's unclear what's happening here.

### DIFF
--- a/src/core/p5.Element.js
+++ b/src/core/p5.Element.js
@@ -232,7 +232,12 @@ p5.Element.prototype.class = function(c) {
  *
  */
 p5.Element.prototype.mousePressed = function(fxn) {
-  adjustListener('mousedown', fxn, this);
+  var eventPrependedFxn = function(event) {
+    this._pInst._setProperty('mouseIsPressed', true);
+    this._pInst._setMouseButton(event);
+    return fxn();
+  };
+  adjustListener('mousedown', eventPrependedFxn, this);
   return this;
 };
 


lmccart:
this is great thanks @j6k4m8. oh I just saw your [EDIT]. yes, I think it would be helpful to have an inline comment here if you don't mind adding it.

j6k4m8:
Thanks @lmccart!! For sure — should I add to this branch for a subsequent merge?

lmccart:
@j6k4m8 sure, that makes sense

limzykenneth:
This PR is causing the issue at #3141. I'm not familiar enough with this part of the codebase but it seems to be lacking certain checks for undefined variables or unbound variables.

j6k4m8:
Looking into this today. Do any contributors know why some elements might
not have these vars populated? A little out of my depth here

meiamsome:
@j6k4m8 @limzykenneth Looks like this is probably a separate but related bug - the p5.dom.js library isn't passing down the p5 instance here: https://github.com/processing/p5.js/blob/6d832afac913bca3a0e615622e47474944addcfb/lib/addons/p5.dom.js#L251

limzykenneth:
@meiamsome That's the part that I was suspecting as well since `p5.Element` will need to have `pInst` defined somewhere and only place I see it explicitly defined is at the constructor. Still, I'm not familiar with the setup of p5.dom so I'll leave it to you or someone else to confirm and fix this. 😄

PR #3104, opened 2018-07-31
Add inline docs to mousePressed listener

j6k4m8:
@lmccart Added docs! Hopefully now it's more clear what's happening here.


lmccart:
@j6k4m8 there's a build process. the lib/p5.* files should not be included, and are .gitignored
https://github.com/processing/p5.js/blob/master/.gitignore#L6

lmccart:
comments are great, thanks!

PR #3110, opened 2018-08-03
Webgl text

### LINKED ISSUE #2183: Implement text() for webgl mode
#### Nature of issue?

- [ ] Found a bug
- [x] New feature request

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

text() only works in 2D mode currently. there is a lot we can leverage from the OpenType library already included here to generate opengl textures for use in rendering. I'll sign up to take this on soon. :)

Spongman:
closes #2183

implements `text()` for the webgl renderer. most of the code is in src/webgl/text.js, however some other changes were required:

- moving most of the `p5.Renderer2D.prototype.text()` and `textAlign()` code into shared methods in `p5.Renderer`, this necessitated:
- changing the `textAlign` and `textBaseline` properties so they are backed directly by variables in the renderer instead of being calculated based on canvas properties. this also simplifies the logic for those properties.
- changes to `push`/`pop`, `_applyTextProperties` methods to accommodate the above.
- changes to various `p5.Font` methods to make them renderer-independent.
- optimization in `p5.RendererGL.prototype.getTexture` which gets called a _ton_.
- add support for creating `p5.Texture` from an `ImageData` object.
- a new private `p5.Shader.prototype.updateTextures()` method that ensures all dirty textures are sent to the GPU.

### DIFF
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -45,7 +45,7 @@ p5.Renderer = function(elt, pInst, isMainCanvas) {
   this._textAscent = null;
   this._textDescent = null;
   this._textAlign = constants.LEFT;
-  this._textBaseline = constants.BOTTOM;
+  this._textBaseline = constants.BASELINE;
 
   this._rectMode = constants.CORNER;
   this._ellipseMode = constants.CENTER;
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -233,6 +233,7 @@ p5.Renderer.prototype.text = function(str, x, y, maxWidth, maxHeight) {
         break;
     }
 
+    var baselineHacked = false;
     if (typeof maxHeight !== 'undefined') {
       switch (this._textBaseline) {
         case constants.BOTTOM:
@@ -240,6 +241,10 @@ p5.Renderer.prototype.text = function(str, x, y, maxWidth, maxHeight) {
           break;
         case constants.CENTER:
           y += (maxHeight - totalHeight) / 2;
+          break;
+        case constants.BASELINE:
+          baselineHacked = true;
+          this._textBaseline = constants.TOP;
           break;
       }
 
@@ -264,6 +269,10 @@ p5.Renderer.prototype.text = function(str, x, y, maxWidth, maxHeight) {
 
       this._renderText(p, line, x, y, finalMaxHeight);
       y += p.textLeading();
+
+      if (baselineHacked) {
+        this._textBaseline = constants.BASELINE;
+      }
     }
   } else {
     // Offset to account for vertically centering multiple lines of text - no
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -44,6 +44,8 @@ p5.Renderer = function(elt, pInst, isMainCanvas) {
   this._textStyle = constants.NORMAL;
   this._textAscent = null;
   this._textDescent = null;
+  this._textAlign = constants.LEFT;
+  this._textBaseline = constants.BOTTOM;
 
   this._rectMode = constants.CORNER;
   this._ellipseMode = constants.CENTER;
@@ -75,6 +77,8 @@ p5.Renderer.prototype.push = function() {
       _textFont: this._textFont,
       _textLeading: this._textLeading,
       _textSize: this._textSize,
+      _textAlign: this._textAlign,
+      _textBaseline: this._textBaseline,
       _textStyle: this._textStyle
     }
   };
@@ -153,6 +157,132 @@ p5.Renderer.prototype.textDescent = function() {
     this._updateTextMetrics();
   }
   return this._textDescent;
+};
+
+p5.Renderer.prototype.textAlign = function(h, v) {
+  if (typeof h !== 'undefined') {
+    this._setProperty('_textAlign', h);
+
+    if (typeof v !== 'undefined') {
+      this._setProperty('_textBaseline', v);
+    }
+
+    return this._applyTextProperties();
+  } else {
+    return {
+      horizontal: this._textAlign,
+      vertical: this._textBaseline
+    };
+  }
+};
+
+p5.Renderer.prototype.text = function(str, x, y, maxWidth, maxHeight) {
+  var p = this._pInst,
+    cars,
+    n,
+    ii,
+    jj,
+    line,
+    testLine,
+    testWidth,
+    words,
+    totalHeight,
+    finalMaxHeight = Number.MAX_VALUE;
+
+  if (!(this._doFill || this._doStroke)) {
+    return;
+  }
+
+  if (typeof str === 'undefined') {
+    return;
+  } else if (typeof str !== 'string') {
+    str = str.toString();
+  }
+
+  str = str.replace(/(\t)/g, '  ');
+  cars = str.split('\n');
+
+  if (typeof maxWidth !== 'undefined') {
+    totalHeight = 0;
+    for (ii = 0; ii < cars.length; ii++) {
+      line = '';
+      words = cars[ii].split(' ');
+      for (n = 0; n < words.length; n++) {
+        testLine = line + words[n] + ' ';
+        testWidth = this.textWidth(testLine);
+        if (testWidth > maxWidth) {
+          line = words[n] + ' ';
+          totalHeight += p.textLeading();
+        } else {
+          line = testLine;
+        }
+      }
+    }
+
+    if (this._rectMode === constants.CENTER) {
+      x -= maxWidth / 2;
+      y -= maxHeight / 2;
+    }
+
+    switch (this._textAlign) {
+      case constants.CENTER:
+        x += maxWidth / 2;
+        break;
+      case constants.RIGHT:
+        x += maxWidth;
+        break;
+    }
+
+    if (typeof maxHeight !== 'undefined') {
+      switch (this._textBaseline) {
+        case constants.BOTTOM:
+          y += maxHeight - totalHeight;
+          break;
+        case constants.CENTER:
+          y += (maxHeight - totalHeight) / 2;
+          break;
+      }
+
+      // remember the max-allowed y-position for any line (fix to #928)
+      finalMaxHeight = y + maxHeight - p.textAscent();
+    }
+
+    for (ii = 0; ii < cars.length; ii++) {
+      line = '';
+      words = cars[ii].split(' ');
+      for (n = 0; n < words.length; n++) {
+        testLine = line + words[n] + ' ';
+        testWidth = this.textWidth(testLine);
+        if (testWidth > maxWidth && line.length > 0) {
+          this._renderText(p, line, x, y, finalMaxHeight);
+          line = words[n] + ' ';
+          y += p.textLeading();
+        } else {
+          line = testLine;
+        }
+      }
+
+      this._renderText(p, line, x, y, finalMaxHeight);
+      y += p.textLeading();
+    }
+  } else {
+    // Offset to account for vertically centering multiple lines of text - no
+    // need to adjust anything for vertical align top or baseline
+    var offset = 0,
+      vAlign = p.textAlign().vertical;
+    if (vAlign === constants.CENTER) {
+      offset = (cars.length - 1) * p.textLeading() / 2;
+    } else if (vAlign === constants.BOTTOM) {
+      offset = (cars.length - 1) * p.textLeading();
+    }
+
+    for (jj = 0; jj < cars.length; jj++) {
+      this._renderText(p, cars[jj], x, y - offset, finalMaxHeight);
+      y += p.textLeading();
+    }
+  }
+
+  return p;
 };
 
 p5.Renderer.prototype._applyDefaults = function() {
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -32,6 +32,8 @@ var defaultShaders = {
   ),
   phongVert: fs.readFileSync(__dirname + '/shaders/phong.vert', 'utf-8'),
   phongFrag: fs.readFileSync(__dirname + '/shaders/phong.frag', 'utf-8'),
+  fontVert: fs.readFileSync(__dirname + '/shaders/font.vert', 'utf-8'),
+  fontFrag: fs.readFileSync(__dirname + '/shaders/font.frag', 'utf-8'),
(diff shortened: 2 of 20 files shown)


src/core/p5.Renderer.js
```diff
@@ -155,6 +159,132 @@ p5.Renderer.prototype.textDescent = function() {
...
+    return this._applyTextProperties();
+  } else {
+    return {
+      horizontal: this._textAlign,
+      vertical: this._textBaseline
+    };
+  }
+};
+
+p5.Renderer.prototype.text = function(str, x, y, maxWidth, maxHeight) {
```
Spongman: this code came from p5.Renderer2D.prototype.text

src/webgl/p5.RendererGL.js
```diff
@@ -951,16 +960,14 @@ p5.RendererGL.prototype._getEmptyTexture = function() {
...
-  };
-  //this.drawMode = constants.TEXTURE;
-  var tex = this.textures.find(checkSource);
-  if (!tex) {
-    tex = new p5.Texture(this, img);
-    this.textures.push(tex);
+  var textures = this.textures;
+  for (var it = 0; it < textures.length; ++it) {
+    var texture = textures[it];
+    if (texture.src === img) return texture;
```
Spongman: using a for loop instead of `Array.find()` which is really slow..

src/webgl/p5.Texture.js
```diff
@@ -131,29 +135,22 @@ p5.Texture.prototype.init = function(data) {
...
+    return false; // nothing to do!
   }
 
   var textureData = this._getTextureDataFromSource();
+  var updated = false;
 
   var gl = this._renderer.GL;
   // pull texture from data, make sure width & height are appropriate
   if (textureData.width !== this.width || textureData.height !== this.height) {
+    updated = true;
```
Spongman: using this flag to de-duplicate the `texImage2D` code

src/webgl/shaders/font.frag
```diff
@@ -0,0 +1,189 @@
+#extension GL_OES_standard_derivatives : enable
```
Spongman: i'll comment this once we have glsl minification in the build

src/core/p5.Renderer.js
```diff
@@ -44,6 +44,8 @@ p5.Renderer = function(elt, pInst, isMainCanvas) {
   this._textStyle = constants.NORMAL;
   this._textAscent = null;
   this._textDescent = null;
+  this._textAlign = constants.LEFT;
+  this._textBaseline = constants.BOTTOM;
```
Spongman: these are stored here now, instead of as native 2d canvas properties.

AdilRabbani:
Oh boy, this is my first time reviewing a commit....please bear with me 😬

src/webgl/p5.RendererGL.js
AdilRabbani: Maybe indexOf would be faster? http://www.andygup.net/fastest-way-to-find-an-item-in-a-javascript-array/

src/webgl/shaders/font.frag
AdilRabbani: Yes, this file needs comments. 👍

src/webgl/p5.RendererGL.js
Spongman: i don't think indexOf would work in this case since we're looking for the element that has a particular (`src`) property. this is that `find()` is intended for - apply a predicate to each element and return the first that matches. but the function call is really slow.

src/webgl/shaders/font.frag
Spongman: i added issue #3114

src/webgl/text.js
```diff
@@ -0,0 +1,693 @@
...
+
+/**
+ * @private
+ * @class ImageInfos
+ * @param {Integer} width
+ * @param {Integer} height
+ *
+ * the ImageInfos class holds a list of ImageDatas of a given size.
+ */
+function ImageInfos(width, height) {
```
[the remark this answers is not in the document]
Spongman: ok, 1543600 should fix the alignment/metrics issues.

src/webgl/shaders/font.frag
Spongman: ... and PR #3116

AdilRabbani:
Can there be a default font for WebGL as well? It feels odd that nothing is displayed if the user doesn't loads a font. Getting 'WEBGL: only opentype fonts are supported' log in the console.
@kjhollen can you take a final look at this as well? 😀

Spongman:
i thought about adding a `z` parameter, but decided against it since other primitives like `arc`, `rect`, `ellipse` & `triangle` only take 2d coordinates. i figure if you're going real 3d positioning of stuff in webgl-mode then you probably should be familiar with `translate()`.

i added an example for `text()`.

AdilRabbani:
This seems a bit tricky. How is 2D mode handling this? I think it seems to be using a default web safe font?

Spongman:
2d mode is using `<canvas>` so, as well as opentype fonts, it can use the fonts built into the browser (or OS). the default font is set here: https://github.com/processing/p5.js/blob/0956fc9c22f8be8ad0e14ca87542cf422c40eebd/src/core/p5.Renderer2D.js#L31

AdilRabbani:
@kjhollen @mlarghydracept This is working great! There is a problem of adding a default font. What do you suggest?

dhowe:
> I'm getting this output on https://p5js.org/reference/#/p5/textLeading example in 2D on your build 
> textAscent() also seems to give unexpected results in 2D mode

if you need more checks for alignments, leading etc, some of the tests in p5.js/test/manual-test-examples/p5.Font/ might be useful, just to verify that 2d, 3d, and processing are all producing the same results...

kjhollen:
thank you for the initial review, @AdilRabbani! I'm happy to take a deeper look at this—I'll do it next week once I settle in from moving a little more, if that works for you @Spongman?

mlarghydracept:
I just had an opportunity to briefly look over this and everything looked fantastic. I can properly review it on Friday.

Spongman:
thanks everyone, no rush ;-)

kjhollen:
or @mlarghydracept can look sooner :)

Spongman:
sorry about that - didn't mean to spam - travis got stuck for some reason on the last commit and i don't know any other way to re-trigger it.

mlarghydracept:
Okay I have been taking some more time with this and realize there are huge number of changes to look over. I will try to do a line-by-line this weekend but for now I have just managed to give it a broad-strokes once over. So far everything works really well but I have concerns that this structure in text.js may make future contributions difficult. Text rendering is a difficult task but I wonder if there are ways that this can be changed slightly to make it less intimidating to future contributors. Namely there is a high density of class declarations and utility functions. This may be unavoidable but it feels different than the rest of the library. Also, why the additions to the p5.Vector class? Maybe I am missing something but couldn't the static methods be used to accomplish this same task without adding to that class?

mlarghydracept:
Hm I see how that probably looks pretty convoluted with the static methods. I still don't really know if it is okay to add these fluent methods to p5.Vector just for this single use.

So I have been trying to think about solutions to the default font issue. This feels like a pretty confusing roadbump for users. I haven't had any epiphanies but I want to keep up a conversation about it because it feels significant. It may be an inescapable aspect of the implementation, in which case we'd need to make sure that the reference clearly explains the need for a loaded font when using WebGL.

test/manual-test-examples/p5.Font/custom/sketch.js
```diff
@@ -1,11 +1,25 @@
+var _webgl = false;
```
mlarghydracept: I'm getting what appear to be errors for this in Chrome. I get console warnings about the number of GL contexts and the dead computer icon inside some of the canvases.

screenshot 426

Spongman: yeah, it looks like some browsers limit the number of concurrent webgl contexts. try edge, it doesn't do this.

Spongman:
ok, i don't mind changing it back to the static methods. the math is actually a little simpler than it used to be.

mlarghydracept:
Switching it back to the static methods seems like a good play. Also consider separating the p5.Font/custom test for GL. I think having some contexts that get destroyed by certain browsers may be confusing to future testers. It also seems like this takes over that test so that if the browser being used can handle GL then it will default to testing only the GL.

Spongman:
> It also seems like this takes over that test so that if the browser being used can handle GL then it will default to testing only the GL.

nope, the 2d stuff is always shown.

Spongman:
ok, i have removed the vector helper functions. please verify that the ugly mess that it has turned into is equivalent to what was there before.

mlarghydracept:
Ah I see now. Maybe don't separate the tests then. I think you should still limit the number of tests though to ensure that there isn't confusion because of browser limitations on simultaneous GL contexts.

As far as the "uglifying". This is only "ugly" or difficult to read because of its insistence on being concise. You could achieve the same functionality without all of the nested static methods, but it would just be a more verbose solution. I understand that the static method approach has a few more "p5.Vector"s in it and I see how the first solution is more pleasant to look at but neither is exceptionally clear to anyone that is looking at it with fresh eyes. I think adding fluent utility methods to the p5.Vector class could be great but perhaps that is its own PR. Then in that PR you could clean up all of the clumsy uses of static methods throughout the codebase. As it stands now I am just trying to minimize the number of changes that happen in any one PR so it doesn't become confusing why other decisions were made by other people in different parts of the codebase. (ie using the static methods rather than the newly introduced fluent methods). Thanks for being flexible to making changes!

Spongman:
ok, i just removed the webgl stuff from that test. i thought it would be useful to play around with, but on second thoughts i don't think it's worth the hassle.

mlarghydracept:
Hi @Spongman okay! I got a chance to return to this PR this morning. I spent a few hours playing around and making small sketches with 3D text. As expected everything is working rather nicely.

I feel like the only remaining issue is the default font. From what I understand there are a few options but they are all slightly less than ideal. What do you think about moving forward with this PR and simply adding some language in the reference for text that explains that text requires a loaded font in WebGL rendering as a stopgap? If we went that route then we could open a separate issue to discuss a best approach to getting a default font. I am particularly drawn to the idea of embedding a base64-encoded binary but I feel like this warrants a more targeted design conversation.

Spongman:
i'm going to put this here, because i don't want to add it to the wiki yet:

-----


# Text
To work with text in webgl mode, you have two options: use the 2d p5.js API to render to an offscreen image, or use the native webgl `text()` method. There are advantages & disadvantages to both which are discussed below:

## Offscreen Renderer
You can draw your text to an offscreen renderer first, and then use it as a texture.

Advantages:
- You can use the same fonts available to the 2d p5.js API, including the browser's built-in fonts (eg, 'mono', 'sans', etc...).
- You can use `stroke()` to draw an outline around the glyphs.
- The performance may be better, especially if you are able to render your text only once in `setup()` and reuse that texture in your `draw()` function.

Disadvantages:
- The fidelity of the text may not be as high as with the `text()` method due to pixellation when zooming or inaccuracies in anti-aliasing when tilting.

Here is an example of using an offscreen renderer to draw text:
```javascript
  var pg;
  function setup(){
    pg = createGraphics(256,256);
  }
  function draw(){
    background(0);
    pg.background(255);
    pg.text('hello world!', 50, 50);
    //pass graphics as texture
    texture(pg);
    plane(100);
 }
```

## webgl `text()`
The webgl version of the `text()` method works very similarly to the 2d version. However there are a few differences:

Disdvantages:
- You can only use use opentype/truetype fonts loaded in your [`preload()`](https://p5js.org/reference/#/p5/preload) function using the [`loadFont()`](https://p5js.org/reference/#/p5/loadFont) method. You must either place those fonts files in a location accessible from your sketch, or use a [CORS-compatible](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) webl URL.
- [`stroke()`](https://p5js.org/reference/#/p5/) is not currently supported.

Advantages:
- The fidelity of the rendered text should be better, especially when zooming & tilting.
- The performance may be better, especially if the text changes regularly and you are unable to cache the offscreen image used in the previous method.

Here is an example of loading an opentype font and using it to draw text with the webgl `text()` method:
```javascript
var myFont;
function preload() {
  myFont = loadFont('assets/AvenirNextLTPro-Demi.otf');
}

function setup() {
  fill('#ED225D');
  textFont(myFont);
  textSize(36);
  text('p5*js', 10, 50);
}
```

kjhollen:
this is good documentation @Spongman. I think we're phasing out wiki documentation in favor of the developer_docs folder, so it probably belongs in a markdown file there, or it could just be added to the web GL architecture docs. we will still need to update the reference for `text()` to indicate the no-default behavior for 3D mode because it will not be what users expect when they migrate from 2D mode.

Spongman:
oh, sorry, i didn't realize the wiki was going away. i think this info is more targetted to users rather than contributors. i could add it to the `text()` reference docs if you think that would be better?

Spongman:
travis stalled again, ugh.

mlarghydracept:
I agree that there it feels that there should be somewhere for more elaborate documentation. Reference docs focus on clarity and simplicity. Not sure about the solution there, but for now I feel that this PR is ready to merge. @kjhollen have any interest in giving this a final glance before I merge? It is a pretty large addition. Thanks for your patience @Spongman!

AdilRabbani:
[WebGL docs](https://github.com/processing/p5.js/blob/master/developer_docs/webgl_mode_architecture.md) would need an update regarding WebGL text in p5.js.

mlarghydracept:
Good call @AdilRabbani!

mlarghydracept:
@Spongman just to clarify - I think that we can move forward with a merge. I have spoken with @kjhollen. The only thing that needs to happen before that is that we need to update the [developer_docs] (https://github.com/processing/p5.js/blob/master/developer_docs/webgl_mode_architecture.md) as per @AdilRabbani's suggestion.

mlarghydracept:
Okay! @Spongman Merging now. Thanks for the final touches, especially the additional comments. It helps demystify quite a lot. This is a large and well-considered contribution! 

Can you make an issue that addresses the default font when you get a chance? It seems that you have already done quite a bit of work to consider possible solutions and it would be good to document these for future contributors.

Spongman:
fantastic, thanks for reviewing this!

Issue #2645, opened 2018-02-13
Feature request: lights() function from Processing


EndBug:
**Nature of issue?**
- [x] New feature request.

**Most appropriate sub-area of p5.js?**
- [x] Lights, Camera

**New feature details:**

In Processing there's a function (`lights()`) that quickly creates a good light effect. You can reproduce it in p5, but it's not as easy and quick as `lights()`.
It would be good to have this function in p5 too.

Spongman:
this is part of #2547

EndBug:
Oh, sorry, I didn't see that.
I'm glad that someone is trying to add this function, are you going to split it and try to pull request again?

kjhollen:
Someone can take this on by creating the new method in src/webgl/p5.RendererGL.js and just having it call `this.ambientLight(128, 128, 128)` and `this.directionalLight(128, 128, 128, 0, 0, -1)` to match the Processing behavior, more or less. some adjustments to the direction might be necessary.

Spongman:
you can just take the code from #2547. it's already tested and working.

IceDune:
@EndBug @Spongman @kjhollen , can i work on this pls?

stalgiag:
Help is still needed with this. The todo is add a function called `lights` that calls the other light functions as @kjhollen explains above. #3112 can be looked at for some clues on how to do example docs and manual tests, but the method only needs to do what is outline in @kjhollen 's comment above.

EndBug:
Hi guys! Seeing the last comment and the interest that this issue gained I decided to try it myself with PR #3553 
At the moment works in Node 10 but has failed the test for Node 8: can someone help me figure out why?

stalgiag:
Hi @EndBug this was unrelated to your code. I restarted the build. This time it was successful. I will open an issue. If this happens again restart the Travis CI build.

PR #3112, opened 2018-08-03
added lights function from processing

IceDune:
added lights method for a feature request.
issue #2645 .


IceDune:
just asking but, is there any change i need to do or is the feature not required?

kjhollen:
hi @IceDune, thanks for submitting this! I need to take a closer look at this, but a couple things I noticed at a glance:

- the `lights()` method here is defined for p5.RendererGL prototype but not on the p5 prototype. it will need to be added there too to ensure that it's accessible in the global scope—it can just refer to what you've written here. I think there are some good examples for how to do this in material.js
- can you add an example to the test/manual-test-examples folder? this would let other devs verify that the code is working as intended
- and add an example to the documentation itself, so that it will show up in the p5 reference :)

IceDune:
@kjhollen , Sure!

kjhollen:
Hi @IceDune, can you let me know if you're interested in making the updates requested here? or if you're busy, would it be ok if I worked on this a bit? thanks!

IceDune:
Hi @kjhollen , i'm sorry i was a bit busy, but could i work on this. i'll try to fix it as soon as possible. thanks!

IceDune:
@kjhollen , for the tests, in which folder under manual-test-examples should i add them to? 
Is it webgl/camera? and for testing this function, just calling `lights()` in `draw()` is enough right?
thank you.

kjhollen:
thanks for the updates @IceDune! for manual-testing-examples I think you can make a new folder in webgl/lights and yes, all it needs to do is show `lights()` in `draw()` and include a couple shapes we can see it working.

IceDune:
hi @kjhollen , i've added the tests, is there any change i need to apply in them? and please let me know how to add the documentation. thank you very much!

kjhollen:
thanks @IceDune! I'll give this a more detailed read soon, but here's a guide on how to generate the inline documentation / examples that will go with the function https://github.com/processing/p5.js/blob/master/developer_docs/inline_documentation.md

IceDune:
hi @kjhollen thank you for telling me how to write the documentation. please let me know if the test I've written is correct so that i can add it as an example. thanks a lot!

IceDune:
hi @kjhollen , may i know if there are any changes that need to be done. thank you.

stalgiag:
Hi @IceDune I am closing this for now so that others can take a crack at it. If you still want to work on it please post in #2645. Otherwise, just want to make sure others get a chance.

PR #3115, opened 2018-08-04
add a textAlign example illustrating vertical alignment

Spongman:
adds a doc sample for `textAlign` that illustrates vertical alignment:

image


dhowe:
also see the alignment examples in p5.js/test/manual-test-examples/p5.Font/, which compare a wide range of p5.js alignments with original processing

PR #3125, opened 2018-08-07
add icons to dev web view

Spongman:
just enables icons for the dev web server:

image

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -339,7 +339,12 @@ module.exports = function(grunt) {
     connect: {
       server: {
         options: {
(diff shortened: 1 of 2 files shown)


lmccart:
I'd prefer to leave this as is rather than add another patch file.

Spongman:
Yeah, the package ptching situation is a pain. I have submitted a PR upstream, maybe it'll stick...

lmccart:
yeah :\ if it does please feel free to resubmit this

Spongman:
it looks like they pushed a new version including this patch...

PR #3127, opened 2018-08-08
Add size limit for loadX methods

### LINKED ISSUE #2981: show error message when loading files that are too big
#### Nature of issue?

- [ ] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [x] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

There are currently two open issues for when the browser fails to load a file because it is too big: #2661 for `loadStrings()` and #2216 for `loadTable()`. I think they are not possible to fix, but it would be good to produce an error message for the user and suggest breaking the data into smaller files as a work around.

TanviKumar:
This is to close #2981 . Uses HEAD request to get the content-length of the file. If found to be greater than 64 MB, error is produced, else the file is fetched as before.

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -27,6 +27,7 @@ require('../core/error_helpers');
  * callback following the syntax specified <a href="https://github.com/camsong/
  * fetch-jsonp">here</a>.
  *
+ * This method is suitable for fetching files up to size of 64MB.
  * @method loadJSON
  * @param  {String}        path       name of the file or url to load
  * @param  {Object}        [jsonpOptions] options object for jsonp related settings
@@ -161,7 +162,16 @@ p5.prototype.loadJSON = function() {
 
       self._decrementPreload();
     },
-    errorCallback
+    function(err) {
+      // Error handling
+      p5._friendlyFileLoadError(5, path);
+
+      if (errorCallback) {
+        errorCallback(err);
+      } else {
+        throw err;
+      }
+    }
   );
 
   return ret;
@@ -180,6 +190,7 @@ p5.prototype.loadJSON = function() {
  * This method is asynchronous, meaning it may not finish before the next
  * line in your sketch is executed.
  *
+ * This method is suitable for fetching files up to size of 64MB.
  * @method loadStrings
  * @param  {String}   filename   name of the file or url to load
  * @param  {function} [callback] function to be executed after <a href="#/p5/loadStrings">loadStrings()</a>
@@ -264,7 +275,16 @@ p5.prototype.loadStrings = function() {
 
       self._decrementPreload();
     },
-    errorCallback
+    function(err) {
+      // Error handling
+      p5._friendlyFileLoadError(3, arguments[0]);
+
+      if (errorCallback) {
+        errorCallback(err);
+      } else {
+        throw err;
+      }
+    }
   );
 
   return ret;
@@ -303,6 +323,7 @@ p5.prototype.loadStrings = function() {
  * object:</p>
  * </p>
  *
+ * This method is suitable for fetching files up to size of 64MB.
  * @method loadTable
  * @param  {String}         filename   name of the file or URL to load
  * @param  {String}         options  "header" "csv" "tsv"
@@ -316,7 +337,7 @@ p5.prototype.loadStrings = function() {
  * @return {Object}                    <a href="#/p5.Table">Table</a> object containing data
  *
  * @example
- * <div class="norender">
+ * <div class='norender'>
  * <code>
  * // Given the following CSV file called "mammals.csv"
  * // located in the project's "assets" folder:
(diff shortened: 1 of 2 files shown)


almchung:
@TanviKumar Aiming for more of a warning than an error message would give a more suitable tone for this case, what do you think?  So maybe use phrases such as 'warning' or 'is recommended...'?

TanviKumar:
@almchung I'll go ahead and update the terms used in the warning. Currently my code doesn't fetch the file if the size exceeds the limit. If we are to give the user just a warning, should I alter the code such that the file is fetched anyway but a warning is given if size is larger than expected?

src/io/files.js
```diff
@@ -1085,6 +1088,68 @@ p5.prototype.httpDo = function() {
       }
     }
 
+    if (method === 'GET') {
+      // Get the size of the file to be fetched using HEAD request
+      return new Promise(function(resolve, reject) {
+        var xhr = new XMLHttpRequest();
```
lmccart: it looks like there are now two HTTP requests being made per httpDo call.. this one here and the request() on line 1153. is this needed? or is there a way to combine?

lmccart: and another one in getFile() called from line 1116?
I'm wondering if this function just needs to be cleaned up and streamlined so it's clear what's going on.
some inline comments would help a lot too.

src/io/files.js
```diff
@@ -1093,8 +1158,72 @@ p5.prototype.httpDo = function() {
...
+        type = 'text';
+      }
+    }
+
+    if (type === 'jsonp') {
+      promise = fetchJsonp(path, jsonpOptions);
+    } else {
+      promise = fetch(request);
+    }
+    promise = promise.then(function(res) {
```
lmccart: this line feels weird to me, can you explain what's going on here?

src/io/files.js
```diff
@@ -941,6 +941,7 @@ p5.prototype.httpPost = function() {
  * For more advanced use, you may also pass in the path as the first argument
  * and a object as the second argument, the signature follows the one specified
  * in the Fetch API specification.
+ * This method will only fetch files upto size of 64MB when "GET" is used.
```
lmccart: add space between "up" and "to"

src/io/files.js
TanviKumar: If a file has to be loaded, and a GET request has to be made, first a HEAD request is made. The HEAD request only fetches the headers and doesn't fetch the entire file so it doesn't take much longer at all. It gives the file size which we use to produce a warning. A normal GET request is then made after the HEAD request is complete to fetch the file the user wants.

So for GET requests, 2 HTTP requests are made, the one in line 1094 and 1116. For other types of requests, just the one in 1116 is made.

src/io/files.js
TanviKumar: Will do!

src/io/files.js
TanviKumar: This is the same code as present before, I'll add another parameter to the getFile() method so the code isn't repeated twice.

TanviKumar:
@lmccart I've pushed some code that makes things a lot cleaner and more modular

meiamsome:
I think `fetch` should be used in preference to XHR - You could just duplicate the current `Request` but set the method to `HEAD`. In either case, that makes the minimum amount of requests for a GET to the same host 2 round-trips and if CORS is used in certain situations this may even be 4 (CORS will send an extra OPTIONS request ahead of time to check that the CORS request is valid before sending it, depending upon if the request is considered 'safe')

As a guess, it may be possible to just read the `Content-Length` header from the fetch response using something like `res.headers.get('Content-Length')` before the body is read, somewhere around here: https://github.com/processing/p5.js/blob/989a25b8b87d06914fd5d6c2a59baee6ca1a7baf/src/io/files.js#L1122
That's entirely untested speculation, though, and by the looks of a quick search you'd have to handle that that header may be `null` if it's unsent or blocked by CORS

TanviKumar:
@meiamsome Will test out getting content length from the fetch response and get back to you.

For the case where no content length information is sent or available, I'm going to add a line about the known size limitations to the documentation of the loadX methods so users are informed beforehand

TanviKumar:
@meiamsome Considering we'll now be fetching the file no matter what the result and only producing a warning, I've changed the code significantly based on your suggestions and also updated much of the documentation. @lmccart @almchung This should seem a lot neater!

meiamsome:
This looks like a good solution to me. I think warning and continuing anyway is the best situation so as to avoid preventing people who are manually changing their browser's limits being restricted.

TanviKumar:
@almchung Do let me know if any changes are needed!

lmccart:
looks great, thanks @TanviKumar and @almchung and @meiamsome for reviewing!

PR #3134, opened 2018-08-09
Fix linter errors

Zalastax:
Just ran `lint:fix` and fixed the `===` problem.
Shouldn't the precommit hook / tests have noticed this 😕?


lmccart:
thanks @Zalastax. I think the tests are not enabled for the manual-test-examples folder to give us great flexibility for those files. that folder is really meant as sort of a scratch area for testing functionality that's hard to test through unit tests.

Issue #1135, opened 2015-11-27
Two sketch instances don't produce same stream of randoms during draw()


toolness:
The following script ought to produce two separate sketches that log identical streams of random numbers, I think:

``` js
var SEED = 1;

function randomLogger(name) {
  return function(p) {
    p.setup = function setup() {
      p.createCanvas(100, 100);
      p.randomSeed(SEED);
    };

    p.draw = function() {
      if (p.frameCount < 10)
        console.log(name + " frame # " + p.frameCount + ": " +
                    Math.floor(p.random(0, 10)));
    };
  };
}

var sketch1 = new p5(randomLogger("sketch1"));;

var sketch2 = new p5(randomLogger("sketch2"));;
```

However, when I run this, I get this in my console:

```
sketch1 frame # 1: 2
sketch2 frame # 1: 2
sketch1 frame # 2: 3
sketch2 frame # 2: 5
```

It seems like the random numbers are synchronized during `setup()` and the first call to `draw()`, but not after that.

mikewesthad:
This is the first time I'm poking around p5 source, but I think I see what's going on here.  
The [p5 constructor](https://github.com/processing/p5.js/blob/ad071c96efef47b752d942c69407d600c4c75fc9/src/core/core.js#L37) doesn't create an independent random generator for each instance.  So the [lcg](https://github.com/processing/p5.js/blob/ad071c96efef47b752d942c69407d600c4c75fc9/src/math/random.js#L16) function, which is used to generate random numbers with a seed, is going to be a shared reference across p5 instances.  They can't have different seeds, and they won't generate independent streams of random numbers. 

The stream of random numbers for a seed of 1 is: `2, 3, 5, 7, ...`.  The reason why you are seeing 2 twice is because the seed is being reset between sketch1's first frame and sketch 2's first frame.  If you add a log to p5 core, you can better see what's going on:

```js
p5.js:25841 The seed is being set to: 1
sketch.js:13 sketch1 frame #1: 2
p5.js:25841 The seed is being set to: 1
sketch.js:13 sketch2 frame #1: 2
sketch.js:13 sketch1 frame #2: 3
sketch.js:13 sketch2 frame #2: 5
sketch.js:13 sketch1 frame #3: 7
sketch.js:13 sketch2 frame #3: 0
```

This aspect of p5 seems like part of the design rather than a bug?

limzykenneth:
@mikewesthad I believe you analysis is correct.

[...]

PR #3139, opened 2018-08-11
Independent instances

Zalastax:
Solves the problem described in #1135. I've found some new problems though, so we should keep it open or create a separate issue and close that one.

There's still a lot more global mutable variables in math/noise, both an LCG that should now use the added _lcg function and some perlin noise variables. I suspect there are global mutable variables in other places, so we need to figure out a smart way of testing that.

### DIFF
--- a/src/core/main.js
+++ b/src/core/main.js
@@ -48,7 +48,7 @@ var p5 = function(sketch, node, sync) {
 
   /**
    * Called directly before <a href="#/p5/setup">setup()</a>, the <a href="#/p5/preload">preload()</a> function is used to handle
-   * asynchronous loading of external files in a blocking way. If a preload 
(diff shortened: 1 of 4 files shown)


Spongman:
`hashCount` is global. not sure whether or not it makes sense to change/keep.

`keysDown` is problematic: #3193

Zalastax:
`hashCount` seems to be a memory optimization. Having it global means that multiple sketches can't create enough ghashes for the page to run out of memory, I guess? There is a todo to remove the limitation so perhaps we can focus on that?

lmccart:
@Zalastax sorry for the delay on this. this looks good but is there a reason the lcg functions are pulled out into a separate file? could you put them in `src/math/random.js`?

PR #3150, opened 2018-08-15
Bugfix p5.DOM: provide pInst

### LINKED ISSUE #3141: button.mousePressed() broken?
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [x] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: p5.js v0.7.1 August 10, 2018
- Web browser and version: Chrome 68.0.3440.106 (Official Build) (64-bit)
- Steps to reproduce this:
Getting the error "`Uncaught TypeError: Cannot read property '_setProperty' of undefined at g.Element.<anonymous> (p5.min.js:7)`" when using the `button.mousePressed();` function.
This also seems to be happening in the p5 example linked below:
https://p5js.org/examples/dom-input-and-button.html

meiamsome:
Fixes #3141, adds regression test

Note that the problem is still visible for elements that weren't created by p5, but this needs a different fix - do we want to assume `new p5.Element()` should bind to the global script or should we check that `this._pInst` exists in the `mousedown` event wrapper? I might prefer promoting `p5._wrapElement` to `p5.wrapElement` and deprectaing explicit use of the `p5.Element` constructor

```html
<button id='test'>test</button>
<script>
function setup() {
  var elt = document.querySelector('#test');
  var elem = new p5.Element(elt);
  elem.mousePressed(function() {});
  elt.dispatchEvent(new Event('mousedown'));
}
</script>
```

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -187,7 +187,7 @@
   p5.prototype._wrapElement = function(elt) {
     var children = Array.prototype.slice.call(elt.children);
     if (elt.tagName === 'INPUT' && elt.type === 'checkbox') {
-      var converted = new p5.Element(elt);
+      var converted = new p5.Element(elt, this);
       converted.checked = function() {
         if (arguments.length === 0) {
           return this.elt.checked;
@@ -200,18 +200,18 @@
       };
       return converted;
     } else if (elt.tagName === 'VIDEO' || elt.tagName === 'AUDIO') {
-      return new p5.MediaElement(elt);
+      return new p5.MediaElement(elt, this);
     } else if (elt.tagName === 'SELECT') {
-      return this.createSelect(new p5.Element(elt));
+      return this.createSelect(new p5.Element(elt, this));
     } else if (
       children.length > 0 &&
       children.every(function(c) {
         return c.tagName === 'INPUT' || c.tagName === 'LABEL';
       })
     ) {
-      return this.createRadio(new p5.Element(elt));
+      return this.createRadio(new p5.Element(elt, this));
     } else {
-      return new p5.Element(elt);
+      return new p5.Element(elt, this);
     }
   };
 
@@ -248,7 +248,9 @@
   function addElement(elt, pInst, media) {
     var node = pInst._userNode ? pInst._userNode : document.body;
     node.appendChild(elt);
-    var c = media ? new p5.MediaElement(elt) : new p5.Element(elt);
+    var c = media
+      ? new p5.MediaElement(elt, pInst)
+      : new p5.Element(elt, pInst);
     pInst._elements.push(c);
     return c;
   }
@@ -1934,7 +1936,7 @@
         this.width = this.elt.offsetWidth;
         this.height = this.elt.offsetHeight;
 
-        if (this._pInst) {
+        if (this._pInst && this._pInst._curElement) {
           // main canvas associated with p5 instance
           if (this._pInst._curElement.elt === this.elt) {
             this._pInst._setProperty('width', this.elt.offsetWidth);
--- a/test/unit/addons/p5.dom.js
+++ b/test/unit/addons/p5.dom.js
@@ -1,3 +1,4 @@
+/* global testSketchWithPromise */
 suite('DOM', function() {
   suite('p5.prototype.select', function() {
     var myp5;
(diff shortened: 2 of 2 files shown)


Spongman:
> should we check that this._pInst exists in the mousedown event wrapper

can't hurt.

meiamsome:
@Spongman I would agree, but I would be tempted to actually log this into the console as it is likely a bug that will affect users. Perhaps we can add something to FES for situations that we know would only be caused by bugs with a link to the GH issues page?

Spongman:
i'm not sure how FES would know about this error condition. currently it just read the yui metadata. are you suggesting some non-javascript DSL for specifying preconditions?

meiamsome:
@Spongman Nah, I was more just thinking we just call it as part of FES. I'd think we could make a method something like the following:

```javascript
p5._bug = function(reason) {
  console.log('FES: p5.js has detected a condition that should never occur: ' + reason);
  console.log('This is likely a bug. You can find help here: <link to GH issues>');
}
```

And then we'd just do this in places where we're checking things cautiously:
```javascript
if (!this._pInst) {
  p5._bug('P5 Element with no p5 instance attached');
} else {
  // the current code
}
```

Spongman:
oh, i see. makes sense.

Spongman:
i would suggest just putting an `if` around the offending lines for now so this can be released quickly.

meiamsome:
Well, it's a separate issue really that I was musing on. This PR is ok to merge to resolve the current issue

lmccart:
thanks!

Spongman:
unfortunately, it turns out that this PR breaks `loadPixels` and `get` on `MediaElements`.

the issue is that those methods defer to `Renderer2D` methods that check `this._pInst || this` in order to get the rendering context. previously (before this PR) a MediaElement's _pInst was always `undefined`, so the `Renderer2D` would pull the `pixels` and associated state (`_pixelsDirty`, etc...) from the `MediaElement` itself. Now, however, `this._pInst` actually points to a `p5` instance, so the `Renderer2D` methods now **wrongly** use _that_ as the rendering context (`pixels`, `_pixelsDirty`, etc...) instead.

i hope that makes sense.

meiamsome:
@Spongman I think that this bugfix still fixes the code to what it should be, though. Which would imply to me that check in `Renderer2D` is buggy? Is that what you are saying here?

Spongman:
well, i _do_ think that the fact that `MediaElement` (and `Image`) defer to `Renderer2D` methods when they aren't themselves derived from `Renderer` _is_ buggy. but it's been like that for a long time.

this behavior is new, though (since this PR). `Element`s _never_ before had a non-`undefined` `_pInst`, so the code in the `Renderer2D` method always did the right thing. this PR now sets `_pInst` which breaks the check in those methods: `this._pInst || this` no longer points to the right thing.

meiamsome:
@Spongman Ok, I just looked at the code and you're right - indeed it looks like the element's `_pInst` was never used before either. The question is, is `_pInst` a property of an `Element` or a `Renderer2D`? I think it is a renderer property that leaked its way into the Element class, and should probably be removed at all. So revert this PR, revert #3102 and then remove the `pInst` argument and `_pInst` element. This would have the affect of reopening #3087 - and I'm not sure there is a good fix for that that wouldn't involve passing down a reference to the p5 instance in the Element class?

It looks the pInst parameter was originally added here: https://github.com/processing/p5.js/commit/4af24d1200f6ab79e004a07caea96149560c1b31#diff-d65c42c409e9a3b3c4c208a193f0520c but the code that was using it was since refactored out of the Element class.

Spongman:
how about just renaming `Element._pInst` to something that doesn't interfere with the renderer stuff?

meiamsome:
@Spongman Yeah, that'd work, but I don't know if a `p5.Element` instance should have state about the sketch it's from, that doesn't really seem to be that meaningful to an element?

Spongman:
oh, actually, that's not going to work (the renaming bit) because `Renderer` derives from `Element`.

the strange thing is that there's code in `Element` that _only_ works if `_pInst` is set:

https://github.com/processing/p5.js/blob/5e550780943306bae0c79cefa95b641a7042fe18/lib/addons/p5.dom.js#L2004-L2005

i'm pretty sure it would be simple to have gotten that to crash before this PR.

----

how about this:

in `Element` constructor:
```javascript
this._pInst = this._ctx = pInst;
```

in `Renderer` constructor:
```javascript
p5.Element.call(...);
this._ctx = this;
```

then `var ctx = this._pInst || this;` just becomes `this._ctx`.

var ctx = (this instanceof p5.Renderer) ? this : this._pInst;

Issue #2616, opened 2018-01-31
visual glitch due to z-buffering and transparent textures


Spongman:
this is just a quick issue to track the fact that z-buffering and semi-transparent textures don't mix. the z-buffer kills subsequently-drawn more-distant pixels, even if the buffered pixel is semi-transparent. the only effective solution is depth-sorting which is probably out of scope.

this is illustrated in the transparentTexture manual test (duplicated here: https://codepen.io/Spongman/pen/eVNOoj?editors=0010)

this repros in 0.5.16 and master since #2610. it's not visible in 0.6.0 because it's masked by #2608.

stalgiag:
This occurs in Processing as well. Adding the "will not fix" label for the above mentioned reasons (depth sorting is out of scope). This label and the "known issue" label will be used to close (for organization purposes) known issues that will not be fixed.

PR #3152, opened 2018-08-16
WebGL GSoC 2018

kjhollen:
This PR merges updates from the Web GL work done by @AidanNelson and @AdilRabbani as part of Google Summer of Code!

New features include:
- implementation of primitives including `arc()`, `curveVertex()`, `bezierVertex()`, `quadraticVertex()`, and `point()`
- implementation/abstraction of camera into p5.Camera object
- new and improved `orbitControl()`, along with `debugMode()` for help visualizing 3D coordinates + motion
- various bug fixes, unit tests, and documentation updates related to webgl

@mlarghydracept and I have been reviewing the code all summer, so I think it is ready to merge to master. Putting up the PR so that Aidan and Adil can have quick look to make sure everything is accounted for and that no unexpected issues occurred. Thanks for all of your great work this summer!!


AdilRabbani:
1) `arcModeAdjust()` can be removed from [helpers.js](https://github.com/processing/p5.js/blob/webgl-gsoc-2018/src/core/helpers.js). It isn't being used now.

2) As I was writing my [blogpost](https://adilrabbani.github.io/blogposts/SummerWithProcessing.html), I wrote an example making a sphere using 3D points. I noticed a weird alpha blending issue.



<p align="center">Note the white boundaries on points</p>

This is only noticeable when the points are big enough. I think this can be resolved by using the discard statement with an if condition, which I tried to avoid previously.

Spongman:
@AdilRabbani i believe you're seeing the same issue as #2616. `discard` may indeed help for the fully-transparent pixels, but you may still see a halo around the circles...

mlarghydracept:
@AdilRabbani I believe that it is acceptable to use an if conditional in the point frag shader. I know it is a faux pas for performance reasons but I have seen some examples online that get decent performance even with conditionals in the frag shader. [Here is one.](https://bl.ocks.org/robinhouston/4b736b08f369a94c419098a72e11c47b). Would you try to do that and see how bad of a performance hit there is? If rendering isn't terribly impacted than it may be a good solution for the moment so that the GSOC branch can get merged while it is still fairly easy to do so. As for the halo that @Spongman is talking about, I think that is a more minor issue and can be addressed later on.

AdilRabbani:
@mlarghydracept on it! Was just waiting for your opinion on the issue.

kjhollen:
Great. Any reason not to merge the branch as-is? I think it's ok if these lingering issues with point() transparency and arcModeAdjust() are resolved in separate, future PRs, so long as they are tracked somewhere.

AdilRabbani:
Yeah @kjhollen  I also wanted to know if I should make the changes and start a separate PR or just commit it here directly?

mlarghydracept:
@kjhollen @AdilRabbani I think it is fine to do it either way. The changes for the transparency fix should be minor so it won't be a big deal on either side.

mlarghydracept:
Okay all cleared up thanks to @AdilRabbani!

AidanNelson:
This looks good to me!

PR #3160, opened 2018-08-17
Fixes GL point alpha-blending issue

AdilRabbani:
Fixes WebGL point issue discussed in #3152


mlarghydracept:
Looks great, fixed the problem, and performance seems to be about the same on my computer! Thanks @AdilRabbani

PR #3161, opened 2018-08-17
removes arcModeAdjust()

AdilRabbani:
Removes arcModeAdjust() from helpers.js as it is not being used in the new implementation of arc #3152


mlarghydracept:
Thanks @AdilRabbani!

Issue #2999, opened 2018-06-08
Instance mode will try to use global functions


meiamsome:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [x] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Details about the bug: 
Code to reproduce:
```javascript
new p5(function(s) {
  s.setup = function() {
    s.background(0);
    setTimeout(function() {
      window.draw = function draw() {
        s.background(200);
      }
      s.draw = null;
    }, 2000);
  }
});
```
This is a heavily reduced version of the issue to show the problem, as it can often be unclear what is happening. In the following code, the global sketch appears animated because the instance mode is actually animating it by calling `draw`, and all of the global `p5` functions are bound to the global instance.
```javascript
function setup() {
  noLoop();
}

function draw() {
  background(100);
  ellipse(mouseX, mouseY, 10);
}

new p5(function(s) {
  s.setup = function() {
    s.background(200);
  }
});
```
I believe this happens for `preload` and `setup` too, but they are harder to verify.

PR #3162, opened 2018-08-17
Instance mode calls global

Zalastax:
The first two commits clean up JSHint / ESLint files.
`createP5Iframe` is needed in the tests I've added so I made it accessible for all tests in the third commit.
The fourth commit tests the problem described in #2999 and the last commit make the tests green again.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,16 +2,13 @@
   "env": {
     "node": true,
     "browser": true,
-    "mocha": true,
     "amd": true,
     "es6": true
   },
   "globals": {
-    "p5": true,
-    "assert": true,
-    "expect": true,
-    "sinon": true
+    "p5": true
   },
+  "root": true,
   "extends": ["eslint:recommended", "prettier"],
   "plugins": ["prettier"],
   "parserOptions": {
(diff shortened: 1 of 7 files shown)


Zalastax:
It was a bit confusing when to use `context` and when to use `this`.
I assume `context` is meant to be where the user defines their callback functions?
I think it would be nice to add
```JS
/**
 * User defined functions that are defined on window if global mode
 * @typedef {Object} UserContext
 * @property {Function} setup
 * @property {Function} draw
 * @property {Function} preload 
 */

/**
 *
 * @method _getContext 
 * @private
 * @return {UserContext}
 */
p5.prototype._getContext = function() {
  if (this._isGlobal) {
     return window;
  }
  return this;
}
```

lmccart:
thanks @Zalastax. can you separate out commits 1+2, commit 3, and commits 4+5? they seem to be addressing different issues.

Zalastax:
I separated out the first commit but there is a dependency chain between the other commits. It's possible to break the chain but the end result would be worse / risk merge conflicts that need to be resolved. If you still want that change to happen, just let me know.

lmccart:
thanks!

Zalastax:
You're absolutely right! I accidentally left in a `suite.only` 😱 https://github.com/processing/p5.js/pull/3162/files#diff-1d1e4669d2175b67408c9b9f1e65c1daR261. If the tests don't work after replacing that with `suite`, revert the merge and I'll get back to fix that soon.

lmccart:
aha that fixed things thanks!

PR #3166, opened 2018-08-19
Rework asynchronous loading test to remove logs

### LINKED ISSUE #3093: global-mode errors fired during mocha tests
(current master)

i'm seeing a bunch of global mode binding errors being fired during the global-mode mocha tests. this is happening both in-browser and on the command line (including travis).

```
p5 had problems creating the global function "_initializeInstanceVariables", possibly because your code is already using that name as a variable. You may want to rename your variable to something else.
You just changed the value of "_initializeInstanceVariables", which was a p5 function. This could cause problems later if you're not careful.
```

meiamsome:
Fixes #3093

Many thanks to @Zalastax for basically solving it, so I have added them as a co-author. The issue is likely due to some cross frame issues with the way attaching handlers works (interestingly the code they came up with was working fine in an actual browser for me). I re-coded the test in a way more similar to the other tests and it appears to work.

### DIFF
--- a/test/unit/core/core.js
+++ b/test/unit/core/core.js
@@ -171,22 +171,17 @@ suite('Core', function() {
 
     test('works when p5.js is loaded asynchronously', function() {
       return new Promise(function(resolve, reject) {
-        createP5Iframe();
+        createP5Iframe(`
+          <script>
+            window.onload = function() {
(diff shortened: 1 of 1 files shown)


lmccart:
thanks!

Issue #3175, opened 2018-08-27
calling pixelDensity() with no args messes with graphic context (calls thereafter behave incorrect)


vogtadi:
#### Nature of issue?

- [X] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [X] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 0.7.1
- Web browser and version: Chrome Version 68.0.3440.106 (Official Build) (64-bit)
- Operating System: MacOSX 10.13.6
- Steps to reproduce this:
As an example, use code in https://p5js.org/reference/#/p5/createGraphics 
add a "var d = pd.pixelDensity();" before the last line
-> circle in example gets scaled
-> expected result:
  - calling pixelDensity without any arguments only READS the parameter

#### Workaround

use the property _pixelDensity instead the function pixelDensity() for reading until this issue has been fixed.

lmccart:
fixed with #3176

PR #3176, opened 2018-08-27
remove side-effect from pixelDensity() when calling the function as read-only

vogtadi:
fixes issue #3175

the function did unconditionally resize the canvas, which has the side effect of clearing it. Calling pixelDensity() as a read-only function should **not** modify the state, i.e. the graphic or anything else.


src/core/environment.js
```diff
@@ -537,10 +537,12 @@ p5.prototype.pixelDensity = function(val) {
       this._pixelsDirty = true;
     }
     returnValue = this;
+    this.resizeCanvas(this.width, this.height, true); // as a side effect, it will clear the canvas
   } else {
     returnValue = this._pixelDensity;
   }
-  this.resizeCanvas(this.width, this.height, true);
+  // not resizing the canvas if we only read the pixelDensity
+  // -> fixes issue #3175
```
lmccart: these two comment lines above can be removed

vogtadi: Agreed. I did only include the manual tests for easier integration of the pull request, so the integrator has some context.

lmccart:
this is great, thank you!

PR #3195, opened 2018-09-08
p5.sound 0.3.9

therewasaguy:
- add `save` and `getBlob` methods to SoundFile
- improve AudioIn documentation
- fix Synth velocity range to prevent clipping/distortion, master limiter has -3 threshold and sharp knee


Spongman:
the tests don't pass on this branch. now those breakages are in master...

PR #3197, opened 2018-09-09
fix webgl get() & RendererGL test

### LINKED ISSUE #3196: webgl: RendererGL test is broken
in master right now, if you run `npm run dev`, and navigate to `/test/test.html` with the debugger console open, you'll see an `Error` thrown when the RendererGL test starts to run. this error isn't caught by the test framework (because it's thrown during the suite setup phase) and so that test suite doesn't run at all.

Spongman:
closes #3196:
- fixes the RendererGL test, the third test of which fails, requiring the following fixes:
- moves `p5.Renderer2D.get()` code to `p5.Renderer` base class (since it's shared, removes the need for funky `p5.RendererGL.get` implementation.
- adds new virtual `p5.Renderer.readPixel` implementation of fetching a single pixel, implements this differently for 2D & GL.


src/core/p5.Renderer.js
```diff
@@ -109,6 +109,94 @@ p5.Renderer.prototype.resize = function(w, h) {
   }
 };
 
+/**
```
Spongman: this is mostly a verbatim copy from p5.Renderer2D.get, except the `instanceof` check is gone, and the `readPixel` call is used instead.

src/core/p5.Renderer.js
```diff
@@ -109,6 +109,94 @@ p5.Renderer.prototype.resize = function(w, h) {
...
+
+  // round down to get integer numbers
+  x = Math.floor(x);
+  y = Math.floor(y);
+  w = Math.floor(w);
+  h = Math.floor(h);
+
+  var sx = x * pd;
+  var sy = y * pd;
+  if (w === 1 && h === 1) {
```
Spongman: no more `instanceof` check

src/core/p5.Renderer.js
```diff
@@ -109,6 +109,94 @@ p5.Renderer.prototype.resize = function(w, h) {
...
+    var imageData, index;
+    if (ctx._pixelsDirty) {
+      imageData = this.readPixel(sx, sy);
```
Spongman: moves single-pixel reading to virtual `readPixel` method

test/unit/webgl/p5.RendererGL.js
```diff
@@ -74,22 +75,26 @@ suite('p5.RendererGL', function() {
   });
 
   suite('get()', function() {
-    myp5.createCanvas(100, 100, myp5.WEBGL);
```
Spongman: it's not valid to use `myp5` here, has to be done inside the tests themselves

Spongman:
i'm not sure your patch is valid. for example, there are other places where `p5.RendererGL.prototype.get` is called (eg MediaElement), and the behavior of that method has now changed.

PR #3201, opened 2018-09-10
travis: test on node 8 and 10

nikolas:
I think now it's pretty much the norm to test on the two main node.js branches like this, which are currently 8.x and 10.x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
-  - "8.9"
+  - "8"
+  - "10"
 addons:
   chrome: stable
 install: npm install


Spongman:
I think it may be worth backing this one out temporarily until Travis fixes the issues it’s having with builds randomly failing without reason. Building twice per commit nearly doubles the chance of build failures.

lmccart:
perhaps we just test node 10 then?

Spongman:
i think that might be best for now. i'd really like to find out why the travis builds are randomly failing, but from a cursory glance it looks like many others are having similar issues and it doesn't seem to be a priority for the travis devs :(

PR #3203, opened 2018-09-11
fix #3200 soundfile dispose error in reference test

### LINKED ISSUE #3200: sound tests are failing
the latest sound library merge #3195 included a broken test that's causing all subsequent travis builds to fail (https://travis-ci.org/processing/p5.js/builds/426166289).

the test is failing because https://github.com/processing/p5.js-sound/issues/318 and https://github.com/processing/p5.js-sound/issues/319

both these issues are fixed in https://github.com/processing/p5.js-sound/pull/320

therewasaguy:
Thanks, @Spongman, for the fix here: https://github.com/processing/p5.js-sound/pull/320


therewasaguy:
hmm, will have a look as it's failing again with 

```
  1) lib/addons/p5.sound.js p5.Reverb documentation example #1 works:
     Error: Failed to set the 'buffer' property on 'ConvolverNode': Cannot set buffer to non-null after it has been already been set to a non-null buffer
```

PR #3204, opened 2018-09-11
Babelify

Spongman:
not really ready for merging, just food for thought:
- runs babelify over the browserified library, converts es6 to es5.
- updates eslint to es6-compatible version
- adds `no-var` & `prefer-const` eslint rules, fixes errors
- uses lambdas instead of inline functions, where appropriate.

there's no attempt to do any es6-style structural changes. just trivial syntactic stuff.


Spongman:
oh yeah, i was hoping this might be useful as part of those discussions, at the very least as a proof-of-concept that it's reasonably easy to incorporate some of the simple es6-isms into the library source while retaining es5 compatibility on the backend.

PR #3205, opened 2018-09-11
Fix command line copy mistake

### LINKED ISSUE #3190: reference docs out of date
the docs @ https://p5js.org/reference/ are currently the ~~0.7.1~~ 0.6.1 😲  docs, whereas the library that it uses, and the download lins are 0.7.2.

i wonder if there's a way to automate pushing the docs simultaneously with the library? also, maybe there's a way to include the release version in the footer of the docs pages?

limzykenneth:
Closes  #3190 

A mistake I made in https://github.com/processing/p5.js/commit/3c944907a1e4794d8bf46f5de7924eef49805d76 which cause the reference to not be copied over to the website repo. After merging this the release step for the documentation should be run to update the reference which by now should be 3 months out of date. 😨

### DIFF
--- a/tasks/release/release-docs.js
+++ b/tasks/release/release-docs.js
@@ -38,7 +38,7 @@ module.exports = function(grunt) {
           console.log('Copying new docs ...');
           return new Promise(function(resolve, reject) {
             exec(
-              'cp -r docs/reference/**/* p5-website/src/templates/pages/reference/',
+              'cp -r docs/reference/ p5-website/src/templates/pages/reference/',
               function(err, stdout, stderr) {
                 if (err) {
                   reject(err);


lmccart:
ah great. I was wondering why that stopped working. I've been copying it manually so it should be pretty up to date despite the grunt release task not working. thanks!

limzykenneth:
@lmccart The data.json files are not up to date and currently are still stuck at 0.6.1 at the website repo https://github.com/processing/p5.js-website/blob/master/src/templates/pages/reference/data.json 

I changed where those files should be copied to so if you had been copying it to the old location it would have been overwritten by the website build step. You will need to copy them over to the above location (`src/templates/pages/reference/`) and rerun the website build step.

PR #3209, opened 2018-09-12
avoid attachListener when disabling default drag/drop behavior

### LINKED ISSUE #3208: dragOver() and drop() not working together?
This is a bug I think I have found with p5.dom.js. Here is an example:

https://editor.p5js.org/icm/sketches/HJuRys8_X

If you drag a file over the canvas, it will turn red (according to how `dragOver()` should work. However, if I comment back in the `drop()` method for receiving the file, the `dragOver()` event ceases to function.)

```javascript
function setup() {
  var c = createCanvas(100, 100);
  background(200);
  textAlign(CENTER);
  text('drop image', width / 2, height / 2);
  c.dragOver(test);  // does not work if line below is added back in
  // c.drop(gotFile);
}

function test() {
  background(255,0,0);
}

function gotFile(file) {
  var img = createImg(file.data);
}
```

I went back and found that the code above works with `0.5.14`, but stops working with `0.5.15` (for both p5.js and p5.dom.js.)

(Addendum! The order seems to matter here. If I call `drop()` before `dragOver()`, `test()` is called but receiving the file breaks.)

Spongman:
closes #3208

this PR does two things (in `p5.Element.prototype.drop()`):
- removes the calls to `stopPropagation()` in the 3 event handlers.
- avoids using `attachListener` when disabling the `dragover` and `dragleave` default behavior since these events may already have been set by the user and `attachListener` will un-register those event handlers.

### DIFF
--- a/src/core/p5.Element.js
+++ b/src/core/p5.Element.js
@@ -1001,26 +1001,22 @@ p5.Element.prototype.drop = function(callback, fxn) {
 
   // Is the file stuff supported?
   if (window.File && window.FileReader && window.FileList && window.Blob) {
-    // If you want to be able to drop you've got to turn off
-    // a lot of default behavior
-    attachListener(
-      'dragover',
-      function(evt) {
-        evt.stopPropagation();
-        evt.preventDefault();
-      },
-      this
-    );
+    if (!this._dragDisabled) {
+      this._dragDisabled = true;
 
-    // If this is a drag area we need to turn off the default behavior
-    attachListener(
-      'dragleave',
-      function(evt) {
-        evt.stopPropagation();
+      var preventDefault = function(evt) {
+        //evt.stopPropagation();
         evt.preventDefault();
-      },
-      this
-    );
+      };
+
+      // If you want to be able to drop you've got to turn off
+      // a lot of default behavior.
+      // avoid `attachListener` here, since it overrides other handlers.
+      this.elt.addEventListener('dragover', preventDefault);
+
+      // If this is a drag area we need to turn off the default behavior
+      this.elt.addEventListener('dragleave', preventDefault);
+    }
 
     // Attach the second argument as a callback that receives the raw drop event
     if (typeof fxn !== 'undefined') {
@@ -1031,7 +1027,7 @@ p5.Element.prototype.drop = function(callback, fxn) {
     attachListener(
       'drop',
       function(evt) {
-        evt.stopPropagation();
+        //evt.stopPropagation();
         evt.preventDefault();
 
         // A FileList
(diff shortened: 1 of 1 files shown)


Spongman:
Retriggering stalled Travis build

Spongman:
unfortunately the retrigger option is only available for repos that you have write access to. for some reason.

Spongman:
https://github.com/travis-ci/travis-ci/issues/887

Spongman:
retriggering again... @Banzaiman, @cotsog

cotsog:
@Spongman I just submitted a PR that should make your builds more reliable: https://github.com/processing/p5.js/pull/3222

We have looked into this issue with NPM's support and our investigation haven't been fruitful, unfortunately.

This is the best we can do right now but it has proven to work for the folks over at Elastic. See https://github.com/elastic/apm-agent-nodejs/pull/558.

src/core/p5.Element.js
```diff
@@ -1001,26 +1001,22 @@ p5.Element.prototype.drop = function(callback, fxn) {
...
-    attachListener(
-      'dragover',
-      function(evt) {
-        evt.stopPropagation();
-        evt.preventDefault();
-      },
-      this
-    );
+    if (!this._dragDisabled) {
+      this._dragDisabled = true;
```
lmccart: can you clarify where this._dragDisabled is used?

src/core/p5.Element.js
```diff
@@ -1031,7 +1027,7 @@ p5.Element.prototype.drop = function(callback, fxn) {
     attachListener(
       'drop',
       function(evt) {
-        evt.stopPropagation();
+        //evt.stopPropagation();
```
lmccart: this line can be removed

src/core/p5.Element.js
```diff
@@ -1001,26 +1001,22 @@ p5.Element.prototype.drop = function(callback, fxn) {
...
+    if (!this._dragDisabled) {
+      this._dragDisabled = true;
 
-    // If this is a drag area we need to turn off the default behavior
-    attachListener(
-      'dragleave',
-      function(evt) {
-        evt.stopPropagation();
+      var preventDefault = function(evt) {
+        //evt.stopPropagation();
```
lmccart: this line can be removed

src/core/p5.Element.js
Spongman: that's all there is. it's just to ensure that the `preventDefault` stuff is only done once per element.

lmccart: ah got it, thanks

PR #3211, opened 2018-09-14
remove _runFrames

Spongman:
this is stale code from when p5 used `setInterval` for animation. it does nothing now.

`this._updateInterval` is never set.


lmccart:
👍

PR #3212, opened 2018-09-14
Stalling `npm install` workaround

cotsog:
Suggestions from Travis CI to make your builds more stable.

Retry npm install up to 3 times and prevent it from stalling using the travis_wait utility function.


lmccart:
cool thanks

PR #3222, opened 2018-09-18
Stalling `npm install` workaround

cotsog:
Add a 5 minutes timeout and retry to command up to 3 times


lmccart:
thanks! I think we've got this one in there now so I'll close this PR

Issue #2571, opened 2018-01-21
Move TypeScript definitions to their own repository


Zalastax:
We just made a release, and the files on GitHub does not contain any TypeScript definitions. p5.zip doesn't have them either. They are, however, available in the [npm package](https://yarnpkg.com/en/package/p5?files).
Is there any way to test the release script properly, without accidentally releasing to npm/GitHub? When I made the modifications to the release script I tested it as best as I could, but it was not enough.

Spongman:
i noticed they are missing from cdnjs, also: https://cdnjs.com/libraries/p5.js

lmccart:
I can add them to github, but my preference is not to add them to the p5.zip to avoid confusion for beginners with many files in the folder.

Zalastax:
Alright. Does that mean that you don't use the automated scripts when making a release? The scripts were modified to add the definition files to both the zip and GitHub, which is not what happened.

lmccart:
I do use the automated scripts, I thought that the files were created erroneously.
I am wondering if the typescript should actually be a separate project rather than coming out of this build. It's feeling like too much for me to maintain properly. What are your thoughts, is this possible?

Zalastax:
I'm sorry that it has increased the burden. It should be possible to maintain the definitions in another repo. It's more complicated, but the responsibility would be split to another maintainer.

The release scripts seem to be working then, so from my side everything is well. I don't think there will be many more changes to the TypeScript generator, just keep writing good and well-formatted documentation. If you want to wait on making a split to a repository for TypeScript, we can just change it so that the zip doesn't contain the definitions.

lmccart:
@Zalastax yes I think this would be much preferred and better for the overall sustainability of the library. I don't really know anything about typescript so it makes it really difficult for me to tell what should be happening and how to debug when things aren't working. is migrating this to a separate repo something you'd be interested in taking on?

Spongman:
I think it would be a shame to remove @toolness's tool from the build process, especially as it provides another really good source of validation for the doc comments. i was going to suggest that we even start having it throw errors during build to help ensure further that the doc comments are consistent during precommit. another repo would certainly work, but the lag between breakage and fix for issues like those would be increased. maybe that's not really a priority...

Zalastax:
I think it would be a shame to remove it as well, but it increases complexity so we might have to maintain it elsewhere for now. My personal preference would actually be for p5 to go ES6 modules, then slowly transition to TypeScript, but that's really not up to me.
[...]

PR #3226, opened 2018-09-20
Make YUIDoc annotations compatible with p5.ts

Zalastax:
YUIDoc comments are split where needed to generate correct overloads in data.json.


Spongman:
wouldn't it be great to have a this validated as part of the p5.js `precommit` ?

Zalastax:
@Spongman wouldn't that require a dependency towards p5.ts? It's not too much trouble to fix afterwards so I don't think it's worth the effort right now.

Spongman:
it would... now that the tool that does the validation is no longer part of this repo.

this is precisely the kind off issue i was worried about here: https://github.com/processing/p5.js/issues/2571#issuecomment-361089453

src/core/shape/vertex.js
```diff
@@ -314,17 +314,29 @@ p5.prototype.beginShape = function(kind) {
...
+/**
+ * @method bezierVertex
+ * @param  {Number} x2
+ * @param  {Number} y2
+ * @param  {Number} x3
+ * @param  {Number} y3
+ * @param  {Number} x4
+ * @param  {Number} y4
+ * @chainable
+ */
```
lmccart: is this signature the same as above in the main yuidoc entry? it can be omitted completely then. only the alternatives are listed below.

src/core/shape/vertex.js
```diff
@@ -706,15 +718,26 @@ p5.prototype.endShape = function(mode) {
...
  * backwards s-shaped black line with 4 pixel thick stroke weight.
  */
+
 /**
  * @method quadraticVertex
  * @param  {Number} cx
  * @param  {Number} cy
- * @param  {Number} [cz] z-coordinate for the control point (for WebGL mode)
  * @param  {Number} x3
  * @param  {Number} y3
```
lmccart: same here

Spongman:
in addition, you might want to double-check that the WEBGL example is making into the docs. in all the other cases where overloaded declarations occur, all the examples appear in the first yui section.

Zalastax:
Good catches @lmccart and @Spongman!
This lead me to also fix rendering of multiple examples: https://github.com/processing/p5.js-website/pull/264.

Issue #3225, opened 2018-09-20
turnAxis has no description


DaforLynx:
#### Nature of issue?

- [ x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ x] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [ x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 
- Web browser and version: Chrome, on school laptop so can't check version for whatever reason. A newer one with the new design. 
- Operating System: macOS Sierra 10.12.6 
- Steps to reproduce this:

**### turnAxis reference article has no description. That is all.**

PR #3245, opened 2018-10-01
fixed entry in contributors

L05:
Avatar URL was missing several characters and thus not appearing.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -182,7 +182,7 @@
     {
       "login": "L05",
       "name": "L05",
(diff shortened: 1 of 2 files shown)


limzykenneth:
@L05 Did you meant to include the fix for #3225 in this PR? If you want to open a seperate PR for it you can make the changes on a different branch on your fork and compare that fork with the master branch here.

lmccart:
thanks @limzykenneth. @l05 is learning about branches now :) we will merge this one and use separate branches in the future

PR #3249, opened 2018-10-04
webgl: fix cylinder normals

### LINKED ISSUE #3248: webgl: cylinder normals are wrong
0.7.2: https://codepen.io/Spongman/pen/mzPBeQ?editors=0010

image

the specular highlight err... highlights this issue. the normals for the top & bottom ring of vertices of a cylinder are wrong.

Spongman:
closes #3248 

don't re-compute normals for cylinders. the _truncatedCone already calculates the correct normals for cylinders, including their end-caps. calling `computeNormals()` is not only unnecessary, it interpolates the normals around the edges, which leads to the weird lighting artifact in the above bug.

here's the fixed result:
image


stalgiag:
Hi @Spongman thanks for this! Would you mind just deleting the unnecessary line and putting a comment to the effect of "//normals computed in call to _truncatedCone" instead? Just to avoid future confusion.

PR #3254, opened 2018-10-05
fix cone normals

### LINKED ISSUE #3250: webgl: cone normals are wrong
0.7.2: https://codepen.io/Spongman/pen/aRNLrj?editors=0010

this one's quite subtle, it's easier to see if you watch the animation above.

image

the issue is how the normals for the top section of the cone are calculated. the fix for this isn't quite as simple as removing the call to `computeNormals()` as this results in even worse artifacts.

Spongman:
closes #3250 

this simplifies the algebra somewhat and ensures that the top cap's central normal is correct.

here's the same sketch from that bug, but with the fixed library: https://codepen.io/Spongman/pen/bmwVRp?editors=0010


stalgiag:
Great job catching such a subtle bug and fixing it while simplifying the math. Thanks @Spongman !

PR #3286, opened 2018-10-27
Changes to add iPad external keyboard arrow key functionality

chrisorban:
I have made changes to keyboard.js that I believe adds iPad external keyboard (apple brand) arrow key functionality without breaking any functionality with other keyboards or systems.

Please consider these changes for merging with the main branch of p5.js.

There are still some issues with the non-arrow-keys on the iPad external keyboard but at least the arrow keys work now. For a lot of the games that I create this is the main functionality I needed. Users can now use the same keyIsDown(LEFT_ARROW), etc. commands in both iPads and regular computers. 

One can press more than one arrow key on the iPad external keyboard and it will detect it correctly in the same way that a normal keyboard would.


chrisorban:
@catarak  @lmccart    Here is a push that has the modifications to keyboard.js to include iPad external keyboard arrow key functionality. It took me a while to figure out how to comply with the style requirements in order to get through the travis check, but it worked just now. Please go ahead and merge it with the main branch.

chrisorban:
I submitted this pull request (without style errors) in November 2018. Since then, the syntax of keyboard.js changed because of the march to p5.js 1.0 and my code failed when I tried to deal with the differences with the p5.js main branch. But after some commits I got through the tests that were failing. This involved things like changing downKey[#] --> this.downKey[#]  and areDownKeys[#] --> this._areDownKeys[#]
This seems like something related to ES6 but I'm not an expert. All I know is that it just works.

So I got through those tests, but as of two weeks ago there is a unit test for the keyboard that my code is failing on. Here is some of the discussion:
https://github.com/processing/p5.js/issues/3567

I kind of wish that we could have gotten this Pull request merged before the unit test for the keyboard was developed. This is now the second oldest Pull request on p5.

lmccart:
Hi @chrisorban  thank you for your work on this. We're not able to support library modifications to support external third-party peripherals because it gets too tricky to determine which ones to add, and this is really a bug with Apple's keyboard itself. We'd suggest making your own library (as I think you have done) and including it in student projects to add the desired modifications. 

However, it does luckily look like this issue will be fixed in the iOS 13 release. https://bugs.webkit.org/show_bug.cgi?id=149054

chrisorban:
I just wanted to follow up to say that I tested the public beta of iOS 13 on my iPad Pro and and the apple brand external keyboard now works fine. EXACTLY like a regular keyboard as far as I can tell. No need to modify p5.js for the sake of the apple keyboard, just encourage folks to upgrade to iOS 13 as soon as they can.

lmccart:
@chrisorban thanks for following up with this info, that is very helpful and great news! 🙏

PR #3303, opened 2018-11-05
resolve dom dependencies for various example snippets, closes #3216

### LINKED ISSUE #3216: some base library reference examples require addon libraries
0.7.2

a few reference examples in the base p5.js library's docs fail to run if the dom addon library isn't loaded. however, there's no indication that the code requires the user to load the addon library.

```
p5.Element.elt
p5.Element.parent
p5.Element.changed
p5.Element.input
p5.Graphics.remove
p5.TypedDict.saveTable
p5.TypedDict.saveJSON
p5.winMouseX|Y
p5.createWriter
p5.saveJSON
p5.saveStrings
p5.loadFont
p5.texture
```

L05:


lib/addons/p5.dom.js
```diff
@@ -242,6 +242,139 @@
     }
   };
 
+  // General handler for event attaching and detaching
+  function adjustListener(ev, fxn, ctx) {
```
lmccart: are these three methods duplicated from p5.Element.js? if so, let's attached these functions to p5.Element as p5.Element_adjustListener and p5.Element._attachListener rather than duplicate the code in two places.

src/core/p5.Element.js
```diff
@@ -28,13 +28,14 @@ p5.Element = function(elt, pInst) {
...
-   * createCanvas(300, 500);
-   * background(0, 0, 0, 0);
-   * var input = createInput();
-   * input.position(20, 225);
-   * var inputElem = new p5.Element(input.elt);
-   * inputElem.style('width:450px;');
-   * inputElem.value('some string');
+   * function setup() {
+   *   var c = createCanvas(50, 50);
+   *   c.elt.style.border='5px solid red';
```
lmccart: the linter is complaining here, I'm not totally sure why your grunt call didn't flag it
but can you add spaces on either side of the `=` here

Issue #3243, opened 2018-10-01
Enhancement: Add more Cursors


juandavidcruzs:
#### Nature of issue?

- [ ] Found a bug
- [x] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [x] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Feature enhancement details:

There are a lot of cursors as seen on: [CSS Cursor Property W3](https://www.w3schools.com/cssref/pr_class_cursor.asp)

However I found that P5.js only has (ARROW, CROSS, HAND, MOVE, TEXT, or WAIT, or path for image)

In `constants.js` we could add new constants for new cursors such as all the resize cursors:

```javascript
/**
 * @property {String} N_RESIZE
 * @final
*/
N_RESIZE: 'n-resize',
```

```javascript
/**
 * @property {String} S_RESIZE
 * @final
*/
S_RESIZE: 's-resize',
```

etc.

SphaeriCactus:
I agree! I would love to see more cursors! :)

lmccart:
I think it might make sense to update the documentation and update native entering of strings, rather than adding more constants for each. this way we're not always falling behind. for example:
```js
cursor('s-resize');
cursor('arrow');
```

 it would be good to also include a link to the CSS cursor property reference page in the reference entry.

lmccart:
this has been addressed with #3306

PR #3306, opened 2018-11-07
DOCS: Extended documentation in core/environment.js ->@method cursor

vipulrawat:
Related to Issue: #3243
Added two things in the documentation:
1. Native string support in `cursor()` i.e. `cursor('cell')`
2. Improved example part to show all possibilities of the cursor.

_New Example:_
cursor


lmccart:
this is great, thank you!

PR #3308, opened 2018-11-08
Undeprecate shuffle()

### LINKED ISSUE #3307: undeprecate shuffle()
I agree with the thinking in #2709 but I wish you'd keep `shuffle()`—maybe reclassify it in the Math/Random section of the reference? There's no easy single-expression equivalent for `shuffle()` in JavaScript and I find myself using it in teaching and projects fairly frequently (way easier than copy/pasting whichever ramshackle fisher-yates implementation has floated to the top of a stack overflow thread).

_Originally posted by @aparrish in https://github.com/processing/p5.js/issues/2709#issuecomment-436620203_

TanviKumar:
To close #3307

### DIFF
--- a/src/utilities/array_functions.js
+++ b/src/utilities/array_functions.js
@@ -194,7 +194,6 @@ p5.prototype.shorten = function(list) {
  * Fisher-Yates Shuffle Algorithm</a>.
  *
  * @method shuffle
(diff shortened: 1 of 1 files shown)


lmccart:
eventually we may want to move this function, but I think we can sort that out when the other array functions are removed completely. thanks @TanviKumar

PR #3312, opened 2018-11-12
Added semicolons whose absence was bugging me

rMazeiks:
Also made a random comment more clear, and added spaces before curly braces


lmccart:
thanks! the changes actually need to be applied here: http://github.com/processing/p5.js-sound in order to stick. do you mind making an additional PR over there?

rMazeiks:
ok, done. Thanks!

Issue #3240, opened 2018-09-28
Feature Request: Return heading between two vectors with v1.heading(v2)


PhilipRL:
#### Nature of issue?

- [ ] Found a bug
- [x] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Details about the bug: 

- p5.js version: 0.7.2

#### Feature enhancement details:
I would find it useful to be able to use heading() to calculate the heading between two vectors by passing another vector as an argument - just like dist() is used to find a distance between two vectors.

If v1 and v2 are vectors, an example would be:
`let a1 = v1.heading(v2);`

Today I solve this by using:
`let a1 = v1.copy().sub(v2).heading();`

Please observe that this is NOT equal to `v1.angleBetween(v2);`!

pakosaldanaort:
@lmccart @PhilipRL  I will work on this  Existing feature enhancement

lmccart:
I'm not seeing a lot of interest in this feature, and we don't have a clear explanation of how it [expands access](https://github.com/processing/p5.js/blob/main/contributor_docs/access.md), so I will close this for now. If an access statement can be added to the issue request, please feel welcome to reopen.

PR #3313, opened 2018-11-12
Add method to calculate heading between two vectors

pakosaldanaort:
I wrote the method **headingVectors()**  to calculate the heading between two vectors, this solve the Feature Request #3240

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -1078,7 +1078,7 @@ p5.Vector.prototype.heading = function heading() {
  * function setup() {
  *   var v1 = createVector(30, 50);
  *   var v2 = createVector(40, 80);
- *   print(v1.headingBetween(v2)); // -0.49394136892
+ *   print(v1.headingBetween(v2)); // 0.07677189126977804
  * }
  * </code>
  * </div>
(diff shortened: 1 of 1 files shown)


lmccart:
Thanks! I think this method could just be called "heading".
We will need inline documentation in order to add this. See guidelines [here](https://github.com/processing/p5.js/blob/master/developer_docs/inline_documentation.md)

pakosaldanaort:
@lmccart  I already added documentation and change the method name to "headingBetween", I didn't change the name to "heading" because javascript is not supported the Overloading

vipulrawat:
@pakosaldanaort Is this supposed to calculate the angle between two vectors `v1` and `v2` or two create a heading vector for a specific angle?

pakosaldanaort:
@vipulrawat is to calculate the angle of rotation between two vectors

pakosaldanaort:
I will work on that @vipulrawat Thanks for your comments :+1:

vipulrawat:
@pakosaldanaort Cool. I am currently assigning WIP label so that others will know that you are currently working on this PR. Remove the label when you're done. :) :+1:

pakosaldanaort:
@vipulrawat @lmccart I already updated the documentation and the examples :+1:

vipulrawat:
@pakosaldanaort Code looks good to me. But is there any special need for this function when the same results can be obtained with [this](https://p5js.org/reference/#/p5.Vector/angleBetween) method. 
Because the results of this code is quite different from `let a1 = v1.copy().sub(v2).heading();` mentioned in #3240. What I am missing?

lmccart:
@pakosaldanaort can you address @vipulrawat's question? we will need to understand this in order to consider merging. thanks

Spongman:
i think the `atan2` method is superior since it avoids the two `sqrt` calls.

even _if_ this isn't merged, i think the `angleBetween` method should use this trig. or if this _is_ merged, the `angleBetween` method should be re-implemented as a call to this method plus some `abs` & 2PI-modulo.

outofambit:
I'm going to close this since this hasn't seen any activity in quite a while. @pakosaldanaort please comment or reopen if you would like to talk more. thanks!

PR #3327, opened 2018-11-19
Update 2d_primitives.js

unjust:
Found the docs a little confusing when trying to create a quad with z coordinates.
Clarify the ordering of arguments for quad.

If I'm not posting this correctly - to the correct repo or into the correct branch, let me know. (I just followed the links at the bottom of the docs)


Spongman:
i'm afraid this isn't going to work. the ordering of the parameters in the docs must match the ordering in the code.

unjust:
ok I must be doing something wrong then because the params don't work as expected for me.

unjust:
Right, any additional args past 8 just get ignored. At first I thought it was an ordering problem.

lmccart:
ok great. closing this and we can follow up with the issue you've posted. linking here for documentation https://github.com/processing/p5.js/issues/3328

Issue #1, opened 2013-05-02
IDEs


fjenett:
Hey Lauren,

not sure how to add things here in an open way, so trying issues for now.

About IDEs: there are some open IDEs around that i think might be interesting for your research here as they embrace JS directly:

Komodo Edit (free, open source version) based on Mozilla XULRunner (like Firefox), comes with SpiderMonkey JS engine included and uses Scintilla as code editor:
http://www.activestate.com/komodo-edit

ICE Coder (pure browser IDE):
http://icecoder.net/

PR #3329, opened 2018-11-20
Support for 3d drawing in quad primitives, fixes #3328

### LINKED ISSUE #3328: 3D renderer only has a 2D implementation of `quad`
Trying to pass z coordinates to `quad` for 3d rendering doesn't work, there is no implementation for 12 args (which would contain `z`)

 https://github.com/processing/p5.js/blob/9851414f747384d708a556285876bc585a2c9e4f/src/webgl/3d_primitives.js#L1027-L1064

_Originally posted by @meiamsome in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0MjMyMDgxNzM4/pull_requests/unread_timeline#issuecomment-440012991_

unjust:
Docs for `quad` say it has support for z args, but it didn't.

- Added support for 12/8 args
- fixed a prettier error that wouldn't let me push

### DIFF
--- a/src/webgl/3d_primitives.js
+++ b/src/webgl/3d_primitives.js
@@ -1167,7 +1167,7 @@ p5.RendererGL.prototype.curve = function(
   for (var i = 0; i <= curveDetail; i++) {
     var c1 = Math.pow(i / curveDetail, 3) * 0.5;
     var c2 = Math.pow(i / curveDetail, 2) * 0.5;
-    var c3 = (i / curveDetail) * 0.5;
+    var c3 = i / curveDetail * 0.5;
     var c4 = 0.5;
     var vx =
       c1 * (-x1 + 3 * x2 - 3 * x3 + x4) +
--- a/src/webgl/3d_primitives.js
+++ b/src/webgl/3d_primitives.js
@@ -1024,30 +1024,61 @@ p5.RendererGL.prototype.rect = function(args) {
   return this;
 };
 
-p5.RendererGL.prototype.quad = function(x1, y1, x2, y2, x3, y3, x4, y4) {
+p5.RendererGL.prototype.quad = function(
+  x1,
+  y1,
+  z1, // x2
+  x2, // y2
+  y2, // x3
+  z2, // y3
+  x3, // x4
+  y3, // y4
+  z3,
+  x4,
+  y4,
+  z4
+) {
+  if (arguments.length === 8) {
+    x4 = x3;
+    y4 = y3;
+    x3 = y2;
+    y3 = z2;
+    y2 = x2;
+    x2 = z1;
+    z1 = z2 = z3 = z4 = 0;
+  }
+
   var gId =
     'quad|' +
     x1 +
     '|' +
     y1 +
     '|' +
+    z1 +
+    '|' +
     x2 +
     '|' +
     y2 +
+    '|' +
+    z2 +
     '|' +
     x3 +
     '|' +
     y3 +
     '|' +
+    z3 +
+    '|' +
     x4 +
     '|' +
-    y4;
+    y4 +
+    '|' +
+    z4;
   if (!this.geometryInHash(gId)) {
     var _quad = function() {
-      this.vertices.push(new p5.Vector(x1, y1, 0));
-      this.vertices.push(new p5.Vector(x2, y2, 0));
-      this.vertices.push(new p5.Vector(x3, y3, 0));
-      this.vertices.push(new p5.Vector(x4, y4, 0));
+      this.vertices.push(new p5.Vector(x1, y1, z1));
+      this.vertices.push(new p5.Vector(x2, y2, z2));
+      this.vertices.push(new p5.Vector(x3, y3, z3));
+      this.vertices.push(new p5.Vector(x4, y4, z4));
       this.uvs.push(0, 0, 1, 0, 1, 1, 0, 1);
       this.strokeIndices = [[0, 1], [1, 2], [2, 3], [3, 0]];
     };
@@ -1136,7 +1167,7 @@ p5.RendererGL.prototype.curve = function(
   for (var i = 0; i <= curveDetail; i++) {
     var c1 = Math.pow(i / curveDetail, 3) * 0.5;
     var c2 = Math.pow(i / curveDetail, 2) * 0.5;
-    var c3 = i / curveDetail * 0.5;
+    var c3 = (i / curveDetail) * 0.5;
     var c4 = 0.5;
     var vx =
       c1 * (-x1 + 3 * x2 - 3 * x3 + x4) +


unjust:
A few things since this is my first pull with this project:

I had a lot of warnings when I ran `grunt`, mostly to do with prettier warnings in files I didn't touch, and 
```No parser and no filepath given, using 'babylon' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.```

Not sure if this is regular noise in the project or something I need to fix.
 
While looking at some other code in this file, like `bezier`, I think I noticed something weird with the way it parses 8 args - but since I haven't used `bezier` I'm not sure if its actually buggy.

File: src/webgl/3d_primitives.js [hunk 1]
```diff
@@ -1136,7 +1167,7 @@ p5.RendererGL.prototype.curve = function(
   for (var i = 0; i <= curveDetail; i++) {
     var c1 = Math.pow(i / curveDetail, 3) * 0.5;
     var c2 = Math.pow(i / curveDetail, 2) * 0.5;
-    var c3 = i / curveDetail * 0.5;
+    var c3 = (i / curveDetail) * 0.5;
```
meiamsome:
```suggestion
    var c3 = i / curveDetail * 0.5;
```
This is why Travis is failing, these brackets are overly verbose. Perhaps we could go for this instead which might be clearer:
```suggestion
    var c3 = 0.5 * i / curveDetail;
```

unjust: I used the “apply suggestion” feature. I think this is the way the file was. 
I originally made the change because Prettier wasn’t allowing me to commit.

unjust:
@meiamsome Please let me know of next steps, if any. I don't have write access, so I guess I wait for someone from the organization to merge.

meiamsome:
@unjust
Looks good to me, we don't have any unit tests on the drawing primitives themselves that I know of, only ones that assert things on the arguments, but that's at a higher level that would miss these cases anyway.  

Thanks for your PR, someone who has the ability to merge will look at it soon!

Re: 1. I'm not sure about this 
Re: 4. I assume you mean that `y3 = x2;` in the `bezier` should be `y3 = z2;`? If so, I agree that looks like a bug to me, you can produce an error example for this and open a new Issue/PR if you'd like?

unjust:
@meiamsome I fixed #1 by using `npm` instead of `yarn` *shrug*.
#4 is a bug, I have a fix and will open an issue.

src/webgl/3d_primitives.js
```diff
@@ -1024,30 +1024,61 @@ p5.RendererGL.prototype.rect = function(args) {
   return this;
 };
 
-p5.RendererGL.prototype.quad = function(x1, y1, x2, y2, x3, y3, x4, y4) {
+p5.RendererGL.prototype.quad = function(
```
lmccart: generally with functions with variable number of args, we leave the arguments unnamed and test arguments.length. I think this could help condense the code a bit. would you be up for updating this function in this way?

unjust: I feel like that would be un-doing the work from this commit 11 months ago
https://github.com/processing/p5.js/commit/41025510d7ddeec704613696eaa283291a69e5da
https://github.com/processing/p5.js/commit/41025510d7ddeec704613696eaa283291a69e5da#diff-dfd440848f7f2b0c2e4a2f2a19aa79b1R834

Not sure what the motivation was from changing it from `arguments` to named `args`.

lmccart: @Spongman can you elaborate on the reasoning on this commit?

Spongman: yeah, mostly because it's significantly simpler to read code using named arguments than using the indexed `arguments` array (whose use is generally considered an anti-pattern unless absolutely necessary). most of those methods (while possibly having optional parameters) do not have overloaded meaning for their arguments, so the use of the `arguments` array is not necessary. also because the use of the `arguments` array can also have significant performance implications - often it excludes the function from being JIT compiled and/or inlined.
i think it's important to distinguish between functions that have variable _numbers_ of arguments, and those that have variable _types_ (or meanings) of arguments. the former don't really need to use the indexed `arguments` array. for the latter, it makes more sense.

lmccart: Thanks @Spongman. As a general rule from the beginning, we've actually used named arguments only when there are not variable numbers nor variables types of arguments. That is, there is only one signature. I realize that this isn't entirely consistent throughout, but I'd like to stay with this for now. We are planning an overhaul to ES6 soon, and we can revisit again soon.

Spongman: hmm... i wasn't aware of that. currently, in the p5 library codebase, the use of the `arguments` array is quite definitely the exception rather than the rule. a regex search for ` \* .*@.*.*\[` will show those functions that take optional parameters. very few of them use the `arguments` array.

lmccart: yes, like I said it is not consistent. but I would like to stick to this rule where we can going forward, until we have the ES6 update conversation.

lmccart: Actually, maybe I wasn't clear, or I mixed that up. If the parameters are optional at the end, where the order remains the same, this would be fine to use named ones. But in this case, we're actually talking about a different ordering, as far as I can tell? So we can stick with arguments here. I think we'll actually also want to update this PR with [multiple signatures](https://github.com/processing/p5.js/blob/master/developer_docs/inline_documentation.md#additional-signatures) in the reference too, for the 2D and 3D versions.

Spongman: yup, that's what i was referring to when i said

> i think it's important to distinguish between functions that have variable numbers of arguments, and those that have variable types (or meanings) of arguments

above.

lmccart: I see, got it. I thought you meant different types as in numbers vs strings etc. Cool looks like we're on the same page then. Thanks for clarifying.

Spongman: indeed. sorry, i probably wasn't clear enough before... i'd like to think that ES6 will help with all this, but unfortunately function overloading was left off the table this time around (AS3 didn't support it either).

lmccart: yeah, it won't exactly save us. but I think it will be a good time to document a few of these guidelines together so there's more clarity around code style and patterns.

unjust: Ok I can't tell where we've landed. I think we've argued for not using `arguments` since this function only has a variable number of args?

> i think it's important to distinguish between functions that have variable numbers of arguments, and those that have variable types (or meanings) of arguments. the former don't really need to use the indexed arguments array.

I think the named arguments would be clearer for now. And any new pattern can be agreed on when we can write in ES6.

lmccart: @unjust we've agreed that named arguments are used in the case where there are either (1) only one signature possible or (2) a variable number of arguments but always the same order and meaning. the arguments object is used in cases where there are variable arguments with different orderings.

so, please remove the named arguments and use the arguments object instead.
also, we will need to add an [additional signature](https://github.com/processing/p5.js/blob/master/developer_docs/inline_documentation.md#additional-signatures) to distinguish the two use cases.

once these two updates have been made, this code can be merged. thanks @unjust!

unjust:
Going to open a new pull request. Realized I should be using feature branches on my fork.

PR #3334, opened 2018-11-21
Changed in release-docs grunt task

vipulrawat:
1. In the task to copy files the error of `No such file or directory` happened everytime even the files were there (and the commands runs perfectly when done manually).
I couldn't find the exact reason for the errors. But now the given solution works. 
2. In the last promise chain even if the error occurs the `console.log('Released on Github')`.

_Error:_
deepinscreenshot_select-area_20181122041052

_Update:_ 
The reason for the error is due the default shell that node child_process api uses is `sh` shell and `sh` shell due to its less feature set is not supporting the copying multiple files using `{file1, file2, file3}` this format. Thus new grunt file changes will target every user either using macos or linux.

### DIFF
--- a/developer_docs/release_process.md
+++ b/developer_docs/release_process.md
@@ -5,6 +5,7 @@
 
 
 ## Requirements
+* Operating System: macOS, Linux or Unix. _(For Windows see at the end of this document)_
 * Logged in NPM CLI : Check if you are logged in by `npm whoami`
 * High Bandwidth : Lots of things to download/pull/push (~190 MB total I presume)
 * An environment variable named __GITHUB_TOKEN__ with the value of your Access Token : Make one by going [here](https://github.com/settings/tokens). Once you have it, in your shell, run `export GITHUB_TOKEN=<your token here>`.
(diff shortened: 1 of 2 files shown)


vipulrawat:
For me this command `cp -r ...src/{data.json,data.min.json,assets} ...dest` is only working in `bash` and `zsh` not in `sh`.

limzykenneth:
@lmccart The step works fine and the files do seemed to be copying correctly. Depending on whether you want full cross platform compatibility or not, we can consider replacing the shell `cp` for a bit of node `fs` code.

vipulrawat:
Yes, If we are targeting for windows users too than using `fs` will be helpful in achieving cross compatibility.
And for linux and mac I think majority of users are in either bash or some newer shell.

lmccart:
Thank you @vipulrawat and @limzykenneth for the review. I think it is ok to not support every platform since it's not a task that's necessary for development in general. However, @vipulrawat could you update the [release process dev doc](https://github.com/processing/p5.js/blob/master/developer_docs/release_process.md) to note this requirement somewhere. And also update anything else related to your changes (if needed). Then this PR should be good to merge.

Issue #3337, opened 2018-11-23
Optimize dist() with distsq() for comparison, without computing square root


ThatIsAPseudo:
#### Nature of issue?

- [ ] Found a bug
- [x] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Feature enhancement details:
The `dist()`function should have an optionnal parameter (or a similar function) in the case we want to get the sqare of the distance between two points, that would avoid computing a square root when we just want to use it in a comparison, i.e. when we don't really want ton know the exact value.

For instance, a simple `distsq()` would complete this task :
```javascript
function distsq(x1, y1, x2, y2) {
   return sq(x1 - x2) + sq(y1 - y2);
}
```
`distsq()` should work in 3D as well.

Remcodejong:
> `distsq()` should work in 3D as well.


This would require the z value of the point as well. Based off the original dist function in p5:
```javascript
e.prototype.distSq = function() {
return e._validateParameters("distSq", arguments),
4 === arguments.length ? return Math.sqrt (d(arguments[2] - arguments[0], arguments[3] - arguments[1])) : 6 === arguments.length ? return Math.sqrt(d(arguments[3] - arguments[0], arguments[4] - arguments[1], arguments[5] - arguments[2])) : void 0
}
````
Something like this?

**Comment by Remcodejong** [comment]
I think that would be exactly the same as the normal distance function, but I could try to rework the suggestion, and ill get back to you

**Comment by ThatIsAPseudo** [comment]
It should be the sum of the square of the differences, shouldn't it? Just like in the `dist` function but without applying `sqrt`

**Comment by Remcodejong** [comment]
Like the variable r2 here? 
https://editor.p5js.org/Remcodejong/sketches/HktaQoYC7

**Comment by Spongman** [comment]
Doesn’t `p5.Vector.dot()` already cover this?

**Comment by ThatIsAPseudo** [comment]
Yeah, exactly like this :
```javascript
let dx = x2-x1;
let dy = y1-y2;
d = (pow(dx,2)+pow(dy,2));
text("dist squared is " + floor(d),10,220);
```

Remcodejong:
[...]

PR #3345, opened 2018-11-26
Implementing distSq() #3337

Remcodejong:
This commit adds the distSq() function as used in #3337.
Tested with the example in #3337 with the distSq() function, instead of line 46 and 47 (from the example). 
Will put in this example later.

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -188,6 +188,50 @@ p5.prototype.dist = function() {
 };
 
 /**
+ * Calculates the distance between two points and squares this value.
+ *
+ * @method distSq
+ * @param  {Number} x1 x-coordinate of the first point
+ * @param  {Number} y1 y-coordinate of the first point
+ * @param  {Number} x2 x-coordinate of the second point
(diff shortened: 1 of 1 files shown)


src/math/calculation.js
```diff
@@ -187,6 +187,50 @@ p5.prototype.dist = function() {
...
+ * @param  {Number} y2
+ * @param  {Number} z2 z-coordinate of the second point
+ * @return {Number}    distance between the two points, squared
+ */
+p5.prototype.distSq = function() {
+  p5._validateParameters('distSq', arguments);
+  if (arguments.length === 4) {
+    //2D
+    return (
+      Math.pow(arguments[2] - arguments[0], 2) +
```
Spongman: i would recommend against using `Math.pow()` if at all possible. it is extremely expensive, especially compared to multiplying a number by itself.

Remcodejong: Ill change it to just the numbers multiplying themselves.

Remcodejong:
Ill close this pull request, ad this function will just be part of a separate library.

PR #3347, opened 2018-11-27
 Use `DOMContentLoaded` instead of `onload` for initializing p5

calebegg:
Documentation:
https://developer.mozilla.org/en-US/docs/Web/Events/DOMContentLoaded

MDN says: 
> It is an incredibly common mistake to use `load` where `DOMContentLoaded` would be much more appropriate

DOMContentLoaded is supported in all modern browsers (IE9+), and can make a
significant difference in page load performance when a page has many
external assets, such as images and iframes.

/cc @jesse-harding


limzykenneth:
Are we sure we don't need any images or stylesheets that is requested in the DOM before p5 starts? I'm not entirely sure myself but we need to make sure that is the case first.

outofambit:
> Are we sure we don't need any images or stylesheets that is requested in the DOM before p5 starts? I'm not entirely sure myself but we need to make sure that is the case first.

@calebegg any thoughts on this? i'm doing a little research to see if this makes sense for p5.

calebegg:
It's up to you. I wrote this patch for a friend and he's used it successfully to significantly speed up his site, so I decided to offer it to you as a courtesy. I've included a link to MDN's explanation of the differences.

I cannot imagine a scenario where this change would break a p5 sketch, but I've experienced breakages with even the "safest" of changes in large projects, as I'm sure have y'all.

Spongman:
my only concern would be drawing text in `setup()` using css-loaded fonts. although, i'm not even sure if this works right now...

meiamsome:
@Spongman wouldn't this kick off the `draw` loop too, even with unloaded assets, which might be problematic.

This could also be a problem for people who are perhaps loading data from image or video elements that are included in the page, right?

Perhaps we could have a `loadPage()` preload method that forces p5 to wait for the `load` event on the page as an opt-in to ensure all such assets exist.

outofambit:
thanks @calebegg!

> This could also be a problem for people who are perhaps loading data from image or video elements that are included in the page, right?

this is my main concern, too @meiamsome. 🤔 i'll put together a test case soon

Spongman:
i believe `HTMLVideoElement` only delays the `load` event if its `delaying-the-load-event flag` attribute is `true`.

stalgiag:
Hm I feel like this doesn't technically speed anything up, it just changes the order for p5 to start. [This test seems like a good explanation of the difference.](http://web.archive.org/web/20150405114023/http://ie.microsoft.com/testdrive/HTML5/DOMContentLoaded/Default.html) Should p5 wait for external assets to load before initializing? I kind of feel that it should. I think it would be rare for someone to need a p5 sketch to load first on a page with a ton of external assets. Whereas I can imagine it being more common for someone to try to grab pixel data from a DOM img in `setup`. I think it is more beginner friendly to handle the latter case.

jesse-harding:
So that which prompted caleb to help me modify the p5.dom.js file was that
I was using a p5 sketch as a banner for my website which includes multiple
embedded Vimeo videos. With the unmodified version, all of the videos load
slowly (and visibly as their relative positioning has not been moved down
by the div container for the p5 sketch being populated).

I love the idea of having a preload method that makes the order
more-visibly optional and not buried.

Thanks for the attention!
-Jesse

stalgiag:
Ah gotchya thanks for the example. I agree that a preload method would be cool. I guess the question then is which should be the default behavior. Load first as in this PR or wait for all external assets in the page to load as in current behavior? There isn't an obvious favorite to me.

jesse-harding:
It seems like the current order makes sense as the default since the
problem in my use case is purely aesthetic and not about getting info from
assets.

stalgiag:
Thanks for the PR and discussion starter @calebegg. I am closing this for now [as I opened an issue](https://github.com/processing/p5.js/issues/3536) that points to the feature request discussed in this PR.

PR #3358, opened 2018-12-04
Support for 3d drawing in quad primitives

### LINKED ISSUE #3328: 3D renderer only has a 2D implementation of `quad`

unjust:
Fixes #3328. Started this PR here https://github.com/processing/p5.js/pull/3329, if you want to read prev discussion.

----

Docs for [quad](http://p5js.org/reference/#/p5/quad) say it has support for z args, but it doesn't.
Added support for 12 args (with z).

*This fixes only when specifying `WEBGL` in canvas mode.
Passing in 12 args without specifying `WEBGL` isn't supported. 
Is it intended to support z in 2D renderer drawing? (sorry if this question doesn't make sense.)

Test code:

```
 <script type="text/javascript" >
        function setup() {
            createCanvas(300, 300, WEBGL);
        }

        function draw() {
            fill('red');
            quad(38, 31, 86, 20, 69, 63, 30, 76);
            translate(50, 50);
            quad(38, 31, 0, 86, 20, 0, 69, 63, 0, 30, 76, 0);
            fill('blue');
            quad(38, 31, -100, 86, 20, -100, 69, 63, -100, 30, 76, -100);
        }
</script>
```
*Before*

screen shot 2018-12-03 at 10 45 09 pm

*After*

screen shot 2018-12-03 at 10 40 53 pm


unjust:
I followed prettier with the indentation recommendations.
Also let me know if there are other conventions I'm not following.

If I can help with `es6` transition please let me know!

unjust:
@outofambit ping! what more should go on with this?

Spongman:
can we do this without the array creation and/or the `forEach` usage? these higher-order functions are very handy, but they're also very expensive.

stalgiag:
@Spongman has a good point since the array creation and `forEach` will occur every frame in most sketches. @unjust maybe you could try separating out how the `_renderer.quad` functions are called in `p5.prototype.quad`? You can check to see if the renderer is 3d there and then pass either zeros or the appropriate arguments depending on the provided argument length. That way `RendererGL.prototype.quad` can always expect 12 arguments. There are other ways to solve this of course but that is one very straightforward and legible option.

src/webgl/3d_primitives.js
```diff
@@ -1010,30 +1010,51 @@ p5.RendererGL.prototype.rect = function(args) {
...
+  x2,
+  y2,
+  z2,
+  x3,
+  y3,
+  z3,
+  x4,
+  y4,
+  z4
+) {
```
unjust: linter/prettier made me format args like this ^

src/core/shape/2d_primitives.js
```diff
@@ -374,6 +374,14 @@ p5.prototype.point = function() {
 p5.prototype.quad = function() {
   p5._validateParameters('quad', arguments);
 
+  // if 3D and we weren't passed 12 args, assume Z is 0
+  if (this._renderer.isP3D && arguments.length !== 12) {
+    Array.prototype.splice.call(arguments, 2, 0, 0);
+    Array.prototype.splice.call(arguments, 5, 0, 0);
+    Array.prototype.splice.call(arguments, 8, 0, 0);
+    Array.prototype.push.call(arguments, 0);
+  }
```
unjust: is this the best way to alter `arguments`?

unjust:
@outofambit @Spongman - hopefully this is more on the right track?

src/core/shape/2d_primitives.js
stalgiag: Hi @unjust This is a good start! The only issue is that Array.prototype.slice creates a new array object every time it is called ([the removed elements are returned or if nothing is removed an empty array is returned](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)). I think that you can safely do something such as:
```
this._renderer.quad.call(
  this._renderer,
  arguments[0], arguments[1], 0, 
  arguments[2], arguments[3], 0, 
  arguments[4], arguments[5], 0, 
  arguments[6], arguments[7], 0
)
```
If I am not mistaken you don't need to modify the arguments array. Also, this maybe makes it easier for someone else to quickly understand what it is that you are doing.

src/webgl/3d_primitives.js
stalgiag: You can use `// prettier-ignore` above a line when you feel confident that it is easier to read without the prettier modifications.

src/core/shape/2d_primitives.js
unjust: thanks ok! not sure why I didn't think of that.

unjust: ok I was altering them because this line in turn uses `arguments`

``` if (this._renderer._doStroke || this._renderer._doFill) {
    this._renderer.quad.apply(this._renderer, arguments);
  }```

I'm not too sure what that line does, do I need to think about the call that's happening there?

Comment (stalgiag) [reply src/core/shape/2d_primitives.js]: I understand the confusion. `call` and `apply` are similar. [Here is a short article that explains the difference.](http://www.syntaxsuccess.com/answer-question/556364e45bd0419c3c4cecea/difference-between-call-and-apply). In my opinion, it is easier to read if you use `call` instead of modifying the existing array and using `apply`. Then you can check if `_renderer.isP3D` is true, and if it is then either use this line above `quad.apply(this._renderer, arguments` or the line I suggested depending on whether the `arguments.length` is 12 or 8 respectively. Does that make sense?

Comment (unjust) [reply src/core/shape/2d_primitives.js]: That makes sense. 
Honestly I overlooked your change from `apply` -> `call` the first time - so I get what you are saying now. Let me know if this is better.

also is there any other way communicate on the p5 project or just through @ mentions?

File: src/core/shape/2d_primitives.js [hunk 3]
```diff
@@ -375,7 +375,18 @@ p5.prototype.quad = function() {
...
+      // if 3D and we weren't passed 12 args, assume Z is 0
+      // prettier-ignore
+      this._renderer.quad.call(
+        this._renderer,
+        arguments[0], arguments[1], 0,
+        arguments[2], arguments[3], 0,
+        arguments[4], arguments[5], 0,
+        arguments[6], arguments[7], 0);
+    } else {
+      this._renderer.quad.apply(this._renderer, arguments);
```
stalgiag: These changes are great! This works well. My only remaining concern is that really unexpected things happen when you are in 2D and you try to pass 12 arguments. I know this isn't new to your pull request as this problem occurs even with quad as-is with the current release. That said, since you are already doing so much to improve it might as well put in a polite check here and if the rendererer isn't 3D and the argument length is 12 output an informative error to the console. What do you think?

stalgiag: Perhaps also put something in the description above. Something to the effect of "z-arguments only work when quad() is used in WEBGL mode"

unjust: ok, I guess its a good idea. I do wonder if we need to do that much checking on behalf of the user.

so have a different if/else before this one? 
or within this last else, just check the arguments, output the error, and call the apply?

stalgiag: I tend to err on the side of caution when functions work differently between 2D and 3D. That said I think you are right on this one. If you just add the explanation that z-arguments only work in WEBGL (see comment below) that should be enough. I think using quad instead of rect already implies an understanding of the coordinate system. Thanks for hearing me out!

unjust: ok hope I put it in the right place.

unjust:
Also @stalgiag - thanks for taking me through this. I'd like to contribute more when I can, and at a quicker pace. 
Right now I feel like I can only work on small tickets in Issues or submit any bugs I find on my own. And I'm depending on prettier and PR feedback for styleguide guidance.

Let me know if I'm missing something where contributors can tune in more to whats now priority. Or any resources to learn about what the js style is. Thanks!

stalgiag:
@unjust this is great! Thanks for the patience with all of the small modification requests.

As for code style, [this is an open issue currently](https://github.com/processing/p5.js/issues/105) we are shooting to have this be part of the 1.0 release. I rely on the linter for most of the formatting concerns as well. Style is a bit of tricky word because it can be expanded to include everything about the solutions that are chosen for particular problems. I think that we are often trying to balance a few different needs:
- a need to have fairly fast code in certain sections of the code base
- a need to have legible, maintainable code
- a need to have a codebase that isn't terribly intimidating to new contributors
I personally never know the right approach to covering all of those bases, so it kind of just comes down to writing something that seems to fit in with the existing code base fairly well and then seeing how people react to it once it is a pull request.

Thanks again for taking care of this!

lmccart:
well put @stalgiag :)

Issue #3376, opened 2018-12-11
Add square() function


L05:
Add square() function for K-12 education/accessibility, as suggested in [this thread](https://github.com/processing/p5.js/issues/2305).

#### Nature of issue?

- [X] New feature request

#### Most appropriate sub-area of p5.js?

- [X] Core/Environment/Rendering

lmccart:
this method could just be an alias that calls `p5.prototype.rect(x, y, s, s)`

montoyamoraga:
would that be adding to the file the src/core/shape/2d_primitives.js the following?
I am trying to figure it out.

p5.prototype.square = function(x, y, s) {
  p5.prototype.rect(x, y, s, s);
}

lmccart:
@montoyamoraga I think this would do it, though I think you want to use `this.rect()` rather than `p5.prototype.rect()` so rectMode() is taken into consideration. and then adding documentation too. documentation could be based on the rect() docs and modified.

lmccart:
resolved with https://github.com/processing/p5.js/pull/3393

PR #3393, opened 2018-12-11
add square() function as alias of rect()

montoyamoraga:
Addressses #3376 
I created a square() function as an alias of rect() and added docs based on the original rect() one.
It is my first time contributing a p5.js function so I'm not sure if I did it right.


lmccart:
looks great, thanks @montoyamoraga!

PR #3394, opened 2018-12-11
ES6/Babelify

Spongman:
- uses `let`/`const` instead of `var`.
- uses lambdas instead of inline functions, where appropriate.
- updates eslint to es6-compatible version
- adds no-var & prefer-const eslint rules, fixes errors
- runs babelify over the browserified library, converts es6 to es5 (so generated library remains compatible)


lmccart:
this is not linked to any open issue

lmccart:
@Spongman We've warned you several times about submitting large codebase changes / refactors without prior discussion on a github issue. It creates an unfair burden on maintainers, disrespects the community process for contributing code, and is just not a sustainable way to move this project forward. I have tried to initiate offline discussions where we could talk through this further and clarify misunderstandings, and you have refused to engage. If this behavior continues, we will have to unfortunately suspend your access to the p5.js repositories.

Spongman:
oh, i'm sorry. i'm just trying to help.

PR #3401, opened 2018-12-13
Change Shape examples to ES6

### LINKED ISSUE #3370: Update Shape Reference Examples on p5js.org to ES6
Update the **Shape** section of the [p5js reference](https://p5js.org/reference/) to ES6. This will entail updating all of the references in source files that have `@module Shape` at the top.

AdilRabbani:
Changes all 'var' to 'let' in reference examples for `@module` Shape. Attempts to close #3370 Please let me know if I missed any other files or there are more changes to be introduced before this can be committed.
It seems that lint is causing a parse error with these changes, which leads to failed test cases.

capture

### DIFF
--- a/src/core/shape/curves.js
+++ b/src/core/shape/curves.js
@@ -155,21 +155,21 @@ p5.prototype.bezierDetail = function(d) {
  * <div>
  * <code>
  * noFill();
- * var x1 = 85,
+ * let x1 = 85,
   x2 = 10,
   x3 = 90,
   x4 = 15;
- * var y1 = 20,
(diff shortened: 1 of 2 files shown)


AdilRabbani:
I guess [this](https://stackoverflow.com/questions/40875615/eslint-first-instance-of-let-results-in-unexpected-token-error) is what needs to be done? @TanviKumar

Spongman:
i have a fix for this coming soon...

Spongman:
#3403 fixes this issue.

PR #3404, opened 2018-12-13
Updated Image Reference Examples on p5js.org to ES6

### LINKED ISSUE #3387: Update Image Reference Examples on p5js.org to ES6
Update the **Image** section of the [p5js reference](https://p5js.org/reference/) to ES6. This will entail updating all of the examples that fall under `@module Image`.

figraham:
Fixes - #3387 

- Updated all `@examples` for `@module Image` to ES6
- Changed var to let
- Changed callbacks to `=>` notation
- Changed idx to index in two examples to make it easier to comprehend
- Fixed `<code></code>` block by changing it to `` ```javascript``` `` so it displays correctly

### DIFF
--- a/src/image/pixels.js
+++ b/src/image/pixels.js
@@ -29,15 +29,15 @@ require('../color/p5.Color');
  * contain the R, G, B, A values of the pixel at (1, 0). More generally, to
  * set values for a pixel at (x, y):
  * ```javascript
- * var d = pixelDensity();
- * for (var i = 0; i < d; i++) {
- *   for (var j = 0; j < d; j++) {
+ * let d = pixelDensity();
+ * for (let i = 0; i < d; i++) {
+ *   for (let j = 0; j < d; j++) {
  *     // loop over
- *     idx = 4 * ((y * d + j) * width * d + (x * d + i));
- *     pixels[idx] = r;
- *     pixels[idx+1] = g;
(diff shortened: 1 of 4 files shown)


src/image/pixels.js
```diff
@@ -427,17 +427,17 @@ p5.prototype.filter = function(operation, value) {
  * Getting the color of a single pixel with get(x, y) is easy, but not as fast
  * as grabbing the data directly from <a href="#/p5/pixels">pixels[]</a>. The equivalent statement to
  * get(x, y) using <a href="#/p5/pixels">pixels[]</a> with pixel density d is
- * <code>
- * var x, y, d; // set these to the coordinates
- * var off = (y * width + x) * d * 4;
- * var components = [
+ * ```javascript
```
lmccart: I don't think this will render correctly. have you tested it?

figraham: Yeah I tested it. It currently displays as code but is hard to read because it doesn't have any line breaks you can see that here:

currentget

I chose to match the syntax that I saw in other places throughout the image docs. Here's the suggested change displayed with `npm run docs:dev`:

newget

I can change it back if you like but I think this is easier to understand.

lmccart: this looks good. thanks for posting the screenshot.

PR #3405, opened 2018-12-14
Convert Math reference examples to ES6

### LINKED ISSUE #3389: Update Math Reference Examples on p5js.org to ES6
Update the **Math** section of the [p5js reference](https://p5js.org/reference/) to ES6. This will entail updating all of the examples that fall under `@module Math`.

vedhant:
this pull request resolves #3389

### DIFF
--- a/src/math/noise.js
+++ b/src/math/noise.js
@@ -74,23 +74,23 @@ var perlin; // will be initialized lazily by noise() or noiseSeed()
  * @example
  * <div>
  * <code>
- * var xoff = 0.0;
+ * let xoff = 0.0;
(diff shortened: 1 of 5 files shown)


src/math/noise.js
```diff
@@ -209,17 +209,17 @@ p5.prototype.noise = function(x, y, z) {
  * @example
  * <div>
  * <code>
- * var noiseVal;
- * var noiseScale = 0.02;
+ * let noiseVal;
+ * let noiseScale = 0.02;
```
lmccart: this could be const

Issue #3375, opened 2018-12-11
Add circle() function


L05:
Add circle() function for K-12 education/accessibility, as suggested in [this thread](https://github.com/processing/p5.js/issues/2305).

#### Nature of issue?

- [X] New feature request

#### Most appropriate sub-area of p5.js?

- [X] Core/Environment/Rendering

vedhant:
Is this to be added to 2D Rendering?

lmccart:
this method could just be an alias that calls `p5.prototype.ellipse(x, y, 2*r, 2*r)`

vedhant:
Yup.
I would like to do it!

AnimeshSinha1309:
Okay, seems like no one is working on this, hope it's okay for me to do this.

nimishagarwal76:
#3419  resolves this issue

lmccart:
addressed with https://github.com/processing/p5.js/pull/3419

PR #3419, opened 2018-12-23
Add circle to core

nimishagarwal76:


nimishagarwal76:
This pull request resolves #3375  .

src/core/shape/2d_primitives.js
```diff
@@ -205,6 +205,38 @@ p5.prototype.ellipse = function(x, y, w, h, detailX) {
...
+ * </div>
+ *
+ * @alt
+ * white circle with black outline in mid of canvas that is 55x55.
+ */
+p5.prototype.circle = function() {
+  var args = Array.prototype.slice.call(arguments, 0, 2);
+  args.push(2 * arguments[2]);
+  args.push(2 * arguments[2]);
+  args = args.concat(Array.prototype.slice.call(arguments, 4));
```
lmccart: in the case of `square()` we had this line in here to pass through the corner radius arguments. in this case, there should be no more than 3 arguments, so this line is not needed

Issue #3425, opened 2018-12-26
Move away from browserify?


outofambit:
While working on #3385, (and from discussions in #3409) I can't help but wonder if this project would be better served by a babel + webpack build process over the current browserify + uglify infrastructure.

I _think_ we can replicate the existing behavior with [babel](https://babeljs.io/), using [babel-preset-env](https://babeljs.io/docs/en/babel-preset-env) and [babel-minify](https://github.com/babel/minify), and [webpack](https://webpack.js.org/) to do the bundling.

We might also be able to use something as simple as [rollup](https://github.com/rollup/rollup), though I am less familiar with that tool.

What do y'all think on a migration like this? Are there other tools worth considering?

limzykenneth:
What would be the benefit of moving from browserify to webpack or rollup? What would be the drawbacks, if any?

bigmstone:
**Disclaimer:** I am a `webpack`/`babel` user which artificially inflates my desire to see it in other places. I also help maintain a `gulp`/`browserify` app as part of my day job, and honestly don't have much bad to say about it. It's mainly set it and forget it. We have all the features I would want from a normal `webpack` setup. Including hot reload. Also use `babelify`, so we have access latest language features.

One weak(ish) reason to switch to `webpack` would be mind share and development resources. `webpack` has orders of magnitude more usage based on GH stars (meh) and NPM downloads (less meh). Having more usage will lead towards the usual suspects: trend towards having new features, able to respond quickly to competing features, quick bug discovery & patching, etc.

I just low-key want to help P5 get better ES6/"Instance" support since ES6 actually made me "okay" with JS. So mainly +1 to bringing `babel` into the mix. Either through `webpack` + `babel` or existing build system + `babelify`.

Spongman:
with webpack, how easy is it to run custom build steps? for example, something like the grunt task defined in `eslint-samples.js` ?

BTW: there's an example of a babelified, ES6 version of p5 over here: #3394. still uses grunt, but hey.

limzykenneth:
If there isn't a strong benefit to switch build tool then the effort may not be worth the reward so that's why I'm asking. 

Out of curiosity, because I haven't have a project that uses webpack or rollup but I want to try, how do they integrate with test runner like mocha or linters like eslint? Do they run with plugins on top or are they completely seperate, ie run with two shell commands (or one npm script with something like `&&`)?

limzykenneth:
[...]

PR #3429, opened 2018-12-27
Rollup build proof of concept

outofambit:
Did a quick proof of concept on replacing browserify with rollup to aid discussion in #3425. This is here mostly to have a convenient place for others to see the diff. :)

I only touched `browserify.js` (and not `combineModules.js`), so I couldn't remove any npm dependencies. However I was able to remove the need for `derequire` and `uglify` in the build, so they could be removed (from `package.json` and `Gruntfile.js`) if `combineModules.js` is also ported to use rollup.

If you want to try it out locally, checkout this branch and then:

```shellsession
$ npm i
$ npm run grunt rollup
$ npm run grunt rollup:min
```

😄

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -471,7 +471,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-contrib-clean');
 
   // Create the multitasks.
-  grunt.registerTask('build', ['browserify', 'browserify:min', 'uglify']);
+  grunt.registerTask('build', ['rollup']);
   grunt.registerTask('lint-no-fix', [
     'yui', // required for eslint-samples
     'eslint:build',
--- a/package-lock.json
+++ b/package-lock.json
@@ -6715,6 +6715,12 @@
         "is-path-inside": "^1.0.0"
       }
     },
+    "is-module": {
+      "version": "1.0.0",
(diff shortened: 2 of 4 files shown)


limzykenneth:
@outofambit Have you tried a Parcel setup? From what I see here it seems that it's including several additional plugins to be able to build what we want. I had a quick try with Parcel simply running `parcel build ./src/app.js` and seems to get a full minified version of the library, it's working as far as my manual test go but I haven't been able to swap it out so that it runs our test suite.

I think Parcel is pretty promising in this sense as I did not have to do any additional thing at all, no plugins no configs and it already minifies and includes json. Not sure if anything else is needed though.

outofambit:
@limzykenneth I haven’t tried Parcel yet, but that sounds promising! I was also disappointed in the number of plugins needed for pretty basic functionality in rollup. (From reading issues in rollup’s repo, there’s a lot of discussion around this.)

When I have some more time I’ll put up another test branch with parcel 🙂

outofambit:
I also found https://github.com/developit/microbundle, a wrapper for rollup that looks much simpler. Might be worth also trying out.

outofambit:
closing this in favor of continuing exploration in #3431 🚀

PR #3434, opened 2019-01-01
Added missing step to custom build documentation, clarified additional details.

coreygo:
Additional details here: https://github.com/processing/p5.js/issues/3425#issuecomment-450760491

### DIFF
--- a/developer_docs/custom_p5_build.md
+++ b/developer_docs/custom_p5_build.md
@@ -14,13 +14,14 @@ Currently, the usage is through invoking a Grunt task manually from the command 
 git clone https://github.com/processing/p5.js.git
 cd p5.js
 npm install
-grunt combineModules:module_x:module_y
+npm run grunt
+npm run grunt combineModules:module_x:module_y
 ```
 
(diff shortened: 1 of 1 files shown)


lmccart:
thank you!

developer_docs/custom_p5_build.md
```diff
@@ -14,13 +14,14 @@ Currently, the usage is through invoking a Grunt task manually from the command
 git clone https://github.com/processing/p5.js.git
 cd p5.js
 npm install
-grunt combineModules:module_x:module_y
+npm run grunt
+npm run grunt combineModules:module_x:module_y
 ```
 
-Here, `module_n` refers to the name of the module which you want to select. Multiple modules must be passed as shown above. Also, these modules must have the same name is their folders in `/src` directory to work correctly.
+Here, `module_n` refers to the name of the module which you want to select. Multiple modules must be passed as shown above. Also, these modules must have the same name is their folders in `/src` directory to work correctly. While `core` is included by default, `core/shape` needs to be included for shapes like line() and other core features to work.
```
coreygo: Erroops, just noticed I missed updating "…name **is** their folders…", it should be "…name **as** their folders…" 🔤 ✅

lmccart:
@coreygo thanks I've made the change

coreygo:
Cheers, have a good one! 😄

PR #3446, opened 2019-01-12
added example to brightness() and changed red() example

hellonearthis:
My first pull request to P5.

Added an example to brightness() to show how using color(hsb) corresponds to the brightness() returned value.

red() I added comments and output values to the example.
I still don't get what the 2nd example is doing as a useful example of red(),  but it's now documented better.

### DIFF
--- a/src/color/creating_reading.js
+++ b/src/color/creating_reading.js
@@ -108,9 +108,22 @@ p5.prototype.blue = function(c) {
  * rect(50, 20, 35, 60);
  * </code>
  * </div>
+ * <div>
+ * <code>
+ * noStroke();
+ * colorMode(HSB, 255);
+ * let c = color('hsb(60, 100%, 50%)');
+ * fill(c);
+ * rect(15, 20, 35, 60);
+ * let value = brightness(c); // A 'value' of %50 is 127.5
+ * fill(value);
+ * rect(50, 20, 35, 60);
+ * </code>
+ * </div>
  *
  * @alt
  * Left half of canvas salmon pink and the right half white.
+ * Left half of canvas yellow at half brightness and the right gray .
  *
  */
 p5.prototype.brightness = function(c) {
@@ -532,13 +545,12 @@ p5.prototype.lightness = function(c) {
  * </code>
  * </div>
  *
- * <div>
+ * <div class="norender">
  * <code>
- * colorMode(RGB, 255);
+ * colorMode(RGB, 255); // Set R in RGB range to 255
  * let c = color(127, 255, 0);
- * colorMode(RGB, 1);
+ * colorMode(RGB, 1); // Set R in RGB range to 1
  * let myColor = red(c);
- * print(myColor);
  * </code>
  * </div>
  *
--- a/src/color/creating_reading.js
+++ b/src/color/creating_reading.js
@@ -115,7 +115,7 @@ p5.prototype.blue = function(c) {
  * let c = color('hsb(60, 100%, 50%)');
  * fill(c);
  * rect(15, 20, 35, 60);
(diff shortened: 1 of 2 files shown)


src/color/creating_reading.js
```diff
@@ -108,9 +108,22 @@ p5.prototype.blue = function(c) {
...
  * </code>
  * </div>
+ * <div>
+ * <code>
+ * noStroke();
+ * colorMode(HSB, 255);
+ * let c = color('hsb(60, 100%, 50%)');
+ * fill(c);
+ * rect(15, 20, 35, 60);
+ * let value = brightness(c); // A 'value' of %50 is 127.5
```
limzykenneth: For English the percentage symbol should go after the number (50% instead of %50)

src/color/creating_reading.js
```diff
@@ -532,13 +545,13 @@ p5.prototype.lightness = function(c) {
  * </code>
  * </div>
  *
- * <div>
+ * <div class="norender">
  * <code>
- * colorMode(RGB, 255);
+ * colorMode(RGB, 255); // Set R in RGB range to 255
```
limzykenneth: Technically not setting only R in RGB to range but all values in RGB, only when providing 4 parameters do they indicate individual colour's range

hellonearthis: I'm not sure because I didn't write this example, I added comments to try and make sense of it but really I think it should be removed as it's so confusing.

src/color/creating_reading.js
hellonearthis: Yes, sorry I missed that.  
Should I change that and put in a new pull request,  I'm new to working with git repos.

lmccart: @hellonearthis you can make the change directly and push your commit to your own repo, and it will update here

src/color/creating_reading.js
outofambit: @hellonearthis how about this? I think this example is worth keeping since setting R, G, and B to the same range is a pretty common use case.
```suggestion
 * colorMode(RGB, 255); // Sets the range for red, green, and blue to 255
```

src/color/creating_reading.js
```diff
@@ -532,13 +545,13 @@ p5.prototype.lightness = function(c) {
...
  * let c = color(127, 255, 0);
- * colorMode(RGB, 1);
+ * colorMode(RGB, 1); // Set R in RGB range to 1
```
outofambit: I think this also needs to communicate that its all three (RGB), not just R. Does that make sense?
```suggestion
 * colorMode(RGB, 1); // Sets the range for red, green, and blue range to 1
```

src/events/keyboard.js
```diff
@@ -325,27 +333,16 @@ p5.prototype._onblur = function(e) {
...
- *   }
- *
- *   if (keyIsDown(UP_ARROW)) {
- *     y -= 5;
- *   }
- *
- *   if (keyIsDown(DOWN_ARROW)) {
- *     y += 5;
- *   }
+ *   if (keyIsDown(LEFT_ARROW)) x -= 5;
```
outofambit: I'm worried about removing the `{` and `}` from these if statements since it may confuse beginners who aren't familiar with JS syntax. ("Are the braces required or not?") Does anyone else have thoughts on this? I don't have a strong opinion.

limzykenneth: I think it is good to keep the braces for clarity.

lmccart:
@hellonearthis this pull request is really close. :) are you still planning to make the last few edits here?

lmccart:
thanks @hellonearthis I've made the remaining changes and merged!

Issue #3447, opened 2019-01-12
setting canvas style position property doesn't work


nebbles:
#### Nature of issue?

- [x] Found a bug

#### Most appropriate sub-area of p5.js?

- [x] Core/Environment/Rendering

#### Which platform were you using when you encountered this?

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: current CDN version
- Web browser and version: 71.0.3578.98 (Official Build) (64-bit)
- Operating System: macOS 10.14.2
- Steps to reproduce this:

Setting the following CSS is the only way to have the canvas bound to the top left corner of the current view.
```css
canvas {
    position: fixed;
    top: 0;
    left: 0;
    z-index: -1;
}
```
the corresponding `setup()` function does not work by itself.
```js
function setup() {
    canvas = createCanvas(windowWidth, windowHeight);
    canvas.style('position', 'fixed');
    canvas.style('top', '0');
    canvas.style('left', '0');
    canvas.style('z-index', '-1');
}
```

So only by setting the styling in CSS does the created canvas actually honour it. When set using `canvas.style()` the canvas does correctly reposition to the top of the page, however, it does not follow the current scroll position as it does with the CSS method.

Ajayneethikannan:
I tried to read the **p5.dom.js** library file and got the following results('sorry if it is a long message'):

The following is the style function from p5.dom.js:

```
p5.Element.prototype.style = function(prop, val) {
    var self = this;

    if (val instanceof p5.Color) {
      val =
        'rgba(' +
        val.levels[0] +
        ',' +
        val.levels[1] +
        ',' +
        val.levels[2] +
        ',' +
        val.levels[3] / 255 +
        ')';
    }

    if (typeof val === 'undefined') {
      if (prop.indexOf(':') === -1) {
        var styles = window.getComputedStyle(self.elt);
        var style = styles.getPropertyValue(prop);
        return style;
      } else {
        var attrs = prop.split(';');
        for (var i = 0; i < attrs.length; i++) {
          var parts = attrs[i].split(':');
          if (parts[0] && parts[1]) {
            this.elt.style[parts[0].trim()] = parts[1].trim();
          }
        }
      }
    } else {
      if (prop === 'rotate' || prop === 'translate' || prop === 'position') {
        var trans = Array.prototype.shift.apply(arguments);
        var f = this[trans] || this['_' + trans];
        f.apply(this, arguments);
      } else {
        this.elt.style[prop] = val;
        if (
[...]

PR #3448, opened 2019-01-12
Fix for p5.Element.prototype.position not applying styling properly

nebbles:
Proposed fix for #3447 where `p5.Element.prototype.position` of p5.dom does not style the element properly as called by e.g. `style('position','fixed')`.

As seen in issue, the problem stems from `p5.Element.prototype.style` however, it seems that the design of this function is to delegate the assignment of transformations to their respective functions.

As such `p5.Element.prototype.position` was modified to be able to handle all cases, explicitly ensuring only valid values for CSS `position` are used.

This is my first PR so hopefully checked all the boxes 🤞

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1621,6 +1621,17 @@
   p5.Element.prototype.position = function() {
     if (arguments.length === 0) {
       return { x: this.elt.offsetLeft, y: this.elt.offsetTop };
+    } else if (
+      arguments.length === 1 &&
+      (arguments[0] === 'static' ||
+        arguments[0] === 'absolute' ||
+        arguments[0] === 'fixed' ||
+        arguments[0] === 'relative' ||
+        arguments[0] === 'sticky' ||
+        arguments[0] === 'initial' ||
+        arguments[0] === 'inherit')
+    ) {
+      this.elt.style.position = arguments[0];
     } else {
       this.elt.style.position = 'absolute';
       this.elt.style.left = arguments[0] + 'px';


limzykenneth:
I think in the case of `style('position','fixed')` it is actually wrong to call `p5.Element.prototype.position` as they are different thing all together. `style` deals with css directly and in css `position` is the way the element is positioned while in p5 or rather `p5.Element.prototype.position` is the coordinates of the element. It shoud either pass through and set the css directly or a new helper function is written (which I don't think is necessary)

Also `p5.Element.prototype.position` is a public method so adding another overload like this could potentially cause problem unless documented.

nebbles:
As referred to in #3447, it does seem to be strange behaviour for `style`:

```js
p5.Element.prototype.style = function(prop, val) {
    var self = this;

    if (val instanceof p5.Color) {
      val =
        'rgba(' +
        val.levels[0] +
        ',' +
        val.levels[1] +
        ',' +
        val.levels[2] +
        ',' +
        val.levels[3] / 255 +
        ')';
    }

    if (typeof val === 'undefined') {
      if (prop.indexOf(':') === -1) {
        var styles = window.getComputedStyle(self.elt);
        var style = styles.getPropertyValue(prop);
        return style;
      } else {
        var attrs = prop.split(';');
        for (var i = 0; i < attrs.length; i++) {
          var parts = attrs[i].split(':');
          if (parts[0] && parts[1]) {
            this.elt.style[parts[0].trim()] = parts[1].trim();
          }
        }
      }
    } else {
      if (prop === 'rotate' || prop === 'translate' || prop === 'position') {
        var trans = Array.prototype.shift.apply(arguments);
        var f = this[trans] || this['_' + trans];
        f.apply(this, arguments);
      } else {
        this.elt.style[prop] = val;
        if (
          prop === 'width' ||
          prop === 'height' ||
          prop === 'left' ||
          prop === 'top'
        ) {
          var numVal = val.replace(/\D+/g, '');
          this[prop] = parseInt(numVal, 10); // pend: is this necessary?
        }
      }
    }
    return this;
  };
```
and so its really the application of position style here that needs the fix in that case.

Happy to make the alteration here instead? Would certainly cause less confusion if `position` is a public method exclusively for setting absolute coordinates.

Ajayneethikannan:
@nebbles  @limzykenneth , I agree that the position function should not be overloaded , but in my humble opinion, it would be nice to see if it supported other position attributes instead of only being restricted to "absolute"

The use case of other properties may arise when the canvas may be present inside another div,
this issue can be considered as an example , 
https://github.com/processing/p5.js-website/issues/394
where the parent div is "relative" , preventing us to set position with respect to window.
My suggestion is an optional  third parameter , so that we can call the function as: 

```some_elt.position(10, 10, ABSOLUTE);```
or 

```some_elt.position.(10, 10, FIXED);```

with default value being absolute.  

Waiting for your thoughts about it ,
Thank you!

limzykenneth:
@Ajayneethikannan I think the proposal of something like `some_elt.position(10, 10, ABSOLUTE);` can be opened as a separate feature request issue for discussions.

lmccart:
thanks @nebbles! we've now addressed #3447 by removing the support for special arg "position" from style, so it can be handled directly by `position()` function. we can continue discussion of the arguments for that function over in #3570.

Issue #3367, opened 2018-12-11
update developer docs to ES6


lmccart:
As part of the migration to ES6, the [developer_docs](https://github.com/processing/p5.js/tree/master/developer_docs) need to have code snippets and language updated from ES5 to ES6. 

All docs currently in the folder have been listed here, but not all of them have code snippets that need updating. These can be checked off once it's confirmed no update is needed.

* [x] [benchmarking_p5.md](https://github.com/processing/p5.js/blob/master/developer_docs/benchmarking_p5.md)
* [x] [creating_libraries.md](https://github.com/processing/p5.js/blob/master/developer_docs/creating_libraries.md)
* [x] [custom_p5_build.md](https://github.com/processing/p5.js/blob/master/developer_docs/custom_p5_build.md)
* [x] [design_principles.md](https://github.com/processing/p5.js/blob/master/developer_docs/design_principles.md)
* [x] [friendly_error_system.md](https://github.com/processing/p5.js/blob/master/developer_docs/friendly_error_system.md)
* [x] [inline_documentation.md](https://github.com/processing/p5.js/blob/master/developer_docs/inline_documentation.md)
* [x] [issue_labels.md](https://github.com/processing/p5.js/blob/master/developer_docs/issue_labels.md)
* [x] [preparing_a_pull_request.md](https://github.com/processing/p5.js/blob/master/developer_docs/preparing_a_pull_request.md)
* [x] [release_process.md](https://github.com/processing/p5.js/blob/master/developer_docs/release_process.md)
* [x] [supported_browsers.md](https://github.com/processing/p5.js/blob/master/developer_docs/supported_browsers.md)
* [x] [webgl_mode_architecture.md](https://github.com/processing/p5.js/blob/master/developer_docs/webgl_mode_architecture.md)

dinesh196:
@Ajayneethikannan  since there are a number of pages, I would also like to help in updating the snippets and thus, this issue.

lmccart:
closed with #3454

PR #3454, opened 2019-01-14
Developer docs updation

Ajayneethikannan:
This pull request attempts to solve the issue #3367 
The changes done are:
conversion of var to let and const,
conversion of normal functions to arrow functions wherever required.

The files which **do not** need updation are:

custom_p5_build.md
design_principles.md
issue_labels.md
preparing_a_pull_request.md
release_process.md
supported_browsers.md
webgl_mode_architecture.md

Some of my observations were:

The last code snippet in the friendly_error_system.md (which contains the attempt to access the global errors in the browser) produces infinite recursive calls and **stack overflow**, 
so maybe that issue can be corrected in a separate pull request . Please correct me if I am wrong

### DIFF
--- a/developer_docs/friendly_error_system.md
+++ b/developer_docs/friendly_error_system.md
@@ -20,7 +20,7 @@ FES provides a generalized error message generation system that developers may u
 * file loading error example:
 ````javascript
 /// missing font file
-var myFont;
+let myFont;
 function preload() {
   myFont = sketch.loadFont('assets/OpenSans-Regular.ttf');
 };
@@ -97,7 +97,7 @@ Note that this will disable the parts of the FES that cause performance slowdown
 * The friendly error system slows the program down, so there is an option to turn it off via setting `p5.disableFriendlyErrors = true;`. In addition, the friendly error system is omitted by default in the minified (`p5.min.js`) version.
 * FES may still result in false negative cases. These are usually caused by the mismatch between designs of the  functions (e.g. drawing functions those are originally designed to be used interchangeably in both 2D and 3D settings) with actual usage cases. For example, drawing a 3D line with 
 ```javascript 
-var x3; // undefined
+const x3; // undefined
 line(0, 0, 100, 100, x3, Math.PI);
 ```
  will escape FES, because there is an acceptable parameter pattern (`Number`, `Number`, `Number`, `Number`) in `line()`'s inline documentation for drawing in 2D setting. This also means the current version of FES doesn't check for the environmental variables such as `_renderer.isP3D`.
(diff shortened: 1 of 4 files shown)


limzykenneth:
It indeed does cause a stack overflow as objects are not copied when assigned to variables.

Ajayneethikannan:
@limzykenneth should I remove that code snippet? I did not update it as it was given as an example of the attempted methods to catch the errors globally.

limzykenneth:
It will be best if you can fix it instead of removing it, so probably saving individual references to `console.log` `console.warn` and `console.error` instead of just the `console` object.

Ajayneethikannan:
@limzykenneth I have fixed the snippet as per your advice, thank you!
I have added few comments to explain what is happening in that piece of code.

Also , we can remove the red colored  error box from the console by returning true in the 
**window.onerror** method. Maybe we can wrap it up in some function  as in this snippet or use callbacks to handle the error after that. If someone could guide me, I would love to take it up as a new issue !

developer_docs/friendly_error_system.md
```diff
@@ -116,10 +116,18 @@ line(0, 0, 100, 100, x3, Math.PI);
 * Global Error catching. It would be very helpful to catch the errors the browser is throwing to the console, so we can match them up with friendly comments. So far we've tried `window.onerror` and the following with no success.
```
lmccart: Is this snippet below something that works or does not work? In the previous version of this text, this paragraph was explaining something that had been tried but does not work.

lmccart: Please add 2 space indentations for this code block below.

Ajayneethikannan: @lmccart  I have done the indentation modification as directed. 

In the previous version of the code , 
```
var original = window.console;
      
["log", "warn", "error"].forEach(function(func){
window.console[func] = function(msg) {
    original[func].apply(original, arguments)
};
``` 

The newly defined function was referencing itself, causing the stack overflow.

This was corrected with help from @limzykenneth , 
making sure that the new functions do not reference themselves , by storing the references to the original function separately. 
```
const original = window.console;
const original_functions  = {
  log: original.log,
  warn:  original.warn,
  error: original.error
}

["log", "warn", "error"].forEach(function(func){
window.console[func] = function(msg) {
//do something with the msg caught by the wrapper function, then call the original function
original_functions[func].apply(original, arguments)
};
});
```

So the updated snippet does not cause the overflow and also catches the errors.

Also , I was not fully able to understand the use of catching global errors ,
but if we want  to replace the red error messages with better looking custom messages, 
we can replace the window.onerror method and finally return true , which may cause the code to run as if it did not detect the error, but also remove the 'unfriendly' error message :smile: 
Maybe to handle errors , we can wrap the window.onerror method with some other function, which returns true.

Thank you!

lmccart: Got it, thank you @Ajayneethikannan. Then can we remove this line above: 
"So far weve tried `window.onerror` and the following with no success."

And update this comment:
```js
//this snippet attempts to wrap window.console methods with a new function to modify their functionality"
```
to

```js
// this snippet wraps window.console methods with a new function to modify their functionality
// it is not currently implemented, but could be to give nicer formatted error messages
```

Then we should be good to merge. Thanks!

lmccart:
🎉

developer_docs/friendly_error_system.md
Ajayneethikannan: I have done the mentioned changes ,  please check it ,
Thank you @lmccart  !

Issue #3450, opened 2019-01-13
Examples involving keyboard events, only one works


hellonearthis:
#### Nature of issue?

- [x] Found a bug

#### Most appropriate sub-area of p5.js?

- [x] Events

#### Which platform were you using when you encountered this?

- [x] Desktop/Laptop

#### Details about the bug: 

Was adding to the event keyboard examples and noticed  that only the 1st example actually functions and the following examples don't function.   see https://p5js.org/reference/#/p5/keyPressed

- Web browser and version: Chrome 71.0.3578.98 (Official Build) (64-bit)
- Operating System: Windows 10
- Steps to reproduce this:
Add a 2nd code example to keyboard.js see https://p5js.org/reference/#/p5/keyPressed

hkirat:
Hi @hellonearthis 
I would like to work on this issue.

sc231997:
@hellonearthis I would like to work on this.

sc231997:
@hellonearthis 
I want to try to fix this. 
Where can I find the code where keyPressed() is implemented?

hellonearthis:
I have no idea where the code is,  i guess the issue is around having to things tapping into the keyboard event.  But on the demos, i can stop both demos and then try to start the 2nd demo with no luck.

Having only one example is the current fix but it's no solution.

limzykenneth:
`keyPressed()` is defined in [here](https://github.com/processing/p5.js/blob/master/src/events/keyboard.js#L176). I would suggest determining whether the problem is limited to the reference page or is it a general bug in the library first before attempting to fix anything.

sc231997:
> `keyPressed()` is defined in [here](https://github.com/processing/p5.js/blob/master/src/events/keyboard.js#L176). I would suggest determining whether the problem is limited to the reference page or is it a general bug in the library first before attempting to fix anything.

In `keyIsDown()` reference page also we have two examples and there the function is working as expected. So maybe there is an issue with the library. 

@limzykenneth can you have the look and tell what may be the issue. 

I am new here so not much idea with code base still learning. :)

Spongman:
It might be helpful to find which git commit (if any) introduced the bug. `git bisect` Can be used for this.

sanketsingh24:
@Ajayneethikannan is this same as #3193, I haven't been following the discussion for some time.

Ajayneethikannan:
[...]

PR #3456, opened 2019-01-15
Removed erroneous example

hkirat:
In Reference to #3450. 

On going through the code, it seems like the keyPressed function might be overriden by the first example. Not sure why this might be happening, but I'm taking a look alongside to try to fix it. 

Meanwhile, I'm pushing this change to remove unexpected behavior from the website, including just one example.


lmccart:
@hkirat I would prefer to fix the behavior where only one example is responding rather than remove an example completely.

PR #3457, opened 2019-01-15
Distance Calculation made easier.

ThunbergOlle:
This is pretty much my first pull request ever. Anyway:

I think there is an easier way of calculating the distance between two points. I think it should be easier to read and understand and also faster by doing this calculation instead. I may be wrong.

It cleans the code and cuts down the code length of the function.

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -803,19 +803,18 @@ function hypot(x, y, z) {
     return Math.hypot.apply(null, arguments);
   }
   if (!z) {
-    var a2 = Math.pow(x, 2);
-    var b2 = Math.pow(y, 2);
-    var c2 = a2 + b2;
-    var hypoten = Math.sqrt(c2);
+    var a2 = x * x;
+    var b2 = y * y;
+    var hypoten = Math.sqrt(a2 + b2);
     // Otherwise use the V8 implementation
     // https://github.com/v8/v8/blob/8cd3cf297287e581a49e487067f5cbd991b27123/src/js/math.js#L217
     return hypoten;
   } else if (z) {
     // Calculate bottom triangle
-    // The calculation of the distance between x,y,z is done by using 2 calculations of pythagoras. 
+    // The calculation of the distance between x,y,z is done by using 2 calculations of pythagoras
     // https://mr-mathematics.com/pythagoras-theorem-in-3d-shapes/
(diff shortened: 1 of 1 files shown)


outofambit:
@Spongman can you please say more? why do you say that?

outofambit:
Also, thank you @ThunbergOlle for submitting this pull request! and congrats on your first one ever! 🎉

Spongman:
because the `Math.pow()` can be significantly more expensive that just multiplying a few numbers and/or taking the reciprocal. some implementations may shortcut around this, but not all do.

Spongman:
also, i'm not quite sure about the math of the 3d case. i'm pretty sure you can remove one of the `sqrt` calls, especially since they're so expensive.

i'm wondering what the rationale is behind changing the existing implementation? it handles overflows correctly, infinities, string parameters, and arbitrary-length argument lists.

ThunbergOlle:
Alright, I'm going to change the `Math.pow` parts to just multiply with itself. The math of the 3d case is indeed unnecessary. I will be changing that later.

The reason for chaning the existing implementation was mainly because of curiosity but I also thought I may had come up with something that may be better, so why not share it.

ThunbergOlle:
I've fixed everything mentioned by @Spongman in the latest commit. 👍

sanketsingh24:
Hey! I can clearly see that your implementation is alot easier to understand for others, but it is fundamental. The earlier implementation is straight from ` v8 `(also pointed out by @Spongman), and takes a lot more precautions and is thought over to be the most optimized one, and is agreed and trusted by the whole community. So I guess its better to keep the old implementation.

Also, Congrats on your first PR.

outofambit:
@sanketsingh24 @Spongman 

Why don't we write up a list of test cases for this function? I understand the existing covers some important edge cases, but I do think this PR increases the readability considerably. It's hard to judge the tradeoffs (or understand what we could reincorporate into this PR).

So far, I'm hearing these cases are covered in the existing implementation:
* overflows/rounding correctly
* infinities
* string parameters
* arbitrary-length argument lists 

Are there others? I think this is a great opportunity to add tests for these cases.

lmccart:
closing this due to inactivity. please feel free to open an issue following up on @outofambit's comment if you'd like to continue work on this.

PR #3468, opened 2019-01-21
Remove stderr check in docs release task

### LINKED ISSUE #3465: Release docs task git behaviour
Continuing from #3464, I think I've managed to track where the problem came from but my git knowledge (maybe even shell knowledge) isn't good enough to understand the behaviour. The problem lies in this block of code in [release-docs.js](https://github.com/processing/p5.js/blob/master/tasks/release/release-docs.js):

```javascript
.then(function() {
          // Add, Commit, Push
          console.log('Pushing to GitHub ...');
          return new Promise(function(resolve, reject) {
            exec(
              'git add --all && git commit -am "Updated Reference for version ' +
                version +
                '" && git push',
              { cwd: './p5-website' },
              function(err, stdout, stderr) {
                if (err) {
                  reject(err);
                }
                if (stderr) {
                  reject(stderr);
                }
                resolve();
              }
            );
          });
        })
        .then(function() {
          console.log('Released Docs on Website!');
          done();
        })
        .catch(function(err) {
          throw new Error(err);
});
```
Most notably this:
```javascript
if (stderr) {
  reject(stderr);
}
```

Apparently git is printing identical message to both `stdout` and `stderr`:
```
remote: 
remote: Create a pull request for 'docs-test' on GitHub by visiting:        
remote:      https://github.com/limzykenneth/p5.js-website/pull/new/docs-test        
remote: 
To https://github.com/limzykenneth/p5.js-website.git
 * [new branch]        master -> docs-test
```
and since `stderr` is truthy it triggers the promise rejection.

I did a quick search online and it seems that printing to `stderr` isn't necessarily an indication of an error occurring (yes? no?) but then in this case why did the docs release task not throw this kind of error before? (or did it?) Did git change its behaviour recently??

I'm quite baffled really, if I am to suggest a fix it would be to just remove the `stderr` `if` block above.

Pinging @lmccart and anyone else with better understanding of git and shell that I do

limzykenneth:
Closes #3465

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "p5",
-  "version": "0.7.2",
+  "version": "0.7.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


lmccart:
thanks @limzykenneth!

PR #3472, opened 2019-01-22
Fixed renderer getting out of sync

### LINKED ISSUE #3265: webgl: setAttributes() breaks resizeCanvas aspect
new in 0.7.2 (doesn't repro in 0.7.1)

- https://codepen.io/Spongman/pen/KGmNrB?editors=1010
- resize the window
- notice the circle is stretched

calling `setAttributes()` causes `resizeCanvas()` to *not* set the aspect ratio of the canvas correctly.

```javascript
function setup() {
  setAttributes('foobar', true);
}

function windowResized() {
}

function draw() {
  background(0);
  ellipse(0, 0, width/4);
}
```

sanketsingh24:
Closes #3265 

- This PR fixes the bug in which the aspect ratio was not being maintained when resizing. The error was the child `_renderer` in `_curCamera` was getting out of sync after `pop()` call in `setattributes()`.   
- @Spongman  suggested the solution for this bug. Also added the sketch I was using in manual tests. 
- Need suggestions for whether the sketch should be merged, and if yes, where it should be present.

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -365,6 +365,12 @@ p5.prototype.setAttributes = function(key, value) {
   this.push();
   this._renderer._resetContext(attr);
   this.pop();
+  if (
+    this._renderer._curCamera._renderer !== this._renderer &&
+    this._renderer._curCamera.cameraType === 'default'
+  ) {
+    this._renderer._curCamera._renderer = this._renderer;
+  }
 };
 
 /**
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -365,10 +365,7 @@ p5.prototype.setAttributes = function(key, value) {
   this.push();
   this._renderer._resetContext(attr);
   this.pop();
-  if (
-    this._renderer._curCamera._renderer !== this._renderer &&
-    this._renderer._curCamera.cameraType === 'default'
-  ) {
+  if (this._renderer._curCamera) {
     this._renderer._curCamera._renderer = this._renderer;
   }
 };
--- a/test/manual-test-examples/webgl/setattribute/index.html
+++ b/test/manual-test-examples/webgl/setattribute/index.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <title></title>
+  <link rel="stylesheet" href="../../styles.css">
+  <script language="javascript" type="text/javascript" src="../../../../lib/p5.js"></script>
+  <script language="javascript" type="text/javascript" src="sketch.js"></script>
+  <script src="../stats.js"></script>
+</head>
+
+<body>
+  <header>
+    <p>Click to enlarge the canvas</p>
+  </header>
+</body>
+
+</html>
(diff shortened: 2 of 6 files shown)


src/webgl/p5.RendererGL.js
```diff
@@ -365,6 +365,12 @@ p5.prototype.setAttributes = function(key, value) {
   this.push();
   this._renderer._resetContext(attr);
   this.pop();
+  if (
+    this._renderer._curCamera._renderer !== this._renderer &&
+    this._renderer._curCamera.cameraType === 'default'
```
Spongman: i'm not sure the `cameraType` check is necessary here.

i think this check can be as simple as
```javascript
if (this._renderer._curCamera) {
  this._renderer._curCamera._renderer = this._renderer;
}
```

sanketsingh24: The idea behind `cameraType` check was taking precautions rather than waiting for a bug to appear but at this moment, I was not able to think of a situation where this could go wrong.

For the second suggestion, I think that could work for sure, didn't think of that while implementing. I think the current code is better in terms of readability, but anyways I will add a comment .

Spongman: i can understand the precaution, but it seems to me that the bug here is that the camera's `_renderer` property is wrong (once the renderer has been switched out), and that this is the case regardless of the `cameraType`. At this point in the code, the old renderer has been discarded - it's no longer attached to the canvas.

Honestly i'd rather that the camera didn't hold onto the renderer at all: the back reference is a design smell and this kind of state-related bug is why.

test/manual-test-examples/webgl/setattribute/sketch.js
```diff
@@ -0,0 +1,12 @@
+function setup() {
+  createCanvas(640, 480, WEBGL);
+  setAttributes('foobar', true);
```
stalgiag: Might as well make this a real attribute

test/manual-test-examples/webgl/setattribute/index.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
```
stalgiag: perhaps this can go in the `/manual-test-examples/webgl/camera/setAttributes`?

src/webgl/p5.RendererGL.js
```diff
@@ -365,6 +365,9 @@ p5.prototype.setAttributes = function(key, value) {
   this.push();
   this._renderer._resetContext(attr);
   this.pop();
+  if (this._renderer._curCamera) {
+    this._renderer._curCamera._renderer = this._renderer;
+  }
```
stalgiag: Great! I would say that it might be a good idea to put a test into `test/unit/webgl/p5.Camera.js`. Something to the effect of:
```
test('Camera is Renderer is correctly set after setAttributes', function() {
      var myCam2 = myp5.createCamera();
      assert.deepEqual(myCam2, myp5._renderer._curCamera);
      myp5.setAttributes('antialias', true);
      assert.deepEqual(myCam2._renderer, myp5._renderer);
    });
```
will probably be good. Though some more general state consistency tests may be in order soon.

sanketsingh24: Will the suite `RendererGL Current Camera` a good place for this test?

stalgiag: Yes that sounds good!

stalgiag:
Thank you for the contribution!

Issue #740, opened 2015-06-10
Arcs being drawn asymmetrically


joecridge:
Arcs seem to be drawn asymmetrically if their start and end angles are not both multiples of QUARTER_PI. The arc is drawn correctly for the first half but deviates quite significantly from a circle between the middle and the end.

workergnome:
Paging @christinedt!

workergnome:
@joecridge—I know that we've been trying to get the arc drawing code right, and it's turning out to be harder than we thought.  @christinedt took a pass through it at the p5jscon, and fixed a problem with it, but apparently we've still got issues.  

Do you have any interest in looking into this yourself?  I think @christinedt is currently busy on other projects, so I don't think that she'd have an issue with you tackling it yourself.  If so, let me know, and I can dump what I know about the issue on you.  If not, no worries—just wanted to know that we've acknowledged this, and we're looking for someone to work on it.

joecridge:
@workergnome Okay, yes I'd be up for giving it a go. I've been using Processing for a long time but I've only recently started to look into its code and p5.js, so any pointers you could give me would be great.

workergnome:
Wonderful!  

The arc drawing code is in the `/src/core/2d_primitives.js` file.  There are also comments there referencing the sources for the math there—a couple papers.

I know that @christinedt found a revised paper fixing some of the existing math, and it seemed to help with some previous problems, and she listed the revision in a comment, but it might not work.

It might also be worth looking at the Java processing code, seeing if their implementation is somehow different and better.  I haven't done that. 

If it turns our that our current technique is totally untenable, we're open to suggestions and complete revisions, as long as there aren't API changes.

Thank you again for looking into this, and if you have any questions feel free to ping me in a comment here!

workergnome:
Also—Have you done the github pull request dance yet?  If you have questions about that, let me know as well.  I'm supposed to have written up instructions, but I haven't quite finished them.

joecridge:
Okay great, I'll see what I can do.

I haven't had a go at the dance yet, any tips?

workergnome:
### Pull Requests:
 
[...]

PR #3475, opened 2019-01-23
Fix: using parametric angles for drawing arc

### LINKED ISSUE #3466: oblong arc() draws angles inconsistently
#### Nature of issue?

- [ x] Found a bug
- [ ] New feature request

- [ ] Color
- [ x] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

- [ x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: v0.7.3
- Web browser and version: brave 0.58.21, chrome 71.0.3578.98, safari 12.0.2 
- Operating System: MacOS Mojave 10.14.2  
- Steps to reproduce this: https://editor.p5js.org/datramt2/sketches/3nZSO_HMn

#### Details

Please take a look at this example:
https://editor.p5js.org/datramt2/sketches/3nZSO_HMn

While increasing the slider from the minimum to maximum value, a green arc() will trace a vertical ellipse from 0 to TWO_PI. Next to it, a blue arc() will trace the same vertical ellipse from PI to 3 PI. I’ve rotated the blue arc() by -PI to cancel out the phase difference between the two arcs, so as to demonstrate the angle differential between them.

In theory—I think(?)—the green arc and the blue arc ought to trace ellipses at the same rate as the user manipulates the slider. As my example shows, there are some inconsistencies in the rate at which the two arcs form when using the slider.

Further investigation revealed that angles fed into arc() that are less than 3/2 PI are drawn according to the subtended angle (i.e., the angle at the center-point that would form were we to connect the edges of the arc to the center). Meanwhile, angles fed into arc() that are greater than 3/2 PI are drawn according to the parametric equation (i.e., x(t) = a cos(t), y(t) = b sin(t)). 

In my example, if we focus on everything that's green and ignore anything that is blue, I have drawn a small circle in the center with a small rotating hand that represents the subtended angle. I’ve also drawn a relatively larger hand that reaches the perimeter of the ellipse. This is the parametric representation. If we drag the slider, we see that the arc follows the subtended angle (little hand) until we’ve reached 3/2 PI, at which point, the arc follows the big hand (parametric angle). 

Unless I am confused about something, to me, this seems unintended. It is worth noting that in processing, arc() uses the parametric equation only. Here is a java snippet to demonstrate that:
```java
void setup() {
  size(600, 600);
  noFill();
}

void draw() {
  background(80);
  stroke(150, 255, 150);
  strokeWeight(1);
  pushMatrix();
  translate(width / 2, height / 2);
  //text((mouseX / PI).toFixed(3) + " PI", -22, 268)
  ellipse(0, 0, 80, 500);
  strokeWeight(10);
  arc(0, 0, 80, 500, 0, 0 + mouseX/float(90)); //0 to 2 PI
  popMatrix();
  
  strokeWeight(4);
  pushMatrix();
  translate(width/2, height/2);
  line(0, 0, cos(mouseX/float(90))*40, sin(mouseX/float(90))*250);
  popMatrix();
}
```

I believe that this p5.js arc() issue is the culprit behind this "closed" bug: #3273. They fixed it by ensuring that the arc was not oblong, but did not fix the underlying issue that I am highlighting here

Thanks, and cheers! <3

vedhant:
-fixes #3466 
-Earlier angle was calculated as subtended angle until 3*PI/2, and after that it was calculated as parametric angles.
-Now it draws arcs as parametric angles (Processing does the same)

### DIFF
--- a/src/core/shape/2d_primitives.js
+++ b/src/core/shape/2d_primitives.js
@@ -118,18 +118,6 @@ p5.prototype.arc = function(x, y, w, h, start, stop, mode, detail) {
     }
   }
 
-  //Adjust angles to counter linear scaling.
-  if (start <= constants.HALF_PI) {
-    start = Math.atan(w / h * Math.tan(start));
-  } else if (start > constants.HALF_PI && start <= 3 * constants.HALF_PI) {
-    start = Math.atan(w / h * Math.tan(start)) + constants.PI;
-  }
-  if (stop <= constants.HALF_PI) {
-    stop = Math.atan(w / h * Math.tan(stop));
-  } else if (stop > constants.HALF_PI && stop <= 3 * constants.HALF_PI) {
-    stop = Math.atan(w / h * Math.tan(stop)) + constants.PI;
-  }
-
   // Exceed the interval if necessary in order to preserve the size and
   // orientation of the arc.
   if (start > stop) {


AdilRabbani:
I think this was originally implemented by @joecridge and he might be the right guy to review this. 👍

vedhant:
@joecridge can u review this?

joecridge:
Thanks for tagging me in.

@vedhant you’re right that this is the location of the problem, but I’m not 100% sure that removing the scaling correction completely is the right solution. Processing’s interpretation of `start` and `stop` are inconsistent in a number of ways (sometimes the two get swapped around; you don’t get the angles you asked for when `start != stop`; etc), and this is something we decided to ‘fix’ for p5 ([original discussion](https://github.com/processing/p5.js/issues/740#issuecomment-113974932)).

IMO the misbehaviour here is because both statements used to have an `else` clause that dealt with the fourth quadrant (and the angles were constrained to `[0,TWO_PI)` beforehand), but that got removed somewhere along the way. I’m refactoring the code around this at the moment to fix some other `arc()` bugs, and I’d like to add it back in there with some proper tests.

So I think the best thing to do for the time being is to add back in the `else` clauses (I think `start = Math.atan(w / h * Math.tan(start)) + constants.TWO_PI`, and similarly for stop). That will fix the inconsistency for all but the cases where start and stop overlap, which is the part I’m dealing with in my PR.

vedhant:
It has been quite a while. @joecridge if you want I can continue working on this PR.

joecridge:
@vedhant Indeed it has! I have just incorporated these changes into #3584 (there‘s a flag for toggling off the angle correction if we decide we want that later) along with fixes for the other arc issues I’m aware of. Thank you for identifying this one and suggesting a patch ❤️. I am sorry for taking forever and generally handling it badly 😅.

stalgiag:
Thanks for the work @vedhant! I am closing this as per #3584.

Issue #3476, opened 2019-01-23
Element.drop requires p5.dom despite being a 'core' feature


mattt-b:
#### Nature of issue?

- [x] Found a bug

#### Most appropriate sub-area of p5.js?

- [x] Core/Environment/Rendering

#### Details about the bug: 

- p5.js version: Tested at least 0.6.1 to 0.7.3

[Element.drop()](https://p5js.org/reference/#/p5.Element/drop) is listed as a `core` feature, but it requires [p5.dom](https://p5js.org/reference/#/libraries/p5.dom) since it depends on [p5.File](https://p5js.org/reference/#/p5.File) which is only exposed in p5.dom.

lmccart:
I would suggest we move `drop()` to the p5.dom library. It feels like a more specialized function. The methods in p5.js are meant to be the most basic set needed to have basic DOM interactions. Since `createFileInput()` is in p5.dom, for example, it would make sense to have `drop()` there too

lmccart:
fixed with https://github.com/processing/p5.js/pull/3477

PR #3477, opened 2019-01-23
Move drop() into p5.dom

### LINKED ISSUE #3474: Problem importing p5.dom in webpack project
#### Nature of issue?

- [x] Found a bug

- [X] Other (specify if possible)
p5.dom

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.7.3
- Steps to reproduce this:

I am using p5js 0.7.3 with webpack. I do not get this error with 0.7..2. When using the global script tags in the Getting Started guides there is no issue.

When importing p5.dom I get an error: `Uncaught ReferenceError: p5 is not defined`, relating to this section: [p5.dom.js#L3225](https://github.com/processing/p5.js/blob/master/lib/addons/p5.dom.js#L3225).

I'm wondering if those two functions are supposed to be moved up inside of the block above where p5 is defined? Moving it manually erases my issues locally, but I don't know enough about p5js or js modules to know if that affects something else.

mattt-b:
Fixes #3474 and #3476 

Moves drop() to be a p5.dom instead of core feature.

Tested in both global and instance mode with the default `<script>` tag method of using p5js and using a build system with the npm package.


lmccart:
thanks @mattt-b!

Issue #3270, opened 2018-10-16
Using AUTO with p5.element.size() sets width or height to 0, rather than "auto"


catarak:
#### Nature of issue?

- [X] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [X] Other (specify if possible) DOM

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version:  0.7.2
- Web browser and version:  Chrome 69.0.3497.100
- Operating System:  MacOSX 10.13.4
- Steps to reproduce this:

The image is rendering with no height, with the height set to `AUTO` in the following code:
```JS
function setup() {
  noCanvas();
  const url = 'https://dog.ceo/api/breeds/image/random';
  loadJSON(url, processData);
}

function processData(data) {
  print(data.message);
  let img = createImg(data.message);
  img.size(500, AUTO);
}
```
If I look at the DOM, the `<img>` was appended with a` height` of `"0"`:
<img width="540" alt="screen shot 2018-10-16 at 3 20 46 pm" src="https://user-images.githubusercontent.com/6063380/47041564-23ee9b80-d157-11e8-8774-eddbd767b52b.png">
If I manually change the `height` to `"auto"` by editing the DOM, it works as expected.

I also tried changing `AUTO` to the String `"auto"`, and I see the same behavior.

Ajayneethikannan:
I would like to work on this issue !

Ajayneethikannan:
I tried to fix the issue and this is what I came up with:
Whenever we use the createImg function to create the image with the given data, the corresponding height and width of the image are initially set to 0, before the image has finished loading by the browser. 
So when we attempt to calculate the automatic height or width, the calculation results in NaN.

Ajayneethikannan:
As I was going through the docs, I noticed that createImg function had a callback parameter,called after the image has loaded , and I believe that the correct way to call the  methods  on the image object would be to call it inside the callback function.
so changing this code : 

``let img = createImg(data.message);
 img.size(500, AUTO);``

to:

[...]

PR #3481, opened 2019-01-25
fixes automatic image size in size function in p5.dom.js

Ajayneethikannan:
This PR tries to solve the issue #3270

The following are the changes made in this PR:

Before assigning the height and width to the image element, the function checks whether the values are numbers or not using isNaN() function . <br/>
Also,

This situation (where sizes are initially zero) occurs only in image elements,
and can be handled separately by using the **image.complete** property (checks whether browser has finished loading the image) , I thought handling the issue using isNaN() would be much shorter.

@lmccart  can you please review the changes?

Thank you!

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -1965,15 +1965,20 @@
   /**
    *
    * Sets the width and height of the element. AUTO can be used to
-   * only adjust one dimension. If no arguments given returns the width and height
-   * of the element in an object.
+   * only adjust one dimension at a time. If no arguments are given, it
+   * returns the width and height of the element in an object. In case of
+   * elements which need to be loaded, such as images, it is recommended
+   * to call the function after the element has finished loading.
    *
    * @method size
    * @return {Object} the width and height of the element in an object
    * @example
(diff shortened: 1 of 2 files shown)


Ajayneethikannan:
@lmccart I have added the documentation , and removed the previously done changes to the size() function.
I request the maintainers to review the changes,
Thank you!

lib/addons/p5.dom.js
```diff
@@ -1965,15 +1965,20 @@
...
+   * to call the function after the element has finished loading.
    *
    * @method size
    * @return {Object} the width and height of the element in an object
    * @example
    * <div class='norender'><code>
-   * var div = createDiv('this is a div');
+   * const div = createDiv('this is a div');
    * div.size(100, 100);
+   * const img = createImg('assets/laDefense.jpg', () => {
```
lmccart: I would suggest making both of these `let` instead of `const`. even though they are not modified again in this example, it is not uncommon that users might do so.

Ajayneethikannan:
@lmccart I have made the requested changes, thanks !

Issue #3463, opened 2019-01-18
createImg causing cross-origin errors on cross-origin image loads.


msawired:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [x] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [x] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 0.7.2
- Web browser and version:  Google Chrome | 71.0.3578.98
- Operating System: OS X Mojave
- Steps to reproduce this:

CreateImg function throws error when images that are loaded from no-cross-origin policy external sites are added into the canvas. When cross-origin images are loaded to the canvas, browser would limit certain access to that data (e.g. using it with ml5js functions, texImage2D, etc.);
Example: https://www.openprocessing.org/sketch/655545 (see browser console for tainted canvas error)

Solution: crossOrigin = 'Anonymous' attribute should be added to createImg function, to prevent any cross-origin issues.
//p5.dom.js, line 435
p5.prototype.createImg = function() {
    p5._validateParameters('createImg', arguments);
	var elt = document.createElement('img');
	elt.crossOrigin = 'Anonymous';

PR #3482, opened 2019-01-26
Added support for cross origin requests

sc231997:
closes: #3463 

elt.crossOrigin = 'Anonymous' attribute added

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -435,6 +435,7 @@
   p5.prototype.createImg = function() {
     p5._validateParameters('createImg', arguments);
     var elt = document.createElement('img');
+    elt.crossOrigin = 'Anonymous';
     var args = arguments;
     var self;
     var setAttrs = function() {


sc231997:
> I think it would be better if you would include a short description along with the issue number in the PR, with the catchphrase "closes: #XXXX", which would help the maintainers to automatically close the related issue 😃. Also the title should not contain just the issue number.

Sorry. Actually, I am new to this. I will keep that in mind from next time.

sanketsingh24:
I think if you would use "closes: #xxxx" or "fixes: #xxxx" it will link the issue.

outofambit:
Thank you for providing some more context @sc231997! and welcome to the p5 repo! 😃

lmccart:
thanks @sc231997!

Ajayneethikannan:
I am very  sorry for commenting on a merged Pull Request, 
I came to know about CORS recently only.
I think setting img.crossOrigin = "Anonymous" always, even if the image is in the same domain,
results in 2 HTTP requests, resulting in unnecessary overhead for same domain images .
(I am not very clear about CORS, please correct me if I am wrong)
If necessary,  we can change it in such a way that we are setting the crossOrigin attribute , only if the src belongs to a different domain.
Thank you!

outofambit:
Hi @Ajayneethikannan, thanks for sharing your concern! Would you please open a new issue for this and we can discuss there? I’m not familiar with that behavior for CORS and I think its worth investigating. Thank you!

Ajayneethikannan:
@outofambit I have opened a new issue, as you suggested for discussion, 
Thank you!

Issue #3390, opened 2018-12-11
i18n for FES messages


kjhollen:
Like #3384, we'd like to be able to change the language for error messages generated from the Friendly Error System. This ticket is to track updating the FES code to support internationalization; tickets for actual translations will be filed in the future.

outofambit:
looking at using https://www.i18next.com/ for this!

Issue #3384, opened 2018-12-11
i18n for console messages


kjhollen:
#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

In various parts of the codebase, there are console error messages for the end user hardcoded in English. We could internationalize messages to provide better support for users who speak other languages. Here are some files where I found such error messages:

[ ] src/webgl/3d_primitives.js
[ ] src/webgl/p5.Camera.js
[ ] src/webgl/p5.RendererGL.js
[ ] src/webgl/p5.Matrix.js
[ ] src/webgl/p5.Shader.js
[ ] src/webgl/p5.Texture.js
[ ] src/core/structure.js
[ ] src/math/p5.Vector.js
[ ] src/typography/loading_displaying.js
[ ] src/typography/p5.Font.js
[ ] src/webgl/p5.Geometry.js

This feature enhancement is not meant to cover translations themselves, but rather the architecture required to allow use of translations in the future.

Bernice55231:
It's been a while, but we, @al6862  @anpanring would like to take over this issue. Could we try it out first based on pull request #3488?

limzykenneth:
@Bernice55231 This is pretty old and I'm not sure still relevant especially considering we have an FES system that can support translated messages as well.

ksen0:
This is approved for implementation in a minor release of 2.x, but it's not quite fleshed out for assignment just yet. To move forward, this issue needs an **implementation proposal** as well as a volunteer to work on it.

_PS: This message is part of cleaning up the [p5.js 2.0 roadmap](https://github.com/orgs/processing/projects/21) prior to release, to make outstanding more more approachable for new contributors. You can join the conversation about this new version on [this Discourse thread](https://discourse.processing.org/t/dev-updates-p5-js-2-0-you-are-here/46130) or this [GitHub thread](https://github.com/processing/p5.js/issues/7488)!_

**If you're new to p5.js contribution, welcome!** Please be sure to read through the [contributor guidelines](https://p5js.org/contribute/contributor_guidelines/), and keep in mind that you should not file a pull request (or start working on code changes) without a corresponding issue or before an issue has been approved for implementation and assigned.

ombalgude:
Is anyone currently working on this? If not, I'd like to take it on. This would be my first contribution to open source.

ksen0:
[...]

PR #3488, opened 2019-01-31
i18n for console and FES messages

ermenauta:
This PR aims to provide a simple architecture proposal to internationalize console errors and FES messages. In reference to #3384  and #3390.

Language files are stored in _src/core/i18n/_ as json key/value entries. The language is selected at runtime, depending on the user browser language. If the user language is not supported (there is no dictionary for the language), then the english language file will be used.

`{
    "HELLO": "Hello World!!!",
    "HELLO_NAME": "Hello {0}!!!"
}`

## Usage
Import the module
`var i18n = require('./i18n/locales');`

The simplest scenario:
`console.log(i18n.__('HELLO')); // Hello World!!!`

For parametric strings we need to pass the array of values as second parameter of the function:
`console.log(i18n.__('HELLO_NAME', ['p5.js'])); // Hello p5.js!!!`

A fallback system act as follows:

- If the key is not found on the current language dictionary, the english dictionary is used
- If the key is not found in the english dictionary, print the key itself

Note: this is my first contribution, all kind of feedback is welcome :)

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -19,6 +19,7 @@ require('./core/shape/2d_primitives');
 require('./core/shape/attributes');
 require('./core/shape/curves');
 require('./core/shape/vertex');
+require('./core/i18n/locales');
 
 // color
 require('./color/color_conversion');
--- a/src/core/i18n/en.json
+++ b/src/core/i18n/en.json
@@ -1,0 +1,4 @@
+{
+    "HELLO": "Hello World!!!",
+    "HELLO_NAME": "Hello {0}!!!"
+}
--- a/src/core/i18n/es.json
+++ b/src/core/i18n/es.json
@@ -1,0 +1,4 @@
+{
+    "HELLO": "Hola Mundo!!!",
+    "HELLO_NAME": "Hola {0}!!!"
+}
--- a/src/core/i18n/locales.js
+++ b/src/core/i18n/locales.js
@@ -34,13 +34,13 @@ var getString = function(str) {
   return dict[str] || defaultDict[str] || str;
 };
 
-function interpolate(str, args) {
+var interpolate = function(str, args) {
   var regex = /{\d}/;
   var _r = function(p, c) {
     return p.replace(regex, c);
   };
   return args.reduce(_r, str);
-}
+};
 
 var localize = function(str, args) {
   var localizedString = getString(str);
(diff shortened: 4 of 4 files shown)


outofambit:
@michelefenu thanks so much for opening this PR (and welcome to the p5js repo 💖)! I'm super excited to see some thinking and work around this topic. Will take a look this weekend and see if I have any thoughts!

michelefenu:
@outofambit thanks for the welcome! I await for your comments :) 

I left things like pluralization out of this implementation to keep thing simple, but if the overall structure is fine I can add them in the future.

limzykenneth:
Overall looks like a good start to me. A few minor points:
1. Any particular reason for the exported function be named `__` instead of something a bit more descriptive?
2. You may want to consider escaping in the interpolation function. I don't think it will cause problem now but a good thing to have just in case. Can ES6's template string be utilized somehow maybe?
3. `getUserLanguage()` function's method of grabbing the first part of the language code will in the most case work but in Chinese for example it won't work as the second part of the Chinese language code (`CN` in `zh-CN`, and `TW` in `zh-TW`) denote the type of script used, despite actually denoting the region. I would suggest a more fine grain and manual approach which is to use regex to pick out those that we support and know will only need their first part then if the regex didn't match, try to match those that need the whole part, then if also not match fallback to English (that reminds me that the website needs to be updated for this as well). But this is all a bit messy still. 

Also consider using `navigator.languages`, it provides a full list of languages that the browser supports and not just the current active one.

michelefenu:
1. Only for brevity, since it's the only exported function and the module does only one thing. But, yeah, we can definitely use a more descriptive name
2. I think the template string might be the cleanest solution
3. Thanks for the suggestions! If implemented properly this will not be so messy :)

Spongman:
IMHO better to use the named function declaration syntax `function name() {}` than the anonymous syntax `var name = function() {}` sine the former gives the functions names in the debugger.

limzykenneth:
> Only for brevity, since it's the only exported function and the module does only one thing. But, yeah, we can definitely use a more descriptive name

Yeah then a more descriptive name would be preferred for clarity in teh codebase

> Thanks for the suggestions! If implemented properly this will not be so messy :)

I didn't mean your implementation is messy, just that dealing with natural languages in code is always gonna be messy. 😄

outofambit:
hey @ermenauta i'm going to close this pull request since its been a while, but feel free to reopen it whenever you're ready to look at this again! ping me if you have any questions :) thanks!

dickinyourass139 (approved):

PR #3489, opened 2019-02-01
Feature: Implemented shininess()

sanketsingh24:
Implemented shininess() method. Based on shininess, which I found to be unmodifiable in the shaders.

Features:

- Changes the shininess factor to the user-specified value.
- Is chainable.

Work Done:

- Implemented the method in material.js.
- Added Documentation and example.
- Added a manual test which is based on [this example](https://p5js.org/examples/3d-materials.html).

Need suggestion on where to place the manual test.

Before: shininess was hard-coded and set to 32.0.
Now: user can change the value of shininess using `shininess()` function. Defaults to 1.0.


stalgiag:
Looks great!

The manual test location is good.

The only change I would request is that the example show a contrasting shininess so it is easier understand what is being changed. Also the stroking may serve as a distraction in the example. [The example that Processing uses](https://processing.org/reference/shininess_.html) might be useful as inspiration here.

sanketsingh24:
@stalgiag please review :smiley:

src/webgl/material.js
```diff
@@ -476,6 +476,49 @@ p5.prototype.specularMaterial = function(v1, v2, v3, a) {
...
+ * </div>
+ * @alt
+ * Shininess on Camera changes position with mouse
+ */
+p5.prototype.shininess = function(shine) {
+  this._assert3d('shininess');
+  p5._validateParameters('shininess', arguments);
+
+  var shader = this._renderer._useLightShader();
+  shader.setUniform('uShininess', shine);
```
stalgiag: I think you should handle when user passes in 0 because this breaks the shader. I would say that 1 is a fine minimum.

src/webgl/material.js
```diff
@@ -476,6 +476,49 @@ p5.prototype.specularMaterial = function(v1, v2, v3, a) {
...
+ * }
+ * function draw() {
+ *   background(0);
+ *   noStroke();
+ *   let locX = mouseX - width / 2;
+ *   let locY = mouseY - height / 2;
+ *   ambientLight(60, 60, 60);
+ *   pointLight(255, 255, 255, locX, locY, 50);
+ *   specularMaterial(250);
+ *   translate(-23, 0, 0);
```
stalgiag: I think you can use -25 and positive 50 for the two x-translations here. Best to keep numbers as "rounded" as possible in examples

src/webgl/material.js
```diff
@@ -476,6 +476,49 @@ p5.prototype.specularMaterial = function(v1, v2, v3, a) {
...
+ *   shininess(1);
+ *   sphere(20);
+ *   translate(46, 0, 0);
+ *   shininess(10);
```
stalgiag: I think you can make this number higher to show a larger range of possible effects

src/webgl/material.js
```diff
@@ -476,6 +476,49 @@ p5.prototype.specularMaterial = function(v1, v2, v3, a) {
...
+ *   background(0);
+ *   noStroke();
+ *   let locX = mouseX - width / 2;
+ *   let locY = mouseY - height / 2;
+ *   ambientLight(60, 60, 60);
+ *   pointLight(255, 255, 255, locX, locY, 50);
+ *   specularMaterial(250);
+ *   translate(-23, 0, 0);
+ *   shininess(1);
+ *   sphere(20);
```
stalgiag: Perhaps worth making these spheres a little more detailed so that the per-vertex lighting doesn't look so odd.

sanketsingh24:
Thanks for the review, I think this PR will be quite affected by #3535, so I will be looking at that first and then making the suggested changes :+1:

src/webgl/material.js
sanketsingh24: @stalgiag Can you please elaborate more on this?

stalgiag: Hm I think you can dismiss this, issues with per vertex lighting looking bad in examples is more or less inevitable. This is a separate problem. But just for clarification, I was suggesting that you pass higher than default values into the optional detail parameters of sphere for these examples to make them look better.

sanketsingh24: Oh, I thought you meant something else by "detailed" :sweat_smile:

stalgiag:
Hi @sanketsingh24 everything looks good. If you want to resolve the small conflict then we can move forward with a merge 😃

sanketsingh24:
Oh I did not notice that :sob: Will fix it now.

sanketsingh24:
@stalgiag Please review

stalgiag:
Thanks @sanketsingh24 !

PR #3491, opened 2019-02-02
fix radians/degrees conversion error in rotate

### LINKED ISSUE #3490: 0.7.3 not obeying angleMode(DEGREES) in WEBGL mode
#### Nature of issue?

- [x] Found a bug

- [x] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.7.3
- Web browser and version: Chrome 71.0.3578.98
- Operating System: Mac 10.14.2

- Steps to reproduce this:

It seems like p5.js 0.7.3 is not obeying angleMode(DEGREES) in WEBGL mode. Rotations stay in Radians. This worked fine in 0.7.2

An easy way to see this is just on the Reference Page examples for "rotate()" and "rotateZ()". If you insert angleMode(DEGREES) anywhere before the rotation, it is obeyed in the rotate() example, but not in the rotateZ() example.

I tried to reproduce this at https://editor.p5js.org/grege2/sketches/TDIyVt9tT but it works ok there. Possibly the Web Editor is still on 0.7.2 or earlier (it would be good to be able to query the version of p5.js that you're running on). 

I've checked it thoroughly in my own app code and it seems a hard error. My WEBGL app rotates fine in numerous DEGREES modes at 0.7.2, but goes crazy at 0.7.3 (because all rotations are 180/PI too large, easily spotted !) I can see from other discussions about angleMode() in the last few months that there has been some looseness in all the _fromRadians / _toDegrees etc calls that has received quite a few fixes recently, possibly there's still a gremlin in there.

I'd be happy to pursue this myself, but if anyone feels they can nail it quickly, feel free. I'll pursue it anyway, it's a good learning exercise.

Regards, GE

Spongman:
closes #3490 

this fixes an error in angle units conversion related to the camera changes.

the matrix methods _only_ _ever_ take radians. this fixes two places where either degrees or mode-dependent units were passed.

### DIFF
--- a/src/webgl/p5.Camera.js
+++ b/src/webgl/p5.Camera.js
@@ -456,7 +456,7 @@ p5.Camera.prototype._rotateView = function(a, x, y, z) {
   centerZ -= this.eyeZ;
 
   var rotation = p5.Matrix.identity(this._renderer._pInst);
-  rotation.rotate(a, x, y, z);
+  rotation.rotate(this._renderer._pInst._toRadians(a), x, y, z);
 
   // prettier-ignore
   var rotatedCenter = [
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -753,7 +753,6 @@ p5.RendererGL.prototype.rotate = function(rad, axis) {
   if (typeof axis === 'undefined') {
     return this.rotateZ(rad);
   }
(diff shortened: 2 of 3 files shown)


grege2:
I would support that. There seems to be more risk than you would expect with these seemingly simple degrees/radians conversions.

stalgiag:
Hi @Spongman thanks for the fix and the tests!

PR #3495, opened 2019-02-04
Add doc note to avoid loop() inside setup().

mhairston:
Documentation change.

I wasted hours trying to debug a sketch in which draw() was somehow unable to access a global variable (it came up as undefined). I had been assuming that I should use noLoop() or loop() in setup(), depending on which I wanted. In fact, I think it used to work. But only noLoop() is OK in setup.

### DIFF
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -76,6 +76,8 @@ p5.prototype.noLoop = function() {
  * within it. However, the <a href="#/p5/draw">draw()</a> loop may be stopped by calling <a href="#/p5/noLoop">noLoop()</a>.
  * In that case, the <a href="#/p5/draw">draw()</a> loop can be resumed with loop().
  *
+ * Avoid calling loop() from inside setup().
+ *
  * @method loop
  * @example
  * <div><code>


Spongman:
calling `loop()` in `setup()` _does_ do weird things, but i'm not seeing the 'unable to access a global variable' error. is there a sketch you can share that points to this behavior?

i see a couple of issues with `loop()` currently:
1) calling `loop()` inside `draw()` when looping is currently enabled, causes a recursive call to draw which can go infinite.
2) calling `loop()` inside `setup()` also calls `draw()` which can cause weird side-effects.

here's a sketch that illustrates this:

https://codepen.io/Spongman/pen/WPEbeB?editors=0011

mhairston:
Here's a reduced case: https://www.openprocessing.org/sketch/664506

The global only gets defined later in setup(). I assume the infinite loop prevents the sketch from getting there. It's not all globals, though -- it works fine if the global is a Boolea, as in this version: https://www.openprocessing.org/sketch/664507

The broken version has an Object.

meiamsome:
@mhairston 
Ah yep, the error you're seeing is `f.status` in particular, `f` is `undefined` in the first `draw` because `loop` is calling the `draw` during `setup` before you initialize `f`, so `f.status` is an error, because `undefined` variables cannot have properties. In your second example though you're just checking the value of `f` - `f === true` which gets converted into `undefined === true` which is valid syntactically and so doesn't cause the error. 

So that error message in particular is based in your sketch, but the underlying error of the `loop()` calling `draw()` during `setup` is one of p5s, if that makes sense?

Issue #3242, opened 2018-09-30
Enhancement: result for loop()


mkrasuski:
#### Nature of issue?

- [ ] Found a bug
- [x] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [x] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [ ] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 
- Web browser and version: 
- Operating System: 
- Steps to reproduce this:

#### Feature enhancement details:
Would be useful to have result for `loop()` as result of the previous state: `true` for enabled looping and `false` for stopped. For example:

```js
if (loop()) noLoop();
```
would switch looping on/off...

mebble:
Yeah not being able to read the state without changing it is a bad idea. An explicit method would be very useful, or perhaps a global variable? p5 already has `keyIsPressed` etc, so a variable like `isLooping` would be consistent with the API.

Also could you explain what you mean by making `loop()` and `noLoop()` chainable?

limzykenneth:
Changing the behaviour of `loop()` and `noLoop()` will probably break a lot of existing code as they are used pretty often so it probably will not be considered, a global variable however I don't see why it won't be possible to have that. Overall I would like to have a way to query if the sketch is looping or not as well, otherwise I will need to have keep track on my own with a toggling variable when I need it.

mebble:
I'm new to p5 but I'd love to work on this. The basic idea is something like the following code snippet
```js
// In /src/core/structure.js
p5.prototype.loop = function() {
  this._loop = true;
  this._setProperty('isLooping', true);  // the new variable
  this._draw();
};
```

limzykenneth:
@mebble Do go ahead with your PR, if you can include tests and documentation then it will help greatly with getting it merge. If you need with with those, checkout the [developer docs](https://github.com/processing/p5.js/tree/master/developer_docs) or ask anyone here.

lmccart:
[...]

PR #3497, opened 2019-02-04
prevent recursive `draw()` calls, explicitly and from within `loop()`

### LINKED ISSUE #3496: loop() issues
re: #3242, #3495

i see a couple of issues with loop() currently:

1) calling loop() inside draw() when looping is currently enabled, causes a recursive call to draw which can go infinite.
2) calling loop() inside setup() also calls draw() which can cause weird side-effects.

here's a sketch that illustrates this:

https://codepen.io/Spongman/pen/WPEbeB?editors=0011

Spongman:
closes #3496 
re: #3242, #3495

this prevents recursive calls of `draw()` from within itself, within `setup()` and from within `loop()` all of which cause various issues discussed above.

### DIFF
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -341,8 +341,7 @@ p5.prototype.redraw = function(n) {
       this._inUserDraw = true;
       try {
         userDraw();
-      }
-      catch (err) {
+      } catch (err) {
         this._inUserDraw = false;
         throw err;
       }
--- a/src/core/structure.js
+++ b/src/core/structure.js
@@ -109,8 +109,12 @@ p5.prototype.noLoop = function() {
  */
 
 p5.prototype.loop = function() {
-  this._loop = true;
-  this._draw();
+  if (!this._loop) {
+    this._loop = true;
+    if (this._setupDone) {
+      this._draw();
+    }
+  }
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/core/structure.js
```diff
@@ -308,6 +312,10 @@ p5.prototype.pop = function() {
  *
  */
 p5.prototype.redraw = function(n) {
+  if (this._inUserDraw || !this._setupDone) {
```
meiamsome: Note that this breaks calling `redraw` from within `draw` manually, which could be useful. However, I'd be willing to say that that is not an important use case.

src/core/structure.js
```diff
@@ -330,7 +338,13 @@ p5.prototype.redraw = function(n) {
       }
       context._setProperty('frameCount', context.frameCount + 1);
       context._registeredMethods.pre.forEach(callMethod);
-      userDraw();
+      this._inUserDraw = true;
+      try {
+        userDraw();
+      } catch (err) {
+        this._inUserDraw = false;
```
meiamsome: This should be in a `finally` block

src/core/structure.js
Spongman: indeed. although the docs above this do state that you probably shouldn't be calling it in `draw` anyway.

src/core/structure.js
Spongman: fantastic, thanks!

stalgiag:
Thanks for the work @Spongman and @meiamsome !

Issue #3494, opened 2019-02-03
circle(radius) breaks my code and teaching resources, future plans for it to change again?


makeyourownalgorithmicart:
Hi

I [previously](https://github.com/processing/p5.js/issues/2305) proposed changes to p5js to make the language easier for younger and non-technical newcomers, in line with the processing foundations own aims and objectives for reach and accessibility. They were reacted at the time.

I went on and created my own [simple.js](https://github.com/makeyourownalgorithmicart/simple.js/wiki) library which implemented changes which addressed issues found from actual teaching experience with very young children, first-time coders, and newcomcers from the arts. It included several changes such as a `repeat()` instruction, consistency aliases, spelling options, friendlier random number function .. and a `circle()` and `square()` function.

The library was included in the popular [openprocessig.org](https://www.openprocessing.org), as an example of uptake.

Since then another [suggestion](https://github.com/processing/p5.js/issues/3375), referring to my original proposal, was made to add `circle()`. This was approved.

Sadly it broke compatibility with my own code. 

My own teaching experience suggests that newcomers / young people find the idea of **diameter** more intuitive for a circle than **radius**. My own code used the third parameter as a **diameter**. The recent p5js 0.7.3 release uses it as a **radius**.

So this breaks my code, and teaching resources.

**So my question is - as go through all that code and teaching resources, will p5js suddently change this behaviour again?**

Personally I wish there had been some consultation - especially amongst those of us who had expended a lot of time and care trying to raise the issue, and the do something about it. I also think some news / awareness / social media discussion around this attempt to make p5js more friendly and accessible would also be good. It would address any perception that the upstream developers are not really that interested in such human issues.

hellonearthis:
Why not just have diameter = radius * 2 or use diameter / 2 when calling circle/ellipse .
The center point of  circles and rectangles are also different unless rectmode is used.
Random looks mostly the same, apart from yours has a floor included. 

With the few number of functions in simple,js it should be easy to update that to work with P5.js

lmccart:
Hey @makeyourownalgorithmicart thanks for your post! We did in fact implement `circle()` directly based on your feedback and are super excited about your library. Sorry it took a while. I think we just got the implementation details not quite right. An enthusiastic contributor responded to the request for circle() addition and it got merged, all before there was a lot of time for comment. We're totally open to considering using diameter, I think this makes some sense and we can discuss. 

[...]

PR #3499, opened 2019-02-05
change 3rd parameter of circle to be diameter instead of radius

montoyamoraga:
This pull request addresses #3494
I changed both the definition and the documentation of the circle() function,
so that its third parameter by default is diameter, instead of radius.

### DIFF
--- a/src/core/shape/2d_primitives.js
+++ b/src/core/shape/2d_primitives.js
@@ -230,7 +230,7 @@ p5.prototype.ellipse = function(x, y, w, h, detailX) {
  * white circle with black outline in mid of canvas that is 55x55.
  */
 p5.prototype.circle = function() {
-  let args = Array.prototype.slice.call(arguments, 0, 2);
+  var args = Array.prototype.slice.call(arguments, 0, 2);
   args.push(arguments[2]);
   args.push(arguments[2]);
   this.ellipse.apply(this, args);
--- a/src/core/shape/2d_primitives.js
+++ b/src/core/shape/2d_primitives.js
@@ -210,18 +210,18 @@ p5.prototype.ellipse = function(x, y, w, h, detailX) {
  * Draws a circle to the screen. A circle is a simple closed shape.
  * It is the set of all points in a plane that are at a given distance from a given point, the centre.
  * This function is a special case of the ellipse() function, where the width and height of the ellipse are the same.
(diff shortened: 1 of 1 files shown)


hellonearthis:
I think it should default to radius and use ellipseMode to change the default to diameter.

This pull would break too much.

src/core/shape/2d_primitives.js
```diff
@@ -230,9 +230,9 @@ p5.prototype.ellipse = function(x, y, w, h, detailX) {
  * white circle with black outline in mid of canvas that is 55x55.
  */
 p5.prototype.circle = function() {
-  var args = Array.prototype.slice.call(arguments, 0, 2);
-  args.push(2 * arguments[2]);
-  args.push(2 * arguments[2]);
+  let args = Array.prototype.slice.call(arguments, 0, 2);
```
meiamsome: `let` is not yet permitted in the source of p5, please use `var` for now

intuitivetextmining:
default should be diameter

intuitivetextmining:
I've made changes which 

- work and are tested with sketches
- also update the documentation

[https://github.com/processing/p5.js/issues/3494#issuecomment-460859136](https://github.com/processing/p5.js/issues/3494#issuecomment-460859136)

src/core/shape/2d_primitives.js
montoyamoraga: thanks! just fixed it.

montoyamoraga:
> I've made changes which
> 
> * work and are tested with sketches
> * also update the documentation
> 
> [#3494 (comment)](https://github.com/processing/p5.js/issues/3494#issuecomment-460859136)

hi! sorry i didn't notice your comment, I think we wrote the same thing?
I just went to the file and changed everything to diameter, and deleted the 2* part.
I think this pull request matches your proposed change, right?

makeyourownalgorithmicart:
thanks @montoyamoraga - let's hope your PR goes through and the next release is soon!

makeyourownalgorithmicart:
thanks everyone!

do we know when the next release is? (and will this be in it?)

I'm asking anyone to rush - but if the timeframe is long then I'll put circle back into my simple.js as I use it for teaching ...

lmccart:
thanks @montoyamoraga. @makeyourownalgorithmicart this will be in the next release. there is no scheduled date as of right now.

PR #3500, opened 2019-02-06
circle() 3rd param changed from radius to diameter

makeyourownalgorithmicart:
discussed here https://github.com/processing/p5.js/issues/3494


sanketsingh24:
Although I didn't take part in the conversation, I do feel this is the same PR as #3499, except that here `let` is not used, and so tests are passed. 

Also, you should use phrases like `fixes: #xxxx ` or `closes: #xxxx' which would close the related issue when the PR is merged.

lmccart:
Thanks @makeyourownalgorithmicart as @sanketsingh24 mentions there's an open PR with this change already so I'd like to respect the work of that contributor and merge that one if we can. In any case, rest assured, this change will definitely be in the next release.

makeyourownalgorithmicart:
thanks @lmccart - that makes sense and is the right thing to do.

PR #3504, opened 2019-02-08
fix MediaElement use of Renderer2D methods

### LINKED ISSUE #3278: loadPixels() does not work for video elements
#### Nature of issue?

- [X] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [X] Other (specify if possible) DOM: Video

- [X] Desktop/Laptop

#### Details about the bug: 

- p5.js version:  0.7.2
- Web browser and version:  Chrome 69.0.3497.100
- Operating System:  Mac OSX 10.13.4
- Steps to reproduce this:

I noticed that [this code](https://github.com/CodingTrain/website/tree/master/Tutorials/P5JS/p5.js_video/11.4_p5.js_brightness_mirror) (it's Coding Train example) doesn't work in 0.7.2, but does work in 0.6.1.

When running loadPixels(), video.pixels is still empty. I tried this for both a video capture and loading in a video from a .mov file.

loadPixels() and pixels does work for images, however.

using video.get(x, y) returns a garbage value.

### LINKED ISSUE #3359: Video Element Do not refresh when using video.get()
#### Nature of issue?

- [ x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x ] Other Capture dom element 

- [ x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 
- Web browser and version: chrome 70.0.3538.110
- Operating System: 
- Steps to reproduce this:windows10

I just try to Reproduce this coding train lesson:
https://editor.p5js.org/leiyufazhuangjia/sketches/rJdRvpQkV
But it seems that the Video Element Do not refresh automatically,the particals always just get the color of the first frame of the video element, is it a bug?

Spongman:
fixes #3278
fixes #3359
re: discussion in #3150

the `loadPixels` manual test example is currently broken for two reasons:
1) the chrome auto-load video blocker, fixed with a `mouseClicked` handler.
2) #3150, which is a necessary fix for the `mousePressed` issue awoke the sleeping bear that is `Image` & `MediaElement` re-use of `Renderer2D` methods.

this PR splits out the `_pInst` and a new `_ctx` renderer field that holds the rendering context for these shared methods. it's initialized in the appropriate constructors:

- for `Renderer` objects, both `_pInst` and `_ctx` are the `p5` instance (or `Graphics`).
- for `Image` objects, ~~both `_pInst` and~~ `_ctx` are the image itself.
- for `MediaElement` objects the `_pInst` is the `p5` instance, and `_ctx` is the element.

i also made `MediaElement.get` use `p5.get` instead of `Renderer2D.get` so it can re-use the parameter conditioning. this got lost in another commit a while back.


meiamsome:
I am not sure I understand why the Image's `_pInst` is itself and not a p5 instance as the name would suggest? Or why not undefined? I feel like `<something>._pInst` should always be a p5 instance, or is this not the case?

Spongman:
you're right, i have removed it. `Image` has not had a `_pInst` member previously. i added it for symmetry, but it's not actually used.

stalgiag:
Hi @Spongman thanks for this chain of fixes. I think that these changes should mostly likely be merged but I just wanted to briefly bring up a bit of confusion that I had when trying to wrap my head around what is going on.

I think that setting  `_ctx` to the `p5` instance for `Renderer` objects feels natural but I have a bit more difficulty following why `_ctx` is the object itself for `MediaElement` and `Image` objects. This makes sense to me when following this series of issues and thinking about `_ctx` in reference to the pixel state but feels a bit confusing otherwise. I think this is mainly because the `_ctx` variable name reads like a short-hand for "context" and I would never expect the "context" for an object to be the object itself. Not sure if I am alone in thinking this is odd but just wanted to bring it up so that we can make sure this is as clear as possible for future devs.

Spongman:
The name `ctx` comes from the existing code in the renderer2d methods that assign a local variable `var ctx = this._pInst || this`.

This PR is just calculating this `ctx` variable in the object’s constructor (where it can be accurately determined). A `_` is prepended to the name as it’s a private member.

We can change the name of you’d like, I just went with what was already there..

stalgiag:
@Spongman that makes sense to me! I think my confusion just stems from trying to follow a problem that centers around a variable name (`_ctx`) that seems to have grown a bit stale. I understand why you kept the name but I am realizing that it might also be a good time to update the variable name. I feel like `ctx` is pretty commonly used for the javascript canvas drawing context. If I take your explanation of what the PR does and change `_ctx` to `_pixelsState` it becomes a little easier to follow why the different objects follow different pixel states.

Here as an example:
* for `Renderer` objects, both `_pInst` and `_pixelsState` point to the `p5` instance (or `Graphics`).
* for `Image` objects, both `_pInst` and `_pixelsState` are the image itself.
* for `MediaElement` objects the `_pInst` is the `p5` instance, and `_pixelsState` is the element.

What do you think? Also open to other suggestions. 😄

PR #3506, opened 2019-02-08
Optimize color to string

### LINKED ISSUE #3505: Color.toString() is slow
0.7.3

i noticed, while profiling `MediaElement.get()` using the ‘loadPixels’ manual test, that a large chunk of the total time is taken up in `Color.toString()`.

if you call `stroke()` or `fill()` in a loop, then this method is called a bunch.

image

Spongman:
fixes #3505 

- don't compute HSB or HSL coefficients unless they are used.
- use `string.concat()` to build return value.

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -246,7 +246,7 @@ p5.Color.prototype.toString = function(format) {
 
     case 'rgba':
     default:
-      return 'rgba('.concat(a[0], ', ', a[1], ', ', a[2], ', ', alpha, ')');
+      return 'rgba('.concat(a[0], ',', a[1], ',', a[2], ',', alpha, ')');
   }
 };
 
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -84,9 +84,6 @@ p5.Color = function(pInst, vals) {
  * canvas with text representation of color
  */
 p5.Color.prototype.toString = function(format) {
-  if (!this.hsba) this.hsba = color_conversion._rgbaToHSBA(this._array);
-  if (!this.hsla) this.hsla = color_conversion._rgbaToHSLA(this._array);
-
   var a = this.levels;
   var f = this._array;
   var alpha = f[3]; // String representation uses normalized alpha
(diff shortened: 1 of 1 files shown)


src/color/p5.Color.js
```diff
@@ -241,7 +246,7 @@ p5.Color.prototype.toString = function(format) {
 
     case 'rgba':
     default:
-      return 'rgba(' + a[0] + ',' + a[1] + ',' + a[2] + ',' + alpha + ')';
+      return 'rgba('.concat(a[0], ',', a[1], ',', a[2], ',', alpha, ')');
```
outofambit: @spongman, do you know this `.concat` approach compares to using a template string? I feel like the template string might be slightly more readable (but i'm not sure if there would be a performance cost).

sanketsingh24: I think I read somewhere that `.concat` works a lot faster with DOM, although I would do a check just to be sure.

sanketsingh24: I just did a [perftest](https://jsperf.com/perftestp5854876528) and `.concat` is faster (probably because of array).

Spongman:
Arent’ template string literals ES6?

src/color/p5.Color.js
meiamsome: As @Spongman mentions in a comment, we can't use template strings in the p5 source yet because we don't have ES6. @sanketsingh24 's perftesting is checking string `+` not template literals. I made a version of the same test with two more ways to calculate the same thing and also avoids the browser optimizing out the operation entirely (Happens on the current Firefox, check the browser scope results on the old one to see that): https://jsperf.com/p5-color-test
Template literals are faster, it's a shame we can't use them, but just using `+` seems to be better performance wise on both chrome and FF that I've tested, but that's potentially differences in the underlying hardware and OS (Linux, Ryzen here)

sanketsingh24: On my system, `.concat` is still faster (Arch Linux)

outofambit: ah sorry @Spongman and thanks @meiamsome. totally skipped my brain.

`.concat` is faster for me on MacOS as well, but not by much. However, I'm fine with this change in the name of consistency, as @Spongman points out. What if we merge this as-is and then revisit when we can use es6?

PR #3519, opened 2019-02-14
Fix loadPixels

Spongman:
~~(based on #3504, diffs: https://github.com/Spongman/p5.js/compare/fix-pInsta...Spongman:fix-loadPixels)~~

fixes #3191 
fixes #3363

this ensures that `_pixelsDirty` is set after drawing operations in `Renderer2D`.

~~I will rebase after #3504 is merged.~~

PR #3520, opened 2019-02-14
Fix get()

### LINKED ISSUE #3493: get() returns pixel/image inconsitently
re: https://codepen.io/Spongman/pen/wNeoJd?editors=0011

here's a table of the return type of `get(x,y,w,h)` based on different values for passed parameters.

```
size\coord             0,0         -1,0        -100,-100   undefined,undefined
undefined x undefined  pixel       pixel       pixel       image       
NaN x NaN              image       image       image       image       
0 x 0                  image       pixel       pixel       image       
1 x 0                  image       image       pixel       image       
1 x 1                  pixel       pixel       pixel       pixel       
2 x 2                  image       image       pixel       image       
```

there's a lot of strange inconsitencies here, and IMO _only_ `get(number, number)` should return a pixel array, and _everything_ else should return an `Image` (or perhaps an error).

ie, it I believe it would be more consistent if it read:

```
size\coord             0,0         -1,0        -100,-100   undefined,undefined
undefined x undefined  pixel       pixel       pixel       image       
NaN x NaN              ?           ?           ?           ?          
0 x 0                  image       image       image       image       
1 x 0                  image       image       image       image       
1 x 1                  image       image       image       image       
2 x 2                  image       image       image       image       
```

this would make the actual returned value match up with what is specified in the docs.

Spongman:
~~(based on #3519, i will rebase after that is merged, diffs: https://github.com/Spongman/p5.js/compare/fix-loadPixels...Spongman:fix-get)~~

fixes #3493

* unifies the `get` logic into a single function
* breaks out the renderer-specific code into a separate `_readPixel` function.
* implements type-specific overloads in `get()`
  * get() : returns an image
  * get(x,y) : returns a pixel
  * get(x, y, w, h) : returns an image
* docs for the above
* `get()` now calls `_validateParameters` (no, it's not a perf issue)
* adds some tests to ensure the above.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -2880,7 +2880,8 @@
     this.setModified(true);
     return this;
   };
-  p5.MediaElement.prototype.get = function(x, y, w, h) {
+  p5.MediaElement.prototype.get = p5.Renderer2D.prototype.get;
+  p5.MediaElement.prototype._getPixel = function() {
     if (this.loadedmetadata) {
       // wait for metadata
       var currentTime = this.elt.currentTime;
@@ -2889,16 +2890,10 @@
       } else {
         this._ensureCanvas();
       }
-
-      return p5.prototype.get.apply(this, arguments);
-    } else if (typeof x === 'undefined') {
-      return new p5.Image(1, 1);
-    } else if (w > 1) {
-      return new p5.Image(x, y, w, h);
-    } else {
-      return [0, 0, 0, 255];
     }
+    return p5.Renderer2D.prototype._getPixel.apply(this, arguments);
   };
+
   p5.MediaElement.prototype.set = function(x, y, imgOrCol) {
     if (this.loadedmetadata) {
       // wait for metadata
--- a/test/unit/image/pixels.js
+++ b/test/unit/image/pixels.js
@@ -39,7 +39,7 @@ suite('pixels', function() {
       img.updatePixels();
     });
 
-    test('works with integers', function() {
+     test('get(x,y) works with integers', function() {
       assert.deepEqual(img.get(25, 25), [255, 0, 0, 255]);
       assert.deepEqual(img.get(25, 26), [0, 0, 255, 255]);
       assert.deepEqual(img.get(0, 0), [255, 0, 0, 255]);
--- a/test/unit/image/pixels.js
+++ b/test/unit/image/pixels.js
@@ -53,7 +53,7 @@ suite('pixels', function() {
       assert.deepEqual(img.get(50, 26), [0, 0, 0, 0]);
     });
 
-    test('get() works', function() {
+    test('get() returns a copy when no arguments are supplied', function() {
       var copy = img.get();
       assert.instanceOf(copy, p5.Image);
       assert.equal(copy.width, img.width);
--- a/test/unit/image/pixels.js
+++ b/test/unit/image/pixels.js
@@ -39,7 +39,7 @@ suite('pixels', function() {
       img.updatePixels();
     });
 
-     test('get(x,y) works with integers', function() {
+    test('get(x,y) works with integers', function() {
       assert.deepEqual(img.get(25, 25), [255, 0, 0, 255]);
       assert.deepEqual(img.get(25, 26), [0, 0, 255, 255]);
       assert.deepEqual(img.get(0, 0), [255, 0, 0, 255]);
--- a/test/unit/image/pixels.js
+++ b/test/unit/image/pixels.js
@@ -46,7 +46,7 @@ suite('pixels', function() {
       assert.deepEqual(img.get(49, 49), [0, 0, 255, 255]);
     });
 
-    test('works when out of bounds', function() {
+    test('get(x,y) returns 0s for out of bounds arguments', function() {
       assert.deepEqual(img.get(25, -1), [0, 0, 0, 0]);
       assert.deepEqual(img.get(-1, 26), [0, 0, 0, 0]);
       assert.deepEqual(img.get(25, 50), [0, 0, 0, 0]);
--- a/test/unit/image/pixels.js
+++ b/test/unit/image/pixels.js
@@ -42,6 +42,39 @@ suite('pixels', function() {
     test('works with integers', function() {
       assert.deepEqual(img.get(25, 25), [255, 0, 0, 255]);
       assert.deepEqual(img.get(25, 26), [0, 0, 255, 255]);
+      assert.deepEqual(img.get(0, 0), [255, 0, 0, 255]);
+      assert.deepEqual(img.get(49, 49), [0, 0, 255, 255]);
+    });
+
+    test('works when out of bounds', function() {
+      assert.deepEqual(img.get(25, -1), [0, 0, 0, 0]);
+      assert.deepEqual(img.get(-1, 26), [0, 0, 0, 0]);
+      assert.deepEqual(img.get(25, 50), [0, 0, 0, 0]);
+      assert.deepEqual(img.get(50, 26), [0, 0, 0, 0]);
+    });
+
+    test('get() works', function() {
+      var copy = img.get();
+      assert.instanceOf(copy, p5.Image);
+      assert.equal(copy.width, img.width);
+      assert.equal(copy.height, img.height);
+
+      assert.deepEqual(copy.get(25, 25), [255, 0, 0, 255]);
+      assert.deepEqual(copy.get(25, 26), [0, 0, 255, 255]);
+      assert.deepEqual(copy.get(0, 0), [255, 0, 0, 255]);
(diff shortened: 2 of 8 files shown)


lib/addons/p5.dom.js
```diff
@@ -2880,7 +2880,8 @@
     this.setModified(true);
     return this;
   };
-  p5.MediaElement.prototype.get = function(x, y, w, h) {
+  p5.MediaElement.prototype.get = p5.Renderer2D.prototype.get;
```
stalgiag: Why not return in the same fashion as p5.Image.get for consistency?

stalgiag: Also I understand that the p5.Renderer2D is more or less just pointing upwards to p5.Renderer.get but is there a reason why you don't make that explicit here so it is easier to know where to go looking for the method logic?

Spongman: i think it's clearer this way to see that it's just reusing another classes implementation.

i think this is preferable to the way that p5.Image does it since in order to determine what it's doing, you have to first parse the code, work out if the argument lists are the same, determine whether or not it matters that the argument list is changed, etc...

Spongman: i did it that way because I think it's `Renderer2D`'s business to know how its `get` is implemented, not `MediaElement`'s. if, down the line, someone decides to implement `Renderer2D.get` differently, this line would still work.

Spongman: how about just adding a line to Renderer2D:
```javascript
// p5.Renderer2D.prototype.get = p5.Renderer.prototype.get;
```

commented out because it's a no-op.

stalgiag: That seems like a good call!

stalgiag: Thanks for the explanation! I know it is breaking from the repo's convention a bit but I agree that this is clearer and we should probably move towards doing this wherever it's applicable. My only concern would be that the single-line approach disappears a little bit in the organization of the code-base when visually scanning, maybe just something that would eventually require consistent placement if this is done a lot. I say change `p5.Image.prototype.get` to match this style but maybe we should get a second opinion first. @outofambit what do you think?

outofambit: I think a better solution here would to have the implementation of `p5.Renderer.prototype.get` be unattached to any class implementation (and so just be `get` in its own file). Each class that uses it could pass in all the necessary data. it looks like that is only:

* `x, y, w, h`
* `pixelsState`
* `canvas`
* `getPixel`?

or maybe

* `x, y, w, h`
* `this`

Each class that uses this implementation (p5.Renderer, p5.Image, p5.MediaElement) could internally reference this single implementation. It looks like a similar pattern would work for `getPixel` as well.

What do you think @Spongman and @stalgiag? Happy to explain more if I'm not totally making sense.

test/unit/image/pixels.js
```diff
@@ -42,6 +42,39 @@ suite('pixels', function() {
...
+    test('get(x,y,w,h) works', function() {
+      for (var w = 1; w < img.width + 5; w += 2) {
+        for (var x = -w * 2; x <= img.width + w * 2; x += 4) {
+          var copy = img.get(x, x, w, w);
+          assert.instanceOf(copy, p5.Image);
+          assert.equal(copy.width, w);
+          assert.equal(copy.height, w);
+          assert.deepEqual(copy.get(0, 0), img.get(x, x));
+        }
+      }
```
sanketsingh24: Wouldn't it be better if there's a test for get(x,y) too?

stalgiag: This is present in several places here unless I am mistaken.

lib/addons/p5.dom.js
Spongman: i would rather keep the implementation in a common base class. the fact that `Image` & `MediaElement` have to 'borrow' the `get` implementation from outside their hierarchy is something i'd rather solve by cleaning up their hierarchies. IMO it's better to move toward increased encapsulation rather than away.

stalgiag: I think I agree with @Spongman here. I can imagine ways in which what you propose would be nice, with shared functions being detached from any class, but my personal feeling is that this would be moving towards a pretty big change for how the library is structured which is something I am always reluctant about.

outofambit: can we add some commented text above that line to explain why its there?

Spongman: i added the comment `// .get() is not overridden`

outofambit: Fair enough. If this pattern is present outside of this PR I'm ok with treading water here. Just to be clear though, my suggested implementation is a move towards encapsulation and away from implicit assumptions about classes and objects in the implementation of `get` and `_getPixel`.

test/unit/image/pixels.js
```diff
@@ -42,6 +42,39 @@ suite('pixels', function() {
...
+    });
+
+    test('works when out of bounds', function() {
+      assert.deepEqual(img.get(25, -1), [0, 0, 0, 0]);
+      assert.deepEqual(img.get(-1, 26), [0, 0, 0, 0]);
+      assert.deepEqual(img.get(25, 50), [0, 0, 0, 0]);
+      assert.deepEqual(img.get(50, 26), [0, 0, 0, 0]);
+    });
+
+    test('get() works', function() {
```
outofambit:
```suggestion
    test('get() returns a copy when no arguments are supplied', function() {
```

test/unit/image/pixels.js
outofambit: they are, but they are lumped in with another test case. it took me a while to find them. going request they be a separate `it` block

test/unit/image/pixels.js
```diff
@@ -42,6 +42,39 @@ suite('pixels', function() {
...
+      var copy = img.get();
+      assert.instanceOf(copy, p5.Image);
+      assert.equal(copy.width, img.width);
+      assert.equal(copy.height, img.height);
+
+      assert.deepEqual(copy.get(25, 25), [255, 0, 0, 255]);
```
outofambit: can these `assert`s be in their own descriptive `it` block since they test different arguments format and resulting behavior?

test/unit/image/pixels.js
sanketsingh24: Ok, these tests are literally littered with `get(x,y)` :sweat_smile: but wouldn't it be better if there was an explicit test for `get(x,y)`, something which @outofambit also suggests?

test/unit/image/pixels.js
Spongman: can you expand on this a bit? i'm not sure what you mean by a 'descriptive it block'. also, these asserts aren't really testing the `get(x,y)` functionality itself (that's testing in a previous test), they're just testing that the pixels in the `copy` are the correct values - it just happens that `get(x,y)` is the clearest way to extract a pixel from an image ;-)

test/unit/image/pixels.js
Spongman: i think the `works with integers` and `works when out of bounds` tests cover this.

test/unit/image/pixels.js
outofambit: >  also, these asserts aren't really testing the get(x,y)

ah ok! that's the context i was missing. i'm good with this test as-is.

test/unit/image/pixels.js
outofambit: ah ok! cool. i'm going to suggest a couple edits to make that clearer. thanks!

test/unit/image/pixels.js
```diff
@@ -42,6 +42,39 @@ suite('pixels', function() {
     test('works with integers', function() {
```
outofambit:
```suggestion
     test('get(x,y) works with integers', function() {
```

test/unit/image/pixels.js
```diff
@@ -42,6 +42,39 @@ suite('pixels', function() {
...
       assert.deepEqual(img.get(25, 25), [255, 0, 0, 255]);
       assert.deepEqual(img.get(25, 26), [0, 0, 255, 255]);
+      assert.deepEqual(img.get(0, 0), [255, 0, 0, 255]);
+      assert.deepEqual(img.get(49, 49), [0, 0, 255, 255]);
+    });
+
+    test('works when out of bounds', function() {
```
outofambit:
```suggestion
    test('get(x,y) returns 0s for out of bounds arguments', function() {
```

test/unit/image/pixels.js
stalgiag: Hi @outofambit and @Spongman are you two okay with committing the two small test description changes so that we can move forward with merge? I don't want to let this and #3521 sit here for too long.

stalgiag:
Thanks @Spongman for the considered implementation! Thanks to all of the reviewers!

PR #3531, opened 2019-02-24
Fix single radius on square()

### LINKED ISSUE #3525: square(x,y,s,r) doesn't work
0.7.3

https://p5js.org/reference/#/p5/square

the middle example should have rounded corners.

image

this fixes it for me:
```javascript
p5.prototype.square = function(x, y, s, tl, tr, br, bl) {
  this.rect(x, y, s, s, tl, tr, br, bl);
};
```

Stefterv:
resolves #3525

slice index accidentally skipped the fourth parameter at index 3

### DIFF
--- a/src/core/shape/2d_primitives.js
+++ b/src/core/shape/2d_primitives.js
@@ -517,7 +517,7 @@ p5.prototype.rect = function() {
 p5.prototype.square = function() {
   var args = Array.prototype.slice.call(arguments, 0, 3);
   args.push(arguments[2]);
-  args = args.concat(Array.prototype.slice.call(arguments, 4));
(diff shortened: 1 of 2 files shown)


Spongman:
can't we do this without all the messing about with `arguments`? such use of the `arguments` array causes deoptimization in many js jits, and IMHO is significantly harder to read than the fix i listed in #3525.

maybe i'm missing something?

outofambit:
hi @Stefterv and thanks so much for this pull request! I've verified this fixes the problem stated in #3525. (and thanks for the tests!! 😻)

@Spongman i agree avoiding `arguments` is a worthwhile refactor, but that extra optimization step shouldn't block this PR from getting merged. i'll open a new issue to track that optimization. thanks!

Issue #3510, opened 2019-02-10
multiple elements having same id


Ajayneethikannan:
#### Nature of issue?

- [x] Existing feature enhancement

#### Most appropriate sub-area of p5.js?

- [x] Core/Environment/Rendering

#### Which platform were you using when you encountered this?

- [x] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [x] Others (specify if possible) - independent of device

#### Details about the bug: 

- p5.js version: 0.7.3 
- Web browser and version: Google Chrome : 72.0.3626.96 (Official Build) (64-bit)
- Operating System: Ubuntu 16.04 
- Steps to reproduce this:

This example sketch can be used to see that p5 allows people to change the ids of multiple elements to be the same.
https://editor.p5js.org/TheSkepticSniper/sketches/PGRAjeMy1

#### Feature enhancement details:
Hello! Currently the id function in p5.Element.js allows users to change id of several elements to the same id. Though having multiple elements with the same Id does not raise errors in the browser, it is not preferable to have it so. Also using classes is a better approach for such an use case. Since it is not wrong to have multiple elements with the same id, I believe we should also not stop the user from having multiple elements with the same id.
Since we can select only one element with the given id, by using the inbuilt function document.getElementById, and also the select method defined in p5.js, this can lead to some undefined behavior too.

#### New feature details:
For this issue, the changes will be made in the id function defined in the p5.Element.js module
The changes:
- Raising a simple warning, with a helpful message for beginners in the console if there is already an element, existing with the id we are providing in the argument.

Ajayneethikannan:
I would love to solve this issue, if it is considered as a valid enhancement. Thank you!

lmccart:
I like this warning idea and we're currently thinking about adding some functionality for Friendly Error System that checks the logic of arguments passed in. That's a longer term effort. For now, a good fix might be to add some info to the documentation recommending that only one element have a particular id per sketch, and directing users to look at class() instead.

lmccart:
resolved with #3534

PR #3534, opened 2019-02-25
Added note for multiple ids in one page

Ajayneethikannan:
Solving #3510

The change made in this PR is : 
Added a note for users that multiple ids cannot be present in a single page , and refer to classes instead.

@lmccart I have made the changes in the documentation as instructed, though it does not close the issue. I request you to review the changes ,

Thank you!

### DIFF
--- a/src/core/p5.Element.js
+++ b/src/core/p5.Element.js
@@ -123,6 +123,9 @@ p5.Element.prototype.parent = function(p) {
  *
  * Sets the ID of the element. If no ID argument is passed in, it instead
  * returns the current ID of the element.
+ * Note that only one element can have a particular id in a page.
+ * The <a href="#/p5.Element/class">.class()</a> function can be used
+ * to identify multiple elements with the same class name.
  *
  * @method id
  * @param  {String} id ID of the element


src/core/p5.Element.js
```diff
@@ -122,7 +122,9 @@ p5.Element.prototype.parent = function(p) {
 /**
  *
  * Sets the ID of the element. If no ID argument is passed in, it instead
- * returns the current ID of the element.
+ * returns the current ID of the element.Note that only one element can have
```
limzykenneth: Missing space before "Note" here.

src/core/p5.Element.js
```diff
@@ -122,7 +122,9 @@ p5.Element.prototype.parent = function(p) {
...
+ * a particular id in a page. Refer to <a href="#/p5.Element/class">.class()</a>
+ * function to set names for multiple elements.
```
limzykenneth: I personally wouldn't call it "name" as for input fields "name" is a very specific attribute and may cause confusion here.

Ajayneethikannan: @limzykenneth Thank you for the suggestion !
Would rephrasing the sentence as ,
".class() function can be used to identify multiple elements with the same class name." make it better ?

limzykenneth: Yeah that would be better

Ajayneethikannan:
@limzykenneth I have done the suggested changes , thank you!

lmccart:
thank you @Ajayneethikannan!

PR #3542, opened 2019-02-27
clean 'prettier-ignore' from examples

### LINKED ISSUE #3541: prettier-ignore displays in two code examples
Currently `prettier-ignore` lines written inside of examples are displaying as comments on the examples. This isn't a pressing issue but could easily be a bit perplexing for people reading the reference. This currently is only a problem in the [bezierDetail example](https://p5js.org/reference/#/p5/bezierDetail) and the incoming RendererGL [applyMatrix example](https://github.com/processing/p5.js/commit/51900bcdb88d96b3df8e0d2de23b0c2285727dba).

If you use `<!-- prettier-ignore-start -->` & `<!-- prettier-ignore-end -->`. The docs build nicely but the linter throws errors. 

Could perhaps be fixed by [`prettier-ignore-start`](https://github.com/prettier/prettier/issues/5287) for JS if it is added, but there is probably another way around this as well.

Spongman:
fixes #3541

just add a step to the docs preprocessor to regex-strip the `// prettier-ignore` comments out of the docs examples.

### DIFF
--- a/docs/preprocessor.js
+++ b/docs/preprocessor.js
@@ -30,6 +30,16 @@ function smokeTestMethods(data) {
   });
 }
 
+function cleanExamples(data) {
+  data.classitems.forEach(function(classitem) {
+    if (classitem.itemtype === 'method' && classitem.example) {
+      classitem.example = classitem.example.map(i =>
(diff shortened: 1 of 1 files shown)


stalgiag:
Wonderful thank you!

PR #3543, opened 2019-02-27
correction for winMouseX, winMouseY examples

Ajayneethikannan:
This fixes https://github.com/processing/p5.js-website/issues/394

- remove unnecessary spacing in the comments
- appending the canvas element to the body of the webpage so that the example is not affected by the
relative css of the website.

Though the change made here acts as a solution, 
I am extremely in support of an extra parameter in the postion function , allowing use to choose between different position options Absolute, static, fixed, relative , sticky.

I request the maintainers to review the PR.


lmccart:
thank you!

Issue #3356, opened 2018-12-03
Transparency looks different in WEBGL/non WEBGL


hamoid:
#### Nature of issue?

- [x] Found a bug

#### Most appropriate sub-area of p5.js?

- [x] Color
- [x] Core/Environment/Rendering
- [x] WebGL

#### Which platform were you using when you encountered this?

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: I couldn't find out. The one today at https://editor.p5js.org/
- Web browser and version: Firefox 63.0.3
- Operating System: ArchLinux
- Steps to reproduce this:
```javascript
function drawOnto(P, x, y, bg) {
  P.background(bg)
  P.noStroke();
  P.fill("rgba(250, 64, 24, 0.5)")
  P.ellipse(x, y, 100, 100);
}
new p5(function(P) {
  P.setup = function() {
    P.createCanvas(200, 100, P.WEBGL);
    drawOnto(P, 0, 50, 0);
  };
});
new p5(function(P) {
  P.setup = function() {
    P.createCanvas(200, 100);
    drawOnto(P, 100, 0, 0);
  };
});
new p5(function(P) {
  P.setup = function() {
    P.createCanvas(200, 100, P.WEBGL);
    drawOnto(P, 0, 50, 255);
  };
});
new p5(function(P) {
  P.setup = function() {
    P.createCanvas(200, 100);
    drawOnto(P, 100, 0, 255);
  };
});
```
![2018-12-03-085651_191x389_scrot](https://user-images.githubusercontent.com/108264/49361045-36a73a80-f6db-11e8-8e55-03c62a66ef2f.png)
Both halfs of the circles should be equal. Half is rendered with WEBGL, half not.

This is the original color: `rgba(250, 64, 24, 0.5)`
These are the resulting colors:
```
Background  WebGL                 Non-WebGL
Black       rgb(157, 87, 72)      rgb(125, 32, 12)
White       rgb(252, 183, 168)    rgb(252, 159, 139)
```
It seems like Non WebGL does a linear interpolation, and WebGL does something else.

stalgiag:
A workaround that fixes this is calling `setAttributes('alpha', false)` on the WebGL canvases. The todo here might be to set 'alpha' to true by default. With 'alpha' set to false you won't be able to have transparency on the backbuffer but this seems like a rare use case, and those users can call `setAttributes`.

I can fix this in #3552 since I am already changing how `setAttributes` sets defaults.

[...]

PR #3552, opened 2019-03-01
setAttributes before GL canvas is made

### LINKED ISSUE #3061: WebGL setAttributes() leads to multiple contexts being created
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

Hi there. :wave: I'm using WebGL and one of the first things I do is turn on MSAA since it dramatically improves the visual quality (and later, if performance starts to become an issue, turn it back off). 

I noticed that the current structure of p5.js does not allow you to create a canvas with specified attributes; instead, you have to create a WEBGL canvas, then call `p5.setAttributes` after to enable MSAA. The result is that the first GL context is destroyed, and a new one created.

This is not a huge deal, although it's a bit wasteful and produces some doubled-up console noise.

```
p5.RendererGL: enabled webgl context
p5.RendererGL: enabled webgl context
```

I think a clean fix might be to allow `setAttributes` to be called before hand (right now it throws an error), thus storing the attributes until the first `createCanvas` call. Thoughts?

### LINKED ISSUE #3180: WEBGL, smooth() and parent() do not seem to work correctly.
Update: The new online editor is nice! So I used it (and a bit of CSS) to help [demo this issue](https://editor.p5js.org/rmichael/sketches/BkOWCDvwQ). :-)

#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.7.1
- Web browser and version: 61.0.2
- Operating System: MacOSX 10.12.6
- Steps to reproduce this:

Apologies if I've misunderstood usage of instance mode, `smooth()`, and `parent()`.

In instance mode, `smooth()`, `parent()`, and the `WEBGL` renderer do not correctly interact.  I have not yet tested this in global mode.  Perhaps there is a problem coordinating the reference returned by createCanvas and that used in the instance?

This does not seem to happen with the `P2D` renderer, presumably because the smooth implementation is different.

Finally, there is a problem with `noSmooth()` and the 2D renderer, it is mentioned at the end of this issue.

Given this code:

```javascript
var sketch = function (s) {

  s.setup = function () {

    // s.smooth(); // No effect; if called here, the shapes will not actually be smoothed.

    var canvas;
    canvas = s.createCanvas(400, 400, s.WEBGL);

    // Case 1
    // s.smooth();  // If called here, the canvas is duplicated -- one will be at the end of the document,
                    // and it will be the one used in draw(), with the grey background and ellipse.
                    // Use the web inspector to see the duplicate canvas, and the console logs the
                    // "p5.RendererGL: enabled webgl context" a second time.

    canvas.parent('test');  // This correctly moves the blank canvas, returned by createCanvas.
                            // If case 1, draw() uses the original canvas, which has remained
                            // at the end of the document.

    // Case 2
    // s.smooth();  // If called here, the canvas is not duplicated and is correctly used by draw,
                    // but it remains at the end of the document and is not moved by .parent()

  }

  s.draw = function () {
    s.background(150, 150, 150);
    s.ellipse(0, 0, 100, 100);
  }
}

new p5(sketch);
```

As a minor matter, it appears `smooth()` has no effect if called before the canvas has been created.  This seems strange and is not mentioned in the documentation.

There are several cases, which depend upon when `smooth()` is called, they are discussed in the code above.  If `smooth()` is not called at all, `parent()` does move (not duplicate) the canvas and it is used by draw() correctly.

In addition, `noSmooth()` and the `P2D` does not appear to work.  2D shapes default to smoothed, but trying to un-smooth them does not work, code below.

```javascript
var sketch = function (s) {

  s.setup = function () {

    var canvas;
    canvas = s.createCanvas(400, 400);
    canvas.parent('logo');
    s.noSmooth();
  }

  s.draw = function () {
    s.background(150, 150, 150);
    s.ellipse(s.width / 2, s.height / 2, 100, 100);
  }
}

new p5(sketch);
```

stalgiag:
This allows `smooth`, `noSmooth` and `setAttributes` to be called before a WEBGL canvas is made. WebGL drawing contexts require their attributes to be set at initialization and reinitialization is not desirable in some cases so this allows attributes to be set before the canvas is first created.

fixes #3061 & fixes #3180


stalgiag:
A few things that aren't ideal:
updating `p5._presetAttributes` in `Renderer2D.prototype.smooth`, this is done because before a WEBGL canvas has been made the Renderer is 2D. This could be moved up to `p5.prototype.smooth` but then it would be called even when the Renderer is GL which is unecessary.

storing the `_presetAttributes` at the `p5` level. Open to other ideas.

@Spongman and @sanketsingh24 please review

src/webgl/p5.RendererGL.js
```diff
@@ -157,6 +143,20 @@ p5.RendererGL.prototype = Object.create(p5.Renderer.prototype);
 // Setting
 //////////////////////////////////////////////
 
+p5.RendererGL.prototype._setAttributeDefaults = function(attr) {
+  var attributes = attr || this._pInst._presetAttributes || {};
```
sanketsingh24: Hi @stalgiag I can see a problem here, look at this case-

```js
function setup() {
  setAttributes('perPixelLighting', true);
  createCanvas(1000, 600, WEBGL);
}

function draw() {
    // code
}

function mousePressed() {
  setAttributes('perPixelLighting', false);
 }
```

Upon pressing the mouse, `antialias` will become `false`, as it reverts to its default value,

Probably using `Object.assign({}, attr, this._pInst._presetAttributes)` will correct the issue.

stalgiag: Hi @sanketsingh24! You are right. Though that resetting to defaults on everything that isn't explicitly set behavior that you are describing is the current implementation but I agree that this isn't necessarily what people should expect. Alternatively we can always update the presetAttributes regardless of renderer and always use those. This will behave a little differently but is probably a more organized implementation.

stalgiag:
Now both the 2D and 3D renderer's modify and store the gl attributes in `p5.instance._glAttributes`. This way the attributes can be modified before making a WebGL canvas. Also, it is now possible to put calls to `smooth()` and `setAttributes()` in the draw loop as they will only actually do work and reinitialize context when there is an actual modification, repeated calls to assign the same setting will be ignored.

stalgiag:
fixes #3356

Issue #3555, opened 2019-03-01
Travis build for Node 8 fail sometimes


stalgiag:
The Travis builds for Node 8 are failing occasionally. The problem seems to be [that this unit test](https://github.com/processing/p5.js/blob/master/test/unit/webgl/p5.Texture.js#L111) occasionally fails because the texture being passed is sometimes [not passing this conditional ](https://github.com/processing/p5.js/blob/master/src/webgl/p5.Texture.js#L311). 

~~Note that this test never fails with Node 8 running on my local machine or Node 10 on Travis.~~

stalgiag:
Correction: this is failing locally.

Probably some discrepancy between the way that textures are stored and retrieved introduced with the merging of #3535 and #3537. (see [RendererGL._tex](https://github.com/processing/p5.js/blob/master/src/webgl/p5.RendererGL.js#L134) vs [RendererGL.textures](https://github.com/processing/p5.js/blob/master/src/webgl/p5.RendererGL.js#L128). I will revert the latter PR merge today if the problem isn't tracked down.

stalgiag:
Update: Apparently this only fails locally sometimes

sanketsingh24:
Ok so problem is [here](https://github.com/processing/p5.js/blob/master/test/unit/assets/nyan_cat.gif), called [here](https://github.com/processing/p5.js/blob/master/test/unit/webgl/p5.Texture.js#L15). This GIF has a size of `50*50 pixels`, so width is `50 pixels` and vice versa, which is not a power of `2`. Sometimes (like, the most of times), for some reason, `texImg1` is `1*1 pixels`, so at those times the test passes without any fuss, but sometimes, GIF loads up and those cases, the test fails. For more info, go [here](https://www.diffchecker.com/KhFb5lvz).

sanketsingh24:
Changing [`setup()`](https://github.com/processing/p5.js/blob/master/test/unit/webgl/p5.Texture.js#L11) to this would make sure `texImg1` is loaded and ready to go, and would make the tests consistent, now just need to replace GIF with one which has dimensions in power of 2s.
```js
  setup(function() {
    myp5 = new p5(function(p) {
      p.setup = function() {
        p.createCanvas(100, 100, p.WEBGL);
        texImg1 = p.loadImage('unit/assets/nyan_cat.gif', function(img) {
          p.texture(img);
        });
        texImg2 = p.loadImage('unit/assets/target.gif');
        // texture object isn't created until it's used for something:
[...]

PR #3556, opened 2019-03-01
Fixed failing test case in p5.texture

sanketsingh24:
Fixes: #3555

Fixed failing test cases, and a typo (Instead of checking height, It was set to width).

### DIFF
--- a/src/webgl/p5.Texture.js
+++ b/src/webgl/p5.Texture.js
@@ -296,7 +296,7 @@ p5.Texture.prototype.setWrapMode = function(wrapX, wrapY) {
   };
 
   var widthPowerOfTwo = isPowerOfTwo(this.width);
-  var heightPowerOfTwo = isPowerOfTwo(this.width);
+  var heightPowerOfTwo = isPowerOfTwo(this.height);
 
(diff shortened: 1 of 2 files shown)


sanketsingh24:
For some reason, these tests were failing before in my local machine, not now, but are failing in travis build now :anger: I'll see into this.

stalgiag:
Hi @sanketsingh24 I think this change should happen as that looks like a typo but probably not the reason for #3555

sanketsingh24:
Yea got that a little late :sweat: fixing #3555 now and then will review your PR

sanketsingh24:
These checks don't fail in my local machine now, ran both node v8.15.0 & v10.15.2, so I am removing the link and just fixing the typo.

stalgiag:
Perfect thanks @sanketsingh24!

I think that eventually these tests should use the  `promisedSketch` / `testSketchWithPromise` methods as @meiamsome suggested in #3555 but I am going to merge this now and open an issue for that change. Thanks again!

sanketsingh24:
I also thought that, but due to urgency, I just went with this implementation. :+1:

PR #3562, opened 2019-03-02
Fix contributions file

EndBug:
I noticed that there was an error when generating contributors due to a missing `"`.
I committed both the fixed `.all-contributorsrc` and the resulting `README.md`


lmccart:
thank you!

PR #3581, opened 2019-03-09
add test for modeAdjust helper function

ihsavru:


test/unit/core/helpers.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * @requires constants
+ * @requires helpers
+ */
+
+var constants = require('../../../src/core/constants');
```
ihsavru: @outofambit I am not quite sure how to import the constants or the helper function here and use them for testing. Using `require` doesn't work, so I am unable to run these tests.

outofambit: hey @ihsavru that's a really good point. i did some digging and it looks like this will be better off being included in the node tests (`test/node/`). sorry i missed this the first time! i tried it out and had some weird issues with mocha. I'm still investigating this, but in the meantime you can take a look at https://github.com/processing/p5.js/blob/master/test/node/test-docs-preprocessor.js for a better example to follow.

ihsavru:
@outofambit moved it to test/node

PR #3582, opened 2019-03-09
Make downKeys variable instance specific, fixes #3193

### LINKED ISSUE #3193: multiple sketches: keyPressed in only one of them, but keyTyped in all of them
related to Issue #1492

#### Nature of issue?

- [X] Found a bug

- [X] Events

- [X] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.7.2
- Web browser and version: Chrome Version 68.0.3440.106 (Official Build) (64-bit)
- Operating System: MacOS 10.13.6
- Steps to reproduce this:
See example in https://codepen.io/anon/pen/bxYWKj

If there is more than one sketch, the keyPressed events are received only by one of them. Issue #1492 mentions, that is working correct as per design (keyPressed is global), but the related event keyTyped does fire in all sketches.

Is it possible, that the "global nature" of the keyEvents changed since June 2016, but has only been adapted for some keyboard related events in p5.js?

keyPressed should fire for all sketches on a web page, like other events related to mouse or keyTyped for regular characters.

Ajayneethikannan:
Fixes #3193 

The changes made in this PR are ,
- Change downKeys variable from global to separate for every instance , so that every instance's keyPressed function is allowed to run .
- Write a unit test, which confirms this behavior.  (with great help from @meiamsome ) 

I request the maintainers to review the PR and suggest the changes , if necessary .
Thank you !

### DIFF
--- a/src/core/main.js
+++ b/src/core/main.js
@@ -575,6 +575,8 @@ p5.prototype._initializeInstanceVariables = function() {
   };
 
   this._pixelsDirty = true;
+
+  this.downKeys = {}; //Holds the key codes of currently pressed keys
 };
 
 // This is a pointer to our global mode p5 instance, if we're in
--- a/src/core/main.js
+++ b/src/core/main.js
@@ -576,7 +576,7 @@ p5.prototype._initializeInstanceVariables = function() {
 
   this._pixelsDirty = true;
 
-  this.downKeys = {}; //Holds the key codes of currently pressed keys
+  this._downKeys = {}; //Holds the key codes of currently pressed keys
 };
 
 // This is a pointer to our global mode p5 instance, if we're in
--- a/src/events/keyboard.js
+++ b/src/events/keyboard.js
@@ -10,12 +10,6 @@
 var p5 = require('../core/main');
 
 /**
- * Holds the key codes of currently pressed keys.
- * @private
- */
-var downKeys = {};
-
-/**
  * The boolean system variable <a href="#/p5/keyIsPressed">keyIsPressed</a> is true if any key is pressed
  * and false if no keys are pressed.
  *
@@ -174,14 +168,14 @@ p5.prototype.keyCode = 0;
  *
  */
 p5.prototype._onkeydown = function(e) {
-  if (downKeys[e.which]) {
+  if (this.downKeys[e.which]) {
     // prevent multiple firings
     return;
   }
   this._setProperty('isKeyPressed', true);
   this._setProperty('keyIsPressed', true);
   this._setProperty('keyCode', e.which);
-  downKeys[e.which] = true;
+  this.downKeys[e.which] = true;
   this._setProperty('key', e.key || String.fromCharCode(e.which) || e.which);
   var keyPressed = this.keyPressed || window.keyPressed;
   if (typeof keyPressed === 'function' && !e.charCode) {
@@ -224,9 +218,9 @@ p5.prototype._onkeydown = function(e) {
  */
 p5.prototype._onkeyup = function(e) {
   var keyReleased = this.keyReleased || window.keyReleased;
-  downKeys[e.which] = false;
+  this.downKeys[e.which] = false;
 
-  if (!areDownKeys()) {
+  if (!this._areDownKeys()) {
     this._setProperty('isKeyPressed', false);
     this._setProperty('keyIsPressed', false);
   }
@@ -304,7 +298,7 @@ p5.prototype._onkeypress = function(e) {
  * been released.
  */
 p5.prototype._onblur = function(e) {
-  downKeys = {};
+  this.downKeys = {};
 };
 
 /**
@@ -381,25 +375,25 @@ p5.prototype._onblur = function(e) {
  */
 p5.prototype.keyIsDown = function(code) {
   p5._validateParameters('keyIsDown', arguments);
-  return downKeys[code];
+  return this.downKeys[code];
 };
 
 /**
- * The checkDownKeys function returns a boolean true if any keys pressed
+ * The _areDownKeys function returns a boolean true if any keys pressed
  * and a false if no keys are currently pressed.
 
- * Helps avoid instances where a multiple keys are pressed simultaneously and
+ * Helps avoid instances where multiple keys are pressed simultaneously and
  * releasing a single key will then switch the
  * keyIsPressed property to true.
  * @private
 **/
-function areDownKeys() {
-  for (var key in downKeys) {
-    if (downKeys.hasOwnProperty(key) && downKeys[key] === true) {
+p5.prototype._areDownKeys = function() {
+  for (var key in this.downKeys) {
+    if (this.downKeys.hasOwnProperty(key) && this.downKeys[key] === true) {
       return true;
     }
   }
   return false;
-}
+};
 
 module.exports = p5;
--- a/src/events/keyboard.js
+++ b/src/events/keyboard.js
@@ -168,14 +168,14 @@ p5.prototype.keyCode = 0;
  *
  */
 p5.prototype._onkeydown = function(e) {
-  if (this.downKeys[e.which]) {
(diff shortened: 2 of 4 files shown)


Ajayneethikannan:
@outofambit thank you very much for reviewing the test !
I am totally up for **simplicity** of the tests !

Along those lines, I think we can make it even more simpler, by making the runSketchesWithKeyPress a synchronous function like this , 
```
function runSketchesWithKeyPress(number) {
  let ready = 0;
  function sketch_function(sketch, resolve) {
    sketch.setup = function() {
      ready++;
      if (ready === number) {
        window.dispatchEvent(new KeyboardEvent('keydown'));
      }
    };
    sketch.keyPressed = function() {
      resolve();
    };
  }

  const promises = [];
  for (var i = 0; i < number; i++) {
    promises.push(promisedSketch(sketch_function));
  }

  return Promise.all(promises);
}
```
thus, the code will have only one async function, making it easier for users to understand .

If we can add a little more complexity for preventing timeouts, 
I think we can change the test as
```
suite('Keyboard Events', function() {
  suite('keyPressed', function() {
    test('multiple keyPressed functions must run once', async function() {
      const sketches = await runSketchesWithKeyPress(2);
      assert.deepEqual(sketches, [1, 1]); //sketches with uncalled keyPressed functions, resolve with 0
    });
  });
});

function runSketchesWithKeyPress(number) {
  let ready = 0;
  let unresolved = [];
  function sketch_function(sketch, resolve) {
    sketch.setup = function() {
      ready++;
      unresolved.push(() => resolve(0)); //to be called if the keyPressed function is not fired.
      if (ready === number) {
        window.dispatchEvent(new KeyboardEvent('keydown'));
        setTimeout(resolveUncalledSketches, 0);
      }
    };
    sketch.keyPressed = function() {
      resolve(1);
    };
  }
  function resolveUncalledSketches() {
    for (var j = 0; j < unresolved.length; j++) {
      unresolved[j]();
    }
  }
  const promises = [];
  for (var i = 0; i < number; i++) {
    promises.push(promisedSketch(sketch_function));
  }

  return Promise.all(promises);
}

```

which still separates the functions, and allows a promised sketch to resolve with 0, if the keyPressed function is not called. (the naming can be changed, for understandability) .

But I believe that a **simpler** test is far better , as suggested by you. 

Can you please suggest what are the changes I should make ,
Thank you very much !

meiamsome:
@outofambit Be wary of putting anything outside of a `suite`, that'll get put on the global scope for all tests which can be problematic.
I totally agree that this test does not read well, and separating it out a bit is good.

My suggestion would be to separate out the test like this:
In the p5 helpers file (test/js/p5_helpers.js) we can add a helper for parallel sketches:
```javascript
function parallelSketches(sketch_fns) {
  var setupPromises = [];
  var resultPromises = [];
  var endCallbacks = [];
  for (var _i = 0; _i < sketch_fns.length; _i++) {
    var i = _i;
    setupPromises[i] = new Promise(function(resolve) {
      resultPromises[i] = promisedSketch(function(sketch, _resolve, _reject) {
        sketch_fns[i](sketch, _resolve, _reject);
        var old_setup = sketch.setup;
        sketch.setup = function() {
          if (old_setup) {
            old_setup();
          }
          resolve();
        }
        endCallbacks[i] = sketch.finish;
      });
    });
  }

  function end() {
    for (var callback of endCallbacks) {
      if (callback) {
        callback();
      }
    }
  }

  return {
    end: end,
    setup: Promise.all(setupPromises),
    result: Promise.all(resultPromises)
  };
}
```
I realise that this implementation is possibly worse than the original... but it means our test becomes far simpler, and mainly focused on what it is actually testing:
```javascript
    test('multiple keyPressed functions must run once', async function() {
      var sketchFn = function(sketch, resolve, reject) {
        let count = 0;
        sketch.keyPressed = function() {
          count += 1;
        };

        sketch.finish = function() {
          resolve(count);
        };
      };
      var parallel = parallelSketches([sketchFn, sketchFn]);
      await parallel.setup; // After this line, ALL sketches are setup
      window.dispatchEvent(new KeyboardEvent('keydown'));
      parallel.end(); // Dispatch the 'end' function of the sketches
      var counts = await parallel.result; // Get the resolved errors. Any reject-ed sketches would be thrown here
      assert.deepEqual(counts, [1, 1]); // Our results object is an array in the same order as parallelSketches
    });
```

Which would mean we can duplicate the tests for other important events (mouse pressed, for example) and anything else dealing with two sketches interacting without duplicating the hard logic. 

Also, if we think a failing test timing out is acceptable (I'm on the fence), we can simplify down to the following:
```javascript
    test('multiple keyPressed functions must run once', async function() {
      var sketchFn = function(sketch, resolve, reject) {
        sketch.keyPressed = resolve;
      };
      var parallel = parallelSketches([sketchFn, sketchFn]);
      await parallel.setup; // After this line, ALL sketches are setup
      window.dispatchEvent(new KeyboardEvent('keydown'));
      await parallel.result; // This result now has no useful data
    });
```
And similar simplifications to the helper:
```javascript
function parallelSketches(sketch_fns) {
  var setupPromises = [];
  var resultPromises = [];
  for (var _i = 0; _i < sketch_fns.length; _i++) {
    var i = _i;
    setupPromises[i] = new Promise(function(resolve) {
      resultPromises[i] = promisedSketch(function(sketch, _resolve, _reject) {
        sketch_fns[i](sketch, _resolve, _reject);
        var old_setup = sketch.setup;
        sketch.setup = function() {
          if (old_setup) {
            old_setup();
          }
          resolve();
        }
      });
    });
  }

  return {
    setup: Promise.all(setupPromises),
    result: Promise.all(resultPromises)
  };
}
```
Though we lose the nicer & clearer error messages, and this technically checks a different condition (That the keyPressed is called _at least_ once instead of _exactly once_)

Let me know what you think about that separation?

outofambit:
> Be wary of putting anything outside of a suite, that'll get put on the global scope for all tests which can be problematic.

ah right! good catch.

> My suggestion would be to separate out the test like this:

@meiamsome this is great, thank you!! i'm on the same page about avoiding timeouts if possible, so I'd prefer your first implementation of `parallelSketches` above.

@Ajayneethikannan what do you think about this approach? if you're on board, perhaps @meiamsome could open a PR to add this `parallelSketches` utility to `master` and then we can use it in this PR? (we could also use it in #3605 😃)

Ajayneethikannan:
@outofambit Sure, I am in !
 I shall wait for parallelSketches utility to be implemented then.
Thank you @outofambit , @meiamsome  !

outofambit:
@Ajayneethikannan awesome!

@meiamsome i'm also happy to put that up in a PR if you'd prefer! just lmk!

outofambit:
@Ajayneethikannan the `parallelSketches` helper is now in `master` so feel free to merge that in and use it whenever you're ready! 🚀

Ajayneethikannan:
I have made the changes , and added comments . Please provide your suggestions :smiley: 
Thank you @outofambit , @meiamsome !

src/core/main.js
```diff
@@ -575,6 +575,8 @@ p5.prototype._initializeInstanceVariables = function() {
   };
 
   this._pixelsDirty = true;
+
+  this.downKeys = {}; //Holds the key codes of currently pressed keys
```
meiamsome: Might we prefer `this._downKeys` to imply this is a private property?

Ajayneethikannan: Oh yes , that is better. I will make the changes .
Thanks !

outofambit: looks good!

Ajayneethikannan:
@meiamsome , I have changed the title , thanks !
We can add a test like the same key variable is in sync for all instances when key is pressed, 
but maybe that should be done in a separate PR.
Otherwise , I think no changes are left.
Thank you !

outofambit:
awesome thank you so much @Ajayneethikannan and @meiamsome! ✨

PR #3588, opened 2019-03-10
Update p5.Element.js

ankur54:
Hi. I have updated the syntax of the code to the latest ES6 version. Please feel free to comment on any additional changes that might be required.

### DIFF
--- a/src/core/p5.Element.js
+++ b/src/core/p5.Element.js
@@ -6,7 +6,7 @@
 
 'use strict';
 
-var p5 = require('./main');
+import * as p5 from "./main";
 
 /**
  * Base class for all elements added to a sketch, including canvas,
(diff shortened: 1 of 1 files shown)


ankur54:
Updated code according to the latest ES6 version.

limzykenneth:
Hi @ankur54, we are currently not migrating the library code base to ES6 yet.

ankur54:
> Hi @ankur54, we are currently not migrating the library code base to ES6 yet.

Oh, but there was an issue raised in GitHub regarding this, and also, there is a GSOC project regarding the shift to ES6. What about them?

ankur54:
Is that the reason why the code is not passing the test cases?

limzykenneth:
We are migrating the library in multiple phases, starting with the reference and examples, that what most of the issues are about. Then we will move onto developer tooling and after that we will then look at how we plan to migrate the source code itself. You can have a look at the [roadmap](https://github.com/processing/p5.js/blob/master/developer_docs/roadmap.md) here.

The test is failing probably because of compatibility issues of the library with ES6 which is why we are putting off this task for now.

ankur54:
So, where can I start contributing? Any suggestions?

limzykenneth:
Have a look at the open issues, especially those that are tagged with `bug`, those would be a good place to start. If you want to work on some ES6 migration you can search for ES6 amongst open issue to see if any reference section or examples that are not migrated yet.

I'm going to close this PR for now.

Issue #3589, opened 2019-03-10
shifting pushMatrix, popMatrix to legacy.js


Ajayneethikannan:
#### Nature of issue?

- [x] Existing feature enhancement

#### Most appropriate sub-area of p5.js?

- [x] Core/Environment/Rendering

#### Which platform were you using when you encountered this?

- [x] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [x] Others (specify if possible) Platform Independent

#### Details about the bug: 

- p5.js version:  0.7.3
- Web browser and version:  Browser independent
- Operating System:  OS independent

#### Feature enhancement details:
By #243 , pushMatrix and popMatrix from processing were removed in favour of push and pop.
My doubts are, 
- Should they be shifted to legacy.js .
- Is the feature printMatrix going to be implemented .

Waiting to hear your suggestions ,
Thank you !

lmccart:
@Ajayneethikannan yes, I think these methods should be moved to legacy.js.
no, we are not planning to implement printMatrix() because the underlying canvas does not track the matrix so to implement this method we would have to add our own tracking of the matrix. we would instead like to direct people to libraries like [this one](https://simonsarris.com/a-transformation-class-for-canvas-to-keep-track-of-the-transformation-matrix/)

Issue #243, opened 2014-06-25
push/popMatrix, push/popStyle, save/restore, push/popState?


lmccart:
I just looked into matrix stuff and realized it's not really implemented.

Canvas supports `save()` and `restore()` which push and pop contexts onto a stack, which consist of the transformation matrix as well as style settings. There are methods for setting the transform of the canvas directly and transforming it in various ways, but there is no default way to access the transformation matrix, meaning it needs to be kept track of internally by hand if you want access.

http://html5.litten.com/understanding-save-and-restore-for-the-canvas-context/

A few options:
1. Keep track of the matrix internally. This means manually multiplying and setting all variables instead of relying on the canvas built in `translate()`, `rotate()`, `transform()` etc. This would slow performance and also leave considerable error for matrix manipulation errors.
2. Have `push/popMatrix()` backed by `save()/restore()`, and keep track of the styles to apply them after calling `restore()`, so that calling `popMatrix()` doesn't reset the style as well as the transform matrix. Probably faster than option 1, but not super elegant to pop off the style just to immediately manually place it back on again.
3. Replace `push/popMatrix()` and `push/popStyle()` with one function `push/popState()` or `push/popContext()`, that is effectively the save as `save()/restore()`, plus maybe a few other things (relating to imageMode, text stuff, etc). I'm in favor of this option as it's the cleanest, but I'd like to hear if there's strong objection from Processing natives.

**It also always felt a little non-intuitive to me that styles carry through from setup to draw and around the draw loop, but transforms do not (ie setup and draw are wrapped with push/popMatrix). I think my proposal would be that setup is wrapped with nothing (everything carries through, transforms and styles), and draw is wrapped with pushState so transforms do not accumulate and styles do not wraparound).

PR #3592, opened 2019-03-11
shift popMatrix, pushMatrix, printMatrix to legacy.js

Ajayneethikannan:
Fixes: #3589

- shift pushMatrix, popMatrix, printMatrix methods to legacy.js
- write a note about how to implement tracking of matrix for printMatrix method

Thanks !

### DIFF
--- a/src/core/legacy.js
+++ b/src/core/legacy.js
@@ -20,4 +20,20 @@ p5.prototype.popStyle = function() {
   throw new Error('popStyle() not used, see pop()');
 };
 
+p5.prototype.popMatrix = function() {
+  throw new Error('popMatrix() not used, see pop()');
+};
+
+p5.prototype.printMatrix = function() {
+  throw new Error(
+    'printMatrix() is not implemented in p5.js, ' +
+      'refer to [https://simonsarris.com/a-transformation-class-for-canvas-to-keep-track-of-the-transformation-matrix/] ' +
+      'to add your own implementation.'
+  );
+};
+
+p5.prototype.pushMatrix = function() {
+  throw new Error('pushMatrix() not used, see push()');
+};
+
 module.exports = p5;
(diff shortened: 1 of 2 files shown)


grege2:
Hi @aj_ryuk, thanks for this PR. I agree it's good to remove unneeded calls, but personally I would recommend not consigning printMatrix() to the dustbin. I think p5.js should actually fully implement printMatrix() and getMatrix(). They really are important tools to debug what's happening with your ModelView matrix. I've been using Iris GL and OpenGL for 30 years and you do need to have "introspection" of the internal state of your GL. If you're doing something complicated and stacking and backtracking transforms, like a "generative art" tree (ie. a physical plant !) you can easily make a mistake. Plus you can make a mistake with the simplest code, rotating something around the screen. A getMatrix() or at least printMatrix() call can help you debug. If you know that your matrix should be like these: [Wikipedia rotate matrix](https://en.wikipedia.org/wiki/Rotation_matrix#Basic_rotations) and it's not, it's a valuable debugging clue. 

Looking forward to other opinions.

Best, GE.

Ajayneethikannan:
Thank you very much for providing your thoughts on this issue @grege2 :smiley: Currently, as per discussions in #243 , the pushMatrix and popMatrix have not been implemented due to errors and performance issue of maintaining the transformation matrix by ourself, and consequently printMatrix. 
I agree that it would be very beneficial for debugging to implement printMatrix, and in this PR, I am only updating the message to say how to implement it in an ad hoc way, by the users. 
In case it is to be implemented , I am all up for it !
Would love to hear others opinions on this .
Thanks !

grege2:
Hi @aj_ryuk. Thanks for the extra explanation. I read through #243. I can see that p5.js doesn't keep track of the matrix, at least not in the traditional way that an OpenGL implementation would, it's held by the Canvas, and it's hard to implement a getMatrix(). Ok, getMatrix()and printMatrix() will have to wait.

Ciao !

PR #3602, opened 2019-03-13
Fix style function in p5.dom.js

### LINKED ISSUE #3548: Calling size() on a p5.graphic object automatically draws it
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.7.3
- Web browser and version: 
- Operating System: Windows
- Steps to reproduce this:

```javascript
function setup() {
    createCanvas(windowWidth, windowHeight, WEBGL)
    graphic = createGraphics(200, 50);
    graphic.size(200, 50)
    graphic.textAlign(CENTER, CENTER);
    graphic.position(width/2, height/2);
}

function draw() {
    background(255);
    graphic.text("HELLO",graphic.width/2,graphic.height/2)
}
```

This is technically a bug, but may be useful for people currently dealing with #2183 . When you create a p5.graphics object and use the `size()` function on it, be it in `setup()` or `draw()`, it will then render over the top of whatever other graphics you are using. It does not need to be called as a texture, it just appears.

This means we now have a one line method to draw 2D text in 3D, but can be very annoying if you actually want to resize a graphics object without its contents appearing onscreen.

Can someone figure out what it is that is causing this, and separate it out into its own function so we can maintain this useful functionality without it being mandatory?

Ajayneethikannan:
Fixes #3548

- Remove setting style using setAttribute and use the style property of the element to set the style instead
  (using setAtrribute to set style [overrides some already set style properties of the element](https://www.w3schools.com/jsref/met_element_setattribute.asp)

- Remove repeated `this.width = aw` and `this.height = aH` line.

Thank you!

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -2010,10 +2010,8 @@
           }
           this.elt.setAttribute('width', aW * this._pInst._pixelDensity);
           this.elt.setAttribute('height', aH * this._pInst._pixelDensity);
-          this.elt.setAttribute(
-            'style',
-            'width:' + aW + 'px; height:' + aH + 'px'
-          );
+          this.elt.style.width = aW + 'px';
+          this.elt.style.height = aH + 'px';
           this._pInst.scale(
             this._pInst._pixelDensity,
             this._pInst._pixelDensity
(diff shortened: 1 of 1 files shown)


Ajayneethikannan:
Thank you for verifying the pull request @outofambit !


>do you think we should look for other places to replace the `setAttribute('style', 'something')` too (in another PR)?

Yes , I think we should check and replace statements where style is set using setAttribute .

Thank you again !

PR #3604, opened 2019-03-13
Changed require to import in app.js

plxity:
Changed all the require statement to 'Import' in app.js


plxity:
Capture
i am getting no error when I "npm run lint" but CI is failing? Can anyone help with this?

Issue #3591, opened 2019-03-11
Dom Drop event does not receive drop Event


lyle:
#### Nature of issue?

- [X] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [X] Dom

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: v0.7.3 January 20, 2019
- p5.dom.js version: v0.4.0
- Web browser and version: Chrome 72.0.3626.121
- Operating System: MacOSX 10.14.3 (18D109)
- Steps to reproduce this:
In this version ( v0.7.3/v0.4.0) using element.drop(fileCallback, eventCallback) the eventCallback never fires.
I created a codepen that shows the bug: https://codepen.io/lyle/pen/aMypPY
This bug does not emerge when using p5.js v0.5.3 / p5.dom.js v0.2.12

lyle:
It looks like this bug was introduced with PR #3303 - the newer p5.Element._attachListener method ensure's there can not be two event types attached at the same time. Considering .drop has two call backs passed in that bind to the same event, we could probably just wrap those two together into one closure that has two callbacks fired.
I am working on . test for the drop method that covers this case.

outofambit:
Thanks @lyle! That implementation sounds reasonable to me! ✨

PR #3605, opened 2019-03-14
Dom Drop's second callback should be called with the raw drop event

lyle:
The dom library has a Element.drop() which is documented to take two
callbacks, the first should be fired for every file dropped, the second
should fire with the raw dom event. The second event was not being
called. I have fixed this bug, improved documentation, and wrote
a mocha test to cover this issue.

Resolves: #3591
See also: #3303

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -2074,8 +2074,8 @@
    * whereas the first callback will trigger for each loaded file.
    *
    * @method drop
-   * @param  {Function} callback  callback to receive loaded file.
-   * @param  {Function} [fxn]     callback triggered when files are dropped.
+   * @param  {Function} callback  callback to receive loaded file, called for each file dropped.
+   * @param  {Function} [fxn]     callback triggered once when files are dropped with the drop event.
    * @chainable
    * @example
    * <div><code>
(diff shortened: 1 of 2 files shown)


lyle:
I have to say that my test code looks bulky to my eye, mocking a Dom drop event seemed to take more code the I expected. But this does test the case of having two callbacks, and counts the amount of times correctly. It is tempting to write a utility for the mocha tests, but I did not find examples in the code base to do so.

lyle:
@outofambit I would appreciate your feedback, esp. around the mocha test.

lyle:
I’m off on vacation right now (sans laptop) so let’s go with this change. As I finish my project in a couple weeks I’ll keep this helper idea in mind.
If I run into any other issues I will start the fix by creating a dom mock helper. 
So please do merge this change. 
What is the process for changing public interfaces, btw? I was thinking of adding the raw event as a second par to the first callback. 
I can write that up someplace and do an implementation.

outofambit:
> I’m off on vacation right now (sans laptop) so let’s go with this change. As I finish my project in a couple weeks I’ll keep this helper idea in mind.
> If I run into any other issues I will start the fix by creating a dom mock helper.
> So please do merge this change.

sounds good! thanks again and have a good vacation!

> What is the process for changing public interfaces, btw? I was thinking of adding the raw event as a second par to the first callback.
> I can write that up someplace and do an implementation.

@lyle yeah i would just start by opening an issue in this repo to propose the change and see what people think before jumping into code. 😃

PR #3609, opened 2019-03-15
p5.dom.js: add hasClass and toggleClass functions

OleksiiBulba:
I added new methods hasClass and toggleClass; added check to method addClass if element already has class; added some tests for methods addClass, removeClass, hasClass and toggleClass.


lib/addons/p5.dom.js
```diff
@@ -1454,6 +1455,68 @@
...
+   *   if (div.hasClass('show')) {
+   *       div.addClass('show');
+   *   } else {
+   *       div.removeClass('show');
+   *   }
+   * }
+   * </code></div>
+   */
+  p5.Element.prototype.hasClass = function(c) {
+    var regex = new RegExp('(?:[a-zA-Z-_\d:]\s|^)' + c + '(?:\s[a-zA-Z-_\d:]|$)');
```
limzykenneth: The tests doesn't seem to like this regex, you should be able to run the prettier step to automatically fix this

lib/addons/p5.dom.js
```diff
@@ -1411,11 +1411,12 @@
    */
   p5.Element.prototype.addClass = function(c) {
     if (this.elt.className) {
-      // PEND don't add class more than once
+      // PEND don't add class more than once*
```
limzykenneth: Why is the extra character added here at the end?

lib/addons/p5.dom.js
OleksiiBulba: I saw that but I tested my regex on https://regexr.com/ -- it is absolutly fine. But suggested ESlint prettier code does not work on that sandbox. So I will dig into it now.

lib/addons/p5.dom.js
OleksiiBulba: Don't mind, just mistype when checked my code. It will be removed. @limzykenneth should I remove comment as well? I think it is reasonable because this comment about my changes.

lib/addons/p5.dom.js
outofambit: @OleksiiBulba stepping back for a second, why do we need a regex for this? From looking at https://developer.mozilla.org/en-US/docs/Web/API/Element/className, we should be able to parse that into a list of classnames with a simple [String.split](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split), and then use [Array.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) to see if the given class is in the list. does that seem reasonable?

Spongman:
Does `element.classList` fall within p5’s compatibility umbrella?

outofambit:
@Spongman ooo i believe it does, thanks for bringing that up! @OleksiiBulba what do you think about using https://developer.mozilla.org/en-US/docs/Web/API/Element/classList?

lib/addons/p5.dom.js
limzykenneth: The few lines of comments here, including the commented out code seems a bit messy and maybe not in consideration anymore if your PR is merged? If so I think it's fine to remove them and if your code need some explanation for future developers then do add some comments in there.

lib/addons/p5.dom.js
OleksiiBulba: @outofambit Can we use method **contains** (**add**, **remove**, **toggle** etc.) from https://developer.mozilla.org/en-US/docs/Web/API/Element/classList ?

OleksiiBulba:
@limzykenneth @outofambit I think checks failed not because of my commits but something else in the code?

lib/addons/p5.dom.js
outofambit: @OleksiiBulba yes! that sounds great :D

lib/addons/p5.dom.js
```diff
@@ -1454,6 +1452,67 @@
...
+  /**
+   *
+   * Checks if specified class already set to element
+   *
+   * @method hasClass
+   * @returns {boolean} a boolean value if element has specified class
+   * @param c {String} class name of class to check
+   * @example
+   * <div class='norender'><code>
+   *
```
meiamsome: Delete this line

lib/addons/p5.dom.js
```diff
@@ -1454,6 +1452,67 @@
...
+   * var div;
+   *
+   * function setup() {
+   *   div = createDiv('div');
+   *   div.addClass('show');
+   * }
+   *
+   * function mousePressed() {
+   *   if (div.hasClass('show')) {
+   *       div.addClass('show');
```
meiamsome: Two less spaces on the indentation here (after the asterisk)

lib/addons/p5.dom.js
```diff
@@ -1454,6 +1452,67 @@
...
+   *   } else {
+   *       div.removeClass('show');
```
meiamsome: Two less spaces on the indentation here (after the asterisk)

lib/addons/p5.dom.js
```diff
@@ -1454,6 +1452,67 @@
...
+  /**
+   *
+   * Toggles element class
+   *
+   * @method toggleClass
+   * @param c {String} class name to toggle
+   * @chainable
+   * @example
+   * <div class='norender'><code>
+   *
```
meiamsome: Remove this blank line

Spongman:
in addition, can we change `removeClass` to something like:
````
this.elt.classList.remove(c);
````

i don't think the regex stuff is necessary

Spongman:
BTW: `classList` _also_ has a `toggle()` method, but i don't think we can use that yet as support is spotty.

OleksiiBulba:
> BTW: `classList` _also_ has a `toggle()` method, but i don't think we can use that yet as support is spotty.

This section https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Browser_compatibility says that almost all browsers (including IE10) have basic support of all methods (toggle method too BUT without the second argument).

lmccart:
thanks @OleksiiBulba!

PR #3613, opened 2019-03-17
Adds a parallel test helper (See #3582)

meiamsome:
As discussed in #3582, this adds a testing helper that should be useful for more complicated tests involving multiple sketches interacting.

### DIFF
--- a/test/js/p5_helpers.js
+++ b/test/js/p5_helpers.js
@@ -22,6 +22,41 @@ function testSketchWithPromise(name, sketch_fn) {
   return test(name, test_fn);
 }
 
+function parallelSketches(sketch_fns) {
+  var setupPromises = [];
+  var resultPromises = [];
+  var endCallbacks = [];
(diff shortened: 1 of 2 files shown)


test/js/p5_helpers.js
```diff
@@ -22,6 +22,41 @@ function testSketchWithPromise(name, sketch_fn) {
   return test(name, test_fn);
 }
 
+function parallelSketches(sketch_fns) {
+  var setupPromises = [];
+  var resultPromises = [];
+  var endCallbacks = [];
+  for (let i = 0; i < sketch_fns.length; i++) {
```
meiamsome: Note that this is `let` to make it clear that the variable is bound inside the loop and won't change in any of the interpreted functions.

The tests have had more leniency on early-adoption of ES6-style things (particularly `async`/`await` for readability reasons) because we control the environment in which the tests are run. I'd rather not have to write closure form here, but am happy to convert if requested.

PR #3621, opened 2019-03-19
renderer has own styles so it can call push/pop itself

### LINKED ISSUE #2946: renderer can't use its own push/pop methods
currently, the p5 instance stacks the styles for renderer and passes an object to p5.Renderer.prototype.pop when p5.prototype.pop is called. this means that the renderer cannot internally use push/pop to maintain its own state; it must have the p5 instance call pop for it.

there are some cases in webgl mode where it would be helpful to let the renderer push/pop it's own state to restore a user texture or shader (for example in image() in #2211). the suggestion here is to let the renderer keep its own style stack instead, and to remove the argument to p5.Renderer.prototype.pop.

I'm planning to implement this soon, let me know if you have any questions. thanks!

stalgiag:
closes #2946 

This moves the style properties to `p5.Renderer`. The only style related settings retained by the `p5.instance` is `_colorMode`. This is based on old design of `p5.instance.styles` holding objects structured like 
```
{
_colorMode: this._colorMode,
 renderer: {all other styles}
}
```
I haven’t worked with `_colorMode` and I assumed that there is some logic for keeping this separate and tried to keep that separation intact as much as possible. Renderers can still call their own push/pop methods since they won't likely be changing the color mode (?). Can move this modification into the Renderer if it makes sense.

Also wrote some simple unit tests to bring WebGL's testing of push/pop closer to full coverage. I realized after writing them in a very verbose fashion that it probably just makes more sense to write them in a more condensed fashion as [in the 2D renderer push/pop unit tests.](https://github.com/processing/p5.js/blob/master/test/unit/core/structure.js#L110). That said, my personal opinion is that the `suite.test` structure of unit tests forces an organized readability and that the long-form tests are fine given this format. I am willing to go back and condense them though.

### DIFF
--- a/src/core/main.js
+++ b/src/core/main.js
@@ -562,7 +562,7 @@ var p5 = function(sketch, node, sync) {
 };
 
 p5.prototype._initializeInstanceVariables = function() {
-  this._styles = [];
+  this._colorModes = [];
 
   this._bezierDetail = 20;
   this._curveDetail = 20;
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -38,6 +38,8 @@ p5.Renderer = function(elt, pInst, isMainCanvas) {
     this._styles = []; // non-main elt styles stored in p5.Renderer
   }
 
+  this.styles = [];
+
   this._textSize = 12;
   this._textLeading = 15;
   this._textFont = 'sans-serif';
@@ -87,7 +89,17 @@ p5.Renderer.prototype.push = function() {
 // a pop() operation is in progress
 // the renderer is passed the 'style' object that it returned
 // from its push() method.
-p5.Renderer.prototype.pop = function(style) {
+p5.Renderer.prototype.pop = function() {
+  var style = this.styles.pop();
+  if (typeof style === 'undefined') {
+    var errorMessage = 'You called pop() without first calling push()!';
+    if (this.isP3D) {
+      //prettier-ignore
+      errorMessage += ' Note that in 3D mode calling setAttributes resets push().';
+    }
+    console.error(errorMessage);
+    return;
+  }
   if (style.properties) {
     // copy the style properties back into the renderer
     Object.assign(this, style.properties);
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -92,12 +92,7 @@ p5.Renderer.prototype.push = function() {
 p5.Renderer.prototype.pop = function() {
   var style = this.styles.pop();
   if (typeof style === 'undefined') {
-    var errorMessage = 'You called pop() without first calling push()!';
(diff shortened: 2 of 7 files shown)


Spongman:
I’m not sure I understand the reasoning for this change. I originally unified the style properties in the single style stack in orde to 1) reduce the amount of allocation during push/pop call and 2) eliminate the chance that the multiple style stacks got out of sync. This seems to undo that, and I’m not quite sure what else it does.

stalgiag:
Hm I was following what I believed to be the todo from the issue listed above. There is a possibility that I misunderstood the issue.

Assuming I didn't misunderstand, it seems to me that there are pros and cons with both the unified p5 level style stacks and the renderer level style stacks. The allocation differences are seem minimal to nonexist. As far as the stacks going out of sync, the webgl style stacks seems to be a supplement to the renderer/2drenderer stack and as written they are maintained in sync. That said, I do think the unified properties approach has some advantages such as being slightly easier to update, and easier maintenance of the styles across resetting/switching of rendering contexts.

Spongman:
i think part of the problem in that issue was that, since the webgl state was stored in the shader attributes directly, pushing and popping the settings in the renderer had no effect (since `pop()` didn't re-apply the values to the shaders). however, since #3535, the state isn't stored in the shader any longer, it's stored in the renderer properties (and updated before the draw call). this _should_ mean that push/pop work somewhat better than before.

stalgiag:
That makes sense. I didn't understand that you had unified this prior and you are right that #3535 makes push/pop work better. If the goal is to make the internal use of push/pop simple so that `image()` for example could be nearly as concise as 
```
p5.RendererGL.prototype.image = function(texture, x, y, w, h) {
  this._pInst.push();
  this._pInst.texture(texture);
  this._pInst.translate(x, y, 0);
  this._pInst.quad(x, y, x + w, y, x + w, y + h, x, y + h);
  this._pInst.pop();
};
```
Then both this branch and the current master are equally successful in this regard (texture gets out of sync with push()/pop() somehow in both). I will give @kjhollen some time to respond and explain a bit more why the separated style stack was requested but if you are right, and #3535 handled most of this issue then I will just cherry pick the unit tests and documentation updates from this PR and work on figuring out why the shader attributes fall out of sync in the draw loop with the above version of `image()` .

stalgiag:
I think it makes sense to close this PR and the issue that it worked to solve. Once #3646 is merged then RendererGL will have a fully functioning `push` and `pop` that it can call itself. The unified style stacks seems a bit easier to maintain and since the desired functionality is already working I don't think it makes sense to split them back out again. If you disagree @kjhollen then I'll happily reopen.

As an afterthought, probably no use in throwing out these unit tests and documentation updates as they are still relevant. When I get a chance I'll cherrypick and open a different PR.

Issue #3622, opened 2019-03-19
setAttributes() resets buffers


stalgiag:
#### Nature of issue?

- [x ] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ x] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x ] Desktop/Laptop
- [ ] Others (specify if possible)

There is currently an edge case in the `master` branch in which `setAttributes()` will silently mess up rendering in WebGL if used after drawing inside of setup. Here is an example sketch:
```
function setup() {
  createCanvas(windowWidth, windowHeight, WEBGL);
  fill(150, 0, 150);
  background(0, 200, 0);
  sphere(100);
  setAttributes('antialias', true);
}
```
If `setAttributes()` is moved anywhere above `sphere()` and `background()` all is well. Likewise, if the code is in `draw()`

This is because when `setAttributes()` is called with settings that are different than the current settings, the rendering context is reset and a new p5.RendererGL object is made. The style and transform properties that are transferred with `push()` and `pop()` carry over during this switch but drawn geometry does not. Buffers can’t be transferred from one context to another. It is possible to rebind based on copy but as far as I can tell that is difficult to do with current architecture.

This only matters when `setAttributes()` is used after drawing inside of `setup()`. If `setAttributes()` is called before drawing any geometry then the drawing will happen in the new renderer. if `setAttributes()` is used in draw then there may be one corrupt frame but it is hidden by the reset of the drawing context anyways and all subsequent drawing happens in new renderer.

This is relegated to very specific uses so I think it is fine to just make this a warning. The todo for now would probably just be to log an error to the console and return without doing anything [inside of `setAttributes()`](https://github.com/processing/p5.js/blob/master/src/webgl/p5.RendererGL.js#L394) when `this._renderer.gHash` is not an empty object `&& !this._setupDone`. Open to other suggestions as well!

stalgiag:
closing this based on #3624. Since this is an edge case a warning should be sufficient.

PR #3624, opened 2019-03-20
Added an error for setattributes on buffers

sanketsingh24:
Partially fixes - #3622 

Added an error for the same.

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -392,15 +392,16 @@ p5.prototype.setAttributes = function(key, value) {
     return;
   }
 
-  if (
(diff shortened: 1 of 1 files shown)


sanketsingh24:
Ok, I quickly looked into it, and made a [test](https://jsperf.com/tester416564199), for me 

`for (var x in obj) { return false; }` was the fastest on Chrome 72.0.3626 and Firefox, but the current implementation is also very close.

stalgiag:
This looks good. I don't think that this is too terribly performance critical because this code will only be run when in setup and actually changing an attribute (conditionals prior to this handle redundant calls). Adding a very small check before a large and rare operation (resetting context) seems acceptable.

PR #3630, opened 2019-03-20
Update DOM Reference Examples to ES6

### LINKED ISSUE #3379: Update DOM Reference Examples on p5js.org to ES6
Update the **DOM** section of the [p5js reference](https://p5js.org/reference/) to ES6. This will entail updating all of the references in source files that have `@module DOM` at the top. In this probably just the [p5.Element class](https://github.com/processing/p5.js/blob/master/src/core/p5.Element.js)

ihsavru:
fixes #3379


lmccart:
thanks @ihsavru!

PR #3633, opened 2019-03-22
Updated ES5 to ES6 in utils/sample-linter.js

abhinavsagar:


Spongman:
please try this again. You can’t just blindly change function definitions to lambda expressions there’s a bunch of stuff here you broke by doing this.

PR #3634, opened 2019-03-22
Add image() for webgl mode

### LINKED ISSUE #2182: Implement image() for webgl mode

vedhant:
fixes #2182

### DIFF
--- a/src/webgl/3d_primitives.js
+++ b/src/webgl/3d_primitives.js
@@ -1532,4 +1532,50 @@ p5.RendererGL.prototype.curveVertex = function() {
   }
 };
 
+p5.RendererGL.prototype.image = function(
+  img,
+  sx,
+  sy,
+  sWidth,
+  sHeight,
+  dx,
+  dy,
+  dWidth,
+  dHeight
+) {
+  this._pInst.texture(img);
+  this._pInst.textureMode(constants.NORMAL);
+
+  // (dx, dy) maps to center of canvas, so it is transformed to the actual (0,0) of canvas
+  var new_dx = dx - this._pInst.width / 2;
+  var new_dy = dy - this._pInst.height / 2;
+
+  var u0 = 0;
+  if (sx <= img.width) {
+    u0 = sx / img.width;
+  }
+
+  var u1 = 1;
+  if (sx + sWidth <= img.width) {
+    u1 = (sx + sWidth) / img.width;
+  }
+
+  var v0 = 0;
+  if (sy <= img.height) {
+    v0 = sy / img.height;
+  }
+
+  var v1 = 1;
+  if (sy + sHeight <= img.height) {
+    v1 = (sy + sHeight) / img.height;
+  }
+
+  this.beginShape();
+  this.vertex(new_dx, new_dy, 0, u0, v0);
+  this.vertex(new_dx + dWidth, new_dy, 0, u1, v0);
+  this.vertex(new_dx + dWidth, new_dy + dHeight, 0, u1, v1);
+  this.vertex(new_dx, new_dy + dHeight, 0, u0, v1);
+  this.endShape(constants.CLOSE);
+};
+
 module.exports = p5;
--- a/src/webgl/3d_primitives.js
+++ b/src/webgl/3d_primitives.js
@@ -1543,6 +1543,8 @@ p5.RendererGL.prototype.image = function(
   dWidth,
   dHeight
 ) {
+  this._pInst.push();
+
   this._pInst.texture(img);
   this._pInst.textureMode(constants.NORMAL);
 
@@ -1576,6 +1578,8 @@ p5.RendererGL.prototype.image = function(
   this.vertex(new_dx + dWidth, new_dy + dHeight, 0, u1, v1);
   this.vertex(new_dx, new_dy + dHeight, 0, u0, v1);
   this.endShape(constants.CLOSE);
+
+  this._pInst.pop();
 };
 
 module.exports = p5;
--- a/src/webgl/3d_primitives.js
+++ b/src/webgl/3d_primitives.js
@@ -1548,10 +1548,6 @@ p5.RendererGL.prototype.image = function(
   this._pInst.texture(img);
   this._pInst.textureMode(constants.NORMAL);
 
-  // (dx, dy) maps to center of canvas, so it is transformed to the actual (0,0) of canvas
-  var new_dx = dx - this._pInst.width / 2;
-  var new_dy = dy - this._pInst.height / 2;
-
   var u0 = 0;
   if (sx <= img.width) {
     u0 = sx / img.width;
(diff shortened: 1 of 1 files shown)


Spongman:
If you’re going to change instance properties, I think you need to make sure they’re reset correctly after you’re done.

stalgiag:
@Spongman's comment brings up the point that all of this should likely take place inside of push/pop to get as close as possible to the 2D mode use for `image`. There may be some unexpected results from using push and pop with drawing currently (see discussion in #3621) but I do think that the best version of this would not effect the state of the rest of the drawing ie 
```
function draw() {
  texture(tex1);
  image(tex2, 0, 0, 100, 100);
  box(50); // should be textured with tex1
}
```
I think this is a great start but the review may take several days because the requested changes may differ depending on what is happening in #3621. That said, requested changes for your PR are unlikely to be major. Thanks for the PR!

vedhant:
@spongman yup you are right. Ill make the changes in a couple of days

src/webgl/3d_primitives.js
```diff
@@ -1532,4 +1532,54 @@ p5.RendererGL.prototype.curveVertex = function() {
...
+  dWidth,
+  dHeight
+) {
+  this._pInst.push();
+
+  this._pInst.texture(img);
+  this._pInst.textureMode(constants.NORMAL);
+
+  // (dx, dy) maps to center of canvas, so it is transformed to the actual (0,0) of canvas
+  var new_dx = dx - this._pInst.width / 2;
```
stalgiag: I don't know about this decision. I understand why you would do it. It may be preferable in some cases to have `image()` draw in 2D and 3D in the same way but I think this is a little misleading for new users. The coordinate system for everything else that is drawn in 3D has an origin at screen center. I think this should stay consistent in 3D rather than staying consistent for this particular function. The user would likely expect the top-left corner of the image to be in the center of the sphere in a sketch like this:
```
function draw() {
  background(255);
  push();
  translate(50, 0, 0);
  sphere(50);
  pop();
  image(grid, 50, 0, 200, 200);
}
```
This is definitely open for debate but this is my instinct on this one.

src/webgl/3d_primitives.js
```diff
@@ -1532,4 +1532,54 @@ p5.RendererGL.prototype.curveVertex = function() {
   }
 };
 
+p5.RendererGL.prototype.image = function(
+  img,
+  sx,
+  sy,
```
stalgiag: I think this is a good start and a flat depth image rendering seems fine for now but might be worth considering if there should be an optional z-argument.

stalgiag: If you don't feel like doing this now this could be a different issue/pull request.

src/webgl/3d_primitives.js
```diff
@@ -1532,4 +1532,54 @@ p5.RendererGL.prototype.curveVertex = function() {
...
+  sWidth,
+  sHeight,
+  dx,
+  dy,
+  dWidth,
+  dHeight
+) {
+  this._pInst.push();
+
```
stalgiag: I think you should either take [the current `imageMode` into account ](https://p5js.org/reference/#/p5/imageMode) or put a [warning in here](https://github.com/processing/p5.js/blob/master/src/image/loading_displaying.js#L529) when the renderer is 3D.

src/webgl/3d_primitives.js
vedhant: The z-argument will only define depth to the image right?

stalgiag: Right so instead of passing 0 as the third argument every time that you call `this.vertex()` you would pass a z-value which determines depth.

vedhant: The signature of ```image()``` in WebGL mode will have an extra argument ```dz``` compared to 2D mode.
When ```image()``` is called by user, first the ```image()``` in loading_displaying.js is called which in turn calls the ```image()``` corresponding to the appropriate renderer. But now that the signature has changed, this will no longer be possible unless the argument ```dz``` is at end.

stalgiag: That could be solved by checking whether the renderer is 3D and checking the argument length in `p5.prototype.image` in loading_displaying but again this could be a future addition to the WebGL image. It is fine to default to 0 depth for now.

src/webgl/3d_primitives.js
```diff
@@ -1532,4 +1532,50 @@ p5.RendererGL.prototype.curveVertex = function() {
...
+  this.beginShape();
+  this.vertex(dx, dy, 0, u0, v0);
+  this.vertex(dx + dWidth, dy, 0, u1, v0);
+  this.vertex(dx + dWidth, dy + dHeight, 0, u1, v1);
+  this.vertex(dx, dy + dHeight, 0, u0, v1);
+  this.endShape(constants.CLOSE);
+
+  this._pInst.pop();
+};
+
```
stalgiag: I think that unit tests should be added for this new method. In can be tricky to catch lack of unit test coverage for new functions down the road. If you don't have the bandwidth for this then an issue should be opened to track that these unit tests don't exist yet. So one of two options:
- add a new simple unit test suite to the [RendererGL](https://github.com/processing/p5.js/blob/master/test/unit/webgl/p5.RendererGL.js) tests. Inside you can have a simple test that [preloads an existing small image asset like this](https://github.com/processing/p5.js/blob/master/test/unit/webgl/p5.Texture.js#L13). Then you should call your new image function with `myp5.image(args)` and use get pixel to make sure that the image is applied as expected. You can test the result of (get(x,y)) in a separate sketch so you know what to assert is the return value of that function when the image is applied properly. If you decide to go this route and need more help feel free to let me know.
- Alternatively, you can open a new issue explaining what needs to happen with adding the unit test and someone else can take care of this. Either option is fine just don't want to lose track of this as a to-do.

src/webgl/3d_primitives.js
stalgiag: I still think there should be a warning [in imageMode](https://github.com/processing/p5.js/blob/master/src/image/loading_displaying.js#L529) when `renderer.isP3D`.

stalgiag:
Hi @vedhant, just a heads up that those last change requests aren't super urgent. There is going to be a release tomorrow and it would be great to get this PR in as part of the release. I will go ahead and merge this as-is tomorrow morning (PST) and if there are any minor changes that still need to happen I'll open issues for them. Thanks!

Issue #3635, opened 2019-03-22
Improve examples of loadPixels


Ajayneethikannan:
#### Nature of issue?

- [x] Found a bug

#### Most appropriate sub-area of p5.js?

- [x] Core/Environment/Rendering
- [x] Image

#### Which platform were you using when you encountered this?

- [x] Others (specify if possible) Device independent

#### Details about the bug: 

- p5.js version:  0.7.3
- Web browser and version:  Browser independent
- Operating System:  OS independent
- Steps to reproduce this:

The reference of loadPixels() of p5.Image and loadPixels() of main canvas can be checked for this. 
-https://p5js.org/reference/#/p5.Image/loadPixels
-https://p5js.org/reference/#/p5/loadPixels

Issues: 
- loadPixels example for main canvas attempts to show how we can copy the pixels of top half of the main canvas, to the bottom half. So in the pixels array, to find the middle index, 
we need to use halfImage = 4 * (d* width) * (d*height/2); 
but instead the img's width and height are used , which seems to be of the same width as the main canvas
- The reverse is done in the loadPixels() example of p5.Image

- The examples used are the same for updatePixels

Open to suggestions,
Thanks!

PR #3637, opened 2019-03-23
improve examples of loadPixels, updatePixels

Ajayneethikannan:
Fixes : #3635

- Changing the calculation logic for halfImage index in p5.Image.prototype.loadPixels and p5.prototype.loadPixels, according to the example . ( loadPixels for image should calculate halfImage , with respect to the image instance, and for main canvas, calculate with respect to the canvas window, taking pixelDensity into consideration )
- Drawing the image, so that it fits the full canvas, so that when an user checks out the example using other images, the code would run as expected for bigger and smaller images

I request the maintainers to review the PR,
Thanks !

### DIFF
--- a/src/image/pixels.js
+++ b/src/image/pixels.js
@@ -519,9 +519,9 @@ p5.prototype.get = function(x, y, w, h) {
  * }
  *
  * function setup() {
- *   image(img, 0, 0);
+ *   image(img, 0, 0, width, height);
  *   let d = pixelDensity();
- *   let halfImage = 4 * (img.width * d) * (img.height * d / 2);
+ *   let halfImage = 4 * (width * d) * (height * d / 2);
  *   loadPixels();
  *   for (let i = 0; i < halfImage; i++) {
  *     pixels[i + halfImage] = pixels[i];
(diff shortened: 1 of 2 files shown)


src/image/pixels.js
```diff
@@ -519,9 +519,9 @@ p5.prototype.get = function(x, y, w, h) {
  * }
  *
  * function setup() {
- *   image(img, 0, 0);
+ *   image(img, 0, 0, width, height);
  *   let d = pixelDensity();
- *   let halfImage = 4 * (img.width * d) * (img.height * d / 2);
+ *   let halfImage = 4 * (width * d) * (height * d / 2);
```
lmccart: I think it's good to leave img.width and img.height here as it's a little more clear to understand how this is half the image.

src/image/pixels.js
```diff
@@ -638,9 +638,9 @@ p5.prototype.set = function(x, y, imgOrCol) {
...
```
lmccart: I think it's good to leave img.width and img.height here as it's a little more clear to understand how this is half the image.

Ajayneethikannan: Thank you for suggesting the changes @lmccart ! I have a doubt, since we are explaining how to use
the pixels array of the main drawing canvas to update it, and are also using pixelDensity of the main canvas, 
I was thinking it would confuse the user which halfImage index we are trying to find (image or main drawing area). 

I would make the changes as your suggestions make it look less complex and clear to the user.
Please provide your thoughts on this, (sorry if I am wrong)
Thank you !

lmccart: @Ajayneethikannan ah I see, yes this is fine to leave as is since we are talking about half canvas image. thanks for explaiing.

PR #3642, opened 2019-03-24
Add support for tdd style in node unit tests

### LINKED ISSUE #3641: support suite/test syntax in node tests
We currently support `suite`/`test` syntax in browser tests, but not in node tests. This should be consistent so the test syntax is consistent.

I looks like this is because we have the line `mocha.setup('tdd');` in our test `html` files but not an equivalent line in our node test setup.

outofambit:
closes #3641 

also centralizes some setup for browser-based tests

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -244,7 +244,8 @@ module.exports = grunt => {
       test: {
         src: ['test/node/**/*.js'],
         options: {
-          reporter: reporter
+          reporter: reporter,
+          ui: 'tdd'
         }
       }
     },
--- a/package-lock.json
+++ b/package-lock.json
@@ -397,6 +397,12 @@
         }
       }
     },
+    "ansi-colors": {
+      "version": "3.2.3",
+      "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-3.2.3.tgz",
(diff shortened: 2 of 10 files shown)


outofambit:
@meiamsome would you mind reviewing this?

meiamsome:
@outofambit Testing locally this doesn't seem to make mocha works like you expect it to in node - running `npm run test` yields `ReferenceError: suite is not defined` in the `mochaTest:test` step if there are `suite` blocks in the node test, and the tests throw `ReferenceError: expect is not defined`  otherwise in the browser tests.

outofambit:
@meiamsome so sorry about that! should be good now. :)

meiamsome:
@outofambit I'm still not sure I follow whether what you're trying to achieve here is working.

As I see the issue, what we're trying to achieve is that you should be able to do add the following:
```javascript
suite('test', function() {
  test('should work', function() {});
});
```
To the bottom of `test/node/helpers.js` and the tests should pass without an issue. If that is the intention, then this doesn't seem to be the case - for me, when trying locally, I see this as throwing the `ReferenceError: suite is not defined` error still. This is when I am running via grunt, but if we're supporting this syntax in the tests, they _must_ work in grunt, otherwise it will break Travis and (I suspect) most people's workflow.
Please let me know if I have misunderstood.

The Chai errors do seem to be gone now, though!

outofambit:
@meiamsome sorry for the back and forth on this and thank you for articulating that test case you were using. i went ahead and ported the existing tests to `tdd` style so CI would fail if the config was not right and i _think_ this is ready, but please let me know if you see anything out of place. 🙇‍♀

PR #3643, opened 2019-03-25
[WIP] Remove redundant call to image function in p5.prototype.background

### LINKED ISSUE #3639: Doubt: p5.prototype.background function in settings.js
#### Nature of issue?

- [x] Found a bug

- [x] Image

- [x] Desktop/Laptop
- [x] Others (specify if possible) Device independent

#### Details about the bug: 

- p5.js version:  0.7.3
- Web browser and version:  Browser independent
- Operating System:  OS independent
- Steps to reproduce this:
This code can be used for reference : 
```
let img;
function preload() {

  img = loadImage('https://i.imgur.com/VyAlcbq.jpg');
}
function setup()
{
    createCanvas(100, 100);
}
function draw() {
  rotate(5);
  background(img); //background translated leaving white space on the canvas.

  //image(img, 0, 0, this.width, this.height);
}

```

If an image is used to create background while transformations are applied before it ,
it can result in the background being translated too, instead of the background filling the screen.

This happens because in [p5.prototype.background](https://github.com/processing/p5.js/blob/master/src/color/setting.js#L176), the image function is directly called without resetting the matrix, which is handled in [p5.Renderer2D.prototype.background](https://github.com/processing/p5.js/blob/master/src/core/p5.Renderer2D.js#L46).

Also , there is a kind of duplicacy of the code , if only one of the methods are ever called. 

Should the implementation be changed that  only p5.Renderer2D.background is called , in case of image being passed as a parameter ? 
Thanks !

Ajayneethikannan:
Fixes #3639 

- Remove calling image function in p5.prototype.background, as it's logic is already handled inside p5.Renderer2D.prototype.background.

As suggested by @lmccart , I will wait for the image functionality to be implemented in WEBGL, after which changes will be made if necessary.

Thank you !

### DIFF
--- a/src/color/setting.js
+++ b/src/color/setting.js
@@ -174,11 +174,7 @@ require('./p5.Color');
  */
 
 p5.prototype.background = function() {
-  if (arguments[0] instanceof p5.Image) {
-    this.image(arguments[0], 0, 0, this.width, this.height);
-  } else {
-    this._renderer.background.apply(this._renderer, arguments);
-  }
+  this._renderer.background.apply(this._renderer, arguments);
   return this;
 };
 


lmccart:
Looks good! Will merge as soon as #3634 is merged.

PR #3644, opened 2019-03-25
setup babelify

### LINKED ISSUE #3615: experiment with babel-preset-env
I'd like to explore using `babel-preset-env` to allow us to consistently support a range of browsers while using whatever ES features we'd like.

I'm not sure how this will impact the bundle size, so I'm going to look into that first.

(builds on #3431)

-----

> There's a lot of other things we could have `babel` do, but I think the strongest case for `babel` in p5js is to ensure backward compatibility with older browser versions (or ones that are the minority in not supporting an ES syntax/feature) using [`babel-preset-env`](https://babeljs.io/docs/en/babel-preset-env).
>
> Supporting a [range of browsers](https://github.com/processing/p5.js/blob/master/developer_docs/supported_browsers.md#our-stated-goal) is much easier with [`babel-preset-env`](https://babeljs.io/docs/en/babel-preset-env). When added to your build process, it will polyfill and transpile your javascript to work on [the browser versions listed in a configuration file](https://babeljs.io/docs/en/babel-preset-env#targets) (which we maintain as part of our repo). It's super cool and would allow us to embrace whatever features of es6+ we find appropriate without limiting the library's (browser) reach.

_Originally posted by @outofambit in https://github.com/processing/p5.js/issues/3425#issuecomment-450445958_

outofambit:
closes #3615 

(we don't have to use `parcel` to use `babel-preset-env`)

[diff between current build output and new build output with this PR for `lib/p5.js`](https://gist.github.com/outofambit/5e66fd8126878328b717197302c6894b). The diff is huge, but it looks like the primary changes are comment formatting (which is good!). This confirms that we're using a compatible set of JS language features and we can add as we go.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,275 @@
         "@babel/highlight": "^7.0.0"
       }
     },
+    "@babel/core": {
+      "version": "7.4.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.4.0.tgz",
+      "integrity": "sha512-Dzl7U0/T69DFOTwqz/FJdnOSWS57NpjNfCwMKHABr589Lg8uX1RrlBIJ7L5Dubt/xkLsx0xH5EBFzlBVes1ayA==",
+      "dev": true,
+      "requires": {
+        "@babel/code-frame": "^7.0.0",
+        "@babel/generator": "^7.4.0",
+        "@babel/helpers": "^7.4.0",
+        "@babel/parser": "^7.4.0",
+        "@babel/template": "^7.4.0",
+        "@babel/traverse": "^7.4.0",
+        "@babel/types": "^7.4.0",
+        "convert-source-map": "^1.1.0",
+        "debug": "^4.1.0",
+        "json5": "^2.1.0",
+        "lodash": "^4.17.11",
+        "resolve": "^1.3.2",
+        "semver": "^5.4.1",
+        "source-map": "^0.5.0"
+      },
+      "dependencies": {
+        "debug": {
+          "version": "4.1.1",
+          "resolved": "https://registry.npmjs.org/debug/-/debug-4.1.1.tgz",
+          "integrity": "sha512-pYAIzeRo8J6KPEaJ0VWOh5Pzkbw/RetuzehGM7QRRX5he4fPHx2rdKMB256ehJCkX+XRQm16eZLqLNS8RSZXZw==",
+          "dev": true,
+          "requires": {
+            "ms": "^2.1.1"
+          }
+        },
+        "ms": {
+          "version": "2.1.1",
+          "resolved": "https://registry.npmjs.org/ms/-/ms-2.1.1.tgz",
+          "integrity": "sha512-tgp+dl5cGk28utYktBsrFqA7HKgrhgPsg6Z/EfhWI4gl1Hwq8B/GmY/0oXZ6nF8hDVesS/FpnYaD/kOWhYQvyg==",
+          "dev": true
+        },
+        "source-map": {
+          "version": "0.5.7",
+          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.7.tgz",
+          "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w=",
+          "dev": true
+        }
+      }
+    },
+    "@babel/generator": {
+      "version": "7.4.0",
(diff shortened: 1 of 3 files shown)


outofambit:
@limzykenneth and @Spongman would you take a look when you have time? thank you!

Spongman:
Do you thjnk it’s possible to make the formatting more closely match what the prettifier currently uses? (Eg remove spaces after function names, etc...)

outofambit:
Looks like babel doesn't support that (https://github.com/babel/babel/issues/189, https://github.com/babel/babel/issues/497), but I can run it through `prettier` after `browserify` is done. I'll add that.

outofambit:
actually there are [a few basic options](https://babeljs.io/docs/en/options#code-generator-options) so i'll try that first and see how close that gets us

limzykenneth:
I'll check back in a couple days, busy setting up a show at the moment.

outofambit:
(i tried running this through prettier (via eslint) after the build was complete, but the file may be too big? it just hung and didn't complete even after a few minutes. i'm going to vote we punt on trying to tighten formatting of the build output to later.

Spongman:
i'm seeing minification in `p5.js` :

````javascript
/**
 * @module Color
 * @submodule Creating & Reading
 * @for p5
 * @requires core
 * @requires constants
 */'use strict';var p5=_dereq_('../core/main');var constants=_dereq_('../core/constants');_dereq_('./p5.Color');_dereq_('../core/error_helpers');/**
````

outofambit:
hmm @Spongman looks like that was because of the `compact` setting i tried for `babel` to tighten up some of the formatting. should be good now.

Spongman:
how about this?

````patch
diff --git a/tasks/build/browserify.js b/tasks/build/browserify.js
index d89ad2c3..8504da23 100644
--- a/tasks/build/browserify.js
+++ b/tasks/build/browserify.js
@@ -3,6 +3,8 @@
 const path = require('path');
 const browserify = require('browserify');
 const derequire = require('derequire');
+const prettier = require('prettier');
+

 const bannerTemplate =
   '/*! p5.js v<%= pkg.version %> <%= grunt.template.today("mmmm dd, yyyy") %> */';
@@ -35,11 +37,23 @@ module.exports = function(grunt) {
         browseified = browseified.exclude('../../docs/reference/data.json');
       }

-      const bundle = browseified
+      let bundle = browseified
         .transform('brfs')
-        .transform('babelify')
+        .transform('babelify', {
+          "presets": [[
+            "@babel/preset-env", {
+              "exclude": ["transform-typeof-symbol"]
+            }
+          ]],
+          generatorOpts: {
+            //compact: isMin,
+            //minified: isMin,
+            retainLines: !isMin
+          }
+        })
         .bundle();

+
       // Start the generated output with the banner comment,
       let code = banner + '\n';

@@ -49,6 +63,15 @@ module.exports = function(grunt) {
           code += data;
         })
         .on('end', function() {
+
+          if (!isMin) {
+            code = prettier.format(code, {
+              printWidth: 80 + 12,
+              singleQuote: true,
+            })
+            .replace(/^          /gm, '');
+          }
+
           // "code" is complete: create the distributable UMD build by running
           // the bundle through derequire, then write the bundle to disk.
           // (Derequire changes the bundle's internal "require" function to
````

outofambit:
@Spongman how do you feel about https://github.com/processing/p5.js/pull/3644/commits/f368cd7b53f678cc99a09f2cf8886d4840c59591? I'd like to keep the babel config out of the grunt task itself, but i think this gets us to about the same place.

Spongman:
yeah, it looks good. 

the other changes (the babelify options, prettier `printWidth`, and whitespace prefix trimming) were an attempt to make the output match the current `p5.js` file. if you look at the diffs with/without those changes, you'll see that the output _with_ those changes much more closely matches the original source code. the diffs are ~ 4MB vs 7MB with most of that difference being in the `data.json` & 3rd-party code. IMO i think it's useful to have the `p5.js` file look reasonably close to the github sources just to help with contributing to the library.

outofambit:
thanks @Spongman! i've made a couple more tweaks but i'm not too worried about the distribution and the source diverging that much. i don't think contributors (especially new ones) look at the built source very often. it's also going to get more pronounced over time as we use more es6 in the source code, and i'm not sure there's much we can do about that.

i'm planning to merge this tomorrow unless additional people would like to review. let me know if that's you! :)

PR #3646, opened 2019-03-25
Change timing for texture binding

### LINKED ISSUE #3636: Multiple textures applied in incorrect order when looping
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- Web browser and version: Chrome and Firefox

So I was following a bug that I thought was related to `push` and `pop` in WebGL mode (#3621 ) but I am realizing that it is a little more widespread. Essentially, the `_tex` value held by RendererGL gets out of sync in the draw loop. In this example:
```
let grid, cat;

function preload() {
  grid = loadImage('../../assets/UV_Grid_Sm.jpg');
  cat = loadImage('../../assets/cat.jpg');
}

function setup() {
  // works as expected with no loop
  // noLoop();
}

// GRID should be on right
// CAT should be on left
function draw() {
  background(150, 0, 150);
  texture(cat);
  box(100);
  translate(100, 0, 0);
  texture(grid);
  box(50);
}
```
I get this result: 
Screenshot (37)

Note that if you uncomment `noLoop()` this works as expected. I have tracked this to the [second time that the draw function is bound](https://github.com/processing/p5.js/blob/master/src/core/main.js#L399). Before this second pass through the draw loop the out put is as expected
Screenshot (38)
Note that in the above screenshot there are still some async operations related to image loading happening.

I will probably continue to dig further into this next week but just wanted to post what I've found here in case anyone has any 'aha' moments or can see where I have a glaring oversight.

stalgiag:
closes #3636 

Okay! I had a lot of trouble tracking this down for some reason. Essentially #3535 introduced a bug that caused all sketches with multiple different textures to apply those textures incorrectly. This escaped testing because with our multiple textures manual test it isn't possible to visually glance and know which object should have which texture. It escaped the unit tests because the unit tests can't test what texture is actually bound to a shader. 

The problem turned out to be that textures were bound when shaders were bound. The shader was binding its texture before the texture uniform had been updated to match the current `_tex` and [an out-of-sync texture was still stored in the uniform](https://github.com/processing/p5.js/blob/master/src/webgl/p5.Shader.js#L219). Just moving the call to `bindTextures` out of `Shader.prototype.bindShader` and into `RendererGL.prototype._setFillUniforms`.

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -1062,6 +1062,7 @@ p5.RendererGL.prototype._setFillUniforms = function(fillShader) {
   var ambientLightCount = this.ambientLightColors.length / 3;
   fillShader.setUniform('uAmbientLightCount', ambientLightCount);
   fillShader.setUniform('uAmbientColor', this.ambientLightColors);
+  fillShader.bindTextures();
 };
 
 p5.RendererGL.prototype._setPointUniforms = function(pointShader) {
--- a/src/webgl/p5.Shader.js
+++ b/src/webgl/p5.Shader.js
@@ -190,7 +190,6 @@ p5.Shader.prototype.bindShader = function() {
   if (!this._bound) {
     this.useProgram();
     this._bound = true;
-    this.bindTextures();
 
     this._setMatrixUniforms();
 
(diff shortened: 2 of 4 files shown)


stalgiag:
Relevant to #3621 and #3624

PR #3647, opened 2019-03-26
Update docs and unit tests to cover WebGL push and pop

stalgiag:
This is just the unit tests and doc updates from the unmerged PR #3621. 

This way there is coverage for most of the WebGL styles with push and pop. Also, the docs now explain which WebGL properties are stored/restored with push & pop.

### DIFF
--- a/test/unit/webgl/p5.RendererGL.js
+++ b/test/unit/webgl/p5.RendererGL.js
@@ -48,6 +48,138 @@ suite('p5.RendererGL', function() {
     });
   });
 
+  suite('push() and pop() work in WEBGL Mode', function() {
+    test('push/pop and translation works as expected in WEBGL Mode', function(done) {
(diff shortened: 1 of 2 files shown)


test/unit/webgl/p5.RendererGL.js
```diff
@@ -48,6 +48,138 @@ suite('p5.RendererGL', function() {
...
+      myp5.push();
+      myp5.shader(fillShader2);
+      assert.equal(fillShader2, myp5._renderer.userFillShader);
+      assert.notEqual(fillShader1, myp5._renderer.userFillShader);
+      myp5.pop();
+      assert.equal(fillShader1, myp5._renderer.userFillShader);
+      done();
+    });
+
+    test('push/pop builds/unbuilds stack properly', function(done) {
```
lmccart: nice :)

Issue #2756, opened 2018-04-03
Documentation of unimplemented 2D vs WebGL


AlexKraken:
#### Nature of issue?

- [ ] Found a bug
- [x ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [x] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [ ] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 
Add documentation of what is and isn't included in the WebGL mode at each release or create a running list within the WebGL wiki page which lists unimplemented functions in WebGL such as blendMode. This would clear up issues for WebGL beginners and provide a single resource to know why these things aren't working. It looks like this has been an issue since 2015: #986
- p5.js version: 
0.6.0
- Web browser and version: 
Safari Version 11.0.2 (13604.4.7.1.3)
- Operating System: 
macOS 10.13.2
- Steps to reproduce this:

```javascript
// must create canvas with WEBGL
blendMode(SCREEN)
fill(0,255,0,100)
ellipse(0,0,100,100)
fill(255,0,0,100)
ellipse(50,50,100,100)
```

stalgiag:
Okay just want to update this with the remaining functions that are current not working with WEBGL mode. Luckily there are only a few now:
`filter()` - needs implementation or at least protection/warning message
`strokeCap()` - implementation will be arduous, has protection and warning message
`strokeJoin()` - implementation will be arduous, needs protection/warning message
`textStyle()` -- this is implied in the documentation since a loaded font is required in GL and the docs say to switch the current font instead of using when `textStyle()` when using a loaded font.

stalgiag:
Okay. I have opened issues for the remaining 2 unprotected, unimplemented WEBGL functions:

filter - #4046 
strokeJoin - #4047 

Almost all functions work in both renderers now. By my count, there are 5 functions that only work in 2D. After those two issues above are filled, all 5 of those functions will log warnings and have proper documentation. I think that means we can close this issue. Thank you for filing the issue!

PR #3650, opened 2019-03-26
blendMode() for WebGL

### LINKED ISSUE #3649: blendMode in WEBGL
- [ ] Found a bug
- [x] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

Just adding an issue to indicate that I am going to add some of the more common `blendMode`s for WEBGL.

I am [using this for reference.](http://www.nutty.ca/articles/blend_modes/). I will probably only implement the modes that can be achieved with a configuration of `gl.blendFunc()` and/or `gl.blendEquation()` in the [same way that Processing has](https://github.com/processing/processing/blob/master/core/src/processing/opengl/PGraphicsOpenGL.java#L6239) though a shader-based solution would also be easy enough to implement.

Related:
#2756
#986

stalgiag:
closes #3649 
related to #2756

implements `blendMode()` for WebGL. Some modes don't work as they would require shader-based solutions and I simply went for blend functions. I don't think the missing modes are a big concern as the working modes in 3D are the same as they are for Processing.

This also includes a manual test and thorough unit tests.

Here is a screenshot of the manual test:
Screenshot (44)

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -405,6 +405,11 @@ module.exports = {
    */
   DIFFERENCE: 'difference',
   /**
+   * @property {String} SUBTRACT
+   * @final
+   */
+  SUBTRACT: 'subtract',
+  /**
    * @property {String} EXCLUSION
    * @final
    */
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -165,8 +165,30 @@ p5.Renderer2D.prototype._getTintedImageCanvas = function(img) {
 //////////////////////////////////////////////
 
 p5.Renderer2D.prototype.blendMode = function(mode) {
-  this.drawingContext.globalCompositeOperation = mode;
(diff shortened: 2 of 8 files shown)


Daandelange:
Hello, thanks a lot for this implementation @stalgiag .
My post is not directly about this feature but offers a possibility to extend the default blendModes.
I guess there's a reason for this feature not to be implemented. _(this is not a feature request)_

Anyways, I'd like to share my custom implementation as a reference. (sorry if this is not the right place).
It's a script that extends p5.js on runtime by overriding 2 prototype functions. (probably not the most flexible way, but easy to use!) It's code reflects the changes included in this pull.

Everything is here: https://editor.p5js.org/daan/sketches/XHj0iUmdU

PR #3652, opened 2019-03-27
Improve norm function reference

Ajayneethikannan:
The changes made in this PR are: 

- Add the guide line, which was not being drawn in subsequent draw calls due to noStroke( ) called below it.
- Rewrite 'refer to second example above' as 'refer to example above' as only one example is present and it already explains about out of range normalized values. 

Thank you !


lmccart:
thanks @Ajayneethikannan!

PR #3655, opened 2019-04-01
Explain Value to the Number Parameter:

code-warrior:
The explanation for an `alpha` value is missing here, but listed at
the URL https://p5js.org/reference/#/p5/color. Thus, that
explanation is copied and pasted here, replacing the regular dash
with an en-dash in the number range.

### DIFF
--- a/src/color/setting.js
+++ b/src/color/setting.js
@@ -153,7 +153,7 @@ require('./p5.Color');
  *                        color mode)
  * @param {Number} v3     blue or brightness value (depending on the current
  *                        color mode)
- * @param  {Number} [a]
+ * @param  {Number} [a]   alpha value relative to current color range (default is 0–255)
  * @chainable
  */
 


Ajayneethikannan:
Hey @code-warrior ! 

Since the explanation for alpha parameter is explained for the second [format](https://github.com/processing/p5.js/pull/3655/files#diff-39ee1ae1ff494d026d5e8118dc36ee44L136) accepted by the background( ) function,I think there is no need to write the description of alpha parameter again in the following formats for the same function. That is why the tests are not passing.

This [link](https://github.com/processing/p5.js/blob/master/developer_docs/inline_documentation.md#additional-signatures) can be used for reference.

Hope it's of help !

code-warrior:
Hi, @Ajayneethikannan,

Thank you for the clarification. I overlooked  the discussion of `alpha` in the **Description** section, because I was looking for a quick, one-sentence description under the **Parameters** section. Seeing an empty description to the Number parameter gave the impression that it was missing.

Thanks, again.

PR #3657, opened 2019-04-02
utilities/conversion.js: add unit tests

ihsavru:
Noticed some inconsistencies during creating unit tests #3658

### DIFF
--- a/test/unit/utilities/conversion.js
+++ b/test/unit/utilities/conversion.js
@@ -1,0 +1,313 @@
+suite.only('Conversion', function() {
+  var myp5;
+
+  setup(function(done) {
+    new p5(function(p) {
+      p.setup = function() {
+        myp5 = p;
+        done();
+      };
+    });
+  });
+
+  teardown(function() {
+    myp5.remove();
+  });
+
+  var result;
+
+  suite('p5.prototype.float', function() {
+    test('should be a function', function() {
+      assert.ok(myp5.float);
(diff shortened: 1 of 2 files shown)


test/unit/utilities/conversion.js
```diff
@@ -0,0 +1,313 @@
...
+      assert.typeOf(result, 'Number');
+      assert.strictEqual(result, 65);
+    });
+
+    test('should return the NaN for empty string', function() {
+      result = myp5.unhex('');
+      assert.isNaN(result);
+    });
+
+    test('should return the NaN for invalid hex string', function() {
```
ihsavru: currently this fails with `AssertionError: expected 202 to be NaN`

test/unit/utilities/conversion.js
```diff
@@ -0,0 +1,313 @@
...
+      assert.typeOf(myp5.hex, 'function');
+    });
+
+    test('should return the hex representation of the number', function() {
+      result = myp5.hex(65);
+      assert.typeOf(result, 'String');
+      assert.strictEqual(result, '00000041');
+    });
+
+    test('should return Infinity for Infinity', function() {
```
ihsavru: This is failing due to unknown behaviour, if it should return Infinity or "INFINITY" or "Infinity"

lmccart:
looks good thanks! I'd like to fix the underlying issues before we merge this in. working on that now.

PR #3662, opened 2019-04-02
reset() for p5.Graphics

### LINKED ISSUE #2586: WEBGL - when used in createGraphics() - light position does not get updated
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Core
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 
When working with WEBGL in a createGraphics environment and population that to a standard P5.js canvas with image() function - some functions may not work as in a normal P5.js WEBGL sketch.

For example I have noticed that the position of pointLight() gets not updated in the sketch, even when I clearly has changed. Is populates light from the original location and that does no change. What I figured "refreshes" the light position, is when I call the function createGraphics() again (but that slows down the performance by a lot).

I made two examples that I ran in the P5 editor sketch. They are and should work identical, but actually don't:

```
function setup() {
}

function draw() {
  background(0);
  directionalLight(250, 250, 250, mouseX-50, 0, 10);
  sphere(25);
}
```

versus

```
var a;

function setup() {
	createCanvas(200, 200);
	a = createGraphics(100, 100, WEBGL);  
}

function draw() {
	// Calling createGraphics on each draw() "helps", kinda
	// Uncomment the next line and see for yourself
	// a = createGraphics(100, 100, WEBGL);  
	
        background(60);
	a.resetMatrix();
	a.background(0);
	a.directionalLight(250, 250, 250, mouseX-50, 0, 10);
        a.sphere(150);
	image(a,0,0);
}
```

- p5.js version: 
https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.6.0/p5.min.js

- Web browser and version: 
Firefox 57.0.4 64-bit

- Operating System: 
Win 10 - 64bit

- Steps to reproduce this:
Just test the first code and than the second code here : https://alpha.editor.p5js.org/

### LINKED ISSUE #3079: Unexpected behavior using transforms with createGraphics
In the normal draw loop, if you use a translate rotate or scale the transforms get reset at the end of the draw loop. For example if you do

````
translate(width/2, height/2);
ellipse(0,0,100);
````

The ellipse will always be drawn in the center of the screen.

If you do this in a createGraphics layer, the transforms will compound so your shape will end up flying off screen.

````
graphics.translate(width/2, height/2);
graphics.ellipse(0,0,100);
````

Is this intentional? I assumed that createGraphics would behave like drawing normally does where transforms are reset at the end of the loop. The behavior occurs in both WEBGL and the 2D renderer. A push and pop will make things normal in the createGraphics layer, but I found it confusing that it didn't reset on it's own. If it is intended, is this documented anywhere? 

I've included an example at the bottom to demonstrate.

#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.6.1
- Web browser and version: 67
- Operating System: OSX 10.13.4
- Steps to reproduce this:

https://alpha.editor.p5js.org/aferriss/sketches/S1g0H_GmX

stalgiag:
fixes #2586 
fixes #3079 

This PR is extremely simple and is meant primarily as a discussion starter. Currently there is nothing equivalent to [Processing's `beginDraw()` and `endDraw()` for PGraphics objects.](https://processing.org/reference/createGraphics_.html). I think it is nice that this isn't necessary in p5 but it introduces a couple of problems as indicated in the issues listed above.

 Here it is [illustrated by this example.](https://editor.p5js.org/stalgiag/sketches/OwZF0PWtO) As soon as you start the sketch the `rect` in the p5.Graphics flies off the bottom right corner. This is because `translate()` is additive. This is reset during `redraw()` for the p5 drawing context but not for any of the p5.Graphics objs that the user may have made. This encountered less often in 2D because translate is more rarely used, but it is heavily relied on in 3D. This lack of resetting also affects the lighting in GL mode.

The suggestion here is to add a utility function named `endFrame()` as per the discussion in #3079. At first, I thought we should just go with something similar to beginDraw/endDraw in Processing but I don't think that is necessary. Those functions do some setting up of default values for the PApplet and [style property reapplication](https://github.com/processing/processing/blob/master/core/src/processing/core/PGraphics.java#L1019) in Processing which I don't think is necessary for us. So the flanking of drawing inside Graphics would be pointless other than to maintain similarity to Processing. As far as I can tell there are two possible approaches for this:

1. Keep track of all of the p5.Graphics objects that are made and automatically reset them in `redraw`
  - pro: fixes the problem without the user needing to understand what is happening and masks the differences between drawing into p5 normally and drawing into a p5.Graphics.
  - con: no user control over when the reset happens.
2. Add a function for people that want to explicitly reset their p5.Graphics. (this PR currently)
  - pro: simple to implement
  - con: requires an understanding of another function

Note: If we decide to go with the version that this PR suggest (#2) then I'll add more clear documentation and unit tests.

### DIFF
--- a/src/core/p5.Graphics.js
+++ b/src/core/p5.Graphics.js
@@ -63,6 +63,21 @@ p5.Graphics = function(w, h, renderer, pInst) {
 p5.Graphics.prototype = Object.create(p5.Element.prototype);
 
 /**
+ * End the drawing loop of p5.Graphics object. This resets the transform
+ * matrices and lighting.
+ *
+ * @method endFrame
+ *
+ */
+p5.Graphics.prototype.endFrame = function() {
+  // just reset the matrix in the Renderer for this p5.Graphics object
+  this._renderer.resetMatrix();
+  if (this._renderer.isP3D) {
+    this._renderer._update();
+  }
+};
+
+/**
  * Removes a Graphics object from the page and frees any resources
  * associated with it.
  *
--- a/src/core/p5.Graphics.js
+++ b/src/core/p5.Graphics.js
@@ -63,14 +63,51 @@ p5.Graphics = function(w, h, renderer, pInst) {
 p5.Graphics.prototype = Object.create(p5.Element.prototype);
 
 /**
- * End the drawing loop of p5.Graphics object. This resets the transform
- * matrices and lighting.
+ * Resets certain values such as those modified by functions in the Transform category
+ * and in the Lights category that are not automatically reset
+ * with graphics buffer objects. Calling this in <a href='#/p5/draw'>draw()</a> will copy the behavior
+ * of the standard canvas.
  *
- * @method endFrame
+ * @method reset
+ * @example
+ *
+ * <div><code>
+ * let pg;
+ * function setup() {
+ *   createCanvas(100, 100);
+ *   background(0);
+ *   pg = createGraphics(50, 100);
+ *   pg.fill(0);
+ *   frameRate(5);
+ * }
+ * function draw() {
+ *   image(pg, width / 2, 0);
+ *   pg.background(255);
+ *   // p5.Graphics object behave a bit differently in some cases
+ *   // The normal canvas on the left resets the translate
+ *   // with every loop through draw()
+ *   // the graphics object on the right doesn't automatically reset
+ *   // so translate() is additive and it moves down the screen
+ *   rect(0, 0, width / 2, 5);
+ *   pg.rect(0, 0, width / 2, 5);
+ *   translate(0, 5, 0);
+ *   pg.translate(0, 5, 0);
+ * }
+ * function mouseClicked() {
+ *   // if you click you will see that
+ *   // reset() resets the translate back to the initial state
+ *   // of the Graphics object
+ *   pg.reset();
+ * }
+ * </code></div>
+ *
+ * @alt
+ * A white line on a black background stays still on the top-left half.
+ * A black line animates from top to bottom on a white background on the right half.
+ * When clicked, the black line starts back over at the top.
  *
  */
-p5.Graphics.prototype.endFrame = function() {
-  // just reset the matrix in the Renderer for this p5.Graphics object
+p5.Graphics.prototype.reset = function() {
   this._renderer.resetMatrix();
   if (this._renderer.isP3D) {
     this._renderer._update();


stalgiag:
@lmccart @outofambit Any opinion on what seems best?

lmccart:
Nice find @stalgia. I think having p5.Graphics as a constant object that isn't automatically reset is a nice feature. For me it fits with the conceptual model of things made with `createX` besides the canvas are static rather than looping. 

I think there's also a performance benefit to this. I've seen people use p5.Graphics objects to create a performance intensive drawing a single time, that's then drawn into canvas on every frame.

My preference then is for option #2. I think we just need to make sure the behavior is clear and the link to the reset function is explained and available in this documentation too.

lmccart:
Is `endFrame()` the right name? Or perhaps it should be `reset()` and also clear the graphics element?

stalgiag:
`reset()` sounds good to me and I think it is a good idea to clear at the same time. Thanks for the input @lmccart

limzykenneth:
I would prefer option 2 as well. At least the way I understand p5.Graphics (or PGraphics in Processing) is that it is an additional drawing buffer that is not animated unless I manually animate it in `draw()`.

However, instead of adding `reset()`, does it overlap with `clear()`? Also does `clear()` reset the transform matrix? Should `reset()` clear the canvas or just reset the matrix and style?

stalgiag:
@limzykenneth that is a good point. `clear()` does currently reset the matrix for the 2D renderer. Maybe it is smarter to just ensure that `clear()` does the same for WebGL and then indicate more clearly in the docs that clear can be used to animate a new frame. p5.Graphics would probably would probably need its own prototype for 'clear()' at that point.

 In my opinion there are a bit of mental gymnastics for the user no matter what because for the currently problematic functions (`translate`, `rotate`, etc) one needs to `clear()` or start a new frame in order for the image to **not** animate. This is pretty contradictory to the base logic of animation. Not sure how that should be articulated in order for it to now be confusing.

stalgiag:
So now this is called `reset()`. I didn't include `clear()` in the function because I can imagine that someone might want to reset the transform matrices and lighting without clearing as happens with the end of the `draw()` loop for standard canvases.

lmccart:
@stalgiag this makes sense to me!

PR #3664, opened 2019-04-03
Fix default value in textToPoints doc

Niels-NTG:
The default value of sampleFactor is actually 0.1, not 0.25. This value can be found at [line 480 in the pathToPoints method](https://github.com/processing/p5.js/blob/master/src/typography/p5.Font.js#L480).


stalgiag:
Thanks for the contribution @Niels-NTG !

PR #3675, opened 2019-04-05
Added parser for STL file

### LINKED ISSUE #2734: loadModel changed to accept STL files
#### Nature of issue?

- [x] New feature request

- [x] Desktop/Laptop

#### New feature details:

The [STL file format](https://en.wikipedia.org/wiki/STL_(file_format)) (stereolithography files) are an old lowest-common-denominator ASCII-based format used in 3D modeling and 3D printing. It would be nice to support loading them through `loadModel()`.

This would be analogous to the previously added feature of "Loading 3d Objects from OBJ" (#904) only for the simple STL file format instead.

In Processing(Java), STL support is implemented in the third-party toxiclibs library, the STLReader class, under the [loadBinary](https://github.com/postspectacular/toxiclibs/blob/6896bc17e85320e886bdf0819c66491ce484728e/src.core/toxi/geom/mesh/STLReader.java#L90) method.

See a previous discussion of p5.js STL support on the forum here: https://forum.processing.org/two/discussion/18812/import-stl

sanketsingh24:
Added a parser for ASCII STL file. 
Fixes #2734 

1. Work Done-
  - STL ASCII files can be parsed and loaded using loadModel().

2. To Do- 
  - Implement same for Binary STL files
  - Check to distinguish

3. Problems- 
  - Loading ArrayBuffer using loadBytes()  or httpdo()

_More to follow, I will update all three points soon, currently this STL parser for ASCII files works as expected._


sanketsingh24:
Actually I have exams from Monday, hence the delay, will start working on this soon after

sanketsingh24:
Added parser for both Binary and ASCII STL files.

Methodology- 
1. File is loaded using `httpDo` method and used as `arraybuffer`.
2. arraybuffer is then checked if its ASCII file or Binary file.
3. If the file type is binary, the arraybuffer is used and model is made.
4. If the file type is ASCII, the arraybuffer is converted into an array of strings and model is made.
5. Model is returned.

I also added two manual tests example, one for each file type, along with a small unit test and modified documentation.

Output from ASCII STL file-
DeepinScreenshot_select-area_20190517173736

Output from binary STL file-
DeepinScreenshot_select-area_20190517173753

src/webgl/loading.js
```diff
@@ -241,6 +265,300 @@ function parseObj(model, lines) {
...
+ *
+ * We need to convert the arrayBuffer to an array of strings,
+ * to parse it as an ASCII file.
+ */
+function parseSTL(model, buffer) {
+  if (isBinary(buffer)) {
+    parseBinarySTL(model, buffer);
+  } else {
+    var reader = new DataView(buffer);
+    var decoder = new TextDecoder('utf-8');
```
stalgiag: TextDecoder is still [flagged as experimental on MDN](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder). I am not sure if this matters a ton as the only unsupported browser is IE which since we are switching to ES6 this year is less important. Edge has unknown compatibility but they are switching over to Chromium so also not super important. Just worth noting and thinking about whether there is an easy alternative. A quick google search reveals mostly polyfills which I think aren't worth it because they would just be removed later this year.

src/webgl/loading.js
```diff
@@ -109,26 +112,47 @@ p5.prototype.loadModel = function(path) {
...
   var self = this;
-  this.loadStrings(
-    path,
-    function(strings) {
-      parseObj(model, strings);
 
-      if (normalize) {
-        model.normalize();
-      }
+  if (fileType === '.stl') {
```
stalgiag: I think if you are going to slice the file extension and test for it, you should probably do the same for `.obj` and then log an error if the file extension isn't either.

src/webgl/loading.js
```diff
@@ -241,6 +265,300 @@ function parseObj(model, lines) {
...
+ * https://en.wikipedia.org/wiki/STL_%28file_format%29#Binary_STL
+ *
+ * Currently there is no support for the colors provided in STL files.
+ */
+function parseBinarySTL(model, buffer) {
+  var reader = new DataView(buffer);
+
+  // Number of faces is present following the header
+  var faces = reader.getUint32(80, true);
+  var r,
```
stalgiag: Per vertex color isn't enabled in GL currently. I think that it might be a nice addition eventually but that seems like its own separate task. It is is probably best to remove all of the color parsing code for now. If iper-vertex-color gets added down the line then this code can get added back in.  If you'd like to archive this work you could push this branch to another before removing the code. Then start an issue here requesting that per-vertex color is added to GL, and add a link to your archive branch for reference. I know this is a bit of work but it would be helpful. Thank you!

test/manual-test-examples/webgl/loadSTL/loadASCII/sketch.js
```diff
@@ -0,0 +1,19 @@
+let teapot;
```
stalgiag: I think that the ASCII and binary manual tests can be combined for ease of viewing. Also, I'd move the models to the assets folder one directory up for consistency.

src/webgl/loading.js
sanketsingh24: [Edge has a compatibility in development](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/encodingstandard/), and for the time being, [I will include a polyfill](https://developers.google.com/web/updates/2012/06/How-to-convert-ArrayBuffer-to-and-from-String), but the polyfill is slower.

src/webgl/loading.js
sanketsingh24: @aferriss and I were discussing this topic exactly. The thing is, If I don't read the color bits if they are present, they will create an offset and those color bits will be computed as vertices, so I have to atleast read them, and I thought If I am already reading them, why not compute them too, as in the future if the feature is added, It will be useful. Performance wise I don't think they affect too much right now. I will add the comments so that in future it will be easier to activate them.

sanketsingh24:
@stalgiag I am able to run both manual tests. Can you tell your specs? Mine are - Arch Linux on Chrome 73.0.3683.86.

stalgiag:
@sanketsingh24 Yes sorry should've thought to mention. It does not work in Chrome 74.0.3729.157 running on Windows 10.

Strangely, I was able to get it working on Firefox 67.0 on the same OS. The HTML relative directory reference wasn't working for whatever reason in FF (might be something to do with my local server) but I moved a copy of the p5 build to the same directory and the manual tests work in FF.

Not sure about Chrome. I can do some more investigation.

src/webgl/loading.js
stalgiag: I don't think you have to include the polyfill. As I said, the transition to ES6 is more or less underway and I think it is okay to stop supporting IE. Maybe just check for TextDecoder and log something about an unsupported browser.

src/webgl/loading.js
stalgiag: That makes sense!

stalgiag:
Continuing to test - works on Firefox 65.0 also. Does not work on different computer running Windows 10 and same version of Chrome. Also does not work on Edge 42.17134.1.0.

If you don't have a computer with Windows installed then I can look into the Chrome problem more when I get a chance.

sanketsingh24:
I actually have a Windows 8 and a Windows 10 installed in a VM, I will surely test on those too. But I highly doubt it will run on Edge because of `TextDecoder`.

test/unit/io/loadModel.js
```diff
@@ -1,13 +1,14 @@
...
   var invalidFile = '404file';
   var validFile = 'unit/assets/teapot.obj';
+  var validSTLfile = 'unit/assets/ascii.stl';
 
   test('_friendlyFileLoadError is called', async function() {
     const _friendlyFileLoadErrorStub = sinon.stub(p5, '_friendlyFileLoadError');
     try {
       await promisedSketch(function(sketch, resolve, reject) {
         sketch.preload = function() {
-          sketch.loadImage(invalidFile, reject, resolve);
```
sanketsingh24: Maybe a typo, this test was testing for loadImage in loadModel.js

stalgiag:
I polyfilled Edge when testing because it was when I thought you might add the polyfill but I think you can just ignore that browser for now.

aferriss:
I just took a look at this on windows 10 in chrome 74 after pulling in master and it needed a small change to get working. It looks like stl files have no way to store texture coordinates, but the fill shader expects there to be texture coordinate buffers bound. 

I'm not sure if it's the perfect solution but just pushing in dummy texture coordinates `model.uvs.push([0,0])` for both ascii and binary parsing got the models to load. It looks like this is what the [obj loader does as well](https://github.com/processing/p5.js/blob/master/src/webgl/loading.js#L214) when there are no uv's found in the model.

sanketsingh24:
@aferriss Thanks for the hard work! I will make the changes ASAP. This method was destined to fail on Edge, but now that Edge is being rebuilt on Chromium, it is less relevant, and also [Edge has a compatibility in development](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/encodingstandard/). For the time being, @stalgiag I think I should include the polyfill, What are your thoughts on this?

stalgiag:
Thanks @aferriss! I think it is fine to include the polyfill. I know that there is going to be a pass at removing a bunch of polyfills soon with the ES6 update. Maybe add a comment explaining the conditions for removing the polyfill since it is specifically tied to Edge development?

sanketsingh24:
@stalgiag where should I place the polyfill? I think it will be [just this file](https://github.com/inexorabletash/text-encoding/blob/master/lib/encoding.js)

stalgiag:
@sanketsingh24 typically polyfills are installed as npm packages. [Here is the info about that polyfill in npm package form.](https://www.npmjs.com/package/text-encoding-polyfill)

sanketsingh24:
Don't know why but this same test is passing for me on my local system for same config.

stalgiag:
Okay! Sorry for flip-flopping on this issue but I think that putting the text decoder polyfill in might be a bit too much for this particular feature 😖 I checked this package on [bundlephobia](https://bundlephobia.com/result?p=text-encoding-polyfill@0.6.7) and realized that it will add 18.1 kb even in minified form to the p5 library. I think this is a little too much to add to the library (even temporarily) if the sole purpose is support for ascii stl loading in Microsoft Edge. Again, sorry about that I should've realized before giving approval to add it. Would you be able to remove the package and just add a console log that says 'ASCII STL loading only works in browsers that support TextDecoder (https://caniuse.com/#feat=textencoder)'. You are welcome to make a revert commit or just add a new removal commit. If this feels too burdensome, we can merge as is and I can make the changes since I am responsible for the mixup. Your call!

Also just so you know, I reset the travis tests. Sometimes they can hang on one of the JSON tests but if you go to the travis website and log in with your github you should be able to restart the tests.

sanketsingh24:
No problem @stalgiag Its a small change! I will remove the polyfill and add the warning.

Also, thanks for the info regarding CI check!

src/webgl/loading.js
```diff
@@ -241,6 +275,308 @@ function parseObj(model, lines) {
...
+ * to parse it as an ASCII file.
+ */
+function parseSTL(model, buffer) {
+  if (isBinary(buffer)) {
+    parseBinarySTL(model, buffer);
+  } else {
+    var reader = new DataView(buffer);
+
+    if (!('TextDecoder' in window)) {
+      console.error(
```
stalgiag: Hi @sanketsingh24 sorry I didn't catch this before but we try not to use `console.error` in this project. I understand that we want the user to know when there model isn't working and that continuing to load it doesn't make sense but it is better if their sketch works as much as possible and this will cause the whole sketch to come to a halt. This is more of a matter of the design, intention, and philosophy of the library. If a user does something like:
```
let myModel;

function preload() {
  myModel = loadModel('myBrokenSTL');
} 

function draw() {
  background(0, 0, 255);
  model(myModel);
}
```
We would want the person who wrote this to see the blue background. Just to give them the feeling that _something_ they are doing is working. Does that make sense?

sanketsingh24: It perfectly makes sense! I will do the requested.

sanketsingh24: @stalgiag Actually `console.error` does not break the sketch, in fact, your code will actually render the blue background with an error message in the console. Maybe you are confused with `throw new error`, which actually stops the execution. Maybe I can change it to `console.warn` if you want?

stalgiag: Oh yes but still better to switch to warn just to make it clear that execution will continue.

PR #3686, opened 2019-04-08
Fix loadJSON() usage with JSONP

### LINKED ISSUE #3684: loadJSON erroring in 0.7.2 & 0.7.3 when using jsonp
#### Nature of issue?

- [X] Found a bug

- [X] Data
- [X] Events
- [X] IO

- [X] Chromebook

#### Details about the bug: 

- p5.js version: 0.7.2 & 0.7.3
- Web browser and version: Google Chrome	74.0.3729.58 (Official Build) beta (64-bit)
- Operating System: ChromeOS
- Steps to reproduce this:
Using the following code
```
let url = "https://en.wikipedia.org/w/api.php" +
           "?action=parse&format=json&prop=langlinks" +
           "&section=0&redirects&page=dog";

function setup() {    
    loadJSON(url, gotData, "jsonp");
  }

function gotData(data) {
  console.log(data);
}
```

In versions prior to 0.7.2 the data loads and is logged to the console. In versions 0.7.2 & 0.7.3 the data is not logged and I get two identical errors in the Chrome error console of 
```
p5.min.js:3 Uncaught (in promise) TypeError: Cannot read property 'get' of undefined
    at VM363 p5.min.js:3
```

However, when I view the Network tab it appears that the json has been loaded from the network.

Example sketches:
[Using 0.7.3](https://editor.p5js.org/piecesofuk/sketches/HjFnwOPve)
[Using 0.7.1 ](https://editor.p5js.org/piecesofuk/sketches/gHNAM5CPO)

limzykenneth:
Fixes #3684 

Disables size check for JSONP requests. I don't know if there is a way to check for JSONP request size or not seeing that it basically creates script tags and I'm not even sure if there is a size limit for JSONP so I just disable the header checks for JSONP requests as they don't return headers.

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -1141,7 +1141,10 @@ p5.prototype.httpDo = function() {
       err.ok = false;
       throw err;
     } else {
-      var fileSize = res.headers.get('content-length');
+      var fileSize = 0;
+      if (type !== 'jsonp') {
+        fileSize = res.headers.get('content-length');
+      }
       if (fileSize && fileSize > 64000000) {
         p5._friendlyFileLoadError(7, path);
       }


meiamsome:
Yep, this should work. It looks like we have no tests for the JSONP features - we should open an issue for that

PR #3694, opened 2019-04-11
loadJSON: add unit tests for jsonp

### LINKED ISSUE #3687: Add unit tests for JSONP features
right now we don't have any tests for JSONP features. lets add some!

cc @meiamsome (#3686)

ihsavru:
fixes #3687

### DIFF
--- a/test/unit/io/loadJSON.js
+++ b/test/unit/io/loadJSON.js
@@ -1,7 +1,9 @@
-suite('loadJSON', function() {
+suite.only('loadJSON', function() {
   var invalidFile = '404file';
   var jsonArrayFile = 'unit/assets/array.json';
   var jsonObjectFile = 'unit/assets/object.json';
+  var jsonpObjectFile = 'unit/assets/object.js';
+  var jsonpArrayFile = 'unit/assets/array.js';
 
   test('_friendlyFileLoadError is called', async function() {
     const _friendlyFileLoadErrorStub = sinon.stub(p5, '_friendlyFileLoadError');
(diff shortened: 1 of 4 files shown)


test/unit/io/loadJSON.js
```diff
@@ -149,4 +151,35 @@ suite('loadJSON', function() {
...
+    });
+    assert.isObject(json);
+  });
+
+  test('passes an array to success callback for array JSONP.', async function() {
+    const json = await promisedSketch(function(sketch, resolve, reject) {
+      sketch.preload = function() {
+        sketch.loadJSON(
+          jsonpArrayFile,
+          { jsonpCallbackFunction: 'resolve' },
```
limzykenneth: The callback function itself doesn't need to be called `resolve` does it? Just the `loadJSON()` success callback calls the promised `resolve` callback.

ihsavru: I was earlier calling jsonCallbackFunction as a success callback which in turn resolved the promise but @meiamsome suggested that I can call resolve directly instead of creating another function. I am naming that callback function `resolve` which is also padding our json

limzykenneth: The `jsonpCallbackFunction` here relates to what the function that is defined in the JSONP response is called. It is sent to the server so that the server can wrap its response in the right function name. 

It does not relate the the callback of `loadJSON` itself, ie the `resolve` a couple lines down from here. That particular `resolve` is calling the `resolve` callback function defined by `promisedSketch`'s callback function above.

In other words, what @meiamsome suggest is to call the `promisedSketch`'s `resolve` as `loadJSON`'s success callback below (instead of calling it in a separate function). The value of `jsonpCallbackFunction` isn't related to the success callback and can be named anything else (and probably should to reduce confusion).

ihsavru: Oh! I am sorry about the confusion, I thought that the `jsonpCallbackFunction` was supposed to be the success callback. I fixed it now (hopefully 😅 )

limzykenneth: All good! 👍

Issue #3674, opened 2019-04-05
keyIsDown() should return false onBlur


ihsavru:
`onBlur()`, we assume that all keys have been released, therefore `keyIsDown(x)` should return false for every key `x`. However, it currently returns undefined because onBlur resets `downKeys` to empty object

```
p5.prototype._onblur = function(e) {
  this._downKeys = {};
};

p5.prototype.keyIsDown = function(code) {
  p5._validateParameters('keyIsDown', arguments);
  return this._downKeys[code];
};
```

WenheLI:
That's true! I can fix it if you haven't dealt the code yet!

ihsavru:
@WenheLI  please go ahead!

lmccart:
fixed with https://github.com/processing/p5.js/pull/3696

PR #3696, opened 2019-04-13
fix onBlur undefined return

WenheLI:
This PR is about https://github.com/processing/p5.js/issues/3674.


lmccart:
thanks @WenheLI

PR #3697, opened 2019-04-13
fix text texture updating

### LINKED ISSUE #3695: webgl text() fails to render new glyphs on second call in v0.8.0
#### Nature of issue?

- [X] Found a bug

- [X] Typography
- [X] WebGL

- [X] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop

#### Details about the bug: 

- p5.js version: v0.8.0
- OS and Web browsers: 
Windows 7x64:
Chrome 73.0.3683.103 
Chrome 75.0.3764.0
Firefox 66.0.3
Firefox 67.0b10
Opera 60.0.3255.27
Android 7:
Chrome 73.0.3683.90
iOS 12.1.4:
Chrome 73.0.3683.68
Safari 12

- Steps to reproduce:
Expected additional calls to text() to draw new glyphs.
Below is a minimal sketch that reproduces the issue.
Conditions are: webgl, additional calls to text(), different glyphs than first call, any browser
Sketch on OpenProcessing:
https://www.openprocessing.org/sketch/698922
(Note: this was working in v0.7.3 For comparison, if p5 v0.7.3 is selected, the sketch works as expected and both the "A" and "B" glyphs are correctly rendered.)

No console errors.

```js
function preload() {
  f = loadFont('inconsolata.otf');
}

function setup() {
  textFont(f);
  textSize(40);
}

function draw() {
  background(100);
  text("A", -30, 0);
	
  // this next call to text() 
  // does not display in v0.8.0
  text("B", 10, 0);
  // Note: if the same glyphs as the first instance
  // are used, it will work.
}
```

Spongman:
closes #3695

#3646 made it so `bindShader()` no longer calls `bindTextures()`. this PR fixes the text rendering code to accommodate that, assuming that that behavior is going to continue.

### DIFF
--- a/src/webgl/text.js
+++ b/src/webgl/text.js
@@ -650,6 +650,7 @@ p5.RendererGL.prototype._renderText = function(p, line, x, y, maxY) {
   var initializeShader = !this._defaultFontShader;
   var sh = this._getFontShader();
   sh.init();
+  sh.bindShader(); // first time around, bind the shader fully
 
   if (initializeShader) {
     // these are constants, really. just initialize them one-time.
@@ -688,7 +689,6 @@ p5.RendererGL.prototype._renderText = function(p, line, x, y, maxY) {
   try {
     var dx = 0; // the x position in the line
     var glyphPrev = null; // the previous glyph, used for kerning
-    var shaderBound = false;
     // fetch the glyphs in the line of text
     var glyphs = font.stringToGlyphs(line);
     for (var ig = 0; ig < glyphs.length; ++ig) {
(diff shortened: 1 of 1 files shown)


stalgiag:
Thanks for the fix @Spongman. Sorry for not catching that text was relying on that differently. I am okay with `bindShader` calling `bindTextures` in the the future but we would have to figure out how to handle the timing so that uniforms are correct before doing that.

Spongman:
no worries. i think it's probably best to leave it the way to have it right now (separated). maybe it's worth revisiting the naming of those internal methods sometime to make them more accurately reflect what's going on inside...

PR #3714, opened 2019-04-29
VSCode Suggested Extensions

outofambit:
This PR adds two config files for vscode:

* `.vscode/extensions.json`, which prompt the user to install a couple vscode extensions when they open the p5 repo (unless they already have them)
* `.vscode/settings.json`, which provides a few settings to make those plugins useful out of the box

The plugins are:

* `eg2.vscode-npm-script`, which allows you to run npm scripts from within vscode
* `esbenp.prettier-vscode`, which runs prettier formatting on save for files edited in vscode.
* `yzhang.markdown-all-in-one`, which adds some handy markdown shortcuts (thinking of `developer_docs/` here)
* `dbaeumer.vscode-eslint`, which will show eslint errors inline while editing code

Additionally, I moved some `prettier` config out of the `eslint` config and into its own config so it can be reused between `eslint` and standalone `prettier`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,16 +15,8 @@
     "ecmaVersion": 2017
   },
   "rules": {
-    "prettier/prettier": [
-      "error",
-      {
-        "singleQuote": true
-      }
-    ],
-    "no-cond-assign": [
-      2,
-      "except-parens"
-    ],
+    "prettier/prettier": ["error"],
+    "no-cond-assign": [2, "except-parens"],
     "eqeqeq": ["error", "smart"],
     "no-use-before-define": [
       2,
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ docs/reference/*
 !*.gitkeep
 examples/3d/
 .idea
-.vscode
 dist/
 p5.zip
 bower-repo/
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "singleQuote": true
+}
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,13 @@
+{
+  // See http://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+  // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+  // List of extensions which should be recommended for users of this workspace.
+  "recommendations": [
+    "eg2.vscode-npm-script",
+    "esbenp.prettier-vscode",
+    "yzhang.markdown-all-in-one"
+  ],
+  // List of extensions recommended by VS Code that should not be recommended for users of this workspace.
+  "unwantedRecommendations": []
+}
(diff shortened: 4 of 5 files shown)


xtydev:
So far the biggest letdown for me has been the [lack of proper IntelliSense](/Microsoft/TypeScript/issues/6942) due to the somewhat dated build process. If only we could ditch UMD [in favour of ES modules](../issues/2503)…

outofambit:
@xtydev thanks for those links! i've been wanting to look into what our options are for getting us some intellisense support...

stalgiag:
@outofambit I am not sure what is happening and this may be specific to my setup but I haven't edited anything about my environment and since this PR the behavior is now broken for me in VSCode. Here is what happens:

When I save a file, a space is inserted after every `function` so :
`p5.prototype.saveGif = function(pImg, filename) {`
becomes
`p5.prototype.saveGif = function (pImg, filename) {`
Which is not correct and causes lint testing to fail. Note this is fixed when I run linting the standard way via `CTRL+SHIFT+P:ESLint Fix all auto-fixable problems` or `npm run lint:fix`.

Currently the only way that I can make commits is to save and then run `npm run lint:fix` and then commit.

I can't find any rule for 'space-before-function-paren' in the project 😕 

I am running Windows 10 Pro with VSCode 1.34

Spongman:
>i've been wanting to look into what our options are for getting us some intellisense support...

@outofambit we tried getting intellisense into the project by generating typescript definition files as part of the build process, and thus providing type information to typescipt-aware editors (like vscode). unfortunately this was moved to a separate project, so it's continuously out of date, and packaged for those that use p5 as a node package, which is approximately nobody. maybe it's time to revisit this?

i build these myself and use them in my own editor:

image

stalgiag:
@Spongman I am very interested in generating typescript definition files. Intellisense would help a ton. Might be worth opening an issue?

outofambit:
I'm happy to discuss this further in an issue, but i have serious reservations about having typescript definitions in the p5 repo, as that definitely raises the barrier to contributing.

We can get _some_ intellisense support from what @xtydev mentions above (https://github.com/processing/p5.js/pull/3714#issuecomment-487653378), which i think will be covered by the work coming out of #3758.

Spongman:
> typescript definitions in the p5 repo

i think there might be a misunderstanding. i'm not suggesting that typescript definitions be kept _in_ the repo, or that _any_ typescript knowledge be required to contribute to p5. i'm just suggesting that we re-introduce the automatic production of typescript definition files (from the jsdoc) as part of the build/release process, so that they would be available on the release page for those that would like to use them in their editor. (ideally they would be shipped in the .zip along with the library in such a way as to provide automatic rich intellisense for typescript-aware editors, but that's a different discussion).

Spongman:
can we revert the eslint stuff from this PR, it's messing up the formatting and killing my battery. we already have the precommit stuff, i think enforcing this in the editor is overkill.

outofambit:
> i think there might be a misunderstanding

@Spongman ah, you are right. thanks for clarifying! i agree with @stalgiag that it would be great to open an issue to outline what you're thinking and get feedback. thanks!

Issue #3564, opened 2019-03-04
Issues with CORS in image


Ajayneethikannan:
#### Nature of issue?

- [x] Existing feature enhancement

#### Most appropriate sub-area of p5.js?

- [x] Image
- [x] Other (specify if possible) - DOM

#### Which platform were you using when you encountered this?
- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version:  0.7.3
- Web browser and version:   All browsers
- Operating System:  All OS
- Steps to reproduce this:

```
let a, b, c, d, canvas;
function setup() {
 a = createImg("https://i.imgur.com/Jvh1OQm.jpg");
 b = loadImage("https://i.imgur.com/Jvh1OQm.jpg");
 c = loadImage("Jvh1OQm.jpg");
 d = createImg("Jvh1OQm.jpg");
 canvas = createCanvas(400, 400);
}

function draw() {
image(a, 0, 0, 400, 400);
image(b, 0, 0, 200, 200);

}
function keyPressed()
{
  let pixel = canvas.get(10,10, 1, 1);
  console.log(pixel);

}
function mousePressed()
{
    saveCanvas('image_cors', canvas, 'jpg');
}

```

#### Feature enhancement details:

- The current implementation of cross origin in p5.dom.js is always setting ,
`
img.crossOrigin = 'Anonymous';
`
This can result in two Http requests, even in the case where the image is in the same domain, leading to potential slowdown. This article can be referred to for this issue : https://webglfundamentals.org/webgl/lessons/webgl-cors-permission.html

- One more issue is that , when we are setting up the crossOrigin parameter for every image, in 
createImg function, the domain from which we are requesting the image , can stop sending us the image, even if it can be used outside the canvas. Since images created by createImg function are not always used inside the canvas, this implementation can block images, which can be rendered inside the img tag, but not inside the canvas, from rendering in the img tag also. So I think we should set this attribute only for images to be used in the canvas, but I am not sure about the implementation (maybe image function can be modified to check for the attribute ? )

( This issue is a follow up to #3482, as a discussion )

salus-sage:
my 2 cents considering browser security restrictions, not sure what will be the best hack for this problem. 
what i found in image/loading_displaying.js
```
 * The path to the image should be relative to the HTML file
 * that links in your sketch. Loading an image from a URL or other
 * remote location may be blocked due to your browser's built-in
 * security.
```
[...]

PR #3715, opened 2019-04-29
Revamped createImg() to solve issues concerning CORS

xtydev:
To solve #3564, `createImg()` in `p5.dom` has been revamped to account for the addition of an optional `crossOrigin` parameter.

Changes involved:
* Reversed 6ed05304fe2da3eef00ab04a0db7d285843aad17, so that we do not use CORS by default.
* Added an optional `crossOrigin` parameter to `createImg()`, enabling the user to explicitly opt for a CORS reuquest, if they intend to use the `` with `<canvas>` without marking it as _tainted_.
* Add corresponding `@param` JSDoc tag documenting the use of the `crossOrigin` parameter.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -419,45 +419,39 @@
    * @method createImg
    * @param  {String} src src path or url for image
    * @param  {String} [alt] alternate text to be used if image does not load
+   * @param  {String} [crossOrigin] <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes">crossOrigin property</a> of the &lt;img&gt; element; use either 'anonymous' or 'use-credentials' to retrieve the image with cross-origin access (for later use with &lt;canvas&gt;; by default, if no value or undefined is passed, CORS is not used
    * @param  {Function} [successCallback] callback to be called once image data is loaded
    * @return {p5.Element} pointer to <a href="#/p5.Element">p5.Element</a> holding created node
    * @example
    * <div class='norender'><code>
-   * createImg('http://p5js.org/img/asterisk-01.png');
+   * createImg('https://p5js.org/assets/img/asterisk-01.png');
    * </code></div>
    */
   /**
    * @method createImg
    * @param  {String} src
    * @param  {Function} successCallback
-   * @return {Object|p5.Element}
+   * @return {p5.Element}
    */
   p5.prototype.createImg = function() {
     p5._validateParameters('createImg', arguments);
     var elt = document.createElement('img');
-    elt.crossOrigin = 'Anonymous';
     var args = arguments;
     var self;
-    var setAttrs = function() {
-      self.width = elt.offsetWidth || elt.width;
-      self.height = elt.offsetHeight || elt.height;
-      if (args.length > 1 && typeof args[1] === 'function') {
-        self.fn = args[1];
-        self.fn();
-      } else if (args.length > 1 && typeof args[2] === 'function') {
-        self.fn = args[2];
-        self.fn();
-      }
-    };
-    elt.src = args[0];
     if (args.length > 1 && typeof args[1] === 'string') {
       elt.alt = args[1];
     }
-    elt.onload = function() {
-      setAttrs();
-    };
-    self = addElement(elt, this);
-    return self;
+    if (args.length > 2 && typeof args[2] === 'string') {
+      elt.crossOrigin = args[2];
+    }
+    elt.addEventListener('load', function() {
+      self.width = elt.offsetWidth || elt.width;
+      self.height = elt.offsetHeight || elt.height;
+      var last = args[args.length - 1];
+      if (typeof last === 'function') last();
+    });
+    elt.src = args[0];
+    return (self = addElement(elt, this));
   };
 
   /**


lib/addons/p5.dom.js
```diff
@@ -419,45 +419,39 @@
    * @method createImg
    * @param  {String} src src path or url for image
    * @param  {String} [alt] alternate text to be used if image does not load
+   * @param  {String} [crossOrigin] <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes">crossOrigin property</a> of the &lt;img&gt; element; use either 'anonymous' or 'use-credentials' to retrieve the image with cross-origin access (for later use with &lt;canvas&gt;; by default, if no value or undefined is passed, CORS is not used
```
limzykenneth: This is a bit verbose for me, the more complete description of this functionality should probably be in the description of the function itself. The parameter description should ideally just say what this is and maybe the possible values for it.

Also you probably shoud add this to the other function signature below as well.

limzykenneth:
I'll test the code locally to make sure it works as expected. Will report back.

The parameter handling is a bit messy as well but that's totally unrelated to this issue so I may open another issue to fix this later.

limzykenneth:
Changes works as expected but there is one problem. In order to use the CORS parameter, you must provide the `alt` parameter and so it creates an ambigous situation where:

```javascript
createImg("https://p5js.org/assets/img/asterisk-01.png", "anonymous");
```

The second parameter above will be treated as the alt text and the image will not be retrieved with the right headers. As such the function signature is no longer

```javascript
createImg(src, [alt], [crossOrigin], [successCallback])
createImg(src, [crossOrigin], [successCallback])
```

but rather some more ambiguous one where whether `alt` is compulsory or not depends on whether `crossOrigin` is given or not while `successCallback` is always optional, something like:

```javascript
createImg(src, [alt], [successCallback])
createImg(src, alt, [crossOrigin], [successCallback])
```

Might be fine if we update the docs to clearly reflect this quirk.

outofambit:
I wonder if keeping `alt` required whenever there is more than one argument might help it be easier for people to reason about? @limzykenneth do you have a preference?

ie
```js
createImg(src)
createImg(src, alt, [crossOrigin], [successCallback])
```

Spongman:
Those overloads are still ambiguous. In order for this to validate correctly you’ll need to have `crossOrigin` be required in the second overload.

lib/addons/p5.dom.js
xtydev: Sure, I'll shorten the parameter description and add an example to illustrate its use.

The point of having a second method signature, IMO, was to allow people to provide a `successCallback` without specifying the optional `alt` parameter. In the same vein, they shouldn't have to specify `crossOrigin` if they just want to supply a `successCallback` in the second method signature, so I think it's better to leave it as is.

xtydev:
> Changes works as expected but there is one problem. In order to use the CORS parameter, you must provide the `alt` parameter and so it creates an ambigous situation where:
> 
> ```js
> createImg("https://p5js.org/assets/img/asterisk-01.png", "anonymous");
> ```
> 
> The second parameter above will be treated as the alt text and the image will not be retrieved with the right headers. As such the function signature is no longer
> 
> ```js
> createImg(src, [alt], [crossOrigin], [successCallback])
> createImg(src, [crossOrigin], [successCallback])
> ```
> 
> but rather some more ambiguous one where whether `alt` is compulsory or not depends on whether `crossOrigin` is given or not while `successCallback` is always optional, something like:
> 
> ```js
> createImg(src, [alt], [successCallback])
> createImg(src, alt, [crossOrigin], [successCallback])
> ```
> 
> Might be fine if we update the docs to clearly reflect this quirk.


`'anonymous'` being treated as `alt` in your example would be the result of incorrect use of the method, not the result of a design flaw in the API, and no, `alt` is always optional regardless of the presence of `crossOrigin`.  To skip the optional `alt` parameter, you'd need to call `createImg()` like this:
```JavaScript
createImg('https://p5js.org/assets/img/asterisk-01.png', undefined, 'anonymous');
```
which is how optional parameters work in TypeScript and other graphics libraries like PixiJS. That is, if you want to omit an optional parameter, you should explicitly pass in an `undefined` in the right place, unless it is the last parameter. Handling optional parameters this way is in fact, might be the best way to avoid ambiguities. I suggest we follow suit and do not allow too much flexibility, which would make parameter handling even messier than it already is.

xtydev:
I agree that adding a detailed example in the docs illustrating the use of `crossOrigin` will definitely provide better clarity. I should be able to start working on it once we have reached a consensus on the method signatures and expected behaviour.

outofambit:
> Those overloads are still ambiguous.

@spongman ah good point. so this would be better:

```js
createImg(src)
createImg(src, alt, crossOrigin, [successCallback])
```

@xtydev I don't think @limzykenneth was saying the API is flawed, so much as the docs don't quite match the API. I agree that passing `undefined` is the most failsafe way to work with optional parameters in JS, but i think that's a complexity we'd like to avoid for (at least most) users of p5js.

limzykenneth:
@xtydev For me

```javascript
createImg(src, [alt], [crossOrigin], [successCallback])
```

means that `alt`, `crossOrigin` and `successCallback` are all optional and any of them can be omitted, ie
```javascript
createImg(src, alt, successCallback)
createImg(src, crossOrigin, successCallback)
```
should both be possible. But obviously (to us, not to someone new to programming) this creates an ambiguous situation where we don't know if the second argument is supposed to be `alt` or `crossOrigin`.

For me it is totally fine to make `alt` a required parameter at least in the case with `crossOrigin` as it is good practice to include it anyway when creating `` tags. I would prefer the suggestion is not `undefined` as that has quite a bit a caveat when used like this, if the user really doesn't want to provide alt text, they can pass in an empty string (`""`) instead.

@outofambit I'm not sure what signatures will be best for both validation and flexibility for the user. I think it may be fine to have more signature tha currently but just so we cover most cases. Maybe something like so:
```javascript
createImg(src, [successCallback])
createImg(src, alt, [successCallback])
createImg(src, alt, crossOrigin, [successCallback])
```
@Spongman You are more familiar than I am with the parameter validation stuff so do have a look.

lib/addons/p5.dom.js
limzykenneth: The signature here is not just to document possible ways to use the function, it is also used by the unminified library to validate the parameters passed to the function by the user. If we want to illustrate a case where both `alt` and `crossOrigin` are omitted but `successCallback` is provided, we can add an additional signature instead.

Spongman:
we just need to ensure that it's possible to disambiguate based on the **types** of the declared parameters. it's not enough to have two parameters that can take different values. also, optional parameters are always omitted right-to-left, meaning you can't omit an optional parameter without also omitting _all_ the parameters to its right.

so this is NOT correct:

> createImg(src, [alt], [crossOrigin], [successCallback])
>
> means that alt, crossOrigin and successCallback are all optional and any of them can be omitted, ie
> 
> createImg(src, alt, successCallback)
> createImg(src, crossOrigin, successCallback)


if you want to be able to drop parameters from the middle, you need to specify multiple overloads.

i would suggest:

````
createImg([src], [alt], [successCallback])
createImg(src, alt, crossOrigin, [successCallback])
````

**Comment by limzykenneth** [comment]
@Spongman Yeah, that is what I get from it and in this case totally wrong. 😝 Which in a way in kind of the point as it is a bit confusing and doesn't reflect that parameters in teh middle can be dropped.

The signature you provided looks good to me, but it does raise a question: is `src` optional?

**Comment by Spongman** [comment]
oh, sorry, `src` is not optional.

**Comment by xtydev** [comment]
> so much as the docs don't quite match the API

Ah, I didn't think there was a mismatch, but now that I've gone through the comments yet another time, I can see that miscommunication arose because we have different interpretations as to what it means for a parameter to be _optional_ in JSDoc. You guys seem to unanimously take _optional_ parameters as parameters that can be skipped entirely, whereas my understanding of the term is more in line with TypeScript and how it's thought of in PixiJS (which uses JSDoc too); the following excerpt may give you a better idea:
```JavaScript
/**
 * Convenience function to set the position, scale, skew and pivot at once.
 *
 * @param {number} [x=0] - The X position
 * @param {number} [y=0] - The Y position
 * @param {number} [scaleX=1] - The X scale value
 * @param {number} [scaleY=1] - The Y scale value
 * @param {number} [rotation=0] - The rotation
 * @param {number} [skewX=0] - The X skew value
 * @param {number} [skewY=0] - The Y skew value
 * @param {number} [pivotX=0] - The X pivot value
 * @param {number} [pivotY=0] - The Y pivot value
 * @return {PIXI.DisplayObject} The DisplayObject instance
 */
setTransform(x = 0, y = 0, scaleX = 1, scaleY = 1, rotation = 0, skewX = 0, skewY = 0, pivotX = 0, pivotY = 0)
{
    this.position.x = x;
    this.position.y = y;
    this.scale.x = !scaleX ? 1 : scaleX;
    this.scale.y = !scaleY ? 1 : scaleY;
    this.rotation = rotation;
    this.skew.x = skewX;
    this.skew.y = skewY;
    this.pivot.x = pivotX;
    this.pivot.y = pivotY;
    return this;
}
```
I actually prefer their way as this would make parameter handling much cleaner.

> I agree that passing `undefined` is the most failsafe way to work with optional parameters in JS, but i think that's a complexity we'd like to avoid for (at least most) users of p5js.

Fair enough, but I'd like you to consider several downsides to the "skippable version" that you're in favour of:
* This would impose further restrictions on what parameters we can have for any function. That is, it's quite impossible to have two consecutive optional parameters of the same type without introducing ambiguity.
* There would be a significant cost to implement parameter handling (i.e. deciding which argument corresponds to which parameter can be tedious), especially for functions with a lot of optional parameters.
* Text editors wouldn't give correct parameter hinting (e.g. in VS Code), which will potentially confuse the user, making it somewhat counterproductive.

xtydev:
> For me it is totally fine to make `alt` a required parameter at least in the case with `crossOrigin` as it is good practice to include it anyway when creating `` tags. I would prefer the suggestion is not `undefined` as that has quite a bit a caveat when used like this, if the user really doesn't want to provide alt text, they can pass in an empty string (`""`) instead.

@limzykenneth Yeah, we might be able to circumvent the issue by making `alt` a required parameter in this case. However, the bigger issue for me is that, sticking to the skippable / droppable definition of _optional parameter_ would mean we'll never be able to have two consecutive optional parameters of the same type in **any function** without introducing ambiguity. I just want to make sure we're all aware of its implications and decide whether that is a worthwhile sacrifice before we commit to making all optional parameters skippable / droppable.

Spongman:
@xtydev 

> You guys seem to unanimously take optional parameters as parameters that can be skipped entirely

@xtydev Can you point to what gave you this impression? We may have another docs bug somewhere, since the validation usually only allows right-to-left dropping of optional parameters (although it doesn’t strictly enforce this in the docs).

xtydev:
> For me it is totally fine to make `alt` a required parameter at least in the case with `crossOrigin` as it is good practice to include it anyway when creating `` tags. I would prefer the suggestion is not `undefined` as that has quite a bit a caveat when used like this, if the user really doesn't want to provide alt text, they can pass in an empty string (`""`) instead.

@limzykenneth Yeah, we might be able to circumvent the issue by making `alt` a required parameter in this case. However, the bigger issue for me is that, sticking to the skippable / droppable definition of _optional parameter_ would mean we'll never be able to have two consecutive optional parameters of the same type in **any function** without introducing ambiguity. I just want to make sure we're all aware of its implications and decide whether that is a worthwhile sacrifice before we commit to making all optional parameters skippable / droppable.

Could you elaborate a bit on the "caveat" you mentioned? Omitting an optional parameter in the middle of a parameter list by passing an `undefined` in its place seems to be quite commonplace, as explained above.

Furthermore, I'd like to point out that in the current method signatures,
```JavaScript
  /**
   * Creates an &lt;img&gt; element in the DOM with given src and
   * alternate text.
   * Appends to the container node if one is specified, otherwise
   * appends to body.
   *
   * @method createImg
   * @param  {String} src src path or url for image
   * @param  {String} [alt] alternate text to be used if image does not load
   * @param  {Function} [successCallback] callback to be called once image data is loaded
   * @return {p5.Element} pointer to <a href="#/p5.Element">p5.Element</a> holding created node
   * @example
   * createImg('http://p5js.org/img/asterisk-01.png');
   * </code></div>
   */
  /**
   * @method createImg
   * @param  {String} src
   * @param  {Function} successCallback
   * @return {Object|p5.Element}
   */
```
if _optional parameters_ were deemed skippable / droppable like you seem to have agreed, the second signature would be considered redundant. Therefore, it looks like there has not been a real consensus yet as to what exactly _optional parameters_ entail in p5.js and it's not too late to make it right 😉.

xtydev:
> Can you point to what gave you this impression?

Oh, it's just that @limzykenneth identified the second argument in
```JavaScript
createImg("https://p5js.org/assets/img/asterisk-01.png", "anonymous");
```
being treated as the alt text as a "problem". It seemed to me he was suggesting that I have effectively made `alt` a _required_ parameter by making it "undroppable".

Incidentally, you did say "Those overloads are **still** ambiguous", which gave me the impression that you too thought my proposed changes were ambiguous in the first place… it **would be** ambiguous if optional parameters were seen as parameters that can be skipped entirely.

xtydev:
> Can you point to what gave you this impression?

@Spongman Oh, it's just that @limzykenneth identified the second argument in
```JavaScript
createImg("https://p5js.org/assets/img/asterisk-01.png", "anonymous");
```
being treated as the alt text as a "problem". It seemed to me he was suggesting that I have effectively made `alt` a _required_ parameter by making it "undroppable".

Incidentally, you also said "Those overloads are **still** ambiguous", which gave me the impression that you too thought my proposed changes were ambiguous in the first place… Well, it **would be** ambiguous if optional parameters were seen as parameters that can be skipped entirely.

lib/addons/p5.dom.js
xtydev: > If we want to illustrate a case where both `alt` and `crossOrigin` are omitted but `successCallback` is provided, we can add an additional signature instead.

I'm not sure if I've understood you correctly, but doesn' the second signature already do that?
https://github.com/processing/p5.js/blob/40b69e699a2f91637cbfb70e866e87def945edae/lib/addons/p5.dom.js#L430-L435

So, why do you think we need yet another signature?

meiamsome:
@xtydev You would need at least a third parameter set: `src, [alt], [crossOrigin], [successCallback]`, `src, alt, successCallback`, `src, successCallback` to maintain compatibility for all options. From where I'm looking this is the only problem with this design in terms of the documented options.

I can see that there could be confusion if someone tries to set cross origin without realizing they would need an alt. We could add extra global parameters - `ANONYMOUS` and `USE_CREDENTIALS` for the options perhaps, and type them such that we can support all the potential overloads:

`src, [alt], [crossOrigin], [successCallback]`
`src, alt, successCallback`
`src, crossOrigin, successCallback`
`src, successCallback`

limzykenneth:
Optional parameters isn't really the right word here. The PixiJS example you gave isn't doing optional parameters either, it is using default parameter values, and before ES6 we would be manually checking if those parameters are `undefined` or not and assign values to them. Optional parameter means that you can omit the parameter entirely from the function call and as @Spongman mentioned should only be omitted from right to left. 

What we are trying to achieve here and in some other functions in p5.js is function overloading, ie the same function having multiple possible signatures. This concept does not exist in Javascript which is why there are all the checks for argument length and argument types in these kind of functions. A result of function overloading is that the function signature is decided by the types of the arguments given to it, their number, and their order (and the function return type in languages that supports it), which means that if you have a function that has two signatures like so:
```
function foo(string a, string b){
}

function foo(string a, string c){
}
```
It is impossible to disambiguate whether you are calling the first signature or the second signature. Which is the problem you observed:
> we'll never be able to have two consecutive optional parameters of the same type in any function without introducing ambiguity

Although there are instances where it is possible to disambiguate them but let's just say for now there aren't. The main difference between default parameter value and function overloading is that with function overloading you can potentially be executing entirely different code depending on the parameters that is passed to the function while default parameter value simply supply a value when it isn't given while executing the same code in the function.

With this function, the only optional parameter is `successCallback` while having three function signatures (@Spongman after some thinking I think we need to add one to what you proposed as you can have `successCallback` without providing `alt`):
```
createImg(src, [successCallback])
createImg(src, alt, [successCallback])
createImg(src, alt, crossOrigin, [successCallback])
```

If the user so wishes, they can totally call the function like how it is called in PixiJS `createImg(src, undefined, "anonymous")` (providing `undefined` is not an optional parameter, you are still providing a parameter which happens to have a value of `undefined`). I am not saying to somehow make `createImg(src, "anonymous")` work. I am merely suggesting that for the example to not call the function with a parameter of `undefined` but with `""` instead.

To be clear, there is no issue with the implementation of `crossOrigin` here. The only thing that would need to be fixed is the documentation. Your implementation is fine.

limzykenneth:
@meiamsome Just saw your reply. I like the extra global parameters idea but according to the specs of `` tag the `crossorigin` attribute can take any random characters which will then default to `anonymous`, ie anything that is not `use-credentials` is treated as `anonymous` unless you omit the attribute. That said, there are no other possible value for it (for now at least) so maybe this idea could work? 🤷‍♂

lib/addons/p5.dom.js
limzykenneth: Let's keep discussion about which signature to include on the main thread. Treat this comment as about the verbosity only.

meiamsome:
@limzykenneth There's three states (Unset, set as "use-credentials" or anything else which is considered "anonymous") - I'd be tempted to ignore the ability of other text entry, because that's true for any HTML attribute - they all have specific defaults in this case. 

The Constant option I suspect we would have to assign them objects or numbers so that it's unambiguous by types. I believe this should work for FES, but maybe it won't? I think an object (`{}`) and `"{}"` (or `1` and `"1"`) would be considered the same for FES atm because they're compared on object keys and thus implicit string cast first (@Spongman). This seems like a fixable bug in FES by changing the object here to an array and just checking if it's in the array: https://github.com/processing/p5.js/blob/master/src/core/error_helpers.js#L275

xtydev:
> Optional parameters isn't really the right word here. The PixiJS example you gave isn't doing optional parameters either, it is using default parameter values,

@limzykenneth FYI, the [JSDoc 3 documentation](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values) refers to the following code as "**an optional parameter and default value**":
```JavaScript
/**
 * @param {string} [somebody=John Doe] - Somebody's name.
 */
function sayHello(somebody) {
    if (!somebody) {
        somebody = 'John Doe';
    }
    alert('Hello ' + somebody);
}
```
So, the PixiJS example I referenced earlier **IS** doing optional parameters. They just happen to be optional parameters with default values. For the purpose of our discussion, optional parameters **without** default values don't work that differently; Here is an example of that:
```JavaScript
/**
 * Calculates the local position of the display object relative to another point.
 *
 * @param {PIXI.IPoint} position - The world origin to calculate from.
 * @param {PIXI.DisplayObject} [from] - The DisplayObject to calculate the global position from.
 * @param {PIXI.IPoint} [point] - A Point object in which to store the value, optional
 *  (otherwise will create a new Point).
 * @param {boolean} [skipUpdate=false] - Should we skip the update transform
 * @return {PIXI.IPoint} A point object representing the position of this object
 */
toLocal(position, from, point, skipUpdate)
{
    if (from)
    {
        position = from.toGlobal(position, point, skipUpdate);
    }
    if (!skipUpdate)
    {
        this._recursivePostUpdateTransform();
        // this parent check is for just in case the item is a root object.
        // If it is we need to give it a temporary parent so that displayObjectUpdateTransform works correctly
        // this is mainly to avoid a parent check in the main loop. Every little helps for performance :)
        if (!this.parent)
        {
            this.parent = this._tempDisplayObjectParent;
            this.displayObjectUpdateTransform();
            this.parent = null;
        }
        else
        {
            this.displayObjectUpdateTransform();
        }
    }
    // simply apply the matrix..
    return this.worldTransform.applyInverse(position, point);
}
```
where `from` and `point` are optional parameters without default values.

> If the user so wishes, they can totally call the function like how it is called in PixiJS `createImg(src, undefined, "anonymous")` (providing `undefined` is not an optional parameter, you are still providing a parameter which happens to have a value of `undefined`).

No, given my proposed implementation, passing in an `undefined` as the second argument would be a way of saying "I want to omit the optional `alt`, and it's okay for an omitted optional parameter (in the form of `undefined`) to be followed by other optional parameters. The takeaway here is, what I mean by _optional parameters_ is in line with PixiJS, TypeScript, etc., but your interpretation differs. It's not hard to find out the difference. I wouldn't say my interpretation is canonical, but it's certainly more mainstream. If you can change your mind around what _optional parameters_ mean, there'd be no need to change the proposed signatures, as they're **unambiguous**.

xtydev:
I've spent WAY too much time on this today 😱. Didn't expect it to take this long. Might not be able to respond in the next week or two. 🙏My apologies.

limzykenneth:
What's stated in JSDoc3 documentation does not contradict what I said. 
```javascript
/**
 * @param {string} [somebody=John Doe] - Somebody's name.
 */
function sayHello(somebody) {
    if (!somebody) {
        somebody = 'John Doe';
    }
    alert('Hello ' + somebody);
}
```

`somebody` here is both optional and it provides default value. Optional means you can leave it out completely when you call the function and leave it out from right to left. So I can call `sayHello()` and `somebody` is optional. **Passing `undefined` is not the same concept as omitting a parameter**. The parameter is not optional if you have to pass `undefined` to it (when it has required parameters after it).

If you look at Typescript:
```typescript
function foo(a?: number, b: number){
}
```
this will not compile as it is passing an optional parameter before a required one, while:
```typescript
function foo(a: number = 1.0, b: number){
}
```
will compile as it is using default parameter value and allows you to call the function with first argument as `undefined`. (https://www.typescriptlang.org/docs/handbook/functions.html) 

Passing `undefined` as optional parameter is not mainstream as it is not a definition in most libraries three.js doesn't do it, paper.js doesn't do it (they do function overload as well), jQuery doesn't do it, vue.js doesn't do it, even PixiJS had a problem with Typescript before because of it (https://github.com/pixijs/pixi.js/issues/5541).

And again https://stackoverflow.com/questions/46958782/why-does-typescript-require-optional-parameters-after-required-parameters this is not a unique point of view that optional means leaving out, and not unique to Javascript/Typescript either (https://php.net/manual/en/functions.arguments.php)

xtydev:
> What's stated in JSDoc3 documentation does not contradict what I said.

Actually, it does. In particular, you said:
>The PixiJS example you gave **isn't doing** optional parameters either, it is using default parameter values

whereas according to the [JSDoc 3 documentation](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values) it is **doing both**.

As for your TypeScript example – of course, required parameters have to come before any optional parameters (not including those with default values) – I don't have any problems with that. For avoidance of doubt, I **never suggested** optional parameters (not including those with default values) can be followed by required parameters.

By saying
> **Passing `undefined` is not the same concept as omitting a parameter** (it is a quirk in Javascript that it has the same behaviour).

You have made it very clear as to where you stand but this is also where our disagreement lies. What I have been suggesting is, **passing `undefined` in the place of an optional parameter is a way of saying "I want to omit this optional parameter"**. If it happens to be the last one in the parameter list, then of course, you can leave it off instead, but if its in the middle of the parameter list, you can pass in an `undefined` to denote the absence of value for that parameter, and then go on passing values for the rest (which would have to be optional as well). For example in TypeScript, we can have:
```TypeScript
function createImg(src: string, alt?: string, crossOrigin?: string, callback?: () => void): HTMLImageElement {
   // Note: different from the createImg() signature I proposed, for demo only
    const img = new Image();
    if (typeof alt === 'string') img.alt = alt;  
    if (typeof crossOrigin === 'string') img.crossOrigin = crossOrigin;
    if (callback) img.addEventListener('load', callback);
    img.src = src;
    return img;
}
```
Notice here both `alt` and `crossOrigin` are optional parameters, as they're each followed by a `?`. This will compile without errors. Want to omit `alt` when calling the function? Sure, you can do that because it is **optional**. All you have to do is:
```TypeScript
createImg('https://p5js.org/assets/img/asterisk-01.png', undefined, 'anonymous');
```
The `@param` JSDoc tag that I proposed in the commit should work very similarly. Well, the return type is `p5.Element` but you should get the idea.

I know I haven't expanded on the disagreement over _optional parameters_, but I really don't have time for that right now 😓, so I guess I'll leave it for next time 💨

Spongman:
The current commit will fail to validate the following call:

````javascript
createImg('foo.png', 'alt tag', function MyCallback() {});
````

outofambit:
hi @xtydev just wanted to ask and see if you think you'll have any time to return to this soon? if not (which is totally fine!), i might take this the last step so we can merge it! thanks!

outofambit:
continuing on at #3890. will do my best to incorporate the feedback given here and request reviews to see what y'all think!

PR #3720, opened 2019-05-02
material.js shader() example

hsab:
addresses #1954
- [x] https://p5js.org/reference/#/p5/shader

extends the loadShader example to allow for toggling between 2 shaders on mouse click/press.

### DIFF
--- a/docs/yuidoc-p5-theme/assets/shader-gradient.frag
+++ b/docs/yuidoc-p5-theme/assets/shader-gradient.frag
@@ -1,0 +1,19 @@
+// Code adopted from "Creating a Gradient Color in Fragment Shader"
+// by Bahadır on stackoverflow.com
+// https://stackoverflow.com/questions/47376499/creating-a-gradient-color-in-fragment-shader
+
+precision highp float; varying vec2 vPos;
+uniform float pos;
+
+void main() {
+
+  vec2 st = vPos.xy + vec2(0,pos);
+
+  vec3 color1 = vec3(1.9,0.55,0);
+  vec3 color2 = vec3(0.226,0.000,0.615);
+
+  float mixValue = distance(st,vec2(0,1));
+  vec3 color = mix(color1,color2,mixValue);
+
+  gl_FragColor = vec4(color,mixValue);
+}
--- a/src/webgl/material.js
+++ b/src/webgl/material.js
@@ -188,6 +188,49 @@ p5.prototype.createShader = function(vertSrc, fragSrc) {
  * @chainable
  * @param {p5.Shader} [s] the desired <a href="#/p5.Shader">p5.Shader</a> to use for rendering
  * shapes.
+ *
+ * @example
+ * <div modernizr='webgl'>
+ * <code>
+ * // Click within the image to toggle
+ * // the shader used by the quad shape
+ *
+ * let mandel;
+ * let gradient;
+ * let useMandel = false;
+ *
+ * function preload() {
+ *   // load the shader definitions from files
+ *   mandel = loadShader('assets/shader.vert', 'assets/shader.frag');
+ *   gradient = loadShader('assets/shader.vert', 'assets/shader-gradient.frag');
+ * }
+ * function setup() {
+ *   createCanvas(100, 100, WEBGL);
+ *   // initially use the gradient shader
+ *   shader(gradient);
+ *   noStroke();
+ * }
+ *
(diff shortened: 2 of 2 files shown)


docs/yuidoc-p5-theme/assets/shader-gradient.frag
```diff
@@ -0,0 +1,19 @@
+// Code adopted from "Creating a Gradient Color in Fragment Shader"
+// by Bahadır on stackoverflow.com
+// https://stackoverflow.com/questions/47376499/creating-a-gradient-color-in-fragment-shader
+
+precision highp float; varying vec2 vPos;
+uniform float pos;
+
+void main() {
+
+  vec2 st = vPos.xy + vec2(0,pos);
```
hsab: @stalgiag I agree, and yet the reason why I opted for the Mandel shader was exactly the familiarity of its reoccurrence throughout the reference manual. But the fact that is it could seem like a blackbox for users/beginners is undeniable. 

Perhaps I should change `pos` to `offset`. Then the shader could  coherently be extended so that:
vec2 st = vPos.xy + vec2(xOffset,yOffset);

color1 and color2 could perhaps be uniforms as well. That way the shader could potentially be reused with a certain level of modularity that is visually distinguishable:
- moves along X || moves along Y || moves on XY plane || static
- full-color control

this could allow two instances of this shader to properly reflect the effects of `shader() ` for this particular example/commit:
shader1: blue-orange moving vertically
shader2: red-green moving horizontally 

what do yo think?

hsab: @stalgiag I agree, and yet the reason why I opted for the Mandel shader was exactly the familiarity of its reoccurrence throughout the reference manual. But the fact that is it could seem like a blackbox for users/beginners is undeniable. 

Perhaps I should change `pos` to `offset`. Then the shader could  coherently be extended so that:
vec2 st = vPos.xy + vec2(xOffset,yOffset);

color1 and color2 could perhaps be uniforms as well. That way the shader could potentially be reused with a certain level of modularity that is visually distinguishable:
- moves along X || moves along Y || moves on XY plane || static
- full-color control

this could allow two instances of this shader to properly reflect the effects of `shader() ` for this particular example/commit:
shader1: blue-orange moving vertically
shader2: red-green moving horizontally 

what do yo think?

hsab: or even better `uniform vec2 offset;` rather than `uniform float xOffset` and `uniform float yOffset`

hsab: as for the `quad()` and `rect()` situation, we could perhaps opt in for something like [this](https://codepen.io/dollee/pen/XEdqyM). Which as a solution, relies primarily on:
`gl_FragCoord.xy / resolution.xy`
This allows for eliminating +-1 values for the quad and simply using `rect(0,0,width,height);`
Yet the ambiguity might come from `setUniform('resolution',[width,height]);`

On a side note, I tried looking for docs on the webgl vs p2d coordinate differences. Are you aware if such a thing exists in the codebase/references?

stalgiag: Thanks for the quick response! I like the offset with two gradient shader instances idea. I understand the the impulse to repeat the Mandel shader. It might be a good idea for familiarity sake but it also seems like a good time to start moving away from it and maybe filing an issue to simplify the other examples away from the Mandel shader as well.

I'll leave it up to you as to whether to change the vertex shader. I don't have extremely strong feelings and it would be `rect(-width/2, -height/2, width, height)` which when combined with setting the resolution uniform doesn't provide the largest gains in terms of simplicity.

As for coordinate differences, the[ getting started with webgl wiki](https://github.com/processing/p5.js/wiki/Getting-started-with-WebGL-in-p5#3d-coordinate-system) explains this a bit. This decision dates back to the first work on WebGL mode in 2015. There are pros and cons to doing this but I agree that maybe it makes sense to try to find another more prominent place for this in documentation. Not sure if there are other mentions that I am missing.

hsab: Thanks for the info, although I think it would have been `rect(-width/2, -height/2, width*2, height*2)`, I decided to just opt in for `quad()`.

I'm gonna close this pull req, and open another one with the latest commit.

Issue #640, opened 2015-05-27
Add getObject() and getArray() methods to p5.Table


workergnome:
Following the discussion in #280, we think it would be useful to have a `getObject()` and `getArray()` methods on p5.Table.  These should return native javascript objects and functions, respectively.  

The syntax of these should probably match `getString()` or `getNumber()`.

christinedt:
Merged!

Issue #280, opened 2014-07-31
p5.Table and p5.TableRow


therewasaguy:
I've begun work on classes that emulate Processing's [Table](https://www.processing.org/reference/Table.html) and [TableRow](https://www.processing.org/reference/TableRow.html) here: https://github.com/lmccart/p5.js/pull/279

Most of the methods from Processing are implemented, but with a few differences that reflect JavaScript. I think there could be room for more differences. For example...

* p5.TableRow could have ```getArray()``` and ```getObject()``` methods. Under the hood, that's what every TableRow is made out of anyway. At least for now... I imagine there are better ways to handle this type of data structure in JS, and would love some feedback.

* I have yet to work on ```saveTable()```. Looking at [filesaver.js](https://github.com/eligrey/FileSaver.js) and [stack overflow](http://stackoverflow.com/questions/17836273/export-javascript-data-to-csv-file-without-server-interaction) for inspiration, curious if anyone has advice.

* There is only one ```get()``` and one ```set()``` rather than ```setInt()```, ```setFloat()```, ```setString()``` . Processing's Table also has typed rows. I don't think this makes sense in JavaScript. But would set/getNumber() be a nice option to convert what might be a String into a Number?

* Processing's loadTable() has an options parameter that can include multiple options, like: ```loadTable("data.csv", "header, tsv")```. Or, the options parameter can be omitted. But in p5, loadTable() also needs an optional callback, so I'm not sure how to order the parameters. Should it be <code>loadTable(path, options, callback) </code>? What if a user wants to use a callback without any options? I'm thinking there could a separate <code>tableOptions()</code> method. For now, loadTable assumes default options of .csv with a Header just to make sure that works.

kevinb7:
@therewasaguy even though JavaScript doesn't differentiate between ints and floats it's nice to have those methods to do type conversion especially downcast float data to integer data.  Also, it maintains API compatibility with other versions of processing as @benfry pointed out.  I think the untyped get/set methods are a useful addition.

lmccart:
@therewasaguy I'm porting gswp right now and looking at some of the data stuff. I noticed there are `get()` and `set()` methods for p5.TableRow, but not for p5.Table. Processing has these for both. ok if I go ahead and add? I think I will also add getInt, getFloat, getString methods as well?

lmccart:
ah I see we have `getNum()` and `getString()` already, so I guess I will just add those same methods to p5.Table as well?

therewasaguy:
[...]

PR #3721, opened 2019-05-02
Table getObject() function now returns an array of objects

### LINKED ISSUE #3277: table.getObject() edit proposal
In `p5.table` we have a `getObject()` function. When called, it creates an object with a property for each line having as property name the row index:
```
{
  "0": {"Title1":"value1", "Title2":"value2"},
  "1": {"Title1":"value1", "Title2":"value2"},
  ...
}
```

However, in most of the other libraries (like as d3.js) CSV are parsed as array containing objects:
```
{
 values: [
    {"Title1":"value1", "Title2":"value2"},
    {"Title1":"value1", "Title2":"value2"},
  ...
}
```
This would be more meaningful in my opinion, since in this way we can use all the vanilla array operators provided by JS.

It would be also more consistent and would provide less errors. Finally, it allows to add other properties to the table object as metadata.

I know it would change a behaviour present in previous versions of P5 but i think it worths it.

mikima:
The function now returns an array of objects.

It is more consistent with other libraries and makes it possible to use vanilla js array operators, such as `foreach()`.

[Unit tests](https://github.com/processing/p5.js/tree/master/developer_docs#unit-tests) were performed and passed.

Closes #3277.

### DIFF
--- a/src/io/p5.Table.js
+++ b/src/io/p5.Table.js
@@ -1270,7 +1270,7 @@ p5.Table.prototype.getString = function(row, column) {
  *
  */
 p5.Table.prototype.getObject = function(headerColumn) {
-  var tableObject = {};
+  var tableObject = [];
   var obj, cPos, index;
 
   for (var i = 0; i < this.rows.length; i++) {


mikima:
@outofambit table has already a [getArray() function](https://p5js.org/reference/#/p5.Table/getArray). It return an array of arrays, without headers.

I had a look to previous discussion on the topic but i found nothing apart the original feature request(#280, #640).

Adding a third function(getArrayOfObjects?) sounds redundant to me, but surely doable.

I still think that the best option is to modify the getObject function in order to get an array of object, which is more ore less the standard way to store data today (correct me if i'm wrong). The current object in not really convenient for iterations over the dataset, which is one of the main uses of this function.

Do you have other ideas? Otherwise i can add the mentioned function.

lmccart:
This issue for this pull request was closed, we're not able to merge at this time.

PR #3722, opened 2019-05-02
material.js shader() example / remove Mandel

hsab:
addresses #1954
- [x] https://p5js.org/reference/#/p5/shader

resolves #3720 

extends the loadShader example to allow for toggling between 2 shaders on mouse click/press.

@stalgiag let me know what you think :thumbsup:

### DIFF
--- a/docs/yuidoc-p5-theme/assets/shader-gradient.frag
+++ b/docs/yuidoc-p5-theme/assets/shader-gradient.frag
@@ -1,0 +1,19 @@
+// Code adopted from "Creating a Gradient Color in Fragment Shader"
+// by Bahadır on stackoverflow.com
+// https://stackoverflow.com/questions/47376499/creating-a-gradient-color-in-fragment-shader
+
+precision highp float; varying vec2 vPos;
+uniform float pos;
+
+void main() {
+
+  vec2 st = vPos.xy + vec2(0,pos);
+
+  vec3 color1 = vec3(1.9,0.55,0);
+  vec3 color2 = vec3(0.226,0.000,0.615);
+
+  float mixValue = distance(st,vec2(0,1));
+  vec3 color = mix(color1,color2,mixValue);
+
+  gl_FragColor = vec4(color,mixValue);
+}
--- a/docs/yuidoc-p5-theme/assets/shader-gradient.frag
+++ b/docs/yuidoc-p5-theme/assets/shader-gradient.frag
@@ -2,18 +2,21 @@
 // by Bahadır on stackoverflow.com
 // https://stackoverflow.com/questions/47376499/creating-a-gradient-color-in-fragment-shader
 
+
 precision highp float; varying vec2 vPos;
-uniform float pos;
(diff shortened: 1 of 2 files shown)


stalgiag:
Cool! I think I had a bit of a misunderstanding about what you were going to do but this looks good. There is something about this feels like it makes more sense as a `setUniform` example since the reused shader makes the second shader object a bit unnecessary. Something like:
```
let shade;
let showRedGreen = false;

function preload() {
  shade = loadShader('shader.vert', 'shader-gradient.frag');
}

function setup() {
  noStroke();
}

function draw() {
  if (showRedGreen) {
    shade.setUniform('colorCenter', [1.0, 0.0, 0.0]);
    shade.setUniform('colorBackground', [0.0, 1.0, 0.0]);
    shade.setUniform('offset', [0, sin(millis() / 2000) + 1]);
  } else {
    shade.setUniform('colorCenter', [1.0, 0.5, 0.0]);
    shade.setUniform('colorBackground', [0.226, 0.0, 0.615]);
    shade.setUniform('offset', [sin(millis() / 2000), 1]);
  }
  shader(shade);
  quad(-1, -1, 1, -1, 1, 1, -1, 1);
}

function mouseClicked() {
  showRedGreen = !showRedGreen;
}
```
would explain setUniform well. But it also works as a `shader` example. There is also the opportunity to do both and to show separate approaches. What do you think?

hsab:
@stalgiag What you have makes perfect sense for setUniform. However it seems to not fully translate the function of `shader()`. Perhaps Mandel was more illuminating in that regard as the user was tangibly interacting with two distinct shaders rather than two instances. 

I highly agree with your final note. I think keeping this example PR as is for `shader()` and also updating the `setUniform` to what you have illustrated is a wholesome solution that underlines different approaches. 

Perhaps we can point the reader to check the other reference example to see the alternative? Do you think this is appropriate for the docs examples? If so should I just add a comment in the example code:

```
// Click within the image to toggle
// the shader used by the quad shape
// visit https://p5js.org/reference/#/p5/setUniform
// for a different approach, utilizing setUniform() function
```

```
// Click within the image to toggle
// the uniform variables used by the shader
// visit https://p5js.org/reference/#/p5/shader
// for a different approach, utilizing shader() function
```

stalgiag:
That sounds good to me. I don't think there are other instances of this kind of cross-referencing examples thing but I think it works well as an instructional tool. 

It is your call about whether to do them as separate PRs. I don't think it is completely necessary since you are editing two related examples.

As an sidenote, this is mostly personal preference but you might find it easier to work with a branch (eg hsab/p5js:shaderMethodExamples). I find it makes keeping my work-in-progress inline with master a bit more manageable.

hsab:
Yeah branching is definitely the way to go. I didn't intend to really commit or do PR was mostly checking the codebase and the grunt tasks and decided to jsut give docs:dev a try. Ended up requesting a PR. Will close this one, branch and recommit the changes.

PR #3731, opened 2019-05-08
rewrite sentence for clarity

lcmolina:
i felt like this sentence could be rewritten to help the reader better understand where the command for fill will be in comparison to the shapes that will be colored by the fill command.

### DIFF
--- a/src/color/setting.js
+++ b/src/color/setting.js
@@ -340,7 +340,7 @@ p5.prototype.colorMode = function(mode, max1, max2, max3, maxA) {
 
 /**
  * Sets the color used to fill shapes. For example, if you run
- * fill(204, 102, 0), all subsequent shapes will be filled with orange. This
+ * fill(204, 102, 0), all shapes run subsequent to the fill command will be filled with the color orange. This
  * color is either specified in terms of the RGB or HSB color depending on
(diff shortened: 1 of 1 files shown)


lmccart:
@lcmolina this is great! I'd suggest replacing the word "subsequent" with "after" to simplify. I think it's a more common word for non-native english speakers.

src/color/setting.js
```diff
@@ -340,7 +340,7 @@ p5.prototype.colorMode = function(mode, max1, max2, max3, maxA) {
 
 /**
  * Sets the color used to fill shapes. For example, if you run
- * fill(204, 102, 0), all subsequent shapes will be filled with orange. This
+ * fill(204, 102, 0), all shapes run subsequent to the fill command will be filled with the color orange. This
```
limzykenneth:
```suggestion
 * fill(204, 102, 0), all shapes drawn after the fill command will be filled with the color orange. This
```

For further clarity.

phemyojo:
shape.filled orange.ok

src/color/setting.js
lcmolina: i agree; "drawn after" definitely provides further clarity than "subsequent", thank you @limzykenneth

lmccart:
🎉🎉

PR #3732, opened 2019-05-08
Provide description for WEBGL

maddyfisher:
Added a description for the WEBGL constant where beginners are able to understand clearly what WEBGL is and why p5.js is using it.

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -16,7 +16,8 @@ module.exports = {
    */
   P2D: 'p2d',
   /**
-   * One of the two render modes in p5.js: P2D (default renderer) and WEBGL. Enables 3D render by introducing the third dimension: Z. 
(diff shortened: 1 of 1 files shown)


lmccart:
@maddyfisher looking great so far! my one suggestion is to break this description into two lines for easier reading. something like this:

```js
* One of the two render modes in p5.js: P2D (default renderer) and WEBGL.
* Enables 3D render by introducing the third dimension: Z. 
```

maddyfisher:
@lmccart yes, I agree, looks much better!

lmccart:
@maddyfisher the linter is confusing here. the problem is not actually the periods, it's the single space after the last character (period or not). so `Delete /`·/`` actually means delete space.

lmccart:
🙌

PR #3739, opened 2019-05-13
Delta time implementation

### LINKED ISSUE #3737: Delta time feature
#### Nature of issue?

- [ ] Found a bug
- [x] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### New feature details:
This has been brought up before in #2180 and although while it is true that we can get the delta time of the last frame with `1000/frameRate()` it requires a bit of extra explanation about why that would give the delta time. Also most game engines also provides a delta time variable in their animation loop so implementing it here would help with moving from one to the other.

It also isn't that hard to implement because p5.js already has the property as [`now - this._lastFrameTime`](https://github.com/processing/p5.js/blob/25ba3ec4e76788fc58c4de9b5166d9300dd516a2/src/core/main.js#L375) (which is how `frameRate()` is derived anyway) we can just expose that as a public variable (named `deltaTime` or `lastFrameTime` maybe) and we're done.

limzykenneth:
Closes #3737 

Implement `deltaTime` as a p5 variable (not unlike `frameCount`). It is only updated when `draw()` is looping which is the same behaviour as `frameRate()`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "p5",
-  "version": "0.7.3",
+  "version": "0.8.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -438,7 +438,8 @@
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/aproba/-/aproba-1.2.0.tgz",
       "integrity": "sha512-Y9J6ZjXtoYh8RnXVCMOU/ttDmk1aBjunq9vO0ta5x85WDQiQfUF9sIPBITdbiiIVcBo03Hi3jMxigBtsddlXRw==",
-      "dev": true
+      "dev": true,
+      "optional": true
     },
     "archiver": {
       "version": "1.3.0",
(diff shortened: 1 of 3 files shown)


stalgiag:
Cool!

One thing that might be worth mentioning  is that both `millis()` and `_lastFrameTime` operate a little unexpectedly in conjunction with `preload`. I am working on something right now that is very timing-dependent and since `millis()` just points to `window.performance.now` there is a bit offness with preloading. Which is to say that I think there is a large delta jump on the first frame. I think this is due to `_lastFrameTime` being updated on construction and not updated again until after passing through `draw` once. This causes the time it takes to get through the preload to come through as the first `deltaTime`. 

This isn't a problem with this PR but just related.

stalgiag:
I put `this._lastFrameTime = window.performance.now();` in `_runIfPreloadsAreDone()` in my working branch to help with this

lmccart:
this is great, thanks @limzykenneth!

PR #3740, opened 2019-05-14
GIF support

### LINKED ISSUE #3380: add GIF support for image() function
Currently, animated .gif files loaded via `loadImage()` and displayed via `image()` will display as static images. A current workaround is to use the `createImg()` function, but this adds complexity and it would be nice for gifs to work with the core `image()` function.

stalgiag:
I mostly want to get thoughts on how it feels from the user side as I consider the code fairly incomplete and I haven't done refinement passes. The branch does pass all current unit tests and doesn't seem to break any manual tests fwiw. I am mostly interested in where expectations are met and where they aren't as per the discussion in #3380. 

[**There is an example on the editor here**](https://editor.p5js.org/stalgiag/sketches/KD9t-ovEN).

Bullet points on implementation:
- adds [omggif](https://github.com/deanm/omggif) library for lightweight decoding (and encoding in future)
- Adds `gifProperties` to p5.Image which is null when not a GIF and has [this structure ](https://github.com/processing/p5.js/compare/master...stalgiag:gifSupport?expand=1#diff-f5034d55fd652b7016c8def87cb18042R896)when it is
- The playback of the GIF is tied to 'realtime' where possible (ie GIF FPS 20, sketch FPS 10 you see every other frame)

closes #3380


src/image/loading_displaying.js
```diff
@@ -77,12 +77,35 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
     pImg.drawingContext.drawImage(img, 0, 0);
     pImg.modified = true;
 
-    if (typeof successCallback === 'function') {
-      successCallback(pImg);
+    //GIF
+    if (path.match(/\.[gif]+$/i)) {
+      var req = new XMLHttpRequest();
```
limzykenneth: `fetch` has support for [`arrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/API/Body/arrayBuffer) as well if you want a syntax more consistent with the loading functions.

src/image/loading_displaying.js
```diff
@@ -77,12 +77,35 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
     pImg.drawingContext.drawImage(img, 0, 0);
     pImg.modified = true;
 
-    if (typeof successCallback === 'function') {
-      successCallback(pImg);
+    //GIF
+    if (path.match(/\.[gif]+$/i)) {
```
limzykenneth: Technically a server can still serve an gif even if the URL doesn't end with `.gif`. I wonder if there's a better way to know if an image is an animated gif or not at this stage?

limzykenneth:
Looks good! I haven't gone through everything but I've left a couple line comments. Also are there support for saving something out as an animated gif directly? Not entirely necessary at this stage though.

src/image/loading_displaying.js
stalgiag: Cool thanks! I will definitely change that.

src/image/loading_displaying.js
stalgiag: Yeah I agree this isn't ideal. I spent a few moments fiddling around with other ideas but didn't know how to do this check without having to load every image even non-gifs as an array buffer. Can definitely give it another shot.

stalgiag:
Great thanks for checking it out! And yeah I don't think this is ready for a detailed review yet. Just trying to make sure the behavior and macro design is somewhere near desired and then I'll do the refactoring, optimization, and remaining features.

There isn't support for saving out an animated gif currently but the omggif library supports encoding so adding this shouldn't be too troublesome and I plan on doing it for this PR.

limzykenneth:
Cool, look forward to seeing more work on this!

src/image/p5.Image.js
```diff
@@ -828,4 +831,87 @@ p5.Image.prototype.save = function(filename, extension) {
   p5.prototype.saveCanvas(this.canvas, filename, extension);
 };
 
+p5.Image.prototype.play = function() {
+  if (this.gifProperties === null || this.gifProperties.playing === true) {
+    return;
+  } else {
+    this.gifProperties.playing = true;
+  }
```
lmccart: could this be simplified to:
```js
if (this.gifProperties) {
   this.gifProperties.playing = true;
}
```

src/image/p5.Image.js
```diff
@@ -828,4 +831,87 @@ p5.Image.prototype.save = function(filename, extension) {
...
+};
+
+p5.Image.prototype.pause = function() {
+  if (this.gifProperties === null || this.gifProperties.playing === false) {
+    return;
+  } else {
+    this.gifProperties.playing = false;
+  }
+};
```
lmccart: could this be simplified to:
```js
if (this.gifProperties) {
   this.gifProperties.playing = false;
}
```

src/image/p5.Image.js
```diff
@@ -828,4 +831,87 @@ p5.Image.prototype.save = function(filename, extension) {
...
+
+p5.Image.prototype.delay = function(d) {
+  if (this.gifProperties === null) {
+    console.error('This is not a gif');
```
lmccart: we generally try to avoid `console.error` to allow things to fail more gracefully. `console.log` should be fine

src/image/p5.Image.js
```diff
@@ -828,4 +831,87 @@ p5.Image.prototype.save = function(filename, extension) {
...
+    return;
```
lmccart: this line could be removed

src/image/p5.Image.js
```diff
@@ -828,4 +831,87 @@ p5.Image.prototype.save = function(filename, extension) {
...
+  //to repeat N times (1<N), should contain integer for loop number
+  var loopCount = gifReader.loopCount();
+  if (loopCount === null) {
+    loopCount = 1;
+  }
+
+  this.gifProperties = {
+    isGif: true,
+    displayIndex: 0,
+    delay: gifReader.frameInfo(0).delay * 10,
```
lmccart: what is the `* 10` for?

src/image/p5.Image.js
```diff
@@ -828,4 +831,87 @@ p5.Image.prototype.save = function(filename, extension) {
   p5.prototype.saveCanvas(this.canvas, filename, extension);
 };
```
lmccart: we'll need some documentation for play(), pause() and some addition to p5.Image() and loadImage() before this can be merged.

lmccart:
looking great! I think the addition of saving to a gif would be a big plus for p5, if it's possible.

src/image/p5.Image.js
stalgiag: Oh yeah I know. I still have a lot to do. I will change it out of draft status when I think it might be near merging status.

src/image/p5.Image.js
stalgiag: Most of these are outdated I think but yes console.error got it! I was going to just route this thru friendly error system for now.

stalgiag:
> looking great! I think the addition of saving to a gif would be a big plus for p5, if it's possible.

Yeah definitely! Thanks for looking over it. Something about the timing between your review and the added commits confused Github a bit but I think I get the gist of each comment, and the overall takeaway.

I have been working on saving for the past two days. Something about extracting the bytes that represent the palette, compressing those bytes into a rgb Uint16Array , and using those to re-encode the GIF is throwing me off. There are other approaches that could be taken but it would be nice to edit the state of the gifs, encode with the same library that is used to decode, and see those changes reflected in the downloaded file.  There isn't an example of anyone using this library to do this anywhere that I have been able to find (extract palette from file, work with file in RGBA, then re-encode back to GIF with same palette). The decoding works really nicely but the encoding isn't very flexible, and is missing a lot of features that would make it easier for people to modify GIFs and encode them later.

For those reasons, I am considering switching to a different library, [readwrite-gif](https://github.com/ksheedlo/readwrite-gif). The decoder is just the GIFReader code from omggif. So the interface wouldn't change. The encoder is much easier to work with however. It has a number of functions that we would have to implement  if we wanted GIF editing and resaving, the most important of which is palette analysis. That way we don't have to retain the palette in any way, and theoretically down someone could add frames with a different color and it would break the encode. I know that library isn't heavily used but I believe it would make a number of things easier.

stalgiag:
Scratch that last comment. I somehow didn't realize that I needed to build an array that had a relational value for each pixel to the palette. ie pixel 0, 4 = colorPalette[2]. Saving works now. I will handle the remaining tasks (testing, documentation, a few more helper methods, cleaning up) on Monday and Tuesday of next week.

stalgiag:
Okay! I wanted to get to unit testing done today but resizing turned out to be a bigger challenge than expected. 

But other than unit tests and a second pass on documentation/code cleanliness this is beginning to feel ready for review.

There are some things that I am intentionally leaving for future pull requests because this one is already becoming unwieldy. Currently the loading and saving and timing all seem to work well and fast. The code doesn't seem terribly obtrusive. In terms of difference in functionality between animated and non-animated p5.Image objects, I was able to get fairly close to parity. Here is rundown based on my testing.

- `mask()` - incorrect on first frame every time the gif changes its frame. This is due to timing. Can probably be an easy fix. Works otherwise, and this is mostly undetectable in my testing.
- `filter()` - same as above
- `resize()` - works well, also works with saving. The scaling is done with a nearest neighbor algorithm because the frames are stored as ImageData and it isn't possible to draw into a tempCanvas and scale the same way non-animated images do. Nearest-neighbor is chosen so as to avoid changing the palette size.
- `updatePixels()`, `loadPixels()` etc - works as far as I can tell in my testing. Also works with saving which is pretty fun. Need to add an overload to `updatePixels()` to have the option of specifying the frame to update, currently it just assumes that you mean the frame that is being displayed.
- `copy()` - works as expected
- `blend()` - works as expected

I am not sure that I will be able to get to the unit tests tomorrow, as I have other work to catch up on. But this should be in a mostly usable, testable, and readable state.

src/image/p5.Image.js
stalgiag: added comment explaining - basically gifs store delay in 1/100 of second so changing to millis here

stalgiag:
I think this is ready for review whenever you get a chance @lmccart and @limzykenneth.

This is pretty large and unwieldy so I wouldn't be surprised if I missed something. I think that it might be difficult to catch every potential problem at this point. Maybe it will make sense to mark it experimental in documentation for a while?

Also the unit test coverage isn't amazing at the moment. Namely I couldn't figure out a great way to test for timing. That may have to be a later todo since it seems pretty involved.

But the functionality from the previous comment still stands, as well as working with `texture` and WebGL.

src/image/p5.Image.js
```diff
@@ -220,6 +222,35 @@ p5.Image = function(width, height) {
...
 
+/**
+ * Helper function for animating GIF-based images with time
+ *
+ */
+p5.Image.prototype._animateGif = function(pInst) {
+  var props = this.gifProperties;
+  if (props.playing) {
+    //to be replaced with deltaTime
```
lmccart: I believe deltaTime has been merged now

src/image/p5.Image.js
```diff
@@ -825,7 +886,261 @@ p5.Image.prototype.isModified = function() {
...
+ *   let frame = gif.getCurrentFrame();
+ *   if (frame % 2 === 0) {
+ *     fill(149, 66, 244);
+ *   } else {
+ *     fill(106, 225, 140);
+ *   }
+ *   image(gif, 0, 0);
+ *   ellipse(frame * 10, height / 2, 40);
+ * }
+ * </code></div>
```
lmccart: I like this, but this example is a little complex to follow. Could you add one more that just displays the current frame number using the text() function overtop of the drawn gif?

src/image/p5.Image.js
```diff
@@ -825,7 +886,261 @@ p5.Image.prototype.isModified = function() {
...
+ * Gets the index for the frame that is currently visible in an animated GIF.
+ *
+ * @method getCurrentFrame
+ * @return {Number}       The index for the currently displaying frame in animated GIF
+ * @example
+ * <div><code>
+ * let gif;
+ *
+ * function preload() {
+ *   gif = loadImage('assets/animated-counting.gif');
```
lmccart: I'm wondering if this gif might actually be a little confusing. Especially in the case of the example addition I proposed. Maybe one that's a clearly looping scene animation would be a more relatable example asset?

src/image/p5.Image.js
```diff
@@ -825,7 +886,261 @@ p5.Image.prototype.isModified = function() {
...
+      );
+    }
+  }
+};
+
+/**
+ * Returns the number of frames in an animated GIF
+ *
+ * @method numFrames
+ * @return {Number}       The number of frames in the animated GIF
```
lmccart: missing example

limzykenneth:
Great work! I am currently a bit busy with some work so I can't review the code in detail but maybe some bugs will only surface when we put it out in the world. I have however manage to do some basic tests. Here's a few questions I have from the minimal tests I did.

The biggest problem I came across is that is there a particular requirement on the animated gif file loaded? From my tests, I am only able to get about 50% of the gifs I downloaded from different places working, the rest all give an error when attempting to load them: `IndexSizeError: Index or size is negative or greater than the allowed amount` from  p5.js:67373. I noticed all the gifs I downloaded from giphy.com are failing like this.

Also mentioned previously, if the URL for the gif resource doesn't end in `.gif` it won't be detected as animated gif and so won't be animated, are you able to find a workaround on that end?

stalgiag:
@limzykenneth Thanks for trying out other GIF's. I must've had a testing set that was offering more flexibility. I wasn't able to recreate the same error as you noted but I was able to recreate that most GIFs from giphy.com weren't working. My error occurred at ImageData construction, but hopefully the underlying problem was the same.

I made the frame array size allocation more accurate and this seemed to fix the problem for the 10 GIFs from giphy.com that I was testing. Sorry about that!

> Also mentioned previously, if the URL for the gif resource doesn't end in .gif it won't be detected as animated gif and so won't be animated, are you able to find a workaround on that end?

I can't think of one but perhaps I am thinking about this the wrong way. Are you saying that we should also offer support for Webp files or perhaps other file-types that have animation data that I am not aware of? Or are you suggesting that we find a way to detect if the resource is a gif even if it is passed without an extension or with a non-GIF extension? If you are asking about the latter, I have searched quite a bit but I don't know of a way to do this without fetching everything passed to `loadImage` as an arrayBuffer. The only way I know of to determine if the file is a GIF other than reading the URL is to look at the first three bytes.

Spongman:
I wonder if it’s possible, instead of using an `Image` element, to use `fetch` first to get a response, then look at the mime-type and then either using `createObjectUrl()` or `arrayBuffer()` on the response depending on the need... it might save a request, although it might also have CORS implications.

stalgiag:
@Spongman interesting suggestion! I can give that a try.

limzykenneth:
@stalgiag Yeah, the last commit fixed the issue with some gifs failing.

I was thinking about maybe fetching then checking MIME type as well but I haven't been able to look closer at how the requests are being made. If you can get that working it would be great.

stalgiag:
@limzykenneth and @Spongman thanks for the tips!

I looked into the suggested approach but this won't work as mimetype the browser determines the mimetype using the extension The browser request is doing something very similar to the regex that I am using currently. So if I look at the  headers for a fetched `loadImage(gifFileNoExtension)` I will have `application/octet-stream` as the mimetype. Similarly if I give that GIF file a `.png` extension, I will get a png mimetype.

It is probably fine to load everything into an array buffer and check it, but it is a not-insignificant amount of extra overhead on image loading. I am unsure of whether it is misleading to be able to load an animated GIF without the correct file extension as the browser operating without p5 will not do this (if you drag and drop a GIF file without a `.gif` extension into a browser it will reject it). I guess the precedent is that this will work with static images and p5? Currently the gif will load as a static image if it has an incorrect or missing extension.

It seems to me like our options are: 

Current behavior - only `.gif` files are loaded with animation properties. GIF files with incorrect or missing extensions will be static.

Support missing extensions - check mimetype on all files, if mimetype is `octet-stream` then load data and check first few bytes to determine if the file is animated or static

Support both missing and incorrect extensions - load file data on all files and check first few bytes to determine file type

limzykenneth:
I'm not sure how the request is structured but instead of looking at the MIME type of the returned file can't we just check the `Content-Type` response header of the response? If it is a gif it should have `image/gif` as the `Content-Type`. Anything that isn't labeled as such should be treated as not a gif, not sure if it applies to direct image link but I imagine it wouldn't he hard to work around(?)

The reason to get this working is not so much that someone didn't add the `.gif` extension to their gif file for some reason but rather to support an arbitrary URI endpoint responding with a gif image:

```javascript
router.get("/file", function(req, res, next){
    res.sendFile("/path/to/animation.gif");
});
```

If such an Express route is opened in a browser it will still be animated.

If checking the header still not workable then maybe just stick to current behaviour for now..

stalgiag:
I'm sorry for the confusion, when I said 'mimetype' in the past post I meant the `Content-Type' section of the `Header` which is a mimetype I believe. Bit confusing!

Anyways, I am retrieving it with 
```
var req = new Request(path, {
    method: 'GET',
    mode: 'cors'
  });
  fetch(path, req).then(function(response) {
    console.log(response.headers.get('content-type'));
  });
```
So we are talking about the same thing. And what I said above is still true, the browser is determining the content-type portion of the header by looking at the extension of the URL. Whatever wizardry is used to render the Image object once it has its `src` set regardless of incorrect content-type is not exposed to us in any meaningful way.

limzykenneth:
The `Content-Type` response header should be set by the responding server and not the browser. With the Express route I've given and the `fetch` code you provided, it does print `image/gif` in the console. Unless I'm misunderstanding something here.

stalgiag:
I am new to Express routes so I am not sure how to recreate what you are talking about. 

The way I am testing currently is with a local server serving a local file. With that setup, the above code will log `image/gif` when the file has a `.gif` extension and `application/octet-stream` when that same file does not have an extension.

limzykenneth:
When a webserver (Express or otherwise) serves an image it usually detects the `Content-Type` automatically with the file extension, unless it is manually defined. So if the server is asked to serve something that doesn't have a recognized extension it will serve it as `application/octet-stream`. But that's not what I think we need to solve, someone naming their gif `animation.gid` on the server I would judge as misnamed file and not something we need to support.

The Express route I've defined is basically when you navigate to `http://localhost:3000/file` it will send that image at `/path/to/animation.gif` to the requester, as long as this part has `.gif` the response will have the `Content-Type` header as `image/gif`, but the user endpoint (`http://localhost:3000/file`) doesn't have `.gif` as its ending so with path sniffing we cannot detect that it is a gif image although it is.

stalgiag:
Ah gotcha! Now I understand what you are trying to accomplish a little bit better. Thank you for the clarity.

I can make the changes easily but I'm not sure how to make a simple test. Any ideas?

limzykenneth:
If you can set up a webserver that can respond to a route with a gif then you are good to go. 

I've been checking out giphy's API but all the responses are JSON so no luck there but if you can find such an endpoint somewhere online, you can skip the set up webserver part.

Spongman:
i think the important thing here is that mime type is coming from the web server: the `.gif` extension isn't required - the url can end with whatever you want.

lmccart:
@stalgiag I think this test endpoint should work: http://someone-spotify.herokuapp.com/gif

stalgiag:
Thank you @lmccart 🙌

stalgiag:
So I made the updates so that loading that GIF via the url you posted works well. Now if a file has a 'image/gif' content-type it will load as an animated GIF.

That particular GIF file is really strange and made me realize a few things. Namely that I wasn't handling disposal codes. Some GIFs only write into a portion of the image with a frame. This disposal code support is different across browsers and there are quite a few possibilities to support. I think that is well out of the scope for this PR but I made everything work with that GIF and the most common disposal codes. I can open an issue for expansion and refinement in this area once this PR is merged.

That file also made me realize that there can be frame level delay. I did not add support for that. It is possible that we will want to in the future, but I am averaging the delay across the frames for now. This should work for 99% of GIFs. It shouldn't be that hard to change the delay timing so that each frame has a delay but that introduces a lot of confusing design challenges with users modifying the delay. For now I think it makes sense to operate with a uniform cross-frame delay and then make a decision about changing it once people use it a bit.

I started to make the example changes that you requested @lmccart  but I felt it would be too time-consuming at the moment to make a nice looping GIF from scratch, and I could not find a resources with decent ones in a size that is useful for reference examples. The only place I can find GIFs that I know for sure are creative commons is on Wikimedia Commons but they have very few. Does anyone know of another resource? If necessary, I can make GIFs from scratch but I won't be able to get to that till next week at the soonest.

limzykenneth:
Thanks for all the effort! This looks good, at least for now. There are a couple places that can probably use some more tweaking but I think at least in terms of public API it should be ok. I still can't do a full tests myself but I'm ok for this to be merged whenever someone else had a look at it.

stalgiag:
Cool! I agree that further tweaking needs to happen and I am happy to open up some issues once this is merged. Just might be easier to refine this in more easily reviewable chunks.

I also just wanted to drop a note that I found some good GIF files on creativecommons.org and I edited them to fit our needs. Now the reference examples have nicer, more relatable GIFs and are more clear as per @lmccart's request.

test/manual-test-examples/p5.Image/GIF/sketch.js
```diff
@@ -0,0 +1,42 @@
+let img0, img1;
+
+let layer0, layer1;
+
+function preload() {
+  img0 = loadImage('./assets/counting-gif@endlessLoop-1secDelay.gif');
+  img1 = loadImage('./assets/counting-gif@1xLoop-0.2secDelay.gif');
```
lmccart: this it nitpicky but can you rename these files to something a little simpler?

lmccart: disregard my last comment, just realized this is a test not a reference example!

lmccart:
@stalgiag this looks good to merge. but I think you said you wanted to merge after we make the release at the end of the month to give it some time in master before putting into a public release?

stalgiag:
Thanks for looking over it at @lmccart! Yeah my original intuition was to delay until after the 0.9.0 release so that the changes could live in master for a while before making it to a public release. There are still a few follow-up issues that I want to work on, such as handling the disposal codes better. I am also open to releasing with 0.9.0 and marking the features as 'experimental', but the first option seems the safest.

lmccart:
🎉🎉🎉

PR #3743, opened 2019-05-15
add a line of documentation about the P2D renderer

outofambit:
just adding a short description to the `P2D` constant to maintain some parity with #3732

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -11,6 +11,7 @@ var PI = Math.PI;
 module.exports = {
   // GRAPHICS RENDERER
   /**
+   * Denotes the default two-dimensional renderer.
    * @property {String} P2D
    * @final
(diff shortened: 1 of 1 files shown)


outofambit:
@maddyfisher does this seem ok to you? open to suggestions on better wording, etc!

src/core/constants.js
```diff
@@ -11,6 +11,7 @@ var PI = Math.PI;
 module.exports = {
   // GRAPHICS RENDERER
   /**
+   * Denotes the default two-dimensional renderer.
```
limzykenneth: Wording wise "denote" may be a bit difficult for people whose English is not their first language.

maddyfisher:
Looks great! I agree with the word changing for "denotes."

PR #3755, opened 2019-05-19
Removes break statements from lighting shaders

aferriss:
For whatever reason, the break statements within loops in the lighting shaders were causing the shaders to fail to execute. 

I amended the shaders to just check if current loop index was less than the light count in all instances that I could find. This also allowed me to reduce the amount of loops being called in the shader.

It might be worth doing some tests to see if this check is really necessary, it might be just as performant to avoid the conditional and do our light calculations all 8 times.

### DIFF
--- a/src/webgl/shaders/light.vert
+++ b/src/webgl/shaders/light.vert
@@ -38,41 +38,41 @@ void main(void){
   float diffuseFactor = 0.3;
 
   vec3 ambientLightFactor = vec3(0.0);
-
-  for (int i = 0; i < 8; i++) {
-    if (uAmbientLightCount == i) break;
-    ambientLightFactor += uAmbientColor[i];
-  }
-
-
   vec3 directionalLightFactor = vec3(0.0);
-
-  for (int j = 0; j < 8; j++) {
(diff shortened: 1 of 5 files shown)


stalgiag:
Strange! Thanks so much for figuring that out. Obviously I can't test for the bug but I believe you that this fix works.

I was able to do some performance testing and I couldn't detect any change. The code is more legible with a single for-loop.

Thanks again for the improvements!

Spongman:
is there any documentation anywhere on shaders not allowing `break;` in loops any more? I would be very interested to know if this is as simple as `break` being broken, or if there's another subtle interaction going on.

aferriss:
@Spongman I think that there is something subtle going on as you suggest because `break` statements seem to run fine on the newer iPhone hardware using the text rendering shader. It wasn't readily apparent to me what exactly was causing the issue, but removing the break statements from the lighting shaders _did_ allow the programs to compile and render.  

I think the extra empty loops and if checks are not ideal so if someone could figure out why the `break`'s were causing issues it'd be great to update them. On the plus side, making this change did allow me to combine the multiple loops into a single loop.

Issue #3760, opened 2019-05-21
webGL text rendering broken on Android 


aferriss:
Text rendering in webGL appears broken on Android in chrome, I'm willing to bet that it's broken on iOS / Safari as well, but haven't had a chance to check yet.

I noticed a number of things in the font.frag shader that look suspicious. 

1. Many mobile devices do not support the derivatives extension. It looks like it's only being used as a hack with `fwidth()` to determine texel size which can be calculated easily as` 1.0 / vec2(width, height)`, and so it is probably better for us to pre-calculate this value on the cpu and pass it to the shader as a uniform. 

2. Some Android devices will fail to compile shaders if non-const values are declared outside of a function block. This shader declares the following in the global scope of the shader, so we should move these into main, and adjust any functions that use them accordingly. 

```
vec2 pixelScale;
vec2 coverage = vec2(0.0);
vec2 weight = vec2(0.5);
```

For example this will fail to compile
```
vec4 test = vec4(0.0);

void main(){
  gl_FragColor = test;
}
```

Whereas this will work fine
```
void main(){
  vec4 test = vec4(0.0);
  gl_FragColor = test;
}
```

3. This shader also uses `break;`  statements inside of a loop ( https://github.com/processing/p5.js/blob/master/src/webgl/shaders/font.frag#L170 and https://github.com/processing/p5.js/blob/master/src/webgl/shaders/font.frag#L198), which might cause problems on iOS devices (see: https://github.com/processing/p5.js/commit/994d718ee1fe6f7a98842687cea701062cec6790). We should try and remove these if possible. 

4. webGL reported a couple different errors on my Android device somewhat inconsistently so there maybe something mixed up in the actual cpu / gl side of things as well. I saw at least one error complaining about a location property being undefined, as well as another error being thrown at drawElements().

5. There are some odd choices with how the code is structured stylistically. In particular are these empty curly bracket blocks at https://github.com/processing/p5.js/blob/master/src/webgl/shaders/font.frag#L159 and https://github.com/processing/p5.js/blob/master/src/webgl/shaders/font.frag#L190. I doubt this is the root cause of the rendering bug but I think they could be removed to make them in line with how some of the other shaders are written. It looks like all they're doing is visually separating blocks of code, so we could just include comments instead. 

[...]

PR #3764, opened 2019-05-23
Fixes webGL text() rendering on android 

aferriss:
Addresses the bugs in https://github.com/processing/p5.js/issues/3760

I think something is still wrong in way that attributes are created and set for text in webGL mode because this fix is more of a hack than a proper solution. Essentially what I'm doing is rebuilding the uv's right before they are bound to the shader. Previously, something was causing gl to lose track of the texcoord buffer, which would cause the drawElements error. 

I also moved some of the variables in the font shader out of the global scope and updated the functions that used them to match. 

I left the derivatives in place since they don't seem to be affecting my phone but I feel that it's worth thinking about changing in the future given the spotty support on mobile hardware.

### DIFF
--- a/src/webgl/shaders/font.frag
+++ b/src/webgl/shaders/font.frag
@@ -57,9 +57,6 @@ int getInt16(vec2 v) {
   return iv.x * INT(128) + iv.y;
 }
 
-vec2 pixelScale;
-vec2 coverage = vec2(0.0);
-vec2 weight = vec2(0.5);
 const float minDistance = 1.0/8192.0;
 const float hardness = 1.05; // amount of antialias
 
@@ -75,7 +72,7 @@ vec4 getTexel(sampler2D sampler, int pos, ivec2 size) {
   return texture2D(sampler, (vec2(x, y) + 0.5) / vec2(size));
 }
 
-void calulateCrossings(vec2 p0, vec2 p1, vec2 p2, out vec2 C1, out vec2 C2) {
+void calulateCrossings(vec2 p0, vec2 p1, vec2 p2, out vec2 C1, out vec2 C2, vec2 pixelScale) {
 
   // get the coefficients of the quadratic in t
   vec2 a = p0 - p1 * 2.0 + p2;
@@ -99,10 +96,10 @@ void calulateCrossings(vec2 p0, vec2 p1, vec2 p2, out vec2 C1, out vec2 C2) {
   C2 = ((a * t2 - b * 2.0) * t2 + c) * pixelScale;
 }
 
-void coverageX(vec2 p0, vec2 p1, vec2 p2) {
+void coverageX(vec2 p0, vec2 p1, vec2 p2, inout vec2 coverage, inout vec2 weight, vec2 pixelScale) {
 
   vec2 C1, C2;
-  calulateCrossings(p0, p1, p2, C1, C2);
+  calulateCrossings(p0, p1, p2, C1, C2, pixelScale);
 
   // determine on which side of the x-axis the points lie
   bool y0 = p0.y > vTexCoord.y;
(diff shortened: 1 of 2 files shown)


Spongman:
I don’t think there’s an easy way to remove the derivatives. Those values are not constant unless the glyph is parallel to the screen.

aferriss:
@Spongman Agreed, but perhaps there's another anti-aliasing solution that doesn't require derivatives? Maybe it's not worth worrying about until someone reports an issue.

Spongman:
yeah, i did check the support before using it: https://webglstats.com/webgl/extension/OES_standard_derivatives

Spongman:
i think it would be better to separate different sets of changes into different PRs. this PR seems to contain two unrelated changes:
1) the shader variable renaming. is there a reason for this change, or is it just a style change?
2) uv rebinding. can you provide a repro case for this issue?

aferriss:
@Spongman 

They're related in the sense that they are both intended to make font rendering more bulletproof on Android devices.

The variables in the shader aren't renamed, just moved out of the global scope of the shader. I've noticed that on some Android devices, placing non-const variables in the global scope can cause the shader to fail to compile. I think it's safer to just move them to the main function and pass them in to any shader that needs them.

For the other change with rebinding uvs, you can try running the simple text example (the star wars one) on am Android and you'll see it throw an error at draw elements. If you follow the error you'll find that webGL complains that the texture coordinate buffer has been enabled but not bound. 

There's probably a better fix for this, since as I mentioned above we shouldn't need to rebind them, but I couldn't determine what was causing my device to lose track of the buffer hence the explicit rebinding.

Spongman:
can you try, in the original code, asserting `g.uvBuffer && gl.isBuffer(g.uvBuffer)` ? before the `this._bindBuffer` calls?

Spongman:
are you able to determine if this is caused by the buffer having not been initialized yet, or if the buffer was initialized but has since been lost for some reason?

aferriss:
@Spongman Didn't have a chance to look at this again until now. I believe that the uv buffer does indeed get initialized in createBuffers, but as you suggest, get's lost somewhere between the creation there and when it gets bound in the text renderer. `g.uvBuffer` is undefined on my android when the renderer tries to bind it, which is what causes the rendering to fail.

Spongman:
ok, i think i understand what the problem is, and this is going to require a bigger fix than just the text stuff, it affects all retained-mode rendering.

the issue is that when creating a retained-mode geometry (any geometry, not just the glyph quad), it's checking the _current_ fill and stroke shaders for which buffers to create and which attributes to enable.

https://github.com/processing/p5.js/blob/6640d149348dea0a1b3df61f0cf8442222403a9f/src/webgl/p5.RendererGL.Retained.js#L171-L172

this means that if you choose a shader that doesn't have a particular attribute, then create a (cached) geometry, that geometry won't have a buffer created for it that corresponds to that attribute. then, if you switch the current shader to one that _does_ have that attribute, it will fail to render because that buffer won't have been created.

i have a branch (https://github.com/processing/p5.js/issues/3620) with some code that removes the shader stuff from the geometry code and delays the buffer creation & attribute binding until just before rendering. i will attempt to rebase that onto master. it might be useful to know if that fixes this buffer issue.

Spongman:
any chance you can try this pen: https://codepen.io/Spongman/pen/QBjQoJ on your android devices?

it's using a build of p5 from my [render-buffers](https://github.com/Spongman/p5.js/tree/render-buffers) branch, compiled p5.js is [here](https://ci.appveyor.com/api/buildjobs/ao7fmg3h5fua0y93/artifacts/lib%2Fp5.js).

aferriss:
@Spongman Your branch and the text example works on my device!

stalgiag:
@Spongman works on my device as well!

@aferriss if you don't mind I think it might make sense to review and possibly merge #3769 first since it seems to address an underlying problem that would fix the uv buffer issue. Afterwards we can make a decision about moving the shader variables out of the global scope. How does that sound?

aferriss:
@stalgiag That sounds like a plan. This PR was only resolving a side effect of a larger problem so if https://github.com/processing/p5.js/pull/3769 fixes that then I'm more than happy to close this.

Spongman:
i do still think it's important to track down the `break;` issues you were seeing, though. for example, the font shader relies on being able to break out of a loop up to 250 bezier strokes in each axis. for most glyphs it's significantly less than that, but more intricate script fonts can approach that. i don't see a significant performance drop on my desktop GPU when forcing it to loop through all iterations, but on different hardware  doing so may cause a noticeable performance loss even on simple fonts.

aferriss:
@Spongman I think the `break` issue was not a problem in the font rendering shader. I tried it on an iOS device and it rendered fine. 

Only the lighting shaders were causing problems when using break. My guess is that it's something related to the non-uniform control flow that I mentioned in https://github.com/processing/p5.js/pull/3755#issuecomment-496104370 but I haven't had time to go take another look. I think that instance, we're only looping 8 times so the performance hit should be relatively negligible.

PR #3772, opened 2019-05-29
core module: add unit tests

ihsavru:


lmccart:
@outofambit this looks good to me if you want to merge it in

test/unit/core/vertex.js
```diff
@@ -0,0 +1,135 @@
...
+      });
+    });
+  });
+
+  suite('p5.prototype.vertex', function() {
+    test('should be a function', function() {
+      assert.ok(myp5.vertex);
+      assert.typeOf(myp5.vertex, 'function');
+    });
+    /*test('missing param #1', function() {
```
outofambit: @ihsavru should this commented out test be removed? or is this still in progress? i'm pretty sure we don't need these two cases, but i might be wrong

ihsavru: I commented those out because vertex doesn't validate it's parameters. So, those tests are failing

outofambit: ah, got it! lets use `test.skip` instead for those instead of commenting them out. (here's the docs and an example for that https://mochajs.org/#inclusive-tests)

Could we also add a code comment before the tests saying why they are skipped?

PR #3775, opened 2019-05-30
unify lighting

### LINKED ISSUE #3774: webgl: gouraud & per-pixel shading are out of sync
0.8.0

it looks like the lighting rules in the two different fill shaders (gouraud and per-pixel) have diverged such that the results are different.

for example, here's the perPixelLighting manual test example side-by-side:

image

differences:
- the per-pixel version is darker,
- the gouraud version is missing the specular highlight from the directional light,
- the per-pixel specular highlights are the wrong color.

Spongman:
closes #3774

this splits out the lighting code into a separate file `lighting.glsl` that is shared between both the gouraud fill shader and the per-pixel shader. it fixes the bugs in that issue and removes the differences highlighted there. it also makes it easier to implement changes to the lighting code in the future, by having a single location for that code.

here's the side-by-side comparison again:
image

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -9,6 +9,11 @@ require('../core/p5.Renderer');
 require('./p5.Matrix');
 var fs = require('fs');
 
+var lightingShader = fs.readFileSync(
+  __dirname + '/shaders/lighting.glsl',
+  'utf-8'
+);
+
 var defaultShaders = {
   immediateVert: fs.readFileSync(
     __dirname + '/shaders/immediate.vert',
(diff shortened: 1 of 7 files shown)


stalgiag:
@Spongman I agree with the fact that some unification and clarity (along with improved function) is necessary with the lighting in the GL Renderer. The most important thing that I want to bring up is that @sanketsingh24 (along with mentor @aferris) will be working pretty extensively on the lighting this summer as part of Google Summer of Code. I think that they should have a role in any kind of redesign decisions that happen with the lighting at this time.

This is the area that I have spent the least time in as GL renderer goes, so my ability to help with high level decisions is a bit hampered here. I have felt for a while that the way that lighting is handled on the CPU side feels a bit outdated relative to the rest of the GL code. I mean this in that it still uses much of the same architecture from the first design of the GL Renderer. This is less related to this specific PR but I do want to advocate for more encapsulation and unification for the lighting code across the board. I think that the lighting state could stand to be a single object held by the renderer rather than a bunch of arrays. This might open up new ideas about shader design as well. Just putting that out there as a parallel concern to what is happening here.

Spongman:
ok, sounds good. just want to add that the `lighting.glsl` stuff isn't really a redesign, just a copy of the existing code.

i would recommend merging this before doing any additional feature-work on the lighting stuff as doing so will mean that the code only needs to be done once, and doesn't need to be tweaked to fit into the two different instances that we have in the gouraud/per-pixel shader pairs.

aferriss:
This looks good to me. I spoke with @sanketsingh24 and we both feel that this is a good change and also makes it easier to work on the lighting shaders going forward. As long as this doesn't break anything I think we can go ahead and merge.

Spongman:
great, thanks for reviewing!

Spongman:
...and silliness ensues...

https://codepen.io/Spongman/full/JqwBRz

image

stalgiag:
Great thanks for chiming in @aferriss. 

I tested on both desktop and mobile and everything works well. There is a [bug with the textureMode manual test and reference example](https://github.com/processing/p5.js/issues/3784) but that is occurring in master currently as well so unrelated to this PR.

I feel good about merging this but just curious if you have any thought on why the BufferDef work is showing up in the diff comparison with this PR? I don't think it'll matter but seems perhaps related to the force-update approach taken to bring this up to master? Not sure! Just wanted to point it out.

Spongman:
@stalgiag  oops, i think i must have messed up a merge somewhere along the line.  thanks for finding that. i'll fix shortly.

stalgiag:
Thanks @Spongman!

PR #3778, opened 2019-06-01
use 'highp int' in font shader

### LINKED ISSUE #3776: webgl: mobile glyph rendering issues
0.8.0
re: #3769 

@stalgiag reported webgl font rendering issues on android mobile devices:

> The text/chinese manual test is still rendering oddly on my mobile device (Android 8, Galaxy s7). I don't know if that has to be addressed in this PR as it is "more broken" in master currently. Just thought it worth noting. Here is a screenshot:

it looks like it might be due to integer precision issues on that device:
> If I change the fragment shader to use high precision floats and also do the integer substitution with floats (ie #if 1 in line 4) then it displays correctly.

Spongman:
closes #3776

increases the precision of int values in the font shader in order to fix the rendering issues on some mobile devices.

### DIFF
--- a/src/webgl/shaders/font.frag
+++ b/src/webgl/shaders/font.frag
@@ -12,7 +12,7 @@ precision mediump float;
 
 #else
   // use native integer math
-	precision mediump int;
+	precision highp int;
 	#define INT(x) x
 
 	int ifloor(float v) { return int(v); }


stalgiag:
Great - thanks for the remote remote debugging!

PR #3783, opened 2019-06-02
Remove .vscode/settings.json

outofambit:
@stalgiag @Spongman can you try this out and see if it fixes the issues you've been having with VSCode since #3714 ? (I still haven't been able to reproduce locally, so my best guess is that there's some conflict between your user settings and the workspace settings. This should resolve that.)

Once we confirm this fixes the issue, i'll re-add `.gitignore` for this file so that we don't commit it accidentally in the future.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ dist/
 p5.zip
 bower-repo/
 p5-website/
+.vscode/settings.json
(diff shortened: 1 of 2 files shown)


stalgiag:
@outofambit Yes! Switching to this branch and reloading VSCode fixes the problem.

PR #3786, opened 2019-06-02
Feature Implementation: lightFalloff()

sanketsingh24:
Implemented lightFalloff

Work done- 
- [x] Implemented falloff in both shaders
- [x] Implemented `lightFalloff()`
- [x] Added a manuel-test-example in /webgl/lights/fallOff
- [x] Unit tests
- [x] Documentation
- [x] Example

To do-

@aferriss 

PS: I will have to do some changes in shader if [unify lighting PR](https://github.com/processing/p5.js/pull/3775) gets merged, but those changes will be small and so I hope that merges soon.


Spongman:
this all looks good. however, i wonder if it's necessary to scale the coefficients so the falloff scales with viewport size?

the old falloff function i had in the per-pixel shader previous was also dependent on viewport size, and if you made the canvas larger/smaller it would make the objects brighter/dimmer.

src/webgl/light.js
```diff
@@ -315,4 +315,35 @@ p5.prototype.lights = function() {
   return this;
 };
 
+/**
+ * @method fallOff
+ * @chainable
+ */
+p5.prototype.fallOff = function(constAtt, linearAtt, quadAtt) {
```
aferriss: Can we name these variables `constantAttenuation, linearAttenuation, quadraticAttenuation` (and also their same properties attached to the renderer object)? I know they're a little on the long side but I think will improve readability and remove any confusion about their purpose especially with `const`.

src/webgl/shaders/light.vert
```diff
@@ -17,6 +17,9 @@ uniform vec3 uPointLightLocation[8];
 uniform vec3 uPointLightColor[8];
 uniform bool uSpecular;
 uniform float uShininess;
+uniform float constAtt;
```
aferriss: I think these should also be `uConstantAttenuation, uLinearAttenuation, uQuadraticAttenuation` prepended with the `u` to keep them consistent with other shader uniforms.

src/webgl/shaders/light.vert
```diff
@@ -59,17 +62,21 @@ void main(void){
     if(i < uPointLightCount){
       vec3 loc = (uViewMatrix * vec4(uPointLightLocation[i], 1.0)).xyz;
       vec3 lightDirection = normalize(loc - mvPosition.xyz);
+      vec3 lightVector = mvPosition.xyz - loc;
 
       float directionalLightWeighting = max(dot(vertexNormal, lightDirection), 0.0);
 
+      float lightDistance = length(lightVector);
+      float fallOffFactor =  1.0 / (constAtt + lightDistance * (linearAtt + quadAtt * lightDistance));
```
aferriss: Is this the correct formula? In your proposal it is written as 
```
  float attenuation = 1.0 / (constantAttenuation + lightDistance * linearAttenuation + (lightDistance * lightDistance) * quadraticAttenuation);
```

Just want to double check.

src/webgl/light.js
```diff
@@ -315,4 +315,35 @@ p5.prototype.lights = function() {
...
+
+  if (linearAtt < 0) {
+    linearAtt = 0;
+  }
+
+  if (quadAtt < 0) {
+    quadAtt = 0;
+  }
+
+  if (constAtt === 0 && (linearAtt === 0 && quadAtt === 0)) {
```
aferriss: Are you doing this to prevent the attenuation from causing divide by zero problems in the shader? Maybe it should be made clear to the user why they cannot set all falloff values to zero by throwing a warning in the console letting them know that constantAtt has been set to 1.

aferriss:
@Spongman I'm also of the opinion that the lighting should appear consistent as the viewport scales. Are you suggesting that we multiply the attenuation values by viewport scale to fix the inconsistent rendering at different window sizes?

src/webgl/shaders/light.vert
Spongman: those two formulas are equivalent. i think the 2nd one is more readable, though. i don't think the optimization is necessary here.

Spongman:
@aferriss yeah something like that. probably not directly by the viewport size, but something that scales the lighting to a standard scale.

something like this, maybe?
```javascript
var viewportSize = p5.Vector.mag(viewport.width, viewport.height); // length of diagonal
var standardViewportSize = 800; // tunable constant
var viewportScale = viewportSize / standardViewportSize; // fraction of standard size

var scaledLinearAttenuation = linearAttenuation / viewportScale;
var scaledQuadraticAttenuation = quadraticAttenuation / (viewportScale * viewportScale);
```

i think the quadratic attenuation uniform needs to be scaled quadratically to keep the resultant scale of `lightDistance` linear.

also, the attenuations might need to be _multiplied_ by `viewportScale` instead of _divided_, it's late and i can't get my head around all the reciprocals right now ;-)

(this _could_ all be done by adding another 'scale' uniform to the shader and scaling the `lightDistance` in there, but i think scaling the attenuation constants in javascript is the better choice.)

src/webgl/shaders/light.vert
sanketsingh24: Yes they are equivalent. I'll rewrite the formula and use the second one though.

src/webgl/light.js
sanketsingh24: I'll log the console.warn with every check.

sanketsingh24:
@Spongman I tried to change the viewport and but it didn't change the lights. Can you please provide the code so I can debug?

aferriss:
@sanketsingh24 @Spongman I think this example demonstrates the effect. The attenuation of the lighting will change as you resize the window. 
https://editor.p5js.org/aferriss/sketches/hZBmLto4_

Spongman:
‘falloff’ is a word meaning the amount that something decreases. ‘fallOff’ implies two words.

Also can we call the method `lightFalloff()` to match processing?

sanketsingh24:
Also, I have a doubt, should `ambientLight()` be also affected by attenuation? On the internet, there's no given rule regarding the above, and some people apply it and some don't, but processing does apply it.

Although the logic to apply falloff to ambient light seems a bit odd to me.

Spongman:
yeah, processing has this somewhat unusual concept of positional ambient lights whose light follows the falloff rules, but not the diffuse or specular lighting rules. it doesn't really map to a physical lighting model, but i guess it could be a useful hack in some circumstances. i would recommend not bothering with it, at least for this PR. maybe in the future if someone asks for it...

aferriss:
@stalgiag Any chance you have time to take a look at this? I think it is good to go.

stalgiag:
Cool! Just ran through it and it looks really great. I feel good about merging this. The only change that comes to mind is possibly adding a click that changes the values being passed to `lightFalloff` in the reference example? This might make the example too complicated but I like how in the manual test I am able to more clearly see the range of possible effects that can be achieved with this function.

Also just an aside @aferriss, I am not sure but I think that you should have merge authority. I am happy to be a second set of eyes whenever you'd like but you are also free to merge Sanket's pull requests. If it makes more sense this can happen into a branch like 'processing/p5.js:lightingGSOC19' but also doesn't need to.

I feel good about this one. @sanketsingh24 feel free to make your own choice about whether to modify the reference example, and @aferriss you should feel free to merge this whenever you'd like.

Awesome work everyone!

aferriss:
@stalgiag Thanks for reviewing! I just checked and I don't have write access to the repo. Who should I talk to about getting that sorted?

lmccart:
@aferriss just sent you an invite to the p5.js team which will give you write access!

sanketsingh24:
@stalgiag I can make an example similar to [`shininess()`](https://p5js.org/reference/#/p5/shininess), Is this good?

stalgiag:
@sanketsingh24 yes that would work well

PR #3796, opened 2019-06-03
Add functions to help work with localStorage

### LINKED ISSUE #3703: Discussion on functions for working with local storage
#### New feature details:

Hello all just wanted to post an issue to track that I am thinking through adding `local_storage.js` for working with [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage). Here are some notes as I start to plan:

- `setItem()` - (could also be named `store()`)
  Not sure what to do here. The options aren't great based on my research. Listing those that have occurred to me (assuming `var myVar`):
  - `setItem('myVar', myVar)`
   This is the most straightforward option for implementation but has smallest benefit in terms of masked complexity. Only benefit is removing the need for 'localStorage.' in front of the method.
  - `setItem('myVar')`
    I don't think this is possible. The only way I can find to do this would require the use of `.callee` which isn't an option. The trouble lies with getting from a string variable name back to the variable value without being able to assume that the variable is global.
  - `setItem({myVar})`
    It is easy enough to make this work and it reduces complexity in some ways but it would likely be a bit confusing for many people.
  - `setItem(myVar)`
    This would be the best but I don't think it is possible.

- `getItem()`
- `removeItem()`
  - these have similar unknowns as `setItem()` but passing in string for variable name would work fine.

- `loadSketch()`
- `saveSketch()`
  - haven't looked into this but the idea is that this could save/load all of the global variables. The question here is whether this makes sense to make available.

- `clear()`
  - straightforward
- `getStorageSize()`
  - Not sure if this is nice to have or not. Would be a pretty heavy synchronous calculation, but each user potentially has a different available storage size.

Any thoughts and discussion are welcome. Thanks!

stalgiag:
This PR is a first pass at adding some [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) features into p5.

This currently adds:
```
storeItem(String - key, String|Number|Object|Boolean|p5.Color - value)
    nothing returned

loadItem(String - key)
    returns String|Number|Object|Boolean|p5.Color

removeItem(String - key)
    nothing returned

clearStorage()
    nothing returned
```

Window.localStorage only accepts string values. So the main benefit of these functions is that they handle mutation of values into strings for storage and back into original data type upon retrieval. This is done by storing two values whenever `storeItem()` is called. The first is the actual data is the passed key/value pair with value in string form:
```
key: key
value: value (as string)
```
 the second is formatted as
```
key: passedKey + 'Type'
value: typeof value
```
This way the value can be changed back into original type before returning in `loadItem()`. I decided to store these into two values  separately rather than simply adding the type info into the first stored item so that people won't run into trouble if they mix and match working with Window.localStorage and these p5 functions.

This still needs unit tests. But just wanted to get opinions and a second set of eyes looking for oversights in case I missed something. No rush since this isn't urgent but would you mind looking over this when you get a chance @outofambit?

closes #3703

### DIFF
--- a/src/data/local_storage.js
+++ b/src/data/local_storage.js
@@ -64,6 +64,9 @@ p5.prototype.storeItem = function(key, value) {
       value = value.toString();
       break;
     case 'object':
+      if (value instanceof p5.Color) {
+        type = 'p5.Color';
+      }
       value = JSON.stringify(value);
       break;
     case 'string':
@@ -138,10 +141,10 @@ p5.prototype.getItem = function(key) {
         break;
       case 'object':
         value = JSON.parse(value);
-        //If an object is meant to be a p5.Color
-        if (typeof value.maxes !== 'undefined') {
-          value = this.color.apply(this, value.levels);
-        }
+        break;
+      case 'p5.Color':
+        value = JSON.parse(value);
+        value = this.color.apply(this, value.levels);
         break;
       case 'string':
       default:
--- a/src/data/local_storage.js
+++ b/src/data/local_storage.js
@@ -10,9 +10,15 @@
 var p5 = require('../core/main');
 /**
  *
- * Stores a value in local storage under the key name. The key can
- * be the name of the variable but it does not have to be. To retrieve
- * stored items see <a href="#/p5/getItem">getItem</a>.
+ * Stores a value in local storage under the key name.
+ * Local storage is saved in the browser and persists
+ * between browsing sessions and page reloads.
+ * The key can be the name of the variable but doesn't
+ * have to be. To retrieve stored items
+ * see <a href="#/p5/getItem">getItem</a>.
+ * <br><br>
+ * Sensitive data such as passwords or personal information
+ * should not be stored in local storage.
  *
  * @method storeItem
  * @for p5
--- a/src/data/local_storage.js
+++ b/src/data/local_storage.js
@@ -54,7 +54,7 @@ var p5 = require('../core/main');
  *
  */
 p5.prototype.storeItem = function(key, value) {
-  if (value === 'undefined') {
+  if (typeof value === 'undefined') {
     console.log('You cannot store undefined variables using storeItem()');
   }
   var type = typeof value;
@@ -122,7 +122,7 @@ p5.prototype.storeItem = function(key, value) {
 p5.prototype.loadItem = function(key) {
   var value = localStorage.getItem(key);
   var type = localStorage.getItem(key + 'Type');
-  if (type === undefined) {
+  if (typeof type === 'undefined') {
     console.log(
       'Unable to determine type of item stored under ' +
         key +
@@ -139,7 +139,7 @@ p5.prototype.loadItem = function(key) {
       case 'object':
         value = JSON.parse(value);
         //If an object is meant to be a p5.Color
-        if (value.maxes.hsb !== undefined) {
+        if (typeof value.maxes.hsb !== 'undefined') {
           value = this.color.apply(this, value.levels);
         }
         break;
--- a/src/data/local_storage.js
+++ b/src/data/local_storage.js
@@ -72,7 +72,7 @@ p5.prototype.storeItem = function(key, value) {
   }
 
   localStorage.setItem(key, value);
-  var typeKey = key + 'Type';
+  var typeKey = key + 'p5TypeID';
   localStorage.setItem(typeKey, type);
 };
 
@@ -121,7 +121,7 @@ p5.prototype.storeItem = function(key, value) {
  */
 p5.prototype.getItem = function(key) {
   var value = localStorage.getItem(key);
-  var type = localStorage.getItem(key + 'Type');
+  var type = localStorage.getItem(key + 'p5TypeID');
   if (typeof type === 'undefined') {
     console.log(
       'Unable to determine type of item stored under ' +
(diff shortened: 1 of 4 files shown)


src/data/local_storage.js
```diff
@@ -0,0 +1,212 @@
...
+    case 'object':
+      value = JSON.stringify(value);
+      break;
+    case 'string':
+    default:
+      break;
+  }
+
+  localStorage.setItem(key, value);
+  var typeKey = key + 'Type';
```
lmccart: I would suggest making this a little more likely to be unique, so there's no chance of it being accidentally overwritten by someone calling `storeItem()` with "Type" as part of the key name. I know this isn't super likely, but it's possible.

Maybe its just a matter of modifying this to "[Type]" or something similar.

lmccart:
This looks great. My one concern is about the `loadItem()` naming. Right now, all `loadX` methods work with preload. This one wouldn't as it is now. We could make it work of course, but it feels arbitrary because we're not loading anything from outside the program. Maybe another way to say this is, right now, "load" in a function name refers to loading something from outside. So maybe something like `getItem()` might make more sense?

stalgiag:
I agree! Thanks for checking it out. I'll change that function name on Monday.

src/data/local_storage.js
```diff
@@ -0,0 +1,212 @@
...
+ * }
+ * </code></div>
+ *
+ * @alt
+ * When you type the key name is displayed as black text on white background.
+ * If you reload the page, the last letter typed is still displaying.
+ *
+ */
+p5.prototype.storeItem = function(key, value) {
+  if (value === 'undefined') {
```
outofambit: i think this check would need to be `value === undefined` to work? (aka no quotations)

src/data/local_storage.js
```diff
@@ -0,0 +1,212 @@
...
+    console.log('You cannot store undefined variables using storeItem()');
```
outofambit: you might be already thinking this, but i think it would be awesome to use the friendly error system for these messages!

src/data/local_storage.js
```diff
@@ -0,0 +1,212 @@
...
+ *
+ */
+p5.prototype.loadItem = function(key) {
+  var value = localStorage.getItem(key);
+  var type = localStorage.getItem(key + 'Type');
+  if (type === undefined) {
+    console.log(
+      'Unable to determine type of item stored under ' +
+        key +
+        'in local storage. Did you save the item with something other than setItem()?'
```
outofambit: i _love_ this messaging!

src/data/local_storage.js
outofambit: i wonder if we should prefix all items and typeItems stored by p5 with something like `[p5-ls]` so it's harder to get accidental collisions too? that might be overkill though.

src/data/local_storage.js
```diff
@@ -0,0 +1,212 @@
...
+      value = JSON.stringify(value);
+      break;
+    case 'string':
+    default:
+      break;
+  }
+
+  localStorage.setItem(key, value);
+  var typeKey = key + 'Type';
+  localStorage.setItem(typeKey, type);
```
outofambit: i think storing the type separately in `localStorage` is a great idea!

src/data/local_storage.js
lmccart: or `if (typeof value === 'undefined') {` might be most clear

src/data/local_storage.js
stalgiag: Yeah I wasn't sure. I didn't have strong feelings either way when first writing this. The impulse to leave the key unchanged was that I thought somebody may try mixing and matching using the localStorage object directly and p5 methods and may become quite confused as to why there is nothing stored at the expected key. That said, maybe we shouldn't allow that because then the lack of type handling would potentially be confusing? Or maybe these are all edge cases, and there is no way to account for every kind of use. Not sure!

src/data/local_storage.js
stalgiag: I do think this could be nice. I am decidedly uncomfortable working with the FES code, but I should do it anyways. I should at least check the parameters with FES. I find that when I am logging a single very specific message that using the FES actually makes things much harder. But in that situation I guess the benefit is that people can turn it off?

src/data/local_storage.js
```diff
@@ -0,0 +1,215 @@
...
+ * @requires core
+ *
+ * This module defines the p5 methods for working with local storage
+ */
+
+'use strict';
+var p5 = require('../core/main');
+/**
+ *
+ * Stores a value in local storage under the key name. The key can
```
lmccart: can you add one more sentence here explaining what local storage is? for example:
"Local storage is saved in the browser and persists between browsing sessions and page reloads."
I wonder if we also should include some kind of security warning. Local storage is pretty insecure so users should not store sensitive info there. Or maybe there is a link we can link to with more info about local storage?

lmccart:
@stalgiag I had one comment about documentation, then I think it's set to merge!

src/data/local_storage.js
stalgiag: Good idea!

 I couldn't find a good link for security stuff. I found a number of pages where folks talk about how insecure local storage is but everything I found was not necessarily appropriate for beginners and/or had information that would be confusing relative to this API for working with it (such as statements like 'localStorage is terrible because it can only store strings despite being a giant JavaScript object'). I don't want to accidentally lead folks down a rabbit-hole of suggestions about cryptographic-signed cookies.

I think your suggestion plus a simple statement about sensitive information should suffice.
"Local storage is saved in the browser and persists between browsing sessions and page reloads. Sensitive data such as passwords or personal information should not be stored in local storage."

lmccart: I think that sounds great

Issue #3803, opened 2019-06-09
Pointer Lock API


jonnytest1:
#### Nature of issue?

- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ x] Core/Environment/Rendering
- [x ] Utilities

#### Which platform were you using when you encountered this?

- [ x] Desktop/Laptop

#### New feature details:
I found this API which removes the cursor from the website and thought this would be a fitting 
addon to p5
demo: https://mdn.github.io/dom-examples/pointer-lock/
demo src: https://github.com/mdn/dom-examples/tree/master/pointer-lock
api doc:     https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API

jonnytest1:
so I created a fork for this functionality but I cant commit it since I dont know how to get eslint to recognize new variables oO ? https://github.com/jonnytest1/p5.js/blob/master/src/events/mouse.js (couldnt double check the docs yet since the data.json file dissapears all the time)

jonnytest1:
ok fixed the docs but the linter still complains about deltaX and deltaY to be undefined oO

lmccart:
closed with #3804 thanks @jonnytest1

PR #3804, opened 2019-06-10
Pointer Lock API

jonnytest1:
added functionality for the pointer lock api : https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API 
in reference to https://github.com/processing/p5.js/issues/3803

### DIFF
--- a/src/core/main.js
+++ b/src/core/main.js
@@ -386,8 +386,8 @@ var p5 = function(sketch, node, sync) {
 
         //reset delta values so they reset even if there is no mouse event to set them
         // for example if the mouse is outside the screen
-        this._setProperty('deltaX', 0);
-        this._setProperty('deltaY', 0);
+        this._setProperty('movedX', 0);
+        this._setProperty('movedY', 0);
       }
     }
 
--- a/src/core/main.js
+++ b/src/core/main.js
@@ -383,6 +383,11 @@ var p5 = function(sketch, node, sync) {
       // NOTE : This reflects only in complete build or modular build.
       if (typeof this._updateMouseCoords !== 'undefined') {
         this._updateMouseCoords();
+
+        //reset delta values so they reset even if there is no mouse event to set them
+        // for example if the mouse is outside the screen
+        this._setProperty('deltaX', 0);
+        this._setProperty('deltaY', 0);
       }
     }
 
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -12,7 +12,7 @@ var p5 = require('../core/main');
 var constants = require('../core/constants');
 
 /**
- * 
+ *
  * The variable deltaX contains the vertical movement of the mouse since the last frame
  * @property {Number} deltaX
  * @readOnly
@@ -978,9 +978,9 @@ p5.prototype._onwheel = function (e) {
  * Use <a href="#/p5/deltaX">deltaX</a> and <a href="#/p5/deltaY">deltaY</a> to get the difference the mouse was moved since
  * the last call of draw</p>
  * <p>Note that not all browsers support this feature</p>
- * <p>This enables you to create experiences that arent limited by the mouse moving out of the screen
- * even if it is repeatedly moved into one direction . </p>
- * <p>For Example a first person perspective experience</p>
+ * <p>This enables you to create experiences that aren't limited by the mouse moving out of the screen
+ * even if it is repeatedly moved into one direction. </p>
+ * <p>For example a first person perspective experience</p>
  *
  * @method requestPointerLock
  * @example
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -13,15 +13,15 @@ var constants = require('../core/constants');
 
 /**
  *
- * The variable deltaX contains the vertical movement of the mouse since the last frame
- * @property {Number} deltaX
+ * The variable movedX contains the vertical movement of the mouse since the last frame
+ * @property {Number} movedX
  * @readOnly
  * @example
  * <div class="notest">
  * <code>
  * let x = 0;
  * function draw() {
- *   x = (x + deltaX) % 100;
+ *   x = (x + movedX) % 100;
  *   background(237, 34, 93);
  *   fill(0);
  *   rect(x, 50, 50, 50);
@@ -31,18 +31,18 @@ var constants = require('../core/constants');
  *
  *
  */
-p5.prototype.deltaX = 0;
+p5.prototype.movedX = 0;
 
 /**
- * The variable deltaY contains the horizontal movement of the mouse since the last frame
- * @property {Number} deltaY
+ * The variable movedY contains the horizontal movement of the mouse since the last frame
+ * @property {Number} movedY
  * @readOnly
  * @example
  * <div class="notest">
  * <code>
  * let y = 0;
  * function draw() {
- *   y = (y + deltaY) % 100;
+ *   y = (y + movedY) % 100;
  *   background(237, 34, 93);
  *   fill(0);
  *   rect(50, y, 50, 50);
@@ -51,7 +51,7 @@ p5.prototype.deltaX = 0;
  * </div>
  *
  */
-p5.prototype.deltaY = 0;
+p5.prototype.movedY = 0;
 /*
  * This is a flag which is false until the first time
  * we receive a mouse event. The pmouseX and pmouseY
@@ -418,8 +418,8 @@ p5.prototype._updateNextMouseCoords = function(e) {
       this.height,
       e
     );
-    this._setProperty('deltaX', e.movementX);
-    this._setProperty('deltaY', e.movementY);
+    this._setProperty('movedX', e.movementX);
+    this._setProperty('movedY', e.movementY);
     this._setProperty('mouseX', mousePos.x);
     this._setProperty('mouseY', mousePos.y);
     this._setProperty('winMouseX', mousePos.winX);
@@ -975,7 +975,7 @@ p5.prototype._onwheel = function(e) {
 /**
  * <p>The function <a href="#/p5/requestPointerLock">requestPointerLock()</a>
  * locks the pointer to its current position and makes it invisible.
- * Use <a href="#/p5/deltaX">deltaX</a> and <a href="#/p5/deltaY">deltaY</a> to get the difference the mouse was moved since
+ * Use <a href="#/p5/movedX">movedX</a> and <a href="#/p5/movedY">movedY</a> to get the difference the mouse was moved since
  * the last call of draw</p>
  * <p>Note that not all browsers support this feature</p>
  * <p>This enables you to create experiences that aren't limited by the mouse moving out of the screen
@@ -1008,9 +1008,45 @@ p5.prototype.requestPointerLock = function() {
     canvas.requestPointerLock || canvas.mozRequestPointerLock;
   if (!canvas.requestPointerLock) {
     console.log('requestPointerLock is not implemented in this browser');
-    return;
+    return false;
   }
   canvas.requestPointerLock();
+  return true;
+};
+
+/**
+ * <p>The function <a href="#/p5/exitPointerLock">exitPointerLock()</a>
+ * exits a previously triggered <a href="#/p5/requestPointerLock">pointer Lock</a> programmatically
+ * for example to make ui elements usable etc
+ *
+ * @method exitPointerLock
+ * @example
+ * <div class="notest">
+ * <code>
+ * //click the canvas to lock the pointer
+ * //click again to exit (otherwise escape)
+ * let locked = false;
+ * function draw() {
+ *   background(237, 34, 93);
+ * }
+ * function mouseClicked() {
+ *   if (!locked) {
+ *     locked = true;
+ *     requestPointerLock();
+ *   } else {
+ *     exitPointerLock();
+ *     locked = false;
+ *   }
+ * }
+ * </code>
+ * </div>
+ *
+ * @alt
+ * black 50x50 rect moves up and down with mouse move while no cursor is visible
+ *
+ */
+p5.prototype.exitPointerLock = function() {
+  document.exitPointerLock();
 };
 
 module.exports = p5;
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -946,7 +946,7 @@ p5.prototype._onwheel = function(e) {
 };
 
 /**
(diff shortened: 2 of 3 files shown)


jonnytest1:
can someone please tell me how I can fix the tester not finding my variables oO ?
/home/travis/build/processing/p5.js/src/events/mouse.js:964:11: 'deltaX' is not defined. [Error/no-undef]
/home/travis/build/processing/p5.js/src/events/mouse.js:965:11: 'deltaY' is not defined. [Error/no-undef]

limzykenneth:
This feature is not supported by [IE11](https://caniuse.com/#search=pointer%20lock%20api), so unless we stop support for IE11 we can't really use this. I'm not sure the behaviour can be polyfilled as well.

jonnytest1:
I found this link which at least says it should be theorethically possible oO ?
https://docs.microsoft.com/en-us/openspecs/ie_standards/ms-pointerlock/a993ae3b-ef76-4fc1-991c-d6f588080152

limzykenneth:
I don't have a windows machine to test, but maybe it just works? I'm not sure.

jonnytest1:
well I did check a bit but havent found a working solution maybe I have an older version oO ?

outofambit:
hey @jonnytest1 thanks for opening this PR! i have a question about the pointer lock api and its use cases. it seems like the p5 api already covers the majority of the functionality with the [events api](https://p5js.org/reference/#group-Events) and [noCursor](https://p5js.org/reference/#/p5/noCursor).

but this seems to be the one thing p5 doesn't cover currently:
> Pointer lock lets you access mouse events even when the cursor goes past the boundary of the browser or screen.

is that right? (this seems useful to me!)

_____

as for IE11 support, i unfortunately couldn't find any polyfills 😢 i'm not sure what we can do, but i'll see if i can think of anything. 🤔

jonnytest1:
yes it locks the cursor to the position it it when you call the lock and makes it invisble adding a onmousemove listener will still trigger events(properties like clientX etc will always stay the same however movementX/movementY will show the difference since the last animationFrame) its very useful for making 1st person games etc

outofambit:
Hi again @jonnytest1 unfortunately I haven't been able to think of anything to make this work for IE11. Perhaps we could log an error message (and do nothing) when `requestPointerLock` is called in a browser that doesn't support it? We do something similar for `createCapture()` if the browser doesn't support `getUserMedia`.

cc @lmccart in case she has thoughts on adding a feature that wouldn't be available on IE11.

stalgiag:
Hi all, I just wanted to chime in and say that I made a [super simple test](https://editor.p5js.org/stalgiag/present/grDsFZPNR) with a build of this PR as per @outofambit's suggestion and after playing around for a bit, I am very pleased with this addition. I think the removal of the window bounds offers up an easy avenue to getting swipe-like interactions, which is useful in both 2D and 3D.

Definitely should be some additional documentation and further consideration given to naming (maybe something more accessible than `requestPointerLock`), but I think this is overall a great feature. It is unfortunate about IE11, but it is my understanding that IE11 support is on its way out as we migrate to ES6? I also think that Microsoft has fully abandoned IE at this point. Either way, best to make sure that documentation is clear on this point and log appropriately.

Thank you @jonnytest1 !

src/events/mouse.js
```diff
@@ -11,6 +11,44 @@
 var p5 = require('../core/main');
 var constants = require('../core/constants');
 
+/**
+ * @property {Number} deltaX
```
lmccart: I like the addition of `deltaX` and `deltaY` but I wonder if there's a better name. I'm thinking of the case where the user looks at the reference page and sees these variables in among the rest. Is there a way to name them to make it more clear what they relate to?

src/events/mouse.js
```diff
@@ -929,4 +969,44 @@ p5.prototype._onwheel = function(e) {
   }
 };
 
+/**
+ * <p>The function <a href="#/p5/requestPointerLock">requestPointerLock()</a>
+ * locks the pointer to its current position and makes it invisible.
+ * Use deltaX and deltaX to get the difference the mouse was moved since
```
lmccart: You can link deltaX and deltaY to their reference pages. See this code snippet as an example of linking: https://github.com/processing/p5.js/blob/master/src/image/loading_displaying.js#L22

src/events/mouse.js
```diff
@@ -929,4 +969,44 @@ p5.prototype._onwheel = function(e) {
...
+ * the last call of draw</p>
+ * <p>Note that not all browsers support this feature</p>
+ *
+ * @method requestPointerLock
+ * @returns {void}
```
lmccart: you can skip the @returns here if there is no return value

src/events/mouse.js
```diff
@@ -929,4 +969,44 @@ p5.prototype._onwheel = function(e) {
...
+ * @example
+ * <div class="notest">
+ * <code>
+ * //click the canvas to lock the pointer
+ * //press ESC to exit lock mode
+ * function draw() {
+ *   background(237, 34, 93);
+ *   fill(0);
```
lmccart: the `fill()` line here is unnecessary since nothing is drawn. it can be removed.

src/events/mouse.js
```diff
@@ -929,4 +969,44 @@ p5.prototype._onwheel = function(e) {
   }
 };
 
+/**
+ * <p>The function <a href="#/p5/requestPointerLock">requestPointerLock()</a>
+ * locks the pointer to its current position and makes it invisible.
```
lmccart: Can you add one more sentence here explaining what this function does / why it's useful? Something like "This enables you to..."

src/events/mouse.js
jonnytest1: oO How about movementX/Y thats what the event calls them or do you think more towards mouseDeltaX ?

src/events/mouse.js
jonnytest1: https://github.com/processing/p5.js/pull/3804/commits/2fcfa818300f93b63b3363efc8670254fe2e0fc6 added links

src/events/mouse.js
jonnytest1: https://github.com/processing/p5.js/pull/3804/commits/2fcfa818300f93b63b3363efc8670254fe2e0fc6 removed return

src/events/mouse.js
jonnytest1: https://github.com/processing/p5.js/pull/3804/commits/2fcfa818300f93b63b3363efc8670254fe2e0fc6 fill removed

src/events/mouse.js
jonnytest1: https://github.com/processing/p5.js/pull/3804/commits/2fcfa818300f93b63b3363efc8670254fe2e0fc6
https://github.com/processing/p5.js/pull/3804/commits/81d7b153e4a978165428688be730d616996c5013  added :

> <p>This enables you to create experiences that aren't limited by the mouse moving out of the screen even if it is repeatedly moved into one direction . </p>
>  <p>For example a first person perspective experience</p>

src/events/mouse.js
outofambit: I like that direction @jonnytest1! What about `movedX`/`movedY`? or maybe `traveledX`/`traveledY`?

src/events/mouse.js
```diff
@@ -929,4 +972,45 @@ p5.prototype._onwheel = function(e) {
...
+ *
+ */
+p5.prototype.requestPointerLock = function() {
+  // pointer lock object forking for cross browser
+  var canvas = this._curElement.elt;
+  canvas.requestPointerLock =
+    canvas.requestPointerLock || canvas.mozRequestPointerLock;
+  if (!canvas.requestPointerLock) {
+    console.log('requestPointerLock is not implemented in this browser');
+    return;
```
outofambit: What do you think about returning `false` here if its not supported? and return `true` at the end of the function if it is? That way developers can tell if its going to work or not and do something else if not? or maybe that's overkill.

src/events/mouse.js
jonnytest1: movedX/Y it is

src/events/mouse.js
jonnytest1: done

outofambit: thank you!

outofambit:
@jonnytest1 thanks for all of this work! i'm feeling pretty good about this PR. @lmccart @stalgiag what do y'all think?

src/events/mouse.js
```diff
@@ -11,6 +11,47 @@
...
+ * }
+ * </code>
+ * </div>
+ *
+ *
+ */
+p5.prototype.movedX = 0;
+
+/**
+ * The variable movedY contains the horizontal movement of the mouse since the last frame
```
stalgiag: I think that maybe 'horizontal' and 'vertical' are switched in this explainer

src/events/mouse.js
```diff
@@ -11,6 +11,47 @@
...
+ * @property {Number} movedY
+ * @readOnly
+ * @example
+ * <div class="notest">
+ * <code>
+ * let y = 0;
+ * function draw() {
+ *   y = (y + movedY) % 100;
```
stalgiag: I really like the look of these two examples but the interaction behavior feels a bit glitchy. I wonder if there is a way to smooth the behavior a little bit without adding too much complexity to the code?

src/core/main.js
```diff
@@ -383,6 +383,11 @@ var p5 = function(sketch, node, sync) {
       // NOTE : This reflects only in complete build or modular build.
       if (typeof this._updateMouseCoords !== 'undefined') {
         this._updateMouseCoords();
+
+        //reset delta values so they reset even if there is no mouse event to set them
+        // for example if the mouse is outside the screen
+        this._setProperty('deltaX', 0);
```
stalgiag: need to be updated to movedX/movedY?

src/events/mouse.js
```diff
@@ -929,4 +972,81 @@ p5.prototype._onwheel = function(e) {
...
+ * @example
+ * <div class="notest">
+ * <code>
+ * //click the canvas to lock the pointer
+ * //click again to exit (otherwise escape)
+ * let locked = false;
+ * function draw() {
+ *   background(237, 34, 93);
+ * }
+ * function mouseClicked() {
```
stalgiag: Make these examples (request/exitPointerLock) show visible change? Maybe an opportunity to demonstrate a simple version of a first person experience?

src/events/mouse.js
```diff
@@ -929,4 +972,81 @@ p5.prototype._onwheel = function(e) {
...
+    console.log('requestPointerLock is not implemented in this browser');
+    return false;
+  }
+  canvas.requestPointerLock();
+  return true;
+};
+
+/**
+ * <p>The function <a href="#/p5/exitPointerLock">exitPointerLock()</a>
+ * exits a previously triggered <a href="#/p5/requestPointerLock">pointer Lock</a> programmatically
```
stalgiag: I don't know if the word 'programmatically' is useful here. Adds more complexity to the sentence without conveying essential information.

src/events/mouse.js
stalgiag: Maybe a good chance to show off a swipe like interaction?

lmccart: also wondering if we can remove the `%` operator, I think this could be unfamiliar to many beginners

jonnytest1: how about an example where it always moves towards the center slowly and then moves according to the mouse movement oO ?

src/core/main.js
jonnytest1: oh thx

src/events/mouse.js
jonnytest1:
```
let x = 0;
function setup(){
   rectMode(CENTER)
}
function draw() {
  if(x > 50){
     x--;
  } else if( x < 50 ){
     x++
  }
  x += floor(movedX);
  background(237, 34, 93);
  fill(0);
  rect(x, 50, 50, 50);
}
```

jonnytest1: this is how i mean slowly moves towards the center and reacts to mouse move

src/events/mouse.js
jonnytest1: that might be a lot of code just saying

stalgiag: It might be but I don't think it has to be. [Here is a link that has it down to about 12 lines of code](https://editor.p5js.org/stalgiag/sketches/grDsFZPNR), which I don't think is outside the realm of the acceptable if this is a primary perk of the feature. (Note that this example uses a build from an older version of this PR thus the deltaX/deltaY)

jonnytest1: https://github.com/jonnytest1/p5.js/tree/examplefirstpersonexperience how about this one ? 
@requestPointerLock

stalgiag: That is definitely cooler but too complicated for a reference example. The example that I posted is much much simpler. You can modify it to match your sensibilities and style but I'd try to keep it under 20 lines.

jonnytest1: > It might be but I don't think it has to be. [Here is a link that has it down to about 12 lines of code](https://editor.p5js.org/stalgiag/sketches/grDsFZPNR), which I don't think is outside the realm of the acceptable if this is a primary perk of the feature. (Note that this example uses a build from an older version of this PR thus the deltaX/deltaY)

stolen :P

src/events/mouse.js
jonnytest1: working demo https://editor.p5js.org/jonathanheindl/present/cY13G4WIr

stalgiag: This looks good! If you have time to change these examples right now, we are about to make the 0.9.0 release and this feature could make it in!

jonnytest1: ah sorry Im operatiing on eu time zone

stalgiag: No sweat! It'll make it into 0.10.0

src/events/mouse.js
```diff
@@ -929,4 +990,86 @@ p5.prototype._onwheel = function(e) {
...
+ *   } else {
+ *     exitPointerLock();
+ *     locked = false;
+ *   }
+ * }
+ * </code>
+ * </div>
+ *
+ * @alt
+ * black 50x50 rect moves up and down with mouse move while no cursor is visible
```
stalgiag: Alright, truly almost done, can you just make these bits of alt text correct again? And make sure there is alt text for every example. Thanks!

jonnytest1: done

stalgiag:
Great! Thank you for this awesome contribution @jonnytest1 !

PR #3810, opened 2019-06-12
events/mouse.js: create unit tests

ihsavru:


test/unit/events/mouse.js
```diff
@@ -250,4 +250,238 @@ suite('Mouse Events', function() {
...
+
+      myp5.mouseWheel = function() {
+        count += 1;
+      };
+
+      window.dispatchEvent(new MouseEvent('wheel'));
+      assert.deepEqual(count, 1);
+    });
+
+    test('doubleClicked functions on multiple instances must run once', async function() {
```
outofambit: it looks like this test case is labelled incorrectly? i see the same description higher in the file and internally this uses `mouseWheel`

PR #3815, opened 2019-06-13
fixed #3691

### LINKED ISSUE #3691: randomSeed() with similar values produces similar sequences
The first value returned from `random()` after a call to `randomSeed(n)` is often very similar to the first value returned from `random()` after a call to `randomSeed(n+1)`. [This sketch](https://editor.p5js.org/allison.parrish/sketches/i8HzVWbtZ) demonstrates the problem:

```
0 62 89 84 64 18
1 62 98 52 68 85
2 62 7 21 72 51
3 62 16 89 75 18
4 62 25 58 79 84
5 62 34 26 83 51
6 62 43 95 86 18
7 62 52 63 90 84
8 62 61 32 94 51
9 62 70 0 98 18
10 62 79 69 1 84
11 62 88 37 5 51
12 62 98 5 9 18
```

The first column is the value passed to `randomSeed()`, the second column is the first value returned from `int(random(100))`, the third column is the next value returned from `int(random(100))`, etc. You can see that the first value returned from `random()` after seeding rounds to the same number. The second value for similar random seeds is also very similar (i.e., `randomSeed(n); random(); console.log(random())` will always display a number roughly 0.9 more than `randomSeed(n+1); random(); console.log(random())`, taking into account the "wrap-around" when the value exceeds 1). This behavior, as far as I can tell, is not typical for pseudorandom number generators.

This behavior becomes a problem when you want to use `randomSeed()` to produce stable random numbers over time and space; i.e., [this sketch](https://editor.p5js.org/allison.parrish/sketches/Y3FrTXQka) sets the random seed for each position in a 2D grid, and should have rectangles of uniformly distributed sizes; instead, it has a noticeable pattern. Likewise, [this sketch](https://editor.p5js.org/allison.parrish/sketches/YcWzf6Upd), which seeds the random number generator with the frame count (source code reproduced below) should probably show the ellipse moving to uniformly distributed positions on the screen; instead, the ellipse moves slowly across the screen as the first value returned from `random()` increases with the value passed to `randomSeed()`:

```
function setup() {
  createCanvas(400, 400);
}

function draw() {
  background(40);
  noStroke();
  fill(255);
  randomSeed(frameCount);
  ellipse(random(width), height/2, 50, 50);
}
```

#### Nature of issue?

- [x] Found a bug

- [x] Math

#### Details about the bug: 

- p5.js version: 0.7.3

Janglee123:
Fixes #3691.
As explained in [comment](https://github.com/processing/p5.js/issues/3691#issuecomment-482475050)

### DIFF
--- a/src/math/random.js
+++ b/src/math/random.js
@@ -16,14 +16,14 @@ var y2 = 0;
 // Linear Congruential Generator
 // Variant of a Lehman Generator
 var lcg = (function() {
-  // Set to values from http://en.wikipedia.org/wiki/Numerical_Recipes
+  // Set to values from glibc(useb by GCC) (https://en.wikipedia.org/wiki/Linear_congruential_generator#Parameters_in_common_use)
   // m is basically chosen to be large (as it is the max period)
   // and for its relationships to a and c
-  var m = 4294967296,
+  var m = 2147483647,
     // a - 1 should be divisible by m's prime factors
-    a = 1664525,
+    a = 1103515245,
     // c and m should be co-prime
-    c = 1013904223,
+    c = 12345,
     seed,
     z;
   return {


src/math/random.js
```diff
@@ -19,11 +19,11 @@ var lcg = (function() {
   // Set to values from http://en.wikipedia.org/wiki/Numerical_Recipes
```
outofambit: could we update this comment to point to whatever library we're copying these values from?

limzykenneth:
Nothing of major from my end. 

One thing I'm wondering about is whether to release this later as part of 0.9.0 instead of under 0.8.x as technically it is kind of a breaking change, ie. if there's a sketch that was using a particular seed for a particular set of random values, those would now need to be reseeded with new values. But then it's not such a big problem anyways.

src/math/random.js
Janglee123: just forgot to check out comment. :smile:

lmccart:
@limzykenneth I think that's fine. we have a 0.9.0 release aiming for end of june

outofambit:
thank you @Janglee123! 🎉

Issue #3806, opened 2019-06-10
Multiple types of materials to support WebGL Geometries


sanketsingh24:
#### Nature of issue?

- [x] Existing feature enhancement

#### Most appropriate sub-area of p5.js?

- [x] WebGL

#### Details 
Currently, in p5, only one of the `specularMaterial()` or `ambientMaterial()` can be applied to a geometry. But many objects in the real world can exhibit both behaviour. Take a look at this sketch from processing-

![Screenshot_2019-06-09_21-55-37](https://user-images.githubusercontent.com/24548786/59205289-4179e500-8bc0-11e9-9f53-915a7faf82a3.png)
[_(Code)_](https://pastebin.com/usCdhdZ3)

In short, the left sphere has just ambient (red) properties whereas the right sphere has both ambient (red) and specular (white) properties.

As I am implementing `emissiveMaterial()` also, we (me and @aferriss ) just got this thought and this would be a good feature. One of the things which we noticed was both [`specularMaterial()`](https://github.com/processing/p5.js/blob/master/src/webgl/material.js#L665) and [`ambientMaterial()`](https://github.com/processing/p5.js/blob/master/src/webgl/material.js#L614) use the same uniform and same variable, [which causes the later one to render](https://editor.p5js.org/sanketsingh24/sketches/vJM9_yMBt).

stalgiag:
Hi @sanketsingh24 and @aferriss, after finishing the summer, what are your thoughts on this issue? Thanks!

aferriss:
@stalgiag @sanketsingh24 I still think this would be a good feature to add, but it's probably not a simple PR. This touches a lot of the webGL rendering pipeline and untangling and sorting things out is probably going to be a bit of an effort. 

I don't think I'll have much time anytime soon to work on this and I'm pretty certain Sanket is also busy in the coming months. Maybe this warrants a help wanted tag?

stalgiag:
Thanks for the thoughts @aferriss !

I agree that this seems a bit rough, and definitely feels like it would need to be part of a larger restructuring. I am looking into these kinds of changes as part of #3794. I am happy to do the work if there is a sound high-level plan for a restructure or redesign, but without this I am reluctant to start messing with this many parts of the code before the 1.0 release. I'd be curious to draw opinions on something like this, but that might require first making a very clear map of the rendering pipeline as it stands now.

[...]

PR #3820, opened 2019-06-14
Feature implementation: emissiveMaterial()

sanketsingh24:
Implemented emissiveMaterial()

Work done- 
- [x] Implemented `uEmissive` uniform in both shaders
- [x] Implemented `emissiveMaterial()` in /src/webgl/material.js
- [x] Added a manual-test-example in /webgl/material/emissive/sketch.js
- [x] Unit tests
- [x] Documentation
- [x] Example

To do-

@aferriss


Spongman:
is this really the direction we want to go, though. i would rather move away from the current practice of overloading the fill color, and toward a more reasonable color model like the one that processing uses where the material has different color properties (ambient, emissive, specular) that all contribute separately to the lighting of the surface. this is much more in line with similar color models that are used in other libraries and 3d modeling/rendering packages.

sanketsingh24:
@Spongman I wouldn't call it overloading, but overwriting, and I also opened an issue regarding this #3806. It would require a lot of refactoring of both shaders and the js part. The shaders themselves are in pretty bad shape right now, and given only one material type is to be rendered, @aferriss and I agreed on this implementation.

Spongman:
oh, i'm sorry, i didn't realize that it had been decided. i would have like to have weighed in beforehand. can you point me to the issue where that discussion occurred?

aferriss:
@Spongman I think the issue Sanket is referring to is the one he referenced (https://github.com/processing/p5.js/issues/3806) but we didn't get any feedback regarding how to move forward. Since we're on a tight timeline for GSOC we decided it would be best to move forward in the paradigm that was already established (as far as we could tell) in p5 of one material type per geometry. 

I think we'd both like to eventually add the ability to use multiple material properties per geometry but since it was a much larger task we decided to punt and try to implement at the end of the summer if we had any leftover time.

Spongman:
yeah, i definitely think this area could use some more discussion before introducing these API changes.

aferriss:
@Spongman Are you asking why they are named `ambientMaterial()` vs. just `ambient()`? I think that these functions are appended with "material" to clearly distinguish it from `ambientLight()` and other lighting functions. Do you have any suggestions for how they could be better named?

As for how this is different than basic.frag, it's not. However I think it makes sense to have this kind of redundancy for setting the color of 3d geometry. I feel like in the context of 3d graphics, fill is the odd one out. If you're already using `ambientMaterial()` and `specularMaterial()` does it not make sense to also use `emissiveMaterial()`rather than `fill()`? 

I also think that this PR starts to lay some basic groundwork for a future where geometry could have multiple material properties.

stalgiag:
I agree with @aferris here. 

Restating points differently or offering my extended thoughts, I think that `fill()` is great for the user as-is, in that it offers a simple intuitive way of getting solid unlit colors especially for folks first making the move from 2D to 3D, but once we are moving into the language of 'materials' I think it makes more sense to offer a different term for the base color for a material. As was pointed out, this is useful mostly once multiple simultaneous materials are achieved. Once this is the case, it becomes a bit confusing for fill to dictate the main color state across all materials instead of just indicating a completely unlit solid color material. There are already many contradictions between the fill color state and material expectations, with `normalMaterial` being the most apparent example. 

And as far as naming, I actually like `ambientMaterial` versus `ambient`. I think the distinction between lighting and material is a little more immediately clear with this than it is in Processing.

aferriss:
Going to merge this as I haven't heard any other dissenting opinions.

Spongman:
@aferriss i find your explanation above to be wholly unsatisfying. now we have two apis for doing essentially the same thing, complicating and confusing the documentation, and 
the rendering pipeline.

i feel this was entirely rushed through with insufficient thought.

aferriss:
@Spongman I'm sorry to hear that. I felt that the PR was in good shape and that there had been sufficient explanation as to the design and intention of the feature. It seemed to me that enough time had passed and since I hadn't heard from you since I was inclined to merge. I'm happy to continue the discussion if you have further thoughts.

stalgiag:
@Spongman luckily no change is permanent! Open-source software is something that changes with time and not in a strictly unidirectional or linear fashion, so let's not exaggerate the severity of any one decision. I think everyone here is happy to talk if you feel like the merits of this design direction should still be discussed.

 For my part in said discussion, I wouldn't say this complicates and confuses the documentation or render pipeline. I do think that there is some confusion but that is because this is a step towards a more fully realized change in which `curFillColor` isn't always the intended material color. Also, it is important to note that there is nothing radical in the doubling of some duties between `fill` and `emmisiveMaterial` as this can be seen in Processing. That said, I'd be happy to try and think through why we might want to break from Processing and develop a different API for multiple simultaneous materials.

Perhaps there is an opportunity for an innovation in design here but it can only be reached through suggestions and discussion.

stalgiag:
@Spongman could you send me an email at the address listed in my profile? I’d be grateful for the opportunity for a more direct conversation outside of the repo. We could discuss any concerns about the way that decisions are made with the library and also talk about what the expectations are for engaging with the community. One-on-one communications seems ideal for the task. Thanks!

PR #3825, opened 2019-06-17
fix quad drawing (textureMode test)

### LINKED ISSUE #3784: textureMode manual test and reference example broken in master
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL

- [x] Desktop/Laptop

#### Details about the bug: 

- p5.js version: master branch
- Web browser and version: 
- Operating System: 
- Steps to reproduce this:
Build master and look at reference example at
test/manual-test-examples/webgl/material/textureMode/ or run `npm run docs:dev` and look at the reference example for textureMode

Screenshot (643)

No messages in console. Oddly the very similar test at manual-test-examples/webgl/immediateMode/texture/ works as expected.

stalgiag:
fixes #3784

I am no expert on the different permutations that are possible with immediateMode drawing but I think this covers everything as well as it was before. No manual test or reference example broken, hollow fill tessellation still working for coplanar shapes and the code is a teeny bit more simple.

### DIFF
--- a/src/webgl/p5.RendererGL.Immediate.js
+++ b/src/webgl/p5.RendererGL.Immediate.js
@@ -198,6 +198,7 @@ p5.RendererGL.prototype.endShape = function(
           this.drawMode === constants.FILL &&
           this.immediateMode._isCoplanar === true)
       ) {
+        this.immediateMode.shapeMode = constants.TRIANGLES;
         var contours = [
           new Float32Array(this._vToNArray(this.immediateMode.vertices))
         ];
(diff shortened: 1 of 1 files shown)


stalgiag:
Okay I left this up for a little bit in case someone knew of a case when this path through the shapeMode conditionals might not work for custom shape rendering in WebGL. Doesn't seem to be the case.

I have tested with all manual tests and every reference example. Everything appears to be the same except the fixing of the bug in #3784. So I am going to merge. Again, not entirely sure that this addresses every permutation, which tells me that we need a better manual test for WebGL immediate mode (which I will address in #3826), but this PR currently fixes the problem without creating any immediately apparent new problems. Feel free to open an issue or comment in this PR, if this breaks something for any reason.

PR #3834, opened 2019-06-24
Add test coverage reporting

### LINKED ISSUE #3385: Measure Unit Test Coverage
Instrument the unit test suite so we can see which parts of the codebase are well tested and which need more coverage.

Some relevant libraries and services that might work for this:
* [Istanbul + nyc](https://istanbul.js.org/docs/tutorials/mocha/)
* [CodeCov](https://codecov.io)

outofambit:
closes #3385 

* adds a third build of p5.js that includes instrumentation to measure code coverage.
* general summary of test coverage is printed out after running tests
* detailed html page of test coverage is generated at `coverage/index.html`


tasks/test/mocha-chrome.js
```diff
@@ -55,17 +57,26 @@ module.exports = function(grunt) {
...
-              if (stats.failures) reject(stats);
-              else resolve(stats);
+            eventHandler1.on('mocha:end', async results => {
+              const { stats, coverage } = results;
+              if (stats.failures) {
+                reject(stats);
+              } else {
+                if (coverage) {
+                  await fs.mkdir('./.nyc_output/', { recursive: true });
+                  await fs.writeFile(
```
meiamsome: Any errors in these will not be caught properly and might be problematic - I would probably consider doing something like
```js
          const { stats, coverage }  = await new Promise(resolve => eventHandler1.on('mocha:end', resolve));
          if (stats.failures) {
            throw new Error(stats);
          }
          if (coverage) {
            await fs.mkdir('./.nyc_output/', { recursive: true });
            await fs.writeFile(
              './.nyc_output/out.json',
              JSON.stringify(coverage)
            );
          }
```

outofambit: how does this look now? i tried the above but had some issues so i tried to find something between that and the current structure.

Issue #3758, opened 2019-05-21
ES6 Transition: Concerns & Progress


hsab:
Hello everyone! 

I'm investigating the means of fully transitioning p5 to ES6. I've already spoken to some of the developers outside of Github and was hoping to shine some light on some of the intricacies that I've observed in the code.

I've noticed there are 42 instances of `module.exports = p5;` in the bundled .js files. And I hope this issue serves as a clarification of the assumption that I have:

From what I understand each `module.exports = p5;` is effectively exposing p5 after extending it in its own respected files/module, be it curves.js, vertex.js, etc. 

Although what seems to be the case is that most modules usually inherit/require from main.js, there are few instances of sequential inheritance of p5. 

For instance:
In `string_functions.js` we have:
```
var p5 = require('../core/main');
require('../core/error_helpers');
....
module.exports = p5;
```
Both `main` and `error_helpers` have their own respective `modules.exports = p5`. However `string_functions` is exporting the instance of p5 inheritted from `main`.

My question in regards to this specific instance is by the end of `string_functions` when we are exporting p5, will this instance have all the prototypal functions defined in `error_helpers`? 

From my understanding, the inheritance scheme seems to be a directed acyclical graph with a single root node (`main.js` - **A** in picture) and a single leaf node which is p5 at its final stage after all the additions to the prototype occur in multiple files ( **F** in picture or equivalently `p5()` in browser).

![1sQcP](https://user-images.githubusercontent.com/11205091/58073047-672e4400-7b56-11e9-84ca-031f21bd9674.png)

I know javascript is independent of the order the code is written unlike C++, and instances and references are resolved during JIT. Given that there are 42 instances of `module.exports=p5` in the bundled lib, how is each instance resolved properly? Is each instance unique, with functionalities unavailable in preceding exports or do they all refer to the same class for lack of a better word?

Moreover given the `import` syntax in ES6, how does this export/require paradigm change, if it does?

Apologies in advance if these questions seem trivial or show a lack of knowledge in JS. I come from different programming languages and JS seems like wild west to me!

Edit: I'm unable to appropriately label my issue. I assume it should be considered a "discussion".

Spongman:
There is only one p5 constructor function (‘class’), and it is passed around (and returned by) all the other various modules. There’s no inheritance going on (except for the Graphics stuff - ignore that), all the p5 methods are just added into the same prototype object

This is impossible to do in ES6-style class definitions.

hsab:
[...]

PR #3843, opened 2019-06-28
Feature implementation: specularLight

sanketsingh24:
Implemented specularLight()

This function separates the diffuse and specular quantities of `pointLight()` and `directionalLight()`. Defaults to (255,255,255) so that this does not become a breaking change.

See manual tests `/webgl/lights/specularLight/` and `/webgl/material/specular_ambient/`.

Work done- 
- [x] Implemented `uDirectionalSpecularColors` and `uPointLightSpecularColors` uniform in both shaders
- [x] Implemented `specularLight()` in /src/webgl/light.js
- [x] Added a manual-test-example in `/webgl/lights/specularLight/`
- [x] Unit tests
- [x] Documentation
- [x] Example

To do-

@aferriss


Spongman:
shouldn't this be called `specularMaterial()` ?

the name `specularLight()` seems to fit in the with the naming scheme for defining lights (eg, `pointLight`, `directionalLight`) whereas this is defining a matrial property.

sanketsingh24:
This method intends to distinguish the specular and diffuse quantities from the `pointLight` and other lighting methods. It affects the lights directly so I thought it would be better to name it `specularLight`, although it is not a light, but then `ambientlight` is too. I am open to suggestions though, and If a better name comes to my mind, i will post about it.

stalgiag:
Hi @sanketsingh24 and @aferriss just a heads up that the transition to ES6 is happening this week. See this [post for a brief overview.](https://github.com/processing/p5.js/issues/3758#issuecomment-508964727)

sanketsingh24:
## specularLight(color)
What this method does is differentiates the specular and diffuse components from the pre-existing lighting methods, which improves the customizability of the sketch. The default value is `(255,255,255)`, which is used when this method is not called at all in the sketch, which prevents this feature to become a breaking change. It affects only the methods which are created after it in the sketch. The method is per-light, so it doesn't add up

Artists will be able to emulate something like this now-
specularLight_sketch
Code-
```js
function setup() {
  setAttributes('perPixelLighting', true);
  noStroke();
}

function draw() {
  background(0);

  shininess(10);
  ambientLight(50);
  specularLight(250, 0, 0);
  pointLight(255, 0, 0, 0, -100, 100);
  specularLight(0, 250, 0);
  pointLight(0, 255, 0, 0, 100, 100);
  specularMaterial(255);
  sphere(80);
}
```

Whereas, this was the previous behavior-
https://editor.p5js.org/sanketsingh24/sketches/Y7mzOuSR_

I think it is very close in behaviour to Processing's `lightSpecular`, and probably behaves in the same way.

aferriss:
@stalgiag Thanks for the heads up re: ES6 changes! @sanketsingh24 and I spoke about this and feel comfortable making the changes needed to resolve any merge conflicts. 

I also wanted to chime in with my understanding of the `specularMaterial()` and `specularLight()` functions and how they differ. 

*Specular Materials*
A specular material is used to set the diffuse color of a geometry's material, and tells the renderer that this material should reflect specular highlights. Currently in p5, all specular highlights are 100% white, regardless of light or material color. 

The color parameters passed to `specularMaterial()` work in tandem with the colors passed to `pointLight()`. Here's a few different scenarios: 

A green point light and a red specular material, the only thing that you will see is the white specular highlight.

aferriss:
@stalgiag @Spongman Just wanted to check in to see if you all had any feedback on this PR. It would be great to get this merged since most of the rest of our GSOC work is tied together with this function. 

@sanketsingh24 And I decided that renaming to `specularColor` made more sense than `specularLight`, but I think we're both still open to other alternatives. 

I feel good about the state of this but thought I would check in to see if anyone else had thoughts.

aferriss:
Thanks for the vote of confidence @stalgiag ! I'm going to merge but if anyone has further suggestions re: naming or clarification of this function down the road please feel free to reach out.

Issue #3830, opened 2019-06-22
p5.dom: removeElements() doesn't remove <option>s


thely:
As above. It looks like removeElements works by adding HTMLCanvasElement to the created elements, which isn't added to any `<option>` tags added to a `<select>`. I had planned to have a static `select` in my HTML and use p5 to swap out lists of options, but since I found no good way to just clear the children, I had to push the `<select>` into p5 as well. So, I guess I have two questions:

1. Shouldn't `removeElements()` clear the p5.dom-generated `<option>`s of a `<select>`, even if the parent `<select>` is in the HTML, not p5?
2. Is there a good way to clear the children of any `select()`ed element in p5?

Edit: using p5 version 0.8.0, as distributed through node. Firefox 67.

nout-kleef:
I'm trying to replicate what you're doing, but I'm wondering: how are you selecting your `<select>` element if you're not adding it using p5?
Are you using plain js (`document.getElementById()`)?
Otherwise, I'm not sure how you plan on targeting that same `<select>` when removing its `<option>`s

thely:
Not with plain JS, with p5's `select()` method.

```html
<select class="some-select"></select>
```

```js
var dropdown = p.select(".some-select");
for (let i = 0; i < optionsList.length; i++) {
  dropdown.option(optionsList[i].name, optionsList[i].val);
}
```

This adds `<option>` elements just fine, but it doesn't remove them when `removeElements()` is called. If that's not how p5.dom is intended to be used, I guess that's just how it is?

nout-kleef:
It seems like `dropdown.option()` does not result in the newly created `<option>` being added to the list of p5 DOM elements.
I've been looking, but I'm unable to find the `.option()` method declaration anywhere.
Do you know the file in which this method is declared?

thely:
I'm pretty sure the `option()` method is a sub-method (?) of `createSelect()` - see [this line](https://github.com/processing/p5.js/blob/4584e993248f603d6d00f60257da3eaad637635a/lib/addons/p5.dom.js#L699) in p5.dom.

Mayank-Mishra0110011:
https://github.com/processing/p5.js/blob/0.8.0/lib/addons/p5.dom.js#L101 on line 101 the `_wrapElement(res)` method only returns a `createSelect()` element without adding the element to `this._elements` array and therefore the `removeElements()` method cannot remove it. Before returning res it should add it to the elements array. So a simple `addElement(res, this)`before the return statement fixes the issue.

PR #3845, opened 2019-06-29
Fixed issue #3830

Mayank-Mishra0110011:
In `select()` method `_wrapElement(res)` method returns a `createSelect()` element without adding the element to `this._elements` array. The `removeElements()` method thus gives an error since the select element created is not a part of the `_elements` array. I added `addElement(res, this)` before the element is returned so that it's pushed to the `_elements` array which fixes the error thrown.

### DIFF
--- a/lib/addons/p5.dom.js
+++ b/lib/addons/p5.dom.js
@@ -98,6 +98,7 @@
       }
     }
     if (res) {
+      addElement(res, this);
       return this._wrapElement(res);
     } else {
       return null;


lmccart:
Thanks! This isn't quite there, because we don't want the whole select element to be added to the elements array, just the parts created by the `option()` function. I've fixed this in https://github.com/processing/p5.js/commit/30cf6fac792ee2df799230ab1a79064151b39c52.

PR #3852, opened 2019-07-01
p5.Graphics and setAttributes

### LINKED ISSUE #3836: setAttributes() breaks rendering of WebGL p5.Graphics
While working on another issue I noticed that `setAttributes()` causes errors with p5.Graphics objects. I have a working branch where I have partially fixed this. In my branch, I have this working:

```
let pg;
function setup() {
  createCanvas(500,500);
  background(0,0,255);
  pg = createGraphics(250, 500, WEBGL);
  pg.setAttributes('alpha', true);
  pg.clear();
  pg.box(100);
}
```
but the rendering is still a bit off when both canvases are 3D as in:
```
let pg;
function setup() {
  createCanvas(500,500, WEBGL);
  background(0,0,255);
  pg = createGraphics(250, 500, WEBGL);
  pg.setAttributes('alpha', true);
  pg.clear();
  pg.box(100);
}
```
I will fix this part as well and open a PR soon. I just wanted an open issue to track this work.

stalgiag:
fixes #3836

Now every combination of 2D and 3D canvases and graphics objects work together with modified and unmodified attributes.

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -209,25 +209,20 @@ p5.RendererGL.prototype._resetContext = function(options, callback) {
   var h = this.height;
   var defaultId = this.canvas.id;
   var isPGraphics = this._pInst instanceof p5.Graphics;
+
+  var replaceCanvas = function(owner) {
+    owner.canvas.parentNode.removeChild(owner.canvas);
+    owner.canvas = document.createElement('canvas');
+    var node = owner._pInst._userNode || document.body;
+    node.appendChild(owner.canvas);
+  };
+
   if (!isPGraphics) {
-    var c = this.canvas;
-    if (c) {
-      c.parentNode.removeChild(c);
-    }
-    c = document.createElement('canvas');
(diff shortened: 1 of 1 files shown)


themike27:
I think this isn't working again.
When I put `pg.setAttributes('alpha', true);` it returns me this error: `Uncaught TypeError: Cannot read property 'push' of undefined`

stalgiag:
Hi @themike27 are you sure that you are using the most recent version of p5? [Here is a working example.](https://editor.p5js.org/stalgiag/sketches/ApPPeBo7K) If you are using the most recent version and you are still experiencing the problem, please open an issue with a code example. Thanks!

themike27:
Oh I just saw that the command `npm i p5` installs the `0.8.0` (six months ago) and not the lastest `0.9.0` (three months ago), do you know how can I install the latest through `npm`?

stalgiag:
Hm odd. You can try `npm install p5@0.9.0`

themike27:
Yes! That was it! Thank you so much!

PR #3860, opened 2019-07-06
Introduce p5.Vector#rem to calculate the remainder/modulo of a vector.

frewsxcv:
https://en.wikipedia.org/wiki/Modulo_operation

This would be helpful for drawing particles that need to wrap when they go offscreen.

For example, instead of:

```js
particle.position.add(this.velocity);
particle.position.set(
  particle.position.x % width,
  particle.position.y % height,
);
```

Now one can write:

```js
particle.position.add(particle.velocity).rem(width);
```

The signature looks like this:

```js
p5.Vector.prototype.rem = function rem(n) {
  ...
};
```

The one-demension (one parameter) behavior matches what we do for `p5.Vector#mult` and `p5.Vector#div`, in contrast to `p5.Vector#add` and `p5.Vector#sub`. One drawback with this design is you can't calculate the remainder of more than one dimension. In the example above, if my canvas was not a square, I would not be able to use `p5.Vector#rem` to wrap particles since the width and height would be different numbers.

I haven't written the docs yet because I'm not yet sure if the one-dimension design is the one we want to go with.

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -570,6 +570,20 @@ p5.Vector.prototype.div = function div(n) {
   return this;
 };
 
+p5.Vector.prototype.rem = function rem(n) {
+  if (!(typeof n === 'number' && isFinite(n))) {
+    console.warn(
+      'p5.Vector.prototype.rem:',
+      'n is undefined or not a finite number'
+    );
+    return this;
+  }
+  this.x %= n;
+  this.y %= n;
+  this.z %= n;
+  return this;
+};
+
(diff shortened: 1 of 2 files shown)


lmccart:
thanks @frewsxcv. new features need to be discussed in issues before they're implemented. you can read more about the process [here](https://github.com/processing/p5.js/tree/master/developer_docs#how-to-contribute). thanks!

frewsxcv:
Ah, my bad, sorry about that! I'll close this and open an issue

PR #3873, opened 2019-07-10
Improving WebGL manual tests

### LINKED ISSUE #3826: Cleanup and Organization of WebGL Manual Test Examples
Hi all! this is just a note to say that I am going to make a pass over the WebGL manual test examples soon and do a clean-up. I will specifically be focusing on removing redundancies, organizing tests into more distinct categories, and adding visible HTML to clarify what expected output is.

GL Renderer development is currently very reliant on manual test examples for detecting bugs and I find that with the current state of the tests it can be arduous to look through everything. I just want to make this a little bit more efficient and clear.

stalgiag:
This is the beginning of an overhaul on the WebGL manual tests to:

- Get the best coverage possible for features that can't be unit tested
- Reduce the number of separate tests that must be looked at to review
- Establish a standard for explanation info in the test so that new users can more easily understand expected output

This is still a work-in-progress. The diffs aren't great to look at because I am mostly deleting these tests in order to recombine them or alter them fundamentally.

closes #3826


stalgiag:
Okay so I think this is ready. Apologies for the large PR. I think that due to the nature of shuffling the various types of manual test coverage around, it made sense to do this in a single PR.

Various large merge conflicts and the overall size of this PR make it hard to completely nail down the formatting but I think this is a first good attempt. This PR does the following:
- Reduces the number of manual tests from 56 to 24 with improved coverage. This was primarily done by deleting redundant tests and combining similar tests.
- Adds `gl-info` ID to test css for explaining what the test should do. This is useful because `header` was getting too large for some of the lengthier explainer texts.
- Sets a loose standard for explainer text. These lengthier explainer texts are useful in that they indicate what is being tested, what should be seen, and what the expected performance is.
- Fixed all the tests that were breaking for reasons unrelated to library bugs

As I said, I'd still like to tweak, format, cleanup and standardize these tests further but this seems like a solid start. I also have already spent an unexpected amount of time on this and would like to move onto other tasks for a while. 

If anyone does want to checkout/review this PR, I'd recommend doing so by viewing the manual tests locally. The test code largely remains unchanged, so instead of doing a line-by-line review, it makes more since to review this in broad strokes by going through the workflow of viewing the tests. Some tests are a bit crowded with information now, but this allows us to check a large number of features with quick glances which seems like the priority with manual tests.

stalgiag:
Okay keeping up with the merge conflicts on this is a little difficult. It might be better to merge if people think it is preferable to have fewer manual tests. The primary next merge conflict that I expect will come from the spotlight PR #3913. @sanketsingh24  if I merge this first would you be willing to add the new spotlight manual test as a row in the `geometryUnlit` test in this PR? If not I can do it and fix the merge conflicts again. The frustum test can continue to be its own test.

Also both @sanketsingh24 and @aferriss could you briefly look over this new set of manual tests and say whether you think it is workable on your end? Again as noted in the above comment, these tests are a little crowded but the goal is get as few tests as possible to glance at. Let me know.

aferriss:
@stalgia ah, I merged spotlight before I saw this. But we'll add the manual tests for both frustum and spotlight in here. I think you meant the `geometryLit` test right? 

The tests look good to me, I agree it's nice to have all these tests in one place.

aferriss:
Also I think there is maybe a test missing for `lights()`

stalgiag:
Oh cool thank you for catching the missing `lights()`. And yes, haha `geometryLit`. I will update accordingly :-)

sanketsingh24:
@stalgiag Shall I push the changes after this gets merged? I think I will need to change `spotLight`, `frustum` and `noLights` according to this PR.

stalgiag:
Hi @sanketsingh24 don't sweat adjusting for this PR. I am currently caught up with other work. I will handle the merge conflicts on my end and introduce your manual tests back into it once I get back to this PR. Just make your manual tests in the traditional way and I will use those as a cue on how to add them back into this PR. Thanks for checking in!

aferriss:
Thanks @stalgiag !

stalgiag:
Okay going to merge the newer manual tests in 1 day to avoid further merge conflicts. Should include everything now, thanks for the patience everyone!

Issue #3877, opened 2019-07-10
replace const with let in reference examples


stalgiag:
Hi all, this is in no way urgent but just wanted to open an issue to discuss the usage of `const` for mutable assign-by-reference variables.

This can be seen in a number of places, [here for example](https://github.com/processing/p5.js/blob/master/src/core/rendering.js#L146) or [here.](https://github.com/processing/p5.js/blob/master/src/core/shape/vertex.js#L381) This isn't always the case [as seen here.](https://github.com/processing/p5.js/blob/master/src/color/p5.Color.js#L799)

This is technically okay since the immutable part of an assign-by-reference variable is the reference and the data is able to change. A [clear explanation is available here.](https://jack.ofspades.com/es6-const-not-immutable/index.html)

Though it is technically okay, I personally feel that it is little more clear to only use `const` for declaring  variables that will not be modified within the scope of that variable. If the data connected to a variable will be changed through use of that variable then it should be declared with `let`.

limzykenneth:
I kinda feel the opposite. JS, even after ES6, is still very quirky and as you mentioned `const` in this case doesn't make the variable immutable but instead just make sure the variable cannot be reassigned. It is something that I thought could have been better named at least. However, this prevention of reassignment is actually quite useful development wise to prevent accidental reusing of the same name for something else. As a rule all variables should be `const` unless it needs to be reassigned in which case it should be `let` (something similar exist in Swift for example with `let` and `var`).

For immutable objects, it would be better to `Object.freeze()` them so that they are truly immutable and the assign them to a `const` for the full C/C++ style constant (almost).

stalgiag:
That is interesting. I guess I take `const` as a suggestion about how to think about the variable. But I guess these reductive conceptual tools are not great to lean on. The protection against reassignment does seem helpful, and definitely more reliable than an implied protection against mutability through the contributor's association with the word constant.

Either way I think some consistency should be added, both in the source and the examples.

hsab:
My justification for using `const`, had to do particularly with reassignment in line with the [description of its capabilities](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const) rather than immutability. In the examples you mentioned, my thinking process was "we shouldn't reassign `verts` but we should be able to append to it, etc. Same applies to `props`. Although it doesn't ensure immutability as @limzykenneth  mentioned, it is a pretty clear indication of the significance of the variable and the fact that it shouldn't be reassigned. 

[...]

Issue #3875, opened 2019-07-10
Use of arrow functions for prototype definition


stalgiag:
Since the ES6 update has been merged already, I am moving a discussion out of #3874 and into an issue so that it is easier to track.

The issue as outlined by @limzykenneth in the pull request discussion:

> Great work @hsab!
> 
> I have a particular question on the use of arrow functions to define prototype members. From what I can find it seems to affect the value of `this` in the member function (arrow function's `this` is defined when the function is evaluated, not when the object is created as such it doesn't point to the object instance created). For a discussion about it [here](https://stackoverflow.com/questions/31755186/es6-arrow-functions-not-working-on-the-prototype) and [here](https://stackoverflow.com/questions/34361379/are-arrow-functions-and-functions-equivalent-exchangeable).
> 
> I can't identify with a quick look whether it is a problem here or not but if the use of arrow function on prototype member isn't really doing anything other than saving bytes it seems a bit superflous to me.

@hsab replied:

> @limzykenneth Thanks!
> 
> This is certainly a valid and concerning point. I will have a thorough look tomorrow.
> Although I was aware of this behavior, I can't recall running into any issues with this.
> Also, just did a quick look and it seems like exactly what you mention is happening in [`main.js`](https://github.com/processing/p5.js/commit/694177b46055032c864feed8f2941e2ab0726824#diff-848b28a6f51426fffc2b0c6b183fa1c0). But these functions (`_start`, `_draw`, `_setup`) are critical (right?) and yet `this` seems to point to object instance during execution. This is quite a curious case. Definitely needs more investigation. If you have an idea of why, I would very much appreciate it.

followed by @limzykenneth : 
> @hsab From my understanding, basically in the `main.js` case you pointed out, at the time the arrow function is evaluated, the value of `this` is the same as the value of `this._start`'s `this` since they are defined in the same scope and so share the same value for `this`. However, when attaching to the prototype, the value of `this` should be defined at the point of object creation (when `new Something()` is called) but with arrow functions, the value of `this` in the function will be defined where the function is defined.
> 
> So in simpler terms, the issue will be centered around the usage of arrow functions on prototype members, the rest should not encounter problems and in my opinion ok to use arrow functions.

Note that this doesn't seem urgent so feel free to continue discussion and response at your own convenience. I just wanted a better place to track the topic. Thanks!

meiamsome:
The conversion is correct in maintaining behaviour - it has converted functions of two forms:
[...]

PR #3874, opened 2019-07-10
ES6 Bootstrap

hsab:
## ES6 Bootstrap
This pull request follows the #3758 discussion; with the aim of transitioning p5 to ES6. It consists of minor changes to the build system to facilitate processing, linting and testing the library with ES6 syntax and standards, as well as major and ubiquitous syntactical modification in line with ES6 features.

It is worthy to note that these transformations are by no means complete, and do not reflect nor implement every possible feature of ES6. They are intended to facilitate a smoother transition to properly and efficiently utilize ES6 features if and when aligned with the community interests and standards. And serve to motivate contributors to gradually conform to the new style and features.

These changes are applied with the following in mind:
- Reduce syntactical ambiguity and JS idiosyncrasies.
- Increase clarity and refrain from obfuscation resulting from new features (e.g.  [Destructuring](http://es6-features.org/#ParameterContextMatching))
 - Prioritize performance when necessary (e.g. [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach))

Moreover, there are many areas of p5 that could benefit from ES6 features beyond the scope of this PR. These features include but are not limited to: 
- Generators
- Number Type Checking
- Set/Map Data-Structures
- Typed Arrays
- etc.

Unfortunately, implementing, testing and benchmarking all these features, was beyond my availability, expertise, and code familiarity.

### Process
After modifying the build system to accommodate the migration, a series of transformations are applied, bootstrapped by [lebab](https://github.com/lebab/lebab), an ES5 to ES6/ES7 transpiler. For every transformation the following process was performed:
- Apply transformation using lebab
- Ensure correct formatting using `npm run lint:fix`
- Manually inspect every single line that was modified and fix/adjust if needed
- Test and build using `npm run grunt`
- Check the example(s) of affected function(s) in the reference doc
- Commit the transformation

In the end, the library was tested against a large random selection of the examples available through [p5.js-web-editor](https://github.com/processing/p5.js-web-editor/) to ensure correct functionality. This was done by cloning the web editor repository and using the generated library files (`p5.js` and `p5-min.js`) instead of the CDN provided 0.8.0 version.

### Issues & Concerns:

- `new p5()` fails to execute when using `combineModules`, although global mode functions as expected. The error thrown is `p5 is not a constructor`
[...]

PR #3881, opened 2019-07-11
Converted docs/documented-method.js to ES6

abhinavsagar:
Part of [#3758](https://github.com/processing/p5.js/issues/3758)

### DIFF
--- a/docs/documented-method.js
+++ b/docs/documented-method.js
@@ -36,8 +36,8 @@
   DocumentedMethod.prototype = {
     // Merge parameters across all overloaded versions of this item.
     _getMergedParams: function() {
-      let paramNames = {};
-      let params = [];
+      const paramNames = {};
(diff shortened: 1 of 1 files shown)


abhinavsagar:
@stalgiag Should the codebase itself be converted to ES6 ?

stalgiag:
Hi @abhinavsagar thanks for the offer but most of this conversion was handled by #3874. That said there will likely be further slight modifications that might need help. Feel free to follow open issues such as #3877 and #3875 to see if there are any new todos about ES6 conversion.

stalgiag:
I think that #3877 in particular is relevant to this particular PR. Definitely the change from `var` will make sense but we may use `const` instead of `let` for this type of variable.

abhinavsagar:
Thanks. I will go through the open issues.

PR #3890, opened 2019-07-14
Add optional crossOrigin parameter to createImg()

### LINKED ISSUE #3564: Issues with CORS in image

outofambit:
closes #3564 
(built off of #3715)

* Added an optional `crossOrigin` parameter to `createImg()`, enabling the user to explicitly opt for a CORS reuquest.
* Simplifies optional arguments by requiring users to include an "alt text" for all images they create. (This doesn't stop them from using an empty string, though.)

-----
thanks to @CleezyITP, @sinabahram, @kjhollen, @lm-n, @lmccart, @oliviaross for recent conversations about accessibility. i think its completely appropriate for p5 to require an alt text for image creation as both an accessibility and educational measure. 💛


limzykenneth:
The order of arguments of the signature is probably causing some problems for the test.

I'm away on a trip at the moment so I can't review this properly just yet but I'll aim to have a close look at this in a week's time. If someone else had a look you can go ahead and merge this, I can always look through it after.

src/dom/dom.js
```diff
@@ -2011,7 +2006,7 @@ p5.Element.prototype.hide = function() {
  * <div class='norender'><code>
  * let div = createDiv('this is a div');
  * div.size(100, 100);
- * let img = createImg('assets/laDefense.jpg', () => {
+ * let img = createImg('assets/laDefense.jpg', 'a test image', '', () => {
```
lmccart: can you make this alt text more descriptive of the image content itself to serve as a good example for users?

src/dom/dom.js
```diff
@@ -1002,7 +997,7 @@ p5.prototype.createInput = function(value, type) {
  * function handleFile(file) {
  *   print(file);
  *   if (file.type === 'image') {
- *     img = createImg(file.data);
+ *     img = createImg(file.data, 'alt text');
```
lmccart: if this is an image not really intended to be viewed, you can use empty string `''` for alt text to indicate that to screen readers

src/dom/dom.js
```diff
@@ -2145,7 +2140,7 @@ p5.Element.prototype.remove = function() {
  * }
  *
  * function gotFile(file) {
- *   img = createImg(file.data).hide();
+ *   img = createImg(file.data, 'a user-uploaded file').hide();
```
lmccart: same thing here as comment on handleFile above

src/dom/dom.js
outofambit: oh good catch. so i looked at the image and literally couldn't figure out what i was looking at 😞, so i used the test image text. i just looked up "la defense" and its a district in paris, so maybe this is a cropped photo of some buildings there? @stalgiag do you have any idea? happy to make this text better, just don't know what this image is...

src/dom/dom.js
outofambit: ohh i did not know that! that's great for this case

src/dom/dom.js
lmccart: lol I couldn't say. in this case, since it's an example of alt text, let's swap the image to something more recognizable to provide a better example of text describing an image

outofambit: that makes so much sense lol

stalgiag: I tried 'search by image' out of curiosity and couldn't find another example of that asset on the internet. @lmccart [mysteriously added it 5 years ago](https://github.com/processing/p5.js/commits/master/docs/yuidoc-p5-theme/assets/laDefense.jpg), and now is suggesting replacing it. I don't like to throw the word conspiracy around but ...

lmccart: woah there if people are whispering conspiracy @REAS must’ve had something to do with it [..](https://github.com/processing/processing-docs/issues/605)

REAS: Uhhhh, it's La Grande Arche de la Défense: https://en.wikipedia.org/wiki/Grande_Arche. Maybe should be a kitten instead?

outofambit: ohhhhhh! wow i would never have figured that out. i thought about deleting the image from the repo to keep the conspiracy going, but we use it in many other examples.

animated gif of a pyramid and an eye above it a lá the illuminati or freemasons (or some other secret society)

for this instance, i've changed it to a wonderful photo of a mountain in https://github.com/processing/p5.js/pull/3890/commits/c5aeea79c81af6c4aaf640a3a3b98b62500ef3f4! yay!

Issue #3892, opened 2019-07-15
webgl/loadSTL manual test is broken in Chrome on Window


stalgiag:
Currently in 'master' the manual test is broken in Chrome 75.0.3770.100 with Windows 10. It works fine in Firefox. The bug seems to occur with both the binary and ascii models. I have tried testing them separately both have the same error.

The error message is:
`GL ERROR :GL_INVALID_OPERATION : glDrawElements: attempt to access out of range vertices in attribute 2`

Which usually means that the buffer for the faces (fill) is not being constructed correctly during loading. [This seems similar to a bug](https://github.com/processing/p5.js/pull/3675#issuecomment-497970726) that appeared to be fixed before merging this feature.

any ideas @sanketsingh24?

aferriss:
@stalgiag @sanketsingh24 I'm 99% certain that the problem comes from the uv buffer not being filled. We just need to fill it up with dummy data as in that comment linked above.

PR #3895, opened 2019-07-16
Fix texture error in STL loader

sanketsingh24:
Fixes: #3892 

Pushing `[0,0]` to `model.uvs` everytime a new vertex is pushed in, as explained [here](https://github.com/processing/p5.js/pull/3675#issuecomment-497970726)

Edit:I don't have a windows machine with me, so I couldn't test this.


aferriss:
Looks good to me.

PR #3905, opened 2019-07-21
Add promise-based preload mechanism

meiamsome:
Part of #2698 - implement a new promise-based preload mechanism.

Related PR which contains some relevant discussion: #3507 

This PR adds a promise-based preload mechanism, along with some tests to assert that the functionality works correctly. Note that due to the tests not being isolated from each other it is hard to test the functionality where these are applied to the `p5.prototype` object, as that is only run once per page. The tests for individual preloads should cover this gap.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -5,6 +5,7 @@ import './core/environment';
 import './core/error_helpers';
 import './core/helpers';
 import './core/legacy';
+import './core/preload';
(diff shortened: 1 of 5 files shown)


outofambit:
hi @meiamsome, i'm going to look at this in the next day or two!

PR #3920, opened 2019-07-28
Converted to ES6 test/node/helpers.js 

abhinavsagar:


test/node/helpers.js
```diff
@@ -1,12 +1,12 @@
-var { expect } = require('chai');
-var constants = require('../../src/core/constants.js');
+import expect from 'chai';
```
outofambit: we have to import `expect` like this:
```suggestion
import { expect } from 'chai';
```

test/node/helpers.js
```diff
@@ -1,12 +1,12 @@
...
+import constants from '../../src/core/constants.js';
```
outofambit: this is how we import all the constants into one object, since each of the individual values are exported individually
```suggestion
import * as constants from '../../src/core/constants.js';
```

abhinavsagar:
Thanks, this was really helpful.

outofambit:
thanks you!

PR #3935, opened 2019-08-06
Fix spelling typos

abhinavsagar:


stalgiag:
Thanks for the contribution!

The apostrophe in "doesn't" is exiting the string. That is why the Travis build is failing. You can escape the apostrophe by putting a backward slash \ before it or it is probably more clear to simply change that word to "does not". It is always welcome to do small typo fixes like this but I think that the unit test messages are a fairly casual space since they are only seen during development, so I wouldn't overthink it too much. Definitely the priority with typos is in public facing text such as documentation and references.

stalgiag:
Thank you!

Issue #3900, opened 2019-07-18
Masks on p5.Graphics


figraham:
#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [X] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [X] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### New feature details:

I was working on a sketch and I realized that you can't mask a p5.Graphics with another one. This is different from what I'm used to in Processing because PGraphics inherits from PImage. It would be a great feature to add so I created a function based on the code in p5.Image. [Here's an example sketch running in web editor](https://editor.p5js.org/figraham/sketches/vOUjYwE3Z).

```
p5.Graphics.prototype.mask = function(inputMask) {
  
  if (inputMask === undefined) {
    inputMask = this;
  }
  const currBlend = this.drawingContext.globalCompositeOperation;

  let scaleFactor = 1;
  if (inputMask instanceof p5.Renderer) {
    scaleFactor = inputMask._pInst._pixelDensity;
  }

  const copyArgs = [
    inputMask,
    0,
    0,
    scaleFactor * inputMask.width,
    scaleFactor * inputMask.height,
    0,
    0,
    this.width,
    this.height
  ];

  this.drawingContext.globalCompositeOperation = 'destination-in';
  p5.Renderer2D.prototype.copy.apply(this, copyArgs);
  this.drawingContext.globalCompositeOperation = currBlend;

}
```

**Is there a reason this hasn't been added yet?** Should I work on writing documentation and then make a pull request with it?

Related to #2841

outofambit:
This looks pretty interesting to me @figraham. Thanks for writing this up (and having a demo!). I'm not aware of any reasons this hasn't been done already but I'll let others (esp. @lmccart and @stalgiag) chime in if there is extra context around this.

figraham:
Sweet! The two things I haven't checked out is a webgl p5.Graphics and at other pixel densities. So I'll try out those two scenarios sometime this week to see if more work would be necessary to make this solution work for everything.

figraham:
[...]

PR #3936, opened 2019-08-06
Fix copy so that it works with a 3D graphics object destination

### LINKED ISSUE #3932: copy() broken on p5.Graphics for WebGL
#### Nature of issue?

- [X] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL

- [X] Desktop/Laptop

#### Details about the bug: 

- p5.js version: 0.9.0
- Web browser and version: Firefox 68.0.1 (64-bit) and Chromium 76.0.3809.87
- Operating System: Linux - Zorin OS 15
- Steps to reproduce this: Run the following code. Or view it in the [web editor](https://editor.p5js.org/figraham/sketches/RH_DbTZij). Will be reproduced when `copy()` function is called through a WebGL p5.Graphics object. Example: `webglGraphics.copy(any valid args will reproduce)`

Following up on @stalgiag suggestion to branch this part of #3900 out in to it's own issue.

`copy()` is undocumented method on p5.Graphics that gets added through [this block of code](https://github.com/processing/p5.js/blob/9748945aa55342944fce9b7244949c59f56eddf3/src/core/p5.Graphics.js#L34).

The original function comes from [here](https://github.com/processing/p5.js/blob/ca92fc457ff836e293916df06964e85f83182368/src/image/pixels.js#L229) and eventually calls a hidden method in Render2D found [here](https://github.com/processing/p5.js/blob/ca92fc457ff836e293916df06964e85f83182368/src/core/p5.Renderer2D.js#L239).

The error that gets thrown is `TypeError: dstImage.drawingContext.drawImage is not a function (sketch: line 26)`. While `dstImage.drawingContext` is defined as a `WebGLRenderingContext` it doesn't have the `drawImage()` function defined. This is where I get lost I can't quite figure out where the `WebGLRenderingContext` is coming from.

This issue might be too far from the current structure of p5.js to get it fixed in a way that makes sense.

```js
let bgGraphics;
let frontGraphics;

function setup() {
  createCanvas(400, 400);
  bgGraphics = createGraphics(width, height, WEBGL);
  frontGraphics = createGraphics(width, height, WEBGL);
}

function draw() {
  
  background(0);
  
  bgGraphics.clear();
  frontGraphics.clear();
  
  bgGraphics.normalMaterial();
  bgGraphics.push();
  bgGraphics.rotateZ(frameCount * 0.01);
  bgGraphics.rotateX(frameCount * 0.01);
  bgGraphics.rotateY(frameCount * 0.01);
  bgGraphics.torus(70, 20);
  bgGraphics.pop();
  //image(bgGraphics, 0, 0);
  
  frontGraphics.copy(bgGraphics, 0, 0, width/2, height/2, 0, 0, width/2, height/2);
  
  frontGraphics.normalMaterial();
  frontGraphics.push();
  frontGraphics.rotateZ(frameCount * 0.01);
  frontGraphics.rotateX(frameCount * 0.01);
  frontGraphics.rotateY(frameCount * 0.01);
  frontGraphics.cone(70, 70);
  frontGraphics.pop();
  image(frontGraphics, 0, 0);
  
}
```
Expected output of above sketch is a cone overlayed on a torus.

stalgiag:
closes #3932 

So there are some quirks with this, thus the draft status for the PR. Basically I tried to find the minimal intervention way to get copy working for WebGL graphics objects. WebGL drawing contexts do not have a way of placing pixel data directly into them, so instead `image` is used to simply create geometry and assign a texture to that geometry. The minimal intervention approach meant that I ended up adding the logic for handling 3D destinations inside of the Renderer2D prototype for copy. This is a bit counter-intuitive so may need to go back further on the chain but the fix would then require a significant amount more code much of which would be redundant.

This seems to work in the most obvious cases including the one listed in the above issue but I only tested for a few things. @figraham would you mind looking this over and testing in a few different ways? I anticipate there will be problems but not sure where to start so help is appreciated. Thanks!

### DIFF
--- a/test/unit/image/pixels.js
+++ b/test/unit/image/pixels.js
@@ -228,7 +228,7 @@ suite('pixels', function() {
       assert.throw(function() {
         let img = myp5.createImage(50, 50);
         img.copy(0, 0, 10, 10, 0, 10, 10);
-      }, 'Signature not supported');
+      });
     });
   });
 });
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -250,11 +250,15 @@ p5.Renderer2D._copyHelper = (
 ) => {
   srcImage.loadPixels();
   const s = srcImage.canvas.width / srcImage.width;
-  const sxMod = srcImage._renderer.isP3D ? srcImage.width / 2 : 0;
-  const syMod = srcImage._renderer.isP3D ? srcImage.height / 2 : 0;
+  // adjust coord system for 3D when renderer
+  // ie top-left = -width/2, -height/2
+  let sxMod = 0;
+  let syMod = 0;
+  if (srcImage._renderer && srcImage._renderer.isP3D) {
+    sxMod = srcImage.width / 2;
+    syMod = srcImage.height / 2;
+  }
   if (dstImage.isP3D) {
-    // adjust coord system for 3D
-    // ie top-left = -width/2, -height/2
     p5.RendererGL.prototype.image.call(
       dstImage,
       srcImage,
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -250,17 +250,32 @@ p5.Renderer2D._copyHelper = (
 ) => {
   srcImage.loadPixels();
   const s = srcImage.canvas.width / srcImage.width;
-  dstImage.drawingContext.drawImage(
-    srcImage.canvas,
-    s * sx,
-    s * sy,
-    s * sw,
-    s * sh,
-    dx,
-    dy,
-    dw,
-    dh
-  );
+  if (dstImage.isP3D) {
+    p5.RendererGL.prototype.image.call(
+      dstImage,
+      srcImage,
+      sx,
+      sy,
+      sw,
+      sh,
+      dx,
+      dy,
+      dw,
+      dh
+    );
+  } else {
+    dstImage.drawingContext.drawImage(
+      srcImage.canvas,
+      s * sx,
+      s * sy,
+      s * sw,
+      s * sh,
+      dx,
+      dy,
+      dw,
+      dh
+    );
+  }
 };
 
 // p5.Renderer2D.prototype.get = p5.Renderer.prototype.get;
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -250,12 +250,16 @@ p5.Renderer2D._copyHelper = (
 ) => {
   srcImage.loadPixels();
   const s = srcImage.canvas.width / srcImage.width;
+  const sxMod = srcImage._renderer.isP3D ? srcImage.width / 2 : 0;
+  const syMod = srcImage._renderer.isP3D ? srcImage.height / 2 : 0;
   if (dstImage.isP3D) {
+    // adjust coord system for 3D
+    // ie top-left = -width/2, -height/2
     p5.RendererGL.prototype.image.call(
       dstImage,
       srcImage,
-      sx,
-      sy,
+      sx + sxMod,
+      sy + syMod,
       sw,
       sh,
       dx,
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -270,8 +270,8 @@ p5.Renderer2D._copyHelper = (
   } else {
     dstImage.drawingContext.drawImage(
       srcImage.canvas,
-      s * sx,
-      s * sy,
+      s * (sx + sxMod),
+      s * (sy + syMod),
       s * sw,
       s * sh,
       dx,
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -198,92 +198,10 @@ p5.Renderer2D.prototype.blend = function(...args) {
   const copyArgs = Array.prototype.slice.call(args, 0, args.length - 1);
 
   this.drawingContext.globalCompositeOperation = blendMode;
-  if (this._pInst) {
-    this._pInst.copy(...copyArgs);
-  } else {
-    this.copy(...copyArgs);
-  }
+
+  p5.prototype.copy.apply(this, copyArgs);
+
   this.drawingContext.globalCompositeOperation = currBlend;
-};
-
-p5.Renderer2D.prototype.copy = function(...args) {
-  let srcImage, sx, sy, sw, sh, dx, dy, dw, dh;
-  if (args.length === 9) {
-    srcImage = args[0];
-    sx = args[1];
-    sy = args[2];
-    sw = args[3];
-    sh = args[4];
-    dx = args[5];
-    dy = args[6];
-    dw = args[7];
-    dh = args[8];
-  } else if (args.length === 8) {
-    srcImage = this._pInst;
-    sx = args[0];
-    sy = args[1];
-    sw = args[2];
-    sh = args[3];
-    dx = args[4];
-    dy = args[5];
-    dw = args[6];
-    dh = args[7];
-  } else {
-    throw new Error('Signature not supported');
-  }
-  p5.Renderer2D._copyHelper(this, srcImage, sx, sy, sw, sh, dx, dy, dw, dh);
-
-  this._pixelsState._pixelsDirty = true;
-};
-
-p5.Renderer2D._copyHelper = (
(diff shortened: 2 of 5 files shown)


aferriss:
@stalgiag I have been wondering about situations like this and other issues related to using `createGraphics()`. Is there a reason why we didn't use framebuffers and instead keep instantiating new webgl contexts? 

Maybe there's something I'm not seeing but it feels like using a webGL framebuffer might simplify some of these problems and be more performant.

stalgiag:
@aferriss interesting suggestion! I am not entirely sure. I think that until this point, webgl graphics objects have been underdeveloped and mostly non-functioning.

I think that the reason that framebuffers haven't been used is because of the existing situation in which graphics objects are similar to p5 objects and are oriented around a renderer which initializes a webgl context upon construction.

A framebuffer would definitely be a lot more performant, but maybe it would require the code to be too different between a GL p5 instance and GL p5 graphics object? I am not entirely sure.

Currently the copy function is using `loadPixels` which does use gl.readPixels. I agree though that it doesn't make sense to keep instantiating webgl contexts.

But I guess one way around a lot of these is problems is if we can guarantee that there is at least one rendering context? Just to understand whether I am following, do you think that in a situation like this:

```
let glGraphics1, glGraphics2;
function setup() {
  createCanvas(100,100);
  glGraphics1 = createGraphics(100, 100, WEBGL);
  glGraphics2 = createGraphics(100, 100, WEBGL);
  glGraphics2.torus(100);
}
```
that the first gl graphics created would create a kind of master gl rendering context and that rendering context would handle everything for rendering the torus out to the framebuffer that is glGraphics2? So the first time any WebGL renderer is made, a context is made, and everything that is reliant on it, whether it is the main canvas or a graphics object, is just a framebuffer?

Also another thing to consider is how much of an edge case it is for someone to have multiple different webgl graphics objects. It might be the case that once someone is doing work that would require multiple framebuffers simultaneously working in a performant fashion that they should be starting to transition towards something like three.js. Not sure!

aferriss:
@stalgiag Ah that's a very good point re: context creation, I hadn't considered that. I bet something like you propose would work but it sounds messy to implement. 

I may be biased but I personally don't think it's such an edge case to have a few of them. Any kind of multi-pass shadery effects (bloom for example) would require having a couple extra graphics objects. I actually think p5 is a nice environment to learn about off-screen buffers and simple shader effects since you can get them setup and running with just a couple lines vs. the many more you'd need in something like three.js. That being said I don't think many people are using p5 for this sort of thing at the moment. 

Anyhow the framebuffer idea was more a "hmm I wonder" kinda thing, so thank you for taking the time to write out a thoughtful response!

figraham:
I've been thinking about this more, and I had the thought that this is probably introducing unnecessary complexity into p5 and going against the way webgl is currently structured.

An alternative solution would be to just have a 2D graphics as an intermediary because as it is `2dGraphics.copy(webglGraphics, ...)` works perfectly. (as seen in [this](https://editor.p5js.org/figraham/sketches/ceXmOkXFf) proof of concept) That way it takes out the complexity and it could be done behind the scenes in a separate library. This might also be a good way to expand on p5.Graphics with methods like `mask()` as described in #3900.

I don't want to discount the work @stalgiag has done but I'm thinking of closing #3900 and #3932 working on library in the couple weeks I have off. That way it can still be accessible to users but doesn't affect the existing structure of p5. I'd love to hear your thoughts @stalgiag, @outofambit and @aferriss.

stalgiag:
@aferriss I am still interested in this. I do think using framebuffers follows normal WEBGL development standards a little more closely. At the moment I am considering opening some issues regarding some fairly structural changes to webgl. Any thoughts like this one are extremely welcome and I'd be happy to talk about whether it makes sense to do a large-scale rewrite to orient the logic around something like this.

Hi @figraham no worries! This work wasn't too involved. Mostly was just a matter of realizing that geometry would need to be made to display pixel data, and that `image()` was already more or less suited. That said I feel that copy should probably work out of the box with WebGL to try and create as much parity as possible between 2D/3D. So that makes this PR relevant still. Unless I am missing something?

PS I am not sure about the blank areas turning black instead of transparent but I can investigate if it makes sense to push forward with this PR.

figraham:
Sounds good. Sorry I didn't realize i had to approve it. I'll close #3900 but I'll leave #3932 open because it relates to this pull request.

As is I think it's alright but the transparency would be nice.

stalgiag:
Okay all the variations on `copy()` work with correct mode-specific coord system with this branch now. I am pasting the code that I was using to test here, it is a bit of an arduous read but the gist is that all four of those functions draws the exact same canvas. This should create the expected behavior between the two modes.

```
let pg;

function setup() {
  // All of these functions draw the same canvas image
  // twoCopyIntoThree();
  // threeCopyIntoThree();
  threeCopyIntoTwo();
  // twoCopyIntoTwo();
}

function twoCopyIntoTwo() {
  createCanvas(400, 400);
  background(150);
  pg = createGraphics(200, 200);
  pg.background(0);
  pg.ellipse(pg.width / 2, pg.height / 2, 100); // center ellipse in coord system
  copy(pg, 0, 0, pg.width, pg.height, 0, 0, pg.width, pg.height); // dest and src use 2D coords
}

function threeCopyIntoTwo() {
  createCanvas(400, 400);
  background(150);
  pg = createGraphics(200, 200, WEBGL);
  pg.background(0);
  pg.ellipse(0, 0, 100);
  copy(
    pg,
    -pg.width / 2, //source is 3D so we use 3D coord system
    -pg.height / 2,
    pg.width,
    pg.height,
    0, //dest is 2D so we use 2D coord system
    0,
    pg.width,
    pg.height
  );
}

function twoCopyIntoThree() {
  background(150);
  pg = createGraphics(200, 200);
  pg.background(0);
  pg.ellipse(100, 100, 100);
  copy(
    pg,
    0, // source is 2D so we use 2D coord system
    0,
    pg.width,
    pg.height,
    -width / 2, // dest is 3D so we 3D coord system
    -height / 2,
    pg.width,
    pg.height
  );
}

function threeCopyIntoThree() {
  background(150);
  pg = createGraphics(200, 200, WEBGL);
  pg.background(0);
  pg.ellipse(0, 0, 100);
  copy(
    pg,
    -pg.width / 2, //source and dest are 3D so we use 3D coord system for both
    -pg.height / 2,
    pg.width,
    pg.height,
    -width / 2,
    -height / 2,
    pg.width,
    pg.height
  );
}
```

One thing that is still up for debate is whether I should move this up the chain in some fashion. Maybe by putting the `_copyHelper` prototype at the `p5` level?. Right now `copy` is routed through the `Renderer2D` prototype even when no `Renderer2D` is involved (such as with GL source and destination). This is against the design principles of the code base but was the easiest way to get visibly working examples.  I should probably shift this up the hierarchy so that everything is handled by the `p5` prototypes to make it easier to understand that this method manages the cases for both renderers.

test/unit/image/pixels.js
```diff
@@ -228,7 +228,7 @@ suite('pixels', function() {
       assert.throw(function() {
         let img = myp5.createImage(50, 50);
         img.copy(0, 0, 10, 10, 0, 10, 10);
-      }, 'Signature not supported');
+      });
```
stalgiag: This error message is thrown but the detection of the content of the error doesn't work because the [parameter validation also logs a message.](https://github.com/processing/p5.js/pull/3936/files#diff-fecaef9926a3d9ed7f9ac61c8e72e2f0R230). Removing the optional message check works fine.

stalgiag:
So this now works in all cases that I can find, including the above tests, all unit tests, and I visually confirmed that all of the reference examples work as before.

Moving the work up to `p5.prototype.copy()` was a bit odd because different objects were using the `p5.Renderer2D.prototype.copy` in different ways by modifying the context or the arguments. It was a bit hard to work my head around.

I have a few ideas about some rules for when work should be delegated to the Renderer level of the hierarchy that may help avoid these kinds of tangles in the future. I will gather notes and pass them along this week.

lmccart:
@stalgiag sorry for the delay. This looks good to me. Moving `copy()` up makes total sense, I think the fact that it was attached to `p5.Renderer2D` was legacy from before we had webgl mode. It's still a little hacky as you note, but it makes more sense at the top level than nested in 2D at least.

Issue #3941, opened 2019-08-10
Documentation for `setAlpha()`


mbforbes:
#### Nature of issue?

- [ ] Found a bug
- [x] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [x] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Feature enhancement details:

I was reading the documentation for [`setAlpha()`](https://p5js.org/reference/#/p5.Color/setAlpha) on `p5.Color`. It would be nice if the value range was specified on this page in the documentation. I needed to read through the example to find out.

Thank you for your consideration!

AndrasGardos:
I want to do this! It's between 0 and 255 right?

AndrasGardos:
@mbforbes, it depends on the color mode. For example, if you set `colorMode(HSB,100);` in the setup,  then 100 will be totally opaque.

AndrasGardos:
https://github.com/processing/p5.js/pull/3942

mbforbes:
Thank you for jumping on this!! So sorry for my slow reply. Adding to the other setter methods as well is a great idea. Awesome you already have a PR in review. Thank you @AndrasGardos !

lmccart:
fixed with https://github.com/processing/p5.js/pull/3942

PR #3942, opened 2019-08-12
Explain ranges of setRed/Green/Blue/Alpha

AndrasGardos:
Adding explanation for ranges of color to some functions.

https://github.com/processing/p5.js/issues/3941

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -249,8 +249,10 @@ p5.Color.prototype.toString = function(format) {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


lmccart:
@AndrasGardos this is great! but I think this text would be better suited to go in the description for each `setX()` method rather than at the parameter listing. would you make this change?

AndrasGardos:
@lmccart Thanks, good idea! Added some description (there was none) along with the explanation.

Issue #3897, opened 2019-07-17
p5js Reference should include entries for JavaScript basics


brysonian:
The reference page on p5js.org seems to assume a basic knowledge of javascript because there aren't pages for things variables and functions. I think it would be hugely helpful to at the very least link to the MDN pages.

I imagine a number of students get lost not knowing where to look when they do get confused about functions, variables, etc.

For context, the processing reference includes pages for int, float, boolean... as well as for, while, etc. 

(see also: #3877)

ProjectGrantwood:
I'm not sure the site needs a whole new set of new reference pages--mostly because there is so much going on in modern javascript, every new reference could be a potential rabbit hole.

However, a one-page reference that differentiates what is the p5js library keywords/syntax from what is pure javascript would be great--with the appropriate links to MDN reference pages, of course. This would be ultra helpful for novice users and probably get to the core of the problem you're expressing.

As a side note, I believe that the original Processing was intended to function more like its own language, not "just a java library"--which is probably why we see int, float, bool, for/while on the Processing reference pages. Somebody correct me if I'm wrong.

brysonian:
That is true about processing, however pedagogically p5 and processing are taught in very much the same way. Most books and syllabi on learning p5 feature sections on functions and variables and iteration, and most learners are not experienced JS coders (or coders at all!) so the library/language distinction isn't a particularly useful one for their needs. In other words, we don't teach JS, then teach p5, they are taught together at the same time.

It seems like a big ask for a new learner, after just a couple of classes, to be expected to know that when they need help with the syntax of `for` loop they need to search for "javascript for loop" or look at MDN (that's overwhelming) but when they want to draw an ellipse, they can look at the online reference. 

There is no need or use in documenting all of modern javascript but that doesn't mean that there isn't a need to provide help on some fundamentals. I also think we are capable of describing how to use `for` in much more supportive and helpful terms than MDN. 

Also note that many of the functions that reproduce built-in JS functionality like [append](http://p5js.org/reference/#/p5/append) are deprecated, which I think is great, but will further exacerbate the issue of how to help new learners find what they need.

stalgiag:
[...]

PR #3953, opened 2019-08-19
Added a doc file(#3897)

singhvisha:
Added a reference.js file that contain explanation of few js keyword and their example.

(#3897)

### DIFF
--- a/src/core/reference.js
+++ b/src/core/reference.js
@@ -1,10 +1,16 @@
 /**
- * file contains few JS library keywords and their examples.
+ * @module Foundation
+ * @submodule Declaration
+ * @for p5
  */
-/*
- * @Decleration  Let
+
+
+/**
+
  * Declares a block scope local variable, optionally initializing it to a value.
  *
+ * @property let
+ * 
  * Syntax:
  * let var1 = value1, var2 = value2
  * @params var1, var2, …, varN
@@ -28,11 +34,11 @@
 
 /**
  *
- * @Decleration const
- *
  * Declares a read-only named constant.
  * Constants are block-scoped, much like variables defined using the 'let' statement.
  * The value of a constant can't be changed through reassignment, and it can't be redeclared.
+ * 
+ * @property const
  *
  * Syntax:
  * const name1 = value1, name2 = value2, ... nameN = valueN;
@@ -49,8 +55,17 @@
  * </div>
  */
 
-/*
- * @ControlStatement  if..else
+ /**
+ * @submodule Control Statement
+ */
+ 
+
+/**
+ * 
+ * The if statement executes a statement if a specified condition is truthy.
+ * If the condition is falsy, another statement can be executed
+ * 
+ * @property if-else
  *
  * syntax:
  * if(condition1)
@@ -77,11 +92,12 @@
  * </div>
  */
 
-/*
- * @Decleration function
- *
+/**
+ * 
  * Declares a function with the specified parameters.
  *
+ * @property function
+ * 
  * syntax:
  * function name(param,param,..., param){
  *    statements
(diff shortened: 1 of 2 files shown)


outofambit:
this is looking good! there's a couple formatting things we need to fix for the tests + linter to pass and i can help with that (probably in the next couple days)!

limzykenneth:
The `package-lock.json` file is changed in this PR which is not expected. @singhvisha Did the npm version you had made those changes upon running `npm install`?

stalgiag:
If it is helpful I find that using `npm ci` ([info link here](https://docs.npmjs.com/cli/ci.html)) instead can help when `npm install` is unexpectedly modifying the `package-lock.json` not sure if that is the precise difficulty here but just a thought.

Other than the problem with the package-lock this is looking pretty exciting. We will probably need to do a collective editing pass to make sure that the language is as accessible to beginners as possible. This is a really great beginning.

limzykenneth:
Upon further inspection, the current `package-lock.json` is out of sync with `package.json` and so when running `npm install` it will repopulate `package-lock.json` with new data. I think the version committed here is correct but it's hard to trace down why the previous changes were made (most came from different commits). We may need to have a round of dependency updates along with `npm audit` very soon.

I'm ok to merge this PR as is.

src/core/reference.js
```diff
@@ -0,0 +1,321 @@
+/**
+ * file contains few JS library keywords and their examples.
+ */
+/*
+ * @Decleration  Let
```
stalgiag:
```suggestion
 * @Declaration  Let
```
Small typo throughout. See [this page for spelling](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements).

src/core/reference.js
```diff
@@ -0,0 +1,321 @@
...
+ * Declares a read-only named constant.
+ * Constants are block-scoped, much like variables defined using the 'let' statement.
+ * The value of a constant can't be changed through reassignment, and it can't be redeclared.
+ *
+ * Syntax:
+ * const name1 = value1, name2 = value2, ... nameN = valueN;
+ * @params name1,name2 ...,nameN   //The constant's name, which can be any legal identifier.
+ * value1,value2 ...,valueN    //The constant's value; this can be any legal expression, including a function expression.
+ *
+ * @exmaple
```
stalgiag:
```suggestion
 * @example
```

stalgiag:
Hi okay as promised I played around a bit and managed to figure out how to get this building the docs as desired. I am copying a small example here:
```
/**
 * @module Foundation
 * @submodule Declaration
 * @for p5
 */

/**
 *
 *
 * @property  let
 *
 * @example
 * <div class="norender">
 * <code>
 * let myVariable = 1;
 * print(myVariable); // writes 1 to the console
 * myVariable = 2;
 * print(myVariable) // writes 2 to the console
 *
 * </code>
 * </div>
 *
 */

/**
 * @submodule Control Statement
 */

/**
 *
 * The if statement executes a statement if a specified condition is truthy.
 *
 * @property  if-else
 *
 * @example
 * <div class="norender">
 * <code>
 * if (a > 0) {
 *  print('positive');
 * } else {
 *  print('negative');
 * }
 * </code>
 * </div>
 *
 */
```
If you put this into `reference.js` and run `npm run docs:dev` you should see the docs at your `localhost:9001`. These are separated by module and submodule. The naming of the module and submodule is up for discussion but I just wanted to get you started with something that builds. In general, I would say try to keep both language and examples as simple as possible which is something that @outofambit can probably help us figure out. Also, I would say that generally this PR doesn't need to be in its absolute perfect final form before merging since others such as @shubymao want to help with this work but it is important to figure out how to make sure that it builds correctly. Would you mind putting your work into a format similar to the one above and testing that you can see everything when running `npm run docs:dev`? Thank you for all of the hard work!

singhvisha:
hi @stalgiag,what you want to reformat the code as you prescribed above and create an another Pull Request.

stalgiag:
Hi @singhvisha whichever is easier for you! You can close this pull request and open another **or** you can modify the `core/reference.js` file and push to this branch.

Issue #3950, opened 2019-08-18
combineModules (custom build) output is not minified


coreygo:
#### Nature of issue?

- [X] Found a bug
- [X] Existing feature enhancement

#### Most appropriate sub-area of p5.js?

- [X] Other (build system)

#### Which platform were you using when you encountered this?

- [X] Desktop/Laptop

#### Details about the bug: 

- p5.js version: v0.9.0 (2f85dda)
- Web browser and version: N/A
- Operating System: MacOS 10.14.6
- Steps to reproduce this:
`npm run grunt combineModules:core/shape:color:math:image`  
which outputs a 2.5M `lib/modules/p5Custom.js`  
compared to the full `p5.min.js` at 543K

#### Feature enhancement details:
I've [created and committed a potential update](https://github.com/processing/p5.js/commit/16ba23368ccf3157e27e4720f56c06e5bad029db) to `tasks/build/combineModules.js` which enables minification using uglify similar to the existing Grunt task for browserify and uglify.

- Steps tested after update:
  - `npm run grunt combineModules:core/shape:color:math:image`  
which outputs a 2.5M `lib/modules/p5Custom.js`

  - `npm run grunt combineModules:min:core/shape:color:math:image`  
which outputs a 1.3M `lib/modules/p5Custom.pre-min.js`

  - `npm run grunt combineModules:min:core/shape:color:math:image uglify`  
which outputs a **207K** `lib/modules/p5Custom.min.js`

The full minified `p5.min.js` is **543K**, `p5.pre-min.js` is 2.9M, and `p5.js` is 4M.

After discussion, I'm happy to create a new pull request and will update the `custom_p5_build.md` doc before doing so.

This could help alleviate any wait on larger build system updates enabling smaller bundles.

coreygo:
Tagging #522 and #3425 as related issues.

coreygo:
I should note that my build can't be merged as-is right now, it breaks things overall… was testing a quick workaround. Will update with something which can be merged.

coreygo:
Okay, I've [pushed an update](https://github.com/coreygo/p5.js/commit/0de9bc4e011332f9677a820fdc212a21fa4ba1a3) (0de9bc4) which seems to do the trick without breaking anything (as far as I can tell). If there's a better way to improve the Grunt tasks enabling "uglified browseified minified" output :wink: for custom builds please let me know. I'm not familiar enough with Grunt or the p5.js build system so might be missing something obvious. 

[...]

PR #3956, opened 2019-08-19
Add combineModules:min enabling minified output with uglify

coreygo:
Addresses: #3950 

  - `npm run grunt combineModules:min:core/shape:color:math:image uglify`  
which outputs a **207K** `lib/modules/p5Custom.min.js`

The full minified `p5.min.js` is **543K**, `p5.pre-min.js` is 2.9M, and `p5.js` is 4M, while `lib/modules/p5Custom.js` is 2.5M.

The custom p5 build doc has also been updated.


coreygo:
My pleasure @outofambit. Closing #3950 now.

PR #3978, opened 2019-08-29
Add some plain JS reference examples

singhvisha:
Formatting the doc file as mentioned.

### DIFF
--- a/src/core/reference.js
+++ b/src/core/reference.js
@@ -1,10 +1,16 @@
 /**
- * file contains few JS library keywords and their examples.
+ * @module Foundation
+ * @submodule Declaration
+ * @for p5
  */
-/*
- * @Decleration  Let
+
+
+/**
+
  * Declares a block scope local variable, optionally initializing it to a value.
  *
+ * @property let
+ * 
  * Syntax:
  * let var1 = value1, var2 = value2
  * @params var1, var2, …, varN
@@ -28,11 +34,11 @@
 
 /**
  *
- * @Decleration const
- *
  * Declares a read-only named constant.
  * Constants are block-scoped, much like variables defined using the 'let' statement.
  * The value of a constant can't be changed through reassignment, and it can't be redeclared.
+ * 
+ * @property const
  *
  * Syntax:
  * const name1 = value1, name2 = value2, ... nameN = valueN;
@@ -49,8 +55,17 @@
  * </div>
  */
 
-/*
- * @ControlStatement  if..else
+ /**
+ * @submodule Control Statement
+ */
+ 
+
+/**
+ * 
+ * The if statement executes a statement if a specified condition is truthy.
+ * If the condition is falsy, another statement can be executed
+ * 
+ * @property if-else
  *
  * syntax:
  * if(condition1)
@@ -77,11 +92,12 @@
  * </div>
  */
 
-/*
- * @Decleration function
- *
+/**
+ * 
  * Declares a function with the specified parameters.
  *
+ * @property function
+ * 
  * syntax:
  * function name(param,param,..., param){
  *    statements
@@ -102,11 +118,11 @@
  *
  */
 
-/*
- *
- * @Decleration return
+/**
  *
  * Specifies the value to be returned by a function.
+ * 
+ * @property return
  *
  * syntax:
  * return expression;
@@ -124,11 +140,12 @@
  *
  */
 
-/*
+/**
  *
- * @Decleration class
  *
  * The class declaration creates a new class with a given name using prototype-based inheritance.
+ * 
+ * @property class
  *
  * syntax:
  *  class name{
@@ -150,12 +167,13 @@
  *
  */
 
-/*
- * @Iteration do..while
+/**
  *
  * Creates a loop that executes a specified statement until the test condition evaluates to false.
  * The condition is evaluated after executing the statement, resulting in the specified statement executing at least once.
  *
+ * @property  do..while
+ * 
  * syntax:
  * do{
  *  statement
@@ -179,13 +197,12 @@
  *
  */
 
-/*
- *
- * @Iteration for
+/** 
  *
  * Creates a loop that executes a specified statement until the test condition evaluates to false.
  * The condition is evaluated after executing the statement, resulting in the specified statement executing at least once.
  *
+ * @property for
  *
  * syntax:
  * for (initialization; condition; final-expression)
@@ -209,11 +226,12 @@
  *
  */
 
-/*
- * @Iteration  for...in
+/**
  *
  * Iterates over the enumerable properties of an object, in arbitrary order.
  * For each distinct property, statements can be executed.
+ * 
+ * @property for..in
  *
  * syntax:
  * for (variable in object)
@@ -241,12 +259,13 @@
  *
  */
 
-/*
- * @Iteration for...of
+/**
  *
  * Iterates over iterable objects (including arrays, array-like objects, iterators and generators),
  * invoking a custom iteration hook with statements to be executed for the value of each distinct property.
  *
+ * @property for..of
+ * 
  * syntax:
  * for (variable of iterable) {
  *  statement
@@ -270,12 +289,12 @@
  *
  */
 
-/*
- * @method parseInt
- *
+/**
  * The parseInt() function parses a string argument and
  * returns an integer of the specified radix (the base in mathematical numeral systems).
  *
+ * @method parseInt
+ * 
  * syntax:
  * parseInt(string, radix);
  *
@@ -292,10 +311,11 @@
  *
  */
 
-/*
- * @method JSON.stringify
+/**
  *
- * The JSON.stringify() method converts a JavaScript object or value to a JSON string,
+ * The JSON.stringify() method converts a JavaScript object or value to a JSON string.
+ * 
+ * @method JSON.stringify
  *
  * syntax:
  * JSON.stringify(value, replacer, space)
--- a/src/core/reference.js
+++ b/src/core/reference.js
@@ -136,29 +136,9 @@
  */
 
 /**
- * It Creates a loop that executes a specified statement until the test condition evaluates to false.
- * The condition is evaluated after executing the statement, resulting in the specified statement
- * executing at least once.
- * @property  do..while
- *
- * @example
- * <div>
- * <code>
- * var result = '';
- * var i = 0;
- * do {
- *   i += 1;
- *   result += i + ' ';
- * } while (i > 0 && i < 5);
- * console.log(result); //'1 2 3 4 5 '
- * <code>
- * </div>
- */
-
-/**
  * Iterates over the enumerable properties of an object, in arbitrary order.
  * For each distinct property, statements can be executed.
- * @property for..in
+ * @property for-in
  *
  * @example
  * <div>
@@ -177,7 +157,7 @@
 /**
  * Iterates over iterable objects (including arrays, array-like objects, iterators and generators),
  * invoking a custom iteration hook with statements to be executed for the value of each distinct property.
- * @property for..of
+ * @property for-of
  *
  * @example
  * <div>
@@ -210,7 +190,8 @@
 
 /**
  * The JSON.stringify() method converts a JavaScript object or value to a JSON string.
- * @method JSON.stringify
+ * @method stringify
+ * @for JSON
  * @param {Object} object :Javascript object that you would like to convert to JSON
  *
  * @example
--- a/src/core/reference.js
+++ b/src/core/reference.js
@@ -321,9 +321,11 @@
  * JSON.stringify(value, replacer, space)
  *
  * [value]: The value to convert to a JSON string.
+ * 
  * [replacer]  //Optional  :A function that alters the behavior of the stringification process, or an array of String and Number objects 
  * that serve as a whitelist for selecting/filtering the properties of the value object to be included in the JSON string.
  * If this value is null or not provided, all properties of the object are included in the resulting JSON string.
+ * 
  * [space]  //optional :A String or Number object that's used to insert white space into the output JSON string for readability purposes.
  * If this is a Number, it indicates the number of space characters to use as white space; this number is capped at 10 (if it is greater, the value is just 10).
  * Values less than 1 indicate that no space should be used.
--- a/src/core/reference.js
+++ b/src/core/reference.js
@@ -321,7 +321,8 @@
  * JSON.stringify(value, replacer, space)
  *
  * [value]: The value to convert to a JSON string.
- * [replacer]  //Optional  :A function that alters the behavior of the stringification process, or an array of String and Number objects that serve as a whitelist for selecting/filtering the properties of the value object to be included in the JSON string.
+ * [replacer]  //Optional  :A function that alters the behavior of the stringification process, or an array of String and Number objects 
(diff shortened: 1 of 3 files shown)


src/core/reference.js
```diff
@@ -0,0 +1,344 @@
+/**
+ * @module Foundation
+ * @submodule Declaration
+ * @for p5
+ */
+
+
+/**
+
```
stalgiag: The docs won't build correctly if there is a space like this here. There can be no empty lines between a `/**` and a `*/`

src/core/reference.js
```diff
@@ -0,0 +1,344 @@
...
+ * Declares a block scope local variable, optionally initializing it to a value.
+ *
+ * @property let
+ * 
+ * Syntax:
```
stalgiag: Explanatory text needs to go above the `@property, @method, etc` line in order to appear.

src/core/reference.js
```diff
@@ -0,0 +1,344 @@
...
+ * let var1 = value1, var2 = value2
+ * @params var1, var2, …, varN
```
stalgiag: I don't think that `@params` is a tag. The non-plural `@param` can be used but only with docs generated for `@method`. Since the idea of a parameter is a bit confusing when talking about declaration I'd suggest erasing this part and trying to ensure that the assignment of value can be explained in the main text. Again, don't worry about perfecting this. I think the goal of this PR should be to get building happening correctly.

src/core/reference.js
```diff
@@ -0,0 +1,344 @@
...
+ * The value of a constant can't be changed through reassignment, and it can't be redeclared.
+ * 
+ * @property const
+ *
+ * Syntax:
+ * const name1 = value1, name2 = value2, ... nameN = valueN;
+ * @params name1,name2 ...,nameN   //The constant's name, which can be any legal identifier.
+ * value1,value2 ...,valueN    //The constant's value; this can be any legal expression, including a function expression.
+ *
+ * @exmaple
```
stalgiag:
```suggestion
 * @example
```

src/core/reference.js
```diff
@@ -0,0 +1,344 @@
...
+ * <div>
+ * <code>
+ * function letTest(){
+ * let x = 1;
+ * if(true){
+ * let x = 2; //different variable
+ * console.log(x); // 2
+ * }
+ * console.log(x); //1
+ * }
```
stalgiag: So this is a bit tricky but all of the code examples have to be linted properly and they can't be autofixed. But they should all follow the formatting rules. So the code example above would become:
```
 * function letTest() {
 *   let x = 1;
 *   if (true) {
 *     let x = 2; //different variable
 *     console.log(x); // 2
 *   }
 *   console.log(x); //1
 * }
```
This can be tricky but if you run `npm run grunt` in your terminal after running `npm run lint:fix` (see final review comment) you will see a line by line breakdown of all expected formatting changes. Note that in this breakdown, a period '.' represents an empty space.

singhvisha: @stalgiag 
Hi,It is very difficult to track all the changes after running the above command.

singhvisha:
@stalgiag
Hi,It is very difficult to track all the changes after running the above command.

singhvisha:
@stalgiag 
Hi, I reformat the doc as you mentioned.But after running the command **npm run grunt** , It is very
difficult  to track the error.

stalgiag:
Hi @singvisha sorry for the small delay! So one way to look at this problem is to take away all of the examples and add them back one at a time simplified. For example, I know that this will pass the tests run by `npm run grunt`:

```
/**
 * @module Foundation
 * @submodule Declaration
 * @for p5
 */

/**

 *
 * @property let
 *
 * @example
 * <div>
 * <code>
 * let x = 2;
 * console.log(x); // prints 2 to the console
 * x = 1;
 * console.log(x); // prints 1 to the console
 * </code>
 * </div>
 *
 */
```
So try starting from there and adding everything else back in one example at a time. When you add something in, try to simplify the example as much as possible. Then run `npm run grunt` and try to solve the problems with formatting one line at a time. If this becomes too frustrating, remember that it isn't essential to get all of your proposed references examples merged with this PR. If we can get a few good examples to start us off then we can add more in with later pull requests. You are setting up the template for later contributions to follow, so making sure that everything works as expected is the only prerequisite to the acceptance of this PR. Good luck!!

singhvisha:
@stalgiag,
hi,I am really sorry for the delay.
I have formatted the file as you have mentioned above.And Now it is not showing any error while running `npm run grunt`.

singhvisha:
@stalgiag.,
Please review it.

stalgiag:
Hi @singhvisha sorry for the delay!

I was trying to figure out a clear way to explain what is happening here. So first, your branch (this PR) isn't passing the tests. We know this because of the red text that says 'All checks have failed' above'. It isn't your fault for not knowing this, since I am assuming that the failed tests spammed new lines into your console, and you didn't see an error so thought it completed correctly. [This is a problem that we are looking into currently ](https://github.com/processing/p5.js/issues/4000). But the main point is that, the tests are only passing when you see this at the end --

Screenshot (130)

Also it should say 'All checks have passed' in green text above this input field.

So to return to why it isn't passing, I would look again at the level of simplicity that I suggested using.
- Avoid using @param.
- Make every example as simple as possible to begin with, if your examples throw errors, the tests will fail.
- Go through your examples one at a time in the browser by running `npm run docs:dev`, keep the console open, see if there are any errors

If you follow these steps, you should be able to get this branch passing the tests.

Good luck!

stalgiag:
Hello @singhvisha, I believe that the main problem here is the use of `square()` in your examples. p5 also has a `square()` function. To be safe, I am copying a working and building version of your file into a suggestion. You can replace everything in reference with this, and then run `npm run grunt`. This works on my machine. You can also then run `npm run docs:dev` and look through your examples and make sure everything looks as you intend.

src/core/reference.js
```diff
@@ -0,0 +1,222 @@
+/**
```
stalgiag: If you replace the code in this file with this, it will build. Mostly just a changes with a bit of formatting and with the `return` example. You should still do some visual reviewing with `npm run docs:dev`
```
/**
 * @module Foundation
 * @submodule Declaration
 * @for p5
 */

/**
 * @property let
 *
 * @example
 * <div>
 * <code>
 * let x = 2;
 * console.log(x); // prints 2 to the console
 * x = 1;
 * console.log(x); // prints 1 to the console
 * </code>
 * </div>
 */

/**
 *
 * Declares a read-only named constant.
 * Constants are block-scoped, much like variables defined using the 'let' statement.
 * The value of a constant can't be changed through reassignment, and it can't be redeclared.
 * @property const
 *
 * @example
 * <div>
 * <code>
 * //define MY_FAV as a constant and give it the value 7
 * const MY_FAV = 7;
 * console.log('my favorite number is: ' + MY_FAV);
 * </code>
 * </div>
 */

/**
 * @submodule Control Statement
 */

/**
 * The 'if' statement executes a statement if a specified condition is truthy.
 * @property if-else
 *
 * @example
 * <div>
 * <code>
 * let a = 4;
 * if (a > 0) {
 *   console.log('positive');
 * } else {
 *   console.log('negative');
 * }
 * </code>
 * </div>
 */

/**
 * @submodule Function
 */

/**
 * Declares a function with the specified parameters.
 * @property function
 *
 * @example
 * <div>
 * <code>
 * function calRectArea(width, height) {
 *   return width * height;
 * }
 * calRectArea(4, 5); //calling the function
 * </code>
 * </div>
 */

/**
 * Specifies the value to be returned by a function.
 * @property return
 *
 * @example
 * <div>
 * <code>
 * function squareNumber(x) {
 *   return x * x;
 * }
 * squareNumber(4); //return 16
 * </code>
 * </div>
 */

/**
 * @submodule Class
 */

/**
 * The class declaration creates a new class with a given name using prototype-based inheritance.
 * @property class
 *
 * @example
 * <div>
 * <code>
 * class Polygon {
 *   constructor(height, width) {
 *     this.name = 'Polygon';
 *     this.height = height;
 *     this.width = width;
 *   }
 * }
 * var polygon = new Polygon(4, 5); //creating new instance of Polygon Class.
 * console.log(polygon.name);
 * </code>
 * </div>
 */

/**
 * @submodule Iterative Statement
 */

/**
 * Creates a loop that executes a specified statement until the test condition evaluates to false.
 * The condition is evaluated after executing the statement, resulting in the specified statement executing at least once.
 * @property for
 *
 * @example
 * <div>
 * <code>
 * for (var i = 0; i < 9; i++) {
 *   console.log(i);
 *   // more statements
 * }
 * </code>
 * </div>
 */

/**
 * It Creates a loop that executes a specified statement until the test condition evaluates to false.
 * The condition is evaluated after executing the statement, resulting in the specified statement
 * executing at least once.
 * @property  do..while
 *
 * @example
 * <div>
 * <code>
 * var result = '';
 * var i = 0;
 * do {
 *   i += 1;
 *   result += i + ' ';
 * } while (i > 0 && i < 5);
 * console.log(result);
 * <code>
 * </div>
 */

/**
 * Iterates over the enumerable properties of an object, in arbitrary order.
 * For each distinct property, statements can be executed.
 * @property for..in
 *
 * @example
 * <div>
 * <code>
 * var obj = { a: 1, b: 2, c: 3 };
 * for (const prop in obj) {
 *   console.log(`obj.${prop} = ${obj[prop]}`);
 * }
 * // Output:
 * // "obj.a = 1"
 * // "obj.b = 2"
 * // "obj.c = 3"
 * </code>
 * </div>
 */

/**
 * Iterates over iterable objects (including arrays, array-like objects, iterators and generators),
 * invoking a custom iteration hook with statements to be executed for the value of each distinct property.
 * @property for..of
 *
 * @example
 * <div>
 * <code>
 * let iterable = 'boo';
 * for (let value of iterable) {
 *   console.log(value);
 * }
 * // "b"
 * // "o"
 * // "o"
 * </code>
 * </div>
 */

/**
 * @submodule JS Method
 */

/**
 * The parseInt() function parses a string argument and
 * returns an integer of the specified radix (the base in mathematical numeral systems).
 * @method parseInt
 * @param {string} string :The value to parse
 * @param {Integer} radix :An integer between 2 and 36 that represents the radix (the base in mathematical numeral systems) of the string.
 *
 * @example
 * <div>
 * <code>
 * parseInt('F', 16);
 * parseInt(21, 8);
 * </code>
 * </div>
 */

/**
 * The JSON.stringify() method converts a JavaScript object or value to a JSON string.
 * @method JSON.stringify
 * @param {Object} object :Javascript object that you would like to convert to JSON
 *
 * @example
 * <div>
 * <code>
 * JSON.stringify('foo'); // '"foo"'
 * JSON.stringify([1, 'false', false]); //'[1,"false",false]'
 * JSON.stringify([NaN, null, Infinity]); //'[null,null,null]'
 * </code>
 * </div>
 */
```

singhvisha: @stalgiag ,
thank you

singhvisha:
@stalgiag 
hi, It has passed all the test.But still when I am running the Command `npm run docs:dev`.And when I am clicking the few example like `for..in, for..of, JSON.stringify`,Its description page is not opening.

stalgiag:
@singhvisha I am not sure off the top of my head how you can render those periods correctly and still get the desired URL. The url encoding for a period (%2E) doesn't seem to point correctly. 

But it doesn't matter as the period use is arbitrary, why not use `-` as you did with `if-else`?

singhvisha:
hi @stalgiag ,now it seems working except the `do-while`.

stalgiag:
Hi @singhvisha looking good! 

One last thing, can you run the following command then commit and push the changes?
`git checkout upstream/master -- package-lock.json`
This will revert the changes to the package-lock file.

After you have done this, we can begin to consider a merge, which can be followed by further edits and additions by other contributors.

singhvisha:
hi @stalgiag ,when I am running the command `git checkout upstream/master -- package-lock.json` and try to commit.It's saying nothing to commit.

stalgiag:
Okay not sure of a simple way to revert the single file then --

@limzykenneth this pull request is a continuation of #3897, in which you said the following:

> Upon further inspection, the current `package-lock.json` is out of sync with `package.json` and so when running `npm install` it will repopulate `package-lock.json` with new data. I think the version committed here is correct but it's hard to trace down why the previous changes were made (most came from different commits). We may need to have a round of dependency updates along with `npm audit` very soon.
> 
> I'm ok to merge this PR as is (for the `package-lock.json` at least).

Is this true for the `package-lock.json` in this PR as well? I don't know enough about the `package-lock.json` to know how to manage the syncing. Thank you for any info!

limzykenneth:
Most of the problems with `package-lock.json` should have been fixed over the past few weeks by dependabot and quick visual inspection of the file committed here seems to match the existing one in the repo. If you want to be sure nothing was messed up you can diff the two versions locally.

stalgiag:
Great thank you @limzykenneth !

singhvisha:
Hi @stalgiag,Could you please do me favour?
Could you please assigne me a new issue(#3990)?

src/core/reference.js
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * Creates a loop that executes a specified statement until the test condition evaluates to false.
+ * The condition is evaluated after executing the statement, resulting in the specified statement executing at least once.
+ * @property for
+ *
+ * @example
+ * <div>
+ * <code>
+ * for (var i = 0; i < 9; i++) {
```
stalgiag:
```suggestion
 * for (let i = 0; i < 9; i++) {
```

stalgiag:
Hi @singhvisha! So I am going to suggest that we close this PR in order to create a clear track on the changes being made here. This will also help you to get a handle on what a clean workflow will look like in the future.

1. Close this pull request.
2. Make backups of both of the current `reference.js` and `local_storage.js` files. Place these somewhere outside of your p5.js folder. This is important so that you do not lose work!
3. Delete your fork and refork. This is rare to do, but I think it is the easiest way to get the history of your `master branch` in sync with the upstream (`processing/p5.js:master`) branch.
4. Follow the instructions in [development process](https://github.com/processing/p5.js/tree/master/developer_docs#development-process) section of the `developer_docs`. Take note that it is better to use `npm ci` instead of `npm install`. This will avoid changes to `package-lock.json`
5. Create a branch for the js reference examples. (`git branch jsReferenceDocs`)
6. Checkout your new feature branch (`git checkout jsReferenceDocs`)
7. Add your backup of `reference.js` from step 2. back into the project.
8. Run `npm run grunt` and make sure that everything is working as expected.
9. Commit your changes, push them, and then open a new pull request based on your `jsReferenceDocs` branch. You should add links to this PR and #3897 in that new PR so there is a trail to follow previous work.

10. Switch back to your master branch. (`git checkout master`) This step is probably unecessary but it is good to go back to the neutral master branch before starting new work.
11. Create a new branch for the local storage p5.Vector work. (`git branch <desiredBranchName>)
12. Checkout that new branch (`git checkout <desiredBranchName>`)
13. Add your changes from the backup of `local_storage_BACKUP.js` from step 2 back into `local_storage.js`
14. Run `npm run grunt` and make sure that everything works as expected.
15. Commit your changes, push them, and then open a new pull request based on this new separate feature branch.

At the end of this process you should have two pull requests open. The work from either pull request shouldn't cross over. They should each have different changes present.

I know this process has been arduous but by the end you should have a good sense of how to contribute in a really nice and easily reviewable way.

singhvisha:
@stalgiag 
Thank you.

PR #3983, opened 2019-09-01
remove stray copies of builds in lib/empty-example/

outofambit:
don't freak out 😄 we don't use these copies of the builds! (they live in `lib/`.)

noticed this when i saw that #3890 was modifying the wrong file


stalgiag:
I don't fully understand what happens with the `lib/empty-example` folder so this might be way of base but does the fact that [the empty sketch](https://github.com/processing/p5.js/blob/master/lib/empty-example/index.html) points to these files matter? Or are they placed there again during some other part of the distribution or build process?

outofambit:
@stalgiag thanks for bringing this up! i double checked and it looks like the empty sketch points to the build artifacts in the enclosing `lib/` folder (note the `"../"`):

https://github.com/processing/p5.js/blob/0ac2b18d925464581b8cd3c65be599a4e46010f3/lib/empty-example/index.html#L8-L10

i ran a build locally just now and i see the those files placed in `lib/` but not in `lib/empty-example`, which i believe is what we want. my understanding of the `lib/empty-example` folder is that it only exists to be packaged up with the distribution zip for the downloads page. (the release process just zips up everything in `lib/`, so the build artifacts and anything else hanging around in that folder are included.)

stalgiag:
Oh odd, truly no point for the dupes! Good catch! Thank you for the explanation @outofambit

PR #3987, opened 2019-09-03
Add erase() and noErase()

### LINKED ISSUE #3759: Draw transparent shapes
#### Nature of issue?

- [ ] Found a bug
- [x] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### New feature details:

My use case:

I put the canvas created by `p5` over an image so that I can draw on it. It worked fluently but when I wanted to implement a rubber, I found no way to make a certain region transparent, e.g., drawing a transparent rectangle/circle.

I ended up with using `p.drawingContext.clearRect()`. Is it possible that `p5` would expose this method, or provide a direct way to draw transparent shapes?

stalgiag:
closes #3759 

I have an [early example in the editor here](https://editor.p5js.org/stalgiag/sketches/1WJC1MtvS). This opens up some neat opportunities for layering graphics, and exposing hidden HTML elements behind the canvas.

#### Currently:
  1. `erase()` activates a state in which all drawing (see 3. for exceptions) is subtractive from its destination canvas
  2. `noErase()` deactivates  that state
  3. `image()` and `background()` do not become erasers during the `erase()` state. This is consistent with `fill()` and prevents behavior that will rarely be desired (using the background or image as an eraser). If someone still wants that behavior they can use `blendMode(REMOVE)` which is added with this PR.
  4. In order to accomplish 3, the most recent change to the `globalCompositingOperation` is stored as `p5.Renderer2D._cachedBlendMode`
  5. The strength (or opacity) of the erasing is dictating by the current fill and stroke color
  6. **No docs/examples/tests yet**
  7. ~~No WEBGL erasing. This might be difficult due to the clearing style of WEBGL. I can give it a shot though.~~

#### Open Questions:
  1. Do we like this interface? ie `erase()`/`noErase()` as opposed to other options (see discussion in issue at top)
  2. Related to 5 above, should the strength of the erase be based on current fill, or should `erase()` accept a parameter of a number between 0 - 255?
  3. Currently `stroke()` also erases. Should we keep this behavior?

#### My opinions
  1. There may be use cases I am missing, but this combined with the option for `blendMode(REMOVE)` seems to work nicely.
  2. Requiring a parameter makes sense  to me(ala `fill()`) but this contradicts my answer to 3 below.
  3. I like that it is possible to erase with `stroke()`. This may create unexpected effects at first but is relatively intuitive and offers a lot of flexibility. Adding a parameter is complicated by this though.

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -437,6 +437,12 @@ export const UP_ARROW = 38;
  */
 export const BLEND = 'source-over';
 /**
+ * @property {String} REMOVE
+ * @final
+ * @default destination-out
+ */
+export const REMOVE = 'destination-out';
+/**
  * @property {String} ADD
  * @final
  * @default lighter
--- a/src/color/setting.js
+++ b/src/color/setting.js
@@ -762,12 +762,35 @@ p5.prototype.stroke = function(...args) {
  * @alt
  * 60x60 TODO!
  */
-p5.prototype.erase = function() {
+p5.prototype.erase = function(opacityFill, opacityStroke) {
   this._renderer._isErasing = true;
+
+  opacityFill = typeof opacityFill == 'undefined' ? 255 : opacityFill;
+  opacityStroke = typeof opacityStroke == 'undefined' ? 255 : opacityStroke;
+
   if (this._renderer.isP3D) {
     this._renderer._cachedBlendMode = this._renderer.curBlendMode;
     this._renderer.blendMode(constants.REMOVE);
+
+    this._renderer._cachedFillStyle = this._renderer.curFillColor.splice(0, 4);
+    this._renderer.curFillColor = [1, 1, 1, opacityFill / 255];
+
+    this._renderer._cachedStrokeStyle = this._renderer.curStrokeColor.splice(
+      0,
+      4
+    );
+    this._renderer.curStrokeColor = [1, 1, 1, opacityStroke / 255];
   } else {
+    // cache the fill style
+    this._renderer._cachedFillStyle = this.drawingContext.fillStyle;
+    const newFill = this.color(255, opacityFill).toString();
+    this.drawingContext.fillStyle = newFill;
+
+    //cache the stroke style
+    this._renderer._cachedStrokeStyle = this.drawingContext.strokeStyle;
+    const newStroke = this.color(255, opacityFill).toString();
+    this.drawingContext.strokeStyle = newStroke;
+
     const tempBlendMode = this._renderer._cachedBlendMode;
     this._renderer.blendMode(constants.REMOVE);
     this._renderer._cachedBlendMode = tempBlendMode;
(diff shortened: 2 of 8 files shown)


stalgiag:
Working with WEBGL in the same way now. [Here is a nearly identical example](https://editor.p5js.org/stalgiag/sketches/aAKV2jFA5).

Note: that the only difference is that the erase fill is not semi-transparent, which I removed because of some odd default color blending that is unrelated to this PR. I will open another issue after this.

lmccart:
this is exciting! I read through it quickly and all seems on the right track. I will respond in the next couple days with some thoughts on the open questions.

limzykenneth:
This looks amazing! 

1. The current interface seems reasonable to me. The combination of `erase()`, `noErase()` and `blendMode(REMOVE)` will require a bit of explaining so I wonder if `blendMode(REMOVE)` can be merged with `erase()` into something like `erase(BACKGROUND)`? If not the current interface is still ok.

2. This kinda contradicts with my suggestion to 1 but I'm ok with its current behaviour as well. Just a wild idea, it would be pretty wild to be able to subtract color with `erase(r, g, b)` I think, though the semantic may be confusing then.

3. Ok with `stroke()` erasing. See 2 for color subtraction which applies to strokes as well.

lmccart:
1. I like these options. I hear @limzykenneth's point but think it may not be too much of an issue to keep both. I imagine the sets of methods would mostly be used in different cases.
2. I think a parameter makes most sense. I find it difficult to wrap my head around `fill()`, `stroke()` and transparency combined.
3. Seems fine to me. I didn't realize this was even possible.

stalgiag:
Okay this should be ready for review now.

I added parameters for erasing strength, docs, and unit tests. I also cleaned up the code and moved the functions down to the Renderer level.

src/core/constants.js
```diff
@@ -436,6 +436,12 @@ export const UP_ARROW = 38;
  * @default source-over
  */
 export const BLEND = 'source-over';
+/**
+ * @property {String} REMOVE
+ * @final
+ * @default destination-out
+ */
+export const REMOVE = 'destination-out';
```
lmccart: one small thing. if we're adding a blendMode option here, can you also add it to this list in the blendMode() reference: https://github.com/processing/p5.js/blob/0.9.0/src/core/rendering.js#L236

lmccart:
The different strengthFill and strengthStroke were unexpected for me. I would've thought, like an eraser, it would have the same strength across both. But it's also interesting to play with. I'm good to keep this and see what emerges. :) I think this is good to merge after we resolve the one comment I left.

src/core/constants.js
stalgiag: Good point thank you!

PR #3992, opened 2019-09-07
fix small random bug

rogerwim:


limzykenneth:
Hi @rogerwim, can you explain more about what this PR is about? Also can you outline the changes to be made in an issue first before filing a PR? Thanks!

PR #3993, opened 2019-09-07
improvements to reference page

CRYSTALMINYICHEN:
Hi, p5 team. This is Crystal Chen from NYU. I was doing summer research with Tega, and we added some descriptions and examples to reference pages for Camera and 2D Primitives. Would you like to see if the change we made was helpful, please? Thank you.


outofambit:
hi @CRYSTALMINYICHEN thanks for sharing this work! i'm going to take a closer look at this soon, but i love the [sketch you link to](https://www.openprocessing.org/sketch/740258) that shows camera vs object. i think it might be a good addition to the tutorial section on the website! @stalgiag @limzykenneth what do you think?

also a side note: it looks like the diff for `3d_primitives.js` duplicates instead of replaces those functions, so we'll need to fix that up before we merge this. i can fix that up if you'd like!

CRYSTALMINYICHEN:
Hi, thanks for your reply!
I would love to add the example sketch to tutorial section on the website, I believe it will help users to visualize the position of a camera and thus have a better understanding of cameras and view in WEBGL mode. And it would be perfect if you can fix the issue in 3d_primitives.js before you merge that. 

Thank you,
Crystal Chen

lmccart:
this looks great, thanks @CRYSTALMINYICHEN and @outofambit!

Issue #3733, opened 2019-05-08
ARC reference 


BerfinA:
#### Nature of issue?

- [x ] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ x] Other (specify if possible)- reference for arc 

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [ x] Desktop/Laptop- web editor 
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 
- Web browser and version: 
- Operating System: 
- Steps to reproduce this:

#### New feature details:

in the description of arc reference it specifies; 

x | Number: x-coordinate of the arc's ellipse
y | Number: y-coordinate of the arc's ellipse
w | Number: width of the arc's ellipse by default
h | Number: height of the arc's ellipse by default
start | Number: angle to start the arc, specified in radians
stop | Number: angle to stop the arc, specified in radians
mode | Constant: optional parameter to determine the way of drawing the arc. either CHORD, PIE or OPEN
detail | Number: optional parameter for WebGL mode only. This is to specify the number of vertices that makes up the perimeter of the arc. Default value is 25.

The arc is always drawn clockwise from wherever start falls to wherever stop falls on the ellipse. Adding or subtracting TWO_PI to either angle does not change where they fall. If both start and stop fall at the same place, a full ellipse will be drawn. 

However, when I put this code to the editor my arc is drawn counter clock wise. This is the case for both using radians and degrees.
```
function setup() {
  createCanvas(400, 400);
}

function draw() {
   background(220);
  arc(100,100,100,100,PI/2,3*PI/2);
}
```

limzykenneth:
I believe counter clockwise is the correct behaviour and the reference probably left it out by mistake.

outofambit:
thanks for clarifying @limzykenneth!

joecridge:
The intended behaviour is clockwise (same as Processing), so the documentation is correct. The snippet above should draw the left half of a circle (since zero degrees is the line starting from the centre of the circle and heading off to the right [in the positive x-direction]).

[...]

Issue #3373, opened 2018-12-11
Update Structure Reference Examples on p5js.org to ES6


mlarghydracept:
Update the **Structure** section of the [p5js reference](https://p5js.org/reference/) to ES6. This will entail updating all of the references in source files that have `@module Structure` at the top.

As a step forward in the transition to ES6 all of the reference examples on p5js.org should be updated to use ES6 standard. This will often mean simply switching 'var' to 'let' or in some cases [using es6 classes](https://github.com/processing/p5.js/issues/2215) where applicable. 

There are tons of resources online for learning about ES6. [Here is an interactive guide](http://stack.formidable.com/es6-interactive-guide/#/) that covers the biggest features.

sanketsingh24:
Would like to work on this issue today :)

PR #4015, opened 2019-09-14
Clarified in the documentation that angles also grow clockwise for arc().

maxthomax:
Addresses #3373, specifically comment https://github.com/processing/p5.js/issues/3733#issuecomment-531463288


lmccart:
thanks @maxthomax!

PR #4016, opened 2019-09-15
Use reject in mocha testing promise

### LINKED ISSUE #4000: Grunt occasionally inserts endless newlines during error
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Development/Testing

- [x] Desktop/Laptop

#### Details about the bug: 

When running `npm run grunt` if a test fails during the documentation testing section (I think only during this and not the unit tests) that causes grunt to spam the terminal with infinite newlines. On my computer this leads to crashing if I don't `CRTL+C` in time, but even if I do this it is then very hard to track down the message for the error since it is often 10k+ lines away by time I catch it.

I have been experiencing this for a couple of weeks, I didn't report it because I thought it was specific to my environment, but now I see that it is happening on Travis as well. [Here is an **unstable** link to an example](https://travis-ci.org/processing/p5.js/jobs/582763763).

- Operating System: Windows 10, Ubuntu 16.0.4(Travis)
- Steps to reproduce this:

Run `npm run grunt` with code that will cause an error during the documentation testing section.

stalgiag:
closes #4000 

The issue above was throwing some wrenches in #3978, so I got a little curious and did a `git bisect`.

Based on that, I think it was something in [this commit](https://github.com/stalgiag/p5.js/commit/894489b0f22b6bbfb8ec48131733548ac6c0ecef). I guessed that it was the `Throw` replacement of `reject`. I think this promise is in a nested async callback, with the parent being `grunt.registerMultiTask( ... async function()`  so `Throw` doesn't [work the same as `reject`](https://stackoverflow.com/questions/33445415/javascript-promises-reject-vs-throw). 

@outofambit does this seem right? Not my most familiar area but it was fun to look around a bit.

### DIFF
--- a/tasks/test/mocha-chrome.js
+++ b/tasks/test/mocha-chrome.js
@@ -59,11 +59,11 @@ module.exports = function(grunt) {
             eventHandler1.emit.bind(eventHandler1)
           );
 
-          await new Promise(async resolve => {
+          await new Promise(async (resolve, reject) => {
             eventHandler1.on('mocha:end', async results => {
               const { stats, coverage } = results;
               if (stats.failures) {
-                throw new Error(stats);
+                reject(stats);
               }
               await saveCoverage(coverage);
               resolve(stats);


limzykenneth:
Not an absolute requirement but promises usually reject with an `instanceof Error` just to make downstream error handling more straightforward.

PR #4022, opened 2019-09-17
add @lisamabley to contributors

outofambit:
hi @LisaMabley, thought i would take a stab at this! does this look ok for you?


LisaMabley:
Yes, it looks fine. Thanks!

PR #4030, opened 2019-09-19
Use random(arr) in loadStrings() inline example

nbogie:
Switched the examples loadStrings() from the inline docs
to use random(arr) rather than ix = floor(random(arr.length)) etc

This makes it simpler to read and the intention clearer to grasp.


lmccart:
super!

Issue #4038, opened 2019-09-25
the round function


mujehoxe:
#### Nature of issue?

- [ ] Found a bug
- [ x] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [x ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- p5.js version: 
- Web browser and version: 
- Operating System: 
- Steps to reproduce this:

#### Feature enhancement details:
the round function can take only one pramater n and then round it .eg : round(2.156) => 2.
i would suggest adding an optional paramater that represents the number of decimal places.eg : round(2.156, 2)  =>  2.2 
i know that there is a simple work around by multiplying the number by 10^2 and round it then dividing the result by 10^2   but just for making the function more general.

CrypticGuy:
@outofambit 
Hi, Can I work on this. 
It would be a nice enhancement to the function. We can add a new parameter to the existing function and just have a default argument, which rounds off to the integer value.

lmccart:
@CrypticGuy sounds good, we would welcome your work on this one

lmccart:
resolved with https://github.com/processing/p5.js/pull/4043

PR #4042, opened 2019-09-28
Added a second Optional Argument To Round Function #4038

ashu8912:
round function can now take two argument first the number to be rounded and second argument is the number of decimal places
this second argument is made optional.


ashu8912:
#4043

PR #4044, opened 2019-09-29
Typo in documentation (acces -> access)

ocwalk:


lmccart:
thanks @ocwalk! if you can make the change over in the [p5.js-sound repository](https://github.com/processing/p5.js-sound/blob/master/src/audioin.js#L70) it will ensure this change sticks. this file you edited is actually copied from that repo's build.

Issue #3973, opened 2019-08-27
bug in angleBetween()


agnius-vasiliauskas:
angleBetween() returns angle between `[0, PI]`, but should return `[-PI,PI]`.
Because now a **pair** of directions can have same angle, however they are different:
Attaching explanation / picture how should be and why.
![angles](https://user-images.githubusercontent.com/4015780/63761275-86610e00-c8c9-11e9-9b0f-e1fae9ccad06.png)

lmccart:
@0x69 can you please provide a code snippet that illustrates the issue? thanks

0x69:
Of course:

```
function setup() {
 let v0 = createVector(1, 0);
 let v1 = createVector(0, 1);
 let v2 = createVector(0, -1);  
 let a1 = v0.angleBetween(v1);
 let a2 = v0.angleBetween(v2);
 print(degrees(a1) + '° between v0,v1');
 print(degrees(a2) + '° between v0,v2');
}
```

>  BAD output:
>  90° between v0,v1 
>  90° between v0,v2
>  EXPECTED output :
>  a)
>  -90° between v0,v1 
>  90° between v0,v2
>  OR
>  b)
>  90° between v0,v1 
>  -90° between v0,v2
>  Depends on how API will choose negative angle tactics - when going
>  clockwise or anti-clockwise from v0 vector to target vector
>

nishantkr18:
@0x69 it seems that `angleBetween()` always returns the absolute angle, although one can know the difference pointed by you, using `let a1 = v0.angleBetween(v2)*Math.sign(v0.cross(v2).z);` when dealing with only x and y (two dimensions).

CrypticGuy:
Hi @stalgiag 
I would like to work on this.

I think it can be solved by either providing the method mentioned by @nishantkr18 in the reference/docs, or we could add a parametrized option to the function, where the angle is displayed clockwise or counter-clockwise, depending on the argument passed. 
If none of the above methods then a new function could be defined to get the counter or counter-clockwise angle.
If these seem like suitable methods, I would like to work on this issue.

0x69:
> @0x69 it seems that `angleBetween()` always returns the absolute angle, although one can know the difference pointed by you, using `let a1 = v0.angleBetween(v2)*Math.sign(v0.cross(v2).z);` when dealing with only x and y (two dimensions).

[...]

PR #4048, opened 2019-09-30
bug in angleBetween() #3973

CrypticGuy:
Hi, 
I've added the changes as discussed in the issue section. I have added a boolean parameter to the function `angleBetween()` for vectors. When set to true absolute value of the angle between them is returned and when set to false the angle between is returned keeping in consideration the sign in a 2D plane.


0x69:
Nice. Should I close issue #3973 now or should I wait merging branch into main or something ?

stalgiag:
Thanks @CrypticGuy !  I realized that the next release is a minor release so it is okay to break backwards compatibility on a function like this. Especially if the fix is simple as with this case (using `abs()`. With that in mind, I think the added parameter might not be necessary. Would you mind removing it and just keeping the logic for evaluating the sign of the return value?

CrypticGuy:
> Thanks @CrypticGuy ! I realized that the next release is a minor release so it is okay to break backwards compatibility on a function like this. Especially if the fix is simple as with this case (using `abs()`. With that in mind, I think the added parameter might not be necessary. Would you mind removing it and just keeping the logic for evaluating the sign of the return value?

Hi @stalgiag 
So, just to confirm, you want me to remove the extra boolean parameter and return the values with the sign. So, if anyone wants to use the function and need the absolute value they can wrap it in `abs()` on their own?

stalgiag:
Hi @CrypticGuy yes that is correct! Thank you!

CrypticGuy:
Hi @stalgiag , 
Sorry for the late response. If I remove the functionality won't it affect the existing test? Should I override them with new expected values or should I modify them with a wrapper of abs() around them?

CrypticGuy:
Hi @stalgiag , 
I have modified the test cases as the build was failing otherwise. 
Currently I have wrapped `angleBetween()` `call in Math.abs()`

stalgiag:
Hi @CrypticGuy this is good for some of the unit tests, but maybe just to make sure that the full -PI to PI range is working, you could replace one or two of the tests with a test that has a value that results in a value between 0 and -PI. Does that make sense? This helps us to test the full new functionality.

CrypticGuy:
Hi @stalgiag , 
I have modified and added a couple of unit test to check the value between the new range. 
I have added test explicitly for value of 0 and PI.

stalgiag:
@all-contributors can you add @CrypticGuy for code and tests?

Issue #4050, opened 2019-10-01
hypot and dist improvements and weird performance results


J-888:
#### Nature of issue?

- [ ] Found a bug
- [X] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [X] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [X] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

- Web browser and version: Google Chrome 77.0.3865.90
- JavaScript version: V8 7.7.299.11
- Operating System: Windows 10 OS Version 1903 (Build 18362.356)

#### Feature enhancement details:
I was working on making the dist function faster in 2 ways. First, many graphic engines have a square dist function to make z sorting (or any distance sorting) way faster, so Im currently working on it to p5.js. The second one would be improving the current hypot function. It checks if the native implementation is available, and if so, uses it (https://github.com/processing/p5.js/blob/master/src/math/calculation.js#L794), and if not, uses the V8 implementation (https://github.com/v8/v8/blob/8cd3cf297287e581a49e487067f5cbd991b27123/src/js/math.js#L217), which I'm currently working on improving by implementing fast paths for 2 and 3 args (the most common in p5.js).

When testing the performance of a few functions (on https://jsperf.com/distance-comparison), I realized that for some reason Math.hypot works way faster than p5.js hypot, when (in this case) it''s just calling Math.hypot. Any idea why?

```js
if (typeof Math.hypot === 'function') {
  return Math.hypot.apply(null, arguments);
}
```

I'm also wondering how could (according to my tests) p5.js hypot run faster when Math.hypot is not available. This doesn't make much sense, especially with the previous weird result.

![jsperf results](https://i.imgur.com/VzjX7p3.png)

Any idea what I could be wrong? jsperf isn't flawless, but this doesn't make any sense to me. I plan on using the repo built in performance tester to double check, but I have never used it do it might take a while.

Thanks in advance!

J-888:
Results after timing all functions on Chrome console, it makes way more sense now:

| Method | Function | Time |
| ------------- | ------------- | -------------:|
[...]

PR #4053, opened 2019-10-02
distSq function, documentation, example and unit tests (100% coverage)

J-888:
distSq is a quick way to compare different distances when the exact distance isn't needed. It performs way better than the regular dist, as noted in https://github.com/processing/p5.js/wiki/Optimizing-p5.js-Code-for-Performance#math-tips

Improvement mentioned in #4050

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -182,6 +182,96 @@ p5.prototype.dist = (...args) => {
 };
 
 /**
+ * Calculates the square of the distance between two points, in either two or three dimensions. Faster if the real distance is not required in the case of comparing distances, etc.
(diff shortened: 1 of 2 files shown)


limzykenneth:
Hi @J-888, the possibility of implementing `distSq` has already been discussed in issue #3337 and the decision there is not to include this. If you feel that there is an argument for its inclusion, please open a new issue.

In general, it is prefered that any PR (especially ones that add new features or changes any existing behaviour) to be discussed in an issue before it is filed, just so that there won't be any wasted work in the case that we decide not to accept said proposal. I'm going to close this now.

PR #4057, opened 2019-10-04
Expand on foundational JS docs

### LINKED ISSUE #3897: p5js Reference should include entries for JavaScript basics

stalgiag:
This PR builds on the [great work ](https://github.com/processing/p5.js/pull/4033)by @singhvisha .

I wanted to try my hand at adding some additional beginner-friendly language to describe some of the key JS basics and supplement the linked MDN pages. This requires trying to figure out how much information to give and how much to withhold. 
As expected, this is a really difficult needle to thread. 

So any and all help with making the descriptions more clear and concise is welcome.
In addition, any higher-level edit suggestions are welcome. This can be anything from formatting to a suggestion to omit or include a specific reference.

Also, I have a couple more specific areas that I welcome opinions on:
- My instinct is to omit the `for-in` and `for-of` references. I am finding it difficult to explain the distinction between them in beginner-friendly language. I also feel that these two are a little on-the-edge of my imagined boundaries for these docs.
- Also, feel that it might be good to omit `parseInt()`. We have[ `int()` ](https://p5js.org/reference/#/p5/int)in the library.
- I don't use any of the p5 functions in the examples. This makes sense to me since we are showing code that should work outside of `setup()` etc. But it means a ton of instances of `console.log`. (which might be better anyways [since `print()` is contested](https://github.com/processing/p5.js/issues/2765))
- I included the `types` even though this is a murky area with JS. I think it is useful for beginners to feel comfortable with these terms. It also helps with cross-referencing when describing what other things are for. (ie classes referencing objects)

I think it is important to nail down the scope of the references and language used within them. So this means I will probably keep passing over these for a bit till they are solid.

Any and all input is welcome, especially from people that have experience teaching these concepts.

Gonna **no-pressure** tag a few people 
@shubymao @singhvisha @lmccart @outofambit @brysonian

closes #3897

### DIFF
--- a/src/core/reference.js
+++ b/src/core/reference.js
@@ -71,7 +71,7 @@
  * @example
  * <div class='norender'>
  * <code>
- * const a = 4;
+ * let a = 4;
  * if (a > 0) {
  *   console.log('positive');
  * } else {
@@ -139,7 +139,7 @@
  * @example
  * <div class='norender'>
  * <code>
- * const myBoolean = false;
+ * let myBoolean = false;
  * console.log(typeof myBoolean); // prints 'boolean' to the console
  * </code>
  * </div>
@@ -157,7 +157,7 @@
  * @example
  * <div class='norender'>
  * <code>
- * const mood = 'chill';
+ * let mood = 'chill';
  * console.log(typeof mood); // prints 'string' to the console
  * </code>
  * </div>
@@ -174,7 +174,7 @@
  * @example
  * <div class='norender'>
  * <code>
- * const num = 46.5;
+ * let num = 46.5;
  * console.log(typeof num); // prints 'number' to the console
  * </code>
  * </div>
@@ -190,7 +190,7 @@
  * @example
  * <div class='norender'>
  * <code>
- * const author = {
+ * let author = {
  *   name: 'Ursula K Le Guin',
  *   books: [
  *     'The Left Hand of Darkness',
@@ -224,7 +224,7 @@
  *     this.width = width;
  *   }
  * }
- * const square = new Polygon('square', 1, 1); // creating new instance of Polygon Class.
+ * let square = new Polygon('square', 1, 1); // creating new instance of Polygon Class.
  * console.log(square.width); // prints '1' to the console
  * </code>
  * </div>
@@ -271,7 +271,7 @@
  * @example
  * <div class='norender'>
  * <code>
- * const person = { fname: 'John', lname: 'Doe', age: 25 };
+ * let person = { fname: 'John', lname: 'Doe', age: 25 };
  * let text = '';
  * let x;
  * for (x in person) {
@@ -291,8 +291,8 @@
  * @example
  * <div class='norender'>
  * <code>
- * const word = 'boo';
- * for (const value of word) {
+ * let word = 'boo';
+ * for (let value of word) {
  *   console.log(value);
  * }
  * </code>
@@ -328,8 +328,8 @@
  * @example
  * <div class='norender'>
  * <code>
- * const myObject = { x: 5, y: 6 };
- * const myObjectAsString = JSON.stringify(myObject);
+ * let myObject = { x: 5, y: 6 };
+ * let myObjectAsString = JSON.stringify(myObject);
  * console.log(myObjectAsString); // prints "{"x":5,"y":6}" to the console
  * console.log(typeof myObjectAsString); // prints 'string' to the console
  * </code>
--- a/src/core/reference.js
+++ b/src/core/reference.js
@@ -57,43 +57,18 @@
  */
 
 /**
- * The equality operator <a href="#/p5/==">==</a>
- * checks to see if two values are equal.
- *
- * Unlike, the strict equality operator, <a href="#/p5/===">===</a>, if
- * the values being compared are not already the same type they will be
- * converted to the same type before comparison.
- *
- * A comparison expression always evaluates to a <a href="#/p5/boolean">boolean</a>.
- *
- * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators">the MDN entry</a>:
- * The equality operator converts the operands if they are not of the same type, then applies strict comparison.
- * If both operands are objects, then JavaScript compares internal references which are equal when operands refer
- * to the same object in memory.
- *
- * @property ==
- *
- * @example
- * <div class='norender'>
- * <code>
- * console.log(1 == 1); // prints true to the console
- * console.log('1' == 1); // prints true to the console
- * </code>
- * </div>
- */
-
-/**
  * The strict equality operator <a href="#/p5/===">===</a>
  * checks to see if two values are equal and of the same type.
- *
- * Unlike, the equality operator, <a href="#/p5/==">==</a>,
- * regardless of whether they are different types
- * the values being compared are not converted before comparison.
  *
  * A comparison expression always evaluates to a <a href="#/p5/boolean">boolean</a>.
  *
  * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators">the MDN entry</a>:
  * The non-identity operator returns true if the operands are not equal and/or not of the same type.
+ *
+ * Note: In some examples around the web you may see a double-equals-sign
+ * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators#Equality">==</a>,
+ * used for comparison instead. This is the non-strict equality operator in Javascript.
+ * This will convert the two values being compared to the same type before comparing them.
  *
  * @property ===
  *
--- a/src/core/reference.js
+++ b/src/core/reference.js
@@ -53,6 +53,136 @@
  */
 
 /**
+ * @submodule Comparison Operators
+ */
+
+/**
+ * The equality operator <a href="#/p5/==">==</a>
+ * checks to see if two values are equal.
+ *
+ * Unlike, the strict equality operator, <a href="#/p5/===">===</a>, if
+ * the values being compared are not already the same type they will be
+ * converted to the same type before comparison.
+ *
+ * A comparison expression always evaluates to a <a href="#/p5/boolean">boolean</a>.
+ *
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators">the MDN entry</a>:
+ * The equality operator converts the operands if they are not of the same type, then applies strict comparison.
+ * If both operands are objects, then JavaScript compares internal references which are equal when operands refer
+ * to the same object in memory.
+ *
+ * @property ==
+ *
+ * @example
+ * <div class='norender'>
+ * <code>
+ * console.log(1 == 1) // prints true to the console
+ * console.log(1 == '1'); // prints true to the console
+ * </code>
+ * </div>
+ */
+
+/**
+ * The strict equality operator <a href="#/p5/===">===</a>
+ * checks to see if two values are equal and of the same type.
+ *
+ * Unlike, the equality operator, <a href="#/p5/==">==</a>,
+ * regardless of whether they are different types
+ * the values being compared are not converted before comparison.
+ *
+ * A comparison expression always evaluates to a <a href="#/p5/boolean">boolean</a>.
+ *
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators">the MDN entry</a>:
+ * The non-identity operator returns true if the operands are not equal and/or not of the same type.
+ *
+ * @property ===
+ *
+ * @example
+ * <div class='norender'>
+ * <code>
+ * console.log(1 == 1) // prints true to the console
+ * console.log(1 == '1'); // prints false to the console
+ * </code>
+ * </div>
+ */
+
+/**
+ * The greater than operator <a href="#/p5/>">></a>
+ * evaluates to true if the left value is greater than
+ * the right value.
+ *
+ *
+ * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators">There is more info on comparison operators on MDN.</a>
+ *
+ * @property >
+ *
+ * @example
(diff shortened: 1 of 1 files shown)


lmccart:
this is great @stalgiag and thanks @singhvisha for the work on this too!
I agree on all the points you mention in your comment, I will give the code a closer look over in the next couple days.

src/core/reference.js
```diff
@@ -5,7 +5,15 @@
...
+ * Creates and names a new variable. A variable is a container for a value.
+ * Variables that are declared with <a href="#/p5/let">let</a> will have block-scope.
+ * This means that the variable only exists within the <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block">
+ * block</a> that it is created within.
+ *
+ *
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let">the MDN entry</a>:
  * Declares a block scope local variable, optionally initializing it to a value.
+ *
  * @property let
```
lmccart: not totally sure `@property` is the right tag, but I couldn't easily see [which yuidoc tag](https://yui.github.io/yuidoc/syntax/index.html) would be more fitting. so if this works best with our reference building, let's use it!

src/core/reference.js
```diff
@@ -175,7 +304,8 @@
  */
 
 /**
- * The parseInt() function parses a string argument and
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt">the MDN entry</a>:
+ * The <a href="#/p5/parseInt">parseInt()</a> function parses a string argument and
  * returns an integer of the specified radix (the base in mathematical numeral systems).
  * @method parseInt
```
lmccart: if we include `parseInt` we should include `parseFloat`, but I'm in favor of omitting as you mentioned already.

src/core/reference.js
```diff
@@ -120,6 +235,19 @@
...
+ * false then the loop will exit.
+ * The third expression is executed at the end of each loop.
+ *
+ * The code inside of the loop body (in between the curly braces) is executed between the evaluation of the second
+ * and third expression.
+ *
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for">the MDN entry</a>:
  * Creates a loop that executes a specified statement until the test condition evaluates to false.
  * The condition is evaluated after executing the statement, resulting in the specified statement executing at least once.
  * @property for
```
lmccart: i think we should include the while loop too

src/core/reference.js
stalgiag: Yeah I am unsure either. I think none of them are perfect but I will experiment with other tags in case something feels a little closer to correct.

src/core/reference.js
```diff
@@ -20,6 +28,14 @@
  */
 
 /**
+ * Creates and names a new constant variable. A variable is a container for a value.
+ * Variables that are declared with <a href="#/p5/const">const</a> will have block-scope.
+ * This means that the variable only exists within the <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block">
+ * block</a> that it is created within.
+ * <a href="#/p5/let">const</a> is different from <a href="#/p5/let">let</a> in that
+ * variables declared with <a href="#/p5/let">const</a> cannot be reassigned or redeclared.
```
limzykenneth: Are these links here all supposed to link to `let`?

src/core/reference.js
```diff
@@ -40,14 +56,22 @@
  */
 
 /**
+ * The if-else statement helps control the flow of your code.
+ * A condition is placed between the parenthesis following 'if'.
+ * When that condition evalues to <a href="https://developer.mozilla.org/en-US/docs/Glossary/truthy">truthy</a>,
+ * the code between the preceding curly braces is run.
+ * Alternatively, when the condition evaluates to <a href="https://developer.mozilla.org/en-US/docs/Glossary/Falsy">falsy</a>,
+ * the code between the curly braces that precedes 'else' is run instead.
```
limzykenneth: > the curly braces that precedes 'else'

probably should have been "the curly braces that comes after 'else'"? As precede means comes before.

src/core/reference.js
```diff
@@ -40,14 +56,22 @@
...
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else">the MDN entry</a>:
  * The 'if' statement executes a statement if a specified condition is truthy.
  * If the condition is falsy, another statement can be executed
  * @property if-else
  *
  * @example
  * <div class='norender'>
  * <code>
- * let a = 4;
+ * const a = 4;
```
limzykenneth: I think we are sticking with using `let` only in examples? #3877

src/core/reference.js
```diff
@@ -129,14 +257,14 @@
...
- *   // more statements
  * }
  * </code>
  * </div>
  */
 
 /**
- * Iterates over the enumerable properties of an object, in arbitrary order.
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in">the MDN entry</a>:
+ * Iterates over the enumerable properties of an <a href="#/p5/object">object</a>, in arbitrary order.
```
limzykenneth: I agree that `for...in` and `for...of` here may be better ommited as the description here taken from MDN are pretty opaque, unless we write our own descriptions of them.

src/core/reference.js
stalgiag: Haha thanks!

src/core/reference.js
stalgiag: Also, "Haha Thanks!" 🤦‍♀

src/core/reference.js
stalgiag: Oh right! Something about there being an actual reference for `const` had me thinking in that way. Will change!

src/core/reference.js
```diff
@@ -20,6 +28,14 @@
  */
 
 /**
+ * Creates and names a new constant variable. A variable is a container for a value.
```
brysonian: The phrase "constant variable" is potentially confusing. I stumble with this too, but maybe something like "Creates and names a new constant. Like a variable, a constant is a container for a value, however constants cannot be changed once they are declared.

src/core/reference.js
lmccart: after thinking about it some more, let's stick with the tags you used unless there's another that makes considerably more sense. I have a feeling using `property` will just play a little nicer with the yuidocs / reference build system we've built

src/core/reference.js
```diff
@@ -20,6 +28,14 @@
...
+ * Variables that are declared with <a href="#/p5/const">const</a> will have block-scope.
+ * This means that the variable only exists within the <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block">
+ * block</a> that it is created within.
+ * <a href="#/p5/let">const</a> is different from <a href="#/p5/let">let</a> in that
+ * variables declared with <a href="#/p5/let">const</a> cannot be reassigned or redeclared.
+ *
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const">the MDN entry</a>:
  * Declares a read-only named constant.
  * Constants are block-scoped, much like variables defined using the 'let' statement.
  * The value of a constant can't be changed through reassignment, and it can't be redeclared.
```
brysonian: is it worth adding "in the same block" or "in the same scope"?

src/core/reference.js
brysonian: or "the curly braces that follow"

src/core/reference.js
```diff
@@ -62,55 +86,146 @@
...
+ *
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function">the MDN entry</a>:
  * Declares a function with the specified parameters.
  * @property function
  *
  * @example
  * <div class='norender'>
  * <code>
- * function calRectArea(width, height) {
+ * function calculateRectArea(width, height) {
```
brysonian: Maybe an example that doesn't use `return` so those concepts aren't muddled?

src/core/reference.js
```diff
@@ -62,55 +86,146 @@
...
+ *
+ * From <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type">the MDN entry</a>:
+ * Boolean represents a logical entity and can have two values: true, and false.
+ *
+ * @property boolean
+ *
+ * @example
+ * <div class='norender'>
+ * <code>
+ * const myBoolean = false;
```
brysonian: For all of the types examples, maybe use `let` per @limzykenneth's comment above.

src/core/reference.js
```diff
@@ -163,7 +292,7 @@
  * <div class='norender'>
  * <code>
  * const word = 'boo';
- * for (let value of word) {
+ * for (const value of word) {
```
brysonian: Just a warning, `for...of` is not supported in Edge.

src/core/reference.js
```diff
@@ -189,17 +319,19 @@
...
  * @for JSON
  * @param {Object} object :Javascript object that you would like to convert to JSON
  *
  * @example
  * <div class='norender'>
  * <code>
- * JSON.stringify('foo'); // '"foo"'
- * JSON.stringify([1, 'false', false]); //'[1,"false",false]'
- * JSON.stringify([NaN, null, Infinity]); //'[null,null,null]'
+ * const myObject = { x: 5, y: 6 };
```
brysonian: Again, maybe use `let` per @limzykenneth's comment above

brysonian:
@singhvisha and @stalgiag this is great!

src/core/reference.js
brysonian: I think also since for...in is almost never what one wants, maybe we should avoid it. Maybe an example of using `Object.keys();` and a `for` loop somewhere would be helpful though?

src/core/reference.js
stalgiag: Sounds good!

stalgiag:
Thank you @limzykenneth , @lmccart , and @brysonian ! 

I think I covered all of the requests. I still need to do wording and clarity passes but this is starting to feel good.

src/core/reference.js
```diff
@@ -40,6 +57,14 @@
  */
 
 /**
+ * The <a href="#/p5/if-else">if-else</a> statement helps control the flow of your code.
```
brysonian: Should the comparison operators be added? Since they are key for if and for? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators similarly, should examples in general use double or triple equals?

stalgiag: Added comparison operator examples. The linter requires the triple equals in all examples see comment below for more.

stalgiag:
I added the comparison operators as per @brysonian 's suggestion.

The tests above will fail because the linter doesn't like the example with the double `==`. I don't think it is possible to disable a linting rule within a sample with our current sample linting system. If you do `/* eslint-disable eqeqeq */ at the top of the file, this will not apply to the sample as the rule-disabling-line isn't extracted by the sample-linter and if it were it would display in the example.

So the few remaining questions:
1. Anybody have a lead on how to disable a linting rule for a single example? My instinct is that it will require modifying [sample-linter.js](https://github.com/processing/p5.js/blob/master/utils/sample-linter.js).
2. Should I also make a reference for `++`, `--`, `&&`, `||`? These are pretty central, but it is also starting to feel a bit crowded.
3. Both `JSON.stringify()` and `console.log()` display incorrectly in the syntax section. See pic:
Screenshot (143)
This is inline with how all of the other class member methods in the library are displaying, but these two are different in that they are typically going to be used with the respective global window property. Because of this it is a bit confusing to see the syntax as `log()`.

brysonian:
does == need to be in there? If the only place it appears in the samples is here then maybe it isn't necessary. There are so few instances where one would use it anyway.

limzykenneth:
I think it's important to include `==` not because it is useful or recommended (though I would say useful if you don't want to care about types) but because there are many examples past, present, and probably future, that are still using it so it is useful to keep it around at least as a contrast to the behaviour of `===` and potential confusion that may arise for people coming from other programming languages.

stalgiag:
Yeah similar to what @limzykenneth is saying, I thought it useful to have `==` just for contrast against `===`. I think the distinction between these two can be a common snag for beginners and people transitioning from strongly-typed languages.

That said, I wonder if perhaps `==` could just be a footnote in the description for `===`. Do you think that would suffice @limzykenneth ?

limzykenneth:
Yeah, I think just some mention of `==` in `===` would be sufficient.

stalgiag:
Okay, `==` hangup resolved.

Quoting myself just to bump the still relevant questions down the discussion chain.
> 2. Should I also make a reference for `++`, `--`, `&&`, `||`, `!==`? These are pretty central, but it is also starting to feel a bit crowded.
> 3. Both `JSON.stringify()` and `console.log()` display incorrectly in the syntax section. See pic:
>    Screenshot (143)
>    This is inline with how all of the other class member methods in the library are displaying, but these two are different in that they are typically going to be used with the respective global window property. Because of this it is a bit confusing to see the syntax as `log()`.

lmccart:
@stalgiag as a cheap hack, if you write it as `window.console.log()` does it display as desired?

stalgiag:
@lmccart this [relates to a problem ](https://github.com/processing/p5.js/pull/3978#issuecomment-531554491)we encountered with the first pass on the docs. Basically a dot can't go in any of the fields.

But in good news, I just figured out that I can add the `@static` property as a hack and we get the desired text in the syntax field. Will push changes in a minute!

stalgiag:
Okay, this feels ready for final review and then merging. 

Those other references (`++`, `--`, `&&`, `||`, `!==`) can potentially become another issue. I don't want to make this PR too large and it seems unwieldy to capture all of the references that might be helpful in one go.

lmccart:
🎉 thanks @stalgiag and everyone else, this is fantastic!

Issue #4047, opened 2019-09-30
strokeJoin() should log warning in WEBGL mode


stalgiag:
Currently when using `strokeJoin()` in WEBGL mode an error occurs and a cryptic message is logged:
`Uncaught TypeError: this._renderer.strokeJoin is not a function`

`RendererGL.prototype.strokeJoin()` should be added and it should log a message that says something like 'Sorry, strokeJoin() is not currently implemented in WEBGL mode.'

Note: Eventually both `strokeJoin()` and `strokeCap()` could be added for WEBGL mode, but this is not a priority.

hydrosquall:
Hi, I'd like to take a pass at this :).

stalgiag:
closed via #4059 🎊

PR #4059, opened 2019-10-05
[feature] Mark strokeJoin as not implemented in WebGL Renderer 

hydrosquall:
## Notes

- Addresses issue #4047
- Used `console.error` rather than `console.warn` for consistency with how we handled this for the `strokeCap` function immediately above this one.


stalgiag:
Great job @hydrosquall ! Thanks for the fix!

PR #4065, opened 2019-10-08
Added a bunch of functions in push() documentation

imrishabh18:
@limzykenneth Hey,check this out.


limzykenneth:
@imrishabh18 Your branch isn't up to date with master (although that isn't much of a problem), but the larger problem is that the latest commit did not have the changes you have made. Have a look at the files changed [here](https://github.com/imrishabh18/p5.js/commit/f332b539f96fb566c43b960a9e62cb4c2c7fcfcb). It can be either the changes are not commited locally or those commits are not pushed to github.

Make sure the changes are there in the files then check `git status`, it should not list any files as modified.

Issue #3487, opened 2019-01-30
Feature request: fract()


ScottGrogin:
#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### New feature details:
I think that having a fract() function to return the fractional part of a number might be a nice addition to the math functions. I was reading through the book of shaders and it was showing how fract() is used in generating noise and I thought it would be a cool function to have as a part p5js. 

I get that if someone wanted this function they could easily write it themselves as its just returning what is passed as a paramater minus the floor of that value, but I feel like it fits in with the rest of the math functions.

Anyway, let me know what you think about adding fract().

Thanks!

ScottGrogin:
> It's not really needed as (n%1).toFixed(4) works well.
> 
> https://stackoverflow.com/a/4512317/45966

I recognize that there are easy ways for people to implement fract() on their own, but there other functions in p5js that are just as easy to implement. You could argue that a function sq() isn't really needed because it just returns the value passed through multiplied by itself.

vedhant:
If this feature is to be added, I would like to do it.

ankur54:
@ScottGrogin Is this issue still open? I would like to work on this asap if possible. Please let me know which file to change. Thank you

limzykenneth:
What is the proposed implementation? Because with something like `(n%1).toFixed(4)` although practically it probably would not cause much problem, you are limited in terms of the number of precision after the decimal point, not to mention that `toFixed` returns a string and not a number (so does it belongs with String Functions such as `nf` or Math Functions such as `floor`?) Simply doing `n%1` would cause floating point precision issue which is why the stackoverflow solution have the `toFixed` portion.

@vedhant @ankur54 If any of you are interested in looking into implementing this, I would suggest fleshing out what your proposed implementation would be in this issue first before heading straight into the source code. Otherwise the likelyhood of said PR is rejected is higher when you did not have a clear idea about the implementation that the community has looked at, because there could very likely be edge cases that you didn't account for that we would have otherwise spotted and fixed while in the discussion phases here on the issue.

sanketsingh24:
[...]

PR #4069, opened 2019-10-11
Added fract() method to calculation.js

dhruvs009:
addresses #3487 (may need some changes still)

 Changes: 

fract() method was added to calculation.js

#### PR Checklist

- [x] `npm run lint` passes

Also, this is the first time I've used a PR template and it was a bit difficult to use it so any help would really be appreciated. (I didn't think screenshots were applicable here, so I removed it)

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -854,7 +854,8 @@ function hypot(x, y, z) {
 p5.prototype.fract = function(num) {
   p5._validateParameters('fract', arguments);
   let sign = 0;
-  if (num === undefined || Math.abs(num) === Infinity) {
+  num = Number(num);
+  if (isNaN(num) || Math.abs(num) === Infinity) {
     return num;
   } else if (num < 0) {
     num = -num;
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -851,10 +851,10 @@ function hypot(x, y, z) {
  * @alt
  * 2 rows of numbers, the first row having 8 numbers and the second having the fractional parts of those numbers.
  */
-p5.prototype.fract = function(num) {
+p5.prototype.fract = function(toConvert) {
   p5._validateParameters('fract', arguments);
   let sign = 0;
-  num = Number(num);
+  let num = Number(toConvert);
   if (isNaN(num) || Math.abs(num) === Infinity) {
     return num;
   } else if (num < 0) {
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -869,11 +869,11 @@ p5.prototype.fract = function(num) {
     num = -num;
     sign = 1;
   }
-  if (String(num).includes('.') && !String(num).includes('e')){
+  if (String(num).includes('.') && !String(num).includes('e')) {
     let toFract = String(num);
-    toFract = Number('0' + toFract.slice(toFract.indexOf('.')))
-    return Math.abs(sign-toFract);
-  } else if (num<1) return Math.abs(sign-num);
+    toFract = Number('0' + toFract.slice(toFract.indexOf('.')));
+    return Math.abs(sign - toFract);
+  } else if (num < 1) return Math.abs(sign - num);
   else return 0;
 };
 
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -841,20 +841,10 @@ function hypot(x, y, z) {
  * function setup() {
  *   createCanvas(windowWidth, windowHeight);
  *   fill(0);
- *   text(String(Infinity), 0, 50);
- *   text(String(fract(Infinity)), 0, 100);
- *   text(String(1e-10), 150, 50);
- *   text(String(fract(1e-10)), 150, 100);
- *   text(String(1.4215e-10), 300, 50);
- *   text(String(fract(1.4215e-10)), 300, 100);
- *   text(String(14213.214e-10), 450, 50);
- *   text(String(fract(14213.214e-10)), 450, 100);
- *   text(String(133232e10), 600, 50);
- *   text(String(fract(133232e10)), 600, 100);
- *   text(String(1.2412), 750, 50);
- *   text(String(fract(1.2412)), 750, 100);
- *   text(String('241.12421'), 900, 50);
- *   text(String(fract('241.12421')), 900, 100);
+ *   text(7345.73472742, 0, 50);
+ *   text(fract(7345.73472742), 0, 100);
+ *   text(1.4215e-15, 150, 50);
+ *   text(fract(1.4215e-15), 150, 100);
  * }
  * </code>
  * </div>
@@ -864,8 +854,9 @@ function hypot(x, y, z) {
 p5.prototype.fract = function(num) {
   p5._validateParameters('fract', arguments);
   let sign = 0;
-  if (num === undefined || Math.abs(num) === Infinity) return num;
-  else if (num < 0) {
+  if (num === undefined || Math.abs(num) === Infinity) {
+    return num;
+  } else if (num < 0) {
     num = -num;
     sign = 1;
   }
@@ -873,8 +864,11 @@ p5.prototype.fract = function(num) {
     let toFract = String(num);
     toFract = Number('0' + toFract.slice(toFract.indexOf('.')));
     return Math.abs(sign - toFract);
-  } else if (num < 1) return Math.abs(sign - num);
-  else return 0;
+  } else if (num < 1) {
+    return Math.abs(sign - num);
+  } else {
+    return 0;
+  }
 };
 
 export default p5;
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -841,21 +841,20 @@ function hypot(x, y, z) {
  * function setup() {
  *   createCanvas(windowWidth, windowHeight);
  *   fill(0);
- *   let nums = [
- *     Infinity,
- *     1e-10,
- *     1.4215e-10,
- *     14213.214e-10,
- *     142144.42142241241212521e10,
- *     133232e10,
- *     1.2412,
- *     '241.12421'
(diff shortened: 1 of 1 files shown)


src/math/calculation.js
```diff
@@ -829,4 +829,58 @@ function hypot(x, y, z) {
...
+ *     142144.42142241241212521e10,
+ *     133232e10,
+ *     1.2412,
+ *     '241.12421'
+ *   ];
+ *   let spacing = 150;
+ *   for (let i = 0; i < nums.length; i++) {
+ *     text(String(nums[i]), i * spacing, 50);
+ *     text(String(fract(nums[i])), i * spacing, 100);
+ *   }
```
lmccart: this example might be a bit hard for beginners to understand because it makes use of arrays and for loops. can you update to make this an example that does a single fract() command?

src/math/calculation.js
```diff
@@ -829,4 +829,58 @@ function hypot(x, y, z) {
...
+ * }
+ * </code>
+ * </div>
+ * @alt
+ * 2 rows of numbers, the first row having 8 numbers and the second having the fractional parts of those numbers.
+ */
+p5.prototype.fract = function(num) {
+  p5._validateParameters('fract', arguments);
+  var sign = 0;
```
lmccart: should be `let`

src/math/calculation.js
```diff
@@ -829,4 +829,58 @@ function hypot(x, y, z) {
...
+  if (num === undefined) return undefined;
+  if (num < 0) {
```
lmccart: these can be `else if`'s

src/math/calculation.js
```diff
@@ -829,4 +829,58 @@ function hypot(x, y, z) {
...
+    num = -num;
+    sign = 1;
+  }
+  if (num === Infinity) return num;
+  else if (
+    String(num).includes('e') &&
+    String(num)[String(num).indexOf('e') + 1] === '-'
```
lmccart: this is sort of confusing to follow. can you simply or break into multiple steps?

src/math/calculation.js
```diff
@@ -829,4 +829,58 @@ function hypot(x, y, z) {
...
+ */
+p5.prototype.fract = function(num) {
+  p5._validateParameters('fract', arguments);
+  var sign = 0;
+  if (num === undefined) return undefined;
+  if (num < 0) {
+    num = -num;
+    sign = 1;
+  }
+  if (num === Infinity) return num;
```
lmccart: this could be combined with check on 868

src/math/calculation.js
```diff
@@ -864,22 +863,17 @@ function hypot(x, y, z) {
  */
 p5.prototype.fract = function(num) {
   p5._validateParameters('fract', arguments);
-  var sign = 0;
-  if (num === undefined) return undefined;
-  if (num < 0) {
+  let sign = 0;
+  if (num === undefined || Math.abs(num) === Infinity) return num;
+  else if (num < 0) {
```
limzykenneth: I'd prefer for the braces to be more explicit here:

```javascript
if (num === undefined || Math.abs(num) === Infinity) {
  return num;
} else if (num < 0) {
  // ....
}
```

src/math/calculation.js
```diff
@@ -864,22 +863,17 @@ function hypot(x, y, z) {
...
-    return Math.abs(sign - num);
-  else if (String(num).includes('.'))
-    return Math.abs(
-      sign - Number('0' + String(num).slice(String(num).indexOf('.')))
-    );
+  if (String(num).includes('.') && !String(num).includes('e')){
+    let toFract = String(num);
+    toFract = Number('0' + toFract.slice(toFract.indexOf('.')))
+    return Math.abs(sign-toFract);
+  } else if (num<1) return Math.abs(sign-num);
```
limzykenneth: Same as above, I'd prefer not to omit the braces here

src/math/calculation.js
```diff
@@ -841,21 +841,20 @@ function hypot(x, y, z) {
...
+ *   text(String(1.4215e-10), 300, 50);
+ *   text(String(fract(1.4215e-10)), 300, 100);
+ *   text(String(14213.214e-10), 450, 50);
+ *   text(String(fract(14213.214e-10)), 450, 100);
+ *   text(String(133232e10), 600, 50);
+ *   text(String(fract(133232e10)), 600, 100);
+ *   text(String(1.2412), 750, 50);
+ *   text(String(fract(1.2412)), 750, 100);
+ *   text(String('241.12421'), 900, 50);
+ *   text(String(fract('241.12421')), 900, 100);
```
limzykenneth: There's too many cases here. The examples are not meant to show all possible uses of the function so just one or two examples will be enough. Also is the `String()` conversion needed here?

limzykenneth:
When the necessary changes are made here I would like to do some local testing as well so maybe hold up on the merge for a bit after the changes are made.

src/math/calculation.js
dhruvs009: I used the `String()` conversion to show that the function also works for undefined input, but if all these cases are not needed I'll remove them as you asked me to.

dhruvs009:
Hey, @limzykenneth I have made the changes you and @lmccart asked me for. I have tested this implementation quite extensively, but there still might be some test cases that would cause problems, do let me know if you find any. Thanks :)

src/math/calculation.js
```diff
@@ -829,4 +829,46 @@ function hypot(x, y, z) {
...
+ * }
+ * </code>
+ * </div>
+ * @alt
+ * 2 rows of numbers, the first row having 8 numbers and the second having the fractional parts of those numbers.
+ */
+p5.prototype.fract = function(num) {
+  p5._validateParameters('fract', arguments);
+  let sign = 0;
+  if (num === undefined || Math.abs(num) === Infinity) {
```
limzykenneth: Can you explain a bit about why there's a check for `undefined` here? Mainly because there are several other cases to handle if you want to handle `undefined` (`null`, `NaN`, `"text string"`, etc). Any of the above values should already be caught with the FES but if you want to catch them and return them as is (the unhandled ones return 0 at the moment), you can just check here if `num` is a Number.

limzykenneth:
I've done my tests and just one last point left from me. It's nearly there so thank you so much @dhruvs009 for your patience while we finish this up.

src/math/calculation.js
dhruvs009: Well, firstly my initial reason to handle `undefined` was that if not it showed an error (expected number of arguments), but then I thought that if by any mistake, I put a `null, NaN, "text string"` in there, I wouldn't want 0 to be returned, because the nature of `fract()` makes sure that 0 is a very important number and can cause a lot of errors in computations. And I think I forgot changing the check to checking if `num` is a number. I think that change would suffice most incorrect inputs?

dhruvs009: Revert back to me on this and I'll make the necessary changes as soon as you do.

limzykenneth: Yeah, just make the check for the number type. I don't think it is necessary to specially handle `undefined` since it would be taken care of by the number type check and it wouldn't suppress the FES message anyway (not that it's necessary).

dhruvs009:
```javascript
function fract(num){
    let sign = 0;
    num=Number(num)
    if (typeof num != Number || Math.abs(num) === Infinity) return num;
    else if (num < 0) {
      num = -num;
      sign = 1;
    }
    if (String(num).includes('.') && !String(num).includes('e')){
      let toFract = String(num);
      toFract = Number('0' + toFract.slice(toFract.indexOf('.')))
      return Math.abs(sign-toFract);
    } else if (num<1) return Math.abs(sign-num);
    else return 0;
};

console.log(fract(7345.73472742))
console.log(fract(41251521531625123e-25))
console.log(fract("HelloWorld"))
console.log(fract("21749028.42154"))
console.log(fract(NaN))
console.log(fract())
``` 

Hey, how about something like this?

dhruvs009: I performed the `Number` typecasting at line 3 to take care of something like example 4.

limzykenneth: `typeof` returns a string so it should be `typeof num != "number"`.

dhruvs009: Oh, right. I'm sorry. I've updated the code accordingly.

dhruvs009: I'm sorry. There still seems to be something wrong with this.

dhruvs009: I've updated the code. I think it should work now (just had a brainfade moment).

dhruvs009:
This seems to be working fine for me. Let me know, if there still are any bugs. Thanks :)

src/math/calculation.js
```diff
@@ -854,7 +854,8 @@ function hypot(x, y, z) {
 p5.prototype.fract = function(num) {
   p5._validateParameters('fract', arguments);
   let sign = 0;
-  if (num === undefined || Math.abs(num) === Infinity) {
+  num = Number(num);
```
limzykenneth: Can the casted value from `Number(num)` be assigned to a new variable instead of reusing `num`? Problem could arise if someone accidentally pass an Object into `fract` and this line will change the variable holding the object into `NaN`.

dhruvs009: I think it works now?

limzykenneth:
@lmccart Do you want to have a final look? I'm ok for this to merge otherwise.

dhruvs009:
@limzykenneth also, do I ask you to add me to the contributors list or do I make a PR myself?

limzykenneth:
I think there's a bot that can do it automatically when merging.

dhruvs009:
Oh it didn't happen on the previous PR that got merged.

lmccart:
@all-contributors please add @dhruvs009 for code, documentation

PR #4079, opened 2019-10-14
adding FES overview closes #3371

### LINKED ISSUE #3371: Add overview to /src/core/error_helpers.js (FES)
Individual functions for the Friendly Error System are well documented in /src/core/error_helpers.js, but an overview of how these functions fit together would make this code easier to understand for contributors looking to get familiar with this part of the library. The idea here is to add a paragraph or list of bullet points of the order of operations for which FES functions are called when.

lmccart:
Resolves #3371

#### Changes: 
Updating /src/core/error_helper.js to include an overview of the FES functions that are called. This could be expanded on further, but I figured having something there was better than nothing. Hopefully this provides a useful starting place. I think rather than adding much more to the overview at the top of the file, more work could be put into documenting the individual functions in this file that are mentioned in the overview outline. Feedback welcome!

#### PR Checklist
- [x] `npm run lint` passes

### DIFF
--- a/src/core/error_helpers.js
+++ b/src/core/error_helpers.js
@@ -1,8 +1,46 @@
 /**
  * @for p5
  * @requires core
+ *
+ * This is the main file for the Friendly Error System (FES). Here is a
+ * brief outline of the functions called in this system.
+ *
+ * The FES may be invoked by a call to either (1) _validateParameters,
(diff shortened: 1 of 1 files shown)


outofambit:
yay thank you!

PR #4117, opened 2019-10-28
friendly autoplay warning

### LINKED ISSUE #3368: Update FES to reflect new Audio Autoplay Policy changes
Update the Friendly Error System to warn if a user is trying to use code that assumes audio will autoplay while it may instead require user interaction. This is a result of Audio Autoplay Policy changes in Chrome (https://developers.google.com/web/updates/2017/09/autoplay-policy-changes#webaudio).

#### Nature of issue?

- [X] Existing feature enhancement

- [X] Other (specify if possible) Friendly Error System

outofambit:
Resolves #3368 

## Changes

Covers two possible ways to have autoplay fail:
- from a programmatic call to `play()`
  - this was pretty straightforward, as this error is explicitly thrown by the browser if the play violates the autoplay policy
- setting `autoplay(true)`, which instructs the browser to autoplay as soon as its ready
  - this was a bit more involved, as there is no event you can listen for this kind of failure. i've written logic that takes a best guess—if autoplay is enabled, it waits for the media to be loaded enough to play, and then sends an error message if the media _doesn't_ actually play in the next half second.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/error_helpers.js
+++ b/src/core/error_helpers.js
@@ -218,6 +218,19 @@ if (typeof IS_MINIFIED !== 'undefined') {
     report(message, method);
   };
 
+  /**
+   * This is called internally if there is a error with autoplay.
+   *
+   * @method _friendlyAutoplayError
+   * @private
+   */
+  p5._friendlyAutoplayError = src => {
+    report(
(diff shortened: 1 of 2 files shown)


lmccart:
👏

PR #4126, opened 2019-11-01
adding style guide notes for yuidoc

lmccart:
Rather than creating a whole styleguide document, I think what we have here is pretty good. I augmented it with some key notes from the #1190, and we can keep adding to it. Are there others you think should be here?

Resolves https://github.com/processing/p5.js/issues/1190


limzykenneth:
Is there somewhere in the docs that mentions the use of `<code>` tags and the `norender` class stuff?

lmccart:
@limzykenneth yeah, that's in the [adding examples](https://github.com/processing/p5.js/blob/master/developer_docs/inline_documentation.md#adding-example-code) section. I'm thinking of these points here as more general / mostly pertaining to the description section.

Issue #3966, opened 2019-08-25
image() in WEBGL affected by the material/lighting conditions


stalgiag:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Which platform were you using when you encountered this?

- [ ] Mobile/Tablet (touch devices)
- [x] Desktop/Laptop
- [ ] Others (specify if possible)

#### Details about the bug: 

This is more of a question about what is expected. Should calls to `image()` in WEBGL mode use the current lighting and material conditions? Currently if you call `image(img)` it behaves similarly to `texture(img); rect()` which is to say that it takes on the current material and lighting state. [Here is an example that shows this.](https://editor.p5js.org/stalgiag/sketches/weaB1OeDZ)

My instinct is that people will usually expect `image()` to appear as unlit, similar to 2D mode.

singhvisha:
hi @stalgiag , Can I work on this issue?

stalgiag:
Hi @singhvisha yes definitely. But first we should decide what the expected behavior is. What do you think?

stalgiag:
Okay I think that this has been open long enough to draw any likely responses. I definitely agree that `image()` should be unlit.

This could be a good opportunity for contribution from new contributors interested in WebGL.

singhvisha:
@stalgiag @aferriss 
hi , Can I work on this issue?

stalgiag:
@singhvisha sounds good!

stalgiag:
Closed via #4128

PR #4128, opened 2019-11-02
Modified the image() function(#3966)

singhvisha:
Modified the image() function so that image don't get affected by material/lighting condition in WEBGL mode(#3966)

### DIFF
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -419,31 +419,6 @@ p5.prototype.image = function(
   _sh *= pd;
   _sw *= pd;
 
-  if (this._renderer.isP3D) {
-    this._renderer.ambientLightColors.length = 0;
-    this._renderer.specularColors = [1, 1, 1];
-
-    this._renderer.directionalLightDirections.length = 0;
-    this._renderer.directionalLightDiffuseColors.length = 0;
-    this._renderer.directionalLightSpecularColors.length = 0;
-
-    this._renderer.pointLightPositions.length = 0;
-    this._renderer.pointLightDiffuseColors.length = 0;
-    this._renderer.pointLightSpecularColors.length = 0;
-
-    this._renderer.spotLightPositions.length = 0;
-    this._renderer.spotLightDirections.length = 0;
-    this._renderer.spotLightDiffuseColors.length = 0;
-    this._renderer.spotLightSpecularColors.length = 0;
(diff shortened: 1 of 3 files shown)


singhvisha:
hi @stalgiag ,sorry for the delay.my college exams were going on.
Can you please review it?

stalgiag:
Hi @singhvisha it might be better to do this in the [p5.RendererGL.prototype.image](https://github.com/processing/p5.js/blob/master/src/webgl/3d_primitives.js#L1698) rather than doing that check for 3D. This will help with organization. Also you may be able to call [`noLights()` ](https://github.com/processing/p5.js/blob/master/src/webgl/light.js#L869)instead to avoid rewriting code.

singhvisha:
Hi @stalgiag ,Can you please review it?

stalgiag:
Hi @singhvisha. Close! But you call `noLights()` after the drawing is complete. This won't have the affect of changing the material/lighting conditions of the image, instead it will just turn off all lights before continued drawing.

It is important to test your changes to ensure they work whenever you edit the source code. I often use the empty manual test example to do this. In this instance a sketch like this should reveal whether your changes are working.

```js
let img;

function preload() {
  img = loadImage('../webgl/assets/cat.jpg')
}

function setup() {
  createCanvas(500, 500, WEBGL);
}

function draw() {
  background(0);
  pointLight(200, 200, 200, 0, 0, 0);
  translate(-200, 0, 0);
  box(100);
  image(img, 150, -50, 100, 100);
  translate(400, 0, 0);
  box(100);
}
```
Note that example only works from `test/manual-test-examples/empty-test/sketch.js`. The above example should have an unlit image with the inside edges of the two boxes lit by the point light.

singhvisha:
hi @stalgiag,
I have run the example locally this time for the changes that I have made in the source code.It seems to working for what you have suggested above.Could you please review the code?

singhvisha:
@stalgiag 
Could you please review the above changes?

stalgiag:
Hi @singhvisha 

Looks good! Good catch on the fact that `noLights()` doesn't set `rendererGL._enableLighting = false;`. That should always be set to false when disabling the lights. It might be a good idea to move that line into the `noLights()` function to ensure that state is always managed correctly.

singhvisha:
hi @stalgiag 
I have added enablingLighting property in light.js.Could you please review?

PR #4156, opened 2019-11-25
Only run GitHub Actions CI on "Pull request" events

outofambit:
## Changes: 

- restricted the `push` event trigger for GitHub actions to only work on the master branch, since its causing duplicate test runs for pull requests. (because we are using the `pull_request` event trigger.)
- some formatting

this means we won't run CI for branches that aren't in a pull request, except for the `master` branch.

#### PR Checklist

- [x] `npm run lint` passes


outofambit:
@gruselhaus how does this look to you?

PR #4164, opened 2019-11-28
use rect() with 3 parameters, similar to how ellipse() works

### LINKED ISSUE #4163: allow rect() use just 3 parameters, like ellipse() - so both words can be hotswapped
#### Nature of issue?

`ellipse()` allows one to provide just 3 parameters ie. `ellipse(x, y, w)` and it auto-magically uses the `w` for the `h` to make a perfect circle. This functionality is really efficient (`circle` and `square` were helpful learning additions too!) – and the pending PR will just add the same feature to `rect()`, since it's confusing when the shape dissapears if swapping the word `ellipse` to `rect` with just three params (since the 4th was required with `rect`). Tested with `rectMode()` and all seems to work fine.

```
// now both will work
ellipse(mouseX, mouseY, 50);
rect(mouseX, mouseY, 50);
```

ffd8:
Resolves #4163 

 Changes: 
As mentioned in issue, simply allows `rect()` to work with just 3 parameters, like `ellipse()` so the words can be easily swapped while sketching. The question has come up on the issue referenced above what happens if one is using rounded corners– however this enhancement is simply to deal with just 3 parameters that `ellipse()` can handle. If one wants to add rounded corners, then the additional height should be added or square used. 

- [x] `npm run lint` passes

### DIFF
--- a/src/core/shape/2d_primitives.js
+++ b/src/core/shape/2d_primitives.js
@@ -560,6 +560,10 @@ p5.prototype.rect = function() {
   p5._validateParameters('rect', arguments);
 
   if (this._renderer._doStroke || this._renderer._doFill) {
+    if (arguments.length === 3) {
+      arguments[3] = arguments[2];
+    }
+
     const vals = canvas.modeAdjust(
       arguments[0],
       arguments[1],
@@ -567,6 +571,7 @@ p5.prototype.rect = function() {
       arguments[3],
       this._renderer._rectMode
     );
+
     const args = [vals.x, vals.y, vals.w, vals.h];
     // append the additional arguments (either cornder radii, or
     // segment details) to the argument list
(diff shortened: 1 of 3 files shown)


src/core/shape/2d_primitives.js
```diff
@@ -560,13 +560,18 @@ p5.prototype.rect = function() {
   p5._validateParameters('rect', arguments);
```
mishazawa: Maybe will be easier to understanding code if rewrite this function with default parameters, f.e.
```js
p5.prototype.rect = function(x, y, w, h = w, ...rest) {
  p5._validateParameters('rect', [x, y, w, h, ...rest]);

  if (this._renderer._doStroke || this._renderer._doFill) {
    const vals = Object.values(canvas.modeAdjust(x, y, w, h, this._renderer._rectMode));
    this._renderer.rect([...vals, ...rest]);
  }
  return this;
}
```

What do you think about this?

ffd8: I wonder if the `rect()` function is older than the `ellipse()` or written by someone else? Since that seems to be the function style used in `ellipse()`. Can't say I have an opinion either way – haha.. just wanted to allow my students to change `ellipse()` to `rect()` with the shape dissapearing.

lmccart:
looks great, thanks! @all-contributors please add @ffd8 for code

PR #4177, opened 2019-12-03
Make animation delay a frame-specific value with GIFs

### LINKED ISSUE #3872: GIF delays are averaged across frames
#### Nature of issue?

- [ ] Found a bug
- [ ] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [x] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

- [ ] Desktop/Laptop

#### Summary
Currently when decoding GIF files, p5 [averages the frame delay across all frames. ](https://github.com/processing/p5.js/blob/master/src/image/loading_displaying.js#L216). This is so that the `delay` value in `gifProperties` can be a single number, and there doesn't need to be a unique delay for every frame. This works for most GIFs, since most GIFs have a consistent delay across frames. 

This issue can serve as a discussion point. If people find that this behavior causes problems, they can post here, and if necessary changes can be made.

#### Potential Fixes
delay can be `integer[]` with a length of `numFrames`
or
A frame can become an object `frame = {imageData, delay}`

stalgiag:
resolves #3872 

GIF frames are now objects with this format:
```js
{image: ImageData, delay: Number}
```

I chose this route instead of making delay an array because I wanted to keep a conceptual link between the frame image and the frame delay. This occasionally results in verbose code but I think the structure of `gifProperties` is more intelligible this way.

With this PR, the [frame-level delay example](https://hamoid.tumblr.com/post/79459820142/twenty-four-point-fifty-five-seconds-attempt-to) given by @hamoid works correctly in p5.

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -371,7 +371,7 @@ p5.Renderer2D.prototype.updatePixels = function(x, y, w, h) {
   h *= pd;
 
   if (this.gifProperties) {
-    this.gifProperties.frames[this.gifProperties.displayIndex] =
+    this.gifProperties.frames[this.gifProperties.displayIndex].image =
       pixelsState.imageData;
   }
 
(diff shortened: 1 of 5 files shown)


lmccart:
👍

PR #4199, opened 2019-12-18
Change object member functions to use regular functions

### LINKED ISSUE #3875: Use of arrow functions for prototype definition

limzykenneth:
Resolves #3875 

 Changes: 

As per the issue, prototype methods are all reverted to using regular function regardless of whether they use `this` in the function body or not. One additional thing I think is necessary but I'm not 100% sure about (90% sure) is static methods of objects. From what I found, in particular this [stackoverflow question](https://stackoverflow.com/questions/52269317/es6-javascript-calling-static-methods-from-classes-with-arrow-functions), it seems it is good practice to use regular functions for static members but I found only reference to `class` syntax static members only.

In any case, none of the converted functions use `this` in their function body so the functionalities will be exactly the same as before.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/color/color_conversion.js
+++ b/src/color/color_conversion.js
@@ -19,7 +19,7 @@ p5.ColorConversion = {};
 /**
  * Convert an HSBA array to HSLA.
  */
-p5.ColorConversion._hsbaToHSLA = hsba => {
+p5.ColorConversion._hsbaToHSLA = function(hsba) {
   const hue = hsba[0];
   let sat = hsba[1];
   const val = hsba[2];
@@ -45,7 +45,7 @@ p5.ColorConversion._hsbaToHSLA = hsba => {
 /**
  * Convert an HSBA array to RGBA.
  */
-p5.ColorConversion._hsbaToRGBA = hsba => {
+p5.ColorConversion._hsbaToRGBA = function(hsba) {
   const hue = hsba[0] * 6; // We will split hue into 6 sectors.
   const sat = hsba[1];
   const val = hsba[2];
@@ -100,7 +100,7 @@ p5.ColorConversion._hsbaToRGBA = hsba => {
 /**
  * Convert an HSLA array to HSBA.
  */
-p5.ColorConversion._hslaToHSBA = hsla => {
+p5.ColorConversion._hslaToHSBA = function(hsla) {
   const hue = hsla[0];
   let sat = hsla[1];
   const li = hsla[2];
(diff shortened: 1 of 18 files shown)


lmccart:
thanks @limzykenneth this all looks good. I think you're correct on using regular functions for all static members.

PR #4229, opened 2020-01-05
Update issue templates

### LINKED ISSUE #4228: Updating to the new issue template workflow
If you view [the ISSUE_TEMPLATE.md file](https://github.com/processing/p5.js/blob/master/ISSUE_TEMPLATE.md) you will see that there is a message encouraging us to update to the new issue template workflow.

This will have some advantages, most important being that it will allow us to offer different templates for different issue subtypes: bug, feature request, feature enhancement. This will allow us to reduce the complexity of the issue template.

stalgiag:
resolves #4228 

This updates the repo to the new issue template workflow. With this PR, users will choose between 3 branching issue templates:
- Found a Bug
- Existing Feature Enhancement
- Feature Request

Each of these templates is different and smaller than the current template. Additionally, they come with default labels.


lmccart:
Cool. I love the interface. It makes me wonder if we might want one more template that is "Coding Question" or similar. that is actually meant to redirect people to discourse.processing.org. The template could simply say "Please post at discourse.processing.org to get coding help rather than opening an issue here." Or is that confusing. Thoughts?
Screen Shot 2020-01-05 at 3 54 16 PM

stalgiag:
@lmccart I think that could work well. I don't think any solution will ever redirect all the correct people, but that one seems friendly enough. As a bonus, maybe this would allow us to remove the paragraph that does similar work in the other templates ("PLEASE NOTE: The github issues are for bugs ...").

lmccart:
@stalgiag yeah I think so, that would simplify those templates in a nice way

stalgiag:
@lmccart done!

PR #4233, opened 2020-01-06
add internationalization tooling

### LINKED ISSUE #3390: i18n for FES messages

outofambit:
Resolves #3390 

## Changes: 

Adds `i18next`, a js library for internationalization to the p5.js codebase. It lets us use translated for text appropriate to the user's browser/computer configuration. It doesn't translate text for us though, that must be done by hand. It provides language detection and translation loading/use.

I've set up `i18next` for use anywhere in the library, with translations files for English and Spanish to start. (I only filled out one Spanish translation right now though. 😢) I've integrated it with all the FES messages _except_ the parameter validation ones, which would be a great next step.

I've also kept this out of the minified build (at least for now) since it affects the initial loading and size of the p5.js library. I think the impacts would be minimal, but I'm playing it safe.

Did a little cleanup in error_helpers.js along the way, too.

## Screenshot

Screen Shot 2020-02-24 at 11 25 27 PM

## PR Checklist

- [x] `npm run lint` passes
- [x] [Unit tests] are included
- [x] [Contributor docs] are included


limzykenneth:
Is the translation text included in the final build itself? I'm a bit worried that it would blow up the size of the library too much especially after including more and more translations over time. 

An idea I had before is to distribute localized versions of the library depending on the browser's language setting that is used to download the library or serving different versions on the web editor. Not sure how viable it is though.

Another idea is to have the translations as an addon that we distribute along with the complete zip package and automatically include on the web editor that can be removed by the user if needed.

outofambit:
> Is the translation text included in the final build itself? I'm a bit worried that it would blow up the size of the library too much especially after including more and more translations over time.

@limzykenneth yep the translation text included in the final build, though only in the minified version since we disable FES in that build anyways. i'm going to see how many different strings there are for FES so we can get an idea of what the scale is here.

I also like your alternatives, especially the translations as an extra download add-on. (i agree automatically including them in the web editor would be ideal, too!) We could also host translations online so sketches can fetch them if needed. That would require the sketch to have internet access, but maybe that's okay?

lmccart:
@outofambit checking in on this one. is this ready to merge or do we want to look into hosting the translations externally?

outofambit:
@lmccart i think the first version could be embedded and we can look into hosting externally aferwards! let me do a little more testing/documenting tonight to make sure this is good!

outofambit:
@lmccart i think this is ready for review + merge!

cc @limzykenneth in case he has time to look at this again!

PR #4239, opened 2020-01-07
Modified light documentation

### LINKED ISSUE #4234: Limited number of lightsources ?
Is the numbers of lightsources (pointlight, directional ...) you can add to a 3-dimensional scene limited in some way?

I am trying to figure out, why some lightsources in my scene are being rendered, while others are not.

AnuragGupta93:
Fixes #4234 
Added maximum number of light possible for these.
* directionalLight()
* pointLight()
* spotLight()

### DIFF
--- a/src/webgl/light.js
+++ b/src/webgl/light.js
@@ -169,6 +169,8 @@ p5.prototype.specularColor = function(v1, v2, v3) {
 
 /**
  * Creates a directional light with a color and a direction
+ * <br/>
+ * Maximum limit of directional possible: 5
  * @method directionalLight
  * @param  {Number}    v1       red or hue value (depending on the current
  * color mode),
@@ -271,6 +273,8 @@ p5.prototype.directionalLight = function(v1, v2, v3, x, y, z) {
 
 /**
  * Creates a point light with a color and a light position
+ * <br/>
+ * Maximum limit of pointLight possible: 5
  * @method pointLight
  * @param  {Number}    v1       red or hue value (depending on the current
  * color mode),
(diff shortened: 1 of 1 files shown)


src/webgl/light.js
```diff
@@ -169,6 +169,8 @@ p5.prototype.specularColor = function(v1, v2, v3) {
 
 /**
  * Creates a directional light with a color and a direction
+ * <br/>
```
lmccart: You can omit these `<br/>` elements and just leave a blank line and it will translate into a line break.

AnuragGupta93: Thanks for the information. Doing right now.

AnuragGupta93: Do I also need to add a workaround for increasing the maximum no. of lights in the documentation?

Means changing the array size in this file and building p5 from source https://github.com/processing/p5.js/blob/e11b4d617973a6cbeb789302ac093d26b8f2e489/src/webgl/shaders/lighting.glsl#L12

stalgiag: No need to add the workaround to the documentation. Thank you!

src/webgl/light.js
```diff
@@ -499,6 +503,8 @@ p5.prototype.lightFalloff = function(
  * light towards the center. Both angle and concentration are optional, but if
  * you want to provide concentration, you will also have to specify the angle.
  *
+ * Maximum limit of pointLight possible: 5
```
stalgiag: This comment is for spotLight

src/webgl/light.js
```diff
@@ -169,6 +169,8 @@ p5.prototype.specularColor = function(v1, v2, v3) {
 
 /**
  * Creates a directional light with a color and a direction
+ *
+ * Maximum limit of directional possible: 5
```
stalgiag: I think that it would be better to use a complete sentence in order to keep with the writing style of the rest of the reference.

Something to the effect of 'A maximum of 5 directional lights can be active at one time.`

src/webgl/light.js
AnuragGupta93: Done

src/webgl/light.js
AnuragGupta93: Done

stalgiag:
@all-contributors please add @AnuragGupta93 for documentation

PR #4246, opened 2020-01-09
added rem() method on p5.Vector

### LINKED ISSUE #3861: Introduce p5.Vector#rem to calculate the remainder/modulo of a vector.
#### Nature of issue?

- [ ] Found a bug
- [x] New feature request

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### New feature details:

(PR with this change: https://github.com/processing/p5.js/pull/3860)

Introduce p5.Vector#rem to calculate the [remainder/modulo](https://en.wikipedia.org/wiki/Modulo_operation) of a vector. . This would be helpful for drawing particles that need to wrap when they go offscreen.

For example, instead of:

```js
particle.position.add(this.velocity);
particle.position.set(
  particle.position.x % width,
  particle.position.y % height,
);
```

Now one can write:

```js
particle.position.add(particle.velocity).rem(width);
```

The signature looks like this:

```js
p5.Vector.prototype.rem = function rem(n) {
  ...
};
```

The one-demension (one parameter) behavior matches what we do for `p5.Vector#mult` and `p5.Vector#div`, in contrast to `p5.Vector#add` and `p5.Vector#sub`. One drawback with this design is you can't calculate the remainder of more than one dimension. In the example above, if my canvas was not a square, I would not be able to use `p5.Vector#rem` to wrap particles since the width and height would be different numbers.

jatin33:
Resolves #3861 

 Changes: 

Adding support for remainder operation for p5.Vector objects.

 Screenshots of the change: 

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -346,17 +346,26 @@ const calculateRemainder2D = function(xComponent, yComponent) {
 const calculateRemainder3D = function(xComponent, yComponent, zComponent) {
   if (xComponent === 0 && yComponent === 0 && zComponent === 0) {
     return this;
-  } else if (xComponent === 0) {
+  } else if (xComponent === 0 && yComponent !== 0 && zComponent !== 0) {
     this.y = this.y % yComponent;
     this.z = this.z % zComponent;
     return this;
-  } else if (yComponent === 0) {
+  } else if (xComponent !== 0 && yComponent === 0 && zComponent !== 0) {
     this.x = this.x % xComponent;
     this.z = this.z % zComponent;
     return this;
-  } else if (zComponent === 0) {
+  } else if (xComponent !== 0 && yComponent !== 0 && zComponent === 0) {
     this.x = this.x % xComponent;
     this.y = this.y % yComponent;
+    return this;
+  } else if (xComponent === 0 && yComponent === 0 && zComponent !== 0) {
+    this.z = this.z % zComponent;
+    return this;
+  } else if (xComponent === 0 && yComponent !== 0 && zComponent === 0) {
+    this.y = this.y % yComponent;
+    return this;
+  } else if (xComponent !== 0 && yComponent === 0 && zComponent === 0) {
+    this.x = this.x % xComponent;
     return this;
   } else {
     this.x = this.x % xComponent;
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -326,6 +326,45 @@ p5.Vector.prototype.add = function add(x, y, z) {
   return this;
 };
 
+/// HELPERS FOR REMAINDER METHOD
+const calculateRemainder2D = function(xComponent, yComponent) {
+  if (xComponent === 0 && yComponent === 0) {
+    return this;
+  } else if (xComponent === 0) {
+    this.y = this.y % x[1];
+    return this;
+  } else if (yComponent === 0) {
+    this.x = this.x % x[0];
+    return this;
+  } else {
+    this.x = this.x % xComponent;
+    this.y = this.y % yComponent;
+    return this;
+  }
+};
+
+const calculateRemainder3D = function(xComponent, yComponent, zComponent) {
+  if (xComponent === 0 && yComponent === 0 && zComponent === 0) {
+    return this;
+  } else if (xComponent === 0) {
+    this.y = this.y % yComponent;
+    this.z = this.z % zComponent;
+    return this;
+  } else if (yComponent === 0) {
+    this.x = this.x % xComponent;
+    this.z = this.z % zComponent;
+    return this;
+  } else if (zComponent === 0) {
+    this.x = this.x % xComponent;
+    this.y = this.y % yComponent;
+    return this;
+  } else {
+    this.x = this.x % xComponent;
+    this.y = this.y % yComponent;
+    this.z = this.z % zComponent;
+    return this;
+  }
+};
 /**
  * Gives remainder of a vector when it is divided by another vector.
  * See examples for more context.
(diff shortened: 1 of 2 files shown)


jatin33:
image

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+ * @method rem
+ * @param {p5.Vector | Number[]}  value  divisor vector
+ * @chainable
+ */
+p5.Vector.prototype.rem = function rem(x, y, z) {
+  const calculateRemainder3D = (xComponent, yComponent, zComponent) => {
+    if (xComponent === 0 && yComponent === 0 && zComponent === 0) {
+      return this;
+    }
+    if (xComponent === 0 && yComponent !== 0 && zComponent !== 0) {
```
lmccart: these can all be `else if`

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+ * </code>
+ * </div>
+ */
+/**
+ * @method rem
+ * @param {p5.Vector | Number[]}  value  divisor vector
+ * @chainable
+ */
+p5.Vector.prototype.rem = function rem(x, y, z) {
+  const calculateRemainder3D = (xComponent, yComponent, zComponent) => {
```
lmccart: do these functions need to be defined within the function? it's preferable to define them outside to minimize execution time

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+    this.y = this.y % yComponent;
+    this.z = this.z % zComponent;
+    return this;
+  };
+
+  const calculateRemainder2D = (xComponent, yComponent) => {
+    if (xComponent === 0 && yComponent === 0) {
+      return this;
+    }
+    if (xComponent === 0 && yComponent !== 0) {
```
lmccart: this can be `else if` and the `&& yComponent !== 0` part can be removed

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+      this.y = this.y % x[1];
+      return this;
+    }
+    if (xComponent !== 0 && yComponent === 0) {
```
lmccart: this can be `else if` and the `xComponent !== 0 && ` part can be removed

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+      this.x = this.x % x[0];
+      return this;
+    }
+    this.x = this.x % xComponent;
```
lmccart: place this block in an `else` for clarity

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+      this.x = this.x % xComponent;
+      this.z = this.z % zComponent;
+      return this;
+    }
+    if (xComponent !== 0 && yComponent !== 0 && zComponent === 0) {
+      this.x = this.x % xComponent;
+      this.y = this.y % yComponent;
+      return this;
+    }
+    this.x = this.x % xComponent;
```
lmccart: place in `else` block for clarity

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+    this.y = this.y % yComponent;
+    return this;
+  };
+
+  if (x instanceof p5.Vector) {
+    if (Number.isFinite(x.x) && Number.isFinite(x.y) && Number.isFinite(x.z)) {
+      var xComponent = parseFloat(x.x);
```
lmccart: these should be `const`

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+      var yComponent = parseFloat(x.y);
+      var zComponent = parseFloat(x.z);
+      calculateRemainder3D(xComponent, yComponent, zComponent);
+    }
+  }
+
+  if (x instanceof Array) {
```
lmccart: `else if`

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+      if (x.length === 2) {
+        calculateRemainder2D(x[0], x[1]);
+      }
+      if (x.length === 3) {
+        calculateRemainder3D(x[0], x[1], x[2]);
+      }
+    }
+  }
+
+  if (arguments.length === 0) {
```
lmccart: `else if`

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+    if (arguments[0] === undefined) {
```
lmccart: can you explain this check?

src/math/p5.Vector.js
```diff
@@ -326,6 +326,141 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+      return this;
+    }
+  }
+
+  if (arguments.length === 1) {
```
lmccart: `else if`

src/math/p5.Vector.js
lmccart: I think you could change these methods to accept an array of length 3 (or 2 for 2D function). As it is now, you end up needing to split up the arrays to call the functions below.

src/math/p5.Vector.js
```diff
@@ -1636,6 +1771,32 @@ p5.Vector.add = function add(v1, v2, target) {
   return target;
 };
 
+// Returns a vector remainder when it is divided by another vector
+/**
+ * @method rem
+ * @static
+ * @param  {p5.Vector} v1 dividend <a href="#/p5.Vector">p5.Vector</a>
+ * @param  {p5.Vector} v2 divisor <a href="#/p5.Vector">p5.Vector</a>
+ * @param  {p5.Vector} target the vector to receive the result
```
lmccart: i would suggest removing this third param, and just making a new vector in the function. is there a reason to have the user pass one in?

test/unit/math/p5.Vector.js
```diff
@@ -275,6 +275,111 @@ suite('p5.Vector', function() {
...
+      test('should return same vector if any vector component is non-finite number', () => {
+        v.rem([2, 3, Infinity]);
+        expect(v.x).to.eql(3);
+        expect(v.y).to.eql(4);
+        expect(v.z).to.eql(5);
+      });
+    });
+
+    suite('p5.Vector.rem(v1,v2)', function() {
+      var v1, v2, res;
```
lmccart: should be `let`

src/math/p5.Vector.js
```diff
@@ -1636,6 +1771,32 @@ p5.Vector.add = function add(v1, v2, target) {
...
+ */
+/**
+ * @method rem
+ * @static
+ * @param  {p5.Vector} v1
+ * @param  {p5.Vector} v2
+ * @return {p5.Vector} the resulting <a href="#/p5.Vector">p5.Vector</a>
+ *
+ */
+p5.Vector.rem = function rem(v1, v2, target) {
```
limzykenneth: The function itself doesn't need to be named, especially considering another member function above have the same name as well so even for debugging it isn't especially useful.

src/math/p5.Vector.js
jatin33: > do these functions need to be defined within the function? it's preferable to define them outside to minimize execution time

- These functions are defined inside because they are borrowing **this** keyword from the parent scope. I will refactor them as helper functions outside the scope. Thanks for pointing that out as I read about inner functions and found that they hamper function execution time. I think the reason behind this might be that the execution state for the inner function is maintained due to closure. Am I thinking right?

src/math/p5.Vector.js
jatin33: Thanks for spoting this out. This change will be done!

src/math/p5.Vector.js
jatin33: Looked at the code again and removed this block and tests are running !

test/unit/math/p5.Vector.js
jatin33: Made the necessary changes.

src/math/p5.Vector.js
limzykenneth: Just noticed member functions in p5.Vector are all named while others throughout the codebase aren't. I think you can leave this for now as this seems to be a separate issue of its own.

src/math/p5.Vector.js
jatin33: > I think you could change these methods to accept an array of length 3 (or 2 for 2D function). As it is now, you end up needing to split up the arrays to call the functions below.

Are you suggesting to write functions in below mentioned way?
```
/**
* @param Number[]
*  returns Number[]
*/
function calculateRemainder2D(array){
}
```

src/math/p5.Vector.js
```diff
@@ -326,6 +326,129 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+    this.x = this.x % xComponent;
+    this.y = this.y % yComponent;
+    return this;
+  }
+};
+
+const calculateRemainder3D = function(xComponent, yComponent, zComponent) {
+  if (xComponent === 0 && yComponent === 0 && zComponent === 0) {
+    return this;
+  } else if (xComponent === 0) {
```
lmccart: does this account correctly for the case where two of the three are 0?

src/math/p5.Vector.js
```diff
@@ -326,6 +326,129 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+    this.x = this.x % x[0];
+    return this;
+  } else {
+    this.x = this.x % xComponent;
+    this.y = this.y % yComponent;
+    return this;
+  }
+};
+
+const calculateRemainder3D = function(xComponent, yComponent, zComponent) {
```
lmccart: is there a reason this takes three arguments instead of one array?

src/math/p5.Vector.js
jatin33: Yes I have covered the edge cases in commit 0b9302d

src/math/p5.Vector.js
```diff
@@ -326,6 +326,138 @@ p5.Vector.prototype.add = function add(x, y, z) {
...
+    this.y = this.y % yComponent;
+    return this;
+  } else if (xComponent !== 0 && yComponent === 0 && zComponent === 0) {
+    this.x = this.x % xComponent;
+    return this;
+  } else {
+    this.x = this.x % xComponent;
+    this.y = this.y % yComponent;
+    this.z = this.z % zComponent;
+    return this;
```
lmccart: could this whole block be rewritten more simply as:
```js
if (xComponent !== 0 ) {
  this.x = this.x % xComponent;
}
if (yComponent !== 0 ) {
  this.y = this.y % yComponent;
}
if (zComponent !== 0 ) {
  this.z = this.z % zComponent;
}
return this;
```

lmccart: and same with 2D method above?

jatin33: made necessary changes bf25493

lmccart:
🎉 thanks for all the revisions of this!

PR #4259, opened 2020-01-18
Adds third argument to position method

### LINKED ISSUE #3570: Additional parameter for position function.
#### Nature of issue?

- [x] Other (specify if possible) DOM

- [x] Others (specify if possible) Device Independent

#### Details about the bug: 

- p5.js version:  0.7.3
- Web browser and version:  Browser independent
- Operating System:  OS independent
- Steps to reproduce this:

Currently , the position function works by providing two numbers x and y, 
and the element on which the function was called , has its position changed to x pixels, and y pixels on the web page, and the position style is set as 'Absolute' 

Current syntax : 
```
some_elt.position(10, 10);
```

**Modification :** 
 
It would be nice to have an extra optional parameter , which will allow us to also set the type of the position we want , like this 
```
some_elt.position(10, 10, ABSOLUTE);
```

```
some_elt.position.(10, 10, FIXED); //always relative to current viewport
```
(The default option would be ABSOLUTE)

The use case of the optional property may arise when the element may be present inside another div,
this issue can be considered as an example ,
processing/p5.js-website#394

Would love to hear your thoughts about it,
Thank you !

jatin33:
Resolves #3570 

 Changes: 

Provided a third optional parameter to position function **positionType**

 Screenshots of the change: 

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -1694,15 +1694,27 @@ p5.Element.prototype.html = function() {
  */
 /**
  * @method position
- * @param  {Number} [x] x-position relative to upper left of window
- * @param  {Number} [y] y-position relative to upper left of window
+ * @param  {Number} [x] x-position relative to upper left of window (optional)
+ * @param  {Number} [y] y-position relative to upper left of window (optional)
+ * @param  {String} positionType it can be static, fixed, relative, sticky, initial or inherit (optional)
  * @chainable
  */
 p5.Element.prototype.position = function() {
   if (arguments.length === 0) {
     return { x: this.elt.offsetLeft, y: this.elt.offsetTop };
   } else {
-    this.elt.style.position = 'absolute';
+    let positionType = '';
+    if (
+      arguments[2] === 'static' ||
+      arguments[2] === 'fixed' ||
+      arguments[2] === 'relative' ||
+      arguments[2] === 'sticky' ||
+      arguments[2] === 'initial' ||
+      arguments[2] === 'inherit'
+    ) {
+      positionType = arguments[2];
+    }
+    this.elt.style.position = positionType || 'absolute';
     this.elt.style.left = arguments[0] + 'px';
     this.elt.style.top = arguments[1] + 'px';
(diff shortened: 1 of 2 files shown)


jatin33:
Hey @lmccart, could you please take a look at the **Test_PR** check, it shows that test coverage is only 14%. I haven't added  **only** keyword to any of the suites yet test coverage has decreased. Could you please suggest why something like this would happen and how to avoid it in the future?

src/dom/dom.js
```diff
@@ -1694,15 +1694,27 @@ p5.Element.prototype.html = function() {
...
+ * @param  {Number} [y] y-position relative to upper left of window (optional)
+ * @param  {String} positionType it can be static, fixed, relative, sticky, initial or inherit (optional)
  * @chainable
  */
 p5.Element.prototype.position = function() {
   if (arguments.length === 0) {
     return { x: this.elt.offsetLeft, y: this.elt.offsetTop };
   } else {
-    this.elt.style.position = 'absolute';
+    let positionType = '';
```
lmccart: you can change this to:
```js
let positionType = 'absolute';
```
and remove the `||` check on line 1717

test/unit/dom/dom.js
```diff
@@ -222,4 +222,16 @@ suite('DOM', function() {
       };
     });
   });
+
+  suite('p5.Element.prototype.position', function() {
+    let paragraph = myp5.createP('out of box');
+    paragraph.position(20, 20, 'static');
+
+    elt = document.createElement('p');
+    elt.appendChild(document.createTextNode('out of box'));
```
lmccart: you can just add this text as the second arg in the line above

src/dom/dom.js
```diff
@@ -1694,15 +1694,27 @@ p5.Element.prototype.html = function() {
  */
```
lmccart: add some explanation of the ability to add position type. right now the description indicates that it will always be position:absolute.
can you also add an example?

PR #4265, opened 2020-01-26
Change the example in asin() and atan() documentation

hellonagi:
because the original example output didn't match the prints

 Changes:
-change the example in asin() and atan() documentation

#### PR Checklist


lmccart:
thanks!

PR #4286, opened 2020-02-15
millis() to be counted per instance

### LINKED ISSUE #4264: millis() behavior
Hi there!

I'm using millis() to manage time on a few sketches. What's happening is that I'm rendering a few sketches and all of them share de same millis() return, even though each one starts at different moments.

I don't know if it is the expected behavior. Accordingly to the reference description of millis(), "the number of milliseconds since starting the program", the program is the p5.js, not the sketch, which makes sense. But I think would be nice to have a millis() value by sketch, ou at least another function with this behavior.

A sample code to show this:

```
const s = ( sketch ) => {

  let x = 100;
  let y = 100;

  sketch.setup = () => {
    sketch.createCanvas(200, 200);
    sketch._mycolor = sketch.color(sketch.random(255));
  };

  sketch.draw = () => {
    sketch.background(0);
    sketch.fill(sketch._mycolor);
    sketch.rect(x,y,50,50);
    sketch.fill(255)
    sketch.text(sketch.millis(), 0,10);
  };
  
  sketch.mousePressed = () => {
    sketch.remove();
    new p5(s);
  }
};

function createSketch () {
  new p5(s);
  setTimeout(createSketch, 2000);
}

createSketch();
```

limzykenneth:
Resolves #4264 

 Changes:

`millis()` will now be counted by each instance of p5 independently starting just before the user's `setup()` function is called.

Also some additional prototype members that were not converted back to regular functions before are now converted.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/main.js
+++ b/src/core/main.js
@@ -201,6 +201,7 @@ class p5 {
       resize: null,
       blur: null
     };
+    this._millisStart = -1;
 
     // States used in the custom random generators
     this._lcg_random_state = null;
@@ -329,6 +330,9 @@ class p5 {
           }
         }
       }
+
+      // Record the time when sketch starts
+      this._millisStart = window.performance.now();
 
       // Short-circuit on this, in case someone used the library in "global"
       // mode earlier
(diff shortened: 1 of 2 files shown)


limzykenneth:
Sorry, had a couple of wrong push. Please ignore those.

lmccart:
this is great, thanks @limzykenneth!

Issue #4284, opened 2020-02-12
Texture on cylinder


matheusb-comp:
When adding an image [`texture()`](https://p5js.org/reference/#/p5/texture) to a cylinder, the image is placed in the cylinder as expected, except for a strange vertical line that looks like the texture image pressed in a very small width.

If the cylinder `detailX` is reduced to 4, it is possible to see that one of the faces have the entire texture image.

Example sketch: https://editor.p5js.org/matheusb-comp/sketches/Rm1CnvPS
![Screenshot from 2020-02-12 00-39-23](https://user-images.githubusercontent.com/1227961/74301184-00c70280-4d31-11ea-8f82-7182f4350463.png)

What is the proper way to apply an image texture to a cylinder, to avoid this vertical line?

aferriss:
I think this is a bug related to the way that the cylinder vertices are constructed. I added a custom shader to visualize the uv's in your example. I believe what's happening is that when the uv's go from 1 back to 0, they get interpolated across the faces of the cylinder rather than snapping immediately from 1 - 0.

I think the fix for this probably requires digging into to the cylinder construction code and figuring out another way to build up the vertices and texture coordinates such that they don't get interpolated. 

![Screen Shot 2020-02-17 at 10 39 40 AM](https://user-images.githubusercontent.com/3698659/74679320-5f95dd00-5172-11ea-8243-5855773b135b.png)

Here's my example. I overlaid another cylinder with it's stroke enabled so that you can see how the triangles are laid out. 
https://editor.p5js.org/aferriss/sketches/ZhRZYjAA

A really hacky temporary fix for @matheusb-comp  is just to increase the subdivisions of the cylinder until the artifact goes away. With enough detail you probably won't be able to notice it

PR #4293, opened 2020-02-17
fix cylinder and cone uv bug

aferriss:
Resolves https://github.com/processing/p5.js/issues/4284

 Changes: Fixed an off by one bug that was preventing uv's on cones and cylinders from being calculated correctly.

#### PR Checklist

- [X] `npm run lint` passes


stalgiag:
Thank you @aferriss !

PR #4322, opened 2020-02-24
Vector mult div

### LINKED ISSUE #4283: Allow mult and div vector math functions to take 2 vectors as arguments

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [X] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

Currently the vector math functions let you add or subtract two vectors together as `p5.Vector.add(v0, v1);` or `p5.Vector.sub(v0, v1)` but for multiplication and division you can only do operations between scalars and vectors `p5.Vector.mult(2, v0)`. 

I think it'd be great if we could add in function signatures for mult and div so that you can do math between two vectors. For example: `p5.Vector.mult(v0, v1)` or `p5.Vector.div(v0, v1)` 

The original mult / div by scalar functionality of these methods would remain the same (non-breaking change).

This would do component wise multiplication or division. Here's another example of how I'm imagining this would work below

```
const v0 = createVector(1, 2);
const v1 = createVector(3, 4);
const v2 = createVector(5, 6);
const multResult = p5.Vector.mult(v0, v1); // (3, 8)
const divResult = p5.Vector.div(v1, v0); // (3, 2)

// Another example 
const anotherMultResult = v2.mult(v1); // (15, 24)
const anotherDivResult = v2.div(v0); // (5, 3)
```

aferriss:
Resolves #4283 

 Changes:
Allows the `p5.Vector.mult()` and `p5.Vector.div()` functions to accept two vectors or optionally to chain a mult / div with another vector.

Examples:
```
// Use the in place method
let a = createVector(4,6);
let b = createVector(2, 3);
a.div(b); // a equals (2, 3)
```
```
// Use the static method
let a = createVector(1, 2);
let b = createVector(2, 3);
let c = p5.Vector.mult(a, b); // c returns (2, 6)
```

#### PR Checklist

- [X] `npm run lint` passes


src/math/p5.Vector.js
```diff
@@ -524,7 +524,8 @@ p5.Vector.prototype.sub = function sub(x, y, z) {
 };
 
 /**
- * Multiply the vector by a scalar. The static version of this method
+ * Multiplies the vector by a scalar, multiplies the x, y, and z components from a vector, or multiplies
```
limzykenneth: Maybe reword this a bit to make the difference between multiplying by a scalar and multiplying by a vector.

src/math/p5.Vector.js
```diff
@@ -584,22 +602,31 @@ p5.Vector.prototype.sub = function sub(x, y, z) {
...
-  if (!(typeof n === 'number' && isFinite(n))) {
-    console.warn(
-      'p5.Vector.prototype.mult:',
-      'n is undefined or not a finite number'
-    );
+p5.Vector.prototype.mult = function mult(x, y, z) {
+  if (x instanceof p5.Vector) {
+    this.x *= x.x || 0;
+    this.y *= x.y || 0;
+    this.z *= x.z || 0;
```
limzykenneth: Why is the type check removed and why does it default to 0 here if falsy? If the value evaluates to falsy (`NaN` probably), it is easier to debug than if it got defaulted to 0.

src/math/p5.Vector.js
```diff
@@ -584,22 +602,31 @@ p5.Vector.prototype.sub = function sub(x, y, z) {
...
     return this;
   }
-  this.x *= n;
-  this.y *= n;
-  this.z *= n;
+  if (x instanceof Array) {
```
limzykenneth: Do we need to cover the case where array is passed in as well?

src/math/p5.Vector.js
```diff
@@ -584,22 +602,31 @@ p5.Vector.prototype.sub = function sub(x, y, z) {
...
+    this.x *= x[0] || 0;
+    this.y *= x[1] || 0;
+    this.z *= x[2] || 0;
+    return this;
+  }
+  const _x = x || 1;
+  const _y = y || x;
+  const _z = z || x;
```
limzykenneth: If it defaults to `x` in case `y` or `z` is not defined while `x` is also undefined, `x` will be `1` while `y` and `z` will be undefined. Also I'm not sure if `x` should default to `1` here if falsy, same reason as above, it makes debugging harder.

src/math/p5.Vector.js
```diff
@@ -658,24 +702,39 @@ p5.Vector.prototype.mult = function mult(n) {
...
-    );
+p5.Vector.prototype.div = function div(x, y, z) {
+  if (x instanceof p5.Vector) {
+    if (x.x === 0 || x.y === 0 || x.z === 0) {
+      console.warn('p5.Vector.prototype.div:', 'divide by 0');
+      return this;
+    }
+    this.x /= x.x || 1;
+    this.y /= x.y || 1;
+    this.z /= x.z || 1;
```
limzykenneth: Same with above regarding default case

src/math/p5.Vector.js
```diff
@@ -658,24 +702,39 @@ p5.Vector.prototype.mult = function mult(n) {
...
+  if (x === 0 || y === 0 || z === 0) {
     console.warn('p5.Vector.prototype.div:', 'divide by 0');
     return this;
   }
-  this.x /= n;
-  this.y /= n;
-  this.z /= n;
+  const _x = x || 1;
+  const _y = y || x;
+  const _z = z || x;
```
limzykenneth: Same here as above.

src/math/p5.Vector.js
aferriss: Sounds good, I'll see what I can do

src/math/p5.Vector.js
aferriss: I was just following the pattern that was already existing in `p5.Vector.add` and `p5.Vector.sub`, neither of which[ do the type checking](https://github.com/processing/p5.js/blob/master/src/math/p5.Vector.js#L310) . I can add the type check back in if you think it's helpful.

src/math/p5.Vector.js
aferriss: Again, also following the pattern laid out in .add and .sub, which both also check if an array was passed in.

src/math/p5.Vector.js
aferriss: In this case, it's probably best to not have the default be 0, since this would cause a divide by zero.

src/math/p5.Vector.js
aferriss: Woops, I think that 1 is supposed to be a 0. 

Maybe I'm not understanding your comment but I think in the case of `x` being defined, and `y` and `z` being undefined, wouldn't they all end up as `x`? I wrote it this way to keep the functionality of being able to provide a single scalar value to multiply against all the vector components. 

```
const val = 2;
const _x =  val || 1; // _x equals 2
const _y = y || val; // y is undefined so _y also equals 2
const _z = z || val; // z is undefined so _z also equals 2
```

One tricky case that I just thought of is what to do in the case that someone provides just both x and y values but doesn't provide a z. 

This also makes me wonder if we should also allow `.add` and `.sub` to be used with a single scalar value like we already did for `.mult and .div`

limzykenneth: My explanation is a bit confusing (it's confusing me as well) so I'll just illustrate in code:

```javascript
const val = undefined; // Not valid but just to illustrate
const _x =  val || 1; // _x equals 1
const _y = y || val; // y is undefined so _y equals val which equals to undefined
const _z = z || val; // z is undefined so _z equals val which equals to undefined
```

Rather should probably be

```javascript
const val = undefined; // Not valid but just to illustrate
const _x =  val || 1; // _x equals 1
const _y = y || _x; // y is undefined so _y equals _x which equals to 1
const _z = z || _x; // z is undefined so _z equals _x which equals to 1
```

And I think default to `1` is actually correct here as if `mult()` is not called with any arguments or called with an undefined argument, the vector should not be modified and in this case is to multiply by 1.

For the case of users providing only `x` and `y` but not `z`, it is technically an undocumented/invalid signature which we don't have to account for. `add()` and `sub()` probably should not be modified to add scalar across all values as that would be a breaking change.

src/math/p5.Vector.js
aferriss: @limzykenneth updated!

src/math/p5.Vector.js
aferriss: @limzykenneth also made some changes to the language here. Let me know if this makes more sense

src/math/p5.Vector.js
aferriss: @limzykenneth I added back in some type checking. it's a bit more 
verbose but I think should cover all the cases (that I could think of)

src/math/p5.Vector.js
aferriss: With type checking, we no longer have this issue of the defaults, since you'll just get back an unchanged vector if something was undefined / nan

limzykenneth:
It's looking good! I'll have a closer look these couple days to see if there's any details left to fix.

limzykenneth:
All's looking good, just one last thing from me. The new signatures need to be updated in teh documentation as well otherwise the parameter validation system may thow warnings with certain valid usage.

I would suggest covering the following cases (for both `mult` and `div`)
```
vector.mult(n); // scalar
vector.mult(x, y);
vector.mult(x, y, z);
vector.mult(v); // vector
vector.mult(arr) // array
```

I don't think there's any other combination but if there is I don't think they should be covered as official API.

aferriss:
@limzykenneth thanks! Good catch, I forgot about the signatures.

The signatures are maybe more complex than they appear because of one of the static methods. 

So in addition to the ones you mentioned we'd also have

```
p5.Vector.mult(v0, v1);
p5.Vector.mult(v0, v1, target);
p5.Vector.mult(v0, arr);
p5.Vector.mult(v0, arr, target);
```

I'm not sure that I understand how those signatures with `target` are supposed to work. 

This seems to work
```
const target = createVector(1,1,1);
const a = createVector(1, 2, 4);
const b = createVector(2, 2, 2);
p5.Vector.div(a, b, target); returns (0.5, 1, 2)
```

but it only works if target was already set to a p5.Vector. If target started out as undefined, it just returns undefined. Is this a bug, or intended behavior?

limzykenneth:
For signatures you can have a look at the existing ones for `add()` where you can combine some with optional arguments, for example:
```javascript
// Instead of
p5.Vector.mult(v0, v1);
p5.Vector.mult(v0, v1, target);

// Just use
p5.Vector.mult(v0, v1, [target]);
```

`target` should work even if it is `undefined`, at least from looking at the source. I'm not immediately sure why it is not working. Might be for another issue though so you don't have to fix so many things in this one PR.

aferriss:
@limzykenneth sounds good, I'll take a look.

I think the reason that `target` is becoming undefined is that when it gets passed in as `undefined` it's getting passed in as a value, instead of as a reference. 

I don't really think there's much we can do about this other than to maybe log a warning that the user hasn't passed in a vector.

limzykenneth:
Cool, as soon as the signatures are updated, I'll approve this PR and let someone else doublecheck before merging. Thanks!

aferriss:
@limzykenneth Ok, added in all the signatures I could think of. The docs page is a little scarier though, but at least it's comprehensive.

Screen Shot 2020-03-06 at 10 58 38 AM

limzykenneth:
Actually there's a few that is redundant, specifically
```javascript
p5.Vector.mult(v, n)
p5.Vector.mult(v0, v1)
p5.Vector.mult(v1, arr)
```
as they are coverd by the optional arguments case already.

aferriss:
@limzykenneth yeah, I included them since they are slightly different in that they would return a vector, vs. modifying something in place. But I'm happy to remove them!

test/unit/math/p5.Vector.js
```diff
@@ -617,9 +617,9 @@ suite('p5.Vector', function() {
 
       test('should not change x, y, z if n is 0', function() {
         v.div(0);
-        expect(v.x).to.eql(1);
-        expect(v.y).to.eql(1);
-        expect(v.z).to.eql(1);
+        expect(v.x).to.eql(v.x);
+        expect(v.y).to.eql(v.y);
+        expect(v.z).to.eql(v.x);
```
limzykenneth: Wouldn't this always be true regardless?

aferriss: Not sure I understand your comment. The test started failing because we changed it so that instead of returning 1 if you divide by zero, we return the original vector.

limzykenneth: If the vector `v` is changed because of dividing by zero, `v.x == v.x`; if it isn't changed because of dividing by zero, still `v.x == v.x`. 

Unless the value of the vector `v` is cached beforehand with `const v2 = v.copy()` and comparing against that cached value, wouldn't any comparison of `v.x` with `v.x` always return true?

aferriss: Oops, duh! Looking at this more closely, that divide by zero test was failing because we weren't catching that case when there was a scalar sent to the function. I updated `div` and added some more tests for divide by zero as well.

limzykenneth:
Looks good. Thanks!

PR #4324, opened 2020-02-25
print() function, even with content, causes print dialog to appear in…

### LINKED ISSUE #4272: print() function, even with content, causes print dialog to appear in firefox

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug: 

- p5.js version: 0.10.2
- Web browser and version: Firefox 72.0.2
- Operating System: MacOS 10.14 (Mojave)
- Steps to reproduce this:

The following code causes the print dialog to appear in Firefox (using [editor.p5js.org](http://editor.p5js.org/)).

```
print("Hello world!")
```

debajitdeb11:
… firefox

fix

Resolves #4272 

 Changes:
 * Fix the issue of print dialog to appear in the browser by using the print() function.
 * Slight changes in the example section

- [ ] `npm run lint` passes

### DIFF
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -15,7 +15,7 @@ p5.prototype._frameRate = 0;
 p5.prototype._lastFrameTime = window.performance.now();
 p5.prototype._targetFrameRate = 60;
 
-const _windowPrint = window.print;
+// const _windowPrint = window.print;
 
 /**
  * The <a href="#/p5/print">print()</a> function writes to the console area of your browser.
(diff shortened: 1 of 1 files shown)


lmccart:
Thanks! print() with content used inside setup and draw works as expected in firefox, and other browsers. So we want to keep the code as is, because we do want to invoke the print dialog if there's no content. The remaining todo on #4272 is to see if we can use the friendly error system to detect a pattern of print("content") outside of setup and draw.

PR #4359, opened 2020-03-10
Prevent multiple init on multiple imports

### LINKED ISSUE #2681: duplicate imports results in dulpicate canvases
p5 v0.6.0

demo: https://codepen.io/Spongman/pen/eVoKqb?editors=1010

based on #2679

if you import the p5.js script twice, you'll get two canvases and presumably a whole bunch of other undefined behavior.

it may be useful to have the p5 library detect this and generate an error.

akshay-99:
Resolves #2681 

 Changes:

- check if p5 has already been initialised globally. 
This is done by checking if a property initialised in the constructor exists in the global object.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/init.js
+++ b/src/core/init.js
@@ -18,9 +18,10 @@ const _globalInit = () => {
   // If that property is already a part of the global object,
   // this code has already run before, likely due to a duplicate import
   if (typeof window._setupDone !== 'undefined') {
-    throw new Error(
+    console.warn(
       'p5.js seems to have been imported multiple times. Please remove the duplicate import'
     );
+    return;
   }
 
   if (!window.mocha) {
--- a/src/core/init.js
+++ b/src/core/init.js
@@ -14,6 +14,15 @@ import { initialize as initTranslator } from './internationalization';
  * @return {Undefined}
  */
 const _globalInit = () => {
+  // Could have been any property defined within the p5 constructor.
+  // If that property is already a part of the global object,
+  // this code has already run before, likely due to a duplicate import
+  if (typeof window._setupDone !== 'undefined') {
+    throw new Error(
+      'p5.js seems to have been imported multiple times. Please remove the duplicate import'
+    );
+  }
+
   if (!window.mocha) {
     // If there is a setup or draw function on the window
     // then instantiate p5 in "global" mode


akshay-99:
@limzykenneth I am sorry for the delay, but please take a look

akshay-99:
Hi @limzykenneth is there something wrong with this PR? If so, I can fix it :sweat_smile:

src/core/init.js
```diff
@@ -14,6 +14,15 @@ import { initialize as initTranslator } from './internationalization';
  * @return {Undefined}
  */
 const _globalInit = () => {
+  // Could have been any property defined within the p5 constructor.
+  // If that property is already a part of the global object,
+  // this code has already run before, likely due to a duplicate import
+  if (typeof window._setupDone !== 'undefined') {
+    throw new Error(
```
limzykenneth: I'm not sure throwing an error is necessary. We just need for it to not initialize in the case that p5.js is already loaded. A warning and an early `return` should be more appropriate I think.

src/core/init.js
```diff
@@ -14,6 +14,15 @@ import { initialize as initTranslator } from './internationalization';
  * @return {Undefined}
  */
 const _globalInit = () => {
+  // Could have been any property defined within the p5 constructor.
+  // If that property is already a part of the global object,
+  // this code has already run before, likely due to a duplicate import
+  if (typeof window._setupDone !== 'undefined') {
```
limzykenneth: Why is it checking for `_setupDone` instead of `p5`? If using instance mode only `p5` will be attached to the window object and nothing else.

akshay-99: @limzykenneth I found that window will have p5 attached to it even before this initialisation code runs. 

Here's what I get when I log `window.p5` at line 19 in the file
```
ƒ p5(sketch, node, sync) {
              var _this = this;
              _classCallCheck(this, p5);
              // PUBLIC p5 PROPERTIES AN…
```
The constructor is already attached to the window object. This is even with single import

src/core/init.js
akshay-99: Okay. But I felt a warning would be apt if we didn't cut the initialisation midway, and that way warn the user about running with dual initialisation. Hence I used an error here. But I am wasn't sure.

src/core/init.js
limzykenneth: Yeah that's right, but with checking anything else that is on the `window` object means that it will only work for global mode and not instance mode.

src/core/init.js
limzykenneth: Either way is fine, I just feel an error is a bit too loud when there is mitigation (stopping second initialization) and so the user doesn't necessarily need to do anything else to continue. Error feels more like something is wrong and it must be fixed before continuing. Just what I feel though.

src/core/init.js
limzykenneth: I'm not sure instance mode is a necessary consideration but it would be great if you can test it out.

akshay-99: @limzykenneth it seems that instance mode is not affected by multiple imports, as it is up to the user to initialise p5 with `new p5(sketch)`. And the users can intentionally create as many instances as they want. When the browser concatenates the scripts to ready them for execution, I think the newer p5 stuff in this case simply replaces the old one with no other side effects.

The trouble in global mode is that the library itself takes the job of doing the `new p5()` initialisation, but with a duplicate import as the library itself is duplicated, this step happens twice. In this case, when the browser concatenates the scripts, `new p5()` is called for a second time for the second import. Hence, the user-defined setup() in sketch.js is called two times, which leads to the two canvases and other weird stuff.

Though yes, the method in this PR will not warn a user about multiple imports if instance mode is used. But as multiple imports here don't seem to break anything, is a warning necessary?

akshay-99: Moreover, I think for instance mode, the state of the global variable will be the same for single or multiple imports so I am not really sure how to catch it.

limzykenneth: If it doesn't affect the behaviour of instance mode then I guess that's fine. Thanks.

src/core/init.js
akshay-99: Should I change this to a warning and an early return then?

akshay-99:
@limzykenneth I have changed the error to a warning instead

akshay-99:
@limzykenneth Please take a look again and tell me if something more needs to be changed.

PR #4365, opened 2020-03-12
Adding template to inline_documentation

### LINKED ISSUE #4364: Inconsistent formats for specifying inline documentation

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Other (Documentation)

The inline documentation which generates the reference documentation does have guidelines for specifying different blocks of documentation but lacks an overall ordered template, due to which in different methods, it creates confusion for newbie developers.

fenilgandhi:
Resolves #4364 

 Changes:
- As the first step, i have added a template which can be used while creating new methods for inline documentation. I have also updated the `inline_documentation.md` file with an image of an ideal example.
- If it is approved, i would convert the existing code to be consistent with the template, module by module.

 Screenshots of the change:
Screenshot from 2020-03-12 01-21-23

#### PR Checklist
- [x] `npm run lint` passes


fenilgandhi:
@lmccart , I have added the template according to your comments. If you approve, i would continue with converting modules to this template.

contributor_docs/inline_documentation.md
```diff
@@ -193,18 +193,21 @@ arc(50, 55, 80, 80, PI+QUARTER_PI, TWO_PI);
...
-You can have multiple examples for one function, just make sure each has it's own `<div>` wrapping. 
+You can have multiple examples for one function, just make sure each example has it's own `@example` block seperated by a single empty line.
+
 ```
 @example
 <div>
 <code>arc(50, 50, 80, 80, 0, PI+QUARTER_PI, OPEN);</code>
 </div>
+
+@example
```
lmccart: there should be only one @example tag, then the html code for as many examples as you'd like

contributor_docs/method.example.js
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @example
+ * <div>
+ * <code>
+ * // "Comment explaining the example."
+ * "Your code goes here."
+ * </code>
+ * </div>
+ *
+ * @example
```
lmccart: there should be only one example tag, so this line can be removed

contributor_docs/method.example.js
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @method "methodName"
+ * @param {"dataType"} "paramName" "Description of the param"
+ * @param {"dataType"} "paramName" "Description of the param"
+ * @chainable
+ */
+p5.prototype.methodName = function() {
+  // Comment explaining the code.
+  'Your code goes here';
+};
```
lmccart: what do you think about adding an additional example to this file that shows everything filled in? I think this might help to interpret the template

contributor_docs/method.example.js
```diff
@@ -0,0 +1,45 @@
...
+ * "Your code goes here"
+ * </code>
+ * </div>
+ *
+ * @alt
+ * "A single line precisely describing the first example"
+ * "A single line precisely describing the second example"
+ */
+
+/**
```
lmccart: wondering if we need a comment here to indicate that this additional signature is used only if needed?

lmccart:
@fenilgandhi looking great! added a few comments inline

contributor_docs/inline_documentation.md
fenilgandhi: Ok, i will remove it.

contributor_docs/method.example.js
fenilgandhi: Ok

contributor_docs/method.example.js
fenilgandhi: Yes, that sounds good. For reference, i have already added a filled copy of the function as an image in the documentation. So i would add it, in the file also.

contributor_docs/method.example.js
fenilgandhi: I think yes, having it written down would make it more clear. I am adding a comment over here, specifying that it is optional.

contributor_docs/inline_documentation.md
```diff
@@ -73,7 +73,7 @@ If the parameter is optional, add square brackets around the name:
 @param {type} [name] Description here.
 ```
 
-If the parameter takes one or more values defined in `constants.js`,
+If the parameter takes one or more values defined in `constants.js` [constants.js](https://github.com/processing/p5.js/blob/master/src/core/constants.js) ,
```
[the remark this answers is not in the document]
fenilgandhi: Ok, i  didn't knew that. I have updated it.

contributor_docs/inline_documentation.md
```diff
@@ -226,12 +229,10 @@ function setup() {
 
 If you need to link to external asset files, put them in [/docs/yuidoc-p5-theme/assets](https://github.com/processing/p5.js/tree/master/docs/yuidoc-p5-theme/assets) and then link to them with "assets/filename.ext" in the code. See the [tint example](http://p5js.org/reference/#/p5/tint).
 
-On some code editors copy-pasting examples in the inline documentation may be awkward and require a lot of adjustments (tabs, asterisks, spaces). You can use this simple tool to beautify the code (and re-indent it to 2 spaces) and make it ready for a copy-paste: 
-[p5examplesFormatter](http://molleindustria.org/p5examplesFormatter/)
-| [download as zip](http://molleindustria.org/p5examplesFormatter/p5examplesFormatter.zip)
+On some code editors copy-pasting examples in the inline documentation may be awkward and require a lot of adjustments (tabs, asterisks, spaces).
```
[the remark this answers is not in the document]
fenilgandhi: I only removed the link because it is no longer alive. It just created confusion. But kept the paragraph as a "watch-out" note for a potential problem.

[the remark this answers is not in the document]
limzykenneth: It's better to have some sort of advice to mitigate the problem of "copy-pasting examples in the inline documentation may be awkward and require a lot of adjustments" otherwise it could create confusion about what people writing examples should do about this problem.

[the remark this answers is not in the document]
fenilgandhi: Do we have any another solution to this problem ? Otherwise i am fine with completely removing it. Shall i do it ?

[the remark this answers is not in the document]
fenilgandhi: Yes,i agree some solution would be nice. But i am not sure what that solution can be. Can we get a copy of the beautifier tool, so that we can simply host it on the repo and just update the link ?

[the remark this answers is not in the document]
limzykenneth: I think we can remove this paragraph as the template should give a good enough idea about what to do in terms of formatting them. Adding the asterisks is a bit troublesome but not too much so and a modern text editor should be able to handle that well enough.

I'm not sure what the beautifier tool is, never used it and this really is the first time I've heard of it.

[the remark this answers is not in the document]
lmccart: let's remove this whole paragraph

fenilgandhi:
@lmccart , @limzykenneth Thanks for your reviews. @lmccart , As the template is pretty much approved, For my next task, i am converting existing examples to this template, module by module.

lmccart:
looks great, thanks @fenilgandhi!

fenilgandhi:
Thanks @lmccart.

PR #4375, opened 2020-03-14
Fix suite structure for p5.Element.prototype.position

### LINKED ISSUE #4374: Format of the tests is wrong for suite p5.Element.prototype.position
As discussed in #4339 

> Also, kinda unrelated to this PR, but the testing framework seems to think that the downloading animated gifs lies in the `p5.Element.prototype.position` suite.

 ```
DOM
    p5.Element.prototype.position
      downloading animated gifs
        p5.prototype.saveGif
          ✓ should be a function
          ✓ should not throw an error
```
> I think it's because of a format issue [here](https://github.com/processing/p5.js/blob/master/test/unit/dom/dom.js#L226), where the `expect` must be enclosed in a test inside the suite.

Or as found [here](https://github.com/processing/p5.js/runs/507291467#step:5:5032),  based on the level of indentation, the testing framework seems to think that` p5.Element.prototype.position` is the parent suite for all subsequent suites.

I don't think it has any other side effects, but felt that it should be fixed nonetheless

akshay-99:
Resolves #4374 

 Changes:
As described in the issue, change the format of suite `p5.Element.prototype.position` to prevent the framework from considering all subsequent suites as being a part of this suite. 
The issue was evident when a subsequent suite, `downloading animated gifs`, was tested using a .only() and the output caused a bit of confusion.

The difference can be observed from the indentation of the very next suite in [this PR](https://github.com/processing/p5.js/pull/4375/checks?check_run_id=507746719#step:5:5045) and on [master](https://github.com/processing/p5.js/runs/507291467#step:5:5033)

This is a minor change I think, as the issue did not seem to have any other side effects
#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/test/unit/dom/dom.js
+++ b/test/unit/dom/dom.js
@@ -224,14 +224,40 @@ suite('DOM', function() {
   });
 
   suite('p5.Element.prototype.position', function() {
-    let paragraph = myp5.createP('out of box');
-    paragraph.position(20, 20, 'static');
+    var myp5;
 
-    elt = document.createElement('p', 'out of box');
-    elt.style.position = 'static';
-    elt.style.left = '20px';
-    elt.style.top = '20px';
-    expect(JSON.stringify(paragraph.elt)).to.eql(JSON.stringify(elt));
+    setup(function(done) {
+      new p5(function(p) {
+        p.setup = function() {
+          myp5 = p;
+          done();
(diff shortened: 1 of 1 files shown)


stalgiag:
Thank you @akshay-99 !

PR #4378, opened 2020-03-15
FES for p5.Vector.add/sub/mult/div/lerp - #4348

### LINKED ISSUE #4348: p5.vector math signatures with the target argument return undefined when passed values

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [X] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug: 

- p5.js version: Master
- Web browser and version: All major browsers
- Operating System: OSX
- Steps to reproduce this:

I noticed that the vector signatures that include a `target` argument only work correctly when a `p5.Vector` is passed in. 

For example this works:

```javascript

let target = new p5.Vector(0, 0, 0);
let v0 = new p5.Vector(1, 1, 1);
let v1 = new p5.Vector(2, 2, 2);

p5.Vector.add(v0, v1, target); 
// target is (3, 3, 3)
``` 

This does not:

```javascript

let target;
let v0 = new p5.Vector(1, 1, 1);
let v1 = new p5.Vector(2, 2, 2);

p5.Vector.add(v0, v1, target); 
// target is undefined
``` 

I believe this is happening because when target is `undefined` it gets treated as a value, and so it never gets changed. When a p5.Vector get's passed in, it gets treated as a reference, and the function works as expected.

Is the signature with `target` even necessary? You can accomplish the same thing by using the return of the vector operation
```javascript
let v0 = new p5.Vector(1, 1, 1);
let v1 = new p5.Vector(2, 2, 2);

let target = p5.Vector.add(v0, v1); // target is (3, 3, 3)
```

The only other solution I can think of here is to log a warning to alert the user that the value they've passed in is not a `p5.Vector`

DivyamAhuja:
Resolves #4348 

 Changes:

added `p5._friendlyError()` call for p5.Vector.add/sub/mult/div/lerp methods in case user provides target parameter but undefined.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -1989,13 +1989,7 @@ p5.Vector.random3D = function random3D() {
  * @static
  * @param  {p5.Vector} v1 a <a href="#/p5.Vector">p5.Vector</a> to add
  * @param  {p5.Vector} v2 a <a href="#/p5.Vector">p5.Vector</a> to add
- * @param  {p5.Vector} target the vector to receive the result
- */
-/**
- * @method add
- * @static
- * @param  {p5.Vector} v1
- * @param  {p5.Vector} v2
+ * @param  {p5.Vector} [target] the vector to receive the result
  * @return {p5.Vector} the resulting <a href="#/p5.Vector">p5.Vector</a>
  *
  */
@@ -2003,6 +1997,12 @@ p5.Vector.random3D = function random3D() {
 p5.Vector.add = function add(v1, v2, target) {
   if (!target) {
     target = v1.copy();
+    if (arguments.length === 3) {
+      p5._friendlyError(
+        'The target parameter is undefined, it should be of type p5.Vector',
+        ''
+      );
+    }
   } else {
     target.set(v1);
   }
@@ -2042,19 +2042,19 @@ p5.Vector.rem = function rem(v1, v2) {
  * @static
  * @param  {p5.Vector} v1 a <a href="#/p5.Vector">p5.Vector</a> to subtract from
  * @param  {p5.Vector} v2 a <a href="#/p5.Vector">p5.Vector</a> to subtract
- * @param  {p5.Vector} target if undefined a new vector will be created
- */
-/**
- * @method sub
- * @static
- * @param  {p5.Vector} v1
- * @param  {p5.Vector} v2
+ * @param  {p5.Vector} [target] if undefined a new vector will be created
  * @return {p5.Vector} the resulting <a href="#/p5.Vector">p5.Vector</a>
  */
 
 p5.Vector.sub = function sub(v1, v2, target) {
   if (!target) {
     target = v1.copy();
+    if (arguments.length === 3) {
+      p5._friendlyError(
+        'The target parameter is undefined, it should be of type p5.Vector',
+        ''
+      );
+    }
   } else {
     target.set(v1);
   }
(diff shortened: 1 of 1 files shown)


src/math/p5.Vector.js
```diff
@@ -1989,20 +1989,20 @@ p5.Vector.random3D = function random3D() {
  * @static
  * @param  {p5.Vector} v1 a <a href="#/p5.Vector">p5.Vector</a> to add
  * @param  {p5.Vector} v2 a <a href="#/p5.Vector">p5.Vector</a> to add
- * @param  {p5.Vector} target the vector to receive the result
- */
-/**
- * @method add
- * @static
- * @param  {p5.Vector} v1
- * @param  {p5.Vector} v2
```
lmccart: why was this signature removed?

src/math/p5.Vector.js
```diff
@@ -2042,19 +2042,19 @@ p5.Vector.rem = function rem(v1, v2) {
  * @static
  * @param  {p5.Vector} v1 a <a href="#/p5.Vector">p5.Vector</a> to subtract from
  * @param  {p5.Vector} v2 a <a href="#/p5.Vector">p5.Vector</a> to subtract
- * @param  {p5.Vector} target if undefined a new vector will be created
- */
-/**
- * @method sub
- * @static
- * @param  {p5.Vector} v1
- * @param  {p5.Vector} v2
```
lmccart: why was this removed?

src/math/p5.Vector.js
```diff
@@ -2207,19 +2220,18 @@ p5.Vector.dist = function dist(v1, v2) {
...
  * @param {Number} amt
- * @param {p5.Vector} target if undefined a new vector will be created
- */
-/**
- * @method lerp
- * @static
- * @param {p5.Vector} v1
- * @param {p5.Vector} v2
- * @param {Number} amt
- * @return {Number}      the lerped value
```
lmccart: why was this removed?

src/math/p5.Vector.js
DivyamAhuja: I remove all of these to make target value [target] (optional) as the functions are always returning resultant vector

src/math/p5.Vector.js
DivyamAhuja: I remove all of these to make target value [target] (optional) as the functions are always returning resultant vector

src/math/p5.Vector.js
DivyamAhuja: @lmccart I remove all of these to make target value [target] (optional) as the functions are always returning resultant vector

DivyamAhuja: Should I revert these changes?

src/math/p5.Vector.js
```diff
@@ -2042,19 +2042,19 @@ p5.Vector.rem = function rem(v1, v2) {
...
  * @param  {p5.Vector} v1 a <a href="#/p5.Vector">p5.Vector</a> to subtract from
  * @param  {p5.Vector} v2 a <a href="#/p5.Vector">p5.Vector</a> to subtract
- * @param  {p5.Vector} target if undefined a new vector will be created
- */
-/**
- * @method sub
- * @static
- * @param  {p5.Vector} v1
- * @param  {p5.Vector} v2
+ * @param  {p5.Vector} [target] if undefined a new vector will be created
```
limzykenneth: New vector won't be created if passed `undefined` so this description needs to change as well.

DivyamAhuja: am still confused about what appropriate description should be for this target parameter....like if target(p5.Vector) is passed it's value will be set to result of function and if it's not passed or undefined it will still return resultant vector..

Any Suggestions about what should one-liner description should be🤔

limzykenneth: The one for `add` just above is fine. If the function signature says it will return then there's no need to say it again here.

limzykenneth:
No issue from me. @lmccart Do you want to have a look over and merge if all is fine?

Issue #4368, opened 2020-03-13
Need to update inline documentation of different modules to be consistent with templates


fenilgandhi:
Since we now have a template for inline documentation across modules needs to be updated to be consistent with the template.

pcgamer1:
@all-contributors please add @pcgamer1 for code

linnhallonqvist:
I would like to try and help with this issue! I will start updating the Typography module.

Qianqianye:
Closing this since we just updated all inline documentation recently.

PR #4383, opened 2020-03-16
[feat] Updated shapes module according to template

fenilgandhi:
To resolve the issue #4368, I am updating the codebase, modules by modules. This PR covers the  [`shape`](https://github.com/processing/p5.js/tree/master/src/core/shape) module.

- Updated inline documentation of shape module to be consistent with template.
- Reworded method descriptions to be simple, consise and clear.
- Refactored examples to be simple and begineer friendly.

 Changes:

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/shape/2d_primitives.js
+++ b/src/core/shape/2d_primitives.js
@@ -98,17 +98,19 @@ p5.prototype._normalizeArcAngles = (
 };
 
 /**
- * Draw an arc to the screen. If called with only x, y, w, h, start, and
- * stop, the arc will be drawn and filled as an open pie segment. If a mode parameter is provided, the arc
- * will be filled like an open semi-circle (OPEN) , a closed semi-circle (CHORD), or as a closed pie segment (PIE). The
- * origin may be changed with the <a href="#/p5/ellipseMode">ellipseMode()</a> function.<br><br>
- * The arc is always drawn clockwise from wherever start falls to wherever stop falls on the ellipse.
- * Adding or subtracting TWO_PI to either angle does not change where they fall.
- * If both start and stop fall at the same place, a full ellipse will be drawn. Be aware that the the
- * y-axis increases in the downward direction, therefore angles are measured clockwise from the positive
+ * Draw an arc to the screen. If called with only x, y, w, h, start and stop,
+ * the arc will be drawn and filled as an open pie segment. If a mode parameter
+ * is provided, the arc will be filled like an open semi-circle (OPEN), a closed
+ * semi-circle (CHORD), or as a closed pie segment (PIE). The origin may be changed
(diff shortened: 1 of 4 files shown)


src/core/shape/2d_primitives.js
```diff
@@ -518,7 +522,7 @@ p5.prototype.quad = function(...args) {
  * @param  {Number} x  x-coordinate of the rectangle.
  * @param  {Number} y  y-coordinate of the rectangle.
  * @param  {Number} w  width of the rectangle.
- * @param  {Number} [h]  height of the rectangle.
+ * @param  {Number} h  height of the rectangle.
```
fenilgandhi: Hey @lmccart , i need some advice over here. I tested the `rect` function and on not providing the 4th argument, the rectangle is not rendered on the canvas and a error messages pops up in the console. But here it was documented as optional, so i changed it to match the current behaviour of the code. Now the tests are failing due to this change. Should i revert back this change or update the test ? I am not sure what i am missing.

fenilgandhi: I think reverting this commit should handle all of the changes. https://github.com/processing/p5.js/commit/98241d6c02c67cacfbe09c8cccfd1bc9356b3cdc#diff-27b6dc80741d2191e0ca1b5030cf73e1

limzykenneth: The code should probably stay the same, it is a documented feature (passing three arguments) and changing it would be considered a breaking change. Maybe look into why FES is not allowing the 3 argument version, maybe the signature isn't registered correctly or something.

DivyamAhuja: Well where did ya test rect() function with 3 arguments? If on web editor then it uses 0.10.2 version of p5.js by default and it works perfectly fine with version 1.0.0 of p5.js . so I think changes for rect() to take 3 arguments were made after release of 0.10.2 version , and maybe that's the reason. After reading source code and testing it on version 1.0.0 I didn't found any visible errors so I think you should revert it back to [h]

Check this example out,
https://editor.p5js.org/DivyamAhuja/sketches/z_PtePeH

DivyamAhuja: @limzykenneth he updated signature, changing h from optional to required so now unit test case with 3 input is expecting 4 input parameters and is throwing error

fenilgandhi: OK, got the error. Yes, i had tested it on the web editor. @DivyamAhuja, you are correct. I will push the necessary changes as well. Thanks for your help @DivyamAhuja , @limzykenneth.

lmccart:
this is all super helpful, thank you!

fenilgandhi:
Glad to help. Thanks.

PR #4395, opened 2020-03-20
[fix] Updated `createSelect` to add DOM Element Support

### LINKED ISSUE #4394: createSelect does not handle existing dom elements

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Other (dom)

#### Details about the bug: 

- p5.js version: 1.0.0
- Web browser and version: Chrome 80.0.3987.122
- Operating System: Manjaro 19.0.2 Kyria (Linux 4.19.108-1-MANJARO)
- Steps to reproduce this:

As per the documentation, [`createSelect`](https://p5js.org/reference/#/p5/createSelect) should be able to take an existing DOM select element and return a `p5.Element` of select type. But the current implementation does not handle such a case and throws an Error. It only handles either a `p5.Element` or creates a new one. Is it documentated wrong or implemented wrong ?

fenilgandhi:
Resolves #4394 

 Changes:
I am sure as what to look up to, as a source of truth but according to my understanding, i have done the following changes.
##### Documentation
- specified methods on seperate lines
- updated their description to match be more explanatory.
- split descriptions of method behaving differently on different parameters.

##### Implementation
- Replaced `var` with `let`.
- Added handling of DOM select, it will wrap up the DOM select in a `p5.Element` and return it.
- Handled cases for validating parameter in methods.

Once the changes in `createSelect` are finalised, i will update the unit tests also.

#### PR Checklist
- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -614,11 +614,13 @@ p5.prototype.createCheckbox = function() {
 /**
  * Creates a dropdown menu &lt;select&gt;&lt;/select&gt; element in the DOM.
  * It also helps to assign select-box methods to <a href="#/p5.Element">p5.Element</a> when selecting existing select box.
- * The .option() method can be used to set options for the select after it is created.
- * The .value() method will return the currently selected option.
- * The .selected() method will return current dropdown element which is an instance of <a href="#/p5.Element">p5.Element</a>
- * The .selected() method can be used to make given option selected by default when the page first loads.
- * The .disable() method marks given option as disabled and marks whole of dropdown element as disabled when invoked with no parameter.
+ * - `.option(name, [value])` can be used to set options for the select after it is created.
+ * - `.value()` will return the currently selected option.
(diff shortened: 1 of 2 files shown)


src/dom/dom.js
```diff
@@ -737,16 +743,18 @@ p5.prototype.createSelect = function() {
...
+        if (this.elt[i].value.toString() === value) {
           this.elt[i].disabled = true;
         }
       }
       return this;
-    } else if (arguments.length === 0) {
-      this.elt.disabled = true;
-      return this;
+    } else {
+      // TODO : Need a default behaviour here.
```
fenilgandhi: What should be done here ?

lmccart: I think we should switch this to remove the undefined check, first check if it's a string, then put the lines 747, 748 under "else"

fenilgandhi: Ok thanks.

lmccart:
looks great, thank you!

PR #4411, opened 2020-03-25
[fix] Splitting methods in dom module

### LINKED ISSUE #4410: Splitting up methods as individual functions in favour of readability and testing
In dom module, [methods CreateDiv, CreateP, CreateSpan](https://github.com/processing/p5.js/blob/f176e7029df9c5f0b34709a378a5e897039927ec/src/dom/dom.js#L376-L400) are bundled up which makes it hard to discover them. It is not good for testing and breaks pattern of how methods should be written. So in favor of ease of readablility and testing, i would like to write them individually.

fenilgandhi:
Resolves #4410 

 Changes:
Split up methods for  createDiv, createP, createSpan

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -348,6 +348,11 @@ function addElement(elt, pInst, media) {
  * createDiv('this is some text');
  * </code></div>
  */
+p5.prototype.createDiv = function(html = '') {
+  let elt = document.createElement('div');
+  elt.innerHTML = html;
+  return addElement(elt, this);
+};
 
 /**
  * Creates a &lt;p&gt;&lt;/p&gt; element in the DOM with given inner HTML. Used
@@ -361,6 +366,11 @@ function addElement(elt, pInst, media) {
  * createP('this is some text');
  * </code></div>
  */
+p5.prototype.createP = function(html = '') {
+  let elt = document.createElement('p');
+  elt.innerHTML = html;
+  return addElement(elt, this);
+};
 
 /**
  * Creates a &lt;span&gt;&lt;/span&gt; element in the DOM with given inner HTML.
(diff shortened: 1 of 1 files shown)


src/dom/dom.js
```diff
@@ -348,6 +348,11 @@ function addElement(elt, pInst, media) {
  * createDiv('this is some text');
  * </code></div>
  */
+p5.prototype.createDiv = function(html) {
+  let elt = document.createElement('div');
+  if (typeof html !== undefined) elt.innerHTML = html;
```
akshay-99: Use a string here. 
`typeof html !== 'undefined'`
`typeof something` gives the type of something but as a string.
This will remove the test failure

src/dom/dom.js
```diff
@@ -361,6 +366,11 @@ function addElement(elt, pInst, media) {
  * createP('this is some text');
  * </code></div>
  */
+p5.prototype.createP = function(html) {
+  let elt = document.createElement('p');
+  if (typeof html !== undefined) elt.innerHTML = html;
```
akshay-99: here too

src/dom/dom.js
```diff
@@ -373,15 +383,11 @@ function addElement(elt, pInst, media) {
...
-  var method = 'create' + tag.charAt(0).toUpperCase() + tag.slice(1);
-  p5.prototype[method] = function(html) {
-    var elt = document.createElement(tag);
-    elt.innerHTML = typeof html === 'undefined' ? '' : html;
-    return addElement(elt, this);
-  };
-});
+p5.prototype.createSpan = function(html) {
+  let elt = document.createElement('span');
+  if (typeof html !== undefined) elt.innerHTML = html;
```
akshay-99: and here

src/dom/dom.js
fenilgandhi: yes, that would solve this. but still the tests fail. I m looking into it.

fenilgandhi: found the issue, this tests are not updated. there is another commit which enhances the tests. Need to find a middle out

Issue #4392, opened 2020-03-18
Enhancing quality of existing tests and adding missing unit tests.


fenilgandhi:
#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Other (Tests)

#### Feature enhancement details:
The test coverage for modules like `dom`, `image` and `io` is pretty low. Also some of the existing test can be improved. 
For more reference, check out #4385, #3567

limzykenneth:
I'm not too sure but wouldn't this issue be covered by #4385 already? Or is this about something else unrelated?

fenilgandhi:
#4385 is a more generalised issue for discussing about the condition of testing as a whole. Whilst i would only be cover 2 or 3 modules and try to improve the quality of existing test and add missing ones. Also since #4385 is used by @akshay-99 to discuss about his GSOC proposal, i thought it would be better to have a seperate issue.

fenilgandhi:
Just for ease of review and merging, i am going to break this work in bundled PRs each with changes of 7-8 methods.

### DOM TESTING

##### NEW PR : create Methods
  - createImg
  - createRadio
  - createColorPicker
  - createInput
  - createFileInput
  - createMedia
  - createVideo
  - createAudio
  - navigator.mediaDevices
  - navigator.mediaDevices.getUserMedia
  - createCapture
  - createElement

##### NEW PR : p5.Element methods I
  - addClass
  - removeClass
  - hasClass
  - toggleClass
  - child
  - center
  - html
  - position
  - _translate

##### NEW PR : p5.Element methods II
  - style
  - attribute
  - removeAttribute
  - value
  - show
  - hide
  - size
  - remove
  - drop

##### NEW PR: p5.MediaElements I
  - constructor
  - play
  - stop
  - pause
  - loop
  - noLoop
  - autoplay
  - volume
  - speed

##### NEW PR: p5.MediaElements II
  - time
  - duration
  - loadPixels
  - updatePixels
  - get
  - set
  - copy
  - mask

##### NEW PR: p5.MediaElements III
  - isModified
  - setModified
  - onended
  - connect
  - disconnect
  - showControls
  - hideControls

##### NEW PR: Cue and File
  Cue
  - addCue
  - removeCue
  - clearCues

  p5.File

PR #4414, opened 2020-03-25
[fix] Fixed bug in dom.removeElements

### LINKED ISSUE #4413: removeElements does not remove all the elements from the DOM

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] DOM
- [ ] Other ()

#### Details about the bug: 

- p5.js version: 1.0.0
- Web browser and version:Chrome 80.0.3987.122
- Operating System: Manjaro 19.0.2 Kyria (Linux 4.19.108-1-MANJARO)

- Steps to reproduce this:
[`removeElements`](https://github.com/processing/p5.js/blob/f176e7029df9c5f0b34709a378a5e897039927ec/src/dom/dom.js#L199-L244) does not remove all the remaining elements. Only a few of the elements at random are removed.

fenilgandhi:
Resolves #4413 
Addresses #4392

 Changes: 
- Updated removeElements to remove all non-canvas elements.
- Updates tests for it.

The current implementation loops through `this._elements` with an index i and once a element which is not a canvas is found, it calls `.remove` on the element. But the catch is that `.remove` is implemented in such a way that it splices `this._elements` and thus `this._elements.length` is obsolete and index `i` is not pointing to the expected element. I tried different approaches to solve this and found the functional approach to be the most clear and efficient, so i replaced the iteration with functional approach.

#### PR Checklist
- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -216,11 +216,10 @@ p5.prototype._wrapElement = function(elt) {
  */
 p5.prototype.removeElements = function(e) {
   p5._validateParameters('removeElements', arguments);
-  for (var i = 0; i < this._elements.length; i++) {
-    if (!(this._elements[i].elt instanceof HTMLCanvasElement)) {
-      this._elements[i].remove();
-    }
-  }
+  // el.remove splices from this._elements, so don't mix iteration with it
+  const isNotCanvasElement = el => !(el.elt instanceof HTMLCanvasElement);
+  const removeableElements = this._elements.filter(isNotCanvasElement);
+  removeableElements.map(el => el.remove());
 };
 
 /**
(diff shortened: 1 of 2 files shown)


lmccart:
great catch!

fenilgandhi:
Thank you, @lmccart. Happy to learn.

bassamanator:
@fenilgandhi Sorry but I don't think the removeElements() function is fixed, at least it isn't for me. When I use it, it still leaves createInput() and createButton() elements on the page. I am using /*! p5.js v1.0.0 February 29, 2020 */

fenilgandhi:
Hey @bassamanator, this changes were applied on 25th March, so i guess you will have to wait for the next release to get these updates.

PR #4418, opened 2020-03-25
A way to verify downloads

akshay-99:
Changes:

Until now there was no way to verify functions that download something to the user's system. I went through the ['file-saver' module](https://github.com/eligrey/FileSaver.js/blob/master/src/FileSaver.js ) which is used in p5 to facilitate downloads and found that it uses `createObjectURL()` function which we can temporarily override in tests to capture the file that is just about to be downloaded. This makes unit testing for download functions possible, without the need to restructure the src function itself.

I have created a helper function in the tests `testWithDownload` which uses the above approach.

I added another helper function: `testUnMinified` , which when used, only runs that test on the unminified library. It is meant to be used for testing invalid arguments on the function.

I felt this was needed because `assert.validationError` is currently defined as, 
```js
assert.validationError = function(fn) {
  if (p5.ValidationError) {
    assert.throws(fn, p5.ValidationError);
  } else {
    assert.doesNotThrow(fn, Error, 'got unwanted exception');
  }
};
```
It would expect **no error** to be thrown when running for the minified version. But some p5 functions will naturally throw an error for wrong arguments when they try to perform operations which the passed arguments simply do not support. 

`testUnMinified` would make things much simpler.

This PR adds tests for:
- saveGif
- saveCanvas
- saveFrames
- saveStrings
- saveJSON
- writeFile
- downloadFile
- save
- saveTable

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/test/js/p5_helpers.js
+++ b/test/js/p5_helpers.js
@@ -24,6 +24,53 @@ function testSketchWithPromise(name, sketch_fn) {
   return test(name, test_fn);
 }
 
+function testWithDownload(name, fn, asyncFn = false) {
+  var test_fn = function(done) {
+    // description of this is also on
+    // https://github.com/processing/p5.js/pull/4418/
+
+    let blobContainer = {};
+
+    // create a backup of createObjectURL
+    let couBackup = window.URL.createObjectURL;
+
+    // file-saver uses createObjectURL as an intermediate step. If we
+    // modify the definition a just a little bit we can capture whenever
+    // it is called and also peek in the data that was passed to it
+    window.URL.createObjectURL = blob => {
+      blobContainer.blob = blob;
+      return couBackup(blob);
+    };
+
+    let error;
+    if (asyncFn) {
+      fn(blobContainer)
+        .then(() => {
+          window.URL.createObjectURL = couBackup;
+        })
+        .catch(err => {
+          error = err;
+        })
+        .finally(() => {
+          // restore createObjectURL to the original one
+          window.URL.createObjectURL = couBackup;
+          error ? done(error) : done();
+        });
+    } else {
+      try {
+        fn(blobContainer);
+      } catch (err) {
+        error = err;
+      }
+      // restore createObjectURL to the original one
+      window.URL.createObjectURL = couBackup;
+      error ? done(error) : done();
+    }
+  };
+
+  return test(name, test_fn);
+}
+
 function parallelSketches(sketch_fns) {
   var setupPromises = [];
   var resultPromises = [];
--- a/test/js/p5_helpers.js
+++ b/test/js/p5_helpers.js
@@ -71,6 +71,11 @@ function testWithDownload(name, fn, asyncFn = false) {
   return test(name, test_fn);
 }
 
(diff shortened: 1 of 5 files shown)


akshay-99:
As an example, I have tested for the filetype of the file that would be downloaded. This can be further extended to verify the contents as well

limzykenneth:
Sounds like a good idea to me.

akshay-99:
@limzykenneth Thanks. Sorry for the late reply I was caught up drafting the proposal. So I will update this PR to add tests for all download functions before I mark ready for review?

limzykenneth:
@akshay-99 You can mark the PR as ready for review whenever you feel it is ready for review, ideally with tests but if not just make sure to make a note of it. If you need help working on this PR, just ask here and we'll see if we can help.

fenilgandhi:
Yes, that is the way to test methods like this, even i have used similar method in some tests of `dom` module. Also, just a good habit, keep the habit of using `assert.deepEqual` instead of `assert.strictEqual`.

akshay-99:
> Also, just a good habit, keep the habit of using assert.deepEqual instead of assert.strictEqual as strict equality is shallow.

Yes but wouldn't using it everywhere be an overkill? The rule I follow is strictEqual for everything I can compare with === (Number, string, etc) and deepEqual for everything else (objects and stuff)

I could use deepEqual everywhere but it just wouldn't feel right to me :see_no_evil: Also I guess it would be a little bit slower.

akshay-99:
@limzykenneth So far I have added tests for `saveGif`, `saveStrings`, `saveJSON`, writeFile, `downloadFile` and `saveTable`. I think only `p5.prototype.save` remains, which I will do tomorrow. I have created a helper `testWithDownload` so that I don't have to repeat the overriding code.

Moreover I realised that since tests that use `assert.validationError` only make sense for the un-minified library where validation of parameters is enabled, some of these might fail when running on the minified library. If I pass incorrect arguments to a function with the un-minified script, I would expect a `ValidationError`. But in the minified script, since validateParameters is not enabled, the function could crash throwing some other exception due to the wrong arguments. It just so happens that the functions on which tests `assert.validationError` was used until now do not crash for wrong arguments.
For example, `p5.prototype.arc` is a function that won't crash for the wrong arguments' tests given [here](https://github.com/processing/p5.js/blob/master/test/unit/core/2d_primitives.js#L36), it just won't draw anything on the screen.

But p5.prototype.saveTable will crash when given wrong arguments.

So to avoid problems in the future I wrote another helper, `testUnMinified`, which when used will only run that test for the un-minified library.

akshay-99:
@limzykenneth #4468 must be merged before I push tests for saveFrames here. The tests are ready but they may fail due to the race-condition we were talking about. Once 4468 is merged, I will push the tests for saveFrames here so that they can be checked by the CI

akshay-99:
@limzykenneth @lmccart I think this is now ready to be reviewed. Tests for all download functions have been added

limzykenneth:
Looks good to me. Thanks.

Issue #4425, opened 2020-03-26
Fixup incorrect implementation of methods in Dom


fenilgandhi:
#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [x] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Details about the bug: 
While writing tests for methods in `dom` module, i noticed more than a few methods have missing edge case handling, incorrect handling of optional arguments or improper documentation of features. So i am creating this generalised issue for `dom` module. With each method that i update from now on, i will comment about what's wrong and why it needed to be fixed.

- p5.js version: 1.0.0
- Web browser and version: Chrome 80.0.3987.122
- Operating System: Manjaro 19.0.2 Kyria (Linux 4.19.108-1-MANJARO)
- Steps to reproduce this: -

fenilgandhi:
[`createInput`](https://github.com/processing/p5.js/blob/master/src/dom/dom.js#L992-L1018) - the arguments are not correctly parsed. If the user wants to set any falsy value to the input, they will not be able to do so.

fenilgandhi:
[`createRadio`](https://github.com/processing/p5.js/blob/master/src/dom/dom.js#L768-L904) 
1. The documentation is not matching with the implementation. The documentations say one param `divId` is allowed, while implementation reads `existing_radios`. 
2. Also `divId` represents id of id and input respectively, so there should be two params or an array but only a string is mentioned.
3. The implementation has an loop of count,prev whose relavance i am not able to figure out.
4. The same code is replicated for `selected` and `value` methods. 
@lmccart , @limzykenneth, i would like to re-implement it from scratch.Can i get a spec for what should be the ideal input params for this method ?

fenilgandhi:
[`createFileInput`](https://github.com/processing/p5.js/blob/master/src/dom/dom.js#L1020-L1091) - It takes 2 input params, a callback and a multiple both of which are optional. The current implementation would not allow the method to be called with only multiple=true and no value of callback.

limzykenneth:
`createInput` - should be pretty straightforward to fix and not a big change so you can probably just go ahead with it.

[...]

PR #4426, opened 2020-03-26
[WIP] [fix] Fixed dom.createInput and added unit tests

fenilgandhi:
Addresses: #4425 

 Changes:
 - Updated misleading documentation to remove `value` as optional argument.
 - Updated implementation of createInput to accept falsy value for the input
 - Added tests for createInput

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -994,8 +994,9 @@ p5.prototype.createColorPicker = function(value) {
  * Use .<a href="#/p5.Element/size">size()</a> to set the display length of the box.
  *
  * @method createInput
- * @param {String} [value] default value of the input box
- * @param {String} [type] type of text, ie text, password etc. Defaults to text
+ * @param {String} value default value of the input box
+ * @param {String} [type] type of text, ie text, password etc. Defaults to text.
+ *   Needs a value to be specified first.
  * @return {p5.Element} pointer to <a href="#/p5.Element">p5.Element</a> holding created node
  * @example
  * <div class='norender'><code>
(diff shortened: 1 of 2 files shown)


src/dom/dom.js
```diff
@@ -1009,11 +1009,11 @@ p5.prototype.createColorPicker = function(value) {
...
  */
-p5.prototype.createInput = function(value, type) {
+p5.prototype.createInput = function(value, type = 'text') {
   p5._validateParameters('createInput', arguments);
-  var elt = document.createElement('input');
-  elt.type = type ? type : 'text';
-  if (value) elt.value = value;
+  let elt = document.createElement('input');
+  elt.type = type;
+  if (value !== undefined) elt.value = value;
```
limzykenneth: I wonder if this check is necessary since `value` is a required parameter, it should never be `undefined`.

fenilgandhi: Yes, it's not. I am updating it.

limzykenneth: What is the existing behaviour that would be broken by this change? ie. If `createInput()` is called without any arguments what does it do? If it doesn't do anything particularly useful I would say that's a bug and not a breaking change. Just in my opinion.

fenilgandhi: But since, it's an input, which is literally used when we want the value from the user. So not all times, should it have a default value. I am having second thoughts about it. I think, i should revert it back to optional param and set default param as empty string. Even i am not sure.

fenilgandhi: I only considered it as a breaking change because already declared `createInput` calls would start showing errors. So, i was wondering if there's some changelog, where i should mention that ? Or is it fine.

limzykenneth: In that case I think it should be reverted to being optional while implementing a check for if `value` is undefined it is set to an empty string. Would that still show errors if reverted like this?

fenilgandhi: No, it won't. Even i feel like it should be reverted. And if reverted, it wouldn't show any errors. Cool, then it is resolved, value is optional.

lmccart:
@fenilgandhi I merged a later PR and it conflicts with this one. can you pull the latest and resolve? I was going to do it myself but didn't want to accidentally lose any of your tests.

fenilgandhi:
Hey @lmccart, i have fixed it. I re-shuffled the test suites to match the sequence of declaration on `src/dom/dom.js`, so that this problem will never arise again and also it would be easy to compare later on.

PR #4432, opened 2020-03-28
Improved_friendlyError to handle namespaced error method names

### LINKED ISSUE #4398: p5._friendlyError() can't provide reference link for methods like p5.Vector.add

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Other (Friendly Error System)

The current `p5._friendlyError(message, method)` call appends `method` string in front of https://p5js.org/reference/#p5/ URL to link to reference site for that method, but this way it can't link to URLs like https://p5js.org/reference/#/p5.Vector/add . So I suggest checking before appending if it has `'p5.'` string in start or not and append it accordingly.

Screenshot from 2020-03-21 01-04-43

ykabusalah:
Resolves #4398

### DIFF
--- a/src/core/error_helpers.js
+++ b/src/core/error_helpers.js
@@ -597,20 +597,22 @@ if (typeof IS_MINIFIED !== 'undefined') {
         const argType =
           arg instanceof Array
             ? 'array'
-            : arg === null
-            ? 'null'
-            : arg.name || typeof arg;
+            : arg === null ? 'null' : arg.name || typeof arg;
         message = `${func}() was expecting ${formatType()} for parameter #${
           errorObj.position
         } (zero-based index), received ${argType} instead`;
         break;
       }
       case 'TOO_FEW_ARGUMENTS': {
-        message = `${func}() was expecting at least ${errorObj.minParams} arguments, but received only ${errorObj.argCount}`;
+        message = `${func}() was expecting at least ${
+          errorObj.minParams
+        } arguments, but received only ${errorObj.argCount}`;
         break;
       }
       case 'TOO_MANY_ARGUMENTS': {
-        message = `${func}() was expecting no more than ${errorObj.maxParams} arguments, but received ${errorObj.argCount}`;
+        message = `${func}() was expecting no more than ${
+          errorObj.maxParams
+        } arguments, but received ${errorObj.argCount}`;
         break;
       }
     }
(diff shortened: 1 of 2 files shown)


lmccart:
@akshay-99 I know you've been working on FES stuff. I went to merge this one but realized the contents of this file had changed significantly and didn't want to mess anything up. this code patch looks good though. would you be up for taking a look and seeing if there's a way to integrate this into your work?

akshay-99:
Hi @lmccart . The contents of `error_helpers.js` were broken up into `validate_params.js`, `file_errors.js` and `fes_core.js`. This particular change would fit in `fes_core.js` after [line 128](https://github.com/processing/p5.js/blob/b3a905936d95e52157dbb80ef053b31ad8eae5b5/src/core/friendly_errors/fes_core.js#L128). 

I can't think of any easier way to merge this right now other than simply moving the code to its new place in the merge commit, but I am not sure if that will still retain the line credits for @ykabusalah if someone else is the one merging in this case. But I think they can be added as a [co-author](https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line) for the merge to retain the line credits.

Github won't show me what lines are conflicting but after pulling the changes and trying to merge this, it seems git thinks that the conflict is in `validate_params.js`. This is because of the fact that `validate_params.js` retains more lines from `error_helpers.js` than the other two files.

lmccart:
Ah yes, I meant rather than trying to merge directly, is it possible to incorporate the logic here to make the fix in what you have written already? @ykabusalah won't have line credit directly, but hopefully will feel ok that the contribution is merged and helping resolve the issue.
alternatively, @ykabusalah if you'd like to open a new PR working off of what's in the repo now, please let us know. I apologize for not getting to merge this earlier!

limzykenneth:
We could resolve the merge conflict manually offline and push the merged branch back to master and it should retain the existing commit history here.

akshay-99:
@limzykenneth  Yes that's what I was saying but the merge commit would involve moving a significant portion of the PR's code to the new place. And that would mean those lines which were moved would be attributed to whoever authored the merge commit. I was wrong about the co-author part, there's no need for that. We could just use `git commit --author="name <email>"` when committing the merge, to make sure those lines which are moved still get attributed to @ykabusalah

lmccart:
I think that sounds fine. it's not such a large change. I would prioritize getting it cleanly merged in over worrying too much about line attribution

akshay-99:
So should I go ahead with merging this offline? I won't be able to push to the master here directly so I would have to make another PR containing the merge commit

lmccart:
yes please :) thank you!

PR #4440, opened 2020-03-30
support creating canvas inside shadow dom

### LINKED ISSUE #4415: Canvas created in shadow dom is hidden indefinitely

- [ ] Color
- [ ] Data
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug: 

- p5.js version: 1.0.0
- Web browser and version: Chrome 80
- Operating System: OSX Mojave
- Steps to reproduce this:
Here's a [codepen](https://codepen.io/johnrobinson/pen/YzXJoBv?editors=1000) showing the problem 

Because the canvas's parent is inside shadow dom, the code [here](https://github.com/processing/p5.js/blob/c1e30d688748543e52aa5ebc241bfc7dfb2f508d/src/core/main.js#L345) is unable to find it to make it visible after setup has run. The current solution is to manually unhide the canvas in setup as shown in the codepen above.

jrobinson01:
Resolves #4415

 Changes:
Instead of relying solely on document querying, this PR includes a change to keep track of canvas's created in specific nodes so that they can be un-hidden and removed appropriately. There may be other places where similar issues occur but I haven't found them yet. 

re: tests - I could not find where the existing functionality of attaching the canvas to a provided node, unhiding it, and removing it were tested, so did not add tests for the updated functionality.

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/main.js
+++ b/src/core/main.js
@@ -10,6 +10,27 @@ import './shim';
 // Core needs the PVariables object
 import * as constants from './constants';
 
+// container object to hold references to parent nodes
+const containers = (function() {
+  let _containers = [];
+  return {
+    add(container) {
+      if (!_containers.includes(container)) {
+        _containers.push(container);
+      }
+    },
+    remove(container) {
+      _containers = _containers.filter(c => c !== container);
+    },
+    contains(container) {
+      return _containers.includes(container);
+    },
+    get() {
+      return _containers;
+    }
+  };
+})();
+
 /**
  * This is the p5 instance constructor.
  *
@@ -229,6 +250,9 @@ class p5 {
       this._events.devicemotion = null;
     }
 
+    // add the node to our global containers
+    node && containers.add(node);
+
     this._start = () => {
       // Find node if id given
       if (this._userNode) {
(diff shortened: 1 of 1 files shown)


fenilgandhi:
Hey, i don't fully understand the underlying code or why is it happening, but in from what i understand, `document.getElementsByTagName` does select all canvas elements in the DOM tree. So i tried adding you `el` to DOM by writing `document.body.appendChild(el)` in the `sketch.setup` method. and the canvas started showing up.

jrobinson01:
The problem is that `document.getElementsByTagName` (or any query mechanism) doesn't pierce the shadow boundary, so can't find any canvas's that are inside shadow dom.

caesarsol:
NOTE: font management is going to be a tricky problem: https://github.com/processing/p5.js/blob/3f0b2f0fe575dc81c724474154f5b23a517b7233/src/typography/loading_displaying.js#L133

Cool that you are working to fix this, thanks!

jrobinson01:
> NOTE: font management is going to be a tricky problem:
I don't think it should be an issue (did not test though). Fonts seem to be global and pierce shadow boundaries.

lmccart:
see comment https://github.com/processing/p5.js/issues/4415#issuecomment-797821162

matt-way:
@lmccart I'd like to reopen this. In your referenced comment you mention that shadow dom is not meant to be shown. I'm not exactly sure what you mean by this. Shadow dom is just a way to isolate web components. We use the shadow dom to isolate dom areas for user created code in our app. When trying to use p5 (instance mode) the canvas is initially hidden and not unhidden.

Before I dig deeper, is there a reason for initially hiding the canvas? And why aren't references kept instead or utilising getElemementByTagName?

As far as access expands go, this limitation means it is extremely difficult for users to utilise p5 in our app.

I'm happy to try create a pr to fix this bug, would just like some context around the choices above.

pixobe:
@matt-way  able to solve this issue? I am facing same issue.

georeith:
+1 can we re-open this, we're unable to use P5js effectively as our renderer lives in the Shadow DOM for stylesheet encapsulation, we have to use unstable hacks to make it work. Surely making P5js work in more environments is expanding access for everyone.

PR #4468, opened 2020-04-07
Resolves #4456, fix saveFrames()

### LINKED ISSUE #4456: saveFrames() if called one after the other will combine and mix up the frames from both into one array

- [ ] Color
- [ ] Data
- [ ] Dom
- [ ] Events
- [x] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:

- p5.js version: 1.0.0
- Web browser and version:  Chrome 79.0.3945.88
- Operating System: Ubuntu 18.04
- Steps to reproduce this:

If I use the following code
```js
function draw() {
 background(mouseX);
 }

 function mousePressed() {
   saveFrames('aaa', 'png', 1, 25, data => {
     print(data);
   });
  saveFrames('bbb', 'jpg', 1, 25, data => {
     print(data);
   });
 }
```
I would expect that since there are two different callbacks, each would get an array of roughly the same size with one callback getting frames with all filenames of aaa and all ext of type png
```js
{
  imageData: "data:image/octet-stream;base64,iVBO....", 
  filename: "aaa0", 
  ext: "png"
}
```
The other should get frames with all filenames of bbb and ext of type jpg
```js
 {
  imageData: "data:image/octet-stream;base64,/9j/4AAQSkZ...", 
  filename: "bbb0", 
  ext: "jpg"
}
```

However in actuality, one array has size 43 and the other has size 0
image

Moreover the contents that should be passed to two different callbacks is merged into the larger array and passed to just one callback
image

Check here https://editor.p5js.org/akshay.padte/sketches/8_58URlX7
I don't think this is the expected behaviour here.

akshay-99:
Resolves #4456

 Changes:

Make `frames` array local to `saveFrames` function, thereby solving a race condition that occurred when two `saveFrames` where called one after the other.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/image/image.js
+++ b/src/image/image.js
@@ -12,13 +12,6 @@
 import p5 from '../core/main';
 import omggif from 'omggif';
 
-// This is not global, but ESLint is not aware that
-// this module is implicitly enclosed with Browserify: this overrides the
-// redefined-global error and permits using the name "frames" for the array
-// of saved animation frames.
-
-/* global frames:true */ let frames = []; // eslint-disable-line no-unused-vars
-
 /**
  * Creates a new <a href="#/p5.Image">p5.Image</a> (the datatype for storing images). This provides a
  * fresh buffer of pixels to play with. Set the size of the buffer with the
@@ -472,8 +465,9 @@ p5.prototype.saveFrames = function(fName, ext, _duration, _fps, callback) {
 
   const makeFrame = p5.prototype._makeFrame;
   const cnv = this._curElement.elt;
+  let frames = [];
   const frameFactory = setInterval(() => {
-    makeFrame(fName + count, ext, cnv);
+    frames.push(makeFrame(fName + count, ext, cnv));
     count++;
   }, 1000 / fps);
 
@@ -525,7 +519,7 @@ p5.prototype._makeFrame = function(filename, extension, _cnv) {
   thisFrame.imageData = imageData;
   thisFrame.filename = filename;
   thisFrame.ext = extension;
-  frames.push(thisFrame);
+  return thisFrame;
 };
 
 export default p5;


akshay-99:
Hi @limzykenneth please can you take a loot at this?

akshay-99:
In my view this needs to be merged before the tests for `saveFrames` are added. As the tests run one after the other in quick succession, the existing race-condition described in #4456  makes them fail. It would be great if someone could review this.

limzykenneth:
I think this should be fine, just one question: if multiple `saveFrames()` are called within 15 frames of each other, what is the behaviour?

akshay-99:
@limzykenneth I am sorry I didn't get your question. Do you mean something like
```js
saveFrames('aaa', 'png', 1, 15, (data)=>console.log(data));
saveFrames('bbb', 'jpg', 1, 15, (data)=>console.log(data));
saveFrames('ccc', 'jpg', 1, 15, (data)=>console.log(data));
```

This with the code on master would yield 3 unequal arrays of sizes around 45, 0, 0 and files aaa, bbb, ccc will all be mixed up into the first array
While the code on this PR gives 3 roughly equal arrays with sizes around 10, 10, 10 with aaa, bbb, ccc being in 3 separate arrays.

If this was not what you meant, could you please clarify?

limzykenneth:
@akshay-99 Shouldn't `saveFrames()` output 15 frames and somehow in this instance it is only outputing 10 frames each? Or am I mistaken here?

akshay-99:
@limzykenneth Ah yes, sorry I made a typing mistake :see_no_evil: . It's indeed 15, 15, 15 not 10, 10, 10
image

Another thing I noticed is sometimes those numbers can be 1 less than the expected if the system is running slow and the interval is cleared before it gets a chance to process the expected number of frames. But this problem also existed on the previous code.

limzykenneth:
That is fine then I think, I'll approve the PR but I'd like to get someone else to have a look as well if you don't mind.

lmccart:
this looks good to me. thanks @limzykenneth for reviewing and @akshay-99 for the fix! some tests for `saveFrames()` would be great if you have the time.

akshay-99:
Thanks @lmccart @limzykenneth ! I will add the tests for saveFrames to #4418 soon

PR #4488, opened 2020-04-29
loadImage() documentation

### LINKED ISSUE #4443: loadImage() documentation - mention base64 string as a parameter
Documentation could state that `loadImage()` accepts a base64 encoded image as a parameter (example: https://stackoverflow.com/a/51162033), as an alternative to the file path. I found this useful to avoid the CORS block when testing locally (if one doesn't wish to setup a webserver).

https://github.com/processing/p5.js/blob/f503b0b81aa86671e8ea900133985b3d26556dac/src/image/loading_displaying.js#L30

sm7515:
Resolves #4443 

 Changes:

added documentation for `loadImage()` to mention base64 string as a parameter
https://github.com/processing/p5.js/blob/f503b0b81aa86671e8ea900133985b3d26556dac/src/image/loading_displaying.js#L30

 Screenshots of the change:

Screen Shot 2020-04-29 at 7 37 02 PM

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -25,6 +25,9 @@ import '../core/error_helpers';
  * that links in your sketch. Loading an image from a URL or other
  * remote location may be blocked due to your browser's built-in
  * security.
+ * <br><br>
+ * You can also pass in the string of a base64 encoded image as an alternative to the file path.
+ * Remember to add "data:image/png;base64," in front of the string.
  *
  * @method loadImage
  * @param  {String} path Path of the image to be loaded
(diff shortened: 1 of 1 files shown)


src/image/loading_displaying.js
```diff
@@ -25,6 +25,9 @@ import '../core/error_helpers';
  * that links in your sketch. Loading an image from a URL or other
  * remote location may be blocked due to your browser's built-in
  * security.
+ * <br><br>
```
lmccart, line 28: you can simply leave this line empty and the empty line after the paragraph break will be added

PR #4524, opened 2020-05-08
docs for `createVideo`: change `video()` to `image()`

joshuahhh:
As examples like https://p5js.org/examples/dom-video-canvas.html suggest, you use the `image()` function to draw a video to the canvas. (There is no `video()` function.)

(I assume a PR template is not require for documentation typo fixes, but please let me know if there's something else I should provide here. Thanks so much!)


lmccart:
thanks!

PR #4563, opened 2020-05-19
supporting existing canvas elements for createGraphics

valentinptc:
Extend the createGraphics function with the ability to pass an existing canvas. As such you can use existing canvas elements for graphics processing.


limzykenneth:
Hi @valentinptc, we generally do not accept PR without a corresponding issue. New features should be discussed in issues and agreed upon before a PR should be filed. Please file an issue for this feature if you wish it to be included in the library. Thanks.

valentinptc:
Here you go: #4564

valentinptc:
@limzykenneth Here is also a discussion from the forum:
https://discourse.processing.org/t/creategraphics-referencing-existing-canvas-context/21067

PR #4568, opened 2020-05-22
Added ssv option in loadTable function

BetaBiest:
FeatureRequest

 Changes:
Just added the option to pick semicolon (';') (Semicolon Seperated Value) as separator for the p5.prototype.loadTable function.

 Screenshots of the change:
Befor:
befor
Added in between:
after

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -421,6 +421,13 @@ p5.prototype.loadTable = function(path) {
           sep = ',';
           separatorSet = true;
         }
+      } else if (arguments[i] === 'ssv') {
+        if (separatorSet) {
+          throw new Error('Cannot set multiple separator types.');
+        } else {
+          sep = ';';
+          separatorSet = true;
+        }
       } else if (arguments[i] === 'tsv') {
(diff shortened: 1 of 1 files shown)


src/io/files.js
```diff
@@ -421,6 +421,13 @@ p5.prototype.loadTable = function(path) {
           sep = ',';
           separatorSet = true;
         }
+      } else if (arguments[i] === 'ssv') {
+        if (separatorSet) {
+          throw new Error('Cannot set multiple separator types.');
```
lmccart, line 426: let's remove these errors throughout, while we're here.
incorrect argument orders should be caught by the friendly error system, not checked within this logic
would you be up for making this change to remove each of the errors here?

BetaBiest: I looked into it but i dont feel confident enough to implement the FES checks.
However I added the new options in the documentation.

lmccart: ok no problem. thanks for making the other edit

PR #4577, opened 2020-05-25
Resolves #4556

### LINKED ISSUE #4556: Style width/height etc throws error when passed number value as argument

- [ ] Color
- [ ] Data
- [x] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:

- p5.js version:  1.0.0
- Web browser and version:  Firefox 76.0.1 (presumably all other supported browsers as well)
- Operating System:  macOS 10.14
- Steps to reproduce this:

```
function setup() {
  createCanvas(400, 400);
  
  let div0 = createDiv('hello');
  div0.style('width', 200+'px'); //works
  div0.style('height', 300); //Uncaught TypeError: val.replace is not a function (sketch: line 6)
  div0.position(50, 50, 'absolute');
}
```

As reported by @mkmori in #3447, calling something like `p5.Element.style('width', 200)` will cause a TypeError to be thrown because the function is expecting a string. Passing a number should be a valid value since it is a valid CSS value (default to pixels).

This should affect `width`, `height`, `top` and `left`. A check should be in place [here](https://github.com/processing/p5.js/blob/master/src/dom/dom.js#L1924) to see if the value is a number of not, if it is don't do the `replace`.

nebbles:
Resolves #4556

 Changes:

- Changes the expected type of the value provided to Element style() method. Since Number is silently rejected by the browser, it is no longer an expected type.
- Improve style setting so that width, height, top, left will correctly update the internal state of Element using the WebAPI. This assures that if the browser silently rejects a numerical value for the value of a style setting, we are fetching the correct value as rendered by the browser.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -1892,12 +1892,13 @@ p5.Element.prototype.style = function(prop, val) {
   }
 
   if (typeof val === 'undefined') {
-    // input provided as single line string
     if (prop.indexOf(':') === -1) {
-      var styles = window.getComputedStyle(self.elt);
-      var style = styles.getPropertyValue(prop);
+      // no value set, so assume requesting a value
+      let styles = window.getComputedStyle(self.elt);
+      let style = styles.getPropertyValue(prop);
       return style;
     } else {
+      // value set using `:` in a single line string
       var attrs = prop.split(';');
       for (var i = 0; i < attrs.length; i++) {
         var parts = attrs[i].split(':');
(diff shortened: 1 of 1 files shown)


lmccart:
thanks! sorry this took so long to merge

Issue #4571, opened 2020-05-24
FES: validateParameters does not detect trailing undefined arguments


akshay-99:
#### Most appropriate sub-area of p5.js?

- [x] Other (Friendly Errors)

As mentioned in #4558

Consider the sketch
```js
let r, g, b;

r = 40;
b = 10;
function setup() {
  createCanvas(400, 400);
  color(r, g, b);
}
```
As expected we get a validation message
```
 p5.js says: color() was expecting Number for parameter #1 (zero-based index), 
received an empty variable instead. If not intentional, this is often a problem with scope: 
[https://p5js.org/examples/data-variable-scope.html] at about:srcdoc:60:3. [http://p5js.org/reference/#p5/color] 
````

But for the following, no validation message is generated
```js
let r, g, b;

r = 40;
g = 10; // now b would be undefined
function setup() {
  createCanvas(400, 400);
  color(r, g, b);
}
```

akshay-99:
Fixing this would also involve changing the code inside `p5.proptotype.save` [here](https://github.com/processing/p5.js/blob/master/src/io/files.js#L1488) .
The problem is that when we call `save()` to save a JSON like this `save(myJSON, 'my.json')`, a call is made to `saveJSON` with the arguments `(args[0], args[1], args[2])`. However, args[2] is optional and is undefined in this case. If we fix this issue, then we would have a `ValidationError` here, resulting in a test failure.

To solve this problem, we could instead apply a check on `arguments.length` in `save()` and only pass the optional third parameter  to `saveJSON` only if `arguments.length === 3`. 

Similarly in selectAll example 2 [here](https://p5js.org/reference/#/p5/selectAll) since `d` and `f` would be `null` as no DOM elements exist that fit the selection criteria, on fixing this issue, this example also gives a `ValidationError`, resulting in a test failure.

So if we were to go ahead and fix this issue, we must either change this example so that `d` and `f` are not `null` anymore, or somehow make the FES bypass this particular example.

@stalgiag what are your thoughts on the selectAll problem?

limzykenneth:
I may be wrong but isn't parameter validation compared against the documentation? In that it doesn't really matter what the source code itself checks for `parameterValidation` just checks the number of arguments and type of arguments against the documentation.

Parameter validation did not throw a warning in the case of `color()` is because with 2 parameters that are numbers, they are still valid (being `color(gray, [alpha])`).

akshay-99:
[...]

PR #4580, opened 2020-05-27
selectAll docs: include missing param types

### LINKED ISSUE #4579: selectAll 2nd param specification misses types
- [x] DOM
- [x] Other (Reference and Friendly Errors)

In `selectAll()`, according to the reference [here](https://p5js.org/reference/#/p5/selectAll), the 2nd param can be of types String, p5.Element or HTMLElement. However, in the syntax of specifying this, only String is mentioned [here](https://github.com/processing/p5.js/blob/master/src/dom/dom.js#L97) and hence the FES thinks only String can be allowed. 
image

It should be like the param in select [here](https://github.com/processing/p5.js/blob/master/src/dom/dom.js#L31) ( the types separated by the | symbol )

akshay-99:
Resolves #4579

 Changes:
Include missing types ( HTMLElement and p5.Element ) for the 2nd parameter in selectAll documentation

Quite a minor change but needed this for working on #4571 :sweat_smile: 

While we are at it, can I also update the example so that we select an element that exists in the reference tests ( perhaps `#mocha` or `head` or `body` instead of `beep` ) ?

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -117,10 +117,14 @@ p5.prototype.select = function(e, p) {
  * a = selectAll('button', '#myContainer');
  *
  * let d = select('#container');
- * a = selectAll('p', d);
+ * if (d) {
+ *   a = selectAll('p', d);
+ * }
  *
  * let f = document.getElementById('beep');
- * a = select('.blah', f);
+ * if (f) {
+ *   a = selectAll('.blah', f);
+ * }
  *
  * a; // unused
  * </code></div>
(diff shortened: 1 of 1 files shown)


lmccart:
looks good, sure go ahead and update the example and we can merge after that

akshay-99:
Thanks @lmccart. Instead of changing the names, I have made it more like the [select example](https://p5js.org/reference/#/p5/select) by adding conditions to check if `d` and `f` exist. This would also work.

PR #4586, opened 2020-05-28
resolves #4585

### LINKED ISSUE #4585: Support for explicitly setting file type of .obj or .stl files

- [ ] Color
- [ ] Data
- [ ] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL

Currently the file type for the model is determined by the file extension of the path given to `loadModel()`. This is a limitation in the case that the model exists on a server outside of ones control but that has a different file extension. In this case, currently it is not possible to load the model even though model would be supported. It would be helpful to be able to explicitly give `loadModel()` the file type of the model being loaded. 

### Example
-  Server hosts a `.obj` file with a randomized file name with no extension. 

`https://www.example.com/aAde3AsdfaHy` 

- Another example is when you have a valid file type but the path requires a signature as query parameter.

`https://www.example.com/model.obj?AWSAccessKeyId=aSftajdS123aS`

ismailnamdar:
Resolves #4585 

 Changes:

- Added `fileType` parameter to `loadModel()`.
- Added case insensitive check for `fileType` to cover also capitalized extensions (e.g`.STL`).

 

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/loading.js
+++ b/src/webgl/loading.js
@@ -34,6 +34,8 @@ import './p5.Geometry';
  *                                     the 3D model object.
  * @param  {function(Event)} [failureCallback] called with event error if
  *                                         the model fails to load.
+ * @param  {String} [fileType]          The file extension of the model
+ *                                      (<code>.stl</code>, <code>.obj</code>).
  * @return {p5.Geometry} the <a href="#/p5.Geometry">p5.Geometry</a> object
  *
  * @example
@@ -96,6 +98,7 @@ import './p5.Geometry';
  * @param  {String} path
  * @param  {function(p5.Geometry)} [successCallback]
  * @param  {function(Event)} [failureCallback]
+ * @param  {String} [fileType]
  * @return {p5.Geometry} the <a href="#/p5.Geometry">p5.Geometry</a> object
  */
 p5.prototype.loadModel = function(path) {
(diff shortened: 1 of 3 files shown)


ismailnamdar:
So, I added `fileType` as the last parameter of the `loadModel()`. Alternatively, we can have an `options` object as a second parameter and we can check if the second parameter is boolean or object. By doing that, we can use both `normalize` and `options`.

src/webgl/loading.js
```diff
@@ -96,29 +98,33 @@ import './p5.Geometry';
...
+    fileType = arguments[4];
   } else {
     normalize = false;
     successCallback = arguments[1];
     failureCallback = arguments[2];
+    fileType = arguments[3];
   }
 
-  const fileType = path.slice(-4);
+  fileType = fileType == null ? path.slice(-4) : fileType;
```
lmccart, line 122: can you just set this on line 109, then it gets overwritten if specified by the if statements below?

ismailnamdar: Thanks. I moved the default definition on line 109 and if `fileType` is given, it will overwrite  the default one.

src/webgl/loading.js
```diff
@@ -106,20 +106,23 @@ p5.prototype.loadModel = function(path) {
...
   let successCallback;
   let failureCallback;
-  let fileType;
+  let fileType = path.slice(-4);
   if (typeof arguments[1] === 'boolean') {
     normalize = arguments[1];
     successCallback = arguments[2];
     failureCallback = arguments[3];
-    fileType = arguments[4];
+    if (arguments[4] != null) {
```
stalgiag, line 114: Can we do a `typeof arguments[4] != 'undefined'` here and on line 121 instead? It is more consistent with the style of the rest of the library.

ismailnamdar: Sure, I added that.

stalgiag:
@all-contributors please add @ismailnamdar for code and testing!

Issue #4583, opened 2020-05-28
FES: Disable parameter validation for internal calls


akshay-99:
#### Most appropriate sub-area of p5.js?

- [x] Other (Friendly Errors)

#### Feature enhancement details:
In #4571, I [mentioned](https://github.com/processing/p5.js/issues/4571#issuecomment-633219382) a number of problems that must also be addressed before addressing the main issue.  In essence, I was talking about p5 functions that pass forward arguments which they didn't even receive in the first place.  Quoting myself from there:
> Fixing this would also involve changing the code inside p5.proptotype.save [here](https://github.com/processing/p5.js/blob/master/src/io/files.js#L1488).
The problem is that when we call save() to save a JSON like this save(myJSON, 'my.json'), a call is made internally to saveJSON with the arguments (args[0], args[1], args[2]). However, args[2] is optional and is undefined in this case. 

> loadModel documentation, example 1.
`ValidationError: loadStrings() was expecting Function for parameter #2 (zero-based index), received an empty variable instead.`
We are not passing any callbacks in the example and that's okay because those are optional parameters. But the code here passes the `failureCallback` argument to `loadStrings` without checking if it exists. And so we see a validation error in `loadStrings`. This can also be fixed via a simple check.

But these are just the visible ones that would result in a test failure. I do not know how many other instances there are in the source where something similar happens. So instead of addressing these directly, I thought they could be addressed alongside another known issue of the FES.  Suppose we run the sketch 
```js
function setup() {
 createCanvas(400, 400);
 let data = {message: 'hi'}, file;
 /* some code */
 saveJSON(data, file);
}
```
We get `🌸 p5.js says: saveJSON() was expecting at least 2 arguments, but received only 1`
and also
`🌸 p5.js says: saveStrings() was expecting String for parameter #1 (zero-based index), received an empty variable instead`
The error for `saveStrings` is confusing here because it was never called by the user. 

I suggest that the FES shouldn't display parameter validation messages when a function is called internally by another p5 function. This would also indirectly "solve" the problems I mentioned in #4571

akshay-99:
My idea for this to see the stack trace [here](https://github.com/processing/p5.js/blob/master/src/core/error_helpers.js#L768) and get the functions in it.
For a p5 function called from the user's sketch we would see a stack
|                                  |
| -----------------------| 
| _friendlyParamError  |
| validateParameters   |
| some p5 functon      |
| some non p5 function |

We can display a validation message in this case.

[...]

PR #4590, opened 2020-05-29
FES: Split error_helpers.js 

akshay-99:
Addresses #4583

 Changes:

The file `error_helpers.js` was growing too large, so as discussed in #4583, I have split it into 3 files inside a new folder `core/friendly_errors`

- validate_params.js
- file_errors.js
- fes_core.js

I also had to change the imports in other files to reflect this.

The blame history for the files is still retained as seen here:
- [validate_params.js](https://github.com/akshay-99/p5.js/blame/master/src/core/friendly_errors/validate_params.js)
- [file_errors.js](https://github.com/akshay-99/p5.js/blame/master/src/core/friendly_errors/file_errors.js)
- [fes_core.js](https://github.com/akshay-99/p5.js/blame/master/src/core/friendly_errors/fes_core.js)

I am not sure if this could be bundled alongside the fix for #4583 or whether this fits better as an independent PR. Until that is decided, I am marking this as draft.

#### PR Checklist

- [x] `npm run lint` passes


akshay-99:
I have pushed the check to prevent validation for internal calls in the latest commit. So for the sketch mentioned in #4583, earlier we would get validation messages for both `saveJSON` and `saveStrings`. Now the one for `saveStrings` will not be displayed.

I decided to use [stacktrace.js](https://www.npmjs.com/package/stacktrace-js) as the regex [here](https://github.com/processing/p5.js/blob/7449231c082681ae69e794e93a63e1615239a991/src/core/error_helpers.js#L767) is very chrome-specific. The regex aims to extract the location in the sketch where the error occurred but it doesn't work on other browsers.
So on chrome, it gives:
`🌸 p5.js says: random() was expecting Number for parameter #0 (zero-based index), received string instead at http://localhost:8100/sketch.js:4:5. (http://p5js.org/reference/#p5/random)`
But on Firefox, I see:
`🌸 p5.js says: random() was expecting Number for parameter #0 (zero-based index), received string instead. (http://p5js.org/reference/#p5/random)`

But importing stacktrace.js directly increased the size of the final p5 library by 200 KB so I instead opted to just borrow the part that we need and put that in with its license and copyright. This increases the size of the final library by around 20 KB.

The test for `p5.Image.blend` initially failed because from [here](https://github.com/processing/p5.js/blob/7449231c082681ae69e794e93a63e1615239a991/test/unit/image/pixels.js#L158)  the call actually goes to `p5.Image.prototype.blend` [here](https://github.com/processing/p5.js/blob/7449231c082681ae69e794e93a63e1615239a991/src/image/p5.Image.js#L783) which then calls `p5.prototype.blend`. The former does not have parameter validation but the latter does. The validation message due to the internal call from `p5.Image.prototype.blend` to `p5.prototype.blend` would be suppressed by this PR and so I had to add validation for `p5.Image.prototype.blend` as well.

akshay-99:
Also, I don't really understand the need for importing `error_helpers.js` in files such as `src/color/creating_reading.js`, `src/core/shape/2d_primitives.js`, `src/core/shape/curves.js`, etc. I thought that it would be required in files which use FES functionality. But some files which also use the FES, notably `src/math/random.js` , `src/dom/dom.js`, `src/image/image.js` don't import it.

I tried commenting out these imports and all the tests still pass and the FES still seems to be working correctly. The functions such as `_validateParameters` get assigned to the p5 object, so shouldn't it be available regardless of the import ( as long as the FES files are imported in `app.js`) ?

akshay-99:
Hi @stalgiag I think I have almost solved #4571 as well. But I have to wait for this to be merged before creating the PR for that otherwise it's tests would fail

stalgiag:
Looking good! Will review today @akshay-99.

akshay-99:
Thanks @stalgiag. I will proceed with #4571 now

PR #4602, opened 2020-06-04
Modified select() and selectAll() functions #4584

### LINKED ISSUE #4584: Using querySelector and querySelectorAll for select and selectAll

- [ ] Color
- [ ] Data
- [x] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

A couple issues pop up when I'm trying to use `select()` and `selectAll()`. Both of them are related to the CSS selection pattern of `.class1, .class2` (ie. selecting elements using multiple queries seperated by commas). 

In the case of `select()` it always returns null instead of the first element selected. For `selectAll()` it returns an empty array instead of array of elements matching the CSS selector.

```javascript
function setup() {
  console.log(document.querySelector("body, html"));
  console.log(select("body, html"));
  
  console.log(document.querySelectorAll("body, html"));
  console.log(selectAll("body, html"));
}
```

I wonder if there's any reason for us to not make `select()` and `selectAll()` a wrapper around `querySelector()` and `querySelectorAll()` respectively. If not I think switching to these API would help simplify the codebase a bit and extend support to whatever the native API support which would cover a lot of the CSS selector syntax.

liampuk:
Resolves #4584

 Changes:

- Modified `select()` and `selectAll()` functions to use `querySelector()` and
`querySelectorAll()` respectively.
- Modified helper function `getContainer()` to use `querySelector()` instead of `getElementById().`
- Added unit tests and modified examples to cover extra functionality provided by these changes.

I have left in the optional container parameter to provide backwards compatibility.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -56,8 +56,8 @@ import p5 from '../core/main';
  */
 p5.prototype.select = function(e, p) {
   p5._validateParameters('select', arguments);
-  var container = getContainer(p);
-  var res = container.querySelector(e);
+  const container = this._getContainer(p);
+  const res = container.querySelector(e);
   if (res) {
     return this._wrapElement(res);
   } else {
@@ -66,8 +66,8 @@ p5.prototype.select = function(e, p) {
 };
 
 /**
- * Searches the page for elements that match the given CSS selector string (can be a class, tag
- * name or combination of both) and returns them as <a href="#/p5.Element">p5.Element</a>s in
+ * Searches the page for elements that match the given CSS selector string (can be an ID a class,
+ * tag name or a combination) and returns them as <a href="#/p5.Element">p5.Element</a>s in
  * an array.
  * The DOM node itself can be accessed with .elt.
  * Returns an empty array if none found.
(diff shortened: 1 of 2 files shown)


src/dom/dom.js
```diff
@@ -19,64 +19,45 @@
...
-    res = container.getElementById(e);
-  } else {
-    res = container.getElementsByTagName(e);
-    if (res.length) {
-      res = res[0];
-    } else {
-      res = null;
-    }
-  }
+  var res = container.querySelector(e);
```
lmccart, line 60: these should be `const` now that we've switched to ES6, not sure how they were missed before

src/dom/dom.js
```diff
@@ -85,17 +66,17 @@ p5.prototype.select = function(e, p) {
 };
 
 /**
- * Searches the page for elements with the given class or tag name (using the '.' prefix
- * to specify a class and no prefix for a tag) and returns them as <a href="#/p5.Element">p5.Element</a>s
- * in an array.
+ * Searches the page for elements that match the given CSS selector string (can be a class, tag
```
lmccart, line 69: can you add ID in here too

src/dom/dom.js
```diff
@@ -112,34 +93,26 @@ p5.prototype.select = function(e, p) {
...
   var arr = [];
-  var res;
   var container = getContainer(p);
-  if (e[0] === '.') {
-    e = e.slice(1);
-    res = container.getElementsByClassName(e);
-  } else {
-    res = container.getElementsByTagName(e);
-  }
+  var res = container.querySelectorAll(e);
```
lmccart, line 115: can you update this function to use `const`?

PR #4606, opened 2020-06-04
FES: Add detection for trailing undefined arguments

### LINKED ISSUE #4571: FES: validateParameters does not detect trailing undefined arguments

akshay-99:
Resolves #4571 

 Changes:

1. Make the FES detect trailing undefined arguments
Consider the following few examples
```js
let r, g, b;
function setup() {
  let r = 50;
  color(r, g, b);
}
```
Here g and b are undefined. But earlier as color with only one argument was still a valid overload, the FES did not log a message. But as it is passed 3 arguments, it is likely that them being undefined is unintentional and so a message should be displayed.

Now it shows:
`🌸 p5.js says: color() was expecting Number for parameter #1 (zero-based index), received an empty variable instead`

`🌸 p5.js says: color() was expecting Number for parameter #2 (zero-based index), received an empty variable instead
`

Similarly:
```js
function setup() {
  let a = document.getElementById('nonexistant');
  select('hgg', a);
}
```
would not have logged an error before this, as the undefined argument is trailing and select(string) is a valid overload. A message here is important to let the user know that `a` is empty ( i.e `notexistant` was not found )

Now it shows:
`🌸 p5.js says: select() was expecting String|p5.Element|HTMLElement for parameter #1 (zero-based index), received an empty variable instead.`

Also for
```js
let r;
circle(10, 10, r);
```
Earlier we used to get 
`🌸 p5.js says: circle() was expecting at least 3 arguments, but received only 2`

Now it correctly shows:
`🌸 p5.js says: circle() was expecting Number for parameter #2 (zero-based index), received an empty variable instead.`

2. Pass the error type to the ValidationError constructor so that it could be accessed in tests
3. Add tests for the edge cases in #2740 
4. Add tests for detection of trailing undefined arguments

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/friendly_errors/validate_params.js
+++ b/src/core/friendly_errors/validate_params.js
@@ -467,10 +467,11 @@ if (typeof IS_MINIFIED !== 'undefined') {
   // tests when expecting validation errors
   p5.ValidationError = (name => {
     class err extends Error {
-      constructor(message, func) {
+      constructor(message, func, type) {
         super();
         this.message = message;
         this.func = func;
+        this.type = type;
         if ('captureStackTrace' in Error) Error.captureStackTrace(this, err);
         else this.stack = new Error().stack;
       }
@@ -537,7 +538,7 @@ if (typeof IS_MINIFIED !== 'undefined') {
           return;
         }
         if (p5._throwValidationErrors) {
-          throw new p5.ValidationError(message);
+          throw new p5.ValidationError(message, func, errorObj.type);
         }
         const location = `${parsed[3].fileName}:${parsed[3].lineNumber}:${
           parsed[3].columnNumber
@@ -603,8 +604,10 @@ if (typeof IS_MINIFIED !== 'undefined') {
     const docs = docCache[func] || (docCache[func] = lookupParamDoc(func));
     const overloads = docs.overloads;
 
-    // ignore any trailing `undefined` arguments
     let argCount = args.length;
+
+    // the following line ignores trailing undefined arguments, commenting
+    // it to resolve https://github.com/processing/p5.js/issues/4571
     // '== null' checks for 'null' and typeof 'undefined'
     // while (argCount > 0 && args[argCount - 1] == null) argCount--;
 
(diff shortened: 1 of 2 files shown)


stalgiag:
Hm - tests passed on this PR before but [after merge](https://github.com/processing/p5.js/runs/751034057?check_suite_focus=true) we are failing on the `selectAll()` documentation. I am looking into it.

akshay-99:
@stalgiag It seems it's because the example was updated again in #4602 after this PR was created.  There needs to be a condition to check if `c` exists before [line 105](https://github.com/processing/p5.js/blame/master/src/dom/dom.js#L105) here and also to see if `b` exists before [line 101](https://github.com/processing/p5.js/blame/master/src/dom/dom.js#L101)

akshay-99:
> Ah yes, now I am remembering that the examples for save, selectAll and loadModels are broken with these changes.

The ones for `save` and `loadModel` are no longer broken as there it was a validation message due to an internal call to a library function from another library function which was suppressed in #4590. The example for selectAll was also fixed in #4580  but it was overwritten in #4602. I am making a PR to fix it.

akshay-99:
Whoops looks like @lmccart already fixed it. Thanks! :smile:

lmccart:
oh sorry I missed this thread!

PR #4618, opened 2020-06-07
Add lazyLog Feature

### LINKED ISSUE #4617: A moderate logging system

- [ ] Color
- [ ] Data
- [ ] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### New feature details:

As the `draw()` function runs unstoppably for many times a second, when we want to log out the value of a variable, it usually results in massive messages in console and weakened performance. I worked on a moderate logging system that will only log the value in console 1. when the value of the particular variable changes, or 2. periodically with the period preset. I hope this feature will be helpful for people to spot useful information in the console and help lift the performance of the code during the debugging process.

peilingjiang:
Resolves #4617 

Changes:
I made changes to the main.js and environment.js to make the `lazyLog()` feature possible. I also make some minor changes to inline_documentation.md in order to resolve a typo and layout problem.

#### PR Checklist

- [x] `npm run lint` passes


src/core/environment.js
```diff
@@ -737,4 +737,80 @@ p5.prototype.getURLParams = function() {
...
+  let argsLength = args.length;
+  let displayName = true;
+
+  if (!argsLength) {
+    // Clear the cached variables
+    this._lazyLogVars = {};
+    this._lazyLogTimers = {};
+    return false;
+  }
+  if (typeof args[argsLength - 1] === 'boolean' && !args[argsLength - 1]) {
```
dhoizner, line 791: i wonder if unpacking the args at entry to `lazyLog` would make it easier to follow how the different cases work. i think i get it, but i would have more confidence making a change to this in the future if the args were unpacked and labeled.

src/core/environment.js
```diff
@@ -737,4 +737,80 @@ p5.prototype.getURLParams = function() {
...
+ */
+/**
+ * @method lazyLog
+ */
+p5.prototype.lazyLog = function(...args) {
+  let argsLength = args.length;
+  let displayName = true;
+
+  if (!argsLength) {
+    // Clear the cached variables
```
dhoizner, line 786: is clearing when no args are passed a common pattern? having a separate function or explicit parameter for this feels like it would be safer.

peilingjiang: Hi @dhoizner, thanks for pointing it out. I opened this pull request before noticing the ongoing discussion around "adding new features only if accessibility increased," and I hope this function could help people access the library better by providing less while more useful and concentrated information in the console, with less computing resources required. With that being said, I also want the interface of it to be as simple as possible. How do you think about `lazyLog(CLEAR)` or a separate `clearLog()`/`clearLazyLog()` function?

dhoizner: hi @peilingjiang i think this would make debugging a lot easier, so i'm all for bringing this in (although i'm not one of the maintainers!). i think that having a `clearLazyLog` makes the most sense to me.

PR #4627, opened 2020-06-09
Fix center() not working with positioned parent elements

### LINKED ISSUE #4623: center() does not work correctly if the element has a parent

- [x ] Core/Environment/Rendering
- [x ] Dom

#### Details about the bug:

- p5.js version:February 29, 202
- Web browser and version: Firefox, Chrome, Edge (does not seem to be a browser problem)
- Operating System: Windows 10 (probably not a OS problem);
- Steps to reproduce this:
The center method does not work correctly if the element has a parent.  In the code 
below, both the canvas and orange division should be center in the pinkish division.
If the parent lines are commented out, the centering seems valid.  The horizontal centering seems OK if the parent is on the left side of the page.  The code can be 
run at [http://brinkje.com/p5js/center.html]
<pre>
function setup() {
  let div3 = createDiv().size(100, 100);
  div3.style("background-color: #DDEEFF; display: inline-block")
  div = createDiv("The division").size(200, 200);
  let div.style("background-color: #FFDDEE; display: inline-block");
  let can = createCanvas(100, 100);
  can.background(220);
  can.parent(div);
  can.text("canvas", 10, 10);
  can.center();
  let div2 = createDiv().size(10, 10);
  div2.style("background-color", "orange");
  div2.parent(div);
  div2.center();
} // setup

peilingjiang:
Resolves #4623

 Changes:
The old `center()` function ignores the offsets of the parent element, thus won't always set the correct position to the child element when the parent was somewhere else on the page instead of positioned at `(0, 0)`.

 Screenshots of the change:
Screen Shot 2020-06-09 at 15 02 38
Before
Screen Shot 2020-06-09 at 15 02 11
After
(Code adapted from #4623)

#### PR Checklist

- [x] `npm run lint` passes


limzykenneth:
Looks good to me. Thanks @peilingjiang

PR #4629, opened 2020-06-11
FES: some i18n improvements and add i18n support for parameter validation messages

### LINKED ISSUE #4616: FES: internationalization for validateParameters

- [x] Other (Friendly errors, internationalization)

#4233 adds internationalization support for FES functionality with the exception of `validateParameters`. Perhaps I could add that in?

Also since the entire internationalization tooling is not used in the minified version, perhaps we can import it conditionally by checking for `window.IS_MINIFIED` and exclude it in the browserify config for the minified build. This would also remove all of `i18next` and `i18next-browser-languagedetector` from the minified script, cutting down the final size quite a bit. 

Also, [this line](https://github.com/processing/p5.js/blob/47da9a508f48d9eeaeb6e9b15b8b4ec8f9ea607b/tasks/build/browserify.js#L42) tries to exclude translations from the minified build, but if we search for something like "Su browser impidío un medio tocar" in the [minified file](https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.0.0/p5.min.js), we can see that the translations are still included

### LINKED ISSUE #4626: FES: Simplifying parameter validation messages

- [x] Other (Friendly errors)

Before starting out with the internationalization for `validateParameters` in #4616, I felt this needed to be discussed first. So the messages for the FES wrong arguments are still ( kind of ) cryptic in the way they describe the position of the arguments. It says something like `parameter #3 (zero-based index)`,  which could be changed to simply to state the ordinal form such as `fourth parameter`. 
This is a very small change but would be nice to have. 

The thing is how this would play out with different languages, as they have different ordinal representations. We could use something similar to the [locales from numeral.js](https://github.com/adamwdraper/Numeral-js/tree/master/src/locales) to display the ordinal short forms (1st, 2nd, 3rd, ...)

Or, since p5 functions don't usually have more than 10 arguments ( not the ones commonly used by beginners at least ), simply having a list of ordinal full forms (first, second, third, ...) for numbers 1 to 10 for all languages in the translation files won't consume much of extra space.

akshay-99:
Resolves #4616 
Resolves #4626 

Also fixed one more bug I came across along the way (  included in the list below )
 
### Changes:

- Simplify friendlyParamError messages to the format mentioned in #4626 .
image

- Add internationalization support for `validateParameters` messages. 

- Remove `i18next`, `i18next-browser-languagedetector` and the translation json files from the minified build, reducing minified library size from 637 KB to 587 KB.
- Support incomplete translation files ( If a string is empty in a translation file, give the corresponding string from the default language file instead. ). Since the spanish translation file is incomplete, the translator used to return empty strings for some keys
image

   Now it defaults to the corresponding English version when it encounters an empty string
image

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/internationalization.js
+++ b/src/core/internationalization.js
@@ -36,6 +36,7 @@ export const initialize = () =>
         nestingPrefix: '$tr(',
         nestingSuffix: ')',
         defaultNS: 'translation',
+        returnEmptyString: false,
         interpolation: {
           escapeValue: false
         },
--- a/src/core/internationalization.js
+++ b/src/core/internationalization.js
@@ -1,6 +1,11 @@
 import i18next from 'i18next';
 import LanguageDetector from 'i18next-browser-languagedetector';
-import resources from '../../translations';
+
+let resources;
+// Do not include translations in the minified js
+if (typeof IS_MINIFIED === 'undefined') {
+  resources = require('../../translations').default;
+}
 
 /**
  * This is our translation function. Give it a key and
--- a/tasks/build/browserify.js
+++ b/tasks/build/browserify.js
@@ -39,7 +39,9 @@ module.exports = function(grunt) {
         browseified = browseified
           .exclude('../../docs/reference/data.json')
           .exclude('../../../docs/parameterData.json')
-          .ignore('../../translations/index.js');
+          .exclude('../../translations')
+          .ignore('i18next')
+          .ignore('i18next-browser-languagedetector');
       }
 
       const babelifyOpts = { plugins: ['static-fs'] };
--- a/translations/es/translation.json
+++ b/translations/es/translation.json
@@ -13,7 +13,28 @@
       "table": "",
       "xml": ""
     },
+    "friendlyParamError": {
+      "type_EMPTY_VAR": "",
+      "type_TOO_FEW_ARGUMENTS": "",
+      "type_TOO_MANY_ARGUMENTS": "",
+      "type_WRONG_TYPE": ""
+    },
+    "location": "",
     "misusedTopLevel": "",
+    "positions": {
+      "p_1": "",
+      "p_10": "",
+      "p_11": "",
+      "p_12": "",
+      "p_2": "",
+      "p_3": "",
+      "p_4": "",
+      "p_5": "",
+      "p_6": "",
+      "p_7": "",
+      "p_8": "",
+      "p_9": ""
+    },
     "pre": "🌸 p5.js dice: {{message}}",
     "welcome": ""
   }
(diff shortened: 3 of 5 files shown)


lmccart:
<3 I have been dreaming about these particular updates for so long!

outofambit:
i'm so happy to see this expanding!!

src/core/internationalization.js
```diff
@@ -31,6 +36,7 @@ export const initialize = () =>
         nestingPrefix: '$tr(',
         nestingSuffix: ')',
         defaultNS: 'translation',
+        returnEmptyString: false,
```
stalgiag, line 39: I am a bit confused about the purpose of this.

translations/es/translation.json
```diff
@@ -13,7 +13,28 @@
       "table": "",
       "xml": ""
     },
+    "friendlyParamError": {
```
stalgiag, line 16: Maybe open an issue for a request for a spanish translation of these once this gets merged?

tasks/build/browserify.js
```diff
@@ -39,7 +39,9 @@ module.exports = function(grunt) {
         browseified = browseified
           .exclude('../../docs/reference/data.json')
           .exclude('../../../docs/parameterData.json')
-          .ignore('../../translations/index.js');
+          .exclude('../../translations')
+          .ignore('i18next')
+          .ignore('i18next-browser-languagedetector');
```
stalgiag, line 44: :-D

translations/es/translation.json
akshay-99: I was thinking of a general issue for all languages. But in #4233 @outofambit and @limzykenneth talked about how including translations as part of the main library file may increase the size quite a lot, and solving this by perhaps serving the translation files sepearately. Do we want this to be addressed first before people start contributing the translations?

tasks/build/browserify.js
akshay-99: I think there's some bug in the browserify `ignore()` method. It doesn't work for relative paths ( I tried using it and contents of the file that was supposed to be ignored were still there in the build ). `exclude()` works. 

Also, the path that is used here, must be exactly the same as that used in the import statements in the source. Which is also why it was `../../docs/` before but now it is `../../../docs` as the file importing from docs (`src/core/error_helpers.js`) became (`src/core/friendly_errors/validateParams.js`)

It took some trial and error to figure this out, but this combination seems to work :see_no_evil:

translations/es/translation.json
stalgiag: Right. We can address that first. I just want to make sure that an issue gets opened so we don't forget that the more general concern needs addressing.

PR #4643, opened 2020-06-18
FES: spelling and caps detection

### LINKED ISSUE #4637: FES: detect capitalization and spelling mistakes

- [x] Other (Friendly Errors)

I have seen that new programmers often take time to register the several naming conventions used in programming such as camelCase for identifier names, ALL_CAPS for constants, etc. And so capitalization mistakes are very common among them but they can spend way more time trying to debug it. For example one may call try to call `createcanvas` instead of `createCanvas`, `Color` instead of `color` and so on. These types of mistakes are easier to spot as the browser itself will show an error like `createcanvas` doesn't exist.
But mistakes in functions which are to be defined by the user such as `setup`, `draw`, `mouseClicked`, etc are harder to spot. For example
```js
function Setup() {
  createCanvas(400, 400);
  background(200);
} 
```
This code snippet won't run as `Setup` has been defined when we needed `setup`. But there will be no error message displayed here and so it can take longer for a beginner to figure it out.  Similarly 
```js
function setup() {
 ...
}
function mouseclicked(){
 ...
}
```
`mouseclicked` will not work but there is no error message.

Spelling mistakes can also be common for someone not  that familiar with English, or there can be regional differences in spelling ( `color` vs `colour` )

I am thinking of using the FES to display a message whenever the user has made a capitalization or spelling mistake and show the nearest match using word distance algorithms. Something like
` 🌸 p5.js says: It looks like you have used mouseclicked. Did you mean mouseClicked ?`
This won't run every time, but only when a spelling/capitalization mistake occurs. So there should be no negative effects on the performance as such. It will function similarly to `helpForMisusedAtTopLevel`

akshay-99:
Resolves #4637 

### Changes:
I have tried the simpler check using `toLowerCase()` as suggested by @limzykenneth .
Seems to work well to capture all kinds of mistakes in capitalization. For the sketch
```js
function Setup() {
  background(100);
}
```
and for the sketch
```js
let sketch = function(p) {
    p.Setup = () => {
        p.background(100);
    }
}
let myp5 = new p5(sketch);
```
We get
```
🌸 p5.js says: It seems that you may have accidently written Setup instead of setup.

Please correct it if it's not intentional. (http://p5js.org/reference/#/p5/setup)
```

And for
```js
function setup() {
  createcanvas(400, 400);
  background(100);
}
```
We get
```
🌸 p5.js says: It seems that you may have accidently written createcanvas instead of createCanvas (at http://localhost:8000/lib/empty-example/sketch.js:2:5).

Please correct it to createCanvas if you wish to use the function from p5.js (http://p5js.org/reference/#/p5/createCanvas)
```

Performance-wise the only concern was `checkForUserDefinedFunctions` as it would have to run the check regardless of whether there is an actual error or not. But it seems to run in under a millisecond and it will run only once, after p5 is initialized. `fesErrorMonitor` shouldn't be a problem for performance as it only runs when we have an actual error.

This only captures capitalization mistakes.
I will try to capture the ones for spelling now and also add tests

### Update: 
Now works for spelling mistakes too. For
```js
function setup() {
  Colour();
}
```
```
🌸 p5.js says: It seems that you may have accidently written Colour instead of color (at http://localhost:8000/lib/empty-example/sketch.js:11:3).

Please correct it to color if you wish to use the function from p5.js (http://p5js.org/reference/#/p5/color)
```

Also added a few tests
#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/friendly_errors/browser_errors.js
+++ b/src/core/friendly_errors/browser_errors.js
@@ -1,0 +1,15 @@
+// Browsers use error strings to build their error messages. These can be
+// different across different browsers. We can build a list of these strings
+// so that the FES is able to extract the required info from the errors given
+// by the browser
+const strings = {
+  ReferenceError: [
+    {
+      msg: '{{}} is not defined',
+      type: 'NOTDEFINED',
+      browser: 'all'
+    }
+  ]
+};
+
+export default strings;
--- a/src/core/friendly_errors/browser_errors.js
+++ b/src/core/friendly_errors/browser_errors.js
@@ -1,7 +1,9 @@
-// Browsers use error strings to build their error messages. These can be
-// different across different browsers. We can build a list of these strings
-// so that the FES is able to extract the required info from the errors given
-// by the browser
+// Different browsers may use different error strings for the same error.
+// Extracting info from them is much easier and cleaner if we have a predefined
+// lookup against which we try and match the errors obtained from the browser,
+// classify them into types and extract the required information. The contents
+// of this file serve as that lookup. The FES can use this to give a simplified
+// explanation for all kinds of errors.
 const strings = {
   ReferenceError: [
     {
--- a/src/core/friendly_errors/browser_errors.js
+++ b/src/core/friendly_errors/browser_errors.js
@@ -8,6 +8,11 @@ const strings = {
       msg: '{{}} is not defined',
       type: 'NOTDEFINED',
       browser: 'all'
+    },
+    {
+      msg: "Can't find variable: {{}}",
+      type: 'NOTDEFINED',
+      browser: 'Safari'
     }
   ]
 };
--- a/src/core/friendly_errors/fes_core.js
+++ b/src/core/friendly_errors/fes_core.js
@@ -55,6 +55,8 @@ if (typeof IS_MINIFIED !== 'undefined') {
 } else {
   let doFriendlyWelcome = false; // TEMP until we get it all working LM
 
+  const errorTable = require('./browser_errors').default;
+
   // -- Borrowed from jQuery 1.11.3 --
   const class2type = {};
   const toString = class2type.toString;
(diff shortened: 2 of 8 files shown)


akshay-99:
I guess this is ready for review. 

So as I mentioned in #4637, there are two types of checks:

- Type 1: 
Mistakes in referencing p5 functions such as `createCanvas`, `random`, `color`, etc. I have used case-insensitive Levenshtein distance to detect these. The threshold is set to 2 for now. i.e it can detect `CREATECANVAS`, `CrateCanvas`,`craeteCanvas`  etc. This is more of a "reactive" check in the sense that it reacts to ReferenceError messages, whenever they occur. It does not activate on its own.
Different browsers use different error strings. It is much easier and cleaner if we divide these into categories so that we can extract the required information about the various kinds of errors.
I have created a file `browser_errors.js`, in which we can add the various error strings used by browsers. We can later match error messages against this to give a simplified explanation for all kinds of errors. For now, I have just added two strings under ReferenceError. The first works for Chrome, Firefox, Edge and Opera. The second works for Safari.

- Type 2:
This is for user-defined functions such as `setup`, `draw`, `preload`, `mouseClicked`, etc. This is a more proactive check, in the sense, it activates without an error message. For this, I have set it up so that it can only detect capitalization mistakes: `SETUP`, `Draw`, 'preLoad', etc. Adding the check for spelling too here would just give a whole bunch of false positives. As the user is allowed to define their own functions which are not related to p5 ( perhaps draw1, myDraw, etc. We should not trigger a message in these cases. ) Also, other libraries may define functions which are close in name to `setup`, `draw`, etc and so showing a message here for misspelling would cause more confusion. So just a capitalization check is enough here.

Type 2 works for both global and instance mode. 
Type 1 works only for global mode. This is because, in instance mode, we won't have a `ReferenceError`. We could have a `TypeError` ( for lets say `p.createcanvas()` ) or no error at all. ( for eg `p.mousey` )
I thought about setting up a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) object, which can detect when a `get()` resolves to `undefined`, and pass that proxy to the sketch instead of p5. But this might be a bit too much and may break a few things. Also, older browsers and IE don't support Proxy can according to [this page](https://babeljs.io/docs/en/learn#proxies) in the Babel docs, it can't be polyfilled either. So I decided to leave and not do that.

src/core/friendly_errors/browser_errors.js
```diff
@@ -0,0 +1,20 @@
+// Browsers use error strings to build their error messages. These can be
+// different across different browsers. We can build a list of these strings
+// so that the FES is able to extract the required info from the errors given
+// by the browser
+const strings = {
+  ReferenceError: [
+    {
+      msg: '{{}} is not defined',
+      type: 'NOTDEFINED',
+      browser: 'all'
```
stalgiag, line 10: why `all` instead of 'Chrome, Firefox, Edge and Opera' as in the comment?

src/core/friendly_errors/browser_errors.js
```diff
@@ -0,0 +1,20 @@
+// Browsers use error strings to build their error messages. These can be
+// different across different browsers. We can build a list of these strings
+// so that the FES is able to extract the required info from the errors given
+// by the browser
```
stalgiag, line 4: I have a bit of trouble comprehending this comment as written. I found the explanation of this file in [this comment](https://github.com/processing/p5.js/pull/4643#issuecomment-647203541) much more clear. Perhaps a rewrite or additional explanation would be helpful.

src/core/friendly_errors/fes_core.js
```diff
@@ -47,12 +47,20 @@ import { translator } from '../internationalization';
 // p5.js blue, p5.js orange, auto dark green; fallback p5.js darkened magenta
 // See testColors below for all the color codes and names
 const typeColors = ['#2D7BB6', '#EE9900', '#4DB200', '#C83C00'];
+let misusedAtTopLevelCode = null;
+let defineMisusedAtTopLevelCode = null;
+
+// the threshold for the maximum allowed levenshtein distance
+// used in misspelling detection
+const EDIT_THRESH = 2;
```
stalgiag, line 55: Maybe this is a levenshtein-specific term but I am thrown off by the word 'edit' here and on L179 `editDistance()`. I  associate 'edit' with writing not reading.

stalgiag: Ah I see now that 'edit distance' is a [technical term you are referencing](https://en.wikipedia.org/wiki/Edit_distance). This is misleading without that knowledge; could potentially be `EDIT_DIST_THRESHOLD`.

src/core/friendly_errors/fes_core.js
```diff
@@ -168,6 +176,200 @@ if (typeof IS_MINIFIED !== 'undefined') {
     console.log(translator('fes.pre', { message }));
   };
 
+  const editDistance = (w1, w2) => {
```
stalgiag, line 179: See above comment -- maybe `computeEditDistance` would be better so that 'edit' doesn't read as the verb.

src/core/friendly_errors/fes_core.js
```diff
@@ -168,6 +176,200 @@ if (typeof IS_MINIFIED !== 'undefined') {
...
+    // An implementation of
+    // https://en.wikipedia.org/wiki/Wagner%E2%80%93Fischer_algorithm
+    // to compute the Levenshtein distance
```
stalgiag, line 182: Maybe just a line to say what this is useful for -- something like 'to detect how dissimilar two strings are'.

src/core/friendly_errors/fes_core.js
akshay-99: Yeah makes sense. I will change the name

src/core/friendly_errors/fes_core.js
akshay-99: Fair enough. I will change this too

src/core/friendly_errors/browser_errors.js
akshay-99: The idea was that there will be many browsers that use one particular string and a few that decide to go their own way by having their unique string.
But sometimes it can also happen that a browser will use the common string for some sub-cases of the same error, and use a unique string for other sub-cases.  For e.g. 
for [this error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cant_access_lexical_declaration_before_init), firefox has a different string, but it also uses `x is not defined` for other error-cases

So as the file grows over the next stages of the project, my idea was to have all browsers first try matching the error of `browser: all` and then if nothing matches there, it will try to look if there is a string marked with its name. The `all` here indicates that all browsers should try if this string matches. Maybe it could be different.

src/core/friendly_errors/fes_core.js
akshay-99: Yes I will add that

src/core/friendly_errors/browser_errors.js
akshay-99: Okay

akshay-99:
Also I just noticed that the error handler is not triggered when `p5.sound.min.js` is in the html file. I am looking into that

akshay-99:
@stalgiag I have made the changes as you suggested. Please let me know your views on [this comment](https://github.com/processing/p5.js/pull/4643#discussion_r443968297). If need be, I will change the `browser: all` thing too.

I have also fixed the problem of `fesErrorMonitor` not running when `p5.sound.min.js` is imported. Although I don't think I completely understand the reason I am presuming it must be because I had only set it to run when [this promise](https://github.com/processing/p5.js/blob/main/src/core/init.js#L57) fails ( an error in the sketch would make this happen ).

But when p5.sound or some other library is present, an error in the sketch would result in [this other promise](https://github.com/processing/p5.js/blob/main/lib/addons/p5.sound.js#L3211) failing. 

So I changed it to instead listen for any PromiseRejectionEvents or any ErrorEvents [here](https://github.com/processing/p5.js/blob/24171b12495ca06ae1f3363722036211f3a5c435/src/core/friendly_errors/fes_core.js#L383)

src/core/friendly_errors/browser_errors.js
stalgiag: Ah that makes sense!

PR #4644, opened 2020-06-18
Append canvas to main

lm-n:
Changes:
Appends canvas to the &lt;main&gt; element. This is important because the &lt;main&gt; element behaves like a main landmark role and can be used by assistive technology to quickly identify and navigate to the dominant part of the body. &lt;main&gt;  doesn't contribute to the document's outline and doesn't affect the DOM's concept of the structure of the page. 

If &lt;main&gt;  does not yet exist in the html, it creates element &lt;main&gt;  in &lt;body&gt;  before appending canvas.

This change doesn't affect user input node cases. 

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/lib/empty-example/index.html
+++ b/lib/empty-example/index.html
@@ -17,6 +17,8 @@
 </head>
 
 <body>
+  <main>
+  </main>
 </body>
 
 </html>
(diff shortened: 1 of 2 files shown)


kjhollen:
Also, do you think it's worth updating the template on https://p5js.org/get-started/ ? I'm not sure!

PR #4654, opened 2020-06-23
Add describe() and describeElement()

lm-n:
Changes:

- create folder in source for accessibility features
- src/accessibility/describe.js adds two functions describe() and describeElement() to the p5.js library that make adding fallback descriptions of the canvas easy for p5.js users. Descriptions generated with these functions can be useful for screen reader users and for search engines.
  - [describe() function](https://www.dsk.lm-n.xyz/p5-describe/describe/index.html)
  - [describeElement() function](https://www.dsk.lm-n.xyz/p5-describe/describeElement/index.html)
- in app.js import './accessibility/describe.js'
- in constants adds constants for functions describe() and describeElement()

#### PR Checklist

- [x] `npm run lint` passes

Pending: 
- Styling of examples in reference. Currently the examples look ugly. @kjhollen and I looked into it but unsure of how to proceed/fix them. @lmccart perhaps you have some ideas of how we could approach this.
Examples of how to use describe()


src/accessibility/describe.js
```diff
@@ -0,0 +1,259 @@
...
+      cnvId +
+      '_Description" role="region" aria-label="Canvas Description"><p id="' +
+      cnvId +
+      '_dsc"></p></div>';
+  } else if (document.getElementById(cnvId + '_dsc') === null) {
+    document
+      .getElementById(cnvId + '_eDsc')
+      .insertAdjacentHTML('beforebegin', '<p id="' + cnvId + '_dsc"></p>');
+  }
+  if (document.getElementById(cnvId + '_dsc').innerHTML !== t) {
```
kjhollen, line 82: It seems safe to just assign without checking, but does this keep it from notifying the screen reader that the value has been updated? (asking for clarity, for me) :)

src/accessibility/describe.js
```diff
@@ -0,0 +1,259 @@
...
+    document.getElementById(cnvId + '_dsc').innerHTML = t;
+  }
+  //If display is LABEL creates a div adjacent to the canvas element with
+  //description text.
+  if (d === 'label') {
```
kjhollen, line 87: use p5 constant LABEL here

src/accessibility/describe.js
```diff
@@ -0,0 +1,259 @@
...
+ * </div>
+ *
+ */
+p5.prototype.describe = function(t, d) {
+  p5._validateParameters('describe', arguments);
+  const cnvId = this.canvas.id;
+  //Creates a sub DOM inside of the canvas element and populates
+  //it with description text.
+  t = this._descriptionText(t);
+  if (document.getElementById(cnvId + '_Description') === null) {
```
kjhollen, line 70: should _Description, _dsc, and other suffixes you've defined here be constants? Might prevent errors in the future if other features are added that depend on these ids.

src/accessibility/describe.js
```diff
@@ -0,0 +1,259 @@
...
+};
+
+/**
+ * Helper function for describe() and describeElement().
+ */
+p5.prototype._descriptionText = function(t) {
+  if (t.endsWith('.') === false) {
+    t = t + '.';
+  }
+  if (/^[A-Z]/.test(t) === false) {
```
kjhollen, line 117: I can never remember regex syntax, so a comment here would be helpful :)

src/accessibility/describe.js
```diff
@@ -0,0 +1,259 @@
...
+ * @module Environment
+ * @submodule Environment
+ * @for p5
+ * @requires core
+ */
+
+import p5 from '../core/main';
+
+/**
+ * Creates a screen-reader accessible description for the canvas in the DOM.
```
kjhollen, line 11: Thinking of a new coder learning about this function. Maybe the part about the DOM can go at the end? I think a new programmer could still use this function without knowing about the DOM and the acronym might be intimidating. I think it's important enough to mention here, but maybe it can go at the end, like it's supplemental info.

src/accessibility/describe.js
```diff
@@ -0,0 +1,259 @@
...
+  if (/^[A-Z]/.test(t) === false) {
+    t = t[0].toUpperCase() + t.slice(1);
+  }
+  return t;
+};
+
+/**
+ * This function creates a screen-reader accessible
+ * description for elements —shapes or groups of shapes that create
+ * meaning together— in the canvas sub DOM. The first paramater should
```
kjhollen, line 126: same comment here as above: suggest moving detail about the canvas sub DOM to the end of the description.

src/accessibility/describe.js
```diff
@@ -0,0 +1,259 @@
...
+ * @param  {Constant} [display] either LABEL or FALLBACK (Optional)
+ *
+ * @example
+ * <div>
+ * <code>
+ * describe('pink square with red heart in the bottom right corner', LABEL);
+ * background('pink');
+ * fill('red');
+ * noStroke();
+ * ellipse(66.6, 66.6, 20, 20);
```
kjhollen, line 41: this is a nitpick, but I think it's ok to just round these numbers for simplicity

src/accessibility/describe.js
```diff
@@ -0,0 +1,259 @@
...
+    document.getElementById(cnvId + n).innerHTML !==
+    '<th scope="row">' + n + '</th><td>' + t + '</td>'
+  ) {
+    document.getElementById(cnvId + n).innerHTML =
+      '<th scope="row">' + n + '</th><td>' + t + '</td>';
+  }
+  //If display is LABEL creates a div adjacent to the canvas element with
+  //a table, a row header cell with the name of the elements,
+  //and adds the description of the element in adjecent cell.
+  if (d === 'label') {
```
kjhollen, line 208: use p5 constant here

src/accessibility/describe.js
```diff
@@ -0,0 +1,259 @@
...
+      document.getElementById(cnvId + n + 'Lbl').innerHTML !==
+      '<th scope="row">' + n + '</th><td>' + t + '</td>'
+    ) {
+      document.getElementById(cnvId + n + 'Lbl').innerHTML =
+        '<th scope="row">' + n + '</th><td>' + t + '</td>';
+    }
+  }
+};
+/**
+ * Helper function for describeElement().
```
kjhollen, line 244: a short description of what updates are being made to the text would be helpful here

src/accessibility/describe.js
```diff
@@ -0,0 +1,300 @@
...
+
+import p5 from '../core/main';
+const ds = '_Description'; //Fallback container
+const fds = '_fds'; //Fallback description
+const ft = '_ft'; //Fallback Table
+const fte = '_fte_'; //Fallback Table Element
+const lb = '_Label'; //Label container
+const lds = '_lds'; //Label description
+const lt = '_lt'; //Label Table
+const lte = '_lte_'; //Lable Table Element
```
kjhollen, line 16: great! one small typo in this comment (Lable)

test/unit/accessibility/describe.js
```diff
@@ -0,0 +1,140 @@
...
+    }
+    p5Container = null;
+  });
+
+  const getInner = elID => {
+    return document.getElementById(elID).innerHTML;
+  };
+
+  suite('p5.prototype.describe', function() {
+    let expected = 'A.';
```
kjhollen, line 35: the describe tests all looks great! another good test to add would be to make sure a period isn't added if the supplied string already has one. additionally, what should the outcome be if the user supplies a string that ends in ! or ?

test/unit/accessibility/describe.js
```diff
@@ -0,0 +1,140 @@
...
+          myp5.describe(myp5.LABEL);
+        },
+        Error,
+        'description should not be LABEL or FALLBACK'
+      );
+    });
+  });
+
+  suite('p5.prototype.describeElement', function() {
+    let expected = '<th scope="row">A:</th><td>B.</td>';
```
kjhollen, line 84: Great! as above: add a test case where the : and . are already present, and any other edge cases that matter.

test/unit/accessibility/describe.js
lm-n: I've added an exception for ! and ? in regex function

lm-n:
> Hi Luis, thanks for your patience while I've been moving! The updates and added tests look great. I found one minor typo and suggest adding a couple more test cases (see comments inline), then this will be ready to go. Great work!

Hi Kate! No worries! Just added the test cases and "!" "?" exceptions in the regex. 😸

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
...
+
+p5.prototype._descriptionText = function(t) {
+  if (t === this.LABEL || t === this.FALLBACK) {
+    throw new Error('description should not be LABEL or FALLBACK');
+  }
+  //if string does not end with '.'
+  if (
+    t.endsWith('.') === false &&
+    t.endsWith('?') === false &&
+    t.endsWith('!') === false
```
lmccart, line 138: these can be simplified to `!t.endsWith('.')` etc

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
...
+p5.prototype.describe = function(t, d) {
+  const cnvId = this.canvas.id;
+  p5._validateParameters('describe', arguments);
+  //Creates a sub DOM inside of the canvas element and populates
+  //it with description text.
+  if (typeof t !== 'string') {
+    return;
+  }
+  t = this._descriptionText(t);
+  if (document.getElementById(cnvId + ds) === null) {
```
lmccart, line 82: simplify to:
`if (!document.getElementById(cnvId + ds)) {`

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
...
+      '" role="region" aria-label="Canvas Description"><p id="' +
+      cnvId +
+      fds +
+      '"></p></div>';
+  } else if (document.getElementById(cnvId + fds) === null) {
+    document
+      .getElementById(cnvId + ft)
+      .insertAdjacentHTML('beforebegin', '<p id="' + cnvId + fds + '"></p>');
+  }
+  if (document.getElementById(cnvId + fds).innerHTML !== t) {
```
lmccart, line 96: i'm not sure this test is necessary. could you just set the `innerHTML`? or is there a performance advantage to not reseting if it's already set?

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
...
+      document.getElementById(cnvId + lds).innerHTML = t;
+    }
+  }
+};
+
+/**
+ * Helper function for describe() and describeElement().
+ */
+
+p5.prototype._descriptionText = function(t) {
```
lmccart, line 130: if this is only used internally, it doesn't need to be attached to `p5.prototype`

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
...
+ * describeElement('Heart', 'red heart in the bottom right corner', LABEL);
+ * fill('red');
+ * ellipse(66.6, 66.6, 20, 20);
+ * ellipse(83.2, 66.6, 20, 20);
+ * triangle(91.2, 72.6, 75, 95, 58.6, 72.6);
+ * </code>
+ * </div>
+ */
+
+p5.prototype.describeElement = function(n, t, d) {
```
lmccart, line 196: for readability, i would suggest spelling out name, text, and display

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
...
+  p5._validateParameters('describeElement', arguments);
+  if (typeof t !== 'string' || typeof n !== 'string') {
+    return;
+  }
+  const cnvId = this.canvas.id;
+  t = this._descriptionText(t);
+  let nm = this._elementName(n);
```
lmccart, line 203: for readability, i would suggest spelling out `elementName`

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
...
+          '<table id="' + cnvId + lt + '"></table>'
+        );
+    }
+    if (document.getElementById(cnvId + lte + n) === null) {
+      let tr = document.createElement('tr');
+      tr.id = cnvId + lte + n;
+      document.getElementById(cnvId + lt).appendChild(tr);
+    }
+
+    if (
```
lmccart, line 270: not sure this check is necessary

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
+/**
+ * @module Environment
+ * @submodule Environment
+ * @for p5
+ * @requires core
+ */
+
+import p5 from '../core/main';
+const ds = '_Description'; //Fallback container
```
lmccart, line 9: for readability it may help to write these out

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
...
+  //check if last character of string n is '.', ';', or ','
+  if (lm === '.' || lm === ';' || lm === ',') {
+    //replace last character with ':'
+    n = n.replace(/.$/, ':');
+    //if string n does not end with ':'
+  } else if (n.endsWith(':') === false) {
+    //add ':'' at the end of string
+    n = n + ':';
+  }
+  //if first character of string is not capitalized
```
lmccart, line 296: does capitalization matter if this is for screen reader? pardon my lack of knowledge about this

src/accessibility/describe.js
```diff
@@ -0,0 +1,304 @@
...
+    ) {
+      document.getElementById(cnvId + lte + n).innerHTML =
+        '<th scope="row">' + nm + '</th><td>' + t + '</td>';
+    }
+  }
+};
+/**
+ * Helper function for describeElement().
+ */
+p5.prototype._elementName = function(n) {
```
lmccart, line 282: if this is only used internally, it doesn't need to be attached to `p5.prototype`

src/accessibility/describe.js
lm-n: If an html element is constantly being reset some screenreaders will read the html element as empty and will not be able to access its content. This why we are reseting only when the content is different.

lmccart: ah! interesting, makes sense. thanks for explaining

src/accessibility/describe.js
lm-n: it doesn't really matter for something like this. after thinking about it I've decided to leave it up to the user and deleted these regex lines.

lm-n:
@lmccart thx for the feedback! I've made all of the suggested changes! 😺 

@kjhollen I still got a couple of new issues with tests —that came up with the new changes — to fix before merge, will comment on them in case you get a chance to look at them. I should be able to work through them before our next meeting.

test/unit/accessibility/describe.js
```diff
@@ -0,0 +1,171 @@
...
+    test('should create description as fallback', function() {
+      myp5.describe(a);
+      let actual = getInner(myID + '_fallbackDesc');
+      assert.deepEqual(actual, expected);
+    });
+    /*test('should not add extra period if string ends in "."', function() {
+      myp5.describe('a.');
+      let actual = getInner(myID + '_fallbackDesc');
+      assert.deepEqual(actual, expected);
+    });*/
```
lm-n, line 49: @kjhollen this is one of the tests that are not working. I've commented out for now. when I try doing what the test does on a p5 sketch it doesn't give me any errors so I have to think through it a bit and figure out what's going on. There are 3 more tests where that I've commented out because, like this one, they keep giving me unexpected errors. If you have any thoughts lmk 😅

kjhollen: ah shoot! ok, I'm checking out the code now and will let you know if the behavior is the same for me.

kjhollen: Ok, I've got a theory. I think it has to do with the way the template strings are being compiled down to ES5 with babel: it's including the indents and linebreaks. I think this is causing a mismatch sometimes.

More here: https://medium.com/dailyjs/how-to-create-multi-line-string-with-template-literals-in-javascript-a3a140d0b0f6

I think you can use a back slash to fix it at the end of each line: https://stackoverflow.com/questions/37321047/wrap-long-template-literal-line-to-multiline-without-creating-a-new-line-in-the

(trying this out now on my end!)

kjhollen: ok, more complex than it looks! I can't get it to work with the minified p5.js file, but the regular one works fine. hmm. any thoughts @lmccart?

akshay-99:
Hey @lm-n! So I took a quick look at the tests ( I hope you don't mind :sweat_smile: ). 
My understanding here is quite limited but I think they are failing at the moment because the `dummy` variable in `describe.js` still remains populated after a test completes. So what I saw was that the first test that called `describe()` ran perfectly but subsequent tests didn't. 

That's probably because even though the `p` tag which was created during a test is deleted in `teardown()`, the `dummy` variable is not cleared. So we won't get a new `p` tag for the next test. And thus the error ` Uncaught Error: Uncaught TypeError: Cannot read property 'innerHTML' of null (http://localhost:9001/test/unit/accessibility/describe.js:45)`. 

If we have a new internal function as 
```js
p5.prototype._clearDummy = function() {
  dummy = { fallbackElements: {}, labelElements: {} };
}
```

and call this in `teardown()`, it should fix this problem.

Also, I didn't see the need for using `setTimeout()` for testing as all of the code in `describe.js` is synchronous.

lm-n:
Thanks @kjhollen  and @akshay-99 ❤️  ❤️ ❤️ ! Akshay you were right! We  had some problems with dummy beyond what you identified. It should work now 👯‍♀️ 👯‍♂️ !

src/accessibility/describe.js
```diff
@@ -235,131 +170,275 @@ p5.prototype.describeElement = function(name, text, display) {
...
+/*
+ *
+ * Helper functions for describe() and describeElement().
+ *
  */
-function _descElementLabelHTML(cnvId, name, inner) {
-  if (!dummy[cnvId + labelContainer]) {
+
+//clear dummy
+p5.prototype._clearDummy = function() {
```
kjhollen, line 230: I'm curious why this method is attached to the prototype but not the fallback elements or label elements. Wondering if it makes sense for fallbackElements and labelElements to be attached to a p5 instance? (I recall that this got added to fix the tests, am wondering if attaching to the instance would also fix that or if there are other possible advantages)

kjhollen:
hey @lm-n I was double-checking this with a merge from `main` branch locally on my machine and hit some snags (that do not appear to be in your code). I repeatedly get an error that says:

`No parser and no filepath given, using 'babel' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.`

Then some of the examples (in p5.Font.js, p5.Image.js, vertex.js, image.js, pixels.js, dom.js, files.js, and p5.Graphics.js) seem to have trouble with the linter (prettier) in the task `eslint-samples:source` build. It looks like the linter now wants a bunch of parenthesis around some math. Wondering if anyone else can advise on what's happening?

I'm not sure if it's safe to merge so I'm going to hold off for now! But knowing that you've got newer work ready to go / review, I'm hoping we can still get this resolved quickly.

lmccart:
@kjhollen all good to merge by me once these last little issues get resolved

kjhollen:
ok, I realized that I had locally run `npm update` because I saw:

``Browserslist: caniuse-lite is outdated. Please run next command `npm update` ``

which updated my package-lock.json and package.json files to versions beyond what the project expected, and caused the issues I wrote about yesterday. I figured this out by doing a clean checkout of the main branch, where everything worked ok until I ran `npm update`, when I also noticed the additional files had been changed. (Posting the solution here in case it's useful to someone later, but my mistake 😅)

Anyway, it was a clean build once I figured this out and all tests pass, so I'm going to merge!

PR #4673, opened 2020-07-03
Decode HTML entities in the documentation file (parameterData.json)

yukienomiya:
**Changes:**
Modified the renderItemDescriptionsAsMarkdown() function in the preprocessor.js so the HTML entities (like "&amp;" for "&") in the class/method/field descriptions are decoded. I used the node-html-entities library to decode the text.

The documentation file (data.json in the p5.js-website repository) is used to create the Reference translation files, so I'm trying to make it as readable as possible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "grunt-mocha-test": "^0.13.3",
     "grunt-newer": "^1.1.0",
     "grunt-simple-nyc": "^3.0.1",
+    "html-entities": "^1.3.1",
     "husky": "^4.2.3",
     "i18next": "^19.0.2",
     "i18next-browser-languagedetector": "^4.0.1",
@@ -145,9 +146,7 @@
     "not dead"
   ],
   "author": "",
-  "dependencies": {
-    "html-entities": "^1.3.1"
-  },
+  "dependencies": {},
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged"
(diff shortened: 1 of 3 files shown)


limzykenneth:
Looks good, just one thing, can the `html-entities` dependency be moved to dev dependencies?

On another note, I've been looking into potential replacement for yuidocs and am trying it out in one of my projects. It is a considerable change but if it works out on my project, it may be viable to be explored for p5.js as well.

package.json
```diff
@@ -145,7 +145,9 @@
     "not dead"
   ],
   "author": "",
-  "dependencies": {},
+  "dependencies": {
+    "html-entities": "^1.3.1"
```
outofambit, line 149: @yukienomiya could we move this to `devDevependencies` please?
(thanks @limzykenneth for noticing this!)

outofambit:
> On another note, I've been looking into potential replacement for yuidocs and am trying it out in one of my projects.

@limzykenneth exciting! out of curiosity, what's the tool called?

limzykenneth:
@outofambit It's [comment-parser](https://github.com/syavorsky/comment-parser), a JSDoc/YUIDoc style comment parser that just outputs the various comment blocks and the associated tags in the block as a javascript object. With a bit of extra work, we could potentially parse the output to fit what we need and repalce `data.json` with it.

package.json
yukienomiya: Done! Thanks @limzykenneth :)

PR #4699, opened 2020-07-19
Adding more checks to StoreItem()

### LINKED ISSUE #4697: storeItem() - prevent the overwrite of a key type

- [ ] Color
- [x] Data
- [ ] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

When a key is stored, its type is also stored whose name is the initial key name followed by `"p5TypeID"`.
If a key named `"someText"` is stored, the data become unreadable if someone store a key as `"someTextp5TypeID"`. Here is an example :
``` javascript
  storeItem("boolFalse", false);
  storeItem("boolFalsep5TypeID", "number");
  print(getItem("boolFalse")); // Output : NaN, Expected : false
```

This error is not likely to happen, but it may be great if we prevent the user to store key name which ends by "p5TypeID".

I can do it, but maybe someone has a better idea about how to prevent such issue (formatting key names, ...).

Cheers

VincentHch:
Resolves #4697 

Hi, 

As explained in the issue, I added another check to the function to avoid data corruption.
I also replaced the `console.log` by `console.error`, it seems a bit more logical to me.
#### PR Checklist

- [X] `npm run lint` passes

### DIFF
--- a/src/data/local_storage.js
+++ b/src/data/local_storage.js
@@ -59,18 +59,18 @@ import p5 from '../core/main';
  */
 p5.prototype.storeItem = function(key, value) {
   if (typeof key !== 'string') {
-    console.error(
+    console.log(
       `The argument that you passed to storeItem() - ${key} is not a string.`
     );
   }
   if (key.endsWith('p5TypeID')) {
-    console.error(
+    console.log(
       `The argument that you passed to storeItem() - ${key} must not end with 'p5TypeID'.`
     );
   }
 
   if (typeof value === 'undefined') {
-    console.error('You cannot store undefined variables using storeItem().');
+    console.log('You cannot store undefined variables using storeItem().');
   }
   let type = typeof value;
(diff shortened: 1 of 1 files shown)


src/data/local_storage.js
```diff
@@ -133,7 +144,7 @@ p5.prototype.getItem = function(key) {
   let value = localStorage.getItem(key);
   const type = localStorage.getItem(`${key}p5TypeID`);
   if (typeof type === 'undefined') {
-    console.log(
+    console.error(
```
lmccart, line 147: we typically avoid using `console.error` so that the code can keep running with a warning in the console

VincentHch: Corrected, thanks!

lmccart: looks great!

PR #4701, opened 2020-07-20
FES + i18n: Load translations from a CDN

### LINKED ISSUE #4694: FES + i18n: Loading translation files from CDN

- [x] Friendly Errors
- [x] Internationalization 

Internationalization was introduced to the FES in #4233. So far we only have English strings for the FES and even [that file](https://github.com/processing/p5.js/blob/b52226962eebf9f58f70eed68c4ce32661677d23/translations/en/translation.json) is starting to get pretty big, and could grow even bigger if more functionality is added or when translations are added for all of [these](https://github.com/processing/p5.js/search?q=console.log+inpath%3Asrc+--filename%3Afes_core.js+-filename%3Areference.js+-filename%3AREADME.md&unscoped_q=console.log+inpath%3Asrc+--filename%3Afes_core.js+-filename%3Areference.js+-filename%3AREADME.md) and [these](https://github.com/processing/p5.js/search?q=throw+new+Error+inpath%3Asrc&unscoped_q=throw+new+Error+inpath%3Asrc) messages.

In the future, when we have translations for more and more languages, all of this may take up a lot of space within the library. 

As @outofambit and @limzykenneth  discussed in #4233, we can host the translation files on a CDN and then fetch whichever file is needed depending on the language detected by `i18next-browser-languageDetector`. The English translation will serve as a fallback in case something goes wrong while fetching from the CDN. Another option could be to package translations into the zip, with instructions on how to load one. 

I am more inclined towards the first option as it does not require extra work and understanding on the user's part. Maybe we can implement the second option too, to have translated messages when running offline.
Translations could be updated with new releases of p5.

I can work on updating the internationalization code, the build process, the release process to implement this.
We should also have another issue to let new contributors know that we are looking for help in adding translations for all languages. :smile:

akshay-99:
Resolves #4694

 Changes:
- [x] Publish translation files to a CDN with every release. I have set it up so that the contents under the path `translations/**` get published to NPM. From there, jsDelivr can cache it to their CDN.
I published a namespaced version to test this out and it seems to work
https://cdn.jsdelivr.net/npm/@akshay-99/p5-test@1.0.1/translations/es/translation.json
https://cdn.jsdelivr.net/npm/@akshay-99/p5-test@1.0.1/translations/en/translation.json

These changes would take effect from the next release. Before then, the configured path ( https://cdn.jsdelivr.net/npm/p5/translations/ ) would give us a 404. This wouldn't be a problem for contributors because of the next point below.

- [x] Use local translation files in `dev` mode. Whenever someone is adding/updating something, they should be able to test it out. Fetching from the CDN will not be helpful as these changes wouldn't have been published yet. I configured grunt and browserify to pack all the translation files in the library when running `browserify:dev` ( i.e with `npm run dev` but not with `npm run build` )

Also, since we have updated our default branch name to `main`, we may have to change something with `np` to let it know this. Otherwise it complains that we aren't on the correct branch and would stop the release. For testing out this PR, I just manually passed in `--any-branch`. But there must be some other way to let `np` know the name of the default branch. 

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/tasks/build/browserify.js
+++ b/tasks/build/browserify.js
@@ -19,6 +19,8 @@ module.exports = function(grunt) {
     function(param) {
       const isMin = param === 'min';
       const isTest = param === 'test';
+      const isDev = param === 'dev';
+
       const filename = isMin
         ? 'p5.pre-min.js'
         : isTest ? 'p5-test.js' : 'p5.js';
@@ -32,9 +34,14 @@ module.exports = function(grunt) {
       // Render the banner for the top of the file
       const banner = grunt.template.process(bannerTemplate);
 
+      let globalsVars = {};
+      if (isDev) {
+        globalsVars['P5_DEV_BUILD'] = () => true;
+      }
       // Invoke Browserify programatically to bundle the code
       let browseified = browserify(srcFilePath, {
-        standalone: 'p5'
+        standalone: 'p5',
+        insertGlobalVars: globalsVars
       });
 
       if (isMin) {
@@ -48,6 +55,10 @@ module.exports = function(grunt) {
           .exclude('./browser_errors')
           .ignore('i18next')
           .ignore('i18next-browser-languagedetector');
+      }
+
+      if (!isDev) {
+        browseified = browseified.exclude('../../translations/dev');
       }
 
       const babelifyOpts = { plugins: ['static-fs'] };
--- a/tasks/build/browserify.js
+++ b/tasks/build/browserify.js
@@ -34,21 +34,21 @@ module.exports = function(grunt) {
       // Render the banner for the top of the file
       const banner = grunt.template.process(bannerTemplate);
 
-      let globalsVars = {};
+      let globalVars = {};
       if (isDev) {
-        globalsVars['P5_DEV_BUILD'] = () => true;
+        globalVars['P5_DEV_BUILD'] = () => true;
       }
       // Invoke Browserify programatically to bundle the code
-      let browseified = browserify(srcFilePath, {
+      let browserified = browserify(srcFilePath, {
         standalone: 'p5',
-        insertGlobalVars: globalsVars
+        insertGlobalVars: globalVars
       });
 
       if (isMin) {
         // These paths should be the exact same as what are used in the import
         // statements in the source. They are not relative to this file. It's
         // just how browserify works apparently.
-        browseified = browseified
+        browserified = browserified
           .exclude('../../docs/reference/data.json')
           .exclude('../../../docs/parameterData.json')
           .exclude('../../translations')
@@ -58,7 +58,7 @@ module.exports = function(grunt) {
       }
 
       if (!isDev) {
-        browseified = browseified.exclude('../../translations/dev');
+        browserified = browserified.exclude('../../translations/dev');
       }
 
       const babelifyOpts = { plugins: ['static-fs'] };
@@ -67,7 +67,7 @@ module.exports = function(grunt) {
         babelifyOpts.envName = 'test';
       }
 
-      const bundle = browseified.transform('babelify', babelifyOpts).bundle();
+      const bundle = browserified.transform('babelify', babelifyOpts).bundle();
 
       // Start the generated output with the banner comment,
       let code = banner + '\n';
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -189,7 +189,7 @@ module.exports = grunt => {
           'src/**/*.vert',
           'src/**/*.glsl'
         ],
-        tasks: ['browserify'],
+        tasks: ['browserify:dev'],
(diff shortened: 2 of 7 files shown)


limzykenneth:
It is possible to publish the files to cdnjs as well, probably just need to make a PR to edit [this file](https://github.com/cdnjs/packages/blob/master/packages/p/p5.js.json).

A fix for `np` might be coming soon: https://github.com/sindresorhus/np/pull/558. In the meantime to release from main, `--any-branch` should be used for now.

akshay-99:
> It is possible to publish the files to cdnjs as well, probably just need to make a PR to edit [this file](https://github.com/cdnjs/packages/blob/master/packages/p/p5.js.json).

Yes. I will look into it.

lmccart:
This looks great. Could we add some link in the `contributor_docs/` to the files that are necessary to update when adding another language? (And any other relevant links) I think it would go in the [internationalization.md](https://github.com/processing/p5.js/blob/main/contributor_docs/internationalization.md) file.

akshay-99:
Hi. Yes. I will add that too.

tasks/build/browserify.js
```diff
@@ -45,6 +52,10 @@ module.exports = function(grunt) {
           .ignore('i18next-browser-languagedetector');
       }
 
+      if (!isDev) {
+        browseified = browseified.exclude('../../translations/dev');
```
stalgiag, line 56: This is nit-picky as it wasn't you that first wrote the variable but since you are interacting with it can you change this variable to `browserified`? The missing r in browser feels like a typo.

tasks/build/browserify.js
```diff
@@ -30,9 +32,14 @@ module.exports = function(grunt) {
       // Render the banner for the top of the file
       const banner = grunt.template.process(bannerTemplate);
 
+      let globalsVars = {};
```
stalgiag, line 35: `globalsVars` or `globals` or `globalVars`?

akshay-99: Ah. Should be `globalVars`, based on the browserify option `insertGlobalVars`. I'll correct that.

tasks/build/browserify.js
akshay-99: Yes, that stuck out a bit for me too. I had initially assumed that it might be some accepted nomenclature for some reason, but seems like this is [the only repo](https://www.google.com/search?q=%22browseified%22) that uses it. So it must be a typo. I'll change that to `browserify`

akshay-99:
> I am working with the assumption that we will have some time to look at this in more depth later to make sure that it is easy enough for others to maintain and build upon.

One part that might be a bit difficult to maintain is [this](https://github.com/processing/p5.js/blob/d3813a31f8ed4c0426cd977cc371c7c5eb528c4f/src/core/internationalization.js#L34). There are already several i18next plugins available that do something similar like [i18next-http-backend](https://github.com/i18next/i18next-http-backend), [i18next-fetch-backend](https://github.com/dotcore64/i18next-fetch-backend). But there were two things which I didn't find a way to implement using these: quickly skip a language if it doesn't exist in the list of supported languages ( don't send a request ), and timeout a request if it takes too much time.

Since we are stalling the initialization waiting for the translations to finish downloading, it may be important that this step doesn't take too much time as well ensure that it doesn't spend time trying to find files for languages which we know aren't supported yet. This may be important when there is a chain of unsupported languages. For example, if someone's language is 
`fr-CA` ( Canadian French ), i18next will try to find `fr-CA` and since that isn't supported yet, try looking up `fr`.  Since that too isn't supported, it finally comes to English. But it just spent time on two requests, which is easily prevented by just checking against a list of supported languages.

akshay-99:
I think adding a language itself should be easy for contributors since ( apart from the actual translation itself ) it only involves adding two extra lines in the code, [here](https://github.com/processing/p5.js/blob/d3813a31f8ed4c0426cd977cc371c7c5eb528c4f/translations/index.js#L26) and [here](https://github.com/processing/p5.js/blob/d3813a31f8ed4c0426cd977cc371c7c5eb528c4f/translations/dev.js#L3).

stalgiag:
> Since we are stalling the initialization waiting for the translations to finish downloading, it may be important that this step doesn't take too much time as well ensure that it doesn't spend time trying to find files for languages which we know aren't supported yet. This may be important when there is a chain of unsupported languages. For example, if someone's browser language is `fr-CA` ( Canadian French ), i18next will try to find `fr-CA` and since that isn't supported yet, try looking up `fr`. Since that too isn't supported, it finally comes to English. But it just spent time on two requests, which is easily prevented by just checking against a list of supported languages.

Thanks for explaining this in detail. I think that quickly skipping unsupported languages should be a priority for future work.

PR #4703, opened 2020-07-21
Merge Accessibility Add-On into p5.js

lm-n:
This PR merges the text output and table output functionalities of [p5.accessibility](https://github.com/processing/p5.accessibility) into p5.js by creating two functions:

-`textOutput()`
`textOutput()` creates a screenreader accessible output that describes the shapes present on the canvas. The general description of the canvas includes canvas size, canvas color, and number of basic shapes in the canvas example: 'Your output is a, 400 by 400 pixels, lavender blue canvas containing the following 4 shapes:'). This description is followed by a list of shapes where the color, position, and area of each shape are described (example: "orange ellipse at top left covering 1% of the canvas"). Each element can be selected to get more details. A table of elements is also provided. In this table, shape, color, location, coordinates and area are described (example: "orange ellipse location=top left area=2").

-`gridOutput()` 
`gridOutput()`, formerly called table output, lays out the content of the canvas in the form of a grid (html table) based on the spatial location of each shape. A brief description of the canvas is available before the table output. This description includes: color of the background, size of the canvas, number of objects, and object types (example: "lavender blue canvas is 200 by 200 and contains 4 objects - 3 ellipses 1 rectangle"). The grid describes the content spatially, each element is placed on a cell of the table depending on its position. Within each cell an element the color and type of shape of that element are available (example: "orange ellipse"). These descriptions can be selected individually to get more details. A list of elements where shape, color, location, and area are described (example: "orange ellipse location=top left area=1%") is also available.

#### PR Checklist

- [x] `npm run lint` passes

[paging @lmccart 4 the 👀  ]


src/accessibility/textOutput.js
```diff
@@ -0,0 +1,146 @@
...
+  return text;
+};
+
+//Builds textOutput table with shape details
+p5.prototype._buildShapeDetails = function(cId) {
+  let shapeDetails = '';
+  let el = 0;
+  for (let x in ingredients) {
+    for (let y in ingredients[x]) {
+      el++;
```
lmccart, line 75: seems a little counterintuitive to increment `el` at the start of the loop, meaning the first id starts with 1 instead of 0, which differs from code logic of counting starting at 0. but I might be missing something here?

src/accessibility/outputs.js
```diff
@@ -0,0 +1,418 @@
...
+  preIngredients = ingredients;
+  ingredients = {};
+  if (cnvConfig.backgroundRGBA !== args) {
+    //cnvConfig.backgroundRGBA = args;
+    cnvConfig.background = this._rgbColorName(args);
+  }
+};
+
+//helper function that gets fill and stroke of shapes
+p5.prototype._accscnvConfig = function(f, args) {
```
lmccart, line 202: maybe this should be `accsCnvCongig`?
also might make the code more readable to spell out canvas (ie `accsCanvasConfig`, `canvasConfig`)

src/accessibility/outputs.js
```diff
@@ -0,0 +1,418 @@
...
+        'afterend',
+        '<section id="' + cIdT + '" class="accessibleOutput"></section>'
+      );
+    if (type === 'txtOut') {
+      let inner = this._createTextOutput(cIdT);
+      document.getElementById(cIdT).innerHTML = inner;
+    }
+    if (type === 'grOut') {
+      let inner = this._createGridOutput(cIdT);
+      document.getElementById(cIdT).innerHTML = inner;
```
lmccart, line 170: I wonder if it's more performant to store key DOM elements internally rather than needing to use `getElementById` lookups for each update?

src/accessibility/color_namer.js
```diff
@@ -0,0 +1,694 @@
...
+  },
+  {
+    h: 95,
+    s: 1,
+    b: 1,
+    name: 'magenta'
+  }
+];
+
+p5.prototype._calculateColor = function(hsv) {
```
lmccart, line 619: if this is only used internally in this file, it doesn't need to be attached to the p5 prototype

src/accessibility/gridOutput.js
```diff
@@ -0,0 +1,159 @@
...
+  if (innerMap !== document.getElementById(cIdT + 'OD').innerHTML) {
+    document.getElementById(cIdT + 'OD').innerHTML = innerMap;
+  }
+  if (innerSDs !== document.getElementById(cIdT + 'SD').innerHTML) {
+    document.getElementById(cIdT + 'SD').innerHTML = innerSDs.details;
+  }
+};
+
+//creates spatial grid
+p5.prototype._buildGrMap = function(cId) {
```
lmccart, line 45: to make this more readable and align with the naming of other functions (ie `_updateGridOutput`), I would suggest writing out `Grid`

src/accessibility/gridOutput.js
```diff
@@ -0,0 +1,159 @@
...
+    Math.round(this.height) +
+    ' pixels, contains ' +
+    numShapes[0] +
+    ' shapes: ' +
+    numShapes[1];
+  return text;
+};
+
+//creates list of shapes
+p5.prototype._buildGridSDs = function(cId) {
```
lmccart, line 115: similar here, to make this more readable, you could write out `_buildGridShapeDetails`

src/accessibility/outputs.js
```diff
@@ -0,0 +1,418 @@
...
+//helper function that sets defaul colors for background
+//fill and stroke.
+p5.prototype._setDefaults = function() {
+  cnvConfig.background = 'white';
+  cnvConfig.fill = 'white';
+  cnvConfig.stroke = 'black';
+};
+
+// return length of lines
+p5.prototype._getLineL = function(args) {
```
lmccart, line 225: consider `_getLineLength` for readability.
but, does the existing `[dist](https://p5js.org/reference/#/p5/dist)` function work for this?

lmccart:
Looking good! Two main thoughts so far:
* I think we will want a document that gives an overview of how the accessibility stuff works in general, placed in the contributor_docs folder. The former [readme](https://github.com/processing/p5.accessibility/blob/master/README.md) can serve as a starting point. What's missing from this doc currently when I think about what's in this PR, is a little more overview of how the different files fit together, and the flow that happens internally. This will be useful for making this work sustainable for future contributors to continue.
* I left some notes throughout, but avoiding abbreviations in variable and function naming will also help ensure this code is readable.

src/accessibility/color_namer.js
```diff
@@ -0,0 +1,694 @@
...
+ * @for p5
+ * @requires core
+ */
+
+import p5 from '../core/main';
+import color_conversion from '../color/color_conversion';
+
+let oghsv;
+
+const xcp = [
```
kjhollen, line 13: thinking of Lauren's comment about the developer docs, it's probably worth a note there (with a short note here) about where these names come from, and whether or not someone can add colors.

src/accessibility/gridOutput.js
```diff
@@ -0,0 +1,159 @@
...
+  }
+  for (let x in ingredients) {
+    for (let y in ingredients[x]) {
+      el++;
+      if (
+        cells[ingredients[x][y].loc.locY][ingredients[x][y].loc.locX] ===
+        undefined
+      ) {
+        cells[ingredients[x][y].loc.locY][ingredients[x][y].loc.locX] =
+          '<a href="#' +
```
kjhollen, line 60: It looks like ES6 has some support for [template strings](https://developers.google.com/web/updates/2015/01/ES6-Template-Strings), which would probably make this kind of block easier to read. I can't remember if we're only supporting ES6 compatible browsers at this time or if we're trying to preserve backwards compatibility. @lmccart is it ok to use ES6 features like this?

src/accessibility/outputs.js
```diff
@@ -0,0 +1,418 @@
...
+ * @submodule Environment
+ * @for p5
+ * @requires core
+ */
+
+// Errors when running example:
+// - point
+// - line
+// - appending does not work
+// Call update only at the end of draw?
```
kjhollen, line 12: seeing your comment here. updating at the end makes sense (otherwise will the reader be notified of changes multiple times?)

src/accessibility/gridOutput.js
lmccart: yes we can use ES6! right now it compiles down to ES5 to support all modern browsers, but it's good to write in ES6 so we are future ready

kjhollen: rad! thank you!

src/accessibility/gridOutput.js
```diff
@@ -6,79 +6,106 @@
...
-          'shape' +
-          el +
-          '">' +
-          ingredients[x][y].color +
-          ' ' +
-          x +
-          '</a>';
+          `  <a href="#${cId}shape${shapeNumber}">${
+            ingredients[x][y].color
+          } ${x}</a>`;
```
kjhollen, line 106: if this is same as above maybe it can be a variable? before the if/else

src/accessibility/gridOutput.js
```diff
@@ -90,70 +117,52 @@ p5.prototype._buildGrMap = function(cId) {
...
-        cId +
-        'shape' +
-        el +
-        '">' +
-        ingredients[x][y].color +
-        ' ' +
-        x +
-        ',';
+      let line = `<li id="${cId}shape${totalShapes}">${
+        ingredients[x][y].color
```
kjhollen, line 141: yeah! this is so much more legible!

src/accessibility/outputs.js
```diff
@@ -5,18 +5,20 @@
  * @requires core
  */
 
-// Errors when running example:
-// - point
-// - line
-// - appending does not work
-// Call update only at the end of draw?
+// To-do:
+// Improve updating to include only shapes that have changed? [leave as an issue]
```
kjhollen, line 9: thinking about this more since we last discussed, this requires:
- updating items that are a match
- removing items that are no longer present
- adding new items

src/accessibility/outputs.js
```diff
@@ -98,92 +99,114 @@ p5.prototype.textOutput = function(display) {
...
- *   //fill(0, 0, 255);
- *   //rect(50, 50, 50, 50);
- *   //ellipse(20, 20, 20, 20);
- *   //x += 0.1;
- * //}
+ * <div>
+ * <code>
+ * let x = 0;
+ * function draw() {
+ *   textOutput();
```
kjhollen, line 118: gridOutput()?

src/accessibility/outputs.js
```diff
@@ -98,92 +99,114 @@ p5.prototype.textOutput = function(display) {
...
-    if (type === 'grOut') {
-      let inner = this._createGridOutput(cIdT);
-      document.getElementById(cIdT).innerHTML = inner;
-    }
-    //this._updateOutput();
+  let container = cnvId + 'accessibleOutput';
+  if (!doc.querySelector(`#${container}`)) {
+    doc.querySelector(
+      `#${cnvId}`
+    ).innerHTML = `<div id="${container}" " role="region" aria-label="Canvas Outputs"></div>`;
```
kjhollen, line 158: extra quote here, between container & role?

src/accessibility/gridOutput.js
```diff
@@ -6,79 +6,106 @@
...
-    'Grid Output<p id="' +
-    cIdT +
-    'Summary" aria-label="grid output summary"><table id="' +
-    cIdT +
-    'OD" summary="grid output content"></table><ul id="' +
-    cIdT +
-    'SD" aria-label="grid output shape details"></ul>';
+p5.prototype._createGridOutput = function(cIdT, cnvId, container, query) {
+  let doc = document.getElementsByTagName('body')[0];
+  let inner = `<div id="${cIdT}">Grid Output<p id="${cIdT}Summary" aria-label="grid output summary"><table id="${cIdT}OD" summary="grid output content"></table><ul id="${cIdT}SD" aria-label="grid output shape details"></ul></div>`;
```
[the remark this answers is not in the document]
lm-n: Tried breaking them but then linter would put them back together!

lm-n:
> Looking good! Two main thoughts so far:
> 
> * I think we will want a document that gives an overview of how the accessibility stuff works in general, placed in the contributor_docs folder. The former [readme](https://github.com/processing/p5.accessibility/blob/master/README.md) can serve as a starting point. What's missing from this doc currently when I think about what's in this PR, is a little more overview of how the different files fit together, and the flow that happens internally. This will be useful for making this work sustainable for future contributors to continue.
> * I left some notes throughout, but avoiding abbreviations in variable and function naming will also help ensure this code is readable.

Thanks Lauren! I've added a doc documenting how different files and functions fit together under contributor docs.

src/accessibility/color_namer.js
lm-n: Just added this to the contributor docs!

contributor_docs/web_accessibility.md
```diff
@@ -0,0 +1,96 @@
...
+
+`gridOutput()`, formerly called table output, lays out the content of the canvas in the form of a grid (html table) based on the spatial location of each shape. A brief description of the canvas is available before the table output. This description includes: color of the background, size of the canvas, number of objects, and object types (example: "lavender blue canvas is 200 by 200 and contains 4 objects - 3 ellipses 1 rectangle"). The grid describes the content spatially, each element is placed on a cell of the table depending on its position. Within each cell an element the color and type of shape of that element are available (example: "orange ellipse"). These descriptions can be selected individually to get more details. A list of elements where shape, color, location, and area are described (example: "orange ellipse location=top left area=1%") is also available.
```
kjhollen, line 17: "formerly" - I think this is a note for people who would have used the p5.accessibility library? suggest removing this or explaining a little clearer.

contributor_docs/web_accessibility.md
```diff
@@ -0,0 +1,96 @@
...
+* `gridOutput()`: This function activates the grid output by setting `this._accessibleOutputs.grid` to `true` and calling `_createOutput('gridOutput', 'Fallback')`. If `LABEL` is passed as a parameter the function also activates the grid output label by setting `this._accessibleOutputs.textLabel` as `true` and calls `_createOutput('gridOutput', 'Label')` for the label.
+* `_createOutput()`: This function creates the HTML structure for all accessible outputs. Depending on the type and display of the outputs the HTML structure created varies. The function also initializes `this.ingredients` which stores all the data for the outputs including: shapes, colors, and pShapes (which stores a string of the previous shapes of the canvas). It also creates `this.dummyDOM` if it doesn't exist. `this.dummyDOM` stores the HTMLCollection of DOM elements inside of `<body>`.
+* `_updateAccsOutput()`: Is called at the end of setup and draw if using accessibleOutputs and calls update functions of outputs(`_updateGridOutput` and `_updateTextOutput`).
```
kjhollen, line 29: worth a note here that this is only called once per draw/setup so as not to overwhelm the screen reader

contributor_docs/web_accessibility.md
```diff
@@ -0,0 +1,96 @@
...
+* `_descriptionText()`: Checks that text is not `LABEL` or `FALLBACK` and ensures text ends with a punctuation mark. If the text does not end with '.', ',', ';', '?', '!', this function adds a '.' at the end of the string. Returns text.
+* `_describeFallbackHTML()`: Creates fallback HTML structure for the canvas. 
+* `_describeLabelHTML()`: This function is only called when the second parameter of `_describe()` is `LABEL`. It creates a div adjacent to the canvas element for the description text.
+
+### describeElement()
+The `_describeElement()` function creates a screen reader accessible description for sketch elements or groups of shapes that create meaning together. The first parameter should be a string with the name of the element, the second parameter should be a string with the description of the element. The third parameter is optional. If specified, it determines how the description is displayed. All element descriptions become part of the sub DOM of the canvas element. If a user passes `LABEL` as a third parameter, an additional div with the element description adjacent to the canvas is created.
```
kjhollen, line 90: remove underscore in `_describeElement()`

contributor_docs/web_accessibility.md
```diff
@@ -0,0 +1,96 @@
...
+* `_describeFallbackHTML()`: Creates fallback HTML structure for the canvas. 
+* `_describeLabelHTML()`: This function is only called when the second parameter of `_describe()` is `LABEL`. It creates a div adjacent to the canvas element for the description text.
+
+### describeElement()
+The `_describeElement()` function creates a screen reader accessible description for sketch elements or groups of shapes that create meaning together. The first parameter should be a string with the name of the element, the second parameter should be a string with the description of the element. The third parameter is optional. If specified, it determines how the description is displayed. All element descriptions become part of the sub DOM of the canvas element. If a user passes `LABEL` as a third parameter, an additional div with the element description adjacent to the canvas is created. 
+
+`_describeElement()` is supported by several functions in [src/accessibility/describe.js](https://github.com/processing/p5.js/blob/main/src/accessibility/describe.js):
```
kjhollen, line 92: here too (remove underscore)

src/accessibility/color_namer.js
```diff
@@ -0,0 +1,700 @@
...
+ * @submodule Environment
+ * @for p5
+ * @requires core
+ */
+
+import p5 from '../core/main';
+import color_conversion from '../color/color_conversion';
+
+//stores the original hsv values
+let oghsv;
```
kjhollen, line 12: suggest a longer, more descriptive variable name here

src/accessibility/describe.js
```diff
@@ -80,41 +78,43 @@ p5.prototype.describe = function(text, display) {
...
+  if (!this.dummyDOM) {
+    this.dummyDOM = document.getElementsByTagName('body')[0];
   }
-
-  //check if text is different
-  if (dummy[cnvId + 'fallbackDesc'] !== text) {
-    //if html structure for description is ready
-    if (dummy[cnvId + 'updateFallbackDesc']) {
+  //check if html structure for description is ready
+  if (this.dummyDOM.querySelector(`#${cnvId}${fallbackDescId}`)) {
```
kjhollen, line 86: seeing this queries makes me wonder if we gain a little performance by having separate variables for each, like:

this.fallbackDescElement
this.labelDescElement

etc. so the check can just be if the appropriate element exists, and if it doesn't, it gets injected into the DOM. hope that makes sense.

src/accessibility/describe.js
```diff
@@ -178,44 +178,45 @@ p5.prototype.describeElement = function(name, text, display) {
...
-  //if it is the first time describeElement() is called
-  if (
-    !dummy.fallbackElements[cnvId + name] ||
-    !dummy.labelElements[cnvId + name]
-  ) {
-    //store copy of body dom in dummy
-    _populateDummyDOM(cnvId);
+  //if there is no dummyDOM
+  if (!this.dummyDOM) {
+    this.dummyDOM = document.getElementsByTagName('body')[0];
```
kjhollen, line 183: does this mean that we always expect the element to be the first item appended to the body? what if the user has specified a different/custom place for the canvas?

src/accessibility/describe.js
```diff
@@ -80,41 +80,40 @@ p5.prototype.describe = function(text, display) {
...
-      ).innerHTML = text;
+      this.descriptions.fallback.innerHTML = text;
     }
   } else {
     //create fallback html structure
-    let results = _describeFallbackHTML(this.dummyDOM, cnvId, text);
-    this.dummyDOM = results;
+    let results = _describeFallbackHTML(this.dummyDOM, cnvId);
+    this.descriptions.fallback = results[0];
+    this.dummyDOM = results[1];
```
kjhollen, line 99: why does this overwrite the `this.dummyDOM` assignment above on line 83?

src/accessibility/describe.js
```diff
@@ -80,41 +80,40 @@ p5.prototype.describe = function(text, display) {
...
-        this.dummyDOM.querySelector(`#${cnvId}${labelDescId}`).innerHTML = text;
+        this.descriptions.label.innerHTML = text;
       }
     } else {
       //create label html structure
-      let results = _describeLabelHTML(this.dummyDOM, cnvId, text);
-      this.dummyDOM = results;
+      let results = _describeLabelHTML(this.dummyDOM, cnvId);
+      this.descriptions.label = results[0];
+      this.dummyDOM = results[1];
```
kjhollen, line 115: same question here as above -- why does this overwrite setting `this.dummyDOM` on line 83?

src/accessibility/describe.js
```diff
@@ -180,43 +179,45 @@ p5.prototype.describeElement = function(name, text, display) {
...
-      ).innerHTML = inner;
+      this.descriptions.fallbackElements[name].innerHTML = inner;
     }
   } else {
     //create fallback html structure
-    let results = _descElementFallbackHTML(this.dummyDOM, cnvId, name, inner);
-    this.dummyDOM = results;
+    let results = _descElementFallbackHTML(this.dummyDOM, cnvId, name);
+    this.descriptions.fallbackElements[name] = results[0];
+    this.dummyDOM = results[1];
```
kjhollen, line 200: overwrites assignment of `this.dummyDOM` on line 182?

src/accessibility/describe.js
```diff
@@ -180,43 +179,45 @@ p5.prototype.describeElement = function(name, text, display) {
...
+    this.descriptions.fallbackElements[name].innerHTML = inner;
   }
   //if display is LABEL
   if (display === this.LABEL) {
+    if (!this.descriptions.labelElements) {
+      this.descriptions.labelElements = {};
+    }
     //if html structure for label element description is ready
-    if (this.dummyDOM.querySelector('#' + cnvId + labelTableElId + name)) {
+    if (this.descriptions.labelElements[name]) {
```
kjhollen, line 209: this is looking much clearer to me — nice work!

src/accessibility/describe.js
```diff
@@ -180,43 +179,45 @@ p5.prototype.describeElement = function(name, text, display) {
...
-        ).innerHTML = inner;
+        this.descriptions.labelElements[name].innerHTML = inner;
       }
     } else {
       //create label element html structure
-      let results = _descElementLabelHTML(this.dummyDOM, cnvId, name, inner);
-      this.dummyDOM = results;
+      let results = _descElementLabelHTML(this.dummyDOM, cnvId, name);
+      this.descriptions.labelElements[name] = results[0];
+      this.dummyDOM = results[1];
```
kjhollen, line 219: overwrites assignment of `this.dummyDOM` on line 182?

src/accessibility/describe.js
```diff
@@ -251,13 +252,13 @@ function _descriptionText(text) {
...
 //creates fallback HTML structure
-function _describeFallbackHTML(dummyDOM, cnvId, text) {
+function _describeFallbackHTML(dummyDOM, cnvId) {
   //if there is no description container
   if (!dummyDOM.querySelector(`#${cnvId + descContainer}`)) {
     //if there are no accessible outputs (see textOutput() and gridOutput())
     if (!dummyDOM.querySelector(`#${cnvId}accessibleOutput`)) {
       //create description container + <p> for fallback description
-      dummyDOM.querySelector('#' + cnvId).innerHTML = `<div id="${cnvId +
+      dummyDOM.querySelector(`#${cnvId}`).innerHTML = `<div id="${cnvId +
```
kjhollen, line 261: before this inner if/else, maybe (on line 258):

``let descContainer = `<div id="${cnvId + descContainer}" role="region" aria-label="Canvas Description"><p id="${cnvId + fallbackDescId}"></p></div>`; ``

then use variable in if/else (unless I missed something when I read this, it looks like the HTML is the same, just how it gets inserted changes in the condition)

src/accessibility/describe.js
```diff
@@ -283,15 +284,13 @@ function _describeFallbackHTML(dummyDOM, cnvId, text) {
   }
   //If the container for the description exists
   if (dummyDOM.querySelector(`#${cnvId}${fallbackDescId}`)) {
-    //update description
-    dummyDOM.querySelector(`#${cnvId}${fallbackDescId}`).innerHTML = text;
-    return dummyDOM;
+    return [dummyDOM.querySelector(`#${cnvId}${fallbackDescId}`), dummyDOM];
```
kjhollen, line 287: is the if statement here necessary? looks like it gets inserted in lines 257-273. there is also not an else clause that returns something if the function was not successful (and it seems other places in the code assume success).

I had to look this up because I wasn't sure how Javascript handles this, but I think it's safe not to pass dummyDOM back. It looks like the dot operator will change the object and its changes will hold outside of the function, if I'm understanding this stack overflow post correctly: https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language

src/accessibility/describe.js
```diff
@@ -321,9 +320,9 @@ function _describeLabelHTML(dummyDOM, cnvId, text) {
       .querySelector(`#${cnvId + labelTableId}`)
       .insertAdjacentHTML('beforebegin', `<p id="${cnvId}${labelDescId}"></p>`);
   }
-  //update description
-  dummyDOM.querySelector('#' + cnvId + labelDescId).innerHTML = text;
-  return dummyDOM;
+  if (dummyDOM.querySelector(`#${cnvId}${labelDescId}`)) {
```
kjhollen, line 323: similar questions/comments here as for _describeFallbackHTML with this return clause and the insertion of the HTML earlier on

src/accessibility/describe.js
```diff
@@ -379,23 +379,22 @@ function _descElementFallbackHTML(dummyDOM, cnvId, name, inner) {
...
-    return dummyDOM;
-  }
+
+  //create a table row for the element
+  let tableRow = document.createElement('tr');
+  tableRow.id = cnvId + fallbackTableElId + name;
+  dummyDOM.querySelector('#' + cnvId + fallbackTableId).appendChild(tableRow);
+  //update element description
+  return [
+    dummyDOM.querySelector('#' + cnvId + fallbackTableElId + name),
```
kjhollen, line 389: similar questions/comments here as for _describeFallbackHTML with this return clause and the insertion of the HTML earlier on

src/accessibility/describe.js
```diff
@@ -430,18 +430,15 @@ function _descElementLabelHTML(dummyDOM, cnvId, name, inner) {
...
-    ).innerHTML = inner;
-    return dummyDOM;
-  }
+  //create a table row for the element label description
+  let tableRow = document.createElement('tr');
+  tableRow.id = cnvId + labelTableElId + name;
+  dummyDOM.querySelector('#' + cnvId + labelTableId).appendChild(tableRow);
+  //update element label description
+  return [
+    dummyDOM.querySelector('#' + cnvId + labelTableElId + name),
```
kjhollen, line 439: similar questions/comments here as for _describeFallbackHTML with this return clause and the insertion of the HTML earlier on

lm-n:
Hi @kjhollen! I've added my project wrap up! Lmk what you think!

src/accessibility/describe.js
```diff
@@ -346,99 +327,88 @@ function _elementName(name) {
...
+      if (!this.dummyDOM.querySelector(`#${cnvId}accessibleOutput`)) {
+        //create container + table for element descriptions
+        this.dummyDOM.querySelector('#' + cnvId).innerHTML = html;
+      } else {
+        //create container + table for element descriptions before outputs
+        this.dummyDOM
+          .querySelector(`#${cnvId}accessibleOutput`)
+          .insertAdjacentHTML('beforebegin', html);
+      }
+      //if describe() has already created the container and added a description
```
kjhollen, line 347: at first, I didn't read this as going with the `else` below because the indentation matched above

contributor_docs/web_accessibility.md
```diff
@@ -83,14 +83,12 @@ The `describe()` function creates a screen reader accessible description for the
...
 
 `describeElement()` is supported by several functions in [src/accessibility/describe.js](https://github.com/processing/p5.js/blob/main/src/accessibility/describe.js):
 * `_elementName()`: Checks that element name is not `LABEL` or `FALLBACK` and ensures text ends with a colon. Returns element name.
 * `_descriptionText()`: Checks that text is not `LABEL` or `FALLBACK` and ensures text ends with a punctuation mark. If the text does not end with '.', ',', ';', '?', '!', this function adds a '.' at the end of the string. Returns text.
-* `_describeElementFallbackHTML()`: Creates fallback HTML structure for the canvas. 
-* `_describeElementLabelHTML()`: This function is only called when the second parameter of `_describeElement()` is `LABEL`. It creates a div adjacent to the canvas element for the description text.
+* `_describeElementHTML()`: Creates fallback HTML structure for the canvas. When the second parameter of `_describeElement()` is `LABEL`, this function creates a div adjacent to the canvas element for the descriptions.
```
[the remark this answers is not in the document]
lm-n: thanks for catching this!

src/accessibility/describe.js
lm-n: umm that's a good point, that's the way lint makes it look. I'll move the comment after the else!

lmccart:
@lm-n This is really fantastic work, I'm so excited about it. A big step forward for the p5.accessibility project. I'm looking forward to continuing this work with you all, in any way we can. Thank you for your time and energy this summer, and for the thorough writeup and documentation. And thanks to @kjhollen for mentoring this summer!

lm-n:
> Hey Luis, this looks like it's ready to go. I found one little typo in the documentation. The comments that you've added are super helpful, and will make it easier for other folks to contribute new features and maintain the existing functionality. Thanks for all your work on this and your many rounds of revisions!

Hi Kate :heart: I've fixed the typo and moved the comments below the else! I think we should be ready to merge :smile: Thanks so much for all your help and support! @lmccart and @CleezyITP thank you :heart:, I'm excited to see what comes next!

LucaDamasco:
Just wanted to pop in and say this rocks. Really excited to be able to more easily share this work with my students and seeing that this was directly integrated into P5 after searching up p5.accessibility today was such an awesome sight! ❤️

Issue #4662, opened 2020-06-26
FES: Friendly messages for global errors


akshay-99:
#### Most appropriate sub-area of p5.js?
- [x] Other (Friendly Errors)

There are a lot of errors that a person may encounter when writing a sketch. Some of these are quite straight forward, while others are more cryptic. It would be nice if the FES can monitor all the errors that arise, and see if it can give a simplified explanation for them and so I browsed through the discourse forum to learn about the errors that usually confuse people.

It may be useful to differentiate between errors that happen in the user's sketch and those that happen within the p5 library. I had first thought about this when I read [this comment](https://github.com/processing/p5.js/issues/3447#issuecomment-628957362). The code inside the library is not within the user's control and so it's not helpful if the focus is put on a line inside the library. A friendlier error message would instead focus on the sketch line that called the library function, and inform the user that it either probably has something to do with the arguments they passed (so they should check the reference once), or they may have stumbled upon some unexpected behaviour of the library ( so they should ask someone, maybe on the forum or here on Github ).  Below is another example error message I came across while browsing through discourse forum [here](https://discourse.processing.org/t/use-an-array-of-color-resolved/6936)
```
Uncaught (in promise) Error: [object Arguments]is not a valid color representation.
    at _main.default.Color._parseInputs (p5.js:43255)
    at new _main.default.Color (p5.js:42293)
    at p5._main.default.color (p5.js:41912)
    at _main.default.Renderer2D.stroke (p5.js:50518)
    at p5._main.default.stroke (p5.js:44066) // The first useful line in the message
    at setup (sketch.js:4) 
    at _setup (p5.js:48328)
    at _runIfPreloadsAreDone (p5.js:48266)
    at p5._decrementPreload (p5.js:48275)
    at p5.<anonymous> (p5.sound.min.js:2)
```
Although many of the argument scenarios can be dealt with by expanding the use of `validateParameters` for more functions, it would impact performance. Some of these errors can arise even if the argument type is correct (as seen in the comment mentioned above).  Also for most errors, much of the stacktrace is not that useful without deep knowledge. It is possible to simplify things by removing details about internal library calls. This can be even applied to errors that happen outside the library.

[...]

PR #4730, opened 2020-08-08
FES: Fix some issues in global error catching

akshay-99:
Addresses #4662 

### Changes:
There was an issue with global error catching in FES. Primarily the event listener for `unhandledrejection` failed to trigger in Chrome if the same-origin policy was not satisfied, or if the protocol used was `file:///` ( running p5 locally, without a local server ). This issue was due to a security policy in Chrome ( and maybe other similar browsers ) that would silently ignore the event listener if the origin of the event and the current origin didn't match. 

To fix this, the idea now is to enclose the calls to user-defined functions ( such as `setup`, `draw`, `preload`, `mouseClicked`, etc) in a `try-catch` block. Since these functions are called from the p5 core, the errors in them are propagated upwards and can be handled in a `catch` block. Once parsed and simplified by `fesErrorMonitor`, we again throw it upwards so that the original error too can appear on the console. The propagation of an error through the call stack is not affected by the security policy, even if different origins are involved ( probably because in this case it can only be seen by functions that directly led to the error )

This enclosing in a try-catch block is done by using a wrapper function `fesErrorWrapper` that wraps around a call to user-defined function from p5. 

#### PR Checklist

- [x] `npm run lint` passes


stalgiag:
I am not sure that the benefit of this change justifies the complexity it adds to the core code. It is not ideal to run all of the core user functions through a wrapper property. 

I would love to get a third opinion on this -- @lmccart would you be willing to add your thoughts? The situation is that FES global error messages do not work with Chrome when a user doesn't use a local server or the online editor. This PR fixes this behavior so that a user can open an html file in Chrome and still get proper FES messages. We encourage using the online editor and development server approaches in most documentation, but of course some amount of new users will inevitably open the html file in a browser anyways.

Do you think that the added complexity in the core (this PR) is worth it to ensure that these users (Chrome with HTML file open) get proper FES messaging?

akshay-99:
@stalgiag I agree that a wrapper function may have gone a bit beyond just a simple enclosing in try catch block as we discussed. But there were a few more instances where the p5 library transfers control to user-code which I did not account for earlier. 

There are a total of 5 instances where the transfer of control happens from p5 to user-code:
1. calling user-defined preload()
2. calling user-defined setup()
3. calling user-defined draw()
4. calling user-defined event handlers ( mouseClicked, mouseMoved, mouseMoved, etc)
5. calling user-defined event handlers attached to a p5.Element ( eg. cnv.mouseClicked(myfunc) )

The first 3 are called directly from p5 code in the normal flow of things, so there is no need for a wrapper function there ( a simple enclosing of the calls in try-catch would work, which would not affect the code complexity much). The last two are triggered upon particular events, where the user-defined function is passed to `addEventListener`. Here the wrapper function comes in handy as instead of passing the user-defined function to `addEventListener`, we can pass the wrapper. The event listener calls the wrapper, which calls the user-defined function. 

As you correctly said, it does have a negative effect on code-readability and understandability ( more-so than if using just a simple try-catch ). I am not aware what per cent of new users run p5 by double-clicking the HTML file so I am not sure if the advantages would outweigh the disadvantages here :sweat_smile:  I am open to views on that.

lmccart:
Hi @akshay-99 thanks for this, and @stalgiag for your consideration of code readability. This is a tough one, but I'd come down on the side of keeping the code simpler rather than trying to account for this case. My thinking is that most beginners now use the web editor for getting started, rather than downloading the zip and trying to double click the index file. This leads me to believe that the users that are linking in p5 from CDN or using the p5 complete download are familiar enough to run a server, or will get quickly pointed in that direction via p5 documentation and/or forum. (I used to see more confusion around running a server on github issues, and that has tapered off a lot with the introduction of the web editor.)

On the other hand, from a code contributor's perspective, the core code can be some of the trickiest to follow because there are so many different things happening. So anywhere we can err on the side of simplicity helps with this. Hope that sounds reasonable.

akshay-99:
Thanks @lmccart for your views on this. So I will close this PR now and add these details in the Limitations section of the FES contributor docs?

When I was reading about this issue and finding ways to fix it, I was confused as to why this FES functionality didn't break on the web editor as well. I initially expected it to break since the origins `editor.p5js.org` and `cdnjs.cloudflare.com` are different. 
I understand now that the web editor adds a `crossorigin` attribute to all script imports behind the scene and that's why this problem is not faced there.

lmccart:
Thanks for your understanding and flexibility @akshay-99! Your suggestion to add this into the Limitations section sounds good. And yes, that's correct re: the web editor as far as I understand.

PR #4733, opened 2020-08-11
p5.js-website update workflow

yukienomiya:
### Changes:
Add a GitHub workflow that gets triggered every time a p5.js release gets published. This workflow will then trigger [a second workflow](https://github.com/processing/p5.js-website/pull/841) on the p5.js-website repository that will update the p5.js documentation in the website's repository.

**This way the data.json file won't need to be manually updated every time there's a new p5.js release.**


limzykenneth:
The steps in [this file](https://github.com/processing/p5.js/blob/main/tasks/release/release-docs.js) should be automating releasing `data.json` to the website when a new version is released. If this PR implements what the above script does then we should remove the old script.

outofambit:
hey @limzykenneth sorry i should have remembered that file when @yukienomiya and i were first discussing this work. this work in combination with processing/p5.js-website#841 does replace this script (and removes the requirement to run locally), so we can remove it.

PR #4756, opened 2020-08-19
Allow drawing point with large scale and small strokeWeight.

### LINKED ISSUE #4707: Shape point() does not work with scale > 1, strokeWeight < 1
#### Most appropriate sub-area of p5.js

#### Details about the bug:

- p5.js version: 1.1.9 July 22, 2020
- Web browser and version: Safari 13.1.2 (15609.3.5.1.3)
- Operating System: MacOS 10.15.6
- Steps to reproduce this: I have a square canvas and want to use it as a coordinate system(-1..+1, -1..+1). So I translate to (width/2, width/2) and scale by (width/2,-width/2). Then I set the strokeWeight to be 1 actual pixel: 2/width. Expectation: lines, circles and points drawn will be 1 pixel wide. This is true for lines and circles, but not points. It seems that the strokeWeight is not used for points, perhaps because it can't be smaller than 1, but the scale *is* respected. This results in mega-points covering a quarter of the canvas (because in scaled coordinates the area is 4). Code that fails:

```
function setup() {
	createCanvas(400, 400);		// square canvas
	background(0);
	noFill();

	const pixperunit = width / 2;	// number of pixels per unit in the desired coordinate system
	translate(pixperunit, pixperunit); // set the origin in the middle
	scale(pixperunit, -pixperunit);	// horizontal and vertical range is now -1..+1
	strokeWeight(1 / pixperunit);	// all lines, arcs and points should be 1 pixel wide

	stroke(255, 0, 0);		// red axes
	line(-1, 0, 1, 0);
	line(0, -1, 0, 1);

	stroke(0, 255, 0);		// green circle
	circle(0, 0, 1);

	stroke(0, 0, 255);		// blue dot
	point(0, 0);
}
```

The resulting picture, which should have one blue pixel in the centre, not a large blue square:
Schermafbeelding 2020-07-23 om 15 04 48

TakumaKira:
### Resolves #4707 Shape point() does not work with scale > 1, strokeWeight < 1

 Changes:

- Remove the conditional statement to replace a circle with the radius below 1px to a rectangle.
- Modify the conditional statement of the test to allow a not entirely white pixel of a dot with the radius below 1px.

#### PR Checklist

- [x] `npm run lint` passes


lmccart:
thanks!

PR #4770, opened 2020-08-26
Implementing p5.Vector setHeading() (Issue #4767)

### LINKED ISSUE #4767: Implementation of p5.Vector setHeading()
I am opening a new issue to implement a `setHeading` method for the p5.Vector class. Currently, p5 contains a function `rotate()`, which rotates the vector `by` a certain value, but no such method exists for setting the vector's rotation to an explicit direction. I currently have something working, and am planning to pull request an implementation.

An implementation would look as follows:
```javascript

p5.Vector.setHeading = function(angle){
  this.rotate(-this.heading())
  this.rotate(angle)
}
```

This is my first time attempting to contribue to open source, so any tips would be greatly appreciated!

samdelong:
This commit adds the `setHeading()` function to the `p5.Vector` class. There are
some remaining issues here. For example, there are no tests. However, there
is an example of it's use.

Resolves #4767 

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -1441,6 +1441,30 @@ p5.Vector.prototype.heading = function heading() {
 };
 
 /**
+ * Rotate the vector to a specific angle (only 2D vectors), magnitude remains the
+ * same
+ *
+ * @method setHeading
+ * @param  {number}    angle the angle of rotation
+ * @chainable
+ * @example
+ * <div class="norender">
+ * <code>
+ * let v = createVector(10.0, 20.0);
+ * // result of v.heading() is 1.1071487177940904
+ * v.setHeading(Math.PI);
+ * // result of v.heading() is now 3.141592653589793
+ * </code>
+ * </div>
+ */
+
+p5.Vector.prototype.setHeading = function setHeading(a) {
+  this.rotate(-this.heading());
+  this.rotate(a);
+  return this;
+};
+
+/**
  * Rotate the vector by an angle (only 2D vectors), magnitude remains the
  * same
  *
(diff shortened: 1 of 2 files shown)


limzykenneth:
Can you include unit tests with this new function as well? Don't worry if not, just let us know here.

samdelong:
Sure, I can do that when I get back to the computer this evening

src/math/p5.Vector.js
```diff
@@ -1440,6 +1440,31 @@ p5.Vector.prototype.heading = function heading() {
...
+ * <code>
+ * let v = createVector(10.0, 20.0);
+ * // result of v.heading() is 1.1071487177940904
+ * v.setHeading(Math.PI);
+ * // result of v.heading() is now 3.141592653589793
+ * </code>
+ * </div>
+ */
+
+p5.Vector.prototype.setHeading = function setHeading(a) {
```
lmccart, line 1461: to address @montoyamoraga's point, let's add a friendly error message check into this one so if people try to use it with 3D vectors it will give an error message to help debug. you can see an example of how to do this [here](https://github.com/processing/p5.js/blob/main/src/core/shape/2d_primitives.js#L173)

lmccart (changes requested):
this looks good to me, once FES check and unit tests are added. thanks @samdelong!

src/math/p5.Vector.js
samdelong: I'm a little confused how the FES system would work here. How would one use the FES system to check if the 'z' component of the object is undefined?

lmccart: you just add the line:
```js
p5._validateParameters('p5.Vector.setHeading', arguments);
```
and it takes care of itself, checking against the @param documentation for this method that you've added

lmccart: ah sorry scratch that, i thought the vector was passed as an argument. so this won't work.

samdelong: No worries. I'm going to push the unit tests and hopefully we can get the FES up and running

lmccart:
thanks @samdelong!

PR #4779, opened 2020-08-29
Allows ES6 sub-classsing of p5

### LINKED ISSUE #4466: ES6 style class extension of p5 instance mode sketch.
It has been previously suggested that the P5.js library may at some point allow instance mode sketches using ES6 style class extensions. Currently, it seems however, that the library does not entirely work using this method as is. 
One of the issues that i found, is that the `sync` argument, (that i believe is supposed to define whether the start of the sketch is sync or async?) is unused. Because the sketch always starts sync, if there are any variables in the sub-class that are passed as constructor arguments, they are not yet initialised as `this.` properties when `setup` is called, since is is called from within the `super` call.
Another issue that I came across was that when `draw` and other async methods are called, they do not seem to be correctly bound to the p5 ojbect, causing issues that prevent drawing etc.
As for the second issue, I was able to rectify it generally by binding all of the async function calls to `this`.
```typescript
let _socket = io();
class sketch extends p5 {
super
    constructor(socket){
        //super must be called before any this assignments. even so, async should be an option?
        super(()=>{}, document.getElementById("sketch"), true/*doesn;t do anything*/);
        this.socket=socket;
    }
    setup(){
        this.socket.on("messgage", doSomething);
        //error this.socket is not defined;
        //if this was async, super returns, socket assigned, then setup called async.
    }
    doSomething(){}
}
new sketch(_socket);
```

Anyways, i know this is a bit long, but I think that it would be much nicer to be able to instantiate p5 as an extended class rather than having to use a seed function to spawn a sketch,

---

edit: forgot some `this.`s

jpdutoit:
Resolves #4466

This pull request allows you to subclass p5 to create a sketch. It wasn't necessary to add any new features to allow this. I only needed to use the same pattern to access the user defined functions as was already used in some parts of the code.

Changes:

- Stop storing user functions in temporary variables to prevent context loss
- Always check _isGlobal to figure out where to find user functions

Example:
```
import p5 from "p5"

class Sketch extends p5 {
  constructor() {
    // Unfortunately you still need to pass the function here,
    // otherwise P5 will think that you want the global style
    super(() => {})
  }

  draw() {
    this.line(0, 0, 100, 100)
    this.line(0, 100, 100, 100)
  }

  keyPressed() {
    console.log("Pressed", this.key)
  }
}

new Sketch()
```

#### PR Checklist

- [x] `npm run lint` passes


limzykenneth (approved):
Changes here shouldn't cause any problems from what I can see. It's a good start for more diverse use case of the library.

limzykenneth:
My thinking is that this increases access but not in a way that is very directly related to p5.js. My view is that for a beginner to coding in general, p5.js is a great place to start learning to code and it is one of the focus of the project (especially on p5.js' usage in education and art) but it is not necessarily something that a coder going beyond starting to learn how to code will want to keep using (although many like myself still do), and while p5.js uses some tricks to make things like global mode possible, it is not necessarily the aim to keep users in the bubble of p5.js indefinitely. I see a user out grows the capability of the library as a good sign in this case.

The way this addition increases access is in this area is by providing a way to incrementally transition from just using p5.js. This could be a way for educators or self learners to teach/learn subclassing and ES6 in general in a more practical and substantial way, for example.

Of course one of the point raised when we set out to only include new features that increases access is that access means different thing to different peolpe and we should figure out what it means for p5.js together, so I can totally accept that my view does not necessarily reflect the definition of access that we are looking for.

outofambit:
hey @limzykenneth so sorry for raising a concern and then disappearing! thanks for sharing your thoughts on accessibility here. i resolved the merge conflicts that have come up in the meantime (its just `.all-contributorsrc`), please merge as you please!

limzykenneth:
Thanks everyone!

christopherball:
I know this comment is on an old thread already merged into the codebase, but I wanted to say adding this ability to extend the Sketch class is **fantastic** as it effectively allows for another pseudo-instanced way to work with p5 without polluting the global namespace, whilst also allowing IDE's like vscode to properly provide autocomplete (compared to using the standard instanced approach where vscode can't infer the type of the p5 instance, so none of the autocomplete functionality works).  Last resort was adding a typescript layer on top but preferred to keep things simpler if possible.  Thanks for this merge!

PR #4817, opened 2020-09-30
Change distance example to be more reasonable (addresses #2865)

samporapeli:
**Changes:**
Previously the example converted float to integer, thus removing all decimal places. However, the text displaying distance has exactly one decimal place. It makes more sense to me to use that decimal place instead of having it be zero all the time. This way the nature of distance function is also better demonstrated (it returns a float instead of an integer).

**Screenshots of the change:**
Before:
Screenshot_20201001_004621
(Notice the .0)
After:
Screenshot_20201001_004320

#### PR Checklist

- [x] `npm run lint` passes
- [x] [Inline documentation] is updated

### DIFF
--- a/src/math/calculation.js
+++ b/src/math/calculation.js
@@ -142,7 +142,7 @@ p5.prototype.constrain = function(n, low, high) {
  *
  *   // d is the length of the line
  *   // the distance from point 1 to point 2.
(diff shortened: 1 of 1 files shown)


limzykenneth (approved):

lmccart:
@all-contributors please add @samporapeli for examples

PR #4824, opened 2020-10-02
Fix GIF timing when the same GIF is used multiple times

### LINKED ISSUE #4818: Animated GIF frame delay off when drawing multiple times per frame

- [ ] Color
- [ ] Data
- [ ] Dom
- [ ] Events
- [x] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

Rendering an animated GIF multiple times in one frame causes the frame rate to increase relative to the number of times the GIF is displayed.  See [example](https://editor.p5js.org/owenroberts/sketches/AK0iCKqXx).

It is possible to adjust the delay based on the number of instances by multiplying the delay of each frame by the number of instances, but only if you know how many instances will be drawn in each frame, and the result is slightly off (see example).

The simplest fix I came up with for this feels a bit hacky but seems to work.  [This commit](https://github.com/owenbmcc/p5.js/commit/7c63a90f05b11e8377d65e53fc785b0c47c7989e) compares the current frameCount to an added property to the GIF image so it doesn't advance the gif more than once per frame.

This assumes that desired way of rendering multiple instances of one animated GIF is they play at the same rate as one instance of the GIF would play.  I think it's what one would expect to see.  If that's not the desired approach, it could be added as an optional argument or something.

Another possibility would be adding a new method to multiply the delay by the number of instances, that seems to work okay but also is a bit off as I mentioned, I haven't looked into why.

For some context, I use p5 in classes I teach and this has come up recently with a student who wanted to create a tiled background of animated GIFs, which causes them to animate 16x faster.  I've used p5.play in the past for sprites, I did a quick test there and it looks like it has the same issue when using sprite sheets.

Using the delay method is a relatively simple sketch level fix but it would be nice if this was either fixed or somehow the way to fix it was documented.

stalgiag:
Resolves #4818 

#### Changes:

Copying comment from #4818 

> 
> So since I didn't think to prepare for this behavior when making the GIF functionality, I decided to take a crack at fixing it. @akshay-99 I mostly took your suggestion. This was what I wanted to do when originally implementing the GIFs but I didn't because of potential differences between when the GIF is loaded and when it is first used. If we check current time against curFrameFirstPaintTime then we will end up skipping the first frame when a user has a delay between loading the GIF and displaying it.
> 
> So I slightly modified the approach to set lastChangedTime (my version of your curFrameFirstPaintTime) to millis() when the GIF is displayed and lastChangedTime is 0. This allows us to set lastChangedTime to 0 on load, reset(), and setFrame() and not concern ourselves with the difference between the time of this modification and display time.
> 
> I also encountered another bug while fixing this. The last frame was being painted first. This was due to the fact that the drawingContext of the p5.Image is used for painting and saving during frame loading for GIFs. So the last frame, the frame that was loaded last was still painted in after load. We don't have the option to load the frames in reverse order unfortunately because of the way that GIF disposal codes work. So I just manually painted the first frame back in after loading is complete.

#### PR Checklist

- [x] `npm run lint` passes


akshay-99:
This seems to work. Though I just realised that using `millis()` may cause another problem. Since rendering a single frame can take up more than one millisecond, there is a possibility that the GIFs may start in sync but then gradually move out of sync. This is because the `millis()` value can differ from one `image()` call to another.
In @owenbmcc 's sketch, ignoring the last row which was intentionally slowed down, the difference of phase becomes very visible after about five minutes of keeping it running
p5_gif_gif

I don't know if it's a significant issue, since hardly anyone will keep looking at the same thing for more than five minutes. It can however can be avoided if we use some other quantity similar to millis but which stays constant throughout the frame. `pInst._lastFrameTime` is one such quantity. It is a direct measure of time which keeps increasing as the sketch progresses. The only problem is that it belongs to the last frame and not the current one.

stalgiag:
Great suggestion @akshay-99 ! I didn't know whether to account for interframe timing differences because I thought it might be hard to know if it was intentional or not. But after seeing your example and testing it a little more, it is apparent that these time differences will only matter in instances where the sketch is left alone for a long period of time with multiple copies of the same GIF painting. The perception of events is at the frame-level but the syncing across multiples will need to be at the millisecond level.

Anyways thanks!

stalgiag:
Also, after writing the above comment. I thought of a situation where the desyncing would be pronounced more quickly with the millis approach and made an example to show (warning: may hang older computers).

[millis for timing](https://editor.p5js.org/stalgiag/sketches/Trjg5mKG3)
[lastFrame + deltaTime for timing](https://editor.p5js.org/stalgiag/sketches/sdOei3Epu)

Thanks for the fantastic catch @akshay-99

akshay-99 (approved):
Looks good!

owenbmcc:
Hi @stalgiag and @akshay-99, thanks for responding to this issue so quickly.
@stalgiag that's awesome that you're at CCNY, i've heard great things from students.

This solution looks good to me.

One concern I have is using `deltaTime` along with `_lastFrameTime` could create an issue in some cases because `deltaTime` can fluctuate while the canvas continues painting using performance.now.  That means if there's a large change in `deltaTime` it can offset the GIF animation delay.

One way to demonstrate this is to run the animated GIF and then switch tabs in the browser for a moment and then return.  Because `deltaTime` jumps ways up for one frame, using `const curTime = pInst._lastFrameTime + pInst.deltaTime;` will set curTime higher for one frame, but then go back to a lower value in the next frame, which causes `props.timeDisplayed` to be set to a higher value that it won't catch up to for a while, which means the GIF stops animating.

I think because `frameCount` only changes after a frame has been painted it is more reliable as a check to advance the next frame in the GIF.

I made an [example](https://editor.p5js.org/owenroberts/sketches/wqmGCcGXk) where you can see how the animated GIF will lag compared to a regular animation in the sketch when using the `curTime = pInst._lastFrameTime + pInst.deltaTime`.

I think in most cases this won't be an issue, but I can imagine that it would be noticeable in some cases.

stalgiag:
Hi @owenbmcc this is a good concern though, in my opinion, it begins to call into question the design of `deltaTime` more than GIF functionality.

When the GIF implementation was first designed, it was decided that GIF playback would be synced to 'realtime' rather than frame time. This decision was central to the design and was unanimously agreed upon by all of the community members that joined the discussion. See issue #3380 and PR #3740 for background info. 

So if we take [your sketch, tie the rendering of the ellipse to "realtime",](https://editor.p5js.org/stalgiag/sketches/OBPOOlOlM) and switch tabs we can begin to see how the question may be more about the functionality of 'deltaTime'. Perhaps we should make `deltaTime` tied to the [document.visibilityState](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState) we can prevent unexpected 'realtime' behavior when switching between tabs?

owenbmcc:
Hi @stalgiag thanks for that context, that makes a lot of sense, that must have been a tricky decision to make but i see the reasoning there, and I see now why my solution using `frameCount` felt hacky at the time, because everything else is based on timing, now that I understand more of how those things work.

Looking at `deltaTime` more, I don't feel that I understand enough of the process going into that to totally weigh in on how to deal with it in these cases, but i think there are more cases than changing tabs that would trigger this problem, like a longer frame painting for whatever reason, or actually a better way of demonstrating this I realized is just to toggle `noLoop` which makes the `deltaTime` spike for one frame when loop starts again, see [example](https://editor.p5js.org/owenroberts/sketches/LO8vidvA2).  I probably wouldn't write a sketch that requires toggling the loop on and off but I can imagine students using that early on when it would provide a way to pause a sketch or something before they may have learned to make a counter or something like that.

I don't know all of the functionality that `deltaTime` provides, but in the specific case of the GIF animation, one way to fix this could be to constrain the `deltaTime` to the expected duration of 1 frame, or just use that value in its place.

That might look something like:

```
const expectedFrameDuration = 1000 / pInst._targetFrameRate;  
const curTime = pInst._lastFrameTime + Math.min(pInst.deltaTime, expectedFrameDuration);
```

I can imagine this might create other issues and I'm not sure if it would fix anything else related to `deltaTime` but it does seem to work for this specific instance of the animated GIF timing, this allows the frames to advance according to the frameRate without the possible spike of the `deltaTime` in these various cases.

owenbmcc:
Thinking about this more, the method I suggested above is not a good solution, I think it's less noticeable with a GIF that has a frame based animation, but constraining `deltaTime` and using it for animation will look weird and choppy if there's a big difference between the values and I think it would be really noticeable if the animation is with a position value like the ellipse in the example.  Maybe constraining it to a range around the targetFrameRate would be better but I'm not sure, I'll try playing with that more. 

Since there are multiple cases where you would need to account for a long delay between frames creating a spike in `deltaTime`, it might need a check based on `document.visibilityState`, along with another check against an unusually large `deltaTime`.

That might make sense as a starting point, but I'm not sure how much complexity is worth introducing here.

lmccart:
thanks @stalgiag!

PR #4829, opened 2020-10-05
Fixes loadStrings so that it can load a large number of lines

### LINKED ISSUE #4828: `loadStrings` fails to load files with large number of lines (>100k)

- [ ] Color
- [ ] Data
- [ ] Dom
- [ ] Events
- [ ] Image
- [x] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:

- p5.js version: main branch as of Oct 4 2020, commit c1640a7cb11c25145e9b9dd561ba27f7b49fcde4
- Web browser and version: Chrome 85.0.4183.121 
- Operating System: macOS 10.15.2 (19C57)
- Steps to reproduce this:

```javascript
let words = [];
function preload() {
  words = loadStrings('https://raw.githubusercontent.com/dwyl/english-words/master/words.txt');
}
function setup() {
  console.log('This will never be executed.');
}
```

- Explanation:

`loadStrings` uses a call to `Array.prototype.push.apply` ([see here](https://github.com/processing/p5.js/blob/main/src/io/files.js#L270)) to populate the return array. When a large number of lines (>100k) is passed to `Array.prototype.push.apply`, many browsers (Desktop Chrome and Mobile Safari at the very least) throw a "Maximum call stack size exceeded" error which stops code execution.

- Solution:

A similar issue was noted and fixed in #2455. I will be committing a PR with a similar fix and a test case.

gregsadetsky:
Resolves #4828

 Changes:
- fixes loadStrings so that it can load files with large number of lines
- adds test case in test/unit/io/loadStrings.js with a 100k line test file
- adds manual test example loading large remote 400k line file

#### PR Checklist
- [x] `npm run lint` passes


src/io/files.js
```diff
@@ -267,7 +267,18 @@ p5.prototype.loadStrings = function(...args) {
...
         .split(/\r/);
-      Array.prototype.push.apply(ret, lines);
+
+      // same logic as in p5.RendererGL.prototype._flatten, see PR#2455
+      if (lines.length > 20000) {
+        // large number of lines can potentially lead to the browser
+        // exceeding its call stack size. better be slow but safe.
+        lines.forEach(line => {
+          ret.push(line);
+        });
```
limzykenneth, line 277: Instead of processing line by line and using forEach which have function calls that could further slow things down, we can process the lines in batches that are large but small enough to not cause a stack overflow. Here's an example implementation from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply)

```javascript
function minOfArray(arr) {
  let min = Infinity;
  let QUANTUM = 32768;

  for (var i = 0, len = arr.length; i < len; i += QUANTUM) {
    var submin = Math.min.apply(null, 
                                arr.slice(i, Math.min(i+QUANTUM, len)));
    min = Math.min(submin, min);
  }

  return min;
}

let min = minOfArray([5, 6, 2, 3, 7]);
```

gregsadetsky: Thanks for your review.

I opted for a simpler solution following the "the best code is no code at all" [mantra](https://blog.codinghorror.com/the-best-code-is-no-code-at-all/) :-)

I ran a trial to compare the current `.push.apply` strategy (which does not work for higher numbers of lines), the simpler `.forEach` approach and the batch approach suggested on MDN. Here are the results:

| Number of lines | Number of trials | `.push.apply` | `.forEach` | batch [1] |
| --- | --- | --- | --- | --- |
| 100 | 10000 | 0.0008 ms | 0.0025 ms | 0.001 ms |
| 1000 | 10000 | 0.0039 ms | 0.0189 ms | 0.0046 ms |
| 10000 | 10000 | 0.0338 ms | 0.1727 ms | 0.038 ms |
| 100000 | 10000 | 1.606 ms | 2.1648 ms | 1.6979 ms |
| 1000000 | 10000 | Error [2] | 27.0131 ms | 17.9381 ms |

The time shown is the _average time per trial_. The total time delta was measured using `(new Date).getTime()` before and after all of the trials, and then divided by the number of trials (10k for all tests) to get this average.

[1] Batch code was `Array.prototype.push.apply(ret, lines.slice(i, Math.min(i+QUANTUM, len)));` with a `QUANTUM` value of 32768, as per MDN example
[2] "Maximum call stack size exceeded" error, as expected for a high number of lines

All tests ran on Desktop Chrome version 85.0.4183.121 on macOS 10.15.2 on a 2015 MacBook Pro 2.5 GHz i7.

---

Considering that the simpler `.forEach` strategy takes 27 ms to insert 1M lines (I'm using a large-ish 31 Mb test file) and that the batch approach takes 18 ms (a modest gain), I would suggest going with `.forEach` considering that the code is simpler to understand and potentially less error-prone than the batch approach.

I'll be happy to implement whatever you deem best, of course.

limzykenneth: I agree that simpler solution should be preferred. What I'm thinking of is to replace the branching structure entirely with just the batch processing, ie. no `if` checks for the length of the input and process using `apply` or `forEach` but rather just use batch processing to handle whatever input that comes in, 10 lines or 1,000,000 lines.

I would suggest just going for `forEach` entirely without checks for using `apply` for less than 20,000 lines, but I'm not sure if batch processing will necessarily add too much complexity in this case without actually trying to implement it myself. In short I want to simplify so that there is no `if` check rather than optimize for performance.

gregsadetsky: I understand.

I based the `if (lines.length > 20000)` on a previous PR that dealt with the same issue -- https://github.com/processing/p5.js/pull/2455 -- to ensure consistency across the codebase.

The two discussed options for batch processing that work with a large number of lines are the following:

1. 
```javascript
lines.forEach(line => {
  ret.push(line);
});
```

2.
```javascript
const QUANTUM = 32768;
for (var i = 0, len = lines.length; i < len; i += QUANTUM) {
  Array.prototype.push.apply(ret, lines.slice(i, Math.min(i+QUANTUM, len)));
}
```

Let me know which one you prefer and I'll make a commit accordingly.

limzykenneth: Let's go with option 2 then. Thanks!

gregsadetsky: Done! Let me know

limzykenneth:
Great thanks!

montoyamoraga:
@all-contributors please add @gregsadetsky for code

PR #4853, opened 2020-10-14
Examples || Added an example for toString() function inside p5.Color and a typo correction

Priya-Pathak:
This adds an example that helps to better understand or visualize the working of toString() function.

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -78,10 +78,7 @@ p5.Color = function(pInst, vals) {
  *
  * <div>
  * <code>
- * createCanvas(200, 100);
- * let myColor;
- * stroke(255);
- * myColor = color(100, 130, 250);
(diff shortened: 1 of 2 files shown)


src/color/p5.Color.js
```diff
@@ -76,8 +76,18 @@ p5.Color = function(pInst, vals) {
  * </code>
  * </div>
  *
+ * <div>
+ * <code>
+ * createCanvas(200, 100);
+ * let myColor;
```
lmccart, line 82: could this line be combined with line 84?

Priya-Pathak: Sure, I will make the changes.
Thanks for the feedback.

lmccart:
thanks! @all-contributors please add @Priya-Pathak for examples

PR #4856, opened 2020-10-15
Fix error when handling error in MediaElement.play

daniel-michel:
Because of the anonymous callback function passed to `promise.catch` `this` was undefined in the function and thus threw the error: `Uncaught (in promise) TypeError: Cannot read property 'src' of undefined`.

 Changes:

I converted the function into an arrow function so `this` keeps the value from outside the function.

#### PR Checklist

- [ ] `npm run lint` passes


lmccart:
thanks!

montoyamoraga:
@all-contributors please add @daniel-michel for code

PR #4865, opened 2020-10-20
Make camera() parameters optional

gohai:
Changes:
The documentation for camera() suggests that all 9 parameters are optional, but passing some - but not all - does break currently. This proposed change lets you call camera() with three parameters for changing only the camera position, or six parameters for camera position & target (without having to specify the not-so-intuitive up vector also). Shouldn't affect current behavior inadvertently (e.g. when calling camera() w/o params).

- [x] `npm run lint` passes
- [ ] Inline documentation is included / updated
- [ ] Unit tests are included / updated (_unit tests all seem to pass_)

### DIFF
--- a/src/webgl/p5.Camera.js
+++ b/src/webgl/p5.Camera.js
@@ -882,13 +882,17 @@ p5.Camera.prototype.camera = function(
   this.eyeY = eyeY;
   this.eyeZ = eyeZ;
 
-  this.centerX = centerX;
-  this.centerY = centerY;
-  this.centerZ = centerZ;
+  if (typeof centerX !== 'undefined') {
+    this.centerX = centerX;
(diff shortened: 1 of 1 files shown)


limzykenneth:
Hi @gohai, we usually require an issue be opened and discussed there before a PR should be filed. Can you open an issue for this in order for us to focus discussion of the problem there first? Thanks.

lmccart:
@limzykenneth I'm in favor of merging this one, it looks like a straightforward enough fix. What do you think?
@gohai In the future, it would be great if you could open an issue first to ensure the community has an opportunity to weigh in. Thanks!

gohai:
Hi - sorry @limzykenneth @lmccart, forgot about responding to that earlier message, will do in future occasions

PR #4877, opened 2020-10-26
Enter key fix for _onkeypress

### LINKED ISSUE #4871: When pressing enter, the key value is set to empty string until enter is released

- [ ] Color
- [ ] Data
- [ ] Dom
- [ x] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:

- p5.js version: 1.1.9

In this [example](https://editor.p5js.org/owenroberts/sketches/g5LthBL4G), focus the canvas and then press the enter key, hold for a moment, and release.  The key value will be blank (empty string) until enter is released.

This is kind of minor but it came up with a student who couldn't figure out why their sketch wasn't working when using `if (key == 'Enter')`.

I checked and it looks like the the `_onkeydown` and `_onkeypress` functions set the `key` property differently.

[_onkeydown](https://github.com/processing/p5.js/blob/main/src/events/keyboard.js#L180):
`this._setProperty('key', e.key || String.fromCharCode(e.which) || e.which);`

[_onkeypress](https://github.com/processing/p5.js/blob/main/src/events/keyboard.js#L285)
`this._setProperty('key', String.fromCharCode(e.which));`

The problem is that `_onkeypress` fires after `_onkeydown` and the value `String.fromCharCode(e.which)` is empty string for the value 13 (the key code for Enter), so `key` is set to empty string until `_onkeyup` fires.

I could make a pull request that substitutes the line used by `_onkeydown` and `_onkeyup` for the one in `_onkeypress` but I wasn't sure if this was intentional.  I don't see anything in the documentation that suggests it is but want to check.

owenbmcc:
Resolves #4871.

Changes:
Updates _onkeypress to match _onkeydown and _onkeyup setting the global key property.  Fixes a bug where the Enter key is saved as empty string in the global key property when pressed and before released.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/events/keyboard.js
+++ b/src/events/keyboard.js
@@ -282,7 +282,7 @@ p5.prototype._onkeypress = function(e) {
     return;
   }
   this._setProperty('_lastKeyCodeTyped', e.which); // track last keyCode
-  this._setProperty('key', String.fromCharCode(e.which));
+  this._setProperty('key', e.key || String.fromCharCode(e.which) || e.which);
 
   const context = this._isGlobal ? window : this;
   if (typeof context.keyTyped === 'function') {


limzykenneth:
Thanks!

PR #4929, opened 2020-12-03
docs: fixed some typos/spellings

covalentbond:
Corrected some spelling/typos

### DIFF
--- a/contributor_docs/friendly_error_system.md
+++ b/contributor_docs/friendly_error_system.md
@@ -131,7 +131,7 @@ function setup() {
 }
 /* 
 FES will show:
-p5.js says: It seems that you may have accidently written "colour" instead of "color" (on line 2 in sketch.js [http://localhost:8000/lib/empty-example/sketch.js:2:3]).
+p5.js says: It seems that you may have accidentally written "colour" instead of "color" (on line 2 in sketch.js [http://localhost:8000/lib/empty-example/sketch.js:2:3]).
 
 Please correct it to color if you wish to use the function from p5.js (http://p5js.org/reference/#/p5/color)
 */
@@ -146,7 +146,7 @@ function preLoad() {
 }
 /* 
 FES will show:
-p5.js says: It seems that you may have accidently written preLoad instead of preload.
+p5.js says: It seems that you may have accidentally written preLoad instead of preload.
 
 Please correct it if it's not intentional. (http://p5js.org/reference/#/p5/preload)
 */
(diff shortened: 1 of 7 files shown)


contributor_docs/friendly_error_system.md
```diff
@@ -131,7 +131,7 @@ function setup() {
 }
 /* 
 FES will show:
-p5.js says: It seems that you may have accidently written "colour" instead of "color" (on line 2 in sketch.js [http://localhost:8000/lib/empty-example/sketch.js:2:3]).
+p5.js says: It seems that you may have accidentally written "colour" instead of "color" (on line 2 in sketch.js [http://localhost:8000/lib/empty-example/sketch.js:2:3]).
```
akshay-99, line 134: Could you make the change in [these](https://github.com/processing/p5.js/search?q=accidently) places too? It's my bad, but quite ironic how a message that's asking the user to correct a spelling mistake in their code itself has a spelling mistake :smile:

covalentbond: So, I actually spent hours trying to do that. But I was unable to edit the .json and .js files because of my workspace End of line settings `LF` and `CRLF`.
It will cause test to fail, if I do it.
If you are one of the maintainers you can add a commit in this pull request itself

akshay-99: Could you add more details to the error? Do only the tests under `misspelling detection` fail or do other tests too? The regex [here](https://github.com/processing/p5.js/blob/3d3e6bb1684c1946bf384b8b6ccf0bf2d9a61cfe/test/unit/core/error_helpers.js#L394) is matching only a part of the string so I don't see how a difference of `\n` would make a difference. 
I just want to see if there is a problem with how these tests are written.

covalentbond: This link might help you https://github.com/processing/p5.js/runs/1491444900

image

covalentbond: Prior to the current forced push commit. This was commited https://github.com/processing/p5.js/commit/00c2d3450032ca3819b62a69d17ac4801bdcdb0b where translation.json file changes were there. But the test were failing

akshay-99: Ah I see. You would have to update both the json and the test files before running the tests.
Test file to be updated is [here](https://github.com/processing/p5.js/blob/main/test/unit/core/error_helpers.js)

[00c2d34](https://github.com/processing/p5.js/commit/00c2d3450032ca3819b62a69d17ac4801bdcdb0b) only changes the json hence the mismatch

expected '\n🌸 p5.js says: It seems that you may have **accidentally** written "stroke" ....' to match /It seems that you may have **accidently** written "strok"/

PR #4967, opened 2020-12-25
Added a friendly error for p5.graphics.size()

### LINKED ISSUE #4956: size() throws error with p5.Graphics

#### Details about the bug:

- p5.js version: 1.1.9
- Web browser and version: latest Firefox and Chrome
- Operating System: macOS 14
- Steps to reproduce this:

`p5.Graphics.size()` throws this error:
```
TypeError: setting getter-only property "canvas"
```

This is specific to p5.Graphics. Other p5.Element-derived objects can resize as expected. 

[Here is an example sketch.](https://editor.p5js.org/stalgiag/sketches/JSwFSnISZ)

highonweb:
Resolves #4956 

 Changes:

- invokes a Friendly error if the size method of a canvas is called with one or more arguments
- It advises to use the resizeCanvas method instead

#### PR Checklist

- [x] `npm run lint` passes


stalgiag:
Thanks @highonweb !

This brings up a quick question. @lmccart are you advocating for not handling resizing of canvases at all with `size()`? I am ok with this but then I think the logic that attempts to handle this (but does not work currently) [on line 2126](https://github.com/processing/p5.js/pull/4967/files#diff-40df360d05e2e2a0928286ee2892aa9d2acd710c8c7f871c8ae6f7229e59e24dL2126) through 2140 should be removed with the addition of this friendly error.

highonweb:
Should I remove that block of code @stalgiag @lmccart

stalgiag:
@highonweb I think it is fine to just remove it. My guess is that the p5.dom library wasn't in strict mode when it was separate and that's what allowed this to work in the past.

`resizeCanvas()` should be a fine workaround. Thanks!

highonweb:
I removed it @stalgiag

src/dom/dom.js
```diff
@@ -2122,29 +2130,6 @@ p5.Element.prototype.size = function(w, h) {
...
-        this.elt.setAttribute('width', aW * this._pInst._pixelDensity);
-        this.elt.setAttribute('height', aH * this._pInst._pixelDensity);
-        this.elt.style.width = aW + 'px';
-        this.elt.style.height = aH + 'px';
-        this._pInst.scale(this._pInst._pixelDensity, this._pInst._pixelDensity);
-        for (prop in j) {
-          this.elt.getContext('2d')[prop] = j[prop];
-        }
-      } else {
-        this.elt.style.width = aW + 'px';
```
stalgiag, line 2142: Looks like you lost the else part of the conditional here.

src/dom/dom.js
```diff
@@ -2113,6 +2113,14 @@ p5.Element.prototype.size = function(w, h) {
   if (arguments.length === 0) {
     return { width: this.elt.offsetWidth, height: this.elt.offsetHeight };
   } else {
+    if (this.elt instanceof HTMLCanvasElement) {
+      p5._friendlyError(
+        'Try using the method resizeCanvas for canvas instead of size method check this for reference',
```
stalgiag, line 2118: Is there a way that we could be more direct with this language? `Use resizeCanvas() instead of size() when changing the size of a canvas or graphics object.`?

stalgiag (changes requested):
Lost the else part of the HTMLCanvasElement conditional and a slight rewording request on the friendly error.

src/dom/dom.js
stalgiag: Note that you don't need to take my suggestion verbatim. It is just meant to illustrate what I mean by direct.

highonweb: yeah ok I will make the required changes

lmccart:
> Thanks @highonweb !
> 
> This brings up a quick question. @lmccart are you advocating for not handling resizing of canvases at all with `size()`? I am ok with this but then I think the code that attempts to handle this (but does not work currently) [on line 2126](https://github.com/processing/p5.js/pull/4967/files#diff-40df360d05e2e2a0928286ee2892aa9d2acd710c8c7f871c8ae6f7229e59e24dL2126) through 2140 should be removed with the addition of this friendly error.

Ah, I didn't realize this is what we were doing. I guess it does make sense for `size()` to work. But would it be simpler to pass the call to `resizeCanvas()` under the hood?

stalgiag:
>  I guess it does make sense for `size()` to work. But would it be simpler to pass the call to `resizeCanvas()` under the hood?

I think this would work fine. The only issue that comes to mind would be canvases made with `createElement('canvas')`. This is an edge case though and should be discouraged since I do not know why someone would want to make a non-p5 HTML canvas inside p5.

@highonweb want to give it a try?

src/dom/dom.js
highonweb: Made both the changes @stalgiag

highonweb:
> > I guess it does make sense for `size()` to work. But would it be simpler to pass the call to `resizeCanvas()` under the hood?
> 
> I think this would work fine. The only issue that comes to mind would be canvases made with `createElement('canvas')`. This is an edge case though and should be discouraged since I do not know why someone would want to make a non-p5 HTML canvas inside p5.
> 
> @highonweb want to give it a try?

yeah I will call that function internally

highonweb:
I have one doubt 
can't we do resizing just like this

```
pg = createGraphics(pgWidth, pgHeight);
pg.width = newWidth;
pg.height = newHeight;
```
checkout this issue
https://github.com/processing/p5.js/issues/2064

stalgiag:
@highonweb yes this makes sense instead of calling `resizeCanvas()`. Good research! Especially because `resizeCanvas()` with p5.Graphics broke on Firefox after updating to version 84.

highonweb:
So can I change the friendly error to do so

stalgiag:
Or if you would like you can try doing that inside the size function. So that 
```js
p5Graphics.size(100, 100);
```
can be a shorthand for:
```js
p5Graphics.width = 100;
p5Graphics.height = 100;
```

highonweb:
yeah that can be done

lmccart:
@highonweb thanks for your work on this! are you still interested in finishing this up?

highonweb:
I deleted the forked repo accidentally

Issue #4852, opened 2020-10-14
Add static versions of vector functions


weslord:
#### Most appropriate sub-area of p5.js?

- [ ] Accessibility (Web Accessibility)
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Feature enhancement details:

Sometimes I want to use a vector's normalized form in an equation, but I don't want to mutate my original vector. In these cases it would be nice to have a static version of `.normalize()`.

For example:
```javascript
// current method
const normalizedVector = myVector.copy().normalize()

// proposed method
const normalizedVector = p5.Vector.normalize(myVector)
```

This would follow the pattern of `.mag()`, `.add()`, and other existing vector functions that have both a mutative prototype version and a static version.

I'd be happy to submit a PR implementing this enhancement, with tests and docs, if you think this would be a valuable addition.

aferriss:
This sounds like a good addition to me. There are already a number of static vector math functions available. My guess is that a few of them were just overlooked. For instance, we're also missing one for `p5.Vector.magSq()`.

You can see the static versions of these [functions here](https://github.com/processing/p5.js/blob/3f0b2f0fe575dc81c724474154f5b23a517b7233/src/math/p5.Vector.js#L2247)

limzykenneth:
I personally don't mind this addition but I do wonder how this fit into the accessibility requirement for adding new features.

aferriss:
I think that this could fit into the accessibility requirement in that it makes the library more predictable and uniform for people new to programming / creative coding / vector math. The current scattershot state of vector math with half of the functions containing static versions, and half that don't feels to me like static functions were just added sporadically as different contributors needed to use them. 

Perhaps it's out of scope for this issue but if we decide not to add these functions, I wonder if the errors / warnings could be more helpful if you tried to use a non-existing static function. Currently if you try to use `p5.Vector.normalize()` FES will tell you that `p5.Vector.normalize` doesn't exist, which is sort of true, but confusing since it does exist in the non-static form.

I went through to take an inventory of the existing static functions, and the missing ones and there's actually quite a lot missing. There are some functions where it doesn't makes sense to have a non-static version (random2D / random3D and fromAngle / fromAngles ) so keep that in mind as well. 

[...]

PR #4972, opened 2020-12-31
Static version of p5.Vector.rotate

siv2r:
Addresses #4852

 **Changes:**
- added static function counterpart of [`rotate()`]
- added example of static implementation of [`rotate()`]

**Changes need to implemented:**
- [ ] start readme file for `src/math`
- [ ] add other static functions (name of those functions will be added here)

#### PR Checklist
- [x] `npm run lint` passes

@limzykenneth @lmccart I have updated the examples of the [`rotate()`] function in the reference page for its static implementation. Is there anything else that should be done for [`rotate()`]?

Once static implementation of `rotate()` looks good. I will start adding the same for other functions. Hopefully, collaborate with @weslord and @covalentbond if they are interested. Finally, start a readme for `src/math`.  

[`rotate()`]: https://p5js.org/reference/#/p5.Vector/rotate

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -2139,6 +2139,34 @@ p5.Vector.mult = function mult(v, n, target) {
 };
 
 /**
+ * Rotates the vector (only 2D vectors) by the given angle, magnitude remains the same and returns a new vector.
+ */
+
+/**
+ * @method rotate
+ * @static
+ * @param  {p5.Vector} v
+ * @param  {Number} angle angle of rotation
+ * @param  {p5.Vector} [target] the vector to receive the result (Optional)
+ */
+
+p5.Vector.rotate = function rotate(v, a, target) {
+  if (!target) {
+    target = v.copy();
+    if (arguments.length === 3) {
+      p5._friendlyError(
+        'The target parameter is undefined, it should be of type p5.Vector',
+        'p5.Vector.rotate'
+      );
+    }
+  } else {
+    target.set(v);
+  }
+  target.rotate(a);
+  return target;
+};
+
+/**
  * Divides a vector by a scalar and returns a new vector.
  */
 
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -1482,6 +1482,19 @@ p5.Vector.prototype.setHeading = function setHeading(a) {
  * </code>
  * </div>
  *
+ * <div class="norender">
+ * <code>
+ * // static function implementation
+ * let v = createVector(10.0, 20.0);
+ * // v has components [10.0, 20.0, 0.0]
+ * let rotated_v = p5.Vector.rotate(v, HALF_PI);
+ * console.log(rotated_v);
+ * // rotated_v's components are set to [-20.0, 9.999999, 0.0]
+ * console.log(v);
+ * // v's components remains the same (i.e, [10.0, 20.0, 0.0])
+ * </code>
+ * </div>
+ *
  * <div>
  * <code>
  * let angle = 0;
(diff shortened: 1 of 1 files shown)


src/math/p5.Vector.js
```diff
@@ -2138,6 +2151,34 @@ p5.Vector.mult = function mult(v, n, target) {
...
+ */
+
+p5.Vector.rotate = function rotate(v, a, target) {
+  if (!target) {
+    target = v.copy();
+    if (arguments.length === 3) {
+      p5._friendlyError(
+        'The target parameter is undefined, it should be of type p5.Vector',
+        'p5.Vector.rotate'
+      );
```
lmccart, line 2173: This friendly error message is a little confusing to me, can you clarify how this works? If you've just copied `target` from `v` above, wouldn't it be defined then?

src/math/p5.Vector.js
```diff
@@ -2138,6 +2151,34 @@ p5.Vector.mult = function mult(v, n, target) {
...
+ * @static
+ * @param  {p5.Vector} v
+ * @param  {Number} angle
+ * @param  {p5.Vector} [target] the vector to receive the result (Optional)
+ */
+
+p5.Vector.rotate = function rotate(v, a, target) {
+  if (!target) {
+    target = v.copy();
+    if (arguments.length === 3) {
```
lmccart, line 2169: Would it make sense to put this check first? I'm getting a little confused reading this in the flow here.

siv2r: I added the check here since other static functions having a `target` variable follows the same fashion. Here are those functions [mult](https://github.com/processing/p5.js/blob/main/src/math/p5.Vector.js#L2126), [div](https://github.com/processing/p5.js/blob/main/src/math/p5.Vector.js#L2126), [lerp](https://github.com/processing/p5.js/blob/main/src/math/p5.Vector.js#L2251). I also feel that it would be better to have the checks later on (in the else statement). Shall I make the change?

lmccart: Ah I see. I think so. @limzykenneth do you see any reason it would be written in this order?

limzykenneth: No particular reason I can see, it is better in the `else` block. The checks could be better written though with the first `if(!target)` check be an arguments length check while the friendly error check be an `instanceof p5.Vector` check in the `else` block.

siv2r: I have made some changes to the checks. Kindly review it.
Thanks

limzykenneth: The new change isn't quite what I mean. It's a bit redundant now with two `copy()`. The friendly error check can just be in the else block of the argument check (ie. nested) without being a separate block.

limzykenneth: That's closer

siv2r: I moved the negation outside the brackets and updated the new changes. How does it look now?

limzykenneth: I'm not 100% sure about the wording of the friendly error (can maybe be even friendlier?) but other than that, the implementation is more or less what I have in mind.

siv2r: I came up with these three possible error statements.
1. Can't assign the resultant vector to the target parameter. It should be of type p5.Vector
2. The target parameter is an invalid 'instanceof' p5.Vector. It should be of type p5.Vector
3. Invalid assignment to the target parameter. It should be of type p5.Vector

The first statement feels friendlier out of the three. We could also replace the 'resultant vector' in the first statement with 'rotated vector', 'vector', or 'vector object'. Thoughts on this?

limzykenneth: Actually I think it's fine for now since the other functions uses similar wording as well.

limzykenneth (approved):

siv2r:
changing the title since @weslord is working on other static functions.

siv2r:
@limzykenneth @lmccart Thanks for taking the time to review this PR!! :cookie:

PR #4989, opened 2021-01-10
Fix inline docs of int(), windowResized() and other event handler functions

### LINKED ISSUE #4988: [docs] fix inline docs of windowResized and other event handler functions etc.
Most appropriate sub-area of p5.js?
- [x] Accessibility (Web Accessibility)

My main intension is to supplement the machine-readable documentation that some other projects (such as [p5.ts](https://github.com/p5-types/p5.ts)) rely on.

---

I've found some points to be fixed in the inline documentation.
Will be happy to create a pull request (not very familiar with it though; will read the contributor docs).

### Description and arguments of `windowResized()`

According to the actual code, the two below can be added:

- Explanation about `return false`,
same as other event handlers such as `mousePressed()`.
Actually this can be any falsy value (while only `false` is valid in other event handlers), but maybe it's better to write it consistently with other event handler functions.
- `@param  {Object} [event]`,
which is an `UIEvent`.
(*ADD*: same for keyboard event handlers such as `keyPressed()`, which lack `@param` tag even though they accept `KeyboardEvent` argument)

### Return type of `mousePressed()` and other event handlers

`mousePressed()` and other similar functions may return either void or `boolean`.
So I'd like to add something like:

```
@return  {*} (optional) `false` to prevent default behavior.
```

Not sure if the type `{*}` is correct (it's also used in `random()` with an `Array` argument);
In TypeScript it would be `any` or `void | boolean`, and I suppose the latter cannot be described in YUIDoc.

### Arguments of `int()` (conversion function)

Not really related to the above, but
`@param {Integer} [radix]` is missing for the case if the first argument is an `Array`.

fal-works:
Resolves #4988

 Changes:

Following improvements in the inline docs.

- `int()`: add missing `@param [radix]`
- `windowResized()`, `keyPressed()` and other keyboard-related functions: add missing `@param [event]`

*EDIT: Following changes have been reverted (see #4988).*

- `windowResized()`: add description about `return false`
- `windowResized()` and other keyboard/mouse/touch-related functions: add missing `@return`

#### PR Checklist

- [x] `npm run lint` passes
- [x] [Inline documentation] is included / **updated**

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2347,6 +2347,130 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "myselfhimself",
+      "name": "Jonathan-David Schröder",
+      "avatar_url": "https://avatars.githubusercontent.com/u/1265346?v=4",
+      "profile": "https://github.com/myselfhimself",
+      "contributions": [
+        "ideas",
+        "code"
+      ]
+    },
+    {
+      "login": "ShenpaiSharma",
+      "name": "Shubham Kumar",
+      "avatar_url": "https://avatars.githubusercontent.com/u/47415702?v=4",
+      "profile": "https://github.com/ShenpaiSharma",
+      "contributions": [
+        "code"
+      ]
+    },
+    {
+      "login": "nakednous",
+      "name": "Jean Pierre Charalambos",
+      "avatar_url": "https://avatars.githubusercontent.com/u/645599?&v=4",
+      "profile": "https://github.com/nakednous",
+      "contributions": [
+        "code",
+        "tool"
+      ]
+    },
+    {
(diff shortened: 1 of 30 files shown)


lmccart:
@fal-works are you still interested in working on this? Is this PR ready to be submitted for merge?

fal-works:
I'm sorry I left this for a while!
I'm still interested, but it's not yet ready; will work on this in a few days.

fal-works:
Now it should be ready.
Sorry for the dirty commit logs; please see "Files changed" for changes

lmccart:
@limzykenneth I know you've been tracking this issue, would you like to take a look over and merge if it's ready? this all looks good to me.

limzykenneth:
Looks good, thanks.

Issue #4866, opened 2020-10-21
texture() ignoring uvs for .obj file


kieranbrowne:
Hi there,

I'm trying to import a basic 3d .obj file and texture into p5js, but the texture appears to be ignoring the uv information in the obj file and misplacing the texture. I've written the equivalent program in processing 3 and it display correctly, so I can be pretty sure it's not an issue with the obj and texture files.

Am I doing something silly, or is this a genuine bug?

#### Most appropriate sub-area of p5.js?

- [ ] Accessibility (Web Accessibility)
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Details about the bug:

- p5.js version: 1.1.9
- Web browser and version: Chrome: 86.0.4240.80 and Firefox: 81.0.2
- Operating System: MacOSX 10.15.6
- Steps to reproduce this: Load an obj file and texture into p5 with loadModel and loadImage.
The p5 code with all the necessary assets can be found here: <https://editor.p5js.org/kbrowne/sketches/o_PYbAgLl>

And here is the equivalent processing 3 code which textures correctly:
```
PShape mod;
PImage tex;

void setup() {
  size(400, 400, P3D);

  mod = loadShape("TreeFinger_L_03.obj");
  tex = loadImage("dif_arm.png");
  mod.setTexture(tex);
  noStroke();
}

void draw() {
  background(220);
  translate(width/2, height/2);
  scale(4);
  rotateX(frameCount * 0.01);
  rotateY(frameCount * 0.01);

  shape(mod, 0, 0);
}
```

stalgiag:
Hi! I know that the [UVs are parsing correctly with some files](https://editor.p5js.org/stalgiag/sketches/ddWWhe1b1) but maybe there is a need to compare the UV parsing here to what is done in Processing to provide greater suport.

Anyone is welcome to take a look.

kieranbrowne:
So I figured out that the y values for the uvs are inverted.

It's possible to reverse them like so:
```js
  for(let i=0; i< mod.uvs.length; i++) {
    mod.uvs[i][1] = 1.-mod.uvs[i][1];
  }
```

And this fixes the problem. See updated sketch: <https://editor.p5js.org/kbrowne/sketches/o_PYbAgLl>

[...]

PR #5017, opened 2021-01-28
#4866 V texture coordinate inversion on OBJ loadModel

myselfhimself:
Testing this as WebGL requires a flipped/inverted V texture coordinate

this is untested for now, unsure it resolves something

### DIFF
--- a/src/webgl/loading.js
+++ b/src/webgl/loading.js
@@ -222,8 +222,10 @@ function parseObj(model, lines) {
         loadedVerts[tokens[0]].push(vertex);
       } else if (tokens[0] === 'vt') {
         // Check if this line describes a texture coordinate.
-        // It will have two numeric parameters.
-        const texVertex = [parseFloat(tokens[1]), parseFloat(tokens[2])];
+        // It will have two numeric parameters U and V (W is omitted).
+        // Because of WebGL texture coordinates rendering behaviour, the V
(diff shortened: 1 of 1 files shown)


myselfhimself:
Tested, see https://github.com/processing/p5.js/issues/4866#issuecomment-769460136
This copies Processing's UV's V dimension inverting behaviour on OBJ model loading. It seems this inverting had been forgotten in the p5.js implementation so far.

myselfhimself:
This could mergeable with some note to be added in a changelog and maybe a note on UV interpretation in the documentation...
Flipping the V dimension (because Y is flipped in WebGL or so), still does not solve all OBJ UV mapping issues, with sometimes images being flipped horizontally and/or vertically without plainly understanding why... If strict following for p5js of the Processing behaviour and API were not a must, maybe some helper flags or functions could allow developers to flip those coordinates easily. Eg. on loadModel(<path>, <normalize scale>, flipU boolean, flipV boolean)..

stalgiag:
@myselfhimself Thank you so much! I really like the idea of adding `flipU` and `flipV`. Perhaps they could be overload parameters in `loadModel`? We don't have any limitations as far as following Processing and I think this could be useful for quick fixes without needing to reexport model files. That can be discussed in another issue if you would like.

@all-contributors please add @myselfhimself for code

PR #5027, opened 2021-02-02
Fix/reset color mode in lights function

vulongphan:
Resolves #4974 

 Changes:

Update `light.js` file [here](https://github.com/vulongphan/p5.js/blob/fix/reset-color-mode-in-lights-function/src/webgl/light.js#L410)

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes
- [x] [Inline documentation] is included / updated
- [ ] [Unit tests] are included / updated

[Inline documentation]: https://github.com/processing/p5.js/blob/main/contributor_docs/inline_documentation.md
[Unit tests]: https://github.com/processing/p5.js/tree/main/contributor_docs#unit-tests

Issue #4998, opened 2021-01-17
Old version of Prettier is incompatible with VSCode extension 


weslord:
#### Most appropriate sub-area of p5.js?

- [ ] Accessibility (Web Accessibility)
- [x] Build tools and processes
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Other (specify if possible)
-> Developer tools / environment

#### Details about the bug:

- p5.js version: 1.1.9
- Web browser and version: N/A
- Operating System: Windows, MacOS
- Steps to reproduce this:

1. Install dependencies with `npm ci`
2. Open project directory in Visual Studio Code
3. Install recommended extensions as prompted by `.vscode/extensions.json`
4. Run command `Format Document` in VS Code

The plugin "Prettier Formatter for Visual Studio Code" will display the following error/warning:
> Your project is configured to use an outdated version of prettier that cannot be used by this extension. Upgrade to the latest version of prettier.

The tricky thing is that Prettier 2.0 has come out in the meantime. They've introduced a few breaking changes to the formatting, such as using trailing commas and inserting a space before the parentheses in `function () {}`. These changes require a bunch of small formatting changes to nearly all of the modules in the project.

Unfortunately, despite the error message, the VS Code `Format Document` command proceeds to format anyways, but it uses a recent version of Prettier rather than the local project version. Since the linting step uses the local version (`1.10.2`), this means running `Format Document` can introduce a bunch of formatting problems and cause `npm run grunt` to fail. (NOTE: This might just be due to how I have VS Code set up personally, falling back to my global settings or something? Can anyone else confirm if this is universal?)

If someone can confirm that this isn't just my system, I think it's worth updating Prettier to save new contributors the potential frustration of formatting errors that are caused by their formatting tools. Thoughts?

#### References:

https://github.com/processing/p5.js/blob/main/.vscode/extensions.json#L8
https://github.com/processing/p5.js/blob/main/package.json#L81
https://github.com/prettier/prettier/issues/3845

weslord:
See https://github.com/processing/p5.js/compare/main...weslord:prettier-2 for branch with potential PR.
[...]

PR #5049, opened 2021-02-18
update lights function

vulongphan:
Resolves #5027

 Changes:

Update `light.js` file

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/light.js
+++ b/src/webgl/light.js
@@ -6,6 +6,7 @@
  */
 
 import p5 from '../core/main';
+import * as constants from '../core/constants';
 
 /**
  * Creates an ambient light with a color. Ambient light is light that comes from everywhere on the canvas.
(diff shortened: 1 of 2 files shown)


vulongphan:
@limzykenneth @stalgiag I have checked the `prettier` version in the `package.json` file and it is the same version as the one in `main`. I have also created this PR which is directly branched from `main` but the issue still persists. Do you have any suggestions?

limzykenneth:
@vulongphan I'm thinking more in terms of the prettier version that is used by other tools on your system. The VSCode plugin for example doesn't work with our version of prettier so maybe something like that packages it's own prettier or using prettier that is installed globally, etc.

vulongphan:
@limzykenneth I really do agree with your speculation. It is true that the prettier version of VSCode I am using does not work with our prettier version. Hence, I have made some appropriate changes in formating and now lint has passed.

Thank you!

src/webgl/light.js
```diff
@@ -409,8 +409,20 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
  */
 p5.prototype.lights = function() {
   this._assert3d('lights');
-  this.ambientLight(128, 128, 128);
-  this.directionalLight(128, 128, 128, 0, 0, -1);
+  if (this._colorMode !== 'rgb') {
```
lmccart, line 412: could you update this to `constants.RGB`? here's an example of how it's done by importing constants and then calling it in [this file](https://github.com/processing/p5.js/blob/bb381febc246c7599381801ae314b3d40330d9d8/src/webgl/3d_primitives.js#L1078)

lmccart: same thing on line 416

stalgiag:
Thanks for working through this @vulongphan . I see that you run into the same problem as #4998 . Really impressed that you powered through.

@lmccart 's point about the color mode constants reminded me that we only have two non-RGB color modes to think about and they both work similarly when making white, so we can actually avoid the whole array copying stuff by just doing something akin to this:

```js
  this._assert3d('lights');
  if (this._colorMode === constants.RGB) {
    this.ambientLight(128, 128, 128);
    this.directionalLight(128, 128, 128, 0, 0, -1);
  } else {
    const maxBright = this._colorMaxes[this._colorMode][2];
    this.ambientLight(0, 0, maxBright);
    this.directionalLight(0, 0, maxBright, 0, 0, -1);
  }
```

Thanks again for sticking around for this entire saga!

vulongphan:
@stalgiag @lmccart I have incorporated both of your feedback. Lint has passed as well. Please check the most recent push [5c0cdaa](https://github.com/processing/p5.js/pull/5049/commits/5c0cdaa362998d03a620eaf44b52b8b3335e0243)

Thank you!

stalgiag:
Amazing! Thank you for your perserverance @vulongphan !

@all-contributors please add @vulongphan for code

PR #5072, opened 2021-03-08
Allows applyMatrix() to take in an array as its first argument

### LINKED ISSUE #5067: applyMatrix() should be able to take in an array of values
This feature would help simplify the process of dealing with matrix data that is already represented by an array of values, allowing the user not to have to worry about passing each value of the matrix individually or knowing how to spread the array.

- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

`applyMatrix()` should be able to take a an array of 6 or 16 values (corresponding to a 2x3 or 4x4 matrix) in addition to the separate parameters it does currently. For a practical example, I'm currently working on a way to easily access pose data from the WebXR API for p5.xr, as can be seen here stalgiag/p5.xr#115 . Pose data is stored on XRRigidTransform.transform.matrix as a Float32Array, holding the 16 values corresponding to the 4x4 matrix. Being able to pass the array directly would allow me to simplify the necessary function calls to draw tracked controllers down to this:
```js
const right = getXRInput(RIGHT);
if (right != null) {
    push();
    fill(255, 255, 255);
    applyMatrix(right.pose);
    box(.15);
    pop();
}
```

Tagging @stalgiag by request.

msub2:
Resolves #5067 

 Changes:

This updates the applyMatrix function in transform.js to also take in an array as its first argument.

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/transform.js
+++ b/src/core/transform.js
@@ -137,13 +137,25 @@ import p5 from './main';
  * </code>
  * </div>
  *
+ * <div>
+ * <code>
+ * function draw() {
+ *   background(200);
+ *   let testMatrix = [1, 0, 0, 1, 0, 0];
+ *   applyMatrix(testMatrix);
+ *   rect(0, 0, 50, 50);
+ * }
+ * </code>
+ * </div>
+ *
  * @alt
  * A rectangle translating to the right
  * A rectangle shrinking to the center
  * A rectangle rotating clockwise about the center
  * A rectangle shearing
+ * A rectangle in the upper left corner
  */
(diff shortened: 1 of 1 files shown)


stalgiag:
Great thank you! This is a small change and is generally useful so I think we can safely merge.

@all-contributors please add @msub2 for code and documentation

PR #5076, opened 2021-03-10
fix: #4993 user-reassigned pixel data

### LINKED ISSUE #4993: "pixels" array of p5.Image cannot be reassigned while canvas one can

- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:

- p5.js version: 1.2.0
- Web browser and version: Firefox 84.0.2 (should apply to all supported browsers)
- Operating System: Linux Pop_OS 20.10
- Steps to reproduce this:

Visit [this example sketch](https://editor.p5js.org/limzykenneth/sketches/0zRWURmXM). The initial expected behaviour is a blank canvas because the `img.pixels` array has been assigned with the `.map` function to be all 0 but the image is still drawn. Replacing the `.map` with a `for` loop works as expected.

Filing this as a bug because this behaviour is different from the canvas itself (see by uncommenting code at the end of the `draw()` function), which reassigning the `pixels` array with `.map` still works.

The problem seems to be outdated variable referencing in p5.Image, possibly affecting p5.Graphics as well but I didn't test.

vsquareg:
Resolves #4993 
###  Issue:
Updating the `pixels` property of objects does not affect the inherent data/pixels used to draw on the canvas.

###  Changes:
- Check if `pixels` property (available to the user for change, with documentation) is sized appropriately for the object.
- Assign this pixel data to `imageData.data`.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -387,6 +387,18 @@ p5.Renderer2D.prototype.updatePixels = function(x, y, w, h) {
   w *= pd;
   h *= pd;
 
+  // checking if user assigned pixel array is appropriate for this object
+  if (pixelsState.pixels.length !== pixelsState.imageData.data.length) {
+    throw new Error(`
+      User assigned pixel array does not match image/canvas dimensions.
+    `);
+  }
+
+  // assigning the pixels to imageData.data (given the above check passes)
+  for (let i = 0; i < pixelsState.pixels.length; i++) {
+    pixelsState.imageData.data[i] = pixelsState.pixels[i];
+  }
+
   if (this.gifProperties) {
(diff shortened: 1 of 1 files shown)


limzykenneth:
Hi @vsquareg, this implementation is going to be too slow especially with larger canvases as it needs to iterate through the whole pixels array whenever `updatePixels()` is called. In general I would prefer general implementation ideas be worked out in the issues first (and often the issue would be assigned so a specific individual) before a PR is filed. I'm closing this for now and we can continue discussions over in the issue.

PR #5098, opened 2021-03-17
fixed-select-option-using-label-issue

### LINKED ISSUE #4854: Label is not clickable on Radio Buttons with createRadio()

- [x] Dom

In radiobutton made with `createRadio()`, when the label is clicked, the RadioButton doesn't get selected.

Current Functionality: 
```html
<input type="radio" value="value-of-option-1" name="radioOption">
<label>Label of Option 1</label>
<input type="radio" value="value-of-option-2" name="radioOption">
<label>Label of Option 2</label>
```

Enhancement: 
 Use unique ID for each `input` `type=radio`, and then use `for` attribute on `label` with that ID
```html
<input type="file" ... id="unique-id">
<label for="unique-id">Label of Option 1</label>
```

Prateek93a:
Resolves #4854

 Changes:

I am using the `id` attribute of the option element to set the `for` attribute of label element.

I wanted to make minimal changes to the code for fixing the issue. Hence I went with approach that @dipamsen suggested. However if required I will update the code to implement the approach @limzykenneth suggested.

GIF of the change:

ezgif com-gif-maker (2)

#### PR Checklist

- [X] `npm run lint` passes


limzykenneth:
@Prateek93a Can the nested label input approach be used here instead? I want to keep the `id` to be something that the user explicitly define only, otherwise it may cause problem when the user attempts to set an `id` on the input element.

Prateek93a:
@limzykenneth Yes, I will update the PR.

Prateek93a:
@limzykenneth I have updated the code to use the nested label input approach. For this, I have also made small updates to the unit tests to work according to the changes.

lmccart:
@limzykenneth if this looks good to you, please feel free to merge

limzykenneth:
This looks good to me and can be merged though I only just noticed that `createCheckbox` is using random ID to associate labels with the checkbox as was done here before. @Prateek93a Are you interested in changing the behaviour of `createCheckbox` to match your implementation here of nested label tag? If so you can push directly to this PR and I'll merge when that's done, otherwise I can merge this as is.

Prateek93a:
@limzykenneth I have made the changes. I have also added some unit tests for `createCheckbox` function.

limzykenneth:
Sorry this took way too long, it slipped through the cracks. I'll review this today.

limzykenneth:
Seems to all work fine. We can revisit if specific problems occur. Thanks @Prateek93a!

Prateek93a:
No problem!

PR #5096, opened 2021-03-16
add version const for 1.3.0

micuat:
Resolves #2525

 Changes:

add version constant - not sure if other files (contributor docs?) needs edit, not to forget to update the const every release (and I am not sure I should have made a PR with 1.3.1...)

#### PR Checklist

- [ ] `npm run lint` passes
- [ ] [Inline documentation] is included / updated
- [ ] [Unit tests] are included / updated

[Inline documentation]: https://github.com/processing/p5.js/blob/main/contributor_docs/inline_documentation.md
[Unit tests]: https://github.com/processing/p5.js/tree/main/contributor_docs#unit-tests

PR #5107, opened 2021-03-18
get version constant from package.json

### LINKED ISSUE #2525: Feature request: check p5.js "version" from code.
#### Nature of issue?

- [x] New feature request

- [x] Core

#### Details about the bug: 

- p5.js version: 0.5.16
- Web browser and version: Chrome 63.0.3239.108 
- Operating System: Mac OS 10.12.6

#### New feature details:

It could be useful to have the p5.js version no. accessible from a properties query in code, ie. be able to read the "0.5.16" as above. In my case, this is just to ensure I'm running on a version no earlier than the one I developed on, a reasonable precaution. However, I'm aware that coding to explicit releases can lead to a minefield of special casing, which we already have mountains of in Web design, eg. the plethora of ugly hacks to deal with old Internet Explorers. Maybe p5.js designers are avoiding this.

(PS. Can I get it cleanly from project.version ? I'm not a JS expert.)

Thanks, GE.

micuat:
Resolves #2525 continuing from https://github.com/processing/p5.js/pull/5096

 Changes:

add version constant taking from `package.json` reflecting @lmccart 's comment https://github.com/processing/p5.js/pull/5096#pullrequestreview-613543846
However the downside is that the whole `package.json` seems to be exposed to the minifined file. This is not a problem as the project is open-source, but I wonder if it causes other issues.

snippets from compiled bundle:

```javascript
      262: [
        function(_dereq_, module, exports) {
          module.exports = {
            name: 'p5',
            repository: 'processing/p5.js',
            scripts: {
              grunt: 'grunt',
              build: 'grunt build',
              dev: 'grunt browserify:dev connect:yui watch:quick',
              docs: 'grunt yui',
              'docs:dev': 'grunt yui:dev',
              test: 'grunt',
              lint: 'grunt lint-no-fix',
              'lint:source': 'grunt lint-no-fix:source',
              'lint:samples': 'grunt lint-no-fix:samples',
              'lint:fix': 'grunt lint-fix',
              'contributors:add': 'all-contributors add',
              'contributors:generate': 'all-contributors generate',
              generate: 'all-contributors generate',
              release: 'grunt release-p5',
              prepublishOnly: 'grunt prerelease'
            },
            'lint-staged': {
              ignore: ['test/js/**/*.js'],
              'Gruntfile.js': 'eslint',
              'docs/preprocessor.js': 'eslint',
              'utils/**/*.js': 'eslint',
              'tasks/**/*.js': 'eslint',
              'test/**/*.js': 'eslint',
              'src/**/*.js': [
                'eslint',
                'node --require @babel/register ./utils/sample-linter.js'
              ]
            },
            version: '1.2.0',
            devDependencies: {
              '@babel/core': '^7.7.7',
              '@babel/preset-env': '^7.10.2',
              '@babel/register': '^7.7.7',
```

```javascript
          /**
           * Version of this p5.js.
           * @property {String} VERSION
           * @final
           */
          var VERSION = _dereq_('../../package.json').version;
```

#### PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -11,7 +11,7 @@ const _PI = Math.PI;
  * @property {String} VERSION
  * @final
  */
-export const VERSION = VERSIONCONST;
+export const VERSION = 'VERSIONCONST';
 
 // GRAPHICS RENDERER
 /**
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -11,7 +11,7 @@ const _PI = Math.PI;
  * @property {String} VERSION
  * @final
  */
-export const VERSION = require('../../package.json').version;
+export const VERSION = VERSIONCONST;
 
 // GRAPHICS RENDERER
 /**
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -11,7 +11,8 @@ const _PI = Math.PI;
  * @property {String} VERSION
  * @final
  */
-export const VERSION = 'VERSIONCONST';
+export const VERSION =
+  'VERSION_CONST_WILL_BE_REPLACED_BY_BROWSERIFY_BUILD_PROCESS';
 
 // GRAPHICS RENDERER
 /**
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -6,6 +6,13 @@
 
 const _PI = Math.PI;
 
+/**
+ * Version of this p5.js.
+ * @property {String} VERSION
+ * @final
+ */
+export const VERSION = require('../../package.json').version;
+
 // GRAPHICS RENDERER
 /**
  * The default, two-dimensional renderer.
--- a/tasks/build/browserify.js
+++ b/tasks/build/browserify.js
@@ -78,6 +78,11 @@ module.exports = function(grunt) {
           code += data;
         })
         .on('end', function() {
+          code = code.replace(
+            'VERSIONCONST',
+            grunt.template.process(`'<%= pkg.version %>'`)
+          );
+
           // "code" is complete: create the distributable UMD build by running
           // the bundle through derequire
           // (Derequire changes the bundle's internal "require" function to
--- a/tasks/build/browserify.js
+++ b/tasks/build/browserify.js
@@ -79,7 +79,7 @@ module.exports = function(grunt) {
         })
         .on('end', function() {
           code = code.replace(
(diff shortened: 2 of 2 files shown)


akshay-99:
Including the package.json file into the built client-side library feels a bit wrong to me, even though as an opensource project we won't have any security implications from doing so. A few other options to consider would be:

1. Using `browserify` to set the version number as a global variable like [here](https://github.com/processing/p5.js/blob/4c194ec0efccf48d32e93b42a1f0b033d3a5c903/tasks/build/browserify.js#L39). Browserify already has access to the version number (It currently uses it to add the top banner). It can then be set as a p5 constant as well.
2.  Using some grunt task like `grunt-string-replace` as seen in [this answer](https://stackoverflow.com/a/20446748) to replace a pattern in the constants file with the version number

limzykenneth:
Is it possible to import the version number using ES6 import statement and let the build step tree shaking the unneeded info from package.json? Not sure if something like that will work at all though. If not then something like option 1 proposed by @akshay-99 would probably be fine as well.

micuat:
thanks for the comments!
@limzykenneth My understanding is that it's only possible by separating the files based on dependency - for example `version.json` referred to by `package.json` so that the script requires `version.json` - which doesn't work in our case. So I chose option 1 ->
@akshay-99 thanks for the spot-on suggestion! I updated the branch to replace the code in grunt - I'm not sure if there's a "cleaner" way to achieve this.

micuat:
oh. test failed because of the undefined const. maybe replacing this with a valid string e.g. `'VERSIONCONST'` would be the solution?

limzykenneth:
I think you can ask the linter to ignore that line as well.

outofambit:
this is awesome @micuat! one tiny request, could we use something like `'VERSION_CONST_WILL_BE_REPLACED_BY_BROWSERIFY_BUILD_PROCESS'` instead of `'VERSIONCONST'` to help communicate how this string is changed in the build process? we don't need anything else before we merge this right?

micuat:
@outofambit thanks <3 yes I totally agree, I also thought that the text should be unique + recognizable if something broke during the grunt process

outofambit (approved):

lmccart:
awesome, thanks everyone!!

PR #5127, opened 2021-03-25
checking for pixel array reassignment inside updatePixels

### LINKED ISSUE #4993: "pixels" array of p5.Image cannot be reassigned while canvas one can

ahujadivyam:
Resolves #4993

 Changes:

checking for pixel array reassignment inside updatePixels and creating new ImageData object accordingly, keeping optimizations related to copying in mind (like if map() is used the resulting UInt8ClampedArray can directly be used by ImageData Constructor and no copying of data will be needed ).

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -281,7 +281,7 @@ p5.Renderer2D.prototype._getPixel = function(x, y) {
 
 p5.Renderer2D.prototype.loadPixels = function() {
   const pixelsState = this._pixelsState; // if called by p5.Image
-
+  
   const pd = pixelsState._pixelDensity;
   const w = this.width * pd;
   const h = this.height * pd;
@@ -391,18 +391,24 @@ p5.Renderer2D.prototype.updatePixels = function(x, y, w, h) {
     this.gifProperties.frames[this.gifProperties.displayIndex].image =
       pixelsState.imageData;
   }
-
-  if (pixelsState.pixels !== pixelsState.imageData.data) {
+  
+  let pixels = pixelsState.pixels
+  if (this._pInst instanceof p5 &&  
+      this._pInst._isGlobal === true) {
+    pixels = globalThis.pixels;
+  }
+  
+  if (pixels !== pixelsState.imageData.data) {
     let imgData;
-    if (pixelsState.pixels instanceof Uint8ClampedArray) {
-      if (pixelsState.pixels.length === pixelsState.imageData.data.length) {
+    if (pixels instanceof Uint8ClampedArray) {
+      if (pixels.length === pixelsState.imageData.data.length) {
         imgData = new ImageData(
-          pixelsState.pixels,
+          pixels,
           pixelsState.width,
           pixelsState.height
         );
       } else {
-        let imgArr = Array.from(pixelsState.pixels);
+        let imgArr = Array.from(pixels);
         imgArr.length = pixelsState.imageData.data.length;
         imgData = new ImageData(
           Uint8ClampedArray.from(imgArr),
@@ -411,10 +417,10 @@ p5.Renderer2D.prototype.updatePixels = function(x, y, w, h) {
         );
       }
     } else {
-      if (Array.isArray(pixelsState.pixels)) {
-        pixelsState.pixels.length = pixelsState.imageData.data.length;
+      if (Array.isArray(pixels)) {
+        pixels.length = pixelsState.imageData.data.length;
         imgData = new ImageData(
-          Uint8ClampedArray.from(pixelsState.pixels),
+          Uint8ClampedArray.from(pixels),
           pixelsState.imageData.width,
           pixelsState.imageData.height
         );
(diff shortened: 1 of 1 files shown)


limzykenneth:
@DivyamAhuja I might be doing something wrong here but with the following code, I still wasn't able to assign to `pixels`.

```javascript
function setup() {
	createCanvas(1280, 720);
	loadPixels();
}

function draw() {
	// for(let i=0; i<pixels.length; i++){
	// 	pixels[i] = 255;
	// }

	pixels = pixels.map((pixel) => {
		return 255;
	});

	updatePixels();
}
```

The `for` loop still works.

DivyamAhuja:
@limzykenneth I guess, I totally ignored the fact that global mode `pixels` was just another reference, so intenal reference was not getting reassigned and code was not working. But I just applied a hacky patch, so let me know what are your views about it. It may not work if someone tries to alter internal pixels array, but for general use case where one is altering just pixel array reference , it will work fine

limzykenneth:
> It may not work if someone tries to alter internal pixels array

What do you mean by that?

DivyamAhuja:
@limzykenneth I meant in this way. As the hacky patch I applied checks if we are in global mode and acting upon instance of p5.Renderer2D, it checks for value of pixels array attached to globalThis( that is window )

https://editor.p5js.org/DivyamAhuja/sketches/Sd50S3ym-

limzykenneth:
I'm currently tied up with something so couldn't review this yet, I'll trr to get to it next week. Do have a look at how the codebase handles variables attached to the global scope (or instance) and see if we can emulate it here.

src/core/p5.Renderer2D.js
```diff
@@ -392,6 +392,40 @@ p5.Renderer2D.prototype.updatePixels = function(x, y, w, h) {
...
+    pixels = globalThis.pixels;
+  }
+
+  if (pixels !== pixelsState.imageData.data) {
+    let imgData;
+    if (pixels instanceof Uint8ClampedArray) {
+      if (pixels.length === pixelsState.imageData.data.length) {
+        imgData = new ImageData(pixels, pixelsState.width, pixelsState.height);
+      } else {
+        let imgArr = Array.from(pixels);
```
limzykenneth, line 406: I wonder if it would be more efficient to use the TypedArray's `set()` method to get the array to the right size, like the one shown [here](https://stackoverflow.com/a/40324738). So instead of creating an array and a Uint8ClampedArray, we just create one Uint8ClampedArray of the right size and copy the existing data over.

limzykenneth (commented):
I've tested out the code and it works from what I can test. Performance-wise I didn't see any significant difference between reassignment and no reassignment of `pixels`, perhaps a bit more garbage collection activity but not significant enough to suggest any noticeable slowdown in practice.

Just one comment inline, not super important, just a bit of micro-optimizing :stuck_out_tongue_closed_eyes: 

Would you be able to write some unit tests that goes with the changes here?

src/core/p5.Renderer2D.js
```diff
@@ -401,21 +401,26 @@ p5.Renderer2D.prototype.updatePixels = function(x, y, w, h) {
...
+          imgArr,
+          pixelsState.imageData.width,
+          pixelsState.imageData.height
         );
       }
     } else {
       if (Array.isArray(pixels)) {
-        pixels.length = pixelsState.imageData.data.length;
+        let imgArr = new Uint8ClampedArray(pixelsState.imageData.data.length);
+        imgArr.set(pixels);
```
limzykenneth, line 421: I think the original implementation for this is simpler but if you just want to keep things consistent then I guess this should be fine as well.

DivyamAhuja: Yeah I thought it'll be fine this way too, if we are doing the same above. :D
As for unit tests, I checked and I think no unit tests exist for `updatePixels()`, so I guess I will write some.

Qianqianye:
Thanks @DivyamAhuja for working on it and @limzykenneth  for reviewing. Just to follow up on this PR, is it ready to be merged after we resolve the conflicts?

PR #5165, opened 2021-04-09
add VERSION const as p5.VERSION

micuat:
continues from https://github.com/processing/p5.js/pull/5107 and resolves https://github.com/processing/p5.js/issues/2525

- [x] `npm run lint` passes

I kept both `p5.VERSION` and `p5.prototype.VERSION` but not sure if we keep both or we just use the former so that it will never be exposed to global scope (no `sketch=>sketch.VERSION` in case of instance). Whichever it is, the new implementation (`p5.VERSION`) should have the priority as it has an advantage that it can be referenced without instantiation.

### DIFF
--- a/src/core/main.js
+++ b/src/core/main.js
@@ -714,6 +714,8 @@ for (const k in constants) {
   p5.prototype[k] = constants[k];
 }
 
+// makes the `VERSION` constant available on the p5 object
+// in instance mode, even if it hasn't been instatiated yet
 p5.VERSION = constants.VERSION;
 
 // functions that cause preload to wait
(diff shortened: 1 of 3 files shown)


outofambit:
hey @micuat, just checking in on this, I'd love to get this merged soon. are you still interested in working on this?

it looks like the only thing left is a decision on the prototype vs instance question above? I think it's probably good to have both. If we only use `p5.VERSION`, I think that would mean it wouldn't be available in the web editor (unless you `this`), which i think is worth preserving. open to other opinions, too!

Oh we need docs too! I can help with that if you'd like.

micuat:
sure! I agree that it should be accessible both ways. Do you think it still needs an edit or is ready for review? I don't think I have access to a build environment / have time to test it right now, so I appreciate if anyone can test it (and edit it if needed).

outofambit:
thanks @micuat! i tested locally and wrote some unit tests to be extra safe. merging now!

PR #5172, opened 2021-04-11
Fixed the documentation for Vector.angleBetween to indicate that it respects angleMode.

sflanker:
Changes:

Modifies the documentation comments for p5.Vector.prototype.angleBetween to indicate that it respects angleMode and will return the angle in either radians or degrees depending on the current value.

#### PR Checklist

Documentation only change.

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -1384,9 +1384,9 @@ p5.Vector.prototype.setMag = function setMag(n) {
 
 /**
  * Calculate the angle of rotation for this vector(only 2D vectors).
- * p5.Vectors created using <a src="#/p5/createVector">createVector()</a>
- * will take the current <a = src="#/p5/angleMode">angleMode</a> into consideration, and give the angle
- * in radians or degree accordingly.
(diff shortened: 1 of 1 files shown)


src/math/p5.Vector.js
```diff
@@ -1549,7 +1549,10 @@ p5.Vector.prototype.rotate = function rotate(a) {
 };
 
 /**
- * Calculates and returns the angle (in radians) between two vectors.
+ * Calculates and returns the angle between two vectors. This function will take
+ * the current <a = src="#/p5/angleMode">angleMode</a> into consideration, and
```
outofambit, line 1553: thanks @sflanker! this looks good, but we need to use `href` for the link to work
```suggestion
 * the current <a href="#/p5/angleMode">angleMode</a> into consideration, and
```

sflanker: Oops, I copied this from the comments on the `heading` function without paying enough attention.

lmccart:
thanks @sflanker!

src/math/p5.Vector.js
outofambit: no problem! thank you!

PR #5225, opened 2021-05-05
fixed transparency of background problem

### LINKED ISSUE #5175: Background image with alpha not working as expected
Using the following simply code, the alpha parameter simply has no effect

```
function setup() {
    ...
    bg = loadImage('bg.jpg');
    ...
}

function draw() {
    ...
    background(bg, 10);
    ...
}
```

If you replace `bg` in `background()` with a colour value, you see trails on screen, which is correct, but with the image, there are no trails, effectively, the alpha value doesn't work. You can test this by changing the alpha value to anything and there is no difference. I tried ranges 0-1, and 0-255.

NB - this issue, https://github.com/processing/p5.js/issues/1906, was closed, with @limzykenneth stating `background()` could not accept an image, which is false, proven by the fact you can indeed load an image using `background()` and the documentation even states you can use an image and an alpha value.

two-ticks:
## Changes:
[example sketch](https://editor.p5js.org/radium.scientist/sketches/_EASCeZ-u)
- change alpha value according to args[1] 
```javascript
p5.Renderer2D.prototype.background = function(...args) {
  this.drawingContext.save();
  this.resetMatrix();
  if (args[0] instanceof p5.Image) {
    if (args[1] >= 0) {
      // set transparency of background
      const img = args[0];
      const pixels = filters._toPixels(img.canvas);
      const tmpCanvas = document.createElement('canvas');
      tmpCanvas.width = img.canvas.width;
      tmpCanvas.height = img.canvas.height;
      const tmpCtx = tmpCanvas.getContext('2d');
      const id = tmpCtx.createImageData(img.canvas.width, img.canvas.height);
      const newPixels = id.data;
      for (let i = 0; i < pixels.length; i += 4) {
        const r = pixels[i];
        const g = pixels[i + 1];
        const b = pixels[i + 2];
        newPixels[i] = r;
        newPixels[i + 1] = g;
        newPixels[i + 2] = b;
        newPixels[i + 3] = args[1];
      }
      tmpCtx.putImageData(id, 0, 0);
      img.canvas = tmpCanvas;
      this._pInst.image(img, 0, 0, this.width, this.height);
    } else {
      this._pInst.image(args[0], 0, 0, this.width, this.height);
    }
```

 ## Screenshots of the change:

image
image

## PR Checklist

- [x] `npm run lint` passes


two-ticks:
@montoyamoraga please review this pull request

montoyamoraga:
hi @two-ticks sorry i'm not familiar with this section of the code,
somebody else will review it :)

limzykenneth:
I think the implementation here is probably going to be way too slow especially with larger canvases since it requires going through the pixels array of the whole canvas. Can this possibly be implemented in another way?

two-ticks:
Maybe we can implement it with `globalAlpha`  also. I also tried to implement this method but implementations failed. `tint()` also uses such method(as in pull request) so it can be an immediate fix.

limzykenneth:
I'm not quite sure what you mean by `globalAlpha`, can you elaborate on this a bit more?

two-ticks:
I have tried something similar to this and some more variations of it. Idea was to change `globalAlpha` value rather than changing alpha value of each pixel. 

```js
  if (args[0] instanceof p5.Image) {
    if (args[1] >= 0) {
      // set transparency of background
      const ctx = this.getContext("2d");
      ctx.globalAlpha = args[1]; // Turn transparency on
      this._pInst.image(img, 0, 0, this.width, this.height);
    } else {
      this._pInst.image(args[0], 0, 0, this.width, this.height);
    }
```

https://github.com/processing/p5.js/blob/374acfb44588bfd565c54d61264df197d798d121/src/core/p5.Renderer2D.js#L197 is using loops

limzykenneth:
Ok, `globalAlpha` seem like a possible fix. As long as there won't be unexpected bug elsewhere as a result of using it, we can have a try at it.

Qianqianye:
Thanks @two-ticks for working on it. @limzykenneth, any recommended next steps to merge this PR?
Are there anything @Palaksharma23 can work on, as @two-ticks mentioned in issue #5175 ?

limzykenneth:
For this we will need an alternate implementation than the one in this PR as it is likely to be too slow. If @Palaksharma23 wants to have a go and @two-ticks isn't currently working on this I'm happy to review any new changes. The solution doesn't have to use `globalAlpha` as mentioned if there are better alternatives.

two-ticks:
I would be very happy if @Palaksharma23 wants to work on this and get assigned. Waiting for @Palaksharma23 to reply.

yashlamba:
Hey @two-ticks @limzykenneth, I am new to p5.js and got issue #5175 in my mail from CodeTriage (thought of investigating it further to learn). I looked at #5225 and the other `globalAlpha` based implementation with some modification (they both work as expected), I couldn't find any other optimal way to do it as well. Happy to test/look further if you can suggest any leads:)

two-ticks:
I think `globalAlpha` implementation would be the fastest, I also couldn't find any other way. If possible share your approach and code so maintainers can review (and assign it to you). @limzykenneth can #5225 issue be assigned to @yashlamba ?

limzykenneth:
I can't assign @yashlamba on #5175 as only participants of the thread can be assigned but @yashlamba feel free to work on this. Thanks!

yashlamba:
Thanks @limzykenneth, I'll go ahead and open a new PR.

Qianqianye:
Thanks @two-ticks and @yashlamba. I will close this PR since it has been fixed on the recent merge of #5683!

PR #5251, opened 2021-05-16
Fixed error in references

aLyonsGH:
Addresses #2865 
In the description of lerpColor(), I changed "the two values where 0.0 equal to the first color" to "the two values where 0.0 is equal to the first color" (I added the "is")


limzykenneth:
Hi @aLyonsGH, I think you have accidentally saved the edited file as a new file at the root of the project folder instead of saving over the original file. Can you try can revert the current changes and save the changes in the original file? Thanks.

aLyonsGH:
Done! I think this is all set now? Thank you for your patience, this is my first contribution to p5.js!

limzykenneth:
I'm not sure why the CI tests are not running, they keep getting cancelled after starting for a few minutes. @outofambit Do you have an idea what's going on?

outofambit (approved):
looks good to me! i re-ran the CI jobs manually, will merge when they finish (i'm sure it'll be fine)

PR #5259, opened 2021-05-23
added missing requried Distortion.process param

BlairCurrey:
Changes:
Adds required parameter to `process` method of `Distortion` object. A `SoundFile` is required but there is no indication in the documents as seen here:
image

### DIFF
--- a/lib/addons/p5.sound.js
+++ b/lib/addons/p5.sound.js
@@ -11762,6 +11762,7 @@ var __WEBPACK_AMD_DEFINE_RESULT__;
    *
    * @method process
    * @for p5.Distortion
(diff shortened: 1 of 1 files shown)


outofambit:
hey @BlairCurrey thanks for submitting this! the p5 sound functionality here is actually imported directly from https://github.com/processing/p5.js-sound/blob/main/src/distortion.js periodically. could you propose your change in that repo please? otherwise it'll get wiped out next time we import from https://github.com/processing/p5.js-sound/. thank you!

i'm gonna close this but let me know if you have any questions or need any help with that. this is definitely worth fixing!

BlairCurrey:
No problem @outofambit , thanks for the clarification.

PR #5272, opened 2021-05-29
Added normal function. Resolves #5022

### LINKED ISSUE #5022: Ability to specify vertex normals when drawing 3d shapes with beginShape and vertex
Currently 3d shapes drawn using beginShape/vertex/endShape have all vertexes assigned a normal of `(0, 0, 1)`. When drawing custom geometry with these functions it should be possible to specify the normal vector for each vertex.

### How would this new feature help [increase access]

Individuals learning the fundamentals of 3d computer graphics for the first time would benefit from being able to use vertex normals when drawing custom geometry using beginShape/vertex/endShape. I think this is more accessible to users who are new to 3d graphics then jumping straight to loading 3d model files, or creating custom p5.Geometry objects. It also presents an opportunity to introduce vector math concepts such as cross product, addition, and normalization.

- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [x] WebGL

#### New feature details:

I propose adding a function `vertexNormal` that sets the normal vector to be used in subsequent calls to `vertex`. This vector would default to `(0, 0, 1)` thereby preserving the current behavior by default. This value would be stored/restored as part of push/pop operations. As far as I know this would only be relevant for drawing using the WebGL renderer, and would have no effect when using other renderers.

sflanker:
Resolves #5022

 Changes:

Added a function `normal` to p5 and `p5.RendererGL` to allow a vertex normal to be explicitly specified for vertices crated with the vertex function..

#### PR Checklist

- [X] `npm run lint` passes

### DIFF
--- a/src/core/shape/vertex.js
+++ b/src/core/shape/vertex.js
@@ -992,4 +992,56 @@ p5.prototype.vertex = function(x, y, moveTo, u, v) {
   return this;
 };
 
+/**
+ * Sets the 3d vertex normal to use for subsequent vertices drawn with
+ * <a href="#/p5/vertex">vertex()</a>. A normal is a vector that is generally
+ * nearly perpendicular to a shape's surface which controls how much light will
+ * be reflected from that part of the surface.
+ *
+ * @method normal
+ * @param  {Vector} vector A p5.Vector representing the vertex normal.
+ * @chainable
+ * @example
+ * <div>
+ * <code>
+ * function setup() {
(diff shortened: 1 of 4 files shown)


stalgiag (approved):
Looks good to me!

Issue #5264, opened 2021-05-25
Stroke odd behaviour for small shapes but up-close camera


JetStarBlues:
#### Most appropriate sub-area of p5.js?

- [x] WebGL

#### Details about the bug:

- p5.js version: 1.3.1 (though I think this bug is not specific to this version)
- Web browser and version: Chrome, Firefox
- Operating System: Windows 10, Ubuntu 20.04
- Steps to reproduce this:
  - Visible on the:
    - [camera() reference page][0] (first example with a plane (white fill, black stroke))
      - ![stroke1](https://user-images.githubusercontent.com/4354703/119559045-64771280-bd5f-11eb-9b0e-fa5201115825.png)
    - [frustum() reference page][1] (example with two cubes(no fill, blue stroke))
      - ![stroke2](https://user-images.githubusercontent.com/4354703/119559079-70fb6b00-bd5f-11eb-81f9-371b338fde7c.png)

It seems by default, the stroke shader draws "small" objects with a thick background. I think this is to help with visibility. For example:

```js
function setup() {
  createCanvas(100, 100, WEBGL);
}

function draw() {
  background(204);
  plane(10, 10);
}
```
generates,  
![stroke_100x100_plane10_defaultCameraPostion](https://user-images.githubusercontent.com/4354703/119559654-33e3a880-bd60-11eb-8948-92341c77e4c8.png)

When the object is larger, the stroke gets thinner.

```js
plane(40, 40);
```
generates,  
![stroke_100x100_plane40_defaultCameraPostion](https://user-images.githubusercontent.com/4354703/119559689-3e05a700-bd60-11eb-80ae-1e0420e4683f.png)

```js
plane(80, 80);
```
generates,  
![stroke_100x100_plane80_defaultCameraPostion](https://user-images.githubusercontent.com/4354703/119559718-46f67880-bd60-11eb-8afe-1f24652f8052.png)

However, this thinning of the stroke does not happen when instead of the object getting bigger, the camera moves closer to the object.

The default camera position is calculated as `(0, 0, (height / 2) / tan(PI / 6))`. For a canvas with height `100`, this is `~(0, 0, 87)`.
```js
let cam;

function setup() {
  createCanvas(100, 100, WEBGL);
  cam = createCamera();
}

function draw() {
  background(204);
  cam.camera(
    0, 0, 87,
    0, 0, 0,
    0, 1, 0
  );
  plane(10, 10);
}
```
generates,  
[...]

PR #5275, opened 2021-05-30
Fix frustum examples

JetStarBlues:
Mentioned in #5264

**Changes:**

Currently the [frustum][2] examples look like this:  
stroke2

The stroke's relative size makes it difficult to tell what you are looking at.  
This PR changes the code in the examples to use `normalMaterial()` instead.  
frustum_normalMaterial2

This is more inline with other examples in the series such as [perspective()][0] and [ortho()][1].

[0]: https://p5js.org/reference/#/p5/perspective
[1]: https://p5js.org/reference/#/p5/ortho
[2]: https://p5js.org/reference/#/p5/frustum

### DIFF
--- a/src/webgl/p5.Camera.js
+++ b/src/webgl/p5.Camera.js
@@ -262,9 +262,7 @@ p5.prototype.ortho = function(...args) {
  * function draw() {
  *   background(200);
  *   orbitControl();
- *   strokeWeight(10);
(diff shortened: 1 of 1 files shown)


aferriss:
Good catch! Looks much better with the normal material.

aferriss (approved):

Issue #5148, opened 2021-04-01
Impossible to set stroke shader without also setting fill shader


stalgiag:
#### Most appropriate sub-area of p5.js?

- [ ] Accessibility (Web Accessibility)
- [ ] Build tools and processes
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

#### Details about the bug:

While investigating #5129, I realized that custom stroke shaders always become the custom fill shader. This is because [`isStrokeShader()`](https://github.com/processing/p5.js/blob/main/src/webgl/p5.Shader.js#L511) returns false on the very first frame because the uniforms have not been loaded into the Shader object yet. This causes [`RendererGL.userFillShader`](https://github.com/processing/p5.js/blob/main/src/webgl/material.js#L260) to be set to the stroke shader. Since it isn't reset every frame, the fill shader will be the custom stroke shader unless manually changed.

There are a few issues at play here. The primary one is that `p5.Shader` objects don't load uniforms during preload. They actually load them during binding. This is fine except we are checking uniforms to determine the type of shader it is.

This could potentially be fixed by initializing the shader object before decrementing the preload but that might cause more problems. Another idea is that we might want to introduce the use of `FILL`, `POINTS`, and `LINES` for the custom shaders in the same way that Processing has. We also might want to consider a different way of auto-detecting the shader type if we want that second argument to be optional.

Thoughts anyone? Maybe @kjhollen or @aferriss?

- p5.js version: latest
- Web browser and version: all
- Operating System: all
- Steps to reproduce this:
[https://editor.p5js.org/stalgia.grigg/sketches/DVQKZTPFS](https://editor.p5js.org/stalgia.grigg/sketches/DVQKZTPFS)

aferriss:
> This could potentially be fixed by initializing the shader object before decrementing the preload but that might cause more problems.

Also, `createShader()` is often used inside of setup, so I think this would cause a problem there as well. Is there a better way to check what kind of shader it is rather than using the uniforms?

I don't know enough about what desktop Processing is doing with the FILL, POINTS, LINES optional param. Is it kind of like setting the gl primitive mode?

stalgiag:
[...]

PR #5286, opened 2021-06-01
Fixes initialization of user-defined stroke shader

### LINKED ISSUE #5285: User-defined stroke shader is not initialized correctly
- [x] WebGL

#### Details about the bug:

Whether a user defined shader (created via [`shader()`][0]) is treated as a fill-shader or a stroke-shader is currently determined by a check for the existence of the `uStrokeWeight` uniform.

[_material.js_][1]  
```js
p5.prototype.shader = function(s) {
  ...

  if (s.isStrokeShader()) {
    this._renderer.userStrokeShader = s;
  } else {
    this._renderer.userFillShader = s;
    this._renderer._useNormalMaterial = false;
  }

  s.init();

  return this;
};
```

[_p5.Shader.js_][2]  
```js
p5.Shader.prototype.isStrokeShader = function() {
  return this.uniforms.uStrokeWeight !== undefined;
};
```

This uniform is only available **after** the call to `s.init()`.

As a result, the current code incorrectly marks a user-defined stroke-shader as a fill-shader on the first invocation of `shader(s)`.
- The `isStrokeShader()` check returns `false`, despite the uniform's presence in the shader source code, because it has not yet been added to `s.uniforms` via `s.init()`.

On the second invocation of `shader(s)`, the shader is correctly recognized as a stroke-shader because the uniform is now present.

However, it is now marked as both a fill-shader and a stroke-shader, and [used for both purposes][3] accordingly.

Here's a [live demo][4] showing this bug.

[0]: https://p5js.org/reference/#/p5/shader
[1]: https://github.com/processing/p5.js/blob/e7f2c143ae283c7983a890e4a2b914203f89c562/src/webgl/material.js#L250
[2]: https://github.com/processing/p5.js/blob/e7f2c143ae283c7983a890e4a2b914203f89c562/src/webgl/p5.Shader.js#L511
[3]: https://github.com/processing/p5.js/blob/f36d825f435e15adbf3969c615ea6e2b04bfe635/src/webgl/p5.RendererGL.Retained.js#L103
[4]: https://editor.p5js.org/acedean/sketches/1t_qtYV1L

JetStarBlues:
Resolves #5285

**Changes:**  
Initializes uniforms before using them to perform `isStrokeShader()` check. See linked issue for more details.

### DIFF
--- a/src/webgl/material.js
+++ b/src/webgl/material.js
@@ -255,14 +255,14 @@ p5.prototype.shader = function(s) {
     s._renderer = this._renderer;
   }
 
+  s.init();
+
   if (s.isStrokeShader()) {
     this._renderer.userStrokeShader = s;
   } else {
     this._renderer.userFillShader = s;
     this._renderer._useNormalMaterial = false;
   }
-
-  s.init();
 
   return this;
 };


stalgiag:
This also resolves #5148. There is still the question of whether the check for `uStrokeWeight` is a valid way to determine if a shader is a stroke shader but that is a separate issue.

stalgiag:
Thanks!

stalgiag:
@all-contributors please add @JetStarBlues for code

JetStarBlues:
> This also resolves #5148

Oh, neat!

PR #5288, opened 2021-06-01
Fix the texture() example

JetStarBlues:
Currently, the size of the box used in the first [`texture()`][0] example is `200`. The canvas however is `100`. The net effect is that we see the interior of the box, which is likely not the intended behaviour.

texture_cur

If the box is scaled down, then we can see the entirety of it on the canvas (from the outisde). Which gives a better example of what a texture looks like when applied to a box.

texture_change

[0]: https://p5js.org/reference/#/p5/texture


stalgiag:
I believe that being inside the box was intentional in the beginning but I agree that this looks better and is a more clear demonstration of what to expect when texturing a box.

PR #5311, opened 2021-06-14
Add example for specularMaterial()

JetStarBlues:
Adds an example for [`specularMaterial()`][0]

Live demo [here][1].

**Screenshot of change**
specularMaterialExample_proposal

\---

Note in the current source code, I found that an [example was already present][2]. It fails to render in the generated documentation because of where it is declared. (On the second parameter variant instead of the first). [Here][3] is a live demo of this hidden example, and below a screenshot of what it renders.
specularMaterialExample_hidden

[0]: https://p5js.org/reference/#/p5/specularMaterial
[1]: https://editor.p5js.org/acedean/sketches/aVTH4VdHd
[2]: https://github.com/processing/p5.js/blob/7375939aa4e4e5ada7381815ace2dcb8a0daa74c/src/webgl/material.js#L728
[3]: https://editor.p5js.org/acedean/sketches/0N81pic7o

### DIFF
--- a/src/webgl/material.js
+++ b/src/webgl/material.js
@@ -712,6 +712,33 @@ p5.prototype.emissiveMaterial = function(v1, v2, v3, a) {
  * @param  {Number} [alpha] alpha value relative to current color range
  *                                 (default is 0-255)
  * @chainable
+ *
+ * @example
+ * <div>
+ * <code>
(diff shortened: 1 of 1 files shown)


outofambit (approved):
thank you @JetStarBlues!

PR #5312, opened 2021-06-14
Improve documentation clarity of light methods

JetStarBlues:
Minor changes to the wording used in the following lighting methods, with the goal of improving clarity.

Borrowed some descriptions from the [Processing Reference][8].

### [ambientLight()][0]

_1) Reword_

Current:

> Creates an ambient light with a color. Ambient light is light that comes from everywhere on the canvas. It has no particular source.

Change:

> Creates an ambient light with the given color.

> Ambient light does not come from a specific direction. Objects are evenly lit from all sides. Ambient lights are almost always used in combination with other types of lights.

_2) Add disclaimer_

Change:

> Note: lights need to be called (whether directly or indirectly) within draw() to remain persistent in a looping program. Placing them in setup() will cause them to only have an effect the first time through the loop.

_3) Be consistent with params_

Use same parameter descriptions as <a href="https://p5js.org/reference/#/p5/color">color()</a> reference.

Current:

> alpha
> the alpha value (Optional)

> gray
> a gray value

> color
> p5.Color: the ambient light color

Change:

> alpha
> alpha value relative to current color range (default is 0-255)

> gray
> Number: number specifying value between white and black

> color
> p5.Color: color as a <a href="#/p5.Color">p5.Color</a>

_4) Fix comments in examples_

### [directionalLight][1]

_1) Reword_

Current:

> Creates a directional light with a color and a direction

> A maximum of 5 directionalLight can be active at one time

Change:

> Creates a directional light with the given color and direction.

> Directional light comes from one direction.
> The direction is specified as numbers inclusively between -1 and 1. For example, setting the direction as (0, -1, 0) will cause the geometry to be lit from below (since the light will be facing directly upwards). Similarly, setting the direction as (1, 0, 0) will cause the geometry to be lit from the left (since the light will be facing directly rightwards).

> A maximum of **5** directional lights can be active at once.

_2) Add disclaimer_

Change:

> Note: lights need to be called (whether directly or indirectly) within draw() to remain persistent in a looping program. Placing them in setup() will cause them to only have an effect the first time through the loop.

_3) Rename `position` parameter to `direction`_

_4) Be consistent with params_

Current:

> v1
> red or hue value (depending on the current color mode),

> v2
> green or saturation value

> v3
> blue or brightness value

> position
> the direction of the light

> color
> Number[]|String|p5.Color:
> color Array, CSS color string, or p5.Color value

> x
> x axis direction

> y
> y axis direction

> z
> z axis direction

Change:

> v1
> red or hue value relative to the current color range

> v2
> green or saturation value relative to the current color range

> v3
> blue or brightness value relative to the current color range

> direction
> direction of light as a <a href="#/p5.Vector">p5.Vector</a>

> color
> p5.Color|Number[]|String:
> color as a <a href="#/p5.Color">p5.Color</a>, as an array, or as a CSS string

> x
> x component of direction (inclusive range of -1 to 1)

> y
> y component of direction (inclusive range of -1 to 1)

> z
> z component of direction (inclusive range of -1 to 1)

### [pointLight][2]

_1) Reword_

Note, I got the description of what a point light is from the [three.js documentation][9], since there was nothing on the [Processing page][10].

Current:

> Creates a point light with a color and a light position

> A maximum of 5 pointLight can be active at one time

Change:

> Creates a point light with the given color and position.

> A point light emits light from a single point in all directions.

> A maximum of **5** point lights can be active at once.

_2) Add disclaimer_

Change:

> Note: lights need to be called (whether directly or indirectly) within draw() to remain persistent in a looping program. Placing them in setup() will cause them to only have an effect the first time through the loop.

_3) Be consistent with params_

Current:

> v1
> red or hue value (depending on the current color mode),

> v2
> green or saturation value

> v3
> blue or brightness value

> x
> x axis position

> y
> y axis position

> z
> z axis position

> position
> the position of the light

> color
> Number[]|String|p5.Color:
> color Array, CSS color string, or p5.Color value

Change:

> v1
> red or hue value relative to the current color range

> v2
> green or saturation value relative to the current color range

> v3
> blue or brightness value relative to the current color range

> position
> position of light as a <a href="#/p5.Vector">p5.Vector</a>

> color
> p5.Color|Number[]|String:
> color as a <a href="#/p5.Color">p5.Color</a>, as an array, or as a CSS string

> x
> x component of position

> y
> y component of position

> z
> z component of position

### [spotLight][3]

_1) Reword_

Note, I got the description of what a spot light is from the [three.js documentation][11], since there was nothing on the [Processing page][12].

I don't have enough background knowledge to understand what the `angle` and `concentration` parameters are supposed to. Do the current descriptions make sense?

Current:

> Creates a spotlight with a given color, position, direction of light, angle and concentration. Here, angle refers to the opening or aperture of the cone of the spotlight, and concentration is used to focus the light towards the center. Both angle and concentration are optional, but if you want to provide concentration, you will also have to specify the angle.

> A maximum of 5 spotLight can be active at one time

Change:

> Creates a spot light with the given color, position, light direction, angle, and concentration.

> A spot light emits light from a single point in one direction along a conical shape. The `angle` parameter is used to determine the radius of the cone. And the `concentration` parameter is used to focus the light towards the center of the cone. Both parameters are optional, however if you want to specify `concentration`, you must also specify `angle`.

> The minimum concentration value is 1.

> A maximum of **5** spot lights can be active at once.

_2) Add disclaimer_

Change:

> Note: lights need to be called (whether directly or indirectly) within draw() to remain persistent in a looping program. Placing them in setup() will cause them to only have an effect the first time through the loop.

_3) Rename `conc` parameter to `concentration`_

_4) Be consistent with params_

Current:

> v1
> red or hue value (depending on the current color mode),

> v2
> green or saturation value

> v3
> blue or brightness value

> x
> x axis position

> y
> y axis position

> z
> z axis position

> rx
> x axis direction of light

> ry
> y axis direction of light

> rz
> z axis direction of light

> angle
> optional parameter for angle. Defaults to PI/3 (Optional)

> conc
> optional parameter for concentration. Defaults to 100 (Optional)

> color
> Number[]|String|p5.Color:
> color Array, CSS color string, or p5.Color value

> position
> the position of the light

> direction
> the direction of the light

Change:

> v1
> red or hue value relative to the current color range

> v2
> green or saturation value relative to the current color range

> v3
> blue or brightness value relative to the current color range

> x
> x component of position

> y
> y component of position

> z
> z component of position

> rx
> x component of light direction (inclusive range of -1 to 1)

> ry
> y component of light direction (inclusive range of -1 to 1)

> rz
> z component of light direction (inclusive range of -1 to 1)

> angle
> angle of cone. Defaults to PI/3

> concentration
> concentration of cone. Defaults to 100

> color
> p5.Color|Number[]|String:
> color as a <a href="#/p5.Color">p5.Color</a>, as an array, or as a CSS string

> position
> position of light as a <a href="#/p5.Vector">p5.Vector</a>

> direction
> direction of light as a <a href="#/p5.Vector">p5.Vector</a>

### [lightFalloff][6]

_1) Reword_

Current:

> Sets the falloff rates for point lights. It affects only the elements which are created after it in the code. The default value is lightFalloff(1.0, 0.0, 0.0), and the parameters are used to calculate the falloff with the following equation:

Change:

> Sets the falloff rate for <a href="#/p5/pointLight">pointLight()</a> and <a href="#/p5/spotLight">spotLight()</a>.

> lightFalloff() affects only the lights which are created after it in the code.

> The `constant`, `linear`, an `quadratic` parameters are used to calculate falloff as follows:

_2) Improve example_

[Separate PR over here][].

### [specularColor][7]

_1) Reword_

Current:

> Set's the color of the specular highlight when using a specular material and specular light.

> This method can be combined with specularMaterial() and shininess() functions to set specular highlights. The default color is white, ie (255, 255, 255), which is used if this method is not called before specularMaterial(). If this method is called without specularMaterial(), There will be no effect.

Change:

> Sets the color of the specular highlight of a non-ambient light (i.e. all lights except <a href="#/p5/ambientLight">ambientLight()</a>).

> specularColor() affects only the lights which are created after it in the code.

> This function is used in combination with <a href="#/p5/specularMaterial">specularMaterial()</a>. If a geometry does not use specularMaterial(), this function will have no effect.

> The default color is white (255, 255, 255), which is used if specularColor() is not explicitly called.

_2) Be consistent with params_

Current:

> value
> a color string

> gray
> a gray value

> values
> an array containing the red,green,blue & and alpha components of the color

> color
> the ambient light color

Change:

> value
> color as a CSS string

> gray
> number specifying value between white and black

> values
> color as an array containing the red, green, and blue components

> color
> color as a <a href="#/p5.Color">p5.Color</a>

_3) Improve example_

[Separate PR over here][].

### [lights][4]

_1) Reword_

Current:

> Sets the default ambient and directional light. The defaults are ambientLight(128, 128, 128) and directionalLight(128, 128, 128, 0, 0, -1).
> Lights need to be included in the draw() to remain persistent in a looping program. Placing them in the setup() of a looping program will cause them to only have an effect the first time through the loop.

Change:

> Places an ambient and directional light in the scene. The lights are set to ...

> Note: lights need to be called (whether directly or indirectly) within draw() to remain persistent in a looping program. Placing them in setup() will cause them to only have an effect the first time through the loop.

### [noLights()][5]

_2) Reword_

Current:

> This function will remove all the lights from the sketch for the subsequent materials rendered. It affects all the subsequent methods. Calls to lighting methods made after noLights() will re-enable lights in the sketch.

Change:

> Removes all lights present in a sketch.

> All subsequent geometry is rendered without lighting (until a new light is created with a call to one of the lighting functions (<a href="#/p5/lights">lights()</a>, <a href="#/p5/ambientLight">ambientLight()</a>, <a href="#/p5/directionalLight">directionalLight()</a>, <a href="#/p5/pointLight">pointLight()</a>, <a href="#/p5/spotLight">spotLight()</a>).

_3) Improve example_

[Separate PR over here][].

### General

_1) Move around variants so that consistent across_

directionalLight (change)

> directionalLight(v1, v2, v3, x, y, z)
> directionalLight(v1, v2, v3, direction)
> directionalLight(color, x, y, z)
> directionalLight(color, direction)

pointLight (no change)

> pointLight(v1, v2, v3, x, y, z)
> pointLight(v1, v2, v3, position)
> pointLight(color, x, y, z)
> pointLight(color, position)

### Screenshots of the change

_ambientLight_
f0_ambientLight

_directionalLight_
f1_directionalLight

_pointLight_
f2_pointLight

_spotLight_
f3_spotLight

_lights_
f4_lights

_noLights_
f5_noLights

_lightFalloff_
f6_lightsFalloff

_specularColor_
f7_specularColor

     
[0]: https://p5js.org/reference/#/p5/ambientLight
[1]: https://p5js.org/reference/#/p5/directionalLight
[2]: https://p5js.org/reference/#/p5/pointLight
[3]: https://p5js.org/reference/#/p5/spotLight
[4]: https://p5js.org/reference/#/p5/lights
[5]: https://p5js.org/reference/#/p5/noLights
[6]: https://p5js.org/reference/#/p5/lightFalloff
[7]: https://p5js.org/reference/#/p5/specularColor
[8]: https://processing.org/reference/
[9]: https://threejs.org/docs/#api/en/lights/PointLight
[10]: https://processing.org/reference/pointLight_.html
[11]: https://threejs.org/docs/#api/en/lights/SpotLight
[12]: https://processing.org/reference/spotLight_.html

### DIFF
--- a/src/webgl/light.js
+++ b/src/webgl/light.js
@@ -9,27 +9,41 @@ import p5 from '../core/main';
 import * as constants from '../core/constants';
 
 /**
- * Creates an ambient light with a color. Ambient light is light that comes from everywhere on the canvas.
- * It has no particular source.
+ * Creates an ambient light with the given color.
+ *
+ * Ambient light does not come from a specific direction.
+ * Objects are evenly lit from all sides. Ambient lights are
+ * almost always used in combination with other types of lights.
+ *
+ * Note: lights need to be called (whether directly or indirectly)
+ * within draw() to remain persistent in a looping program.
+ * Placing them in setup() will cause them to only have an effect
+ * the first time through the loop.
+ *
  * @method ambientLight
- * @param  {Number}        v1      red or hue value relative to
- *                                 the current color range
- * @param  {Number}        v2      green or saturation value
- *                                 relative to the current color range
- * @param  {Number}        v3      blue or brightness value
- *                                 relative to the current color range
- * @param  {Number}        [alpha] the alpha value
+ * @param  {Number}        v1       red or hue value relative to
+ *                                   the current color range
+ * @param  {Number}        v2       green or saturation value
+ *                                   relative to the current color range
+ * @param  {Number}        v3       blue or brightness value
+ *                                   relative to the current color range
+ * @param  {Number}        [alpha]  alpha value relative to current
+ *                                   color range (default is 0-255)
  * @chainable
  *
  * @example
  * <div>
  * <code>
  * createCanvas(100, 100, WEBGL);
- * ambientLight(0);
- * ambientMaterial(250);
+ * ambientLight(0); // black light (no light)
+ * ambientMaterial(255, 127, 80); // coral material
  * sphere(40);
  * </code>
  * </div>
+ * @alt
+ * sphere with coral color under black light
+ *
+ * @example
  * <div>
  * <code>
  * function setup() {
@@ -37,15 +51,22 @@ import * as constants from '../core/constants';
  * }
  * function draw() {
  *   background(51);
- *   ambientLight(100); // white light
- *   ambientMaterial(255, 102, 94); // magenta material
+ *   ambientLight(255); // white light
+ *   ambientMaterial(255, 127, 80); // coral material
  *   box(30);
  * }
  * </code>
  * </div>
  * @alt
- * evenly distributed light across a sphere
- * evenly distributed light across a rotating sphere
+ * sphere with coral color under white light
+ */
+
+/**
+ * @method ambientLight
+ * @param  {Number}        gray    number specifying value between
+ *                                  white and black
+ * @param  {Number}        [alpha]
+ * @chainable
  */
 
 /**
@@ -56,21 +77,14 @@ import * as constants from '../core/constants';
 
 /**
  * @method ambientLight
- * @param  {Number}        gray   a gray value
- * @param  {Number}        [alpha]
- * @chainable
- */
-
-/**
- * @method ambientLight
  * @param  {Number[]}      values  an array containing the red,green,blue &
- *                                 and alpha components of the color
+ *                                  and alpha components of the color
  * @chainable
  */
 
 /**
  * @method ambientLight
- * @param  {p5.Color}      color   the ambient light color
+ * @param  {p5.Color}      color   color as a <a href="#/p5.Color">p5.Color</a>
  * @chainable
  */
 p5.prototype.ambientLight = function(v1, v2, v3, a) {
@@ -90,25 +104,30 @@ p5.prototype.ambientLight = function(v1, v2, v3, a) {
 };
 
 /**
- * Set's the color of the specular highlight when using a specular material and
- * specular light.
+ * Sets the color of the specular highlight of a non-ambient light
+ * (i.e. all lights except <a href="#/p5/ambientLight">ambientLight()</a>).
  *
- * This method can be combined with specularMaterial() and shininess()
- * functions to set specular highlights. The default color is white, ie
- * (255, 255, 255), which is used if this method is not called before
- * specularMaterial(). If this method is called without specularMaterial(),
- * There will be no effect.
+ * specularColor() affects only the lights which are created after
+ * it in the code.
  *
- * Note: specularColor is equivalent to the processing function
+ * This function is used in combination with
+ * <a href="#/p5/specularMaterial">specularMaterial()</a>.
+ * If a geometry does not use specularMaterial(), this function
+ * will have no effect.
+ *
+ * The default color is white (255, 255, 255), which is used if
+ * specularColor() is not explicitly called.
+ *
+ * Note: specularColor is equivalent to the Processing function
  * <a href="https://processing.org/reference/lightSpecular_.html">lightSpecular</a>.
  *
  * @method specularColor
  * @param  {Number}        v1      red or hue value relative to
- *                                 the current color range
+ *                                  the current color range
  * @param  {Number}        v2      green or saturation value
- *                                 relative to the current color range
+ *                                  relative to the current color range
  * @param  {Number}        v3      blue or brightness value
- *                                 relative to the current color range
+ *                                  relative to the current color range
  * @chainable
  * @example
  * <div>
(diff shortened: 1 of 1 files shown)


src/webgl/light.js
```diff
@@ -9,68 +9,82 @@ import p5 from '../core/main';
...
+ *   ambientLight(255); // white light
+ *   ambientMaterial(255, 127, 80); // coral material
  *   box(30);
  * }
  * </code>
  * </div>
  * @alt
- * evenly distributed light across a sphere
- * evenly distributed light across a rotating sphere
+ * sphere with coral color under white light
```
stalgiag, line 61: This is a box not a sphere.

src/webgl/light.js
```diff
@@ -9,68 +9,82 @@ import p5 from '../core/main';
...
  * <div>
  * <code>
  * function setup() {
  *   createCanvas(100, 100, WEBGL);
  * }
  * function draw() {
  *   background(51);
- *   ambientLight(100); // white light
- *   ambientMaterial(255, 102, 94); // magenta material
+ *   ambientLight(255); // white light
```
stalgiag, line 54: I would try to keep as many visual elements similar between the block ambient light example and the white example (ie same shape, same background color).

stalgiag (commented):
It has been a long delay but luckily your thoughtful and thorough doc rewrites are still just as relevant. I requested one minor change. Thank you!!

src/webgl/light.js
JetStarBlues: good note, updated!

src/webgl/light.js
JetStarBlues: Good point. Updated both examples to be identical except for the single light color difference.

JetStarBlues:
@stalgiag  PR updated.
Not sure what the standard usage is for the `Resolve Conversation` button. Should I use it when I've updated the PR according to associated comments? Does it generate a notification?

stalgiag:
Thank you for this fantastic and thoughtful documentation edits @JetStarBlues !

PR #5314, opened 2021-06-15
Improve clarity of `specularColor()` example

JetStarBlues:
The current example does not do the best job of showcasing `specularColor`. Since the colors of the pointlights are set to the same value as the specular color, it is not clear what the specular color adds to the scene. For example, if you comment out the lines with calls to `specularColor()`, the change the omission makes is not obvious.

_current_
specularColorExample_current

_current with `specularColor` lines commented out_
specularColorExample_current_commentedOut

The example I have replaced it with aims to make the relationship more clear. Here is a [live version][0]. The light is given a white color, while the specular color is set to red. By clicking the mouse, you can toggle whether the specular color is set to red, or not set (and thus the default of white).

_change with red specular color_
specularColorExample_proposal

_change with default specular color_
specularColorExample_proposal_commentedOut

Change as docs:
specularColorExample_proposal_docs

[0]: https://editor.p5js.org/acedean/sketches/pTxAuDjfy


src/webgl/light.js
```diff
@@ -113,27 +113,40 @@ p5.prototype.ambientLight = function(v1, v2, v3, a) {
...
- *   pointLight(0, 255, 0, 0, 50, 50);
- *   specularMaterial(255);
- *   sphere(40);
+ *
+ *   ambientLight(60);
+ *
+ *   // add point light to showcase specular color
+ *   let locX = mouseX - width / 2;
+ *   let locY = mouseY - height / 2;
+ *   if (setSpecularColor) {
```
stalgiag, line 131: maybe conditional should go under the pointlight call so that the comment above reads like the intro to a relevant chunk?

src/webgl/light.js
```diff
@@ -113,27 +113,40 @@ p5.prototype.ambientLight = function(v1, v2, v3, a) {
  * @example
  * <div>
  * <code>
+ * let setSpecularColor = true;
```
stalgiag, line 116: `setRedSpecular`? In simple examples it feels good to see as much info as possible about the impact of a boolean in its name.

stalgiag (changes requested):
two tiny comments

JetStarBlues:
@stalgiag bumping this PR for review. Previously updated as per requested changes.

stalgiag (approved):

PR #5315, opened 2021-06-15
Improve clarity of `lightFalloff()` example

JetStarBlues:
In the current example, the translations to place the spheres are done globally (instead of using `push` and `pop`). As a result, the light positions are affected by these translations.

As a consequence of this, the color of the rightmost sphere is affected more by the light position than by falloff.

This is what the two spheres look like in the **current** example if the lines setting `lightFalloff()` are commented out:

falloffExample_current_commentedOut

What should happen instead is that both spheres are equally lit. Here is what they look like with `lightFalloff()` commented out in the **proposed** example:

falloffExample_proposal_commentedOut

In the proposed changes (live version [here][0]), the difference between the two spheres becomes attributable to falloff only.

Change as docs:
falloffExample_proposal_docs

[0]: https://editor.p5js.org/acedean/sketches/7lkU8Q9cv

### DIFF
--- a/src/webgl/light.js
+++ b/src/webgl/light.js
@@ -443,17 +443,26 @@ p5.prototype.lights = function() {
  *   noStroke();
  * }
  * function draw() {
+ *   ortho();
  *   background(0);
+ *
  *   let locX = mouseX - width / 2;
  *   let locY = mouseY - height / 2;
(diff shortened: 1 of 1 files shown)


stalgiag:
Great change!

PR #5319, opened 2021-06-17
Point to wiki in text() reference page

JetStarBlues:
Adds a note that points to the [wiki][0] as a resource for learning more about using text in WEBGL mode. There is a fair amount of subtlety that cannot be squeezed onto the [`text()` reference page][1].

textAddendum

[0]: https://github.com/processing/p5.js/wiki/Getting-started-with-WebGL-in-p5#text
[1]: https://p5js.org/reference/#/p5/text


lmccart:
@JetStarBlues for accessibility reasons, it's best to make the link text descriptive. Maybe you could update "here" to say "in the wiki"?

JetStarBlues:
Thank you for pointing this out!

outofambit (approved):
thank you!

PR #5320, opened 2021-06-17
Add `vertex()` variant to documentation

JetStarBlues:
# Add `vertex()` variant in documentation

The `vertex(x, y, u, v)` variant is currently missing from the [`vertex()` reference page][0]. In this variant, the omitted `z` defaults to `0`. It can be seen in use on the [`textureMode()` reference page][1].

_variants before change_
vertex_before

_variants after change_
vertex_proposal

[0]: https://p5js.org/reference/#/p5/vertex
[1]: https://p5js.org/reference/#/p5/textureMode

### DIFF
--- a/src/core/shape/vertex.js
+++ b/src/core/shape/vertex.js
@@ -958,9 +958,17 @@ p5.prototype.quadraticVertex = function(...args) {
  * @method vertex
  * @param  {Number} x
  * @param  {Number} y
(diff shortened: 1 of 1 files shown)


outofambit (approved):
@JetStarBlues thank you! i didn't know jsdoc would interpret an optional argument in the middle correctly like this. 😲

PR #5328, opened 2021-06-21
fix(typos): fix typos show up in docs

pearmini:


montoyamoraga (approved):
thanks so much!

montoyamoraga:
@all-contributors please add @pearmini for docs

pearmini:
> thanks so much!

You are welcome :)

src/core/main.js
```diff
@@ -377,7 +377,7 @@ class p5 {
         !this._loop ||
         time_since_last >= target_time_between_frames - epsilon
       ) {
-        //mandatory update values(matrixs and stack)
+        //mandatory update values(matrixes and stack)
```
limzykenneth, line 380:
```suggestion
        //mandatory update values(matrices and stack)
```
This should be `matrices` if it is plural or `matrix` if it is singular. Not sure whether it is singular or plural though.

limzykenneth (changes requested):

src/core/main.js
montoyamoraga: i looked it up here and both matrixes and matrices are correct?
https://thewordcounter.com/plural-of-matrix/
:)

limzykenneth: Hmm...well in that case if this is referring to a plural case we can accept as is then.

pearmini: So now sshould I change matrixes to matrices?

limzykenneth: No worries, it's fine. I'll merge as is. Thanks!

PR #5348, opened 2021-07-09
Add class method and tests for p5.Vector.toString()

weslord:
Addresses #4852

####  Changes
Add tests for `p5.Vector.prototype.toString()`
Add class method and tests for `p5.Vector.toString()`

#### Notes
Adding class method `p5.Vector.toString()` for consistency's sake, as per #4852, but noticed that this has the small side effect of overriding the inherited default method from `Function.prototype.toString()`, which returns the code that defines a function. I think it's unlikely anyone depends on that original behaviour, but since it's a breaking change I figured I'd split this particular change off of the rest of the static/class methods in #5044. This way we can discuss it separately.

One weird ramification is that defining `p5.Vector.toString = (v) => v.toString()` caused problems with unit tests for instanceof p5.Vector - I haven't dug in to figure out why, but I did find out that `p5.Vector.toString = (v) => String(v)` seems to work just fine. Strange, right?

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -2364,4 +2364,23 @@ p5.Vector.normalize = function normalize(v, target) {
   return target.normalize();
 };
 
+/**
+ * Returns a string representation of a vector v. This method is useful for
+ * logging vectors in the console.
+ * Note that the static method p5.Vector.toString() overrides the existing
+ * inherited method Function.prototype.toString().
+ */
+/**
+ * @method toString
+ * @static
+ * @param  {p5.Vector}    v the vector to stringify
+ * @return {String}       the string representation
+ */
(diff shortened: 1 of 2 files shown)


Qianqianye:
Thanks @weslord! Add the Math area stewards @limzykenneth, @jeffawang, @AdilRabbani to review the PR.

limzykenneth:
Because `toString` is usually a built in function with specific expectation, can the implementation here be made so that it returns what it would have returned originally when no arguments are passed to the static function? (This is also the reason why the `.toString()` version doesn't work btw)

Qianqianye:
Hi @weslord, we'd like to follow up with this pull request. Will you be able to respond to @limzykenneth's comments? Please let us know if you need any clarification. Thanks.

PR #5353, opened 2021-07-15
Fixed bug were cached colors are not cleared

0xJonas:
#### Changes:
When using the `setRed()/setGreen()/setBlue()/setAlpha()` methods of `p5.Color`, cached HSL and HSB values are not cleared properly. This leads to the following bug:

```javascript
const c = color(255, 0, 0);
console.log(lightness(c)) // 50
c.setRed(0);
console.log(lightness(c)); // bug: still 50
```
The first call to `lightness()` computes the HSL values and caches them, the second call retrieves the cached values even though the RGB values have changed. The fix is simple: Clear the `hsba `and `hsla `fields of `p5.Color` when calling `setRed()/setGreen()/setBlue()/setAlpha()`.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -365,6 +365,10 @@ p5.Color.prototype._calculateLevels = function() {
   for (let i = array.length - 1; i >= 0; --i) {
     levels[i] = Math.round(array[i] * 255);
   }
+
+  // Clear cached HSL/HSB values
+  this.hsla = null;
+  this.hsba = null;
 };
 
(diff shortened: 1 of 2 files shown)


joecridge:
Nice catch!

It might be better to test that the value returned from `lightness(c)` changes when you call `c.setRed(100)` etc (rather than looking at the caches directly), but either way this LGTM – thanks :)

montoyamoraga:
@all-contributors please add @0xJonas for bug

PR #5146, opened 2021-03-31
Add text wrapping function

lawreka:
Resolves #5081
Addresses #4652 
Follow-up to #5108 

Changes:

Adds **textWrap** function to specify how lines of text are wrapped within a text box. This requires a max-width set on the text area, specified in [text()](https://p5js.org/reference/#/p5/text) as parameter `x2`.
* WORD wrap style only breaks lines at spaces. A single string without spaces that exceeds the boundaries of the canvas or text area is not truncated, and will overflow the desired area, disappearing at the canvas edge.
* CHAR wrap style breaks lines wherever needed to stay within the text box.

WORD is the default wrap style, and both styles will still break lines at any line breaks (\n) specified in the original text. The text area max-height parameter (y2) also still applies to wrapped text in both styles, lines of text that do not fit within the text area will not be drawn to the screen.

I may have gone a little overboard adding documentation and examples, specifically the last one with the little screaming stick-man demonstrating the current lack of truncation. He should be removed if additional functionality is added in the future to deal with truncation and hyphenation, but for now I think he shows kind of a fun way to work with #4652.

I also added an example using Chinese characters, since as @dhowe mentioned in discussion, textWrap(CHAR) is essential for displaying text in character-based languages nicely.

 Screenshots of the change:
Screenshot 2021-03-31 at 21 56 56

#### PR Checklist
- [x] `npm run lint` passes
- [x] [Inline documentation] is included / updated
- [x] [Unit tests] are included / updated

PR #5366, opened 2021-07-28
Fixing ReferenceError: totalHeight is not defined throw by textAlign(…

### LINKED ISSUE #5360: Undefined totalHeight in text function when setting vertical alignment

- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [x] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:

Using textAlign(CENTER, CENTER) causes 'ReferenceError: totalHeight is not defined' when using the text function.

- p5.js version: 1.4.0
- Web browser and version: Google Chrome	91.0.4472.164 (Official Build) (x86_64)
- Operating System: macOS Version 10.15.7 (Build 19H1030)
- Steps to reproduce this:
```
function setup() {
  createCanvas(400, 400);
}

function draw() {
  textAlign(CENTER, CENTER)  
  text("example text", width/2, height/2, 300, 100)
}
```

Console error:
```
Uncaught ReferenceError: totalHeight is not defined
    at _main.default.Renderer2D._main.default.Renderer.text (p5.js:64912)
    at _main.default.Renderer2D.text (p5.js:66355)
    at p5._main.default.text (p5.js:93356)
    at draw (24871d84-9d19-45cd-b2fe-c6baebca1e56:7)
    at p5._main.default.redraw (p5.js:70809)
    at _draw (p5.js:63046)
```

malviys:
Fixing ReferenceError: totalHeight is not defined throw by textAlign(___, CENTER) and textAlign(___, BOTTOM)
Resolves #5360 

 Changes:
Removed undefine variable ```totalHeight``` from src/core/p5.Renderer.js causing ReferenceError while calling 
textAlign_(___, CENTER) and textAlign(_____, BOTTOM)

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/p5.Renderer.js
+++ b/src/core/p5.Renderer.js
@@ -269,11 +269,11 @@ p5.Renderer.prototype.text = function(str, x, y, maxWidth, maxHeight) {
 
       switch (this._textBaseline) {
         case constants.BOTTOM:
-          shiftedY = y + (maxHeight - totalHeight);
+          shiftedY = y + maxHeight;
           y = Math.max(shiftedY, y);
           break;
         case constants.CENTER:
-          shiftedY = y + (maxHeight - totalHeight) / 2;
+          shiftedY = y + maxHeight / 2;
           y = Math.max(shiftedY, y);
           break;
         case constants.BASELINE:


outofambit:
@malviys thank you for working on this! could you please include screenshots of using `textAlign` to help confirm that we now have the correct behavior (besides not erroring)? thank you!

(also ccing @lawreka in case you have thoughts on the approach here!)

lawreka:
Thanks for tagging me here, @outofambit ! It looks like I left these uninitialized variables in when refactoring the text wrapping :( 

This fix looks like it does eliminate the ReferenceError, but I'm digging in a little bit into #5146 to see if I missed something important about how `totalHeight` was used. I would expect in this example that if the text is vertically aligned to the center of the box (where the line is drawn through), the whole multi-line block of text would shift up a little bit.

Screenshot 2021-08-24 at 18 52 34

As it is, only the first line is conforming to the vertical-alignment rule. I would assume also with text-align set to BOTTOM that we don't see the last line aligned with the bottom, but the first line.

If you want to get rid of the error quickly and plan to release soon, this is a good fix, but I'm also happy to revisit the text wrapping function and see what got lost in re-implementation.

malviys:
Here is code screenshot of ```textAlign```. Putting CENTER or BOTTOM as any parameter throws ReferenceError, now it is working fine. @lawreka I checked in source code that variable ```totalHeight``` is not declared and used anywhere.

p5 js - reffrence error code
p5 js - reffrence error sketch

Snowman-s:
I had been struggling with this problem for a while. Thanks for coming up with a solution!

Here are my thoughts:

Presumably, the `totalHeight` was the actual height of the text as it was drawn.

This value would have been used to draw the entire text across multiple lines at the specified position, not just a single line.
(In fact, before #5146,  `totalHeight` was declared and used.)

I noticed that it is not easy to calculate the `totalHeight`, because the height of the text changes for each `textWrapStyle`.
You probably need to split the text and calculate what the height will be beforehand (if you use `totalHeight`).

Snowman-s:
@malviys @lawreka, 

It would be great if you could tell me what the current status of this Pull Request is. Thanks!

lawreka:
> You probably need to split the text and calculate what the height will be beforehand

Yes, this would require a more in-depth solution, which would probably necessitate its own PR separate from this one. I thought I would have time to work on that when I commented 17 days ago but unfortunately this was not the case :( 

>  what the current status of this Pull Request is

I think that will be up to @outofambit to approve, I'm happy with @malviys solving the error in this PR and to open a new issue/follow-up PR to actually implement the correct vertical alignment.

outofambit:
hey @lawreka @Snowman-s @malviys! thanks for the great conversation here. it seems like we should merge this PR to fix the immediate error and then follow up with more work to re-implement `totalHeight` for multiline bottom aligned text areas. i'll merge this now and file an issue to followup on the second part. thanks again!

outofambit (approved):

PR #5394, opened 2021-08-29
Update dom.js comments

WayGuan:
Changes:
- Updated a comment in dom.js regarding the position( ) method. The original comment was a copy-and-paste which is not suitable here.


limzykenneth:
Looks good. Thanks!

PR #5415, opened 2021-09-15
Inline examples updated to remove "Object reference chain is too long" error

### LINKED ISSUE #5367: Unable to build library using grunt. Fatal error: Protocol error (Runtime.callFunctionOn): Object reference chain is too long.

- [x] Build tools and processes
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:
Unable to build library using grunt

- p5.js version: 1.4.0
- Web browser and version: Google Chrome Version 92.0.4515.107 (Official Build) (64-bit)
- Operating System: Windows 10 Build 19043.1110
- Node version: 16.5.0 
- Steps to reproduce this:
```bash
npm run grunt
```

Console error:
```bash
Fatal error: Protocol error (Runtime.callFunctionOn): Object reference chain is too long
```

almchung:
Resolves #5367 (maybe 😅)

 Changes:
Some examples from `math/p5.Vector` and `utilities/array_functions` used `print([Object]);`. 
Changed the lines to print simple string instead: `print([Object].toString());`

This now runs without the `Object reference chain is too long` error (using puppeteer v10.2.0).

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -274,7 +274,7 @@ p5.Vector.prototype.copy = function copy() {
  *
  * let v3 = p5.Vector.add(v1, v2);
  * // v3 has components [3, 5, 7]
- * print(v3);
+ * print(String(v3));
  * </code>
  * </div>
  *
@@ -384,7 +384,7 @@ const calculateRemainder3D = function(xComponent, yComponent, zComponent) {
  *
  * let v3 = p5.Vector.rem(v1, v2);
  * // v3 has components [1, 1, 1]
- * print(v3);
+ * print(String(v3));
  * </code>
  * </div>
  */
@@ -482,7 +482,7 @@ p5.Vector.prototype.rem = function rem(x, y, z) {
  *
  * let v3 = p5.Vector.sub(v1, v2);
  * // v3 has components [1, 1, 1]
- * print(v3);
+ * print(String(v3));
  * </code>
  * </div>
  *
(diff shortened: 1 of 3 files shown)


limzykenneth:
It does solve it but I'd prefer not to alter the examples and set limit on them as they are running fine on the website itself.

almchung:
@limzykenneth  The parts of the example code I modified are all inside `print()` and happening inside the console, so I thought it would not alter the webpage examples too much.

limzykenneth:
Yeah, that's not quite what I mean. I mean editing the examples in general to resolve this issue and also if someone authors an example that prints a vector they likely need to convert it to string as well which is a bit unintuitive.

almchung:
I completely agree that the original examples are way much more intuitive and elegant.
On the other hand, the modified ones do highlight that `p5.Vector` is a class and its `toString()` method, which is designed for logging vectors in the console. 

I don't have a strong preference for either way and actually have mixed feelings about my solution presented here. Would there be a way to conditionally exclude certain examples only for the headless test?

limzykenneth:
I'm more inclined to investigate if playwright that you mentioned in the issue would work for us. There's also the open issue on puppeteer that I opened which may help as well.

almchung:
That sounds great. I will file a separate issue on Playwright after I run some basic setup tests.

PR #5424, opened 2021-09-22
Update attributes.js

UnityOfFairfax:
Changes:

Minor copy edits to docstrings.


lmccart:
@UnityOfFairfax it looks like there are some linting issues, could you try building on your end and see if you can resolve them? I think they may just be extra whitespace issues.

PR #5439, opened 2021-10-05
Added wrap-up of gsoc-21 @satyasaibhushan

satyasaibhushan:
added wrap-up file and add it to the list in Readme.md


satyasaibhushan:
Also, for the assets used while writing the wrap-up, I had to use the links to my repo.
I think maintaining an `assets` folder inside the `project_wrapups` folder might come in handy for all the students.

outofambit:
thank you @satyasaibhushan! could you add [alt text](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-alt) for those two images in the last section? then i can merge. thank you!

PR #5443, opened 2021-10-08
add internal functions for changing translation language in i18next

outofambit:
This is some foundational work for allowing the web editor to change the translation language used for FES messages in a running p5.js sketch. These changes don't expose any functionality externally, I've just written wrappers around the underlying `i18next` functions that make these changes so its easier for others to work with within p5.js.

cc @almchung

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/internationalization.js
+++ b/src/core/internationalization.js
@@ -167,3 +167,28 @@ export const initialize = () => {
   // we have the translation files.
   return i18init;
 };
+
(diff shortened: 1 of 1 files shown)


almchung:
This is great! I'll wait for the merge and then add the web editor - FES i18n integration features to this.

PR #5454, opened 2021-10-17
improve the zooming interaction

yangshuzhan:
Changes:
Improve the zooming interaction in webgl mode. Now you can zoom smoothly.

### DIFF
--- a/src/webgl/interaction.js
+++ b/src/webgl/interaction.js
@@ -88,16 +88,16 @@ p5.prototype.orbitControl = function(sensitivityX, sensitivityY, sensitivityZ) {
     this._setProperty('wheelDefaultDisabled', true);
   }
 
-  const scaleFactor = this.height < this.width ? this.height : this.width;
-
+  const scaleFactor = 50;
   // ZOOM if there is a change in mouseWheelDelta
-  if (this._mouseWheelDeltaY !== this._pmouseWheelDeltaY) {
+  if (this._mouseWheelDeltaY !=0) {
     // zoom according to direction of mouseWheelDeltaY rather than value
     if (this._mouseWheelDeltaY > 0) {
       this._renderer._curCamera._orbit(0, 0, sensitivityZ * scaleFactor);
     } else {
       this._renderer._curCamera._orbit(0, 0, -sensitivityZ * scaleFactor);
     }
+    this._mouseWheelDeltaY=0;
   }
 
   if (this.mouseIsPressed) {
--- a/src/webgl/interaction.js
+++ b/src/webgl/interaction.js
@@ -102,6 +102,8 @@ p5.prototype.orbitControl = function(sensitivityX, sensitivityY, sensitivityZ) {
 
   if (this.mouseIsPressed) {
     // ORBIT BEHAVIOR
+    if(Math.abs(this.mouseX-this.pmouseX)>50||Math.abs(this.mouseY-this.pmouseY)>50)
+      return this;//when noloop() is disabled,need a frame to reset pmouse position 
     if (this.mouseButton === this.LEFT) {
       const deltaTheta =
         -sensitivityX * (this.mouseX - this.pmouseX) / scaleFactor;


outofambit:
hi @yangshuzhan could you share more about the problem you are solving with this PR? if you are able to attach before and after gifs to show how zooming works differently, that would be very helpful!

yangshuzhan:
> hi @yangshuzhan could you share more about the problem you are solving with this PR? if you are able to attach before and after gifs to show how zooming works differently, that would be very helpful!

Sorry that I didn't describe it clearly. 
before:
zoom-before
after:
zoom-now

stalgiag:
Thanks for this @yangshuzhan and for the documentation of the example. I do find that this improves the quality of interaction for the zooming with `orbitControls()`. I would advocate for merging but it would be useful if there was an issue proposing this change. This is outlined in [this section of the Contributors Docs](https://p5js.org/contributor-docs/#/README?id=github-issue-flow) in the 'If you discover a bug or have an idea for a new feature you'd like to add...' bullet.

Once you open an issue to propose this change then you can fix the linting and testing errors. There is more info on this in the Contributors Docs (linked above). If you have trouble, feel free to ping me.

Qianqianye:
Thanks @yangshuzhan! Like @stalgiag mentioned, an issue needs to be opened before a pull request is opened, then you can tag the issue in the pull request. You can read more about it in the [GitHub Issue Flow](https://github.com/processing/p5.js/tree/main/contributor_docs#github-issue-flow). This is necessary for tracking the development and keeping discussions clear. Thanks!

src/webgl/interaction.js
```diff
@@ -88,20 +88,22 @@ p5.prototype.orbitControl = function(sensitivityX, sensitivityY, sensitivityZ) {
     this._setProperty('wheelDefaultDisabled', true);
   }
 
-  const scaleFactor = this.height < this.width ? this.height : this.width;
-
+  const scaleFactor = 50;
```
davepagurek, line 91: Do you think we can make this scale factor be relative to the width and height? It's also used to orbit via click and drag, and when the canvas is much larger than 50, it means dragging across the canvas rotates more than one would expect. Here's an example of the current behaviour, and then after a quick change to try to make it something relative to the canvas size:

zoomscalefactor

src/webgl/interaction.js
```diff
@@ -88,20 +88,22 @@ p5.prototype.orbitControl = function(sensitivityX, sensitivityY, sensitivityZ) {
...
   // ZOOM if there is a change in mouseWheelDelta
-  if (this._mouseWheelDeltaY !== this._pmouseWheelDeltaY) {
+  if (this._mouseWheelDeltaY !=0) {
```
davepagurek, line 93: Nice, this definitely feels a lot smoother!

src/webgl/interaction.js
```diff
@@ -88,20 +88,22 @@ p5.prototype.orbitControl = function(sensitivityX, sensitivityY, sensitivityZ) {
...
     } else {
       this._renderer._curCamera._orbit(0, 0, -sensitivityZ * scaleFactor);
     }
+    this._mouseWheelDeltaY=0;
   }
 
   if (this.mouseIsPressed) {
     // ORBIT BEHAVIOR
+    if(Math.abs(this.mouseX-this.pmouseX)>50||Math.abs(this.mouseY-this.pmouseY)>50)
+      return this;//when noloop() is disabled,need a frame to reset pmouse position
```
davepagurek, line 106: Makes sense, I see that if I start off with `noLoop()` and then call `loop()` in e.g. `mousePressed`, the orbit jumps. Maybe we can rephrase the comment a bit to make this clearer, something like:
"Ignore abnormally large mouse movements, which might occur if the draw loop has been paused and just became unpaused"

davepagurek (changes requested):
Thanks for this change, the zooming motion definitely is smoother! I think before merging, we just need two things:
- Right now, in making zoom smoother, we accidentally also made orbiting much more sensitive, so we should make the scale amount be relative to the canvas size again
- The eslint check is currently failing due to these issues, which we'll need to resolve:
    ```
  /home/runner/work/p5.js/p5.js/src/webgl/interaction.js:96:30: Expected '!==' and instead saw '!='. [Error/eqeqeq]
  /home/runner/work/p5.js/p5.js/src/webgl/interaction.js:108:1: This line has a length of 84. Maximum allowed is 80. [Error/max-len]
  /home/runner/work/p5.js/p5.js/src/webgl/interaction.js:109:84: Trailing spaces not allowed. [Error/no-trailing-spaces]
  ```
  Additionally, while not an error, it would be great if we could match the spacing in the rest of the repo by putting spaces around operators, and adding spaces between `if` and the condition.

inaridarkfox4231:
excuse me.
I have suggestions.
here is sample code: [improve zooming](https://editor.p5js.org/dark_fox/sketches/aVrbgFmL9)

There are three main changes:

change0:
change default value of sensitivityZ to 1

change1:
Change the scale factor for zoom to 0.03
There may be differences depending on the specifications of the mouse, but even if there is, it will be difficult to deal with. At least this value was comfortable in my environment.
The processing for mousewheel delta is the same.

change2:
Change the scale using logarithms.
Change the lower bound of the distance to 0.001.
Using logarithms allows us to treat small and large bounds equally.
Also, the lower limit of the smaller one is slightly smaller considering the case where you want to create a picture on a small scale.

davepagurek:
I'm going to close this PR for organizational purposes as https://github.com/processing/p5.js/pull/6105 also addresses similar issues, but feel free to open a new issue if there are more changes you'd like to see made!

PR #5479, opened 2021-11-24
feat: enable Github Issue Forms

### LINKED ISSUE #5414: Github Issue Forms
Make filing an issue easier and remove the need to navigate potentially confusing markdown structure of existing issue template.

- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

I just came across this: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms and it definitely is an improvement over our existing issue template (no more relying on markdown to select sub-area to report) and much easier/cleaner to use from the end user perspective.

Not sure how stable the feature is currently as it is still in beta but would very much like to see this implemented for p5.js.

willmartian:
Resolves #5414 

 **Changes:**

Replaced the existing Markdown issue templates with GitHub issue forms. I believe that the forms make it easier to contribute to p5.js, which furthers its mission of improving access. 

To preview the new functionality, feel free to [open an example issue in my fork.](https://github.com/willmartian/p5.js/issues/new/choose)

 Screenshots of the change:

**User view upon selecting "new issue":**
image
**User view upon selecting "Existing Feature Enhancement":**
image


.github/ISSUE_TEMPLATE/found-a-bug.yml
```diff
@@ -0,0 +1,51 @@
...
+      - label: Image
+      - label: IO (Input/Output)
+      - label: Localization
+      - label: Math
+      - label: Unit Testing
+      - label: Typography
+      - label: Utilities
+      - label: WebGL
+      - label: Other (specify if possible)
+- type: textarea
```
limzykenneth, line 28: Can this be type "input" instead as it should be a simple field.

.github/ISSUE_TEMPLATE/found-a-bug.yml
```diff
@@ -0,0 +1,51 @@
...
+  attributes:
+    label: p5.js version 
+    description: You can find this in the first line of the p5.js file.
+  validations:
+    required: false
+- type: textarea
```
limzykenneth, line 34: "input" here as well.

.github/ISSUE_TEMPLATE/found-a-bug.yml
limzykenneth, line 28: Can this be type "input" instead as it should be a simple field?

.github/ISSUE_TEMPLATE/found-a-bug.yml
limzykenneth, line 34: "input" here as well.

.github/ISSUE_TEMPLATE/found-a-bug.yml
```diff
@@ -0,0 +1,51 @@
...
+  attributes:
+    label: Web browser and version 
+    description: In the address bar, on Chrome enter "chrome://version", on Firefox enter "about:support". On Safari, use "About Safari".
+  validations:
+    required: false
+- type: textarea
```
limzykenneth, line 40: "input" here as well.

.github/ISSUE_TEMPLATE/found-a-bug.yml
```diff
@@ -0,0 +1,51 @@
...
+  attributes:
+    label: Operating System 
+    description: "Ex: Windows/MacOSX/Linux/Android/iOS along with version"
+  validations:
+    required: false
+- type: textarea
```
limzykenneth, line 46: Would be nice if it can be some kind of code field but not sure if that is possible at all.

limzykenneth (changes requested):
This looks better than I expected and I think really would make not just authoring an issue but also for us going through an open issue much easier!

A few changes inline.

@outofambit @Qianqianye Would you like to make a look at this as well?

.github/ISSUE_TEMPLATE/found-a-bug.yml
willmartian: Hmm I don't think there is a code block field (although that is a good feat request for GitHub). However, I can add a default value to the field: 

image

Which is then rendered as the following:

image

What do you think? In terms of a friendly experience for first timers, I am not sure if having a default code block is better or no markdown at all is better.

limzykenneth: Is it able to add a blank line in the template shown to the user right above `### Snippet`? Just to make things a bit neater.

outofambit: I like the code block! maybe inside of the ticks it could say

```js
# Paste your code here :)
```

so if you get intimidated by the code block formatting, there's a clear invitation for where to put your example?

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,38 @@
+name: 🌱 New Feature Request
+description: This template is for requesting a new feature be added.
+labels: [feature request]
+body:
+- type: textarea
+  attributes:
+    label: Increasing Access 
+    description: How would this new feature help [increase access](https://github.com/processing/p5.js/blob/main/contributor_docs/access.md) to p5.js?
```
outofambit, line 8: I like the idea behind this being required, but I'd also like to give people an out if they don't know. I've seen folks submit an idea with this question blank and have others chime in with their own illuminating ideas. Maybe we could do something like this?

```suggestion
    description: How would this new feature help [increase access](https://github.com/processing/p5.js/blob/main/contributor_docs/access.md) to p5.js? (If you're not sure, you can type "Unsure" here and let others from the community offer their thoughts.)
```

outofambit (commented):
thanks for doing this @willmartian, I'm really excited to get this in! left a few comments, let me know what you think? :D

willmartian:
Thanks for the reviews all! And the patience, I stopped looking at GitHub over the holiday :) 

I think I implemented all of the requested changes--let me know if there is anything else @outofambit

limzykenneth (approved):

limzykenneth:
Thanks! I'll merge this in a couple days if there are no further comments.

limzykenneth:
Merged. There's a few things I may want to tweak but for now, let's give this a go!

PR #5484, opened 2021-11-28
Fixed GIF Freezing After Returning to Unfocused Tab

### LINKED ISSUE #5434: Animated GIFs freeze after the browser tab is off screen.

- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [X] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

### Details about the bug:

- p5.js version: 1.4.0
- Web browser and version: Brave Version 1.30.86 Chromium: 94.0.4606.61 (Official Build) (x86_64)
- Operating System: MacOSX 10.15.7

#### Repro Steps

1. Open a sketch that displays an animated GIF using `image()` in every call to `draw()` with all default settings.
1. Make sure the GIF is set to animate forever (no limit on the number of loops).
1. Switch to a different desktop (so that the browser window is off screen), or switch to a different tab within the same browser window.
1. Switch back to the original browser tab.

#### Expected

The GIF should still be animating

#### Actual

The GIF is frozen for some period of time (the longer the tab was off screen the longer it is frozen).

#### Example

[https://editor.p5js.org/Kumu-Paul/sketches/T_SbGe-yK](https://editor.p5js.org/Kumu-Paul/sketches/T_SbGe-yK)

```javascript
let img;

function preload() {
  img = loadImage("https://www.paulwheeler.us/files/catbus-2.gif");
}

function setup() {
  createCanvas(300, 300);
}

function draw() {
  background(200);

  scale(0.4, 0.4);

  image(img, 10, 10);
}
```

#### Cause

[p5.Image.js line 227](https://github.com/processing/p5.js/blob/main/src/image/p5.Image.js#L227)

```
const curTime = pInst._lastFrameTime + pInst.deltaTime;
```

This use of `deltaTime` is incorrect. This line should just be:

```
const curTime = pInst.millis();
```

frigorific44:
Resolves #5434 

 Changes:
GIFs now only use `_lastFrameTime` to track when to show the next frame. A sum with `deltaTime` was used before to estimate what the current time was in a way which would keep frames across all GIFs synchronized, but had the side effect of freezing a GIF when returning from a different tab or window, as the estimate would use the large `deltaTime` between the previous frames to estimate a `curTime` which has not yet been reached. Everything is effectively the same, except GIF frames are no longer translated by the running-average of `deltaTime`.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/image/p5.Image.js
+++ b/src/image/p5.Image.js
@@ -224,7 +224,7 @@ p5.Image = function(width, height) {
  */
 p5.Image.prototype._animateGif = function(pInst) {
   const props = this.gifProperties;
-  const curTime = pInst._lastFrameTime + pInst.deltaTime;
+  const curTime = pInst._lastFrameTime;
   if (props.lastChangeTime === 0) {
     props.lastChangeTime = curTime;
   }


stalgiag:
Thank you for this contribution! Returning to this after a long time away and trying to wrap my head around the messy stuff done here in the name of 'real-time'. I sometimes return to #3380 as that was the original spec of the GIF implementation.

First things first, I am not sure but I believe that `+ deltaTime` was in place to account for the time between the current frame and the previous frame. This is a minor sync issue and is probably a worthy trade to keep the GIFs from breaking with unfocused tabs.

Second, this PR is affected by a different but similar problem. It seems that when a GIF is not looping and the tab is unfocused long enough to reach the end, p5 never paints the final frame. This wasn't introduced by you but is likely a side effect of fixing the main issue with unfocused tabs.

I would be willing to merge this if we open another issue that clearly explains the known imperfections of the current implementation.

stalgiag:
Thank you!!

PR #5487, opened 2021-12-01
p5.renderer: allow default renderer to be temporarily replaced

osresearch:
Changes:
This patch adds the `p5.renderer()` method to make it easier to replace the default renderer with a `p5.Graphics` object.  When the default renderer is replaced with an off-screen `p5.Renderer` object, the normal drawing calls will render into the off screen object instead of the canvas. The method returns the original renderer, which can be restored later and the off screen renderers can then be combined into a single image.

The intended use is for combining multiple sketches that call methods like `rect()` or `line()` directly instead of through a `p5.Graphics` argument.

#### PR Checklist
- [X] `npm run lint` passes

### DIFF
--- a/src/core/rendering.js
+++ b/src/core/rendering.js
@@ -351,4 +351,45 @@ p5.prototype.blendMode = function(mode) {
  * white ellipse with shadow blur effect around edges
  */
 
+/**
+ * Temporarily override the default p5 renderer.
+ *
+ * @method renderer
+ * @param  {p5.Renderer} new default renderer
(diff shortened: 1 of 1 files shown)


limzykenneth:
Hi @osresearch, we usually require an issue to be filed and discussed there before any PR should be filed, especially for new feature requests which have additional requirement of meeting our [accessibility requirement](https://github.com/processing/p5.js/blob/main/contributor_docs/access.md).

I'm going to close this PR for now but please do open an issue for this feature request. Thanks!

PR #5503, opened 2021-12-12
Made changes referencing issue #5374

### LINKED ISSUE #5374: radio.selected(value) does not work with radio buttons holding "value" and "label"

- [ ] Color
- [ ] Data
- [x] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:

- p5.js version: 1.4.0
- Web browser and version:  Chrome and 92.0.4515.131 (Official Build)  (64 bit) (cohort: Stable)
- Steps to reproduce this:

Hi all,
In the following code, the "red" button is not selected. Am I missing something?
```
let radio; 

function setup() {
  radio = createRadio(); 
  radio.option(1, "red"); 
  radio.option(2, "blue"); 
  radio.option(3, "green");
  radio.selected(1); 
}

function draw() {

}
```

reejuBhattacharya:
Resolves #5374

 Changes: changed the comparison in the values of the selected function to compare between two strings.

 Screenshots of the change:
image

#### PR Checklist

- [x] `npm run lint` passes


reejuBhattacharya:
Hi, 
I would love it if you would check my pull request. Tagging @outofambit (DOM steward) and @limzykenneth (for actively suggesting solutions to this issue)

reejuBhattacharya:
@outofambit , @limzykenneth.  
Tagging because it has been 29 days. Would love some feedback on this pull request.

lib/empty-example/sketch.js
```diff
@@ -1,7 +1,7 @@
...
-function draw() {
-  // put drawing code here
-}
\ No newline at end of file
+    // put setup code here
+  }
+  
+  function draw() {
+    // put drawing code here
+  }
```
limzykenneth, line 7: Can the changes here be removed as it is not related to the issue?

src/dom/dom.js
```diff
@@ -953,7 +953,7 @@ p5.prototype.createRadio = function() {
       }
     } else {
       for (const option of self._getOptionsArray()) {
-        if (option.value === value) {
+        if (option.value === value.toString()) {
```
limzykenneth, line 956: I'm having second thought about this as it doesn't work in all scenarios. I think it may be better to note in the documentation recommending only string values be used for `.option` and `.selected`. But that's just what I'm thinking right now, not sure if it's the best solution here.

lib/empty-example/sketch.js
reejuBhattacharya: Actually, this is not really a change. I used the sketch.js file in the empty example to test my changes in the local machine. After I was done, I replaced it with the original template that it had.

src/dom/dom.js
reejuBhattacharya: My thought process about this was: since `option.value` will always be a string, we could check it strictly with a string too. This way, the user can put a string, integer or character value as the `value` and it will convert it to a string and then make comparisons with the original value, which was automatically converted to a string by the HTML.
Could you help me with figuring out where this might not work? 
Thanks for replying!

limzykenneth: `null` does not work and all objects will get casted into `[object Object]`. If the documentation does not make clear the expected value, casted or otherwise, would be, implicit casting into strings may not be enough to avoid potential confusion.

lib/empty-example/sketch.js
limzykenneth: It is registered here as a change because the lines highlighted have additional white spaces at the beginning which shouldn't be there.

src/dom/dom.js
reejuBhattacharya: I understand the problem now. I will make changes to the documentation to resolve this issue, and make a pull request.
Appreciate the help!

PR #5578, opened 2022-01-22
Updated some JsDoc definitions

Gaweph:
- Minor fixes to JsDoc comments that are preventing typescript generation
- Added Method overload for `applyMatrix()` - added support for 4x4

(JsDoc definition are used in the generation of the typescript files @ https://github.com/p5-types/p5.ts used by https://github.com/DefinitelyTyped/DefinitelyTyped)


src/core/transform.js
```diff
@@ -25,14 +25,45 @@ import p5 from './main';
...
+ * @param  {Number} d_2x3 numbers which define the 2×3 matrix to be multiplied
+ * @param  {Number} e_2x3 numbers which define the 2×3 matrix to be multiplied
+ * @param  {Number} f_2x3 numbers which define the 2×3 matrix to be multiplied
+ * @chainable
+ * 
+ */
+  
+/**
+* @method applyMatrix
+* @param  {Number} a_4x4 numbers which define the 4x4 matrix to be multiplied
```
Gaweph, line 47: Rationale:

`p5.RendererGL.applyMAtrix(arguments)` is called from this method which then expects either 6 or 16 values

```js
p5.RendererGL.prototype.applyMatrix = function(a, b, c, d, e, f) {
  if (arguments.length === 16) { // 4x4 matrix
    p5.Matrix.prototype.apply.apply(this.uMVMatrix, arguments);
  } 
};
```

Gaweph:
@limzykenneth Is there anything further needed for this PR?

src/core/transform.js
```diff
@@ -25,12 +25,37 @@ import p5 from './main';
...
+ * @chainable
+ */
+/**
+ * @method applyMatrix
+ * @param  {Number} a_2x3 numbers which define the 2×3 matrix to be multiplied
+ * @param  {Number} b_2x3 numbers which define the 2×3 matrix to be multiplied
+ * @param  {Number} c_2x3 numbers which define the 2×3 matrix to be multiplied
+ * @param  {Number} d_2x3 numbers which define the 2×3 matrix to be multiplied
+ * @param  {Number} e_2x3 numbers which define the 2×3 matrix to be multiplied
+ * @param  {Number} f_2x3 numbers which define the 2×3 matrix to be multiplied
```
limzykenneth, line 38: Why is the parameter renamed here from the original `a, b, c,...`?

Gaweph: The YuiDoc generator is having problems when the params of overloads have the same name.  Would it be better to be all together?

```
  * @param  {Number} a numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} b numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} c numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} d numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} e numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} f numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} [g] numbers which define the 4x4 matrix to be multiplied
  * @param  {Number} [h] numbers which define the 4x4 matrix to be multiplied
  * @param  {Number} [i] numbers which define the 4x4 matrix to be multiplied
  * @param  {Number} [j] numbers which define the 4x4 matrix to be multiplied
  * @param  {Number} [k] numbers which define the 4x4 matrix to be multiplied
  * @param  {Number} [l] numbers which define the 4x4 matrix to be multiplied
  * @param  {Number} [m] numbers which define the 4x4 matrix to be multiplied
  * @param  {Number} [n] numbers which define the 4x4 matrix to be multiplied
  * @param  {Number} [o] numbers which define the 4x4 matrix to be multiplied
  * @param  {Number} [p] numbers which define the 4x4 matrix to be multiplied
  ```

Comment (limzykenneth) [reply src/core/transform.js:38]: What are the problems it is having? As we have other functions with overloads using same parameter names elsewhere and they work fine for the most part (`loadJSON()` for example).

Comment (Gaweph) [reply src/core/transform.js:38]: @limzykenneth I shall investigate further sir :)

Comment (Gaweph) [reply src/core/transform.js:38]: Apparently it breaks when you try to give descriptions to the overloaded params, the `loadJSON()` example does not have descriptions for the params.

Shall I combine all variables into a single method with optional params?  Or should I create better variable names and 
have 3 distinct methods (I have committed a new file change that has nicer names) let me know what you think I should do :)

Comment (limzykenneth) [reply src/core/transform.js]: `loadJSON()` does have descriptions for each of the parameters. It probably clearer to use the reference page as a guide instead of the inline docs.

The descriptions can only be defined once for each parameter meaning the parameter with the same name cannot have two different descriptions (which make sense when we look at the layout of the reference page, all parameters are listed and described together). If the overload have new parameters that hasn't been described, they can have descriptions as well (see `tint()` for this case).

Comment (Gaweph) [reply src/core/transform.js]: > The descriptions can only be defined once for each parameter meaning the parameter with the same name cannot have two different descriptions (which make sense when we look at the layout of the reference page, all parameters are listed and described together). If the overload have new parameters that hasn't been described, they can have descriptions as well (see `tint()` for this case).

This is correct.  What shall we do in this case?  

Option 1) Overload method has different param names and thus different param descriptions (as with tint())
Option 2) Overload for 4x4 matrix has new descriptions for param g - p
Option 3) Remove param descriptions for both overloads that access 6 or 16 values and add to overall method description

Any advice on what is needed to get this PR in would be appreciated :)

Comment (limzykenneth) [reply src/core/transform.js]: In this case only a-f would present overlap of names with other overloads which means that these will need one description each that covers when they are used for 2x3 or 2x4 cases, ie `numbers which define the 2×3 or 4x4 matrix to be multiplied`. g-p are only used in 4x4 so they can have a 4x4 specific description ie `numbers which define the 4x4 matrix to be multiplied`. Finally array is its own case so can have its own description, ie `arr an array of numbers - should be 6 or 16 length (2*3 or 4*4 matrix values)`.

Which is to say for descriptions we are close here already, you just need to add the description back in instead of leaving them completely blank or changing their names.

**Comment by limzykenneth** [comment]
It would be great to have a graphic like the one currently used for the 2x3 matrix in the description for the 4x4 matrix as well.

**Comment by Gaweph** [comment]
> It would be great to have a graphic like the one currently used for the 2x3 matrix in the description for the 4x4 matrix as well.

Do you mean this example?

```js
 *   applyMatrix(  ct, 0.0,  st,  0.0,
 *                0.0, 1.0, 0.0,  0.0,
 *                -st, 0.0,  ct,  0.0,
 *                0.0, 0.0, 0.0,  1.0);
 ```
 
 Ironically this is an example of a 4x4 matrix (16 params)

**Comment by Gaweph** [comment]
@limzykenneth 4x4 image added.  I think this will work fine without any param descriptions, the images highlight what the properties are for :)

Let me know if you think we need anything further on this

**Comment by limzykenneth** [comment]
I would still prefer the text description to be present.

**Comment by Gaweph** [comment]
> I would still prefer the text description to be present.

@limzykenneth OK I think it now has the best of all options.  There is an example image for both 6 and 16 array sizes.  Each overload has clear description and param names aren't reused so there is no confusion regarding what neds to be passed into each method.

Let me know if you think anything further is needed

**Comment by Gaweph** [comment]
@limzykenneth Do you ned any further changes on this?  Once this PR goes in I can generate the new p5 definitelytyped d.ts definitions

File: src/core/transform.js [hunk 3]
```diff
@@ -24,13 +24,41 @@ import p5 from './main';
...
+ * @param  {Number} b7  numbers which define the 4x4 matrix to be multiplied
+ * @param  {Number} b8  numbers which define the 4x4 matrix to be multiplied
+ * @param  {Number} b9  numbers which define the 4x4 matrix to be multiplied
+ * @param  {Number} b10  numbers which define the 4x4 matrix to be multiplied
+ * @param  {Number} b11  numbers which define the 4x4 matrix to be multiplied
+ * @param  {Number} b12  numbers which define the 4x4 matrix to be multiplied
+ * @param  {Number} b13  numbers which define the 4x4 matrix to be multiplied
+ * @param  {Number} b14  numbers which define the 4x4 matrix to be multiplied
+ * @param  {Number} b15  numbers which define the 4x4 matrix to be multipliedo
+ * @param  {Number} b16  numbers which define the 4x4 matrix to be multiplied
```
limzykenneth, line 61: I'm afraid this is still not quite what I mean. The original parameter names for a 2x3 matrix should be retained and reused for a 4x4 matrix with descriptions that fit both 2x3 and 4x4 at the same time. Then in the signature definition for 4x4 the name that were present in 2x3 were repeated without description and the additional parameters will have unique 4x4 only description.

Something like this:
```
/**
  * @method applyMatrix
  * @param  {Number} a numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} b numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} c numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} d numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} e numbers which define the 2×3 or 4x4 matrix to be multiplied
  * @param  {Number} f numbers which define the 2×3 or 4x4 matrix to be multiplied
  */
/**
  * @method applyMatrix
  * @param  {Number} a
  * @param  {Number} b
  * @param  {Number} c
  * @param  {Number} d
  * @param  {Number} e
  * @param  {Number} f
  * @param {Number} g numbers which define the 4x4 matrix to be multiplied
  * @param {Number} h numbers which define the 4x4 matrix to be multiplied
  * @param {Number} i numbers which define the 4x4 matrix to be multiplied
  * @param {Number} j numbers which define the 4x4 matrix to be multiplied
  * @param {Number} k numbers which define the 4x4 matrix to be multiplied
  * @param {Number} l numbers which define the 4x4 matrix to be multiplied
  * @param {Number} m numbers which define the 4x4 matrix to be multiplied
  * @param {Number} n numbers which define the 4x4 matrix to be multiplied
  * @param {Number} o numbers which define the 4x4 matrix to be multiplied
  * @param {Number} p numbers which define the 4x4 matrix to be multiplied
```

Unless there is some other reasons this could not work?

Gaweph: Apologies, I mis understood :)  This has been done.  Let me know if you are now happy with the changes or if anything else is needed.

src/core/transform.js
```diff
@@ -24,13 +24,41 @@ import p5 from './main';
  * > <img style="max-width: 150px" src="assets/transformation-matrix.png"
  * alt="The transformation matrix used when applyMatrix is called"/>
  *
+ * <img style="max-width: 150px" src="assets/transformation-matrix-4-4.png"
```
limzykenneth, line 27: 150px width is a bit too small for this matrix. You can preview this using `grunt yui:dev`.

limzykenneth:
Almost there, just one more thing.

Gaweph:
> Almost there, just one more thing.

Increased size, you are right, was way too small :)

Gaweph:
> Almost there, just one more thing.

@limzykenneth Ping.  Are we happy with this now sir?

PR #5591, opened 2022-02-05
Added portuguese (brazilian) translation for FES.

absurd-web:
Relevant: #4233
### Changes:

I've added a portuguese (brazilian) translation for FES. I revised it and I am reasonably satisfied. Other speaker's suggestions would be appreciated.
<hr>

#### PR Checklist
- [x] `npm run lint` passes

### DIFF
--- a/translations/dev.js
+++ b/translations/dev.js
@@ -1,7 +1,7 @@
 export { default as en_translation } from './en/translation';
 export { default as es_translation } from './es/translation';
 export { default as ko_translation } from './ko/translation';
-
+export { default as pt_translation } from './pt/translation';
 /**
  * When adding a new language, add a new "export" statement above this.
  * For example, if we were to add fr ( French ), we would write:
@@ -13,4 +13,4 @@ export { default as ko_translation } from './ko/translation';
  *
  * "es_MX" is the language key whereas "translation" is the filename
  * ( translation.json ) or the namespace
-*/
+ */
--- a/translations/index.js
+++ b/translations/index.js
@@ -24,5 +24,6 @@ export default {
 export const languages = [
   'en',
   'es',
-  'ko'
+  'ko',
+  'pt'
 ];
--- a/translations/pt/translation.json
+++ b/translations/pt/translation.json
@@ -1,0 +1,72 @@
+{
+  "fes": {
+    "autoplay": "Uma mídia não foi autorizada a reproduzir (de '{{src}}'), provavelmente devido à política de reprodução automática do navegador.\n\n+ Para mais informações, visite: {{url}}",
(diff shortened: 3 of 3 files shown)


almchung:
Hi @absurd-web ! Welcome to p5.js, and thank you for taking the time to work on the Portuguese translation of Friendly Errors. I have some questions.

Would it be a good idea to specify the locale with a region designator (`pt_BR`)? What do you think?  Also, what would be a good [fallback language](https://www.i18next.com/principles/fallback) for `pt`/`pt_BR`?

And lastly, I also agree with you it would be great if we can get a second look from another Portuguese (Brazilian) speaker. Do you know a local group that can help you find a peer-reviewer? @outofambit and @Qianqianye do you know anyone who can help us on this? Or, we can deploy this version first and then let feedback/revisions happen naturally.

tiagohermano (approved):
Everything seems fine to me. Nice work 👍🏽

Qianqianye:
> Would it be a good idea to specify the locale with a region designator (`pt_BR`)? What do you think? Also, what would be a good [fallback language](https://www.i18next.com/principles/fallback) for `pt`/`pt_BR`?

Thanks for contributing @absurd-web. Wondering if you have might have thoughts for the question @almchung raised here?Thank you!

Qianqianye:
Hi @absurd-web, we'd like to follow up with this pull request, and we are wondering if you could respond to @almchung's previous comments. 

Thanks @tiagohermano for the translation review!

Meanwhile, I'm inviting this year's GSoC FES contributor @Ayush23Dash and mentors @nbriz and @almchung to this conversation, so we can decide how to move forward with this PR. Thanks.

Ayush23Dash:
I think having a region designator would definitely be helpful @almchung .
At the same time, it would be great if we could get feedback from Portuguese language contributors about having a [fallback language](https://www.i18next.com/principles/fallback) for `pt_BR`, if it is necessary or not; if it is not necessary in Portuguese Langugage's case then I believe we won't require the region designator as well(i.e. `pt_BR`)

absurd-web:
Sorry about this very late response.

@Qianqianye I am not really sure about the fallback. If British English falls back to American English (or general, international, english), then Brazilian would be the fallback to a hypothetical European Portuguese translation, as I am assuming the number of speakers is the metric used in this case. 
In any case I don't think an European Portuguese translation would look all that different, the same for the other lusophonic countries.
For these reasons I don't really think a designator is needed, @almchung .

Ayush23Dash:
Thanks @absurd-web \!!

In this case, I believe the current PR is good to go, without having a fallback language @almchung @Qianqianye

PR #5604, opened 2022-02-16
Issue radiobutton

### LINKED ISSUE #5603: The selected function for radio buttons created with createRadio does not work when making multiple changes to which option is selected.

- [ ] Color
- [ ] Data
- [X] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

### p5.js version

1.4.1

### Web browser and version

Chromium 98, Safari 15.3, FireFox 96

### Operating System

MacOS

### Steps:
1. Create a radio button with createRadio
2. Add multiple options
3. Cycle through those options repeatedly using the `selected(value)` function

Expected: you should be able to cycle through the options repeatedly

Actual: the radio buttons get stuck on the last option

### Snippet:

```js
const options = ["foo", "bar", "baz"];

function setup() {
  noCanvas();
  noLoop();

  let radio = createRadio();
  options.forEach((v) => radio.option(v));
  radio.selected("foo");

  let ix = 0;
  let nextButton = createButton("Next");
  nextButton.mouseClicked(() => {
    const opt = options[++ix % options.length];
    radio.selected(opt);
  });

  let fixButton = createButton("Fix");
  fixButton.mouseClicked(() => {
    // override radio.selected with a version that works properly
    let self = radio;
    // Here's the fix:
    self.selected = function(value) {
      let result = null;
      if (value === undefined) {
        for (const option of self._getOptionsArray()) {
          if (option.checked) {
            result = option;
            break;
          }
        }
      } else {
        // deselect all first (Google Chrome wigs out when multiple options have the checked attribute set to true)
        self._getOptionsArray().forEach(option => option.removeAttribute("checked"));
        for (const option of self._getOptionsArray()) {
          if (option.value === value) {
            option.setAttribute("checked", true);
            result = option;
            break;
          }
        }
      }
      return result;
    };
  });
}

function draw() {}
```

reejuBhattacharya:
Resolves #5603 

####Screenshots of the change:

The issue, as illustrated below, is whenever we cycled through the radio button selections in the code, the `checked=true` option is not removed after each subsequent change. Therefore, after one cycle, all options have `checked=true` and the program malfunctions.

before

The expected behavior: 

after
 

#### PR Checklist

- [x] `npm run lint` passes

Acknowledgement: Credit goes to @sflanker for solving the problem in the issue itself, I simply implemented it selectively in the source code.

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -943,7 +943,6 @@ p5.prototype.createRadio = function() {
   };
 
   self.selected = function(value) {
-    // console.log('Hello');
     let result = null;
     if (value === undefined) {
       for (const option of self._getOptionsArray()) {
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -943,6 +943,7 @@ p5.prototype.createRadio = function() {
   };
 
   self.selected = function(value) {
+    // console.log('Hello');
     let result = null;
     if (value === undefined) {
       for (const option of self._getOptionsArray()) {
@@ -952,6 +953,12 @@ p5.prototype.createRadio = function() {
         }
       }
     } else {
+      // forEach loop to uncheck all radio buttons before
+      // setting any one as checked.
+      self
+        ._getOptionsArray()
+        .forEach(option => option.removeAttribute('checked'));
+
       for (const option of self._getOptionsArray()) {
         if (option.value === value) {
           option.setAttribute('checked', true);
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -984,7 +984,10 @@ p5.prototype._onwheel = function(e) {
  * Note that not all browsers support this feature.
  * This enables you to create experiences that aren't limited by the mouse moving out of the screen
  * even if it is repeatedly moved into one direction.
- * For example, a first person perspective experience.
+ * For example, a first person perspective experience. It is recommended to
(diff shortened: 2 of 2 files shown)


sflanker:
It turns out that browser actually track the checked state for radio button inputs using a JavaScript property separately/in addition to the `checked="true"` attribute 🤦‍♂️. So even with the fix here things will be broken if the user interactively changes the selection. Update:

```
        // When clearing the state
        self._getOptionsArray().forEach(option => {
          option.checked = false;
          option.removeAttribute("checked");
        });
```

```
            // When setting the selection
            option.setAttribute("checked", true);
            option.checked = true;
```

reejuBhattacharya:
Oh ok...So, as I understand it, we will have to update the JavaScript using `option.checked` and simultaneously update the HTML using the `setAttribute` and `removeAttribute` functions.....

Ok, making the changes as you mentioned pronto. Thanks for the update!

limzykenneth:
From what I can see the issue seems to be fixed now with this PR. @sflanker Would you like to double check this before I merge?

sflanker:
The change looks good to me.

PR #5643, opened 2022-03-24
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/docs/preprocessor.js
+++ b/docs/preprocessor.js
@@ -10,7 +10,7 @@ function smokeTestMethods(data) {
 
       if (
         classitem.access !== 'private' &&
-        classitem.file.substr(0, 3) === 'src' &&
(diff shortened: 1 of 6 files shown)


montoyamoraga:
@all-contributors please add @CommanderRoot for code

limzykenneth:
Looks good. Thanks!

PR #5682, opened 2022-05-27
Fixes #5160

### LINKED ISSUE #5160: angleMode not saved/restored by push/pop

- [ ] Color
- [X ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [X ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:

My library sets angleMode to degrees, which can have side effects if user has set it to radians.
You might say, ok reset to radians after use, but what if user had set to degrees?
Two fixes:
Have angleMode return previous angleMode so i can save and restore
or
Save/restore angleMode in push/pop

KevinGrajeda:
Changes:

- added return when angleMode() is called without parameters Resolves #5160 
- added validation for angleMode() parameters
- added tests and documentation

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/trigonometry.js
+++ b/src/math/trigonometry.js
@@ -284,7 +284,6 @@ p5.prototype.radians = angle => angle * constants.DEG_TO_RAD;
  * Calling <a href="#/p5/angleMode">angleMode()</a> with no arguments returns current anglemode.
  * @method angleMode
  * @param {Constant} mode either RADIANS or DEGREES
- * @chainable
  * @example
  * <div>
  * <code>
@@ -318,7 +317,6 @@ p5.prototype.angleMode = function(mode) {
   } else if (mode === constants.DEGREES || mode === constants.RADIANS) {
     this._angleMode = mode;
   }
-  return this;
 };
 
 /**
--- a/src/math/trigonometry.js
+++ b/src/math/trigonometry.js
@@ -281,9 +281,10 @@ p5.prototype.radians = angle => angle * constants.DEG_TO_RAD;
 /**
  * Sets the current mode of p5 to the given mode. Default mode is RADIANS.
  *
+ * Calling <a href="#/p5/angleMode">angleMode()</a> with no arguments returns current anglemode.
  * @method angleMode
  * @param {Constant} mode either RADIANS or DEGREES
- *
+ * @chainable
  * @example
  * <div>
  * <code>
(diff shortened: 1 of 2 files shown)


KevinGrajeda:
@limzykenneth I would like to have a review on this pr, thank you👍

KevinGrajeda:
added a test for the validation of parameters in anglemode and now i dont have any failed tests

src/math/trigonometry.js
```diff
@@ -281,9 +281,10 @@ p5.prototype.radians = angle => angle * constants.DEG_TO_RAD;
 /**
  * Sets the current mode of p5 to the given mode. Default mode is RADIANS.
  *
+ * Calling <a href="#/p5/angleMode">angleMode()</a> with no arguments returns current anglemode.
  * @method angleMode
  * @param {Constant} mode either RADIANS or DEGREES
- *
+ * @chainable
```
limzykenneth, line 287: Why is this function made chainable? For most p5.js API there isn't much sense for functions to be chainable as they usually are not exposed as object instance methods.

limzykenneth:
I'm not sure how this PR addresses #5160 which has more do to with `push()` and `pop()` instead of parameter validation.

src/math/trigonometry.js
KevinGrajeda: I really don't know the use of chaining in this type of functions i just used other mode functions as a reference: 

https://github.com/processing/p5.js/blob/e32b45367baad694b1f4eeec0586b910bfcf0724/src/core/shape/attributes.js#L139-L141
https://github.com/processing/p5.js/blob/e32b45367baad694b1f4eeec0586b910bfcf0724/src/core/shape/attributes.js#L36-L38
https://github.com/processing/p5.js/blob/e32b45367baad694b1f4eeec0586b910bfcf0724/src/color/setting.js#L241-L246

KevinGrajeda:
In #5160 it was mentioned that adding return type to angleMode would be a solution
>Adding a return to `angleMode()` call without argument should be a doable solution.
https://github.com/processing/p5.js/issues/5160#issuecomment-822802333

if this doesn't solve that issue i think it's still useful having that return and the parameter validation was just missing.

Qianqianye:
Thanks @KevinGrajeda. Hey @limzykenneth, what do you think of the current status of this PR? Thank you!

src/math/trigonometry.js
limzykenneth: Can `@chainable` be removed here? Thanks.

KevinGrajeda: I think it's done

limzykenneth:
Looks good. Thanks!

Qianqianye:
Thanks @KevinGrajeda and @limzykenneth ✨

Qianqianye:
@all-contributors please add @KevinGrajeda for code

PR #5683, opened 2022-05-28
Fixed background transparency with Images

### LINKED ISSUE #5175: Background image with alpha not working as expected

yashlamba:
Co-authored-by: TwoTicks <adityasiddheshwar.adisid@gmail.com>

Resolves #5175
Continues #5225

@two-ticks suggested the `globalAlpha` implementation in #5225, this PR implements it.

## Changes:

[Example](https://editor.p5js.org/yashlamba/sketches/gFMiuAr45)
Replicated the example from #5225, to demonstrate consistency with the other solution.

- Setting alpha in background now works with images.
- Alpha parameter range is consistent with the docs (0-255), wanted to point this out since `globalAlpha` uses 0-1 range.

 Screenshots of the change:

#### PR Checklist
image
image

- [ ] `npm run lint` passes

### DIFF
--- a/src/core/p5.Renderer2D.js
+++ b/src/core/p5.Renderer2D.js
@@ -47,7 +47,14 @@ p5.Renderer2D.prototype.background = function(...args) {
   this.resetMatrix();
 
   if (args[0] instanceof p5.Image) {
-    this._pInst.image(args[0], 0, 0, this.width, this.height);
+    if (args[1] >= 0) {
+      // set transparency of background
+      const img = args[0];
+      this.drawingContext.globalAlpha = args[1] / 255;
+      this._pInst.image(img, 0, 0, this.width, this.height);
+    } else {
+      this._pInst.image(args[0], 0, 0, this.width, this.height);
+    }
   } else {
     const curFill = this._getFill();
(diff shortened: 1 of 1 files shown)


yashlamba:
@two-ticks, added you as a co-author, I hope you don't mind. In case you want me undo that, please let me know :)

limzykenneth (approved):

limzykenneth:
Looks good. Thanks!

PR #5693, opened 2022-06-15
Add anchor tag to web safe font

tlsaeger:
Add anchor tag to web safe font, so users don't have to scroll on the page and are directly taken to the list of available fonts.
I came across this minor issue and just fixed it, without opening an issue so no issue number. 

 Changes:

Minor change, to direct users to web safe fonts using an anchor tag. 

 Screenshots of the change:

#### PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/src/typography/loading_displaying.js
+++ b/src/typography/loading_displaying.js
@@ -280,7 +280,7 @@ p5.prototype.text = function(str, x, y, maxWidth, maxHeight) {
 /**
  * @method textFont
  * @param {Object|String} font a font loaded via <a href="#/p5/loadFont">loadFont()</a>,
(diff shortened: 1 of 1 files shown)


limzykenneth:
The original link does include the anchor fragment and works correctly on Firefox. It doesn't seem to work reliably in Chrome, would this be a Chrome specific problem that should be fixed on Chrome's end? Or is there some edge case that prevents redirected anchor fragment to work in Chrome?

tlsaeger:
Hi @limzykenneth yeah I tried it out in Chrome, but if it should work it might be a problem on my end. I will close the PR.

PR #5694, opened 2022-06-19
Update `saveFrames` documentation

### LINKED ISSUE #5685: SaveFrames() misses some frames

- [X] Accessibility (Web Accessibility)
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [X] Utilities
- [ ] WebGL

### p5.js version

v1.4.1 

### Web browser and version

Google Chrome 102.0.5005.61

### Operating System

Mac OS

### Steps:
1. visit https://p5js.org/reference/#/p5/saveFrames
2. open the console and you will find that the total frames obtained are 22, it should be  25 though, if fps is provided as 25 and duration is 1 sec then the total frames obtained should be 25.

### Snippet:

```js

function draw() {
 background(mouseX);
 }

 function mousePressed() {
 saveFrames('out', 'png', 1, 25, data => {
   print(data);
 });
 }

```

jesi-rgb:
## Issue
Resolves #5685, in which some discussion was made regarding the behaviour of the `saveFrames` function. @endurance21 and I were inspecting the code and thought it was not working properly, but @limzykenneth went ahead and clarified why the arguments are set up like that, but stated that there was no proper explanation for that within the docs. So that is what this change is aiming to solve! 

I tried to be explicit enough in my explanation for the reason to be clear, but not too technical so newcomers can properly understand the problem. Let me know if I it needs change for it to be either clearer and more explicit or less technical and more friendly!

## Changes

1. https://github.com/jesi-rgb/p5.js/blob/21f5e3db573cad9ef4a00821a5468b31549656a6/src/image/image.js#L72
	1. Fixes an issue with the linter (adding parentheses to the division)
2. https://github.com/jesi-rgb/p5.js/blob/21f5e3db573cad9ef4a00821a5468b31549656a6/src/image/image.js#L424
	3. Adds further explanation as to why the function works like this
4. https://github.com/jesi-rgb/p5.js/blob/21f5e3db573cad9ef4a00821a5468b31549656a6/src/image/image.js#L435
	5. Adds further explanation in the arguments description to make sure people know what to expect.

## Screenshots of the change:

The changes I added are highlighted.

image

image

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/image/image.js
+++ b/src/image/image.js
@@ -69,7 +69,7 @@ import omggif from 'omggif';
  * let img = createImage(66, 66);
  * img.loadPixels();
  * let d = pixelDensity();
- * let halfImage = 4 * (img.width * d) * (img.height / 2 * d);
+ * let halfImage = 4 * (img.width * d) * ((img.height / 2) * d);
  * for (let i = 0; i < halfImage; i += 4) {
  *   img.pixels[i] = red(pink);
  *   img.pixels[i + 1] = green(pink);
--- a/src/image/image.js
+++ b/src/image/image.js
@@ -69,7 +69,7 @@ import omggif from 'omggif';
  * let img = createImage(66, 66);
  * img.loadPixels();
  * let d = pixelDensity();
- * let halfImage = 4 * (img.width * d) * ((img.height / 2) * d);
+ * let halfImage = 4 * (img.width * d) * (img.height / 2 * d);
  * for (let i = 0; i < halfImage; i += 4) {
  *   img.pixels[i] = red(pink);
  *   img.pixels[i + 1] = green(pink);
--- a/src/image/image.js
+++ b/src/image/image.js
@@ -420,15 +420,20 @@ p5.prototype.saveGif = function(pImg, filename) {
  *  as an argument to the callback function as an array of objects, with the
  *  size of array equal to the total number of frames.
  *
- *  Note that <a href="#/p5.Image/saveFrames">saveFrames()</a> will only save the first 15 frames of an animation.
+ *  Note that <a href="#/p5.Image/saveFrames">saveFrames()</a> will only save the first 15 seconds of an animation.
+ *  The arguments `duration` and `framerate` are constrained to be less or equal 15 and 22, respectively, which means you
+ *  can only download a maximum of 15 seconds worth of frames at 22 frames per second, adding up to 330 frames.
+ *  This is done in order to avoid memory problems since a large enough canvas can fill up the memory in your computer
+ *  very easily and crash your program or even your browser.
+ * 
  *  To export longer animations, you might look into a library like
  *  <a href="https://github.com/spite/ccapture.js/">ccapture.js</a>.
  *
  *  @method saveFrames
  *  @param  {String}   filename
  *  @param  {String}   extension 'jpg' or 'png'
- *  @param  {Number}   duration  Duration in seconds to save the frames for.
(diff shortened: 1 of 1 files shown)


src/image/image.js
```diff
@@ -420,15 +420,20 @@ p5.prototype.saveGif = function(pImg, filename) {
  *  as an argument to the callback function as an array of objects, with the
  *  size of array equal to the total number of frames.
  *
- *  Note that <a href="#/p5.Image/saveFrames">saveFrames()</a> will only save the first 15 frames of an animation.
+ *  Note that <a href="#/p5.Image/saveFrames">saveFrames()</a> will only save the first 15 seconds of an animation.
```
limzykenneth, line 423: I don't think we need this line anymore with the explanations below.

limzykenneth (commented):
Just one comment inline

src/image/image.js
```diff
@@ -69,7 +69,7 @@ import omggif from 'omggif';
  * let img = createImage(66, 66);
  * img.loadPixels();
  * let d = pixelDensity();
- * let halfImage = 4 * (img.width * d) * (img.height / 2 * d);
+ * let halfImage = 4 * (img.width * d) * ((img.height / 2) * d);
```
limzykenneth, line 72: This is not passing linting and should be reverted

limzykenneth (changes requested):

src/image/image.js
jesi-rgb: that's interesting because on the other hand, the linter would not let me commit unless I made that change... could it be that i have another version of the linter installed? I think I am using the exact setup that the contributor_docs mentions, but clearly i missed something here

limzykenneth: Hmm...I double check that locally.

jesi-rgb: Made sure to setup and run everything again just in case I missed something but here is the output from the log:

```
[18:21:29] Running tasks for src/**/*.js [failed]
[18:21:29] → ✖ node --require @babel/register ./utils/sample-linter.js found some errors. Please fix them and try committing again.
/Users/jesi/repositorios/p5.js/src/image/image.js:72:43: Replace `img.height·/·2` with `(img.height·/·2)` [Error/prettier/prettier]
```

You can see it is asking for the change I made. That said, this only happens on commit, but this is not triggering when `npm run dev` is running and the console shows other errors. 

Are the linters in each system different?

limzykenneth: I've tested locally and the original version works while the one in the PR failed as the CI test did. Does running `npm test` clear for you? Do you have prettier separately installed on your system somewhere that is of a different version than what is used here?

jesi-rgb: sorry for the delay! did not get notified.

indeed, if I run npm test with my change, it prompts an error removing exactly that.

I changed it back to what it should be and now everything in `npm test` runs fine.

jesi-rgb:
Seems like this is already working fine! It let me commit with the correct configuration!

src/image/image.js
limzykenneth: @jesi-rgb Still have this pending then we can merge.

jesi-rgb: Got it and also a minor typo!

limzykenneth:
Looks good. Thanks!

PR #5704, opened 2022-07-01
Add an isNegative function

thecodingcookery:
I don't know if this is in the correct place, please tell me! But I've added a `isNegative()` function, though.
I don't have an example for this function, but I've tested it out, and it works!

### DIFF
--- a/src/math/math.js
+++ b/src/math/math.js
@@ -46,4 +46,9 @@ p5.prototype.createVector = function(x, y, z) {
   }
 };
 
+p5.prototype._isNegative = function(value) {
+  let string = value.toString();
+  return string[0] == '-';
+}
+
 export default p5;


msed21:
There seems to be an unnecessary underscore in the front of the functions name. Additionally, an easier and faster way to check if a number is negative is to simply check if it less than 0.

limzykenneth:
Hi @thecodingcookery, we usually require an issue to be made and discussed before a PR can be open, especially for new features which will require an argument to be made for how the new feature in question increase access to p5.js.

In this particular case, this function is too specific and I don't think it fit very well with the rest of the library. If there is a similar groupings of functions like these, they will fit better as an addon library instead of being in the core p5.js library.

For the above reasons, I'm going to close this PR for now. Feel free to open a new issue if you wish to discuss further. Thanks!

PR #5709, opened 2022-07-03
Implement `saveGif` as a native p5 function

### LINKED ISSUE #5118: Feature for saving gif files
#### New feature details:

I am getting this request quite a lot from my students. They want to be able to share their work on social media in animated form. I know that [`saveFrames`](https://p5js.org/reference/#/p5/saveFrames) exists but students have to already be relatively advanced to use that and it doesn't encode to gif.

Other options like using ccapture.js, require a number of advanced skills that students often don't have during their first experiences with coding.

Since GIF support was enabled we already have most of the work in place for this feature. [saveGif](https://github.com/processing/p5.js/blob/main/src/image/image.js#L187) already exists as a private method for saving animated p5.Image objects as GIFs. To enable this feature, we would have to add a separate public method (probably `saveGif` and the private one would be renamed) that stores frames over a certain number of frames and then turns them into an animated p5.Image object. This object could then be sent to the private method which could handle all of the encoding.

It would enable students to share their work in animated form without needing advanced skills. 

- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [x] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

jesi-rgb:
#### 🚧 This branch will be the work in progress for the implementation of the public `saveGif` function for the **'22 GSoC** edition. 🚧

# Save Gif

resolves #5118 

This PR **resolves issue #5118**, in which a thorough discussion is had about whether p5.js should include a native GIF encoding solution or not. 

## How it works
In the previously mentioned issue some proposals were discussed about how the interface for this functionality should be. The easiest I found to both implement and use was the following. The `saveGif` function admits at most 3 arguments for now:

```javascript
saveGif(filename, duration, delay)
```

- `filename` for the name of the file downloaded to your computer
- `duration` duration **in seconds** that you wish to download from your animation
- `delay` delay **in seconds** that you wish to wait before recording the animation. The animation will nevertheless have a duration of what you specified on the `duration` argument.

As an example, I think it is best to use it within the mouse/keyboard events:
```javascript
 function mousePressed() {
   if (mouseButton === RIGHT) {
     saveGif('mySketch', 3, 1);
   }
 }
```

This function will take all the current frames of the present animation, save them in a buffer and then proceed to generate a palette, apply this palette to every frame, optimize for transparency and encode it in a GIF blob, that will then force to download.

Functionality can be extended in lots of places but I believe that the current state works very conveniently by adding just 3 extra lines of familiar code. I plan on keep contributing to this functionality even after the GSoC to make it as convenient as possible.

 ## Changes

1. Most important change is in [src/image/loading_displaying.js](https://github.com/jesi-rgb/p5.js/blob/ae03db0c62863387933f5dfd88ed3cb7dfbde411/src/image/loading_displaying.js), which is where the main `saveGif` function lives. Alongside this function, the functions `_generateGlobalPalette` and `_pixelEquals` were also added.
2. Second important change is in [src/image/image.js](https://github.com/jesi-rgb/p5.js/blob/73acfb183744179ea03a5f8f8e6bd1778205b6d7/src/image/image.js), where the old `saveGif` function was renamed to a more obscure and explicit `encodeAndDownloadGif`.
	1. Alongside this change, the file [src/image/p5.Image.js](https://github.com/jesi-rgb/p5.js/blob/73acfb183744179ea03a5f8f8e6bd1778205b6d7/src/image/p5.Image.js) was also changed to consider this new function name.
	2. Same goes with the test [test/unit/image/downloading.js](https://github.com/jesi-rgb/p5.js/blob/629d7448d4dc0cf562bc9214ab6cc76d04f5a680/test/unit/image/downloading.js), which was updated to evaluate the `encodeAndDownloadGif` function.

### ⚠ Some other changes
Some other changes were made for the sake of either speed or convenience. These changes are completely opinionated and may include breaking changes that I may not be aware of, so please consider them carefully. 

1. The `package.json` file is modified to include the package [gifenc](https://www.npmjs.com/package/gifenc), created by [Matt DesLauriers](https://twitter.com/mattdesl). He personally reached out to me regarding this topic and it turns out that the package includes two key functions that make the GIF rendering not only much easier, but also much faster.
	1. Naturally, the `package-lock.json` file also changes.
2. The `Gruntfile.js` is also changed. In order to include Matt's package, we needed to first compile the javascript code to babel code, before _uglifying_ it. The uglify module wouldn't process the ES6 notation for some reason, although I've seen in some places that the project is already accounting for that. That is the `babel` module that appears in lines like `grunt.loadNpmTasks('grunt-babel');`
	1. Also, the `ecmaVersion` wouldn't pass the `npm run build` nor `npm run lint` nor any other command of this kind altogether. So I bumped up the version to 8 and it seems to be working fine.

Again, all these changes are considered from a rather ignorant and novel standpoint, so maybe I am making very questionable decisions here. I ask the community to pay special attention here, though it does not need to be said that I will try my best for this to be as close to perfect as it can be.
	
	

 ## Screenshots of the change:

These are some screenshots of the tool in use. We wrote the function keyPressed and we are tracking the 's' key. When pressed, a gif will be rendered and downloaded. A beautiful sketch by @TomasMiskov is presented.

SCR-20220808-ixh

During the saving and rendering, a little paragraph appears indicating the progress. I feel like this is very important since, even though it usually only takes a couple of seconds, a couple of seconds in web time is an almost infinite amount of time.

SCR-20220808-ixk

The resulting gif is presented here, just shy over 1MB in size: 
mySketch (36)

## PR Checklist

- [x] `npm run lint` passes


jesi-rgb:
@limzykenneth @Qianqianye don't know exactly who I have to ping for this but this is review-ready! let me know if I have to proceed further before the review or do something else. hope you have a great day :)

src/image/loading_displaying.js
```diff
@@ -158,6 +159,341 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+    frames.push(data);
+    count++;
+
+    p.html(
+      'Saved frame <b>' +
+        frames.length.toString() +
+        '</b> out of ' +
+        nFrames.toString()
+    );
+    await new Promise(resolve => setTimeout(resolve, 0));
```
davepagurek, line 310: This is a clever way to keep the UI updating, nice!

src/image/loading_displaying.js
```diff
@@ -158,6 +159,341 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+      let lastPixel = [
+        lastFramePixels[p],
+        lastFramePixels[p + 1],
+        lastFramePixels[p + 2],
+        lastFramePixels[p + 3]
+      ];
+
+      // if the pixels are equal, save this index to be used later
+      if (_pixelEquals(currPixel, lastPixel)) {
+        matchingPixelsInFrames.push(parseInt(p / 4));
```
davepagurek, line 363: Since `p` is already a number, I think you can use `Math.floor()` instead of `parseInt()`. Although since `p` is iterating by 4, I think maybe we don't need to do anything extra to ensure it's an integer here?

src/image/loading_displaying.js
```diff
@@ -158,6 +159,341 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+        gl.drawingBufferWidth,
+        gl.drawingBufferHeight,
+        gl.RGBA,
+        gl.UNSIGNED_BYTE,
+        pixels
+      );
+
+      data = _flipPixels(pixels);
+    } else {
+      data = this.drawingContext.getImageData(0, 0, this.width, this.height)
```
davepagurek, line 297: I think there's a slight performance cost to getting and storing the pixels array each frame. This is fine for sketches that use `frameCount`, but for ones that use `millis()` (so they drop frames to keep timing correct rather than letting timing slow down to not drop frames), do you think it's worth changing how `millis()` works while saving a gif? We could use fake time increments so the sketch thinks only `1000/targetFrameRate` milliseconds pass each frame to ensure the resulting gif doesn't look sped up because it took slightly longer than usual to render each frame. (Maybe this could be something for a follow-up PR since on its own this feature is super useful)

davepagurek (commented):
Great work on this! I tested it out a bit, it's easy to use and the resulting gifs are great quality 🙂

src/image/loading_displaying.js
jesi-rgb: I had no idea `millis()` worked like that! As for your question, I am not sure what to do about it.

Another related thing is that the delay functionality implemented will only work with sketches that use `frameCount`, as you said. I had a sketch from a friend that just updated the speed from some particles by adding up a value each frame and, since the simulation was not depending on the `frameCount` value, it just does nothing. 

Although it doesn't break, it does not behave as expected but have no idea how to solve it. :/

jesi-rgb:
Thanks so much for the review, @davepagurek!

src/image/p5.Image.js
```diff
@@ -890,7 +890,7 @@ p5.Image.prototype.isModified = function() {
  */
 p5.Image.prototype.save = function(filename, extension) {
   if (this.gifProperties) {
-    p5.prototype.saveGif(this, filename);
+    p5.prototype.encodeAndDownloadGif(this, filename);
```
endurance21, line 893: 🚧   this might be a serious breaking point, we have to take care of it    🚧

test/unit/image/downloading.js
```diff
@@ -320,3 +320,33 @@ suite('p5.prototype.saveFrames', function() {
...
+    assert.typeOf(myp5.saveGif, 'function');
+  });
+  test('should not throw an error', function() {
+    myp5.saveGif('myGif', 3, 2);
+  });
+  testWithDownload('should download a GIF', async function(blobContainer) {
+    myp5.saveGif(myGif, 3, 2);
+    await waitForBlob(blobContainer);
+    let gifBlob = blobContainer.blob;
+    assert.strictEqual(gifBlob.type, 'image/gif');
```
endurance21, line 350: can we add more checks of downloaded gif parameters?
something like this pseudoCode 
```js
let gif= saveGif(filename, seconds, delay) ;

assert(gif.properties.duration === seconds )

```

endurance21 (changes requested):
@jesi-rgb  
A huge blast of work and looks so impressive   🥇 .

Its a huge  work to review so i have chosen to perform this task in chunks :), 
What i have checked now is design pattern of your api, general coding practise and  some dependecies related code,i have left some comments, lets resolve those comments and then we will move further completing all chunks of review.

src/image/loading_displaying.js
```diff
@@ -158,6 +159,341 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+ *   // this will download the first two seconds of my animation!
+ *   saveGif('mySketch', 2);
+ * }
+ * </code>
+ * </div>
+ *
+ * @alt
+ * animation of a circle moving smoothly diagonally
+ */
+p5.prototype.saveGif = async function(fileName, seconds = 3, delay = 0) {
```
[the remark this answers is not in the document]
davepagurek: We probably shouldn't force it to be an integer in case, for example, we have a perfect loop in 90 frames, or 1.5s. (Although I think forcing it to be an integer would make sense if we use frames as a unit instead of seconds, which could also be an option)

Gruntfile.js
```diff
@@ -523,10 +533,14 @@ module.exports = grunt => {
   grunt.loadNpmTasks('grunt-contrib-clean');
   grunt.loadNpmTasks('grunt-simple-nyc');
 
+  //this library converts the ES6 JS to ES5 so it can be properly minified
+  grunt.loadNpmTasks('grunt-babel');
```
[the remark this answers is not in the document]
jesi-rgb: So sorry! Missed that one. I'll check if babelify works and remove babel!

src/image/loading_displaying.js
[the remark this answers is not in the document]
jesi-rgb: Using frames as units is incredibly useful for advanced users and not intuitive for newcomers. Having the ability to switch between the two seems ideal. How do we do that? Using another argument like `frameUnits = true`?

src/image/loading_displaying.js
```diff
@@ -158,6 +159,341 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+  let gifFrameDelay = 1 / _frameRate * 1000;
+
+  // constrain it to be always greater than 20,
+  // otherwise it won't work in some browsers and systems
+  // reference: https://stackoverflow.com/questions/64473278/gif-frame-duration-seems-slower-than-expected
+  gifFrameDelay = gifFrameDelay < 20 ? 20 : gifFrameDelay;
+
+  // because the input was in seconds, we now calculate
+  // how many frames those seconds translate to
+  let nFrames = Math.ceil(seconds * _frameRate);
```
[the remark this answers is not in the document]
jesi-rgb: Just because I thought it was a bit more convenient to have some extra frames (because ceil rounded to higher integer). If you have extra unnecessary frames, you can drop them down easily, but if you missed the last frames, you need to re-render the whole thing again.

Does it make sense? Or is it better to floor it?

src/image/loading_displaying.js
```diff
@@ -158,6 +159,341 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+  let nFramesDelay = Math.ceil(delay * _frameRate);
+
+  //   initialize variables for the frames processing
+  var count = nFramesDelay;
```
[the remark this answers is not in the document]
jesi-rgb: Being honest, that was at a very early stage and didn't know much of the difference between let and var, haha. Will fix it immediately!

src/image/loading_displaying.js
```diff
@@ -158,6 +159,341 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+  // we start on the frame set by the delay argument
+  frameCount = nFramesDelay;
+
+  const lastPixelDensity = this._pixelDensity;
+  this.pixelDensity(1);
+
+  // We first take every frame that we are going to use for the animation
+  let frames = [];
+
+  if (document.getElementById('progressBar') !== null)
```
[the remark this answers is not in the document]
jesi-rgb: Sounds good! But how would the `progressCallback` function work? Return the number of frames that we already processed and let the user do whatever they want with that info?

src/image/p5.Image.js
jesi-rgb: I made sure to substitute the old `saveGif` function with the encodeAndDownloadGif, which is the original function that was already present in the repo. Even the tests were considered.

It indeed is risky though! I'll keep close attention to that!

test/unit/image/downloading.js
jesi-rgb: Sure! I just couldn't think of more tests at the time, but I'll add as much as I can.

src/image/loading_displaying.js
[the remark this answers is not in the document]
davepagurek: Maybe adding an optional object with extra options at the end? e.g.:
```js
function(fileName, time, { delay = 0, units = 'seconds' }) {
  const frames = units === 'seconds' ? time * 60 : time
  // etc
}

// Usage:
saveGif('test.gif', 30, { units: 'frames' })
```

Might make it easier to read at call sites with the key: value syntax rather than having `true` as a parameter without extra context around it. P5 does something similar when there are advanced options, e.g. in `textToPoints`: https://p5js.org/reference/#/p5.Font/textToPoints

src/image/loading_displaying.js
davepagurek: Oh interesting, do you have a link to that sketch or an example that works similarly?

Also for `millis`, I think a solution would be, when we start recording, to replace the implementation of `millis` with a fake implementation. something like:
```js
// Replace with a fake implementation
const oldMillis = p5.prototype.millis
let time = millis()
p5.prototype.millis = () => time

while (count < nFrames + nFramesDelay) {
  time += 1000 / this._targetFrameRate // Fake time increment
  // render the frame here
}

// Put the old function back
p5.prototype.millis = oldMillis
```

src/image/loading_displaying.js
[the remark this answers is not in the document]
jesi-rgb: Sounds ideal!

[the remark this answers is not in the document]
endurance21: yes definetly makes sense. @jesi-rgb we can do that !

src/image/loading_displaying.js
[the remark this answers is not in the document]
endurance21: given the main focus is on the gif and its appearence, i think an extra frame wont be producing much glitch.
we can go with it.

src/image/loading_displaying.js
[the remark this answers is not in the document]
endurance21: no problem.
always on for better !

src/image/loading_displaying.js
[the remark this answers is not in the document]
endurance21: yes, we shall do that.

but it has a downside as well , suppose if someone passes a callback function which is taking too much to finish, that would slow down the our processing too.

src/image/p5.Image.js
endurance21: can you please make a gist of what this function is doing  and  how your replacement does not affects  this functionality in terms of programming.

also did you run every examples using this function, perhaps this would be present in examples folder!

test/unit/image/downloading.js
endurance21: sure, keep up the good work , eh!

Qianqianye:
Great work @jesi-rgb! Thanks @davepagurek for the review. 
As @jesi-rgb 's mentor @endurance21, please feel free to merge the PR when it's ready. Thank you!

package.json
```diff
@@ -67,6 +67,7 @@
     "grunt-mocha-test": "^0.13.3",
     "grunt-newer": "^1.1.0",
     "grunt-simple-nyc": "^3.0.1",
+    "grunt-babel": "^8.0.0",
```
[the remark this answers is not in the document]
jesi-rgb: # Main problem
To understand the main problem, let me explicitly state it here. The current build pipeline takes all the code and runs `browserify` on it, which makes a single file, and that is passed on to `uglify`, which makes it very small, browser ready.


`Uglify` does not understand ES6 code, so it needs to be transpiled before it gets here. This is `browserify's` (and more specifically,` babelify's`) job. `Browserify` can be coupled with babelify to transpile between javascript versions before compiling and bundling. 

# Babelify and options
Inside the `tasks/build/browserify.js` file we have, among other things, this important variable:

```javascript
const babelifyOpts = {
		// add some options
        plugins: ['static-fs']
      };
```

which lets us modify the behaviour of `babelify`. As of right now, it only has a plugin called `static-fs`, which converts all asynchronous file system calls to their "hardcoded" version, statically inlining these calls.

We can pass in more options to change the behaviour:

```javascript
const babelifyOpts = {
		// add some options
        plugins: ['static-fs'],
        presets: ['@babel/preset-env'],
        global: true,
      };
```

This uses the `present-env` preset, recommended for our use case and the `global` option tells `babelify` to also process and transpile code coming from `node_modules`. The default behavior does not do that.

# No solution found

After playing with every combination, some problem arises.

- If we don't include `global`, we cannot have `node_modules` transpiling so future dependencies from other contributors that also work similarly to `gifenc` won't work during build time.
- If we don't include the `static-fs`, `mochaChrome` breaks:
```
Running "mochaChrome:yui" (mochaChrome) task
p5.js translator called before translations were loaded
p5.js translator called before translations were loaded
[stuck here forever]
```
- Including `static-fs` AND `global` breaks the `static-fs` since it cannot compile other libraries (like `opentype`) with their file calls.
```
Fatal error: /Users/jesi/repositorios/p5.js/node_modules/opentype.js/dist/opentype.js: Not able to statically evaluate the expression(s) for babel-plugin-static-fs.
Try changing your source code to something that can be evaluated at build-time, e.g.
    const src = fs.readFileSync(__dirname + '/foo.txt', 'utf8');
 while parsing file: /Users/jesi/repositorios/p5.js/node_modules/opentype.js/dist/opentype.js
```

# My initial solution
Initially, I simply included the `grunt-babel` option since the main problem was I could not get `Uglify` to work. This worked fine immediately, albeit not knowing why and not knowing that `babelify` was already present.

For what it's worth, the `grunt-babel` has not given absolutely any problem, since it is set after the `browserify` task which means it gets to transpile the already bundled app. I believe this can open a door for future contributors that won't have to deal with this topic but just develop a better experience for p5.js.

If this is not enough to justify the `grunt-babel` inclusion, my only guess is that I should not include `gifenc` at all and figure out another way of using that code. Totally open for discussion!

src/image/p5.Image.js
jesi-rgb: `encodeAndDownloadGif` is exactly the same function that was before, just that it was called `saveGif`. They call it as _private_, in the sense that the normal users won't know about the function. That's why it does not appear in the documentation

Gruntfile.js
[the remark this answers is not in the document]
endurance21: taking jesus solution as of now and moving ahead...

src/image/loading_displaying.js
```diff
@@ -158,6 +159,363 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+ * }
+ * </code>
+ * </div>
+ *
+ * @alt
+ * animation of a circle moving smoothly diagonally
+ */
+p5.prototype.saveGif = async function(
+  fileName,
+  duration,
```
endurance21, line 222: since the second argument is signifying both time and number of frame, is it okay to name it as duration? 

how about, `count` or `number` ?

src/image/loading_displaying.js
```diff
@@ -158,6 +159,363 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+  // we start on the frame set by the delay argument
+  //   frameCount = nFramesDelay;
+
+  const lastPixelDensity = this._pixelDensity;
+  this.pixelDensity(1);
+
+  // We first take every frame that we are going to use for the animation
+  let frames = [];
+
+  if (document.getElementById('progressBar') !== null)
```
endurance21, line 264: progressBar sounds so common name, lets name it something that would be unique that perhaps user wont be using .


something like - p5.gif.progressBar

src/image/loading_displaying.js
```diff
@@ -158,6 +159,363 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+    // if we have a WEBGL context, initialize the pixels array
+    // and the gl context to use them inside the loop
+    gl = document.getElementById('defaultCanvas0').getContext('webgl');
+    pixels = new Uint8Array(gl.drawingBufferWidth * gl.drawingBufferHeight * 4);
+  }
+
+  // stop the loop since we are going to manually redraw
+  this.noLoop();
+
+  while (count < nFrames + nFramesDelay) {
```
endurance21, line 289: 1. count variable is not intuitive ( what type of count ?) 
2. nframes+ nFrameDelay is making this less perciveable,cannot we make another variable and name it significantly like let xyz = a +b , whree xyz would be some intuitive name rather than writing a+b.

src/image/loading_displaying.js
```diff
@@ -158,6 +159,363 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+  let gl;
+  if (this.drawingContext instanceof WebGLRenderingContext) {
+    // if we have a WEBGL context, initialize the pixels array
+    // and the gl context to use them inside the loop
+    gl = document.getElementById('defaultCanvas0').getContext('webgl');
+    pixels = new Uint8Array(gl.drawingBufferWidth * gl.drawingBufferHeight * 4);
+  }
+
+  // stop the loop since we are going to manually redraw
+  this.noLoop();
```
endurance21, line 287: cannot we let the loop run as it is and meanwhile we extract the pixel colors ?
using this maybe - https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Pixel_manipulation_with_canvas

src/image/loading_displaying.js
```diff
@@ -158,6 +159,363 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+
+  while (count < nFrames + nFramesDelay) {
+    /*
+      we draw the next frame. this is important, since
+      busy sketches or low end devices might take longer
+      to render some frames. So we just wait for the frame
+      to be drawn and immediately save it to a buffer and continue
+      */
+    this.redraw();
```
endurance21, line 296: same as above

endurance21 (changes requested):
@jesi-rgb  good follow up buddy!
left some comments, for next iteration.

src/image/loading_displaying.js
jesi-rgb: `Count` or `number` are, in my honest opinion, more confusing. In the end, the animation will have a duration of either 30 seconds or 30 frames, which, I think, is understood correctly. Also, the default behaviour is seconds, so those who do not mess with the options will be comfortable. 

What do you think?

src/image/loading_displaying.js
jesi-rgb: Sounds great, will fix it ASAP!

src/image/loading_displaying.js
jesi-rgb: We are doing exactly what you say! We are enabling the loop back just before extracting the pixel colors and start encoding.

What we cannot do is loop while recording the frames, because we need all the frames in order, one by one, and we extract the frame information from the current Canvas context.

davepagurek: I think it would be possible to use the existing `loop()`, but I believe that would require augmenting `draw()` to keep track of the export frame count, and to record a frame at the end of each export frame. That said, I think I personally prefer this method where we manually call `redraw()`, as it keeps `draw()` simple and also it means you can read the gif saving function more easily without having to jump around the codebase to see the data flow.

jesi-rgb:
@endurance21 Well! The example errors we talked about are fixed! To be honest, there was nothing to fix on the first place. 

Either grunt or firefox or something was caching the old code and preveting the new code to be used in the example. I just removed every p5.js file under /lib and re-built again.

Unless something else can be added, I think we are ready!

endurance21 (changes requested):

src/image/loading_displaying.js
```diff
@@ -158,6 +159,387 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
...
+
+  // check the mode we are in and how many frames
+  // that duration translates to
+  const nFrames = units === 'seconds' ? duration * _frameRate : duration;
+  const nFramesDelay = units === 'seconds' ? delay * _frameRate : delay;
+  const totalNumberOfFrames = nFrames + nFramesDelay;
+
+  // initialize variables for the frames processing
+  let frameIterator = nFramesDelay;
+  frameCount = frameIterator;
```
endurance21, line 279: I am wondering if it is good to set the system variable from here? and also could you please elaborate why it is necessary ?

jesi-rgb: `frameCount` is a variable exposed by the p5.js library upon instancing. Accesing `frameCount` returns the index of the frame we are currently in, and we can overwrite it in order to _travel_ to a given frame.

This is part of the delay feature. We calculate how many frames we should wait for before recording. Imagine setting `delay = 3`, which makes it `3 * 60 = 180` (assuming a frameRate of 60). We have to wait until frame 180 to start recording. Cool, simply set `frameCount = 180` and then keep on as usual. 

This also makes sure that the user, if `delay = 0`, can record their sketch from the very beginning without extra hassle, since this would effectively set `frameCount = 0`, even if the animation has already started.

davepagurek: Btw, could this be related to the sketch you mentioned earlier that doesn't work correctly when recording? If a sketch depends partially on `frameCount` but also has some updates that don't (e.g. a physics simulation that updates each `draw()`), I wonder if rewinding `frameCount` might put the sketch into an unexpected state. Maybe there should be a way to keep counting from the current state? e.g. passing `undefined` instead of a number?

jesi-rgb: Yes, it probably will do that. But for sketches not based on that, I cannot think of a way to somehow attach their time component to something. 

I don't know how other systems approach this, but I thought at least we could have support for those who use `frameCount` as a way of dealing with time, since it is a common practice in shader languages. I myself find using it a lot, and also learned about it in some Coding Train video, which may increase the chance of people using it.

This does not solve the problem at all, of course, but at least we are considering a subset of the users for this matter.

I don't know if implementing a robust solution for every sketch is very difficult, but being already familiar with the codebase, I can work on improving it over the next year independently of GSoC or other events, if I can find a way to do it!

davepagurek: For sure! I think just supporting frameCount for now is totally fine, as long as we leave space in the API to support other things in the future without breaking older code. So the future uses to consider are:
- sketches that rely on state updates in `draw()`: I did some tests, it seems like doing state updates in `draw()` works fine for gif recording! so this is handled already
- sketches that rely on `millis()`: that can be done later if we provide a mock implementation of `millis`, which won't affect any of the code you've already written, so everything's good here for this PR!
- sketches that mix `frameCount` with state updates (or `millis()`): the only issue is that rewinding will bring `frameCount` out of sync with the others. If we don't rewind, then they're in sync. We don't have to support no rewinding in this PR, e.g. in the future we could support it by supplying `delay: undefined` and removing the check to ensure it's a number. If that looks ok, then everything here is good without changes 🙂  I just want to make sure we're aware of it in case there's a different API we want for it in the future

davepagurek:
Also, I notice that gifs turn out all black if I call `saveGif` inside `setup()`. I'm not sure exactly why this is happening yet, since [this is all that happens between setup() and draw()](https://github.com/processing/p5.js/blob/7d193c34665b8771596f01c8774e1e4c2d7f064b/src/core/main.js#L346-L359). I suspect it might have something to do with the first `draw` is getting called after `setup` finishes regardless of `noLoop()` being called, so there could be some double-drawing? e.g. adding this to the top of `saveGif` fixes it:

```js
if (this._setupDone) {
  await new Promise((res) => window.requestAnimationFrame(res));
}
```

That's a bit of a hack though, so in any case, I think it's also fine to add something to the documentation saying it has to be called outside of `setup` for now.

jesi-rgb:
@davepagurek yes, I knew that calling within setup causes problems and wanted to make it possible for those that may unintentionally put it there to just work. 

Unfortunately, the snippet of code you shared didn't work for me! I'll just state that in the documentation for now, while we manage some ways to make this work :)

davepagurek:
> Unfortunately, the snippet of code you shared didn't work for me! I'll just state that in the documentation for now, while we manage some ways to make this work :)

Sounds good! Also, that's interesting that it doesn't work, I guess there's more to it than I thought 🤔 Definitely out of scope for this PR then! If you feel like sending the sketch it doesn't work for, I can poke around in it and see if I can debug some more!

endurance21 (approved):

endurance21:
@jesi-rgb  great work!
I have reviewed the PR and tested it locally,  and happy to say that the animation to gif feature is working just fine and is in the stage to be made available to the public.

Disclaimer : 
This PR contains the one basic discussion regarding the "babel parsing the es6-es5 code and the role of uglify ", which is still a matter of discussion within the group, however for now the solution provided by this pr seems reasonable to that discussion and if a better solution is discovered in near future, it shall definitely be incorporated.

PR #5735, opened 2022-07-24
Noting that the p5.js number generator is only sudo-random

thecodingcookery:
In the example of the random function I noted that the random number generator is only sudo-random.

### DIFF
--- a/src/math/random.js
+++ b/src/math/random.js
@@ -110,7 +110,7 @@ p5.prototype.randomSeed = function(seed) {
  * let words = ['apple', 'bear', 'cat', 'dog'];
  * let word = random(words); // select random word
  * text(word, 10, 50); // draw the word
- * describe(`word displayed at random. Either apple, bear, cat, or dog.`);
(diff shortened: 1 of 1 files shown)


Qianqianye:
Thank you! As a reminder, an issue needs to be opened before a pull request is opened, then you can tag the issue in the pull request. You can read more about it in the [GitHub Issue Flow](https://github.com/processing/p5.js/tree/main/contributor_docs#github-issue-flow). This is necessary for tracking development and keeping discussion clear. Thanks!

thecodingcookery:
Sorry, I thought that something as small as this would not need a github issue. I will try to keep this in mind!

Qianqianye:
Please go ahead and open an issue @thecodingcookery. Thank you!

limzykenneth:
I think in this case it is not very important to note that this is pseudorandomness as to humans it will be indistinguishable from true randomness and it's not being used for cryptographic purposes.

Also to nitpick a bit 😝, the spelling should be `psuedo` instead of `sudo`.

Qianqianye:
I will close this PR for now for organizational purpose, and please feel free to open a new PR after an issue is opened. Thank you!

PR #5737, opened 2022-07-28
Permit zero-temporary-value buffer binding

rbtcollins:
When using custom 3D renderers, p5 can support much larger models, but this particular bit of code showed up as a hotspot: I was passing in a Float32 already, and it was getting cloned.

Without this change, my particular workload runs at 7-12 fps. With it 45-50fps.

In principle p5 can make use of this itself, though having webGL ready representations as the source of truth is a much larger change - but even so, taking steps in that direction seems broadly useful.

 Changes:

 Screenshots of the change:

#### PR Checklist

- [ ] `npm run lint` passes

Fixes: https://github.com/processing/p5.js/issues/5739

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -1337,7 +1337,12 @@ p5.RendererGL.prototype._bindBuffer = function(
   if (!target) target = this.GL.ARRAY_BUFFER;
   this.GL.bindBuffer(target, buffer);
   if (values !== undefined) {
-    const data = new (type || Float32Array)(values);
+    let data;
+    if (ArrayBuffer.isView(values)) {
+      data = values;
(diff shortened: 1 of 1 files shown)


Qianqianye:
Thank you @rbtcollins ! As a reminder, an issue needs to be opened before a pull request is opened, then you can tag the issue in the pull request. You can read more about it in the [GitHub Issue Flow](https://github.com/processing/p5.js/tree/main/contributor_docs#github-issue-flow). This is necessary for tracking development and keeping discussion clear. I will close this PR for now for organizational purpose, and please feel free to open a new PR after an issue is opened. Thanks!

rbtcollins:
@Qianqianye hoop jumped through, you should be able to re-open this PR now.

Issue #5741, opened 2022-07-29
Test issue labeler


Qianqianye:
### Most appropriate sub-area of p5.js?

- [X] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [X] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Localization Tools
- [ ] Friendly Errors
- [ ] Other (specify if possible)

### Steps:
1.
2.
3.

### Snippet:

```js

// Paste your code here :)

```

PR #5731, opened 2022-07-21
Add automatic labelling to issues

stampyzfanz:
Resolves #5636

 Changes:

Issues are now automatically labelled as per the check boxes ticked in the issue form which all types of issues include other than discussion issues. Additionally, the issue form is reordered as requested by [Qianqianye](https://github.com/Qianqianye).

A prototype is [here](https://github.com/stampyzfanz/TestingGithubActions/issues) which works as expected. 

 Screenshots of the change:

image
image

PR #5744, opened 2022-07-30
Fix issue labeler permissions and regex

stampyzfanz:
Fixed 2 bugs found by @Qianqianye in https://github.com/processing/p5.js/issues/5731#issuecomment-1200007906

 Changes:

1) The regex was relaxed to allow capital and lowercase Xs in the checkboxes and allow arbitrary whitespace between them and the issue name because issue forms have inconsistent markdown for some reason. Qianqianye noticed that #5741 didn't product expected results which has been fixed in [my fork](https://github.com/stampyzfanz/p5.js/issues/5).
2) The write permission was granted to the workflow using the [permissions key](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) so that it can actually add the label. This was only added to the labeler workflow and should not impede the security of any other GitHub Actions. 

Something else I noticed was a warning in the GitHub Action from the use of an undefined key `enable-versioned-regex`, however, the key is [reccomended by the docs](https://github.com/github/issue-labeler). Additionally, it still ends in an error message [which is a known bug](https://github.com/github/issue-labeler/issues/32). However, it still works as expected and can label issues in [my fork](https://github.com/stampyzfanz/p5.js/issues) even after I reduced the permissions of the `GITHUB_TOKEN` to try to emulate this repository. 

Just let me know if I should create an issue for this.

 Screenshots of the change:

#### PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,32 +1,32 @@
 "Area:Accessibility":
-  - '\[X\] Accessibility'
+  - '\[[xX]\]\s*Accessibility'
 "Area:Color":
-  - '\[X\] Color'
+  - '\[[xX]\]\s*Color'
 "Area:Core":
-  - '\[X\] Core'
+  - '\[[xX]\]\s*Core'
 "Area:Data":
-  - '\[X\] Data'
+  - '\[[xX]\]\s*Data'
 "Area:DOM":
-  - '\[X\] DOM'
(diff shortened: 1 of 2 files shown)


Qianqianye:
Thank you @stampyzfanz! I did a test in issue #5745 and the labels totally worked. Thanks!

Issue #5308, opened 2021-06-13
WEBGL - Is ambientMaterial() the same as fill()?


JetStarBlues:
### How would this new feature help increase access
- Document behaviour

#### Most appropriate sub-area of p5.js?
- [x] Other (documentation)

#### Details:
I'm currently working on improving the material documentation.

The current documentation for [`ambientMaterial()`][0] does not mention [`fill()`][1]. I'm wondering if the two are equivalent since they have identical behaviour. See [here][2] for live demo.

I.e. is `fill(255, 0, 0)` the same as `ambientMaterial(255, 0, 0)`?

If they are equivalent, I would like to mention this in the documenation.

[Link][3] to `fill` (WEBGL) source code.
[Link][4] to `ambientMaterial` source code.

[0]: https://p5js.org/reference/#/p5/ambientMaterial
[1]: https://p5js.org/reference/#/p5/fill
[2]: https://editor.p5js.org/acedean/sketches/ir0F1NMLg
[3]: https://github.com/processing/p5.js/blob/1250864d63e6353233789f0059f62ebe0188a2a5/src/webgl/p5.RendererGL.js#L632
[4]: https://github.com/processing/p5.js/blob/1250864d63e6353233789f0059f62ebe0188a2a5/src/webgl/material.js#L636

stalgiag:
I didn't design this and I find it a bit too subtle but there is a difference. `ambientMaterial` sets the material to a specific shader. `fill()` does not do this. Both set the `curFillColor` so this what makes there relationship a bit confusing. For example:

```js
  specularMaterial(0, 0, 255);
  fill(255, 0, 0);
  sphere(100);
```
results in a sphere with a red specular material applied.
whereas
```js
  specularMaterial(0, 0, 255);
  ambientMaterial(255, 0, 0);
  sphere(100);
```
results in a sphere with a red ambient material applied.

Perhaps a slight redesign or a better articulation of the distinction in the docs is in order but this is my understanding of the difference currently.

JetStarBlues:
Huh. Thank you for the insight.

So the current `fill()` behaviour in WEBGL mode can be described as:

> Sets the color of the current material

\---

_Sidenote:_ is it correct to state that the default material of a sketch (when not explicitly created with a call to `specularMaterial` etc.) is `ambientMaterial`?

\---
[...]

PR #5774, opened 2022-08-30
Feature Implemeted -- Multiple Material support for geometry

### LINKED ISSUE #3806: Multiple types of materials to support WebGL Geometries

ShenpaiSharma:
Resolves #3806 #5308

 Changes:

Implemented a few new uniforms (uSpecularMatColor, uAmbientMatColor, uEmissiveMatColor) which store the RGB colors for different materials and fill the geometry with different color properties and thus contributing separately to the lighting of the surface.

So finally, we have decided to move away from the current overwriting of fill colors to the geometry, towards the approach that Processing uses where the material has different color properties (ambient, emissive, specular) that all contribute separately to the lighting of the surface.

 Screenshots of the change:

Before:
image

[Code](https://editor.p5js.org/ShenpaiSharma/sketches/Rg-58jQmM)

After:
image

[Code](https://editor.p5js.org/cfoss/sketches/xTZuZViai)

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/lib/empty-example/sketch.js
+++ b/lib/empty-example/sketch.js
@@ -4,18 +4,30 @@ function setup() {
 }
 
 function draw() {
-  background(100);
+  background(0);
   // translate(width/2, height/2);
 
-  specularColor(255, 255, 255);
-  directionalLight(127, 127, 127, cos(frameCount * 0.1), 1, -1);
+  // specularColor(255, 255, 255);
+  // directionalLight(127, 127, 127, cos(frameCount * 0.1), 1, -1);
+
   ambientLight(255, 255, 255);
 
-  // fill(255, 0, 0);
-  // emissiveMaterial(0, 255, 0);
-  ambientMaterial(0, 0, 255);
-  specularMaterial(0, 255, 0);
-  // emissiveMaterial(255, 0, 255);
-  shininess(100);
+  fill(0, 0, 255);
+  // emissiveMaterial(255, 0, 0);
+  // ambientMaterial(0, 0, 255);
+  // specularMaterial(0, 0, 255);
+  // shininess(32);
   sphere(100);
 }
+
+
+// function setup() {
+//   createCanvas(400, 400, WEBGL);
+//   noStroke();
+// }
+
+// function draw() {
+//   background(0);
+//   directionalLight(127, 127, 127, cos(frameCount * 0.1), 1, -1);
+//   sphere(100);
+// }
--- a/lib/empty-example/sketch.js
+++ b/lib/empty-example/sketch.js
@@ -1,7 +1,20 @@
 function setup() {
-  // put setup code here
+  createCanvas(600, 600, WEBGL);
+  noStroke();
 }
 
 function draw() {
-  // put drawing code here
+  background(0);
+  // translate(width/2, height/2);
+
+  // SpecularColor(255, 255, 255);
+  directionalLight(127, 127, 127, cos(frameCount * 0.1), 1, -1);
+  ambientLight(255, 255, 255);
+
+  // fill(255, 0, 0);
+  // emissive(255, 0, 0);
+  ambientMaterial(0, 0, 255);
+  specularMaterial(255, 0, 0);
+  shininess(100);
+  sphere(100);
 }
--- a/lib/empty-example/sketch.js
+++ b/lib/empty-example/sketch.js
@@ -4,17 +4,18 @@ function setup() {
 }
 
 function draw() {
-  background(0);
+  background(100);
   // translate(width/2, height/2);
 
-  // SpecularColor(255, 255, 255);
+  specularColor(255, 255, 255);
   directionalLight(127, 127, 127, cos(frameCount * 0.1), 1, -1);
   ambientLight(255, 255, 255);
 
   // fill(255, 0, 0);
-  // emissive(255, 0, 0);
+  // emissiveMaterial(0, 255, 0);
   ambientMaterial(0, 0, 255);
-  specularMaterial(255, 0, 0);
+  specularMaterial(0, 255, 0);
+  // emissiveMaterial(255, 0, 255);
   shininess(100);
   sphere(100);
 }
--- a/lib/empty-example/sketch.js
+++ b/lib/empty-example/sketch.js
@@ -1,33 +1,7 @@
 function setup() {
-  createCanvas(600, 600, WEBGL);
-  noStroke();
+  // put setup code here
 }
 
 function draw() {
-  background(0);
-  // translate(width/2, height/2);
-
-  // specularColor(255, 255, 255);
-  // directionalLight(127, 127, 127, cos(frameCount * 0.1), 1, -1);
-
-  ambientLight(255, 255, 255);
-
-  fill(0, 0, 255);
-  // emissiveMaterial(255, 0, 0);
-  // ambientMaterial(0, 0, 255);
-  // specularMaterial(0, 0, 255);
-  // shininess(32);
-  sphere(100);
+  // put drawing code here
 }
-
-
-// function setup() {
-//   createCanvas(400, 400, WEBGL);
-//   noStroke();
-// }
-
-// function draw() {
-//   background(0);
-//   directionalLight(127, 127, 127, cos(frameCount * 0.1), 1, -1);
-//   sphere(100);
-// }
--- a/src/webgl/material.js
+++ b/src/webgl/material.js
@@ -701,7 +701,7 @@ p5.prototype.normalMaterial = function(...args) {
 };
 
 /**
- * Sets the current material as an ambient material of the given color.
+ * Sets the ambient color of the material.
  *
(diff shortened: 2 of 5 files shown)


Qianqianye:
Thanks @ShenpaiSharma. I am inviting the GSoC mentor @calebfoss to review this PR.

calebfoss:
Awesome work, @ShenpaiSharma!

I'm thinking it would be helpful to update the inline documentation a bit to clarify that specular, emissive, ambient, and fill can be combine, but normal cannot.  

Perhaps in normalMaterial, we can add a sentence to the effect of "This material cannot be combined with other materials and will override the current material settings." In ambientMaterial, I think it would clarify to revise "Sets the current material as an ambient material of the given color. The ambientMaterial() color is the color the object will reflect under any lighting." to something like "Sets the ambient color of the current material. Objects will reflect this color under any lighting. The ambient color is combined with the current fill color as well as colors set with emissiveMaterial() and specularMaterial()." And similar adjustments to emissiveMaterial and specularMaterial.

Also something I'd be interested to hear more thoughts on before merging this PR:
- Combining material qualities/colors raises the question of how to turn these qualities off. What if, for example, you want to give one shape a specular color but then you want the following shape to not use specular? In this implementation, the way you would do that is to call specularMaterial(0). I'm wondering if that is sufficiently intuitive or if we might want to add methods called noSpecularMaterial, noEmissiveMaterial, and noAmbientMaterial to echo the 2D api?
- Now that these methods set qualities of the current material, rather than setting the renderer to use a new material, would it be more intuitive to drop the "material" in their names (i.e. specular(), emissive(), and ambient() like in Processing)? This could also help clarify the difference between these methods and normalMaterial().
@kjhollen @stalgiag @aferriss - If you have a chance, it would be helpful to hear your perspective!

kjhollen:
Congrats @ShenpaiSharma!

> * Now that these methods set qualities of the current material, rather than setting the renderer to use a new material, would it be more intuitive to drop the "material" in their names (i.e. specular(), emissive(), and ambient() like in Processing)? This could also help clarify the difference between these methods and normalMaterial().
>       @kjhollen @stalgiag @aferriss - If you have a chance, it would be helpful to hear your perspective!

I think that early on we discussed adding the word "material" to match more present-day tools that use shaders (Unity, three.js, etc) so that people familiar with other platforms would have a consistent search term. This vocabulary was less common in the early days of Processing, so I think it's ok not to match here! Curious what others think.

aferriss:
> Combining material qualities/colors raises the question of how to turn these qualities off. What if, for example, you want to give one shape a specular color but then you want the following shape to not use specular? In this implementation, the way you would do that is to call specularMaterial(0). I'm wondering if that is sufficiently intuitive or if we might want to add methods called noSpecularMaterial, noEmissiveMaterial, and noAmbientMaterial to echo the 2D api?

I might be wrong, but aren't these function calls affected by calls to push() and pop() ? Do you think that's not enough of a handle to help people control turning on and off material properties?

calebfoss:
@kjhollen, thanks for that insight! Sounds like leaving the method names as is would be more clear, particularly with the adjustments to documentation.

> I might be wrong, but aren't these function calls affected by calls to push() and pop() ? Do you think that's not enough of a handle to help people control turning on and off material properties?

@aferriss, that's correct! My thinking is that stroke() and fill() are also affected by push() and pop(), but we still have noStroke() and noFill(). That being said, push() and pop() may be more familiar to folks using webGL mode, since they'll need to use transformations to position models anyway.

Here's an example of drawing two spheres without using push() and pop(), one with specularMaterial and one without:
https://editor.p5js.org/cfoss/sketches/9Ocp292Ks

aferriss:
@calebfoss Thanks for the context and the example! I definitely see the value in what you're proposing and it makes sense to me. 

Might be worth starting another thread with this proposal just to keep things clean here, since I don't think your proposed feature would be a blocking issue for this PR.

stalgiag:
> Congrats @ShenpaiSharma!
> 
> > * Now that these methods set qualities of the current material, rather than setting the renderer to use a new material, would it be more intuitive to drop the "material" in their names (i.e. specular(), emissive(), and ambient() like in Processing)? This could also help clarify the difference between these methods and normalMaterial().
> >   @kjhollen @stalgiag @aferriss - If you have a chance, it would be helpful to hear your perspective!
> 
> I think that early on we discussed adding the word "material" to match more present-day tools that use shaders (Unity, three.js, etc) so that people familiar with other platforms would have a consistent search term. This vocabulary was less common in the early days of Processing, so I think it's ok not to match here! Curious what others think.

For the previous behavior, it felt nice to break from Processing and use "material" for the reasons that @kjhollen listed here. Though now that these methods don't switch the set of shaders used for rendering, the word 'material' at the end feels less accurate because they aren't changing what I would think of as the material but instead setting attributes of a single material. It almost feels like if we keep the term 'material' then the sequence should be reversed to communicate that the material is being modified not swapped (ie: `materialSpecular()`, `materialEmissive()` ). This feels clunky though because the unique segment of the name is at the end. This is a conundrum and the Processing names begin to sound better.

calebfoss:
Thanks, @aferriss! Happy to get that started once this is merged.

> For the previous behavior, it felt nice to break from Processing and use "material" for the reasons that @kjhollen listed here. Though now that these methods don't switch the set of shaders used for rendering, the word 'material' at the end feels less accurate because they aren't changing what I would think of as the material but instead setting attributes of a single material. It almost feels like if we keep the term 'material' then the sequence should be reversed to communicate that the material is being modified not swapped (ie: `materialSpecular()`, `materialEmissive()` ). This feels clunky though because the unique segment of the name is at the end and the differentiation from instances where the material is actually being swapped becomes more confusing. This is a conundrum and the Processing names begin to sound better as I consider it. But I am open to anything and will follow the group's lead.

@stalgiag, thanks for these thoughts! Yes, I think if I were seeing specularMaterial() for the first time, I would assume that the renderer would now use a new, shiny material, which is exactly what it's been doing until this feature. So I was a bit concerned if we change these methods' behaviors without changing their names, that might cause some confusion.

So here's what these variations look like when producing this [example](https://editor.p5js.org/cfoss/sketches/IhkuJnMhn):
image

A
```
ambientMaterial(128, 0, 0);
emissiveMaterial(0, 0, 128);
specularMaterial(0, 128, 0);
sphere(width/4);
```
B

```
materialAmbient(128, 0, 0);
materialEmissive(0, 0, 128);
materialSpecular(0, 128, 0);
sphere(width/4);
```
C
```
ambient(128, 0, 0);
emissive(0, 0, 128);
specular(0, 128, 0);
sphere(width/4);
```
While I certainly see the value in being consistent with other tools (as @kjhollen referenced) and in being consistent with how things have been named already, looking at these options, I think C reads the most clear.

A bonus benefit if we do decide to add new methods to turn these off is that `noSpecular()` is simpler to type and read than `noSpecularMaterial()` and I think clearer as well.

kjhollen:
Being mindful of the GSoC project timeframe, I don't think it's essential to change the method names immediately as a condition of this merge. 

We could decide to open a new issue and track renaming the functions as future work, as I think there's a little more to do beyond just changing the method names. For example, can we deprecate the methods as they exist now, so that we don't break anyone's code on short notice? Would we want some custom friendly error messages that indicate that the method name was/is changing?

In terms of methods for turning off types of materials, we do also have `resetShader()` and should think about how that relates to built-in materials.

Is there any additional discussion needed before merging this PR? I know @aceslowman is eager to use these new features in a web GL tutorial. Please let me know if I can help with any further review!

calebfoss:
@kjhollen this makes sense, and thank you for the additional info. This is now merged, and I'm happy to open up renaming conversations in a new thread. Sorry for holding things up!

Congratulations and thank you, @ShenpaiSharma !!!

kjhollen:
thanks @calebfoss and congrats @ShenpaiSharma!! 🥳

stalgiag:
Great work @ShenpaiSharma !

Qianqianye:
Wonderful work @ShenpaiSharma and @calebfoss 🎉

PR #5781, opened 2022-09-01
Allow setting the image width and height. Addresses #5779

annalog-codes:
Allow setting the image width and height to support proper loading of SVG images as canvas images.
This allows for better scalling of image elements relative to a given screen / canvas size.

SVG files can already be loaded using this method. However, currently the intrinsic size stored in an SVG file is used to determine its size, which works against the idea of using SVGs-files as scalable templates for canvas images. Allowing for parameters for the desired width and height in which the SVG should be interpreted renders the SVG file correctly and allows for easy scalability.

I have tested this code and it seems to work fine.

 Changes:

Two new parameters, `width` and `height`, have been added to `p5.prototype.loadImage`. If they are given they are used as the width and height for the canvas and image in the drawingContext, instead of the intrinsic image width and height.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -65,7 +65,13 @@ import '../core/friendly_errors/fes_core';
  * image of the underside of a white umbrella and grided ceililng above
  * image of the underside of a white umbrella and grided ceililng above
  */
-p5.prototype.loadImage = function(path, successCallback, failureCallback) {
+p5.prototype.loadImage = function(
+  path,
+  successCallback,
+  failureCallback,
+  width,
+  height
+) {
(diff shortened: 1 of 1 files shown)


limzykenneth:
Hi @annalog-codes, thanks for filing this PR. However in general we require issues should be discussed as approved for implementation before a PR should be filed. I'm going to close this for now and will reopen once the issue has been approved by stewards or maintainer.

You should include in the issue the proposed fix in more detail.

annalog-codes:
Hi @limzykenneth, thank you for explaining the proper workflow. I'll edit my initial comment on the issue with more details.

PR #5784, opened 2022-09-03
Image fit features

Manpreet-Singh001:
Addresses [#5720](https://github.com/processing/p5.js/issues/5720) 
 Changes:
Allows users to `CONTAIN/COVER` the specified destination box without distorting the aspect ratios.

 Screenshots of the change:
`CONTAIN`
`CONTAIN CENTER CENTER`
image
image
`CONTAIN CENTER TOP`
image
`COVER`
`COVER CENTER CENTER`
image
`COVER RIGHT CENTER`
image

#### PR Checklist

- [x] `npm run lint` passes


Manpreet-Singh001:
This is a working prototype. It doesn't include unit-tests, since I couldn't think of what the tests should look like. Here's how the logic works. 
`CONTAIN`
Since we, want to contain the entire image we find the img/destination ratio that is the biggest and decrease the destination box's width and height to the required ratio to allow the whole image to be shown with maximum allowed height or width.
`COVER`
This works a little differently from the `CONTAIN` since for images that are smaller than the destination box, we cannot get away with increasing the destination box's width/height to scale the image since it will not be what the user would expect when specifying the `dw, dh` we were able to change these values in `CONTAIN` cause they were smaller than user specified and didn't interrupt with the desired behavior. So, the image needs to be resized by the minimum factor of the ratios as to cover the whole container.

I am fairly new to open-source so any suggestions on how to improve this are welcome.

src/image/loading_displaying.js
```diff
@@ -412,25 +488,30 @@ p5.prototype.image = function(
...
+  xAlign,
+  yAlign
 ) {
   // set defaults per spec: https://goo.gl/3ykfOq
 
   p5._validateParameters('image', arguments);
 
   let defW = img.width;
   let defH = img.height;
+  yAlign = yAlign || constants.CENTER;
```
Manpreet-Singh001, line 502: default is `CENTER` if not specified by user

src/image/loading_displaying.js
```diff
@@ -412,25 +488,30 @@ p5.prototype.image = function(
...
     // video no canvas
     defW = img.elt.videoWidth;
     defH = img.elt.videoHeight;
   }
 
-  const _dx = dx;
-  const _dy = dy;
-  const _dw = dWidth || defW;
-  const _dh = dHeight || defH;
+  let _dx = dx;
```
Manpreet-Singh001, line 511: Since the `fit` mode needs to change these variables to allow fitting the entire image

src/image/loading_displaying.js
```diff
@@ -461,7 +542,43 @@ p5.prototype.image = function(
...
+        _dh,
+        _sw,
+        _sh
+      );
+      vals.x = x;
+      vals.y = y;
+      vals.w = w;
+      vals.h = h;
+    }
+  } else {
```
Manpreet-Singh001, line 579: `imageMode` shouldn't affect the image shown when `fit` mode is specified.

davepagurek: I think we still want `imageMode` to affect positioning when `fit` is specified, and it should just refine the positioning. of the source image within the destination box specified. e.g. starting from this non-`fit` code:

<table>
<tr>
<td>

```js
rectMode(CENTER)
imageMode(CENTER)
rect(0, 0, dw, dh)
image(img, 0, 0, dw, dh, 0, 0, sw, sh)
```

</td>
<td>

</td>
</tr>
</table>

adding some fit parameters, I would expect it to look like this:

<table>
<tr>
<td>

```js
rectMode(CENTER)
imageMode(CENTER)
rect(0, 0, dw, dh)
image(img, 0, 0, dw, dh, 0, 0, sw, sh, COVER, RIGHT, CENTER)
```

</td>
<td>

</td>
</tr>
</table>

but without taking into account `imageMode`, it looks like this currently:

<table>
<tr>
<td>

```js
rectMode(CENTER)
imageMode(CENTER)
rect(0, 0, dw, dh)
image(img, 0, 0, dw, dh, 0, 0, sw, sh, COVER, RIGHT, CENTER)
```

</td>
<td>

</td>
</tr>
</table>

src/image/loading_displaying.js
```diff
@@ -280,6 +280,79 @@ function _createGif(
...
+ * @param {Number} sx
+ * @param {Number} sy
+ * @param {Number} sw
+ * @param {Number} sh
+ * @returns {Object}
+ */
+
+function _imageCover(img, xAlign, yAlign, dw, dh, sx, sy, sw, sh) {
+  const r = Math.max(dw / sw, dh / sh, 1);
+  img.resize(img.width * r, img.height * r);
```
davepagurek, line 332: I think we aren't able to use `img.resize` in this method because not all input types have this method (e.g. `p5.MediaElement`, the return type of `createVideo`). Also, `resize` modifies the image object, and I think we want to keep these methods from modifying the input.

src/image/loading_displaying.js
```diff
@@ -280,6 +280,79 @@ function _createGif(
...
+ * @param {Number} dy
+ * @param {Number} dw
+ * @param {Number} dh
+ * @param {Number} sw
+ * @param {Number} sh
+ * @returns {Object}
+ */
+
+function _imageContain(xAlign, yAlign, dx, dy, dw, dh, sw, sh) {
+  const r = Math.max(sw / dw, sh / dh, 1);
```
davepagurek, line 297: Is the `Math.max(..., 1)` there to prevent the image from stretching past its original pixel size? If so, I think we can take it out, because the intended behaviour is to allow images to uniformly scale possibly above its original dimensions to fit or fill the container.

davepagurek:
Thanks for taking this on, and nice work on this! I tested different combinations of fits and fills and the alignment logic works well! I left a few comments about handling more edge cases. Let me know if I can clarify anything!

src/image/loading_displaying.js
Manpreet-Singh001: Currently, it doesn't take `imageMode` into account and produces the result as if `imageMode` is always `CORNER`. I was not able to reproduce this image going out of bound. I would appreciate if you could send me the `src` for this sketch. Also, my question regarding `imageMode` is if `imageMode` is `CENTER` and the user inputs 
`image(img, 0, 0, dw, dh, 0, 0, sw, sh, COVER, RIGHT, CENTER` should we consider `yAlign: center` from the center of the image.

src/image/loading_displaying.js
Manpreet-Singh001: It calculates, the minimum needed ratio that we need to scale the destination box by, so that we don't distort the image aspect ratio. Let me, know if it still confuses you.

davepagurek:
Also, here's an idea of how we can unit test this: All `image()` calls end up calling `this._renderer.image()` at the end. Maybe we can make assertions on the parameters that get passed to that function? We could do this with [spies](https://sinonjs.org/releases/latest/spies/), something like:

```js
test('COVER when the source image is taller than the destination', function() {
  let src = myp5.createImage(20, 100);
  sinon.spy(myp5._renderer);
  myp5.image(src, 0, 0, 300, 400, 20, 100, myp5.COVER);
  assert(myp5._renderer.image.calledOnce);
  assert.equal(
    myExternalLibrary._doNetworkCall.getCall(0).args[3], // sw
    20
  );
  assert.equal(
    myExternalLibrary._doNetworkCall.getCall(0).args[4], // sh
    20 * (400 / 300)
  );
  // etc
});
```

src/image/loading_displaying.js
```diff
@@ -539,6 +539,17 @@ p5.prototype.image = function(
 
   let vals = { x: _dx, y: _dy, h: _dh, w: _dw };
   if (fit) {
+    switch (this._renderer._imageMode) {
```
davepagurek, line 541: This is close to what we want! I think the image mode would change the destination box size (and shift the other way for `CENTER`, i.e.:
```js
switch (this._renderer._imageMode) {
  case constants.CENTER:
    _dx -= _dw * 0.5;
    _dy -= _dh * 0.5;
    break;
  case constants.CORNERS:
    _dw -= _dx;
    _dh -= _dy;
    break;
}
```

...although I wonder if we can reduce the complexity by using `modeAdjust`, since that does these adjustments already? Something like this, where you pass `vals.w`, `vals.h`, `vals.x` and `vals.y` for `dw`, `dh`, dx`, and `dy`:

```js
const vals = canvas.modeAdjust(_dx, _dy, _dw, _dh, this._renderer._imageMode);

if (fit) {
  if (fit === constants.COVER) {
    const { x, y, w, h } = _imageCover(
        xAlign,
        yAlign,
        vals.w, // Replacing dw
        vals.h, // Replacing dh
        _sx,
        _sy,
        _sw,
        _sh
      );
      // etc
```

test/unit/image/loading.js
```diff
@@ -443,3 +443,56 @@ suite('displaying images', function() {
...
+    assert.equal(myp5._renderer.image.getCall(0).args[4], 400); // sh
+  });
+
+  test('COVER when source image is smaller than destination', function() {
+    let src = myp5.createImage(20, 100);
+    sinon.spy(myp5._renderer, 'image');
+    myp5.image(src, 0, 0, 300, 400, 0, 0, 20, 100, myp5.COVER);
+    assert(myp5._renderer.image.calledOnce);
+    assert.equal(myp5._renderer.image.getCall(0).args[3], 20); //  sw
+    assert.equal(myp5._renderer.image.getCall(0).args[4], 400 / 15); // sh
```
davepagurek, line 496: Great work on these tests!

src/image/loading_displaying.js
```diff
@@ -280,6 +280,73 @@ function _createGif(
...
+ * @param {Number} dh
+ * @param {Number} sx
+ * @param {Number} sy
+ * @param {Number} sw
+ * @param {Number} sh
+ * @returns {Object}
+ */
+
+function _imageCover(xAlign, yAlign, dw, dh, sx, sy, sw, sh) {
+  const r = Math.max(dw / sw, dh / sh, 1);
```
davepagurek, line 329: Oh I just noticed, if you use an image that's larger than the destination, it doesn't scale down to minimally cover (because that would involve a scale lower than 1.) I think if we remove the `, 1` here then it will handle that case though!

src/image/loading_displaying.js
Manpreet-Singh001: I tried using the existing `modeAdjust` function but the way it works is by shifting the `dx or dy` so as the image appears to be drawn from the center. The reason I couldn't use it to is because it in fitMode `COVER` moving away the destination box makes the image not cover the whole required box, which seems a bit counter intuitive.

src/image/loading_displaying.js
Manpreet-Singh001: Will fix this, I assumed that minimal cover was not required. But, looking at css `cover` it does actually minimally covers the larger image.

src/image/loading_displaying.js
Manpreet-Singh001: I think the complexity can be addressed by taking this logic outside into one function call.

davepagurek: > in fitMode COVER moving away the destination box makes the image not cover the whole required box

I think `sx` and `sy`, `sw`, and `sh` only change what part of the source image gets put into the destination box, wherever the destination box is, so when you shift the destination box, the source image shifts with it. This would be consistent with how `imageMode` works with `image` when you don't provide any fit parameters: changing from `imageMode(CORNER)` to `imageMode(CENTER)` just shifts where the image is drawn, but the image itself is the same as before.

Let me know if I'm misinterpreting what your concern is, but I think your `_imageCover` and `_imageContain` functions will produce the expected output if we give them updated `dx`, `dy`, `dw`, and `dh` values, either from manually applying shifts as we're doing currently, or by giving them the outputs of `modeAdjust`.

src/image/loading_displaying.js
```diff
@@ -536,56 +579,33 @@ p5.prototype.image = function(
...
-      vals.x = x;
-      vals.y = y;
-      vals.w = w;
-      vals.h = h;
-    }
-  } else {
-    vals = canvas.modeAdjust(_dx, _dy, _dw, _dh, this._renderer._imageMode);
-  }
+  let vals = canvas.modeAdjust(_dx, _dy, _dw, _dh, this._renderer._imageMode);
+  vals = _imageFit(
```
davepagurek, line 583: Nice, this code looks great!

davepagurek:
I'll run the code again this evening to double check that everything works as expected, but unless anything new comes up, this seems good to go! Great work!

The one final thing we might want to do is add an example to the comments of `image()` showing usage of fitting so that readers of the p5 docs can see how it might be used. Maybe we can add an example sketch that shows one COVER and one CONTAIN so they can see the difference?

davepagurek:
Just played around with it some more, this looks good to me! I think once we have an example in the docs, this is good to go :grin:

src/image/loading_displaying.js
```diff
@@ -424,6 +424,14 @@ function _sAssign(sVal, iVal) {
  * to explain further:
  * <img src="assets/drawImage.png"></img>
  *
+ * This function can also be used to draw images without distorting the orginal aspect ratio,
+ * by adding 9th parameter fit which can either be COVER or CONTAIN.
```
davepagurek, line 428: Super minor grammar suggestion: I think a few commas on this line and the next will make this a tad easier to read:
```diff
-  * by adding 9th parameter fit which can either be COVER or CONTAIN.
+  * by adding 9th parameter, fit, which can either be COVER or CONTAIN.
-  * CONTAIN as the name suggests contains the whole image within the specified destination box
+  * CONTAIN, as the name suggests, contains the whole image within the specified destination box
 ```

**Comment by davepagurek** [comment]
Thanks for adding those examples!

File: src/image/loading_displaying.js [hunk 11]
```diff
@@ -490,6 +498,36 @@ function _sAssign(sVal, iVal) {
...
+ * <div>
+ * <code>
+ * let img;
+ * function preload() {
+ *   img = loadImage('assets/laDefense50.png'); // dimensions of image are 50 x 50
+ * }
+ * function setup() {
+ *   // COVER the whole destination box without distorting the image's aspect ratio
+ *   // COVER the specified destination box which is of dimension 100 x 100
+ *   // Display the CENTER part of the image for both LEFT and TOP corner
```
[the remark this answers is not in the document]
Manpreet-Singh001: Thanks for the suggestion, this makes it a lot more easier to read.

davepagurek (approved):
I think this is all ready to go now! Thanks for the time you put into this feature 🙂

davepagurek:
@all-contributors please add @Manpreet-Singh001  for 📖💻⚠️

davepagurek:
@all-contributors please add @Manpreet-Singh001 for doc, code, test

PR #5812, opened 2022-09-25
Include bundled third party libraries and p5.js' license in p5.js and p5.min.js

### LINKED ISSUE #5615: p5.min.js does not include the licenses for bundled third party libraries (or itself)
### Topic

The released [p5.min.js](https://github.com/processing/p5.js/releases/download/v1.4.1/p5.min.js) file contains no license for p5js nor any licenses for the third party dependencies. However, searching through [p5.js](https://github.com/processing/p5.js/releases/download/v1.4.1/p5.js) there are a few licenses left in there. (e.g. search for the text "MIT License"

I don't have much knowledge of licensing and just trying to ensure a sketch I'm releasing will be compliant. I believe all third party licenses need to be included either in the p5.min.js file or I need some way to include them within my own bundle.

Is there a switch on the minifcation process that should be preserving these in the minified file? Also, should the p5.js LGPL license not also be included within the minified file?

stampyzfanz:
Resolves #5615

### Changes:

1.  Ensure inline licenses from bundled third party libraries are kept in p5.min.js. 
2.  Ensure that the LGPL license remains in at the beginning of p5.js & p5.min.js 

### Some things I'm not sure about
- p5.js' entire 498 line LGPL license is used at the beginning of the files. Should a shorter version be used instead? Also, the entire license is kept within the browserify.js and Gruntfile.js files but maybe it should read from the license.txt file.
- I may have missed where some licensed code was used in my searches of "MIT" and "copyright".
- There is licensed code used in src/webgl/shaders/line.vert was already included because it wasn't minified due to prior processing and encoding as a string. If you search p5.min.js for "Ben Fry and Casey Reas" you can that it remains in the code, however I do not know how to tell Uglify how to preserve specific line breaks. Someone more skilled than me could implement the line breaks if it is deemed important although it may not even be possible in Grunt's Uglify, the tool p5 uses for minification.
- For some reason, designated comments are only preserved if they proceed a function with a signature that is kept as is in the minified file. In src/dom/dom.js, there was no way I could work out how to preserve the comment directly above the internal Cue class. Because of this, I moved it underneath but there may be a better solution I couldn't find.
image

- I had to precede all the preserved multiline comments with single line comments so that the indentation would be correct. I believe that the single line comments aren't serving their purpose to explain the code around them, but I kept them just so that the output in p5.min.js would be correct. Maybe their purpose should be explained more clearly in the comments.
- Should we add this to the developer documentation so people know how to preserve comments relating to licenses, and if so where?
- Should we include the entirety of MIT's license each time it is referred to? I made sure to use the same amount of information as there was previously, but after reflection of what MIT's license says, it seems that the whole license should be reproduced each time.

Note that other libraries such as p5.sound.js may need to do this as well, I have not checked them.

Selected extracts of output of minification. 
image
image

#### PR Checklist

- [x] `npm run lint` passes (It doesn't pass locally due to line endings but it passes on GitHub)
- [ ] [Unit tests] are included / updated (I don't think this project has unit tests for the build process. Let me know if this is required.)


limzykenneth:
There's two aspect of this I would like to bring up here.

---

The first is about including license text itself in the bundled library. I have look around and the biggest problem is there is no guidance whatsoever about how to deal with bundled dependencies licenses in JS, even from the OSI, and any other advices I can find always starts with "I'm not a lawyer and this is not legal advice" which I don't know how to interpret in terms of is this advice trustworthy or not. 

Many of said advice is as you implemented here which is to include full licenses in the bundled code. However I have also looked at dozens of open source libraries, up to including the most starred ones on cdnjs such as Vue, React, TailwindCSS as well as projects similar to p5.js such as Three.js, paper.js, pixi.js, Tone.js, none of them include full license text and at most just some license header about copyright holder and the type of license the code is distributed under at the top. Most don't have bundled third party dependencies but for the few that I found that did also only included a simple header and not the full license text. Some of the headers include links to the license text, most don't.

Absence of any other credible advice I really can't tell what is the "right" way to do this. In my view, including all full license text unnecessarily bloats out the bundled library, making it harder for end user to view and navigate the source code, not to mention adding many additional bytes of data that the end user need to download on what for them may be very limited bandwidth. If convention is to just include header and not full text I don't quite see the value of going against it absent authoritative advice otherwise.

---

The second is about the implementation details. Adding the full license text directly in the Gruntfile.js is a bit excessive and potentially difficult to maintain. We can instead either keep the license text as separate files or, even better, request them with direct web link to the respective projects license text file on build so even if the project update their license text we can still stay up to date without needing to modify things on our side.

The addition of single line comments to preserve line breaks is also not ideal. It is not uglify that removes the line break but rather browserify itself because the AST it uses does not preserve that information (I'm not sure if other build tools' AST preserve such information or not as I haven't looked into it but it is the case with browserify), which means that it doesn't know whether it is appropriate to insert a newline at the specific location. I need to confirm what the behaviour is like without the single line comments addition to know what the problems as solutions are though.

Many of these are also subject to whether we just do header license declaration or full license text as described in the point above.

stampyzfanz:
I used grunt to read the entire LGPL license from license.txt as per your second point. 

Since when I tested it last, I tried to get Prettier to work (still doesn't) and reinstalled Grunt. Now I can't reproduce the error in the indentation when taking away the single line comments.

This is what occured lat time

limzykenneth:
Prettier has since been removed from the build process of p5.js. The new prettification tool we use (`pretty-fast`) doesn't format comments but for most cases looks fine as it just preserves the original indentation of the comments.

Also as an additional curve ball to the original first point, opentype.js which is an important dependency for us has two external dependencies with their own licenses that are only briefly mentioned in a header comment.

limzykenneth:
Sorry to revive this after so long but I want to move towards resolution for this. I've looked at some advice and how other libraries are doing it and the general sense is that the code we distribute is covered automatically by copyright and thus by the license(s), regardless of whether it is included in the file or not (eg. the license usually have wording of applying to copy of a portion of the code). The obligation is to "publish ... copyright notice and disclaimer of warranty" which depending on reading, make available on Github repo publicly counts.

Through cdnjs I found the following: React includes copyright notice and a sentence refering to license file in source tree; Vue does not include any license text whatsoever; Three.js includes copyright notice plus an [`SPDX-License-Identifier`](https://spdx.dev/ids/); jQuery has a copyright notice with a link to license page on their website.

I'm proposing two options: 1. React style copyright notice with explainer that our license plus third party license can be found in our repo; 2. Not including any license text at all.

stampyzfanz:
No worries, I agree that it's important to reduce the amount of stale pull requests. I don't have much experience with copyright so I don't want to commit to one option, but I think that either will suit this project well.

limzykenneth:
Let's go with no having specific license text for now which is the current behaviour, I'll close this PR then. Thanks for all the inputs.

PR #5839, opened 2022-10-19
getTargetFrameRate issue #5762

### LINKED ISSUE #5762: Add a function to get the target frame rate
### Increasing Access

I am developing a library for recording p5.js animations and this proposal is to improve it.
Many already use this library, and improving its usability should increase the number of sketches posted to Twitter, Instagram, etc., and thus more access to p5.js.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

### Feature request details

I am developing a library to record p5.js animations.
In a recording library, it makes sense in many use cases to match the recording frame rate to the target frame rate in p5.js.

But it looks like the only way to get the target frame rate is to reference a variable called `_targetFrameRate`. And this is probably a private variable, not intended for user use. Libraries that depend on private variables in p5.js could suddenly stop working if p5.js is updated, which should be avoided if possible.

It would be great if p5.js supported a function to get the target frame rate as a public API, such as `getTargetFrameRate()`.

sz245:
Our team (@sz245, @jorge-mor, @4slk4) worked on issue #5762. Thank you to @donaldong, our mentor, for guiding us through the process.

Added a method that returns the frame that the user sets (_targetFrameRate) when calling the function getTargetFrameRate().
Resolves #5762.

#### Changes:
/src/core/environment.js - line 331-338

#### Code change:
``` js
/**
 * Returns the current _targetFrameRate.
 * @private
 * @return {Number} current _targetFrameRate
 */
p5.prototype.getTargetFrameRate = function() {
  return this._targetFrameRate;
};
```

#### PR Checklist

- [x] `npm run lint` passes
  - Manual browser test: 
  - frameRate() prints out actual frameRate and getTargetFrameRate() prints out _targetFrameRate using console.log()
  - Example - Use code from the following link and add the attached code below to the bottom of function draw(): https://p5.javascript.org.pl/amused-loan?template=none 
   
``` js
    let j = frameRate(20);
    console.log("frameRate(): ", j.getFrameRate())
    console.log("targetFrameRate(): ", j.getTargetFrameRate())
```
   

### DIFF
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -329,6 +329,16 @@ p5.prototype.setFrameRate = function(fps) {
 };
 
 /**
+ * Returns the current targetFrameRate.
+ * @private
+ * @return {Number} current targetFrameRate
+ */
+p5.prototype.getTargetFrameRate = function() {
+  return this._targetFrameRate;
+};
+
+
+/**
  * Hides the cursor from view.
  *
  * @method noCursor
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -329,15 +329,16 @@ p5.prototype.setFrameRate = function(fps) {
 };
 
 /**
- * Returns the current _targetFrameRate variable.
+ * Returns _targetFrameRate variable. The default _targetFrameRate is set to 60.
+ * This could be changed by calling frameRate() and setting it to the desired
+ * value. When getTargetFrameRate() is called, it should return the value that was set.
  * @method getTargetFrameRate
- * @return {Number} current _targetFrameRate
+ * @return {Number} _targetFrameRate
  * @example
  * <div><code>
  * function draw() {
  *   frameRate(20);
- *   console.log("frameRate: ", frameRate())
- *   console.log("targetFrameRate: ", getTargetFrameRate());
+ *   text(getTargetFrameRate(), width / 2, height / 2);
  * }
  * </code></div>
  */
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -329,9 +329,17 @@ p5.prototype.setFrameRate = function(fps) {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


limzykenneth:
Is this function meant to be part of the private? If it is intended for public use it would be great to have a more complete documentation as well as not marking it as `@private`.

sz245:
@limzykenneth The function is meant for public use and is now marked as `@method`. The in-line documentation has also been updated and an example has been included.

src/core/environment.js
```diff
@@ -328,6 +328,23 @@ p5.prototype.setFrameRate = function(fps) {
...
+/**
+ * Returns the current _targetFrameRate variable.
+ * @method getTargetFrameRate
+ * @return {Number} current _targetFrameRate
+ * @example
+ * <div><code>
+ * function draw() {
+ *   frameRate(20);
+ *   console.log("frameRate: ", frameRate())
+ *   console.log("targetFrameRate: ", getTargetFrameRate());
```
limzykenneth, line 340: Can these be displayed without using `console.log`? As it is not common to have the browser console open while viewing the documentation. You can do something similar to `frameCount()`'s example perhaps?

limzykenneth (changes requested):
One more changes inline.

Would you be able to add unit tests for this functions? No worries if not, we can add it later.

src/core/environment.js
sz245: Yes, will work on that.

limzykenneth:
Looks good. Thanks!

PR #5846, opened 2022-10-25
NaN and infinity check for isNumber

### LINKED ISSUE #5675: Suggestion:  include an error message when drawing shapes with `NaN` values.
### Increasing Access

If the user tries to draw a shape where any of the arguments are `NaN` it does not draw anything.  I would expect that it would log an error using the friendly error system so that the user knows why their code didn't do what they intended.  This would make it much easier for beginners to debug their code.

- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [X] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

Initially I assumed that p5.js was somehow catching or avoiding a DOM error.  Unfortunately it seems like the canvas API itself does not give any errors or warnings when encountering `NaN` values (SVG does ☹️).  So p5.js would need to check and validate arguments on its own.

We would want to include an `NaN` check in all shape functions such as `rect()`, `line()`, etc.

Here is an example code but it's very uninteresting.  It doesn't do anything.
```
function setup() {
  createCanvas(400, 400);
  vanillaJs();
}

function draw() {
  background(220);
  fill(0);
  const myVar = undefined + 1;
  rect(0, myVar, 100, 100);
  line(100, 100, myVar, 100);
}

function vanillaJs() {
  const canvas = document.createElement('canvas');
  canvas.width = 400;
  canvas.height = 400;
  document.body.appendChild(canvas);
  const ctx = canvas.getContext('2d');
  ctx.fillStyle = 'red';
  const myVar = undefined + 1;
  ctx.fillRect(0, myVar, 100, 100);
}
```

oscarczer:
Resolves #5675 

 Changes:

Implemented an additional checker for null strings and infinite values in isNumber within validate_params.js

 Screenshots of the change:

Screen Shot 2022-10-25 at 11 28 20 pm

#### PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/src/core/friendly_errors/validate_params.js
+++ b/src/core/friendly_errors/validate_params.js
@@ -344,7 +344,7 @@ if (typeof IS_MINIFIED !== 'undefined') {
   const isNumber = param => {
     switch (typeof param) {
       case 'number':
-        return true;
+        return isFinite(param);
       case 'string':
         return !isNaN(param);
       default:
--- a/src/core/friendly_errors/validate_params.js
+++ b/src/core/friendly_errors/validate_params.js
@@ -346,7 +346,7 @@ if (typeof IS_MINIFIED !== 'undefined') {
       case 'number':
         return isFinite(param);
       case 'string':
-        return !isNaN(param);
+        return !(isNaN(param) || param == "null");
       default:
         return false;
     }
--- a/test/unit/core/error_helpers.js
+++ b/test/unit/core/error_helpers.js
@@ -177,6 +177,28 @@ suite('Error Helpers', function() {
         'ValidationError type is correct'
       );
     });
+
+    testUnMinified('line: null string given', function() {
+      let err = assert.throws(function() {
+        p5._validateParameters('line', [1, 2, 4, "null"]);
+      }, p5.ValidationError);
+      assert.strictEqual(
+        err.type,
+        'WRONG_TYPE',
+        'ValidationError type is correct'
+      );
+    });
+
+    testUnMinified('line: infinite value given', function() {
+      let err = assert.throws(function() {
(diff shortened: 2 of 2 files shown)


src/core/friendly_errors/validate_params.js
```diff
@@ -344,9 +344,9 @@ if (typeof IS_MINIFIED !== 'undefined') {
   const isNumber = param => {
     switch (typeof param) {
       case 'number':
-        return true;
+        return isFinite(param);
```
limzykenneth, line 347: I don't think finite number check should be implemented here as mentioned in the original issue there could still be utility for using `Infinity` in certain circumstances and if drawing functions are ever to enforce finite only numbers they should be done in the drawing functions themselves instead of in the general case here.

src/core/friendly_errors/validate_params.js
```diff
@@ -344,9 +344,9 @@ if (typeof IS_MINIFIED !== 'undefined') {
...
       case 'string':
-        return !isNaN(param);
+        return !(isNaN(param) || param == "null");
```
limzykenneth, line 349: There is another special case that needs to be accounted for here which is when `param === ""` or `param === " "` for any number of spaces in the string. Accounting for this can potentially add some complexity, alternatively we can attempt to do `parseFloat(param)` first to directly test if the result is `NaN` which also have the benefit of possibly not needing to use different switch cases here.

limzykenneth (changes requested):

Issue #5292, opened 2021-06-04
Error when using loadJSON should catch error if there is an errorCallback


dievardump:
#### Most appropriate sub-area of p5.js?

- [ ] Accessibility (Web Accessibility)
- [ ] Build tools and processes
- [ ] Color
- [ ] Core/Environment/Rendering
- [x] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Other (specify if possible)

#### Details about the bug:

- p5.js version: 1.3.1
- Web browser and version: Firefox 88.0.1 
- Operating System: Linux
- Steps to reproduce this:
```
let meta = { 
  // default value
  metadata: {}
};

function preload() {
  const uri = 'unknown.json';
  
  function onLoad(data) {
    meta.metadata = data;
  }
  
  function onError(error) {
    console.log(error);
  }
  
  loadJSON(uri, onLoad, onError);
}

function setup() {
  createcanvas(windowWidth, windowHeight);
  fill(255);
  rect(0, 0, 250, 250);
}
```
I am trying to fetch a file. This file can be present or not, therefore I have default data already set. 

Since I provide an `onError` callback, I would expect the error to be catched and passed to the onError callback.

However, here, the error is passed to onError but is not catched, which stops the current process. 
-> This is not the default behavior an errorCallback should have. If provided, it explicitly means I wish to handle the error myself and it should not propagate.

limzykenneth:
Possibly related to #5032

chepecarlos:
same error in version 1.6.0

PR #5918, opened 2022-12-25
Made image loading failure in preload() not to stop whole process

### LINKED ISSUE #5032: loadImage can brake the whole sketch
If you take the default example and add this:

```js
function preload() {
	loadImage("banner_512x128.mp4", (img)=> {

	},
	(err)=> {
		console.log("mp4 you idiot!");
	});
}
```

Then you get, as expected "mp4 you idiot!" and "p5.js says: It looks like there was a problem loading your image. Try checki....".

I know this is plain stupid, but it also means a corrupt image can break a whole sketch.

r21nomi:
Resolves #5032 

### Issue
- `setup()`, `draw()` was not called if an image loading in `preload()` failed
  - It's inconvenient for the case like the image was on server and network connection issue

### Changes
- Made image loading failure in `preload()` not to stop whole process
  - `setup()`, `draw()` will be called if there was an failure callback in `p5.loadImage()`
  - `setup()`, `draw()` will not be called if there was no failure callback in `p5.loadImage()`

I know @limzykenneth says [this might be breaking changes](https://github.com/processing/p5.js/issues/5032).
So please check this changes are right way or not.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -104,6 +104,7 @@ p5.prototype.loadImage = function(path, successCallback, failureCallback) {
           e => {
             if (typeof failureCallback === 'function') {
               failureCallback(e);
+              self._decrementPreload();
             } else {
               console.error(e);
             }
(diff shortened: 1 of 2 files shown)


limzykenneth:
I think this is fine to fix now that I had a look at it. There is no easy semantic way to continue the sketch after a handled error halted the sketch and a error halted sketch will probably not be counted as a semantic sketch so changing its behaviour won't be a breaking change.

#5292 Can be fix in the same manner as well (along with other load functions with the same behaviour)

PR #5944, opened 2023-01-08
Update mocha to latest version

limzykenneth:
Addresses #5939

 Changes:

There is a small API change that prevents mocha from exiting correctly which has been fixed. As such mocha has been updated to the latest version.

Also to note is that NPM has finally fixed the install bug that prevents up from upgrading node version and now with node 18 and npm 8.19.3, the install works fully. I haven't test which precise version of npm start to have the fix but in any case we don't have to suggest node 14 anymore and just skip to node 18. `package-lock.json` has been regenerated accordingly.

#### PR Checklist

- [x] `npm run lint` passes


Qianqianye:
Thanks @limzykenneth!

davepagurek (approved):
Thanks for looking into this! Just checked it out, looks good!

Issue #5114, opened 2021-03-23
Clean-up logging in build scripts


dhowe:
Not a critical issue, but there are massive numbers of warnings and debugging output in the bill scripts that make it difficult to tell if things are working correctly. Would be nice to clean these up. A bit of output is shown below but it even worse in the tests:
```

$ grunt
(node:47751) Warning: Accessing non-existent property 'cat' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:47751) Warning: Accessing non-existent property 'cd' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'chmod' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'cp' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'dirs' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'pushd' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'popd' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'echo' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'tempdir' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'pwd' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'exec' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'ls' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'find' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'grep' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'head' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'ln' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'mkdir' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'rm' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'mv' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'sed' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'set' of module exports inside circular dependency
(node:47751) Warning: Accessing non-existent property 'sort' of module exports inside circular dependency
[...]

PR #6028, opened 2023-02-22
Build output cleanup

Ucodia:
Addresses #5114

#### Changes

First change upgrades `simple-git` package version replaces usage of `require('simple-git/promise'')` by `require('simple-git')` which fixes the following warning at the top of the build output:

```
=============================================
simple-git has supported promises / async await since version 2.6.0.
 Importing from 'simple-git/promise' has been deprecated and will
 report this error until the next major release of version 4.

To upgrade, change all 'simple-git/promise' imports to just 'simple-git'
=============================================
```

Second change upgrades the transitive dependency `istanbul-reports` from `grunt-simple-nyc` package which fixes the following warning at the bottom of the build output:

```
Handlebars: Access has been denied to resolve the property "statements" because it is not an "own property" of its parent.
You can add a runtime option to disable the check or this warning:
See https://handlebarsjs.com/api-reference/runtime-options.html#options-to-control-prototype-access for details
Handlebars: Access has been denied to resolve the property "branches" because it is not an "own property" of its parent.
You can add a runtime option to disable the check or this warning:
See https://handlebarsjs.com/api-reference/runtime-options.html#options-to-control-prototype-access for details
Handlebars: Access has been denied to resolve the property "functions" because it is not an "own property" of its parent.
You can add a runtime option to disable the check or this warning:
See https://handlebarsjs.com/api-reference/runtime-options.html#options-to-control-prototype-access for details
Handlebars: Access has been denied to resolve the property "lines" because it is not an "own property" of its parent.
You can add a runtime option to disable the check or this warning:
See https://handlebarsjs.com/api-reference/runtime-options.html#options-to-control-prototype-access for details
```

#### Screenshots of the change:

Not applicable.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8723,27 +8723,6 @@
         "node": ">=0.12.0"
       }
     },
-    "node_modules/handlebars": {
-      "version": "4.7.7",
-      "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.7.7.tgz",
-      "integrity": "sha512-aAcXm5OAfE/8IXkcZvCepKU3VzW1/39Fb5ZuqMtgI/hT8X2YgoMvBY5dLhq/cpOvw7Lk1nK/UF71aLG/ZnVYRA==",
-      "dev": true,
-      "dependencies": {
-        "minimist": "^1.2.5",
-        "neo-async": "^2.6.0",
-        "source-map": "^0.6.1",
-        "wordwrap": "^1.0.0"
-      },
-      "bin": {
-        "handlebars": "bin/handlebars"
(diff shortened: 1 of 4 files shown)


limzykenneth:
We are not actually using those release scripts anymore and they along with `simple-git` can probably be removed. I'll merge this anyway for now though.

limzykenneth:
The release process are all handled via Github Actions CI whenever a version tag is pushed. It is documented here: https://github.com/processing/p5.js/blob/main/contributor_docs/release_process.md

Issue #6038, opened 2023-02-28
Fatal error When running npm test on fresh github clone


shamun-khatri:
### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [X] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors
- [ ] Other (specify if possible)

### Operating System

Windows 

### Steps to reproduce this

While setting up p5.js locally i am getting this error. whn i press command npm test after npm ci.
### Snippet:

Running "eslint-samples:source" (eslint-samples) task
Fatal error: Cannot read properties of undefined (reading 'prefixLength')

```

**Comment by limzykenneth** [comment]
Can we get the node.js and npm version you have?

**Comment by aditya-shrivastavv** [comment]
I am facing the same problem. 
node version - "18.12.1"

**Comment by limzykenneth** [comment]
@aditya-shrivastavv Can you share the npm version as well? The OS version could be helpful too.

**Comment by aditya-shrivastavv** [comment]
nvm version = **1.1.10**

OS version-
Windows 11
version = **22H2**

Another problem I am having is with prettier, The problem is so annoying that I had to disable prettier. When I save the file so many changes happens.
for example -

```js
function foo() {

}
// turns to 
function foo () {  // gap inserted

}

const obj = {
 prop1: value1,
 prop2: value2
}
// turns to
const obj = {
 prop1: value1,
 prop2: value2,  // comma automatically inserted
}
```

and so many changes like this
prettier latest version

monmon2003:
Include this line of code inside rules in .eslintrc.json file : "linebreak-style": ["error", "windows"]
On including this, npm test works
![2023-03-03](https://user-images.githubusercontent.com/122162780/222529555-838f67fa-7bbd-4b84-b8d6-ceeeb17701da.png)

limzykenneth:
@aditya-shrivastavv I can't fully replicate, if you can find out a bit more why that's happening it would be great. We do not use prettier for the formatting of p5.js and I don't recommend the use of prettier here.

[...]

PR #6044, opened 2023-03-03
removed .prettierrc file

aditya-shrivastavv:
### Changes:

since **prettier** dependency was _removed_ from the project then their is no use of **.prettierrc** file, that should be _removed._ this thing is already discussed with @limzykenneth on issue #6038 

### Impact:

1. no impact other files or anything.
2. will be helpful to all the contributors who have **vscode prettier** extension installed, after the file will be _removed_ the _extension_ will not get automatically activated thus no unnecessary style related changes will take place. I was facing this problem so I had to disable my extension manually.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,3 +1,0 @@
-{
-  "singleQuote": true
-}


limzykenneth:
Looks good. Thanks!

PR #6076, opened 2023-03-17
rename identifier `globalFunctions` to `classesWithGlobalFns`

aditya-shrivastavv:
Related to #6055 

 Changes:
Since `globalFunctions` identifier refers to an array that contains classes whose functions are exposed in the global scope, the name of the identifier is not accurate and might be confusing, So I renamed it to `classesWithGlobalFns` which is very clear.

Already discussed with @davepagurek in #6055


davepagurek (approved):
Thanks for following up and making this change!

PR #6095, opened 2023-03-31
Simplified isNumber function

### LINKED ISSUE #5675: Suggestion:  include an error message when drawing shapes with `NaN` values.

anpanring:
Resolves #5675

Changes: Going off of discussion surrounding Issue #5675 and PR #5846, this simplifies the isNumber parameter validation function using parseFloat and isNaN to return false when users use null, NaN, or an empty string (w/ or without spaces) as a parameter when drawing shapes. It still accepts Infinity for special circumstances.

Screenshots of the change:
Screenshot 2023-03-31 at 1 19 58 PM

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/test/unit/core/error_helpers.js
+++ b/test/unit/core/error_helpers.js
@@ -189,9 +189,9 @@ suite('Error Helpers', function() {
       );
     });
 
-    testUnMinified('line: infinite value given', function() {
+    testUnMinified('line: NaN value given', function() {
       let err = assert.throws(function() {
-        p5._validateParameters('line', [1, 2, 4, Infinity]);
+        p5._validateParameters('line', [1, 2, 4, NaN]);
       }, p5.ValidationError);
       assert.strictEqual(
         err.type,
--- a/test/unit/core/error_helpers.js
+++ b/test/unit/core/error_helpers.js
@@ -177,6 +177,28 @@ suite('Error Helpers', function() {
         'ValidationError type is correct'
       );
     });
+
+    testUnMinified('line: null string given', function() {
+      let err = assert.throws(function() {
+        p5._validateParameters('line', [1, 2, 4, 'null']);
+      }, p5.ValidationError);
+      assert.strictEqual(
+        err.type,
+        'WRONG_TYPE',
+        'ValidationError type is correct'
+      );
+    });
+
+    testUnMinified('line: infinite value given', function() {
+      let err = assert.throws(function() {
+        p5._validateParameters('line', [1, 2, 4, Infinity]);
+      }, p5.ValidationError);
+      assert.strictEqual(
+        err.type,
+        'WRONG_TYPE',
+        'ValidationError type is correct'
+      );
+    });
   });
 
   suite('validateParameters: trailing undefined arguments', function() {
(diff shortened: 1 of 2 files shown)


limzykenneth:
This implementation does not take into account some edge case behaviour of `parseFloat`, specifically `parseFloat` can parse a string as long as it starts with a number so `parseFloat("100abcd")` will get parsed into `100` and the whole implementation here will regard `"100abcd"` as a number when it is not. There are also number base annotated strings that cannot be handled by `parseFloat` as well but that is less of an issue.

test/unit/core/error_helpers.js
```diff
@@ -177,6 +177,28 @@ suite('Error Helpers', function() {
...
+    testUnMinified('line: infinite value given', function() {
+      let err = assert.throws(function() {
+        p5._validateParameters('line', [1, 2, 4, Infinity]);
+      }, p5.ValidationError);
+      assert.strictEqual(
+        err.type,
+        'WRONG_TYPE',
+        'ValidationError type is correct'
+      );
+    });
```
limzykenneth, line 201: CI is failing on this test case, I don't think we need to cover for `Infinity` as well since `Infinity` is a valid number that can be useful in certain circumstances (eg. `min()` `max()`)

anpanring: My bad, that makes sense! I just pushed a commit to slightly change that test to check for NaN instead of Infinity since that should return a ValidationError and checking for NaN was originally the main objective.

limzykenneth:
Looks good. Thanks!

PR #6105, opened 2023-04-14
Improve zooming behavior with orbitControl()

### LINKED ISSUE #6104: Improve scaling specs when executing webgl's orbitControl()
### Increasing Access

Currently, webgl's orbitControl() has a stuttering issue when scaling with mouse scroll.

The reason for this is that the _mouseWheelDeltaY is not directly used for judgment in the orbitControl() function, and for some reason the difference from the previous one is taken. This is unnecessary processing.

```js
if (this._mouseWheelDeltaY !== this._pmouseWheelDeltaY) {
  // zoom according to direction of mouseWheelDeltaY rather than value
  if (this._mouseWheelDeltaY > 0) {
    this._renderer._curCamera._orbit(0, 0, sensitivityZ * scaleFactor);
  } else {
    this._renderer._curCamera._orbit(0, 0, - sensitivityZ * scaleFactor);
  }
}
```
Another problem is that the distance to the gaze point is changed by addition and subtraction when scaling in _orbit().

```js
camRadius += dRadius;

// prevent zooming through the center:
if (camRadius < 0) {
  camRadius = 0.1;
}
```
With this, if the scale is small, it will move too much, and if the scale is large, the distance will not change easily, making it difficult to operate comfortably.
Usability can be compromised, for example, when executing code such as:
```js
function setup() {
  createCanvas(800, 640, WEBGL);

  // case:1 (small scale)
  perspective(PI/3, width/height, 0.1, 10);

  // case:2 (large scale)
  //camera(0, 0, 1000, 0, 0, 0, 0, 1, 0);
  //perspective(PI/3, width/height, 100, 10000);
}

function draw() {
  background(220);
  orbitControl();

  // case:1
  box(0.2);

  // case:2
  //sphere(200);
}
```
Also, there is no necessity for the number 0.1. Smaller scales are possible, and larger scales are practically meaningless.

I would like to make suggestions to improve these problems.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

First, it is unnatural that only sensitivityZ is set to 0.5, so set it to 1.

before:
```js
  sensitivityZ = 0.5;
```
after:
```js
  sensitivityZ = 1;
```

Next, we will only use _mouseWheelDeltaY to determine if it is 0, and we will specify a scaling factor of 0.02 (not sure if this is a good value).

before:
```js
if (this._mouseWheelDeltaY !== this._pmouseWheelDeltaY) {
  // zoom according to direction of mouseWheelDeltaY rather than value
  if (this._mouseWheelDeltaY > 0) {
    this._renderer._curCamera._orbit(0, 0, sensitivityZ * scaleFactor);
  } else {
    this._renderer._curCamera._orbit(0, 0, - sensitivityZ * scaleFactor);
  }
}
```
after:
```js
const zoomScaleFactor = 0.02;
  // zoom according to direction of mouseWheelDeltaY rather than value
  if (this._mouseWheelDeltaY > 0) {
    this._renderer._curCamera._orbit(0, 0, sensitivityZ * zoomScaleFactor);
  } else {
    this._renderer._curCamera._orbit(0, 0, - sensitivityZ * zoomScaleFactor);
  }
}
this._mouseWheelDeltaY = 0;
```

Filally, In _orbit(), when changing the distance from the gaze point, take the base 10 logarithm and change it there. Also, for the distance, refer to cameraNear and cameraFar and limit the movement accordingly.

before:
```js
camRadius += dRadius;

// prevent zooming through the center:
if (camRadius < 0) {
  camRadius = 0.1;
} 
```
after:
```js
camRadius = Math.log10(camRadius);
camRadius += dRadius;
camRadius = Math.pow(10, camRadius);

// prevent zooming through the center:
if (camRadius < this.cameraNear) {
  camRadius = this.cameraNear;
}
if (camRadius > this.cameraFar) {
  camRadius = this.cameraFar;
}
```

I have provided sample code with these changes applied. You can check it here.
[improve zooming](https://editor.p5js.org/dark_fox/sketches/aVrbgFmL9)

inaridarkfox4231:
## Overview

Currently, scaling with webgl's orbitControl() results in jerky behavior.
This is because the judgment method is inappropriate. This is because the change in WheelDeltaY is detected when only the change in WheelY should be detected.
But the problem doesn't stop there. In spite of the fact that behavior that does not depend on the scale is required, the change in the distance from the gaze point is executed by addition and subtraction.
This pull request is to fix these issues.

Resolves #6104

 ## Changes:
First, set sensitivityZ to 1 and prepare a separate scale for zooming.
The decision should be made using the change in WheelY instead of the change in WheelDeltaY.
Furthermore, by changing the distance using a logarithm, the behavior is not affected by the size of the scale.
Finally, let's limit the distance using near and far clips. Because 0.1 doesn't make sense for some scales.

 ## Screenshots of the change:
sample code: [improve zooming](https://editor.p5js.org/dark_fox/sketches/aVrbgFmL9)

#### PR Checklist

- [x] `npm run lint` passes


inaridarkfox4231:
i don't know the cause... What is happened?

inaridarkfox4231:
If you upload a video, it will become heavy, so please check the difference in behavior with this tweet. It shows how the scaling behavior is improved by applying the patch.
[tweet sample](https://twitter.com/inaba_darkfox/status/1646918984720326659)

davepagurek (approved):
Thanks for making these changes, it definitely feels better

inaridarkfox4231:
thanks very much for merge! ('ω')
There is still room for improvement in orbitControl(), so I would like to propose some more in the future.

Qianqianye:
Thanks @inaridarkfox4231 and @davepagurek 🎉

PR #6142, opened 2023-05-16
Create images.json

SembeiNorimaki:
JSON file containing an example array of two images.  Needed for the example of preloading images from a JSON file

Resolves #[Add issue number here]

 Changes:

 Screenshots of the change:

#### PR Checklist

- [ ] `npm run lint` passes
- [ ] [Inline documentation] is included / updated
- [ ] [Unit tests] are included / updated

[Inline documentation]: https://github.com/processing/p5.js/blob/main/contributor_docs/inline_documentation.md
[Unit tests]: https://github.com/processing/p5.js/tree/main/contributor_docs#unit-tests

PR #6143, opened 2023-05-16
Update loading_displaying.js

SembeiNorimaki:
A useful example of using callbacks when preloading images.

Sometimes we don't directly have the names of the images, but those have to be loaded from a JSON file.

Loading a JSON file without callbacks will make the data not available immediately, therefore avoiding the preloading of images.

With this example, we can preload a JSON file and use the callback to then preload the list of images, so everything gets ready when the `setup()` function is called.

I find this example very useful and I think it should be included either here, or in the `preload()` documentation

 Changes:

 Screenshots of the change:

#### PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -61,6 +61,29 @@ import '../core/friendly_errors/fes_core';
  * }
  * </code>
  * </div>
+ * <div>
+ * <code>
+ * // JSON file containing a list of files to load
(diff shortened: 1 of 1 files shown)


limzykenneth:
Closing this as per comment in #6142. Please do open a new issue for this feature request for dicussions first. Thanks.

PR #6167, opened 2023-05-29
WebGL mode documentation improvements

wong-justin:
(This PR is a working draft that's part of the GSoC 2023 project 'Supporting shader-based filters in p5.js")

Initial changes:
- add basic summary of differences between P2D and WEBGL mode
- add links to WEBGL mode tutorial content

 Screenshots:
- Before: 

- After: 

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -22,8 +22,18 @@ export const VERSION =
  */
 export const P2D = 'p2d';
 /**
- * One of the two render modes in p5.js: P2D (default renderer) and WEBGL
- * Enables 3D render by introducing the third dimension: Z
+ * One of the two render modes in p5.js, most known for 3D rendering.
+ *
+ * Some differences and new features to expect compared to the default render mode <a href="/#/p5/P2D">P2D</a>:
+ * - The <strong>coordinate system</strong> is centered and uses an optional z-dimension
+ * - <strong>Textures</strong>, <strong>materials</strong>, and <strong>lighting</strong> can be applied to shapes
+ * - You can control the <strong>camera</strong> that views the scene
+ * - <strong>Text</strong> is drawn differently
+ * - You can use <strong>shaders</strong>
+ *
+ * To learn more about WEBGL mode, see:
+ * - <a href="https://github.com/processing/p5.js/wiki/Getting-started-with-WebGL-in-p5">Getting started with WebGL in p5</a>, a wiki page on the p5.js github
+ * - <a href="https://p5js.org/learn/#:~:text=Getting%20Started%20in%20WebGL">Getting Started in WebGL</a>, a series of tutorials in the p5.js "Learn" reference section
  * @property {String} WEBGL
  * @final
  */


src/core/constants.js
```diff
@@ -22,8 +22,18 @@ export const VERSION =
...
- * One of the two render modes in p5.js: P2D (default renderer) and WEBGL
- * Enables 3D render by introducing the third dimension: Z
+ * One of the two render modes in p5.js, most known for 3D rendering.
+ *
+ * Some differences and new features to expect compared to the default render mode <a href="/#/p5/P2D">P2D</a>:
+ * - The <strong>coordinate system</strong> is centered and uses an optional z-dimension
+ * - <strong>Textures</strong>, <strong>materials</strong>, and <strong>lighting</strong> can be applied to shapes
+ * - You can control the <strong>camera</strong> that views the scene
+ * - <strong>Text</strong> is drawn differently
+ * - You can use <strong>shaders</strong>
```
wong-justin, line 32: I was thinking about adding links to the bold features, like `<a href="https://github.com/processing/p5.js/wiki/Getting-started-with-WebGL-in-p5#text"><strong>Text</strong></a> is drawn differently`

davepagurek: Linking sounds good to me, for what it's worth!

src/core/constants.js
```diff
@@ -22,8 +22,18 @@ export const VERSION =
  */
 export const P2D = 'p2d';
 /**
- * One of the two render modes in p5.js: P2D (default renderer) and WEBGL
- * Enables 3D render by introducing the third dimension: Z
+ * One of the two render modes in p5.js, most known for 3D rendering.
+ *
+ * Some differences and new features to expect compared to the default render mode <a href="/#/p5/P2D">P2D</a>:
+ * - The <strong>coordinate system</strong> is centered and uses an optional z-dimension
```
davepagurek, line 28: Also, just a heads up, for now bullet lists in our docs render just as newlines in the reference. I don't think we have an issue to track this yet so I made an issue for it here: https://github.com/processing/p5.js/issues/6170

src/core/constants.js
aceslowman: I agree that it is helpful to link out to more info for each of those topics and I appreciate this as a brief list of differences. I think that replacing the bold text with links though might not be very clear, I wonder if things can be reworded in a way that lets you describe where the information can be found. Something like:

`The coordinate system is centered and uses an optional z-dimension, which you can read more about in the <a href="">"Getting Started With WebGL"</a> tutorial in the Learn page.`

Additionally, it would help to include a little more text for "Text is drawn differently" and "You can use shaders". For text, the reference page for [text()](https://p5js.org/reference/#/p5/text) describes some of the issues, maybe you can rephrase it into something like:

`Text in WebGL requires opentype/truetype fonts loaded in your [preload()](https://p5js.org/reference/#/p5/preload) function using the [loadFont()](https://p5js.org/reference/#/p5/loadFont) method. More information about using text in this mode can be found in the [wiki](https://github.com/processing/p5.js/wiki/Getting-started-with-WebGL-in-p5#text).`

And it might be helpful to briefly define shader, as some users might not be familiar:

`WebGL mode makes it possible to use shaders, which are programs that can be used for creating a variety of effects and graphics that take advantage of hardware acceleration.`

src/core/constants.js
```diff
@@ -22,8 +22,18 @@ export const VERSION =
  */
 export const P2D = 'p2d';
 /**
- * One of the two render modes in p5.js: P2D (default renderer) and WEBGL
- * Enables 3D render by introducing the third dimension: Z
+ * One of the two render modes in p5.js, most known for 3D rendering.
```
aferriss, line 25: A small nit but can we phrase this a little differently to reflect WEBGL's usage rather than it's notoriety. 

`One of the two render modes in p5.js, used for computationally intensive tasks like 3d rendering and drawing shaders.`

src/core/constants.js
```diff
@@ -22,8 +22,18 @@ export const VERSION =
...
+ *
+ * Some differences and new features to expect compared to the default render mode <a href="/#/p5/P2D">P2D</a>:
```
aferriss, line 27: Can we simplify the language here:
`WEBGL differs from the default P2D renderer in the following ways:`

src/core/constants.js
aferriss, line 28: Also echoing what Austin said. I think for this list it would be good to put the topic as the first word, and then follow it with a sentence or two about how it is different. For example:

**Coordinate System** - When drawing in WEBGL mode, the origin point (0,0,0) is located at the center of the screen. More information about WEBGL [coordinates and transformations can be found on the learn page](https://p5js.org/learn/getting-started-in-webgl-coords-and-transform.html)

**Next topic** - one or two sentences describing the differences with a link to more info if possible.

src/core/constants.js
aferriss, line 32: I think we can also add a couple more elements to this list.
**3D Shapes** - WEBGL mode can be used to draw 3D shapes like cubes, spheres, cones, and more. (with links to each shape). 

**Graphics Acceleration** - WEBGL mode uses the graphics card and may help you boost performance of your sketch or draw an increased number of shapes to the screen at once.

Any other major differences worth mentioning here?

src/core/constants.js
davepagurek: ok once this is merged https://github.com/processing/p5.js/pull/6177 you'll get bullets next to your list items!

wong-justin:
Thanks for all the feedback!
Updated content:

image

wong-justin (commented):
This one took me a while to realize. Maybe a comment should be left for future contributors to know that documentation will not be generated for p5.RendererGL.prototype definitions?

src/webgl/3d_primitives.js
aferriss, line 1: I think we can leave this developer documentation in place, despite the fact that it doesn't render into the actual docs. Adding a note that explains that these docs are not rendered would be good though!

wong-justin:
The past few commits are changes to drawing function docs, starting with `line()` and `arc()`. I'm trying to strike a balance between useful changes and consistency.

I also noticed that the season of docs contributor in [#6178](https://github.com/processing/p5.js/issues/6178) is making a style guide for clear and consistent docs and snippets, and he'll be starting with 2D primitives too. So I feel like I should only make a few priority changes for now.

src/core/shape/2d_primitives.js
```diff
@@ -375,6 +375,17 @@ p5.prototype._renderEllipse = function(x, y, w, h, detailX) {
...
+ * <div>
+ * <code>
+ * createCanvas(100, 100, WEBGL);
+ * strokeWeight(7);
+ * line(0, 0, 0, 30, 0, 0);
+ * line(0, 0, 0, 0, 30, 0);
+ * line(-30, -30, -30, 30, 30, 30);
+ * describe('two lines mark the flat x and y axes while a third line spans diagonally through the third dimension');
+ * </code>
+ * </div>
```
wong-justin, line 396: The other code examples only used the x and y parameters for [line()](https://p5js.org/reference/#/p5/line), so I added one using the z parameters for WEBGL mode. 

I figure it's a good goal to show all the possible parameters in the code examples, which would mean having a 3D example for the 2D primitives.

Having these scaffolded examples might naturally help users transition to WEBGL. Eg. they might know the WEBGL coordinate system is different just from their time browsing 2D primitives examples.

aferriss (commented):
Thanks for making the changes @wong-justin I think things are looking good! 

Given that some doc style changes are around the corner, maybe it makes sense to separate out the code examples for webGL shape functions into a separate PR that can be updated to match the new style when / if those changes are made. 

On the one hand I don't want to get blocked by that work, but also it would be a shame to have to make these changes once now and then once again when new guidance is available. What do you think @aceslowman ?

src/core/shape/2d_primitives.js
aferriss: Agree, it's nice to see the webGL versions with the updated signatures! If we don't think it's too much code I think it could be nice to color these lines in rgb (xyz) gizmo fashion like so https://editor.p5js.org/aferriss/sketches/qCsZ08lJ8

src/core/shape/2d_primitives.js
```diff
@@ -172,6 +172,15 @@ p5.prototype._normalizeArcAngles = (
...
  *
+ * <div>
+ * <code>
+ * createCanvas(100, 100, WEBGL);
+ * arc(0, 0, 80, 80, 0, PI + QUARTER_PI, PIE, 7);
+ * describe(
+ *   'ellipse with a quarter missing, and the outline is segmented instead of curved'
+ * );
+ * </code>
+ * </div>
```
wong-justin, line 183: The round 2D primitives have a `detail` parameter, so technically for completeness they could use an example like this to show it in action.

But I think it would distract from useful parameters and examples, and `detail` doesn't seem like a popular feature for users, and `detail` is already documented well in `curveDetail()` and `bezierDetail()` and the 3D primitives.

Maybe this should make it into the list of WEBGL vs P2D differences? Something like:
- **Vertex Count** - When drawing curved shapes in `WEBGL` mode, you can specify how many vertices are drawn with a `detail` parameter.

aferriss: Instead of `Vertex Count` I think we can use p5's existing language and call it `Shape Detail`. 

Also I don't think additional examples detracts and to be honest I didn't realize the round 2d shapes even had that parameter. I think it would be a good addition, and as you note, the 3d shapes include this, so why not 2d as well?

https://p5js.org/reference/#/p5/sphere

aceslowman:
> Thanks for making the changes @wong-justin I think things are looking good!
> 
> Given that some doc style changes are around the corner, maybe it makes sense to separate out the code examples for webGL shape functions into a separate PR that can be updated to match the new style when / if those changes are made.
> 
> On the one hand I don't want to get blocked by that work, but also it would be a shame to have to make these changes once now and then once again when new guidance is available. What do you think @aceslowman ?

@aferriss @wong-justin I think separating it out into a new PR would be helpful, it could also be a good opportunity to communicate a bit with the style docs contributor. Personally I think continuing with doc changes makes sense, even if there is some revision needed down the line.

Also, I think that the text changes made on the WEBGL reference page are a great improvement, I don't feel like it's excessively wordy and it does a good job at linking out to further reading.

nickmcintyre:
@wong-justin @aferriss @aceslowman the changes you've made thus far are great improvements. I won't begin editing WebGL docs until August, so please go ahead with your work. I'll share a draft of the style guide later this week and will ping you for feedback.

Here is the draft **Wording** section:

---

Write simple, declarative sentences. Brevity is a plus: get to the point.

Write in present tense: "Returns an object that...", rather than "Returned an object that..." or "Will return an object that...".

Start comments in upper case. Follow regular punctuation rules:

```javascript
// Draws a fractal from a Julia set.
function drawFractal(c, rad, maxIter) {
  // ...
}
```
Communicate to the reader the current way of doing things, both explicitly and implicitly. Use the idioms recommended in this guide. Reorder code samples to emphasize favored approaches if needed. The documentation should be a model for best practices and approachable for beginners.

Documentation has to be brief but comprehensive. Explore and document edge cases. What happens for the each combination of arguments? What bugs are most likely to appear in a beginner's code?

Spell names correctly: p5.js, CSS, HTML, JavaScript, WebGL. When in doubt, please have a look at some authoritative source like their official documentation.

---

In the docs, I propose we stick with WebGL unless you're referencing the `WEBGL` constant.

How should we incorporate [canvas descriptions](https://p5js.org/contributor-docs/#/web_accessibility?id=user-generated-accessible-canvas-descriptions)?

PR #6169, opened 2023-05-30
Improvement of orbitControl(): Behavior when going out of canvas during operation

### LINKED ISSUE #6168: Improve the behavior of orbitControl(): when the mouse or touch pointer is outside the canvas
### Increasing Access

Currently, orbitControl() is not executed when operating outside the canvas.

```js
  // If the mouse is not in bounds of the canvas, disable all behaviors:
  const mouseInCanvas =
    this.mouseX < this.width &&
    this.mouseX > 0 &&
    this.mouseY < this.height &&
    this.mouseY > 0;
  if (!mouseInCanvas) return;
```
It's a natural constraint because you have to do other things outside the canvas.
However, it is a little inconvenient to completely prohibit the behavior, because the operation will stop even if it goes out of the canvas while operating in the canvas.

With EasyCam, rotation and moving will continue even if you go outside while operating inside the canvas.

https://github.com/processing/p5.js/assets/39549290/3e71ec39-2ce7-4913-9e1c-26289140a148

Even in vRoidHub, which I use as a reference for behavior, the behavior continues even if I go outside during operation.

https://github.com/processing/p5.js/assets/39549290/a1d0f5c6-e440-4347-934c-2277abab0aba

on the other hand...

https://github.com/processing/p5.js/assets/39549290/a20ef09c-8070-42df-997f-b0a2d8377e57

So I would like to implement this specification.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [X] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

First, prepare a flag to manage whether the interaction is ongoing.

```js
// Flags for recording the state of zooming, rotation and moving
this._renderer.executeRotateAndMove = false;
this._renderer.executeZoom = false;
```
Then remove the code regarding mouseInCanvas. Instead, prepare a flag to check if the mouse or touch pointer is inside the canvas.

```js
// Flag whether the mouse or touch pointer is inside the canvas
let pointersInCanvas = false;
```

Touches are calculated separately for each case according to the number of pointers. Then, when the touch starts inside the canvas, the process starts, and when the touch ends outside the canvas, the process ends.

```js
    /* for touch */
    // if length === 1, rotate
    // if length > 1, zoom and move
    if (movedTouches.length === 1) {
      const t = movedTouches[0];
      deltaTheta = -sensitivityX * (t.x - t.px) / scaleFactor;
      deltaPhi = sensitivityY * (t.y - t.py) / scaleFactor;
      // For one point, calculate only for that point.
      pointersInCanvas =
        (t.x > 0 && t.x < this.width && t.y > 0 && t.y < this.height);
    } else {
      const t0 = movedTouches[0];
      const t1 = movedTouches[1];
      const distWithTouches = Math.hypot(t0.x - t1.x, t0.y - t1.y);
      const prevDistWithTouches = Math.hypot(t0.px - t1.px, t0.py - t1.py);
      const changeDist = distWithTouches - prevDistWithTouches;
      // move the camera farther when the distance between the two touch points
      // decreases, move the camera closer when it increases.
      deltaRadius = -changeDist * sensitivityZ * touchZoomScaleFactor;
      // Move the center of the camera along with the movement of
      // the center of gravity of the two touch points.
      moveDeltaX = 0.5 * (t0.x + t1.x) - 0.5 * (t0.px + t1.px);
      moveDeltaY = 0.5 * (t0.y + t1.y) - 0.5 * (t0.py + t1.py);
      // If there are no less than two points, calculate the first two points.
      pointersInCanvas =
        (t0.x > 0 && t0.x < this.width && t0.y > 0 && t0.y < this.height) &&
        (t1.x > 0 && t1.x < this.width && t1.y > 0 && t1.y < this.height);
    }
    // Initiate an interaction if touched in the canvas
    if (this.touches.length > 0 && pointersInCanvas) {
      this._renderer.executeRotateAndMove = true;
      this._renderer.executeZoom = true;
    }
    // End an interaction when the touch is released while outside the canvas.
    if (this.touches.length === 0 && !pointersInCanvas) {
      this._renderer.executeRotateAndMove = false;
      this._renderer.executeZoom = false;
    }
```

For the mouse, it's calculated the same as the traditional mouseInCanvas. This also starts processing when the mouse down occurs in the canvas.
Regarding the end of processing, it will end when the mouse is released, whether inside or outside the canvas.

```js
    /* for mouse */
    // if wheelDeltaY !== 0, zoom
    // if mouseLeftButton is down, rotate
    // if mouseRightButton is down, move

    // For mouse, it is calculated based on the mouse position.    
    pointersInCanvas =
      (this.mouseX > 0 && this.mouseX < this.width) &&
      (this.mouseY > 0 && this.mouseY < this.height);

      // zoom the camera depending on the value of _mouseWheelDeltaY.
      // move away if positive, move closer if negative
      deltaRadius = this._mouseWheelDeltaY * sensitivityZ;
      deltaRadius *= mouseZoomScaleFactor;
      this._mouseWheelDeltaY = 0;
      // start zoom when the mouse is wheeled within the canvas.
      if (pointersInCanvas) this._renderer.executeZoom = true;
    } else {
      // quit zoom when you stop wheeling.
      this._renderer.zoomFlag = false;
    }
    if (this.mouseIsPressed) {
      if (this.mouseButton === this.LEFT) {
        deltaTheta = -sensitivityX * (this.mouseX - this.pmouseX) / scaleFactor;
        deltaPhi = sensitivityY * (this.mouseY - this.pmouseY) / scaleFactor;
        moveDeltaX = this.mouseX - this.pmouseX;
        moveDeltaY = this.mouseY - this.pmouseY;
      }
      // start rotate and move when mouse is pressed within the canvas.
      if (pointersInCanvas) this._renderer.executeRotateAndMove = true;
    } else {
      // quit rotate and move if mouse is released.
      this._renderer.executeRotateAndMove = false;
    }
```

All acceleration processes will only run if these flags are true. For example, for rotation processing:

```js
  // rotate process
  if ((deltaTheta !== 0 || deltaPhi !== 0) &&
  this._renderer.executeRotateAndMove) {
    // accelerate rotate velocity
    this._renderer.rotateVelocity.add(
      deltaTheta * rotateAccelerationFactor,
      deltaPhi * rotateAccelerationFactor,
      0
    );
  }
```

It is necessary to manage the state with flags. Otherwise, it would be impossible to distinguish between starting an interaction inside and going outside, and trying to start an interaction outside. The latter should prevent interaction.

In fact, you can use addEventListener to set the start of the event to canvas and the end to window, but remodeling p5.js's orbitControl() in such a framework is costly I can't because it's too big. Instead, it is handled by having a flag in the renderer.

[orbitControl_outsideBehavior](https://openprocessing.org/sketch/1941568)

https://github.com/processing/p5.js/assets/39549290/18536629-8592-4cb9-964a-7b7240cd394c

I would like to create a pull request with this content.

inaridarkfox4231:
The current orbitControl() is designed to stop processing suddenly if it goes out of the canvas while operating inside the canvas.

I don't think it's a very clean behavior, so I'm trying to improve this.

Resolves #6168

## Changes:
Stop using mouseInCanvas and instead use pointersInCanvas, which indicates whether the mouse or touch pointer is inside the canvas.
Use this flag to start processing when you interact within the canvas.
Prepare some flags for p5.Renderer so that even if you go outside the canvas during the interaction, the processing will not end as long as the interaction continues.

### Current behavior

https://github.com/processing/p5.js/assets/39549290/3959d5b4-999b-4283-b0f1-c0818553fa2e

### Expect behavior
OpenProcessing: [orbitControl_outsideBehavior](https://openprocessing.org/sketch/1941568)
p5Editor: [orbitControl_outsideBehavior](https://editor.p5js.org/dark_fox/sketches/-4WRlDrSf)

https://github.com/processing/p5.js/assets/39549290/aff115bb-bf2a-4c9e-a900-9951e86cc548

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/interaction.js
+++ b/src/webgl/interaction.js
@@ -191,7 +191,7 @@ p5.prototype.orbitControl = function(
     // if mouseLeftButton is down, rotate
     // if mouseRightButton is down, move
 
-    // For mouse, it is calculated based on the mouse position.    
+    // For mouse, it is calculated based on the mouse position.
     pointersInCanvas =
       (this.mouseX > 0 && this.mouseX < this.width) &&
       (this.mouseY > 0 && this.mouseY < this.height);
--- a/src/webgl/interaction.js
+++ b/src/webgl/interaction.js
@@ -54,7 +54,7 @@ import * as constants from '../core/constants';
  */
 
 // implementation based on three.js 'orbitControls':
-// https://github.com/mrdoob/three.js/blob/dev/examples/js/controls/OrbitControls.js
+// https://github.com/mrdoob/three.js/blob/6afb8595c0bf8b2e72818e42b64e6fe22707d896/examples/jsm/controls/OrbitControls.js#L22
 p5.prototype.orbitControl = function(
   sensitivityX,
   sensitivityY,
@@ -150,13 +150,16 @@ p5.prototype.orbitControl = function(
     /* for touch */
     // if length === 1, rotate
     // if length > 1, zoom and move
+
+    // for touch, it is calculated based on one moved touch pointer position.
+    pointersInCanvas =
+      movedTouches[0].x > 0 && movedTouches[0].x < this.width &&
+      movedTouches[0].y > 0 && movedTouches[0].y < this.height;
+
     if (movedTouches.length === 1) {
       const t = movedTouches[0];
       deltaTheta = -sensitivityX * (t.x - t.px) / scaleFactor;
       deltaPhi = sensitivityY * (t.y - t.py) / scaleFactor;
-      // For one point, calculate only for that point.
-      pointersInCanvas =
-        (t.x > 0 && t.x < this.width && t.y > 0 && t.y < this.height);
     } else {
       const t0 = movedTouches[0];
       const t1 = movedTouches[1];
@@ -170,18 +173,15 @@ p5.prototype.orbitControl = function(
       // the center of gravity of the two touch points.
       moveDeltaX = 0.5 * (t0.x + t1.x) - 0.5 * (t0.px + t1.px);
       moveDeltaY = 0.5 * (t0.y + t1.y) - 0.5 * (t0.py + t1.py);
-      // If there are no less than two points, calculate the first two points.
-      pointersInCanvas =
-        (t0.x > 0 && t0.x < this.width && t0.y > 0 && t0.y < this.height) &&
-        (t1.x > 0 && t1.x < this.width && t1.y > 0 && t1.y < this.height);
     }
-    // Initiate an interaction if touched in the canvas
-    if (this.touches.length > 0 && pointersInCanvas) {
-      this._renderer.executeRotateAndMove = true;
-      this._renderer.executeZoom = true;
-    }
-    // End an interaction when the touch is released while outside the canvas.
-    if (this.touches.length === 0 && !pointersInCanvas) {
+    if (this.touches.length > 0) {
+      if (pointersInCanvas) {
+        // Initiate an interaction if touched in the canvas
+        this._renderer.executeRotateAndMove = true;
+        this._renderer.executeZoom = true;
+      }
+    } else {
+      // End an interaction when the touch is released
       this._renderer.executeRotateAndMove = false;
       this._renderer.executeZoom = false;
     }
@@ -260,7 +260,7 @@ p5.prototype.orbitControl = function(
       0
     );
   }
-  if (this._renderer.rotateVelocity.mag() > 0.001) {
+  if (this._renderer.rotateVelocity.magSq() > 0.000001) {
     this._renderer._curCamera._orbit(
       this._renderer.rotateVelocity.x,
       this._renderer.rotateVelocity.y,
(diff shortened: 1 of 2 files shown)


inaridarkfox4231:
In the case of touch, there seems to be no problem with only check one point, so it may be corrected later.

inaridarkfox4231:
It may not need to check pointersInCanvas for the quit  conditions for touch interaction, too.

inaridarkfox4231:
Only one touch pointer needs to be examined, so I rewrote it that way.

Also, we removed the in-canvas condition from the end condition of the touch interaction, as it is unnecessary.

I was using mag() to check whether it is positive value or not, but magSq() is ok and faster, so I rewrote it.

Lastly, the link on the Three.js site that was mentioned as a reference for orbitControl() was broken, so I rewrote it with an unbroken link. I didn't actually refer to this, but it's not good to have a broken link, so I put up a link that exists.

davepagurek (approved):
Thanks for updating this, I didn't notice when testing on fullscreen canvases but this definitely feels better on little canvases!

inaridarkfox4231:
Thank you for merge!('ω') The sample sketch exists in reference is so small, so I wanted to avoid frequent stop rotating and moving.

PR #6173, opened 2023-05-31
convert some missing class.

### LINKED ISSUE #3758: ES6 Transition: Concerns & Progress

wuyudi:
Resolves #3758

More class to es6, this is a continue to https://github.com/processing/p5.js/pull/6075

 Changes:

I use `p5\..*?.prototype\..*?` to find them.

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/3d_primitives.js
+++ b/src/webgl/3d_primitives.js
@@ -174,8 +174,8 @@ p5.prototype.box = function(width, height, depth, detailX, detailY) {
         [20, 21],
         [22, 23]
       ];
-      for (let i = 0; i < cubeIndices.length; i++) {
-        const cubeIndex = cubeIndices[i];
+
+      cubeIndices.forEach((cubeIndex, i) => {
         const v = i * 4;
         for (let j = 0; j < 4; j++) {
           const d = cubeIndex[j];
@@ -192,7 +192,7 @@ p5.prototype.box = function(width, height, depth, detailX, detailY) {
         }
         this.faces.push([v, v + 1, v + 2]);
         this.faces.push([v + 2, v + 1, v + 3]);
-      }
+      });
     };
     const boxGeom = new p5.Geometry(detailX, detailY, _box);
     boxGeom.computeNormals();
--- a/src/accessibility/gridOutput.js
+++ b/src/accessibility/gridOutput.js
@@ -49,10 +49,7 @@ function _gridMap(idT, ingredients) {
   let shapeNumber = 0;
   let table = '';
   //create an array of arrays 10*10 of empty cells
-  let cells = Array(...Array(10)).map(function() {});
-  for (let r in cells) {
-    cells[r] = Array(...Array(10)).map(function() {});
-  }
+  let cells = Array.from(Array(10), () => Array(10));
   for (let x in ingredients) {
     for (let y in ingredients[x]) {
       let fill;
--- a/src/accessibility/gridOutput.js
+++ b/src/accessibility/gridOutput.js
@@ -49,9 +49,9 @@ function _gridMap(idT, ingredients) {
   let shapeNumber = 0;
   let table = '';
   //create an array of arrays 10*10 of empty cells
-  let cells = Array.apply(null, Array(10)).map(function() {});
+  let cells = Array(...Array(10)).map(function() {});
   for (let r in cells) {
-    cells[r] = Array.apply(null, Array(10)).map(function() {});
+    cells[r] = Array(...Array(10)).map(function() {});
   }
   for (let x in ingredients) {
     for (let y in ingredients[x]) {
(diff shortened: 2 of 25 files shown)


src/webgl/3d_primitives.js
```diff
@@ -174,8 +174,8 @@ p5.prototype.box = function(width, height, depth, detailX, detailY) {
         [20, 21],
         [22, 23]
       ];
-      for (let i = 0; i < cubeIndices.length; i++) {
-        const cubeIndex = cubeIndices[i];
+
+      cubeIndices.forEach((cubeIndex, i) => {
```
limzykenneth, line 178: Not sure if it is still true or relevant but `forEach` tends to be slower than pure for loops.

wuyudi: https://measurethat.net/Benchmarks/Show/2090/0/array-loop-vs-foreach-vs-map#latest_results_block it seems that forEach now is faster.

limzykenneth: Just tried that benchmark on my computer and for Chrome foreach is faster but for Firefox for loops are about 5-6x faster. A bit surprised Chrome is optimized for foreach though. 🤔
Screenshot 2023-06-01 at 16 11 30

wuyudi:
[Why there is a remaining `p5.RendererGL.prototype.tessyVertexSize = 12;`](https://stackoverflow.com/questions/34185643/ecmascript-6-syntax-for-declaring-prototype-default-values-using-class)

wuyudi:
I think now it's ok to merge :eyes:

jeffscottward:
I'm late to the conversion party but just wondering why use a classes?

Definitely agree we have to get away from prototype but at least in the react.js (react team moved away after using it) and node.js worlds, classes are frowned upon. It's a jarring context switch for a language that is designed to be functional.

Is using classes required for this switch away from prototype?

davepagurek:
> Is using classes required for this switch away from prototype?

I would say that it's not required, but is the easiest change, since the previous code was already using classes but just using an ES5 syntax for it, so we could change to `class` syntax without updating any call sites.

jeffscottward:
> > Is using classes required for this switch away from prototype?
> 
> 
> 
> I would say that it's not required, but is the easiest change, since the previous code was already using classes but just using an ES5 syntax for it, so we could change to `class` syntax without updating any call sites.

Oh that is smart actually. Ok fair. :) I might try to throw AI at this and see if it can blaze through.

jeffscottward:
> > > Is using classes required for this switch away from prototype?
> > 
> > 
> > I would say that it's not required, but is the easiest change, since the previous code was already using classes but just using an ES5 syntax for it, so we could change to `class` syntax without updating any call sites.
> 
> Oh that is smart actually. Ok fair. :) I might try to throw AI at this and see if it can blaze through.

Uh oh..
https://github.com/processing/p5.js/blob/main/contributor_docs/archive/es6-adoption.md#browser-compatibility-and-transpiling

http://incaseofstairs.com/six-speed/

Assuming we are using Babel - Classes is a *massive* perf hit. 8x? wow.
What about typescript?

Also - based on Contrib docs - seems like a master Issue for ES6 conversion is missing? These one of PR's leave the conversation sporadic in closed issue comments. Should I make one?

jeffscottward:
🚨 🚨  Please move ES6 conversion commentary over to https://github.com/processing/p5.js/issues/6371 🚨 🚨

PR #6183, opened 2023-06-06
Changed em tags (asterisks) to multiplication symbols

Acha0203:
Changes:
Changed em tags to multiplication symbols in src/core/transform.js

before:
@param  {Array} arr an array of numbers - should be 6 or 16 length (2\*3 or 4\*4 matrix values)

after:
@param  {Array} arr an array of numbers - should be 6 or 16 length (2×3 or 4×4 matrix values)
 
#### PR Checklist

- [x ] `npm run lint` passes
- [x ] [Inline documentation] is included / updated
- [x ] [Unit tests] are included / updated

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3983,7 +3983,7 @@
     "node_modules/boom": {
       "version": "0.4.2",
       "resolved": "https://registry.npmjs.org/boom/-/boom-0.4.2.tgz",
-      "integrity": "sha1-emNune1O/O+xnO9JR6PGffrukRs=",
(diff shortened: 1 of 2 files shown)


inaridarkfox4231:
Thank you for noticing mistake!
That's right, but there are other mistakes.

エックスとばつ

```js
 /**
 * @param  {Number} a numbers which define the 2×3 or 4x4 matrix to be multiplied
 */
```
As you can see from this, the multiplication symbol "×" and the alphabet "x" are mixed. So I think you should fix this as well. 
I think the multiplication symbol is more appropriate.

Acha0203:
I have changed the characters of x to multiplication symbols. Thank you!

Acha0203:
I have entered the command you said and committed the results. If you have any problems or questions, please let me know. Thank you!

limzykenneth:
Looks good. Thanks!

PR #6187, opened 2023-06-07
fix missing extends

wuyudi:
Resolves https://github.com/p5-types/p5.ts/issues/27

 Changes:

fix missing extends in class doc.

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -2449,8 +2449,9 @@ class Cue {
  * @class p5.MediaElement
  * @constructor
  * @param {String} elt DOM node that is wrapped
+ * @extends p5.Element
  */
-p5.MediaElement = class MediaElement extends p5.Element {
+class MediaElement extends p5.Element {
   constructor(elt, pInst) {
     super(elt, pInst);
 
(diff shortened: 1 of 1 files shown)


src/dom/dom.js
```diff
@@ -3609,7 +3610,9 @@ p5.MediaElement = class MediaElement extends p5.Element {
 
     this._prevTime = playbackTime;
   }
-};
+}
+
+p5.MediaElement = MediaElement;
```
limzykenneth, line 3615: Might be just me but I don't quite understand what this particular syntax is doing compared to how it was before, can you perhaps explain a bit here?

wuyudi: I read the p5.sound 's code, then I found that it's organized like this.

Pros: better hint for the editor. Since now all the `p5.class` becomes any, but the `class` itself has proper hints.
Cons: Need to type the name more times.

the effect is the same.

wuyudi:
@limzykenneth I thought this could be merged now.

PR #6216, opened 2023-06-16
Created fix for camera settings getting reset each time resizeCanvas() is called

### LINKED ISSUE #6203: resizeCanvas resets the camera

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

### p5.js version

1.6.0

### Web browser and version

Chrome 114.0.5735.110 (Official Build) (64-bit)

### Operating System

Windows 10

### Steps:
1. Create a camera object
2. Resize the canvas

### Snippet:

```js

function windowResized() {
}

let cam = createCamera(); // create a new camera instance with custom properties and then resize the window and it'll reset multiple of the camera's attributes like its position & rotation

```

RandomGamingDev:
Created fix for camera settings getting reset each time resizeCanvas() is called

Resolves #6203.

 Changes:

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes


RandomGamingDev:
This is the solution for issue https://github.com/processing/p5.js/issues/6203.

davepagurek:
Thanks for taking this on!

Do you think it's equivalent to take out the else branch here? https://github.com/processing/p5.js/blob/d2b5d5e153c9b2a66b7cb80f2c41aecb350d2ac2/src/webgl/p5.Camera.js#L1615-L1626

If it is equivalent, it might be better to make the code change there to avoid doing some work then undoing it

davepagurek:
Also, just a heads up, looks like we get some linter errors currently:

image

RandomGamingDev:
About the resize function, tbh I'm not very sure, but they don't look like they do the same things is all of I have to say ig lol.
About the linting thing sorry about that. I made sure that the code was of the same style, but wasn't able to check directly since lint was complaining about the different line endings.

inaridarkfox4231:
The way you do it, you'll run into problems using orbitControl() in this way.

sample:[resizeCanvas_BUGFIX](https://editor.p5js.org/dark_fox/sketches/fSy118Gro)


```js
function setup() {
  //ortho(-width/2,width/2,-height/2,height/2,0.1,4000);
}

function draw() {
  background(0);
  orbitControl();
  box(100);
}

function windowResized() {
}
```

https://github.com/processing/p5.js/assets/39549290/9aee6965-b556-48a0-853b-4ae56d35ba61

Immediately after resizing, there seems to be no effect, but when orbitControl() is executed, the camera returns to the default.
However, when performing the orbitControl() interaction, the camera changes based on its default state.
This is because _resize() is performed on _curCamera inside resizeCanvas. At that time, the eyeX and eyeY information will return to the default.

```js
p5.RendererGL.prototype.resize = function (w, h) {
  _main.default.Renderer.prototype.resize.call(this, w, h);
  this.GL.viewport(0, 0, this.GL.drawingBufferWidth, this.GL.drawingBufferHeight);
  this._viewport = this.GL.getParameter(this.GL.VIEWPORT);

  this._curCamera._resize(); //resize pixels buffer

  var pixelsState = this._pixelsState;
  if (typeof pixelsState.pixels !== 'undefined') {
    pixelsState._setProperty('pixels', new Uint8Array(this.GL.drawingBufferWidth * this.GL.drawingBufferHeight * 4));
  }
};
```

In orbitControl(), the camera is moved based on information such as eyeX and eyeY, so even if you change only the matrix, it will be overwritten with information based on eyeX and eyeY.

However, writing code to copy eyeX and eyeY one by one is troublesome. So, in this case, I think you should copy the camera and substitute it as it is.

```js
p5.prototype.resizeCanvas2 = function(w, h, noRedraw) {
  p5._validateParameters('resizeCanvas', arguments);
  if (this._renderer) {
    // save canvas properties
    const props = {};
    for (const key in this.drawingContext) {
      const val = this.drawingContext[key];
      if (typeof val !== 'object' && typeof val !== 'function') {
        props[key] = val;
      }
    }

    // save camera
    const cam = this._renderer._curCamera.copy();
    
    this._renderer.resize(w, h);
    this.width = w;

    // restore camera.
    this._renderer._curCamera = cam;
    this._renderer._curCamera.projMatrix.set(this._renderer.uPMatrix);

    // reset canvas properties
    for (const savedKey in props) {
      try {
        this.drawingContext[savedKey] = props[savedKey];
      } catch (err) {
        // ignore read-only property errors
      }
    }
    if (!noRedraw) {
      this.redraw();
    }
  }
  //accessible Outputs
  if (this._addAccsOutput()) {
    this._updateAccsOutput();
  }
};
```

Now running orbitControl() will not make the camera the default.

https://github.com/processing/p5.js/assets/39549290/cd73969e-0fbc-4cfa-85a7-099dd904ed3b

Note that this code I proposed defaults the projection matrix to the camera. I think this is debatable.

The projection matrix and the camera matrix have different roles. It may be possible to use the same camera matrix, but that is not the case for the projection matrix. Because there is an aspect ratio problem.
If the projection matrix remains the same when changing the canvas size, the screen will be distorted because the aspect ratio cannot be handled in this way.

```js
function setup() {
  ortho(-width/2,width/2,-height/2,height/2,0.1,4000);
}

function draw() {
  background(0);
  orbitControl();
  box(100);
}

function windowResized() {
}
```

https://github.com/processing/p5.js/assets/39549290/36101337-5a5e-458d-9596-5b5f95c9153a

I don't think there's anything I can do about this, so I think the only way to deal with it is to write code that sets the projection matrix immediately after resizeCanvas.

RandomGamingDev:
I combined my previous work with your advice [inaridarkfox4231](https://github.com/inaridarkfox4231) and now `orbitControl()` seems to work fine.

inaridarkfox4231:
However, multiplying the 0th and 5th elements of the projection matrix by the appropriate ratio values ​​might work around this problem. This is because this component of the projection matrix contains information about the aspect ratio of the screen.

inaridarkfox4231:
'custom' is added only for ortho and frustum, in which case the 0th and 5th elements of the projection matrix reflect the width and height, respectively. So, if you copy the projection matrix exactly as it is before resizing, and then multiply these components by the appropriate ratio values, you should be able to prevent the screen from being distorted.

But, of course, this doesn't apply to the camera as a global variable, so it's all about the renderer's camera. Perhaps, after giving the camera width and height information, let setCamera() check whether resizing has been performed, and if so, write code to change the projection matrix appropriately It might work. But I don't know if it's a backwards compatible change.

RandomGamingDev:
`custom` is also applied for `perspective()`. Turns out I made a typo and forgot to use `!==` instead of `===` so sorry about that part. I wanted to add the part where it only updates the projection matrix if the camera is "custom" (aka has had its projection matrix manually modified by some p5.js function) since [davepagruek](https://github.com/davepagurek) recommended that I add it so that the camera, if not of type custom, will continue to dynamically update to match the screen dimensions.

inaridarkfox4231:
I see, I didn't know perspective camera's cameraType can be also 'custom', sorry. I think that's better.
I'll leave the rest up to you and the reviewer.
I was concerned about the behavior of orbitControl() when resizeCanvas(), that's why I wanted to give my opinion.

RandomGamingDev:
lol dw it's fine, I didn't realize that there was a problem with `orbitControl()` since I typically just deal with the matrices and didn't see the other values getting interacted with much. I should've probably thought about cases like that.

davepagurek:
Just thought of a few more things to consider, one is that Framebuffers also have a default camera that gets resized here, so we probably want to implement this in both spots:

https://github.com/processing/p5.js/blob/63bea68eb3af40d3e6f450c03bd76bdc2442be31/src/webgl/p5.Framebuffer.js#L677-L685

Also, if we copy the camera and set `this._renderer._curCamera = cam` with the copy, users who have stored their camera in a local variable might be confused when editing their camera no longer affects how their sketch rendered.

> About the resize function, tbh I'm not very sure, but they don't look like they do the same things is all of I have to say ig lol.

The `p5.Camera` `resize` method is the one responsible for overwriting the properties we want to persist, so I was thinking maybe we can edit that to not overwrite them? i.e. just update the aspect ratio in the projection matrix? Just a suggestion though.

One other thing, it would be great to add a unit test -- maybe something like, make a canvas, set `frustum()`, then resize it, and get the pixels; and then make a second canvas already at that final size, set `frustum()` with the same parameters, and assert that the pixels are the same between the two.

RandomGamingDev:
tbh I didn't know that p5.Framebuffer was part of the p5.js library lol. I'll implement smth for that when I can. Also yeah, I sorta forgot about the outside reference to camera part. Originally I only copied the matrices because of that, but ig I just forgot. I'll make it so that it just copies the attributes later. I think I'll probably just create a class and iterate over each of the key value pairs in order to write them. Their attributes like the matrices however would have to have their references replaced which would be a bit annoying if someone were to have cached those tho lol, altho at this point basically unavoidable. I'll try to write smth for the unit test too when I can.

davepagurek:
I think it's ok to replace the matrix properties, we also haven't publicly documented their existence, so I think it's fair game to modify that without it causing a breaking change 🙂

inaridarkfox4231:
Indeed, when dealing with pan() and tilt(), it may be a problem if the local object reference is rewritten. I'm sorry.

RandomGamingDev:
lol it's fine. also yeah ig it'd probably be fine since they rn't documented.

RandomGamingDev:
I'll be a bit busy, but I'll add those to the pull request when I can.

inaridarkfox4231:
I would like to mention a few points.

First of all, resizeCanvas() is used in both 2D and webgl, but the problem now is the behavior of the camera, so if you want to change the specifications, I feel that p5.RendererGL's resize() is better (in that case, there is no need to check if _curCamera is undefined).

Next, if it is clear which issue you are dealing with, I think it would be better to put the number properly instead of leaving it as "Add issue number here" (not compulsory, but that I think it's easier for other contributors to participate in this pull request).

Then I'll explain why I wanted to copy the camera beforehand. That's because the use case I envisioned was a situation where only the default camera was used, and I wanted the results of moving with orbitControl() to be continuous before and after.

I thought it would be convenient if this specification change could be realized that, so I tried to express my opinion, but to be honest, I think it would be better to prepare a global camera and write such processing on my own, I feel. Because it's more flexible.

With the method of copying in advance, there was an opinion that if you are using a global camera, the processing for that camera will not be reflected. That's true, but when using the global camera, the user can access the camera directly, unlike the default camera, so I think you should write your own resizing process. This is because it is up to the user to decide what to do when resizing. I feel like there isn't much to decide on this side. Also, even if you are dealing with multiple cameras, only _curCamera can decide what to process on this side.

This pull request is certainly based on issue #6203, but the issue does not clearly state what kind of use case and what is the specific problem, so the direction of the discussion is unclear, I feel. But the issue and the pull request were filed by the same person, so there's no need to go back to the issue here. I think it would be good for the proponent to raise a specific use case and think about what kind of specification changes can be made to solve the problem that occurs when resizing the canvas.

Then it's better to consider whether the spec change will affect other use cases, such as the situation where multiple global cameras are used, or the case of p5.Framebuffer, I think.

I hope that whatever changes are made in the end will be productive.

RandomGamingDev:
The problem right now is indeed about the behavior of the camera, but the problem with the camera is the fact that `resizeCanvas()` not only resizes the canvas, but also the camera and is the issue that's trying to be addressed here although yes, we'd probably want this to other resize functions that reset the camera too.
Also yes, `resizeCanvas()` is in both 2D and WebGL mode, which is why we're checking for whether the camera's frustum/projection matrix has been previously modified so that we can keep compatibility with the traditional canvas.
Yes, it'd be more flexible, but for someone who hasn't messed around with p5.js before and especially for someone new to programming as a whole. The camera resetting for no reason in a way that requires your own custom fix they'd probably be at least decently annoyed and might just quit the project, since the whole idea of implementing some code separate from your own code that interacts with the p5.js code that interacts with the graphics goes against the idea of abstraction in a way that makes it easy for everyone.
The use case as displayed in issue [#6203](https://github.com/processing/p5.js/issues/6203) is just `resizeCanvas()` resetting the camera since having the camera reset every time you say, resized the window of a dynamically resizing canvas would be decently annoying.
Adding the camera resizing feature to p5.Framebuffer.js will happen. I'm just a bit busy rn.
Also yeah, we're gonna have to deal with the other cameras since they do still get reset.
tbh I should've probably been trying to find the root cause of why the CPU values are getting reset by a canvas change that should only be affecting the GPU values. I'll try to find that out later when I can. Hopefully that'll be able to address the root cause so we don't have worry about all of this since otherwise we'd be forced to just copy every single camera every single time.

inaridarkfox4231:
It might be easier to understand if you provide sample code. I think that it is too short and it is difficult to convey the intention.

inaridarkfox4231:
I'm also busy, so I'll end with this reply.

First, although it may contradict what I have stated above, here is my current view.

Having the default camera set to the default state for its size every time the canvas is resized is a very natural behavior. It's not a bug. If it does not change the camera matrix even if the size is changed, it may be a setting that does not match the size of the canvas, and some people may feel uncomfortable. It can be a breaking change for such people. you have to take that into account. So, I would like to withdraw my opinion above. I'm going to write the code myself to solve it.

To you, this behavior is a bug, but I think it's not very persuasive unless you write specific, working code to explain it. Create a sketch and post a link, or even a video, I think you need something. Sketches are necessary to check what happens when you change the specifications, so can't say you don't have to prepare anything. I think you need to prepare something.

I'm not familiar with p5.Framebuffer so I can't say anything. Since it was the reviewer who brought up this topic, I think you should ask the reviewer.

To avoid using the default settings, you can prepare data in advance and use it again, but since the size of the canvas changes, it would be impossible to use everything as it is. They need to think about it properly. Writing code is do that. We cannot do everything for them. Maybe they need to talk to someone or do some research, I think. Trial and error is necessary depending on the purpose.

It is inherently natural behavior, but certifying it as a bug and implementing some kind of specification change entails risks and responsibilities. I think it is necessary to think about it properly. Please consult with the reviewer and draw a better conclusion. 

That's all.

↑ I'm sorry, but I think you should fix this. Please follow the minimum etiquette when making a pull request.

RandomGamingDev:
I fixed the Resolves #[Add issue number here]. Also, I discussed with [davepagurek](https://github.com/davepagurek) and he seems to agree that it's a bug. For an example, a lot of programs oftentimes end up using the camera to store their values for movement among other things and even those that don't most likely don't cache them separately outside of the camera. For a 3d application like a game or any other sort of software a simple resize could easily break their code base. Plus, even if we were to say that since the WebGL uniforms reset so everything else should reset too, if you look at the function p5.js is clearly making an effort to cache and restore canvas values, so it's probably more so of a bug that they managed to do it for basically everything except for the camera.

RandomGamingDev:
As for creating a sketch for this, I feel like it's unnecessary. I could create a sketch demonstrating a moving camera that gets reset upon the window getting resized due to wanting to resize with the window, but that'd only display the same thing as what I said. If u mean getting a full project that has a problem with this, I feel that that's unnecessary.

inaridarkfox4231:
I can't help it, so I'll give you a suggestion. Unlike you, I'm not very handy, so I can't solve the problem without a sketch. So I'd like to make a patch, put a link of sketch, post a video of the result, and have it discussed and validated. excuse me.

The reason I am asking you to make a sketch is that I have no idea how you came to propose this specification change. If it's because you've hit a wall in making some concrete sketch, then just show it. That sets the direction of the discussion, but you can't do that, so this discussion doesn't move forward.

This pull request is not a so-called stack-overflow. If the problem can be solved with technique, there is no need to change the specification. Do you just want to revamp the specification by labeling it as a bug and claiming it as an improvement? I think it's not a contribution.

## My proposal:

First, let the renderer record the previous width and height in resizeCanvas().

```js
p5.prototype.resizeCanvas = function(w, h, noRedraw) {
  p5._validateParameters('resizeCanvas', arguments);

  /* ~~~~~~~ */

    // record previous width/height.
    this._renderer.prevWidth = this.width;
    this._renderer.prevHeight = this.height;
    this.width = w;
    this.height = h;
    // Make sure width and height are updated before the renderer resizes so
    // that framebuffers updated from the resize read the correct size
    this._renderer.resize(w, h);

   /* ~~~~~~~ */
};
```

Then change the camera's _resize() function to:

```js
p5.Camera.prototype._resize = function() {
  // If we're using the default camera, update the aspect ratio
  this._computeCameraDefaultSettings();

  if (this.cameraType === 'default') {
    this._setDefaultCamera();
  } else {
    this.aspectRatio = this.defaultAspectRatio;
    const oldAspectRatio =
      this._renderer.prevWidth / this._renderer.prevHeight;
    this.projMatrix.mat4[0] *= oldAspectRatio / this.aspectRatio;
    if (this._isActive()) {
      this._renderer.uPMatrix.mat4[0] = this.projMatrix.mat4[0];
    }
  }
}
```

If the cameraType is 'default', it will do the same as before. If 'custom', run defaultSettings on a new canvas, then multiply the first element of the projection matrix by the ratio of the previous aspect ratio and new aspect ratio. This procedure is for the perspective case, but it also works for ortho and frustum. Watch the video below.

[resize_problem_DEMO](https://editor.p5js.org/dark_fox/sketches/N5t-X7p8k)

https://github.com/processing/p5.js/assets/39549290/cec3809d-fc40-4c8c-b68e-f60961016399

I'll explain a few.

## Regarding the fact that the screen does not change when resizing in the horizontal direction

If 'custom', the current behavior is to fix the FOV, update only the aspect ratio, and call perspective. That's why. Applying similar behavior to other projection modes will result in similar screen changes in other projection modes.
For example, vRoidHub adopted such behavior. So I don't think it's unnatural.

## the same behavior as before when the camera type is 'default'

If you don't care about the camera, and if your sketch depends on the width and height of the screen, then simply adapting to the new canvas should suffice. Assuming such a case, I think it is better not to change the specifications.
When moving the camera, I think you should set the camera type to custom by executing a command such as perspective(PI/3) in advance. Of course, this is not a very natural way to do it, so that's the problem. However, if the type becomes custom just by moving the camera, I think people who want it to go back to the default when they change the canvas will be frustrating.

## As for p5.Framebuffer

It's dave pagurek who brought up the topic of p5.Framebuffer, so I'd better leave it to him. As for my opinion, it looks like it's just doing a _resize() on the camera. So I don't think there is any problem with this approach. I don't know the details.

## About the camera as a global variable

It would seem fine if appling _resize() to the camera, but it's a private function so It can't. What to do about it is out of my hands.

However, this is just a suggestion and if there are any problems, they will be corrected. Maybe it will be discarded. However, you don't seem to be contributing to the solution of this problem at all, so if you're having trouble with no ideas, I think adopting them is one way.

I am also busy, so I will refrain from further suggestions.

inaridarkfox4231:
For example, by calling _resize() for a camera that is not _curCamera like this, it is possible to support resizing even for cameras that are not active. But can't get _curCamera, so can't use this method.

[resize_problem_DEMO2](https://editor.p5js.org/dark_fox/sketches/WwGUB28kn)

https://github.com/processing/p5.js/assets/39549290/55bcfcbb-880f-4d32-99ca-6fbe2332982f

Maybe we could do this with a function like isActive() that gets a boolean value for whether or not it is active, and a global camera function like _resize(), It would be difficult in now environment.

Again, it's originally your job to make these suggestions and verify them. I have no right to rewrite the source code. Since you are the one who made the proposal, I strongly hope that you will take full responsibility for your response.

inaridarkfox4231:
Is your intended implementation ie removing the concept of 'default'? Then you should do this:

```js
p5.Camera.prototype._resize = function() {

  this._computeCameraDefaultSettings();
  this.aspectRatio = this.defaultAspectRatio;
  const oldAspectRatio =
    this._renderer.prevWidth / this._renderer.prevHeight;
  this.projMatrix.mat4[0] *= oldAspectRatio / this.aspectRatio;
  if (this._isActive()) {
    this._renderer.uPMatrix.mat4[0] = this.projMatrix.mat4[0];
  }
}
```

If you no longer need a concept like 'default' that resets the camera, then this is the way to go. However, even if you propose this, the specification change will not be realized without the reviewer's permission in the end. I'll leave the rest to you.

inaridarkfox4231:
If you don't want resetting the camera to be the default behavior, this is the way:

[resize_problem_DEMO3](https://editor.p5js.org/dark_fox/sketches/cF0NvF8NJ)
```js
p5.RendererGL.prototype.resize = function (w, h) {
  p5.Renderer.prototype.resize.call(this, w, h);
  this.GL.viewport(0, 0, this.GL.drawingBufferWidth, this.GL.drawingBufferHeight);
  this._viewport = this.GL.getParameter(this.GL.VIEWPORT);

  this._curCamera._resize(); //resize pixels buffer
  this._curCamera.cameraType = 'custom';

  var pixelsState = this._pixelsState;
  if (typeof pixelsState.pixels !== 'undefined') {
    pixelsState._setProperty('pixels', new Uint8Array(this.GL.drawingBufferWidth * this.GL.drawingBufferHeight * 4));
  }
};
```

https://github.com/processing/p5.js/assets/39549290/e2656fa2-51c9-4188-9b92-48bed8f250d0

This will make the default camera custom. No matter how much you move it, the specification change I suggested earlier will be applied, so the camera matrix will not be destroyed.
The concept of 'default' remains. It will be given when perspective() is called with no arguments. The concept is never used unless you explicitly grant it.

I will leave the rest to you.

> As for creating a sketch for this, I feel like it's unnecessary. I could create a sketch demonstrating a moving camera that gets reset upon the window getting resized due to wanting to resize with the window, but that'd only display the same thing as what I said. If u mean getting a full project that has a problem with this, I feel that that's unnecessary.

Did you understand the need to make a sketch?

inaridarkfox4231:
Regarding the camera as a global variable, if you change the createCamera() specification as follows so that all cameras created are custom, processing like perspective (PI/3) becomes unnecessary.

[resize_problem_DEMO4](https://editor.p5js.org/dark_fox/sketches/EjykFK_Gf)
```js
p5.prototype.createCamera = function () {
  this._assert3d('createCamera');
  const _cam = new p5.Camera(this._renderer);

  // compute default camera settings, then set a default camera
  _cam._computeCameraDefaultSettings();
  _cam._setDefaultCamera();

  _cam.cameraType = 'custom';

  // set renderer current camera to the new camera
  this._renderer._curCamera = _cam;

  return _cam;
};
```

But nothing happens if you don't call _resize(), so you can't fix the problem about that.

RandomGamingDev:
I'm confused how you wouldn't have understood the change and it's reason, but here's a duplicate of the sketch used originally in case you want it: https://editor.p5js.org/PotatoBoy/sketches/7NVWhsA7P.

RandomGamingDev:
The issue that I attempted to solve was the same one as before, where resizing of the camera resets the camera. Also, sorry for the late responses and the lack of commits recently. I've been busy.
My response afterwards was regarding why the camera behaved this way in the first place, since, the resetting of the uniforms relating to the WebGL context on the GPU typically wouldn't effect values relating to it on the CPU, but for some reason both get reset which I will do more research into when I can.

RandomGamingDev:
As for moving the camera, I believe that only the commands affecting the projection matrix turn the camera to type "custom".

RandomGamingDev:
The reason for checking for custom in the previous commit originally was because some people wouldn't care about the camera.

inaridarkfox4231:
I have understood well that I cannot help you. thank you very much.
I'll leave the rest to davePagurek.

inaridarkfox4231:
Are you the one raising the issue? Or are you the one to solve it? I don't know which one. If you're not the solver, explain the problem well to someone else and let them handle it. Because you are busy.

inaridarkfox4231:
My suggestion didn't seem to make sense to you, so I'll delete it. I'm sorry.
It seems that, I have misunderstood the nature of the problem. I can't get the nature of the problem myself, so I leave the rest to someone who can do it.

I looked at your sketch, but it's a mess and I have no idea what the problem is. I think you should start by clearing your head first.

[DEMO](https://editor.p5js.org/dark_fox/sketches/EBKJCF7zo)

https://github.com/processing/p5.js/assets/39549290/6870efc0-1849-4df6-8a4a-ae9761c42bd3

I can't see what the problem is. Leave it to someone who can see it.

If davePagurek says it's a bug, it's a bug. I don't think you can solve it, so I think you should leave it to him.

RandomGamingDev:
That was just the sketch for the previous fix, but what I'm talking about is the root issue behind both GPU & CPU values getting reset.

davepagurek:
I think it makes sense to not reset the position of a manually-placed camera (as opposed to the default camera, which I think should continue to be updated due to the default z of the camera being different at different resolutions in order to achieve a consistent field of view. Changing that would probably be another discussion.)

The talk before about the camera's `_resize` method is because that's the thing that's resetting the camera values. That might also be the cause of what you're talking about with updating uniforms, since that method may update camera matrices, and we send those to the GPU each frame? Also, my comment about Framebuffers is just that those have their own `resize` method which in turn resizes their camera, similar to `resizeCanvas` resizing the main canvas camera, so we should probably make those two cases work the same way, whatever we do. Updating the implementation of `_resize` would work for both, I believe.

I think it would be good to make a unit test out of it to make sure the default camera updates but a manually positioned camera doesn't, to help clarify the behaviour we want.

inaridarkfox4231:
I still don't know the direction of this discussion, but if the current specification change passes, the behavior of cameras with cameraType custom will be like this:

https://github.com/processing/p5.js/assets/39549290/b92b0676-e868-42b7-b60a-3e9970077931

```js
function setup() {
  perspective(PI/3);
}

function draw() {
  background(220);
  
  orbitControl();
  
  //rotateX(frameCount * 0.01);
  //rotateY(frameCount * 0.01);
  push();
  fill(0);
  rect(0, 0, 400, 400);
  pop();

  box(100);
}
```
Since the aspect ratio has not been updated, it will stretch and shrink horizontally, is this the intended behavior? Or does cameraType custom mean that it behaves like that? If so, that's fine.

Also, you're checking to see if _curCamera is undefined, but I think this is not  necessary if you're modifying RendererGL's resize() function.

I don't understand what you mean by "resize" discussed here. Therefore, I simply thought that the purpose of the discussion was to make the behavior that the appearance does not change even if the screen size is changed.

https://github.com/processing/p5.js/assets/39549290/8bd63009-ec50-4b0c-9db6-ab1e8687a6b1

But it seems that was not the case. CPU, GPU, uniform, framebuffer seems to be involved. It was not a discussion I could engage in. I'm sorry. Sorry.

davepagurek:
> Since the aspect ratio has not been updated, it will stretch and shrink horizontally, is this the intended behavior? Or does cameraType custom mean that it behaves like that? If so, that's fine.

I think for `frustum` and `ortho` stretching makes sense (you specify the left/right/top/bottom, and those stretch to fit to the viewport regardless of whether the aspect ratio is the same), and also for custom perspective (you manually specify an aspect ratio; if you pick one different from the main canvas, your content stretches.) So I think the only case where we want it to prevent stretching is if you made a camera and did not specify left/right/top/bottom or a specific aspect ratio. 

> I don't understand what you mean by "resize" discussed here.

I hope I'm not complicating the discussion too much by bringing in Framebuffers, `fbo.resize(width, height)` is like an equivalent to `graphic.resizeCanvas(width, height)` but just for framebuffers. Ideally, however cameras work in graphics is also how they work in framebuffers. I don't think we need to consider them when talking about what the behaviour should be, whatever makes sense for graphics makes sense for framebuffers too. If e.g. we add code to the `resizeCanvas` implementation, we'll probably just want to add the same code to `Framebuffer.resize` as well.

inaridarkfox4231:
I see, custom means manually changing the projection mode after resizing. I finally understood that. thank you very much.

inaridarkfox4231:
Perhaps I misunderstood the meaning of the names "default" and "custom". The "default" of projection mode means that the projection function is not called.
"Custom" means calling the projection function. Even if you just change the near and far clips, it will be "custom".
However, the "default" and "custom" are related to the projection mode, but apart from that, there is the concept of "default" and "custom" for "what should be the behavior when resizing".
I understood that the purpose of this discussion is to match the two types of "default" and "custom". In other words, if the projection mode is "Custom", you must also customize the resizing process.
However, I thought the goal was to give any projection mode a natural "default" behavior, and allow post-resize processing to customize it if needed.
It seems that was the cause of the misunderstanding. I'm sorry to have disturbed you.

inaridarkfox4231:
> Also, if we copy the camera and set this._renderer._curCamera = cam with the copy, users who have stored their camera in a local variable might be confused when editing their camera no longer affects how their sketch rendered.

I have tried to deal with this issue. First, let's see what the current behavior is. I prepared a global variable camera and tried moving it with pan() and tilt(). As a result, it certainly stopped working when resized.

[DEMO_3](https://editor.p5js.org/dark_fox/sketches/jlHNBcDhh)
```js
let myCam;

function setup() {
  myCam = createCamera();
}

function draw() {
  background(220);

  if (mouseIsPressed) {
    const panValue = (mouseX-width/2)*0.06/width;
    myCam.pan(panValue);
    const tiltvalue= (mouseY-height/2)*0.06/height;
    myCam.tilt(-tiltvalue);
  }

  push();
  fill(0);
  rect(0, 0, 400, 400);
  pop();

  box(100);
}

function windowResized() {
}
```

https://github.com/processing/p5.js/assets/39549290/11ecde42-2ab7-473a-95f3-42adec07d6d1

The cause is that the copied camera was assigned to this._renderer._curCamera. Since the reference to the global camera is removed, it will not move even if you operate it.

```js
    // reapply camera
    if (this._renderer._curCamera !== undefined) {
      this._renderer._curCamera = cam;
      if (this._renderer._curCamera.cameraType === 'custom') {
        this._renderer.uPMatrix = this._renderer._curCamera.projMatrix;
      }
    }
```

Rewrite this with the set() that was implemented yesterday as follows.

```js
    // reapply camera
    if (this._renderer._curCamera !== undefined) {
      if (this._renderer._curCamera.cameraType === 'default') {
        cam._computeCameraDefaultSettings();
        cam.perspective();
      }
      this._renderer._curCamera.set(cam);
    }
```
That is, if cameraType is 'default', you can reset the default by calling perspective() with no arguments. At that time, I need a new default value, so I'm calling a function _computeCameraDefaultSettings().
All we're doing here is deciding whether to use the default projection matrix of the new canvas size, or the same as before.
So, after setting the information appropriately in cam based on that, you can copy the contents with set(). set() does not destroy references. Also, since it's active, the renderer's uP is automatically updated with the projMatrix.

https://github.com/processing/p5.js/assets/39549290/625d8f73-43f7-4577-91ea-dc59885f77c5

In this way, _curCamera is updated by camera operation even after resizing.
It works fine even if the cameraType is custom.

inaridarkfox4231:
But if I fix the _resize() of the camera, I get the same behavior.
[DEMO_4](https://editor.p5js.org/dark_fox/sketches/J3tav4llp)
```js
p5.Camera.prototype._resize = function() {
  const cam = this.copy();
  if (this.cameraType === 'default') {
    cam._computeCameraDefaultSettings();
    cam.perspective();
  }
  this.set(cam);
}
```
I think it's easier than fixing resizeCanvas(). Because we don't need to check if the _curCamera property is undefined.
I'm not sure if this works for the framebuffer as well.

inaridarkfox4231:
But if I fix the _resize() of the camera, I get the same behavior.
[DEMO_4](https://editor.p5js.org/dark_fox/sketches/J3tav4llp)
```js
p5.Camera.prototype._resize = function() {
  if (this.cameraType === 'default') {
    this._computeCameraDefaultSettings();
    this.perspective();
  }
}
```
If I think about it, neither set() nor copy() is necessary. All we need to do here is make the projection mode default if 'default'.
The reason why it was necessary to use copy() or to obtain the matrix is because I was trying to change the specifications while not messing with the cameras _resize() function. However, since the _resize() function is only called when resizing is executed, it would have been better to have aimed to fix this from the beginning.

Also, no matter how simple the example is, it is important to make a sample and check it. I don't know if the behavior would have improved without it. Much easier than just discussing code. Because you can't be sure you're on the right path.

Thank you for creating the example. It is thanks to you that the verification went smoothly.

I think it's easier than fixing resizeCanvas(). Because we don't need to check if the _curCamera property is undefined.
I'm not sure if this works for the framebuffer as well.

RandomGamingDev:
Sorry for the late response, I've been a bit busy lol. Yeah, I looked at the demo and it looks it works properly and would be the same solution, but probably apply more uniformly across p5.js. Thanks for the help!

RandomGamingDev:
> Thanks for taking this on!
> 
> Do you think it's equivalent to take out the else branch here?
> 
> https://github.com/processing/p5.js/blob/d2b5d5e153c9b2a66b7cb80f2c41aecb350d2ac2/src/webgl/p5.Camera.js#L1615-L1626
> 
> If it is equivalent, it might be better to make the code change there to avoid doing some work then undoing it

Sorry about that lol. I should've probably checked. I didn't know that that was the cause of the issue lol. I'll create a commit removing the else branch instead.

src/webgl/p5.Camera.js
```diff
@@ -1617,11 +1617,6 @@ p5.Camera = class Camera {
     if (this.cameraType === 'default') {
       this._computeCameraDefaultSettings();
       this._setDefaultCamera();
-    } else {
```
davepagurek, line 1620: Currently framebuffers also have a duplicate version of this method: https://github.com/processing/p5.js/blob/53119bfc5c6f3f57c4791647d5d09d1977b24dc4/src/webgl/p5.Framebuffer.js#L55-L57

The duplication exists right now because it replaces `this._renderer.width`and height with `this.fbo.width` and height. If we remove the `else` branch here, I think we can just remove the overloaded `_resize` method from the framebuffer camera class, since the implementation would now be the same as for normal cameras.

davepagurek (commented):
Sorry for the delay on this! I think this change makes sense, do you think we make the same change to the frambebuffer camera class too?

It would also be great to add a unit test where we make two cameras, one default perspective and one custom, and show that after resizing the canvas, only one has a changed matrix.

src/webgl/p5.Camera.js
RandomGamingDev: Sorry for the late response. This looks like it'd work perfectly. I'll remove the overloaded `_resize()`.

RandomGamingDev:
btw should we making it so that moving the camera sets `cameraType` to custom?

davepagurek:
> btw should we making it so that moving the camera sets cameraType to custom?

I think this is maybe a bit more complicated than we though. I'm just realizing there are some other parts of the codebase where we basically treat `default` as meaning `perspective`, which is clear isn't the case all the time:

https://github.com/processing/p5.js/blob/a940cd560579821a3bde97693bcf2ef3dcabfcca/src/webgl/p5.Shader.js#L278-L286

Right now we use `cameraType` to determine:
1. Do we reset the projection matrix when the size changes?
2. Do we reset the camera position when the size changes?
3. Do we render lines with thickness scaling with perspective or not?

For default cameras, the answers are YYY; for custom perspective, the answers are NNY; for default perspective with location changes, the answers are maybe YNY; for ortho cameras, the answers are NNN.

So maybe we need to make different properties for (1) whether it's using perspective or not, (2) whether it uses custom position, and (3) whether it uses custom projection?

RandomGamingDev:
Maybe. Although I wonder why we're using strings for these in the first place. Why not a more enum like system?

davepagurek:
I think in this case, an enum is maybe not the right data type for this if we want to store multiple properties. If we're considering having something like `cameraType = { perspective: true, autoPosition: false, autoProjection: false }`, since JS objects use strings as keys anyway, then it would probably make sense to continue using strings for these keys.

In general, other things that look like enums currently (e.g. CENTER, LEFT, RIGHT for text alignment) are also just strings, but we export named constants for easier access, so we could always do the same and make named variables for each string constant.

RandomGamingDev:
Me and @davepagurek decided that it'd be best to make it so that the position (including the rotation) doesn't reset at all, but keep the view matrix at it is so that it can resize alongside the canvas. I'll add this alongside a constructor with an option to choose between `default` and `custom` cameras which shows whether or not the view matrix will reset when the canvas does. It will be `default` by default mostly for backwards compatibility reasons.
I'll implement this when I have time, but tbh that might take a while lol

RandomGamingDev:
If any people have questions or disagree with any of the changes proposed comment below.

davepagurek:
btw @RandomGamingDev I made a little sketch here to test out what it feels like to use a fixed camera distance vs a fixed field of view (what we currently do.)

https://github.com/processing/p5.js/assets/5315059/73fef336-ae3f-4211-9f13-a9e39cef680f

Live: https://editor.p5js.org/davepagurek/sketches/TC3kWHOWk

It looks like using a fixed camera distance has some benefits actually? There are less changes when you resize the camera, so that's arguably a plus. And then for our purposes, it means we would only need to update the perspective matrix for default perspective cameras when the window size changes.

Here's another quick sketch to verify that if you make a camera, resize the canvas, then click the play button to restart with the camera in that new size, it looks visually the same as it did after starting at a different size + resizing:

https://github.com/processing/p5.js/assets/5315059/5bdb369e-4660-472e-a794-ff8eff18a36a

Live: https://editor.p5js.org/davepagurek/sketches/CTKUAAmMZ

RandomGamingDev:
I'll probably create another pull request to add more complex features like that and just use this pull request to fix the issue with as little as possible and as close to the issue itself. That way other people can more easily chip in and bring up their thoughts on the changes.

davepagurek (approved):
I think this looks good to go now! Thanks for your patience and the long review process for this one, it ended up being a more complicated issue than we though 😅

Issue #4820, opened 2020-10-01
Suggestion: provide shader for each p5js filter


dhowe:
Suggestion: make available the set of filters in p5js (currently THRESHOLD, GRAY, OPAQUE, INVERT, POSTERIZE, BLUR, ERODE, DILATE) as shaders (with the same names), so that users can more easily port sketches between default and WEBGL renderers. 

USE-CASE: Student has performance problems in sketch using filters. Instructor would like to suggest WEBGL, but the filter doesn't exist as a shader (or is not easily found).

Another option would be to have the filter call actually work in WEBGL by loading/invoking the correct shader automatically.

#### Most appropriate sub-area of p5.js?

- [x] Accessibility (Web Accessibility)
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Other (specify if possible)

stalgiag:
Interesting idea! @aferriss has already done most (all?) of the hard work [for this in this repo.](https://github.com/aferriss/p5jsShaderExamples)

Typically, I would direct a student to this repo when I was suggesting this route. But perhaps it is worth the addition to library size (likely small) to have these as p5.Shader constants.

I think that performance is really important to consider right now and definitely falls under the umbrella of accessibility. Working with students through Zoom, I am painfully aware of the hard limits that students are hitting with performance. This is exacerbated by the fact that many students are using home computers for the first time when they have used school lab computers (which often have superior hardware) in the past.

For sketches that rely heavily on images with filters, the performance gains with WEBGL and shaders will be very significant.

dhowe:
> Interesting idea! @aferriss has already done most (all?) of the hard work for this in this repo.

Exactly -- I think it would largely be a question of syncing up the names. And seems a good thing to do whether or not we want to have them invoked automatically when use the GL renderer.

outofambit:
This sounds like a great idea to me! 💜

JalokimFreelance:
Hey, I would like to help with that issue, would just the .vert and .frag files be enough? Or do I have to convert them somehow into Shader object?

outofambit:
@JalokimFreelance i think that would be enough! i'm 90% sure we already have some build process for inlining shader files into the p5.js lib.

JetStarBlues:
[...]

PR #6237, opened 2023-06-23
Adds createFilterShader() and custom shader support to the webGL filter() function 

wong-justin:
Addresses: [#4820](https://github.com/processing/p5.js/issues/4820), [#5571](https://github.com/processing/p5.js/issues/5571)

## Initial changes:
- `filter()` accepts a shader parameter
- Added property for storing current filter shader

## Will also change:
- Apply the current filter shader and render to main canvas / parent renderer
- Add shaders for constants like `GRAYSCALE`, etc
- Clarifying docs to show new functionality

`p5.Framebuffer` is an alternative graphics layer to pursue if `p5.Graphics` doesn't work. Either one is nice if it gives a simple way to call `shader()`, renders to main with something like `texture() plane()` / `copyPixels()`, and gives resizing and other coverage for free. Right now I started a framebuffer approach after running into two difficult p5.Graphics errors.

(This PR is a working draft that's part of the GSoC 2023 project 'Supporting shader-based filters in p5.js", mentored by @aferriss and @aceslowman)

---

- [x] npm run lint passes


src/webgl/p5.RendererGL.js
```diff
@@ -879,29 +879,44 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    // not sure why userShader[0] is the actual shader object
+    // and userShader is Arguments instead
+    userShader = userShader[0];
+    // perhaps necessary: bind the shader to the pg renderer, not the main
+    // let userShaderCopy =
+    //   new p5.Shader(pg._renderer, userShader._vertSrc, userShader._fragSrc);
+
+    // apply shader to pg
+    // TODO: shader isn't being applied properly
+    pg.shader(userShader);
```
aferriss, line 906: I think user shader will need to be created using the pg context here. I wonder if we can just extract the shader strings from an already loaded shader and reload it using `pg.createShader()`

src/image/pixels.js
```diff
@@ -469,8 +469,19 @@ p5.prototype._copyHelper = (
...
  */
+
+/**
+ * @method filter
+ * @param {p5.Shader}  shaderFilter  A shader that's been loaded
+ */
 p5.prototype.filter = function(operation, value) {
   p5._validateParameters('filter', arguments);
+
+  if (this._renderer.isP3D) {
```
aferriss, line 480: I think we should continue this discussion about whether or not webGL filter should attempt to be used even if the source sketch was not in webGL model

src/webgl/p5.RendererGL.js
```diff
@@ -874,12 +879,43 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    //   this.filterShader = args[0]
+    // else
+    //   this.filterShader = map(args[0], {GRAYSCALE: grayscaleShader, ...})
+    //   filterOperationParameter = undefined or args[1]
+
+    let userShader = args[0];
+    // Bind the shader to the pg renderer,
+    // since when the user makes it with createShader() it binds to the main
+    userShader =
+      new p5.Shader(pg._renderer, userShader._vertSrc, userShader._fragSrc);
```
aferriss, line 902: We probably don't wanna create a shader every frame, can we cache it somehow?

src/image/pixels.js
```diff
@@ -469,8 +469,19 @@ p5.prototype._copyHelper = (
  * image of a brickwall
  * image of a brickwall with less detail
  */
+
+/**
+ * @method filter
+ * @param {p5.Shader}  shaderFilter  A shader that's been loaded
```
aferriss, line 475: Add in example, alt text, supporting documentation, etc.

src/webgl/p5.RendererGL.js
aferriss, line 902: Should we provide users with a default vertex shader?

src/webgl/p5.Shader.js
```diff
@@ -382,6 +382,18 @@ p5.Shader = class {
    * canvas toggles between a circular gradient of orange and blue vertically. and a circular gradient of red and green moving horizontally when mouse is clicked/pressed.
    */
   setUniform(uniformName, data) {
+    let other = this._renderer.filterShader;
```
aferriss, line 385: Maybe I'm not understanding, but isn't this unnecessary? I think you already set the uniform in the filter function. If it is necessary, can we check if the filter layer exists first before executing the code here?

src/webgl/p5.RendererGL.js
```diff
@@ -874,12 +879,54 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    let pg = this.filterGraphicsLayer;
+
+    // TODO, handle different signatures:
+    // if args[0] is a p5.Shader,
+    //   this.filterShader = args[0]
+    // else
+    //   this.filterShader = map(args[0], {GRAYSCALE: grayscaleShader, ...})
+    //   filterOperationParameter = undefined or args[1]
+
+    if (typeof args[0] === 'string') {
```
aferriss, line 899: Is this the right check? I think we need to know if args[0] is a p5.shader, or a filter constant ( BLUR, INVERT, etc), and if not, we print an error. Aren't the filter constants strings?

src/webgl/p5.Shader.js
aferriss: Another thought, what about a case where you have two of the same shader, but want to pass different uniforms to them, or use them in different contexts throughout your sketch, (for ex an invert shader running on both a piece of geometry in the main sketch and an invert filter for the filter function) Would this code support that behavior?

src/webgl/p5.RendererGL.js
wong-justin: I think so. Here are some related thoughts: #6144 
I'm leaning towards the simplest option, just exposing a default vertex shader as a string constant.

src/webgl/p5.RendererGL.js
wong-justin: Yes, that's the check. Although there shouldn't need to be a third case handling errors, since the args get validated in the parent `filter()` function: [pixels.js#L472](https://github.com/processing/p5.js/blob/3b0d1328b96a6d46dfc77e3f33e97077934967fa/src/image/pixels.js#L472)

Right now the string / filter constant case has an error placeholder. It will be replaced in the next PR that implements the shader versions of filter constants. And the current code handles the p5.Shader case.

src/image/pixels.js
wong-justin: Agreed, I think that's the primary value of this whole project, speeding up existing `filter()` calls in P2D sketches. So P2D sketches would have to create a hidden webGL context if `filter` is used.

I think shader filter usage should be automatic while including an opt-out. Something like `filter(CONSTANT, useWebGL=false)`?

And I think that should be included in the next PR that implements the webGL shaders for filter constants.

src/webgl/p5.Shader.js
wong-justin: Great point, I didn't think of that use case. 

On transferring uniforms from the main shader to the secondary shader: A better way of detecting when to copy uniforms might be checking for a child / copy reference, something like `if this.has duplicateShader, then this.duplicateShader.setUniform(...) instead`. And that reference would be set when the duplicateShader is created. I think that should work - not sure why I thought of this other way first.

The whole thing feels a little icky though - first having to create another webgl graphics context, and now having to duplicate shaders since the user's `createShader()` call is bound to the main renderer, and there doesn't seem like a decent way to rebind an existing shader to a different webgl context. And I think there's a little overhead to compiling another shader, although it's been a negligible effect in my testing. I think this duplication complexity would be avoided with the framebuffer approach.

src/webgl/p5.RendererGL.js
davepagurek: Another thought, we could make a function to make a shader that only takes in a fragment shader and automatically uses that default vertex shader? Like createFilterShader()?

src/image/pixels.js
aceslowman: I agree about the opt-out, most people would benefit from the hidden webGL context, but it feels like there would be some negative side-effects eventually if there was no way to avoid that. Am I correct in thinking that there might be situations where a shader heavy project might benefit from strategically shifting some effects to the CPU?

src/webgl/p5.RendererGL.js
aceslowman: A default vertex shader as a string constant honestly makes sense to me, although I'd like to hear other thoughts. Would the idea be that they would use createShader() like normal, but pass in the new constant?

src/image/pixels.js
```diff
@@ -458,6 +463,44 @@ p5.prototype._copyHelper = (
...
+ *   void main() {
+ *     vec2 uv = vTexCoord;
+ *     uv.y = 1.0 - uv.y;
+ *     vec4 sampledColor = texture2D(tex0, uv);
+ *     float gray = luma(sampledColor.rgb);
+ *     gl_FragColor = vec4(gray, gray, gray, 1);
+ *   }`
+ * );
+ * background('RED');
+ * filter(myShader);
```
aferriss, line 499: I think instead of just turning a red canvas grey it would be better to show off the feature but filtering an image that has been previously drawn

src/image/pixels.js
aferriss: I think we can check if webgl support is available and then attempt to filter on the GPU for the filter constants. Agree that it would be good to opt out if possible but it does add some slight code complexity here

src/webgl/material.js
```diff
@@ -183,6 +183,65 @@ p5.prototype.createShader = function(vertSrc, fragSrc) {
...
+ * Creates a new <a href="#/p5.Shader">p5.Shader</a> object.
+ *
+ * Note, shaders can only be used in WEBGL mode.
+ *
+ * @method createFilterShader
+ * @param {String} fragSrc source code for the fragment shader
+ * @returns {p5.Shader} a shader object created from the provided
+ *                      fragment shader.
+ * @example
+ * <div modernizr='webgl'>
```
aferriss, line 198: Is this modernizer line in other examples?

src/webgl/material.js
```diff
@@ -183,6 +183,65 @@ p5.prototype.createShader = function(vertSrc, fragSrc) {
...
+ *   uniform sampler2D tex0;
+ *
+ *   float luma(vec3 color) {
+ *     return dot(color, vec3(0.299, 0.587, 0.114));
+ *   }
+ *
+ *   void main() {
+ *     vec2 uv = vTexCoord;
+ *     // gets flipped vertically
+ *     uv.y = 1.0 - uv.y;
```
aferriss, line 213: Should we preflip the y uv so that the user doesn't need to deal with this?

src/webgl/material.js
```diff
@@ -183,6 +183,65 @@ p5.prototype.createShader = function(vertSrc, fragSrc) {
...
+ *   createCanvas(100, 100, WEBGL);
+ *   let s = createFilterShader(frag);
+ *   background('RED');
+ *   filter(s);
+ *   describe('a plain gray canvas');
+ * }
+ * </code>
+ * </div>
+ */
+p5.prototype.createFilterShader = function(fragSrc) {
```
aferriss, line 228: I like it, nice and simple!

src/webgl/p5.Shader.js
aferriss: The overhead to create and compile a single other shader should be minimal, I don't think we need to worry too much since it's ideally only happening once in the lifetime of a sketch

src/webgl/material.js
aceslowman: Just curious, would the user be able to understand that this is being done for them?

wong-justin: Yeah, after trying the pre-flipped `vTexCoords.y` it for a week I came across passing mouseY as a uniform and forgetting to invert the coordinates. So the user will eventually find out that they have to flip y, and the inconsistency might be confusing between when it's done for them (maybe this one time out of convenience) and when it's not (the rest of the time).

src/webgl/material.js
```diff
@@ -211,29 +211,53 @@ p5.prototype.createShader = function(vertSrc, fragSrc) {
  * <div modernizr='webgl'>
  * <code>
  * function setup() {
- *   let frag = `precision mediump float;
- *   varying mediump vec2 vTexCoord;
+ *   let fragSrc = `precision highp float;
```
wong-justin, line 218: Are we good to use `highp` as the default in examples? I guess it's only certain old mobile hardware that would care about mediump, and it's becoming less of a matter over time.

aferriss: Yep, no issues there!

src/webgl/material.js
```diff
@@ -184,16 +184,19 @@ p5.prototype.createShader = function(vertSrc, fragSrc) {
...
 /**
- * Like <a href="#/createShader">createShader()</a>, but with
- * a default vertex shader so that only a fragment shader is needed.
- * Creates a new <a href="#/p5.Shader">p5.Shader</a> object.
+ * Creates a new <a href="#/p5.Shader">p5.Shader</a> object with only a fragment shader, intended for creating image effects on the canvas.
+ * Like <a href="#/createShader">createShader()</a>, but with a default vertex shader included.
  *
- * The fragment shader is given a uniform, or variable, called `tex0`.
+ * Note:
+ * - The fragment shader is given a uniform, or variable, called `tex0`.
```
aferriss, line 191: Can we change the language to clarify that the default input is a texture called tex0?

Something like: "The fragment shader is provided with a single texture input uniform called `tex0`"

src/webgl/material.js
```diff
@@ -191,17 +191,7 @@ p5.prototype.createShader = function(vertSrc, fragSrc) {
  * The fragment shader is given a uniform, or variable, called `tex0`.
  * This is created specificially for filter shaders to access the canvas contents.
```
aferriss, line 192: Maybe change the language slightly here to say that this function is intended to be used with the `filter()` function. 

"createFilterShader is intended to be used in tandem with the <a href="filter function...">`filter()`</a> for filtering the entire contents of a canvas when using webGL mode.

src/webgl/p5.RendererGL.js
```diff
@@ -872,12 +877,51 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
-    // https://github.com/aferriss/p5jsShaderExamples
-    // https://itp-xstory.github.io/p5js-shaders/#/
-    p5._friendlyError('filter() does not work in WEBGL mode');
+  filter(args) {
+    // Couldn't create graphics in RendererGL constructor
+    // (led to infinite loop)
+    // so it's just created here once on the initial filter call.
+    if (!this.filterGraphicsLayer) {
+      this.filterGraphicsLayer =
+        new p5.Graphics(this.width, this.height, constants.WEBGL, this._pInst);
```
aferriss, line 886: Could it not be working with pgraphics because this._pInst is referring to the main sketch?

wong-justin: Yeah I think that's it. Would the alternative be making a new `p5` instance there?

Here's some logs:
image

wong-justin: Also trying to reference the other time in the source where a p5.Graphics is created, in `src/core/rendering.js`:
```
p5.prototype.createGraphics = function(w, h, renderer, canvas) {
  ...
  return new p5.Graphics(w, h, renderer, this, canvas);
}
```

src/webgl/p5.RendererGL.js
```diff
@@ -872,12 +877,51 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    // Copy the user shader once on the initial filter call,
+    // since it has to be bound to pg and not main
+    let isSameUserShader = (
+      this.filterShader !== undefined &&
+      userShader._vertSrc === this.filterShader._vertSrc &&
+      userShader._fragSrc === this.filterShader._fragSrc
+    );
+    if (!isSameUserShader) {
+      this.filterShader =
+        new p5.Shader(pg._renderer, userShader._vertSrc, userShader._fragSrc);
```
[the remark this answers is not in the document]
wong-justin: Agreed to skip this tiny optimization for now

aferriss (approved):

PR #6255, opened 2023-07-06
Image Light Feature - GSOC 2023

AryanKoundal:
This PR is for the work under the GSOC 2023 project "Improving p5.js WebGL/3d functionality by adding support for image-based lighting in p5.js WebGL mode."

### Project Description:
This project aims to add lightning to a 3D object, using the surrounding environment as a single light source, which is generally called Image-Based Lightning. In simple words, one can very quickly drop in an image from real life to use as surrounding lights, rather than continuously tweaking the colors and positions of ambient, point, etc lights. 

### Changes:
This includes the whole progress broken down into steps. The boxes will be checked off according to the progress made.
- [x] CubeMap Convolution
- [x] PBR and Irradiance Lighting
- [x] Pre-filtering environment map
- [x] Pre-computing the BRDF

### PR Checklist
- [x] `npm test`
- [x] `npm run build` passes
- [x] `npm run lint` passes
- [x] Inline documentation is included/updated
- [x] Examples included/updated
- [x] Functions Descriptions and Documentation is included/updated

The project is completed and here are some screenshots and videos demonstrating the work.

Image of Example 1
example 1

Video of Example 1
https://github.com/processing/p5.js/assets/77334487/44b30c77-33c1-41d0-ada5-282424978832

Image oF Example 2
example 2

Video of Example 2
https://github.com/processing/p5.js/assets/77334487/a0a6b3f9-b25b-451f-961e-b2970cb9e907


src/webgl/shaders/imageLight.frag
```diff
@@ -0,0 +1,113 @@
+precision mediump float;
+varying vec3 localPos;
+
+// the HDR cubemap converted (can be from an equirectangular environment map.)
+uniform sampler2D environmentMap;
+varying vec2 vTexCoord;
+
+const float PI = 3.141;
+
+vec2 normalToEquirectangular( vec3 v){
```
davepagurek, line 10: Do we use this function?

src/webgl/shaders/imageLight.frag
```diff
@@ -0,0 +1,113 @@
...
+  // if phi is 0, then there are no x, z components
+  float theta = 0.0;
+  // else 
+  theta = acos(v.x / sin(phi));
+  float sinTheta = v.z / sin(phi);
+  if (sinTheta < 0.0) {
+    // Turn it into -theta, but in the 0-2PI range
+    theta = 2.0 * PI - theta;
+  }
+  theta = theta / (2.0 * 3.14159);
```
davepagurek, line 36: Maybe where we use 3.14159 right now we can use the `const float PI` we defined earlier, and add more decimals to that?

davepagurek (commented):
Looking good so far!

src/webgl/shaders/imageLight.frag
AryanKoundal: Nope, I actually re-wrote the function in line 22. The function N2E is the one being used for now. I think I should remove the function on line 10 and rename the function at line 22.

AryanKoundal: Nope, I actually re-wrote the function in line 22. The function N2E is the one being used for now. I think I should remove the function on line 10 and rename the function at line 22.

src/webgl/shaders/imageLight.frag
AryanKoundal: Sure, I'll clean up this file. It has some extra code as well.

src/webgl/p5.RendererGL.js
```diff
@@ -446,6 +448,17 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     this.spotLightAngle = [];
     this.spotLightConc = [];
 
+    // repetition
+    this.textures = new Map();
```
davepagurek, line 452: I think we don't need to declare this any more now that we have `blurryTextures` below

src/webgl/light.js
```diff
@@ -494,6 +494,33 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
   return this;
 };
 
+p5.prototype.imageLight = function(img){
+  // ! if the graphic is always true, do we really need this section?
```
davepagurek, line 498: If we're going to use `img` as a light source, it means we'll need the blurry version of its texture in order to pass into the shader. If `this.blurryTextures(img)` is defined, then we've already made the blurry texture in a previous frame and don't need to do anything more here. If not though, that's when we need to run the commented out code below to make it, and then after line 512 we'll need to store it with `this.blurryTextures.set(img, newGraphic)`

davepagurek: We could alternatively do this if statement right before we do `shader.setUniform('equiRectangularTextures', this.blurryTextures.get(this.activeImageLight));` later on. The idea is that we need that texture to exist by that point

src/webgl/p5.RendererGL.js
```diff
@@ -1701,6 +1715,14 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     return tex;
   }
 
+  // used in imageLight
+  // create a new texture from the img input
+  getBlurryTexture(input){
+    const blurrytex = new p5.Texture(this, input);
```
davepagurek, line 1721: I think the main difference between the blurry textures we store here and the textures already used by RendererGL is that for normal textures, we just make a new p5.Texture directly from the input. For this though, the input is non-blurry, and we want to store the blurry version, so somewhere (maybe in here actually?) we should run the shader to generate the blurry image onto a graphic and store that instead.

src/webgl/p5.RendererGL.js
```diff
@@ -1796,6 +1821,21 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    //set uniform values
+    if( this.activeImageLight == null ){
+      console.log('activeImageLight prop is null');
+    }
+    shader.setUniform('uUseImageLight', this.activeImageLight != null );
+    // true
+    if (this.activeImageLight) {
+      // this.activeImageLight has image as a key
+      // look up the texture from the blurryTexture map
+      shader.setUniform('equiRectangularTextures', this.blurryTextures.get(this.activeImageLight));
```
davepagurek, line 1835: If we decide to make `getBlurryTexture` compute and store the blurry texture if it doesn't yet exist, instead of doing `blurryTextures.get()` here, we could do `this.getBlurryTexture(this.activeImageLight)`

src/webgl/p5.RendererGL.js
TanviKumar, line 1721: nit: blurryTexture or inputBlurryTexture

src/webgl/p5.RendererGL.js
AryanKoundal: Sure, I will remove this. I was going to ask you about this, that's the reason I marked it as repetition.

src/webgl/shaders/lighting.glsl
```diff
@@ -33,8 +33,16 @@ uniform float uConstantAttenuation;
...
 
+// setting from  _setImageLightUniforms()
+// boolean to initiate the calculateImageDiffuse and calculateImageSpecular
+uniform bool uUseImageLight;
+// storing the texture
+uniform sampler2D equiRectangularTextures;
+
 const float specularFactor = 2.0;
 const float diffuseFactor = 0.73;
+const float PI = 3.141;
```
TanviKumar, line 44: Can this be used from another common constant file? Would be better to have a value closer to real PI, since 3.14159 is closer to 3.142

src/webgl/shaders/lighting.glsl
```diff
@@ -67,6 +75,44 @@ LightResult _light(vec3 viewDirection, vec3 normal, vec3 lightVector) {
   return lr;
 }
 
+float map(float value, float min1, float max1, float min2, float max2) {
```
TanviKumar, line 78: nit: A clearer name for this if possible

src/webgl/shaders/lighting.glsl
```diff
@@ -67,6 +75,44 @@ LightResult _light(vec3 viewDirection, vec3 normal, vec3 lightVector) {
...
+  // z = r sin(phi) sin(theta)
+  float phi = acos( normal.y );
+  // if phi is 0, then there are no x, z components
+  float theta = 0.0;
+  // else 
+  if(  abs(sin(phi)) > 0.0 ){
+    theta = acos( normal.x / sin(phi) );
+  }   
+
+    vec2 newTexCoor = vec2(
```
TanviKumar, line 94: Identation is off

src/webgl/p5.RendererGL.js
AryanKoundal: We actually refactored this function completely.

src/webgl/shaders/lighting.glsl
AryanKoundal: I don't think we have a file for constants because mostly constants are taken via the "MATH" object but it isn't available in shader languages according to my knowledge.  

We can do `#define PI 3.141592` at the top of this file.
Another possible option is to do  `#define PI 3.141592` in the `\webgl\shaders\webgl2Compatibility.glsl` file.

I am not sure which one is better.

src/webgl/shaders/lighting.glsl
AryanKoundal: The `map(vale, min1, min2, max1, max2)` function is converting the range of value from [min1 to max1] to [min2 to max2].

So should we make it something like `convertRange()` or `mapRange()` maybe?

src/webgl/shaders/lighting.glsl
AryanKoundal: Thanks, corrected it :)

src/webgl/shaders/lighting.glsl
davepagurek: This function mirrors https://p5js.org/reference/#/p5/map so maybe we could link to that reference page + add a comment description of what it does?

src/webgl/shaders/lighting.glsl
davepagurek: For now since we only use this here and we don't use pi anywhere else in a shader yet (to my surprise, haha) I think we can keep it in this file for now, and if we find we need to use it in another file too (maybe we need it for the specular shader?) we can factor it out into a new shader constants file at that point

src/webgl/shaders/lighting.glsl
```diff
@@ -67,6 +79,57 @@ LightResult _light(vec3 viewDirection, vec3 normal, vec3 lightVector) {
...
+  vec3 worldNormal = normalize(vNormal);
+  vec3 lightDirection = normalize( vViewPosition - worldCameraPosition );
+  vec3 R = reflect(lightDirection, worldNormal);
+  vec2 newTexCoor = mapTextureToNormal( R );
+#ifdef WEBGL2
+  vec4 outColor = textureLod(environmentMapSpecular, newTexCoor, levelOfDetail);
+#else
+  vec4 outColor = TEXTURE(environmentMapSpecular, newTexCoor);
+#endif
+  return outColor.xyz;
```
davepagurek, line 130: To darken the dark parts:

```suggestion
  return pow(outColor.xyz, 10.0);
```

davepagurek: I guess more specific to p5: we sort of want to try to match the look that we get with point lights when using the same `specularMaterial()` and `shininess()`.

src/webgl/shaders/imageLightDiffused.frag
```diff
@@ -0,0 +1,82 @@
...
+	vec3 irradiance = vec3(0.0);  
+	vec3 up	= vec3(0.0, 1.0, 0.0);
+	vec3 right = normalize(cross(up, normal));
+	up = normalize(cross(normal, right));
+
+	//  We specify a fixed sampleDelta delta value to traverse the hemisphere; decreasing or increasing the sample delta will increase or decrease the accuracy respectively.
+	const float sampleDelta = 0.025;
+	float nrSamples = 0.0;
+  
+	for(float phi = 0.0; phi < 2.0 * PI; phi += sampleDelta)
```
davepagurek, line 62: Btw, I think the indentation looks off on Github because this line uses tabs and the next line uses spaces. Maybe vscode has as button to convert everything to spaces?

src/webgl/shaders/lighting.glsl
```diff
@@ -67,6 +79,59 @@ LightResult _light(vec3 viewDirection, vec3 normal, vec3 lightVector) {
...
+  vec3 R = reflect(lightDirection, worldNormal);
+  vec2 newTexCoor = mapTextureToNormal( R );
+#ifdef WEBGL2
+  vec4 outColor = textureLod(environmentMapSpecular, newTexCoor, levelOfDetail);
+#else
+  vec4 outColor = TEXTURE(environmentMapSpecular, newTexCoor);
+#endif
+  // this is to make the darker sections more dark
+  // png and jpg usually flatten the brightness so it is to reverse that
+  return pow(outColor.xyz, 10.0);
```
davepagurek, line 132: Looks like tests are failing because `pow` needs both parameters to be the same type :') So maybe this needs to be:
```suggestion
  return pow(outColor.xyz, vec3(10.0));
```

src/webgl/light.js
```diff
@@ -494,6 +494,12 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
   return this;
 };
 
+p5.prototype.imageLight = function(img){
+  // setting activeImageLight so that the setUniform is executed
+  this._renderer.activeImageLight = img;
```
davepagurek, line 499: oh I just noticed, I think right now this doesn't get reset in `push`/`pop`, so we'd need to add to the `push` method here to do something similar to this: https://github.com/processing/p5.js/blob/d713b35e79866c5bf79fbcf338ff8761da846248/src/webgl/p5.RendererGL.js#L1583-L1583
e.g.
```js
properties.activeImageLight = this.activeImageLight;
```

src/webgl/light.js
```diff
@@ -494,6 +494,12 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
   return this;
 };
 
+p5.prototype.imageLight = function(img){
+  // setting activeImageLight so that the setUniform is executed
```
davepagurek, line 498: Maybe we can be a bit more specific here and say that this is what `_setImageLightUniforms` looks at to see if it needs to send uniforms to the fill shader

src/webgl/p5.RendererGL.js
```diff
@@ -468,6 +473,17 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     this.spotLightAngle = [];
     this.spotLightConc = [];
 
+    // Null if no image light is set, otherwise, it is set to a p5.Image
```
davepagurek, line 476: Maybe we can rephrase this into sentences so it's a bit easier to read? Also, for each map, we should say what the keys and values are. E.g. for this one, it maps a `p5.Image` used by `imageLight()` to a `p5.Graphics` containing the combined light it sends to each direction.

src/webgl/p5.RendererGL.js
```diff
@@ -468,6 +473,17 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    // made a blurrytexture attribute to be used in imageLight function
+    // this is to lookup image from blurrytexture Map rather from a
+    // texture map
+    this.diffusedTextures = new Map();
+    // map to store the created textures to prevent mis calculation
```
davepagurek, line 481: This one maps the input `p5.Image` to a `p5.MipmapTexture`. Worth mentioning that we use a mipmap texture because it can contain light from multiple roughness levels.

src/webgl/p5.RendererGL.js
```diff
@@ -468,6 +473,17 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    this.specularTextures = new Map();
+    // property to be set true after imageLight is called
```
davepagurek, line 483: Slight update to this: we don't set it to true, we set it to a `p5.Image`. Instead of checking if it's true, we check if it's not null

src/webgl/p5.RendererGL.js
```diff
@@ -601,6 +617,7 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     this.curStrokeCap = constants.ROUND;
     this.curStrokeJoin = constants.ROUND;
 
+    // repetition
```
davepagurek, line 620: Do we still need this line?

src/webgl/p5.RendererGL.js
```diff
@@ -1874,6 +1905,85 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     this.textures.set(src, tex);
     return tex;
   }
+  /*
+    *  used in imageLight,
+    *  To create a blurry image from the input non blurry img,
+    *  Add it to the blurryTexture map,
+    *  Returns the blurry image
+   */
+  getBlurryTexture(input){
```
davepagurek, line 1914: Maybe we can rename this to `getDiffusedTexture` for consistency?

src/webgl/p5.RendererGL.js
```diff
@@ -1874,6 +1905,85 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     this.textures.set(src, tex);
     return tex;
   }
+  /*
+    *  used in imageLight,
+    *  To create a blurry image from the input non blurry img,
+    *  Add it to the blurryTexture map,
```
davepagurek, line 1911: I think we renamed `blurryTexture` to `diffusedTextures`. Also we can mention that we only create it if it does not already exist.

src/webgl/p5.RendererGL.js
```diff
@@ -1874,6 +1905,85 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+  getBlurryTexture(input){
+    // if one already exists for a given input image
+    if(this.diffusedTextures.get(input)!=null){
+      return this.diffusedTextures.get(input);
+    }
+    // if not, then create one
+    let newGraphic; // maybe switch to framebuffer
+    // draw the blurry image
+    // set the shader on the graphic and set the shader on the image
+    // this._renderer._applyTextProperties(img, newGraphic);
```
davepagurek, line 1923: I think we can take out this one

src/webgl/p5.RendererGL.js
```diff
@@ -1874,6 +1905,85 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    // make small width, hardcode to 200px
+    let smallWidth = 200;
+    //let smallWidth = input.width;
```
davepagurek, line 1926: This can also be removed

src/webgl/p5.RendererGL.js
```diff
@@ -1874,6 +1905,85 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    // if one already exists for a given input image
+    if(this.diffusedTextures.get(input)!=null){
+      return this.diffusedTextures.get(input);
+    }
+    // if not, then create one
+    let newGraphic; // maybe switch to framebuffer
+    // draw the blurry image
+    // set the shader on the graphic and set the shader on the image
+    // this._renderer._applyTextProperties(img, newGraphic);
+    // make small width, hardcode to 200px
```
davepagurek, line 1924: We could mention that it's OK that it's small because it's going to be really blurry and smooth anyway

src/webgl/p5.RendererGL.js
```diff
@@ -1874,6 +1905,85 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    newGraphic.rectMode(newGraphic.CENTER);
+    newGraphic.rect(0, 0, newGraphic.width, newGraphic.height);
+    this.diffusedTextures.set(input, newGraphic);
+    return newGraphic;
+  }
+
+  /*
+   *  used in imageLight,
+   *  To create a texture from the input non blurry image,
+   *  Creating 8 different levels of textures according to different sizes and atoring them in `levels` array
```
davepagurek, line 1948: Maybe worth mentioning that they're different roughness levels too. The rougher it is, the smaller the size, because rougher surfaces are blurrier/less detailed

src/webgl/p5.RendererGL.js
```diff
@@ -1874,6 +1905,85 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    newGraphic.noStroke();
+    newGraphic.rectMode(newGraphic.CENTER);
+    newGraphic.rect(0, 0, newGraphic.width, newGraphic.height);
+    this.diffusedTextures.set(input, newGraphic);
+    return newGraphic;
+  }
+
+  /*
+   *  used in imageLight,
+   *  To create a texture from the input non blurry image,
```
davepagurek, line 1947: Similar to above, this only creates it if it doesn't already exist

src/webgl/p5.RendererGL.js
```diff
@@ -1920,6 +2030,9 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     fillShader.setUniform('uEmissive', this._useEmissiveMaterial);
     fillShader.setUniform('uShininess', this._useShininess);
 
+    // calling the _setImageLightUniforms from here
```
davepagurek, line 2033: This one probably doesn't need a comment, since it doesn't add much over the code itself

src/webgl/p5.RendererGL.js
```diff
@@ -1970,6 +2083,31 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     fillShader.bindTextures();
   }
 
+  // getting called from _setFillUniforms
+  _setImageLightUniforms(shader){
+    //set uniform values
+    if( this.activeImageLight == null ){
```
davepagurek, line 2089: This is just for debugging, right? We can probably remove this if block now

src/webgl/p5.RendererGL.js
```diff
@@ -1970,6 +2083,31 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+      console.log('activeImageLight prop is null');
+    }
+    shader.setUniform('uUseImageLight', this.activeImageLight != null );
+    // true
+    if (this.activeImageLight) {
+      // this.activeImageLight has image as a key
+      // look up the texture from the blurryTexture map
```
davepagurek, line 2096: Another spot where we can rename to `diffusedTextures` for consistency

src/webgl/p5.RendererGL.js
```diff
@@ -1970,6 +2083,31 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+      // change the name of shader of diffused to prevent confusion
```
davepagurek, line 2097: I think we can take this note out now

src/webgl/p5.RendererGL.js
```diff
@@ -1970,6 +2083,31 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+      let diffusedLight = this.getBlurryTexture(this.activeImageLight);
+      // change the name of this uniform as diff for diffused and specular
```
davepagurek, line 2099: I think we can take this one out too

src/webgl/p5.RendererGL.js
```diff
@@ -1970,6 +2083,31 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+      shader.setUniform('environmentMapDiffused', diffusedLight);
+      let specularLight = this.getSpecularTexture(this.activeImageLight);
+      // shine >= 1 so
+      // 0-1*8 = 0-8
+      // 0-8 * 20 = 0-160
+      let roughness = 20/this._useShininess;
+      shader.setUniform('levelOfDetail', roughness*8);
```
davepagurek, line 2106: We should mention that the 8 here is because we have 8 levels in the mipmap

src/webgl/shaders/imageLight.vert
```diff
@@ -0,0 +1,34 @@
+// add precision
```
davepagurek, line 1: Did we want to copy the precision from the fragment shader to here?

src/webgl/shaders/imageLightDiffused.frag
```diff
@@ -0,0 +1,82 @@
+precision mediump float;
```
davepagurek, line 1: oh actually we probably want this to be `highp`; this basically gets ignored by the webgl driver on desktop, but on mobile it's actually used, and in my experience any time we have normals, the precision makes a big difference on the visual fidelity
```suggestion
precision highp float;
```

src/webgl/shaders/imageLightDiffused.frag
```diff
@@ -0,0 +1,82 @@
...
+varying vec3 localPos;
+
+// the HDR cubemap converted (can be from an equirectangular environment map.)
+uniform sampler2D environmentMap;
+varying vec2 vTexCoord;
+
+const float PI = 3.14159265359;
+
+// not needed
+vec2 normalToEquirectangular( vec3 v){
```
davepagurek, line 11: We can remove this function since we aren't using it

src/webgl/shaders/lighting.glsl
```diff
@@ -67,6 +79,64 @@ LightResult _light(vec3 viewDirection, vec3 normal, vec3 lightVector) {
   return lr;
 }
 
+// converts the range of "value" from [min1 to max1] to [min2 to max2]
+float map(float value, float min1, float max1, float min2, float max2) {
+  return min2 + (value - min1) * (max2 - min2) / (max1 - min1);
+}
+
+vec2 mapTextureToNormal( vec3 normal ){
```
davepagurek, line 87: oh should we be using your implementation of `nTOE` from the other fragment shaders instead of this? If I recall correctly, this is the older implementation that has a bug where it only uses the 0-PI range, which you fixed in your updated `nTOE`.

I think using `nTOE` would also mean we can remove the `map` function above since this is all that uses it.

src/webgl/shaders/imageLightDiffused.frag
AryanKoundal: I'll manually do this, vscode has a feature for this but it's messes up the whole file. It's kinda like prettier, the way prettier messes up more than it cleans :")

src/webgl/shaders/lighting.glsl
AryanKoundal: Done, i'm testing after the changes. If nToE works, which it will very likely. I'll remove the `mapTextureToNormal` one.

src/webgl/shaders/lighting.glsl
```diff
@@ -67,6 +79,96 @@ LightResult _light(vec3 viewDirection, vec3 normal, vec3 lightVector) {
...
+  theta = acos(v.x / sin(phi));
+  float sinTheta = v.z / sin(phi);
+  if (sinTheta < 0.0) {
+    // Turn it into -theta, but in the 0-2PI range
+    theta = 2.0 * PI - theta;
+  }
+  theta = theta / (2.0 * 3.14159);
+  phi = phi / 3.14159 ;
+  
+  vec2 angles = vec2( phi, theta );
```
davepagurek, line 126: Looks like something might be up still:
image

I noticed that we had a note for diffused light to use the word normal instead of `R`. When I use that, it looks like this:
image

I tried using a two-toned background to check against and it seems like it's kind of flipping it:
image


That looks a bit more correct, but it looks like the green is on the side? I think that might be because in other files, we used theta and phi differently, but here, we want to return `vec2(theta, phi)` instead of the other way around (I think that makes sense that the x axis is theta here since that's the one whose range is 2*PI. That's different than in the other shaders, which is a little odd, but at least we tested those independently and they seem fine.) That looks like this:

image

so that still seems 90 degrees off, but in a different axis this time. I tried cycling -90 degrees in the theta axis. This seems to look right, and that's using `vec2 angles = vec2( fract(theta + 0.75), 1.0 - phi );`
image

With just diffuse: 
image

With specular:
image

```suggestion
  vec2 angles = vec2( fract(theta + 0.75), 1.0 - phi );
```

src/webgl/shaders/lighting.glsl
```diff
@@ -67,6 +79,96 @@ LightResult _light(vec3 viewDirection, vec3 normal, vec3 lightVector) {
...
+  // use worldNormal instead of R
+  // added nToE instead of this, remove this completely if the
+  // new function works
+  // vec2 newTexCoor = mapTextureToNormal( R );
+  vec2 newTexCoor = nToE( R );
+  vec4 texture = TEXTURE( environmentMapDiffused, newTexCoor );
+  // return texture.xyz;
+  // this is to make the darker sections more dark
+  // png and jpg usually flatten the brightness so it is to reverse that
+  return pow(texture.xyz, vec3(10.0));
```
davepagurek, line 146: I tried changing `imageLight()` with `lights()` in a sketch to see if they could be roughly equivalent. I think 10 might be a bit too aggressive for the diffuse light unfortunately. I tried playing around with some other curves:

<table>
<tr>
<th>Lighting call</th>
<th>

```glsl
pow(
  texture.xyz,
  vec3(10.0)
);
```

</th>
<th>

```glsl
texture.xyz;
```

</th>
<th>

```glsl
smoothstep(
  vec3(0.),
  vec3(0.8),
  texture.xyz
);
```

</th>
<tr>
<td>

```js
imageLight(img);
```

</td>
<td>

image

</td>
<td>

image

</td>
<td>

image

</td>
</tr>
<tr>
<td>

```js
lights()
```

</td>
<td>

image

</td>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

So although it's not physically accurate or anything, maybe doing this will help make it easier for people to switch from `lights()` to `imageLight()`? This maps texture values from 0-0.8 to 0-1 along a smooth curve:
```suggestion
  return smoothstep(vec3(0.0), vec3(0.8), texture.xyz);
```

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
...
+ * <div class="notest">
+ * <code>
+ * let img;
+ * let slider;
+ * function preload() {
+ *   img = loadImage('assets/laDefense.jpg');
+ * }
+ * function setup() {
+ *   createCanvas(windowWidth, 400, WEBGL);
+ *   slider = createSlider(0, 800, 100, 1);
```
davepagurek, line 558: I feel like I don't notice much change in visual after going halfway through the slider, so maybe for this example a range of 0-400 would be better?

davepagurek (commented):
The documentation is looking great! just some very minor punctuation things for that.

I did a bit of debugging of the `nToE` function. It looks like it has to be a bit different from the other files to work properly? but it seems like there's a value that makes it work.

I also left some suggestions about the power-of-10 thing for diffuse light, let me know what you think!

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
   return this;
 };
 
+/**
+ * Creates an image light with the given image.
+ *
+ * The image light stimulates light from all the directions.
+ * This is done by using the image as a texture for a relatively
+ * very huge light in a form of a sphere. This sphere contains
```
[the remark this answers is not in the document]
AryanKoundal: done

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
...
+ * or encapsulates the whole scene/drawing.
+ * It will have different effect for varying shininess of the
+ * object in the drawing.
+ * Under the hood it is mainly doing 2 types of calculations,
+ * first one is creating an irradiance map from the
```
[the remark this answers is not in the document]
AryanKoundal: done

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
...
+ * environment map of the input image.
+ * Second one is managing reflections based on the shininess
```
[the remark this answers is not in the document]
AryanKoundal: done

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
...
+ * or roughness of the material used in the scene.
+ *
+ * Note: The image's diffuse light will be affected by fill()
+ * and the specular reflections will be affected by specularMaterial()
+ * and shininess()
```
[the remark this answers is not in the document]
AryanKoundal: done

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
...
+ * @param  {p5.image}    img  image for the background
+ * @example
+ * <div class="notest">
+ * <code>
+ * let img;
+ * function preload() {
+ *   img = loadImage('assets/laDefense.jpg');
+ * }
+ * function setup() {
+ *   createCanvas(windowWidth, 400, WEBGL);
```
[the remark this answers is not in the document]
AryanKoundal: done

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
...
+ * }
+ * function draw() {
+ *   background(220);
+ *   imageMode(CENTER);
+ *   image(img, 0, 0, width, height);
```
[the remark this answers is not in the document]
AryanKoundal: done

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
...
+ *   ambientLight(50);
+ *   imageLight(img);
+ *   // This will use specular once we add it
+ *   specularMaterial(20);
+ *   // shininess(slider.value());
```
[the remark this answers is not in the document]
AryanKoundal: done

src/webgl/light.js
AryanKoundal: makes sense, done

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
...
+ *   // This will use specular once we add it
+ *   specularMaterial(20);
+ *   shininess(slider.value());
+ *   // orbitControl();
+ *   noStroke();
+ *   scale(2);
+ *   // sphere(50)
+ *   rotateX(frameCount * 0.005);
+ *   rotateY(frameCount * 0.005);
+ *   box(50);
```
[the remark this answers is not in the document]
AryanKoundal: Done, removed the rotate functions, and changed the shape to sphere.

src/webgl/light.js
```diff
@@ -494,6 +494,12 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
   return this;
 };
 
+p5.prototype.imageLight = function(img){
```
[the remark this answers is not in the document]
AryanKoundal: done

src/webgl/light.js
AryanKoundal: done

src/webgl/p5.RendererGL.js
AryanKoundal: done

src/webgl/shaders/lighting.glsl
AryanKoundal: done, changed the `nToE` name to `mapTextureToNormal` for better understanding name.

src/webgl/shaders/lighting.glsl
AryanKoundal: Changed this line, but do i have to change the line 
```
  // use worldNormal instead of R
  vec2 newTexCoor = mapTextureToNormal( R );
```
Did you use worldNormal or R in this function?

src/webgl/shaders/lighting.glsl
AryanKoundal: Done for `caluculateImageDiffuse`, should I do this for `calculateImageSpecular` are well?

src/webgl/light.js
```diff
@@ -494,6 +494,99 @@ p5.prototype.pointLight = function(v1, v2, v3, x, y, z) {
   return this;
 };
 
+/**
+ * Creates an image light with the given image.
+ *
+ * The image light stimulates light from all the directions.
```
davepagurek, line 500:
```suggestion
 * The image light simulates light from all the directions.
```

src/webgl/shaders/lighting.glsl
davepagurek: I think specular looks ok how it is, the inconsistency is fine as long as it looks ok.

src/webgl/shaders/lighting.glsl
davepagurek: > Did you use worldNormal or R in this function?

I think that one should be `worldNormal`.

I double checked the suggested code and I think maybe I was testing without that by accident though -- I think when using the world normal, to get the example with the stripes to work, it would be this:

```glsl
vec2 angles = vec2( fract(theta + 0.25), 1.0 - phi );
```

(Basically, 0.25 instead of 0.75, which is like rotating by pi.)

davepagurek (commented):
The examples look great!

src/webgl/light.js
AryanKoundal: > 

done

davepagurek (commented):
One last thing, do you think for the example image we use, we should use an actual sphere map image that wraps correctly so users know what kind of input this is intended for? e.g. something from https://polyhaven.com/hdris since the license for these is free and requires no attribution?

src/webgl/shaders/lighting.glsl
```diff
@@ -67,6 +79,63 @@ LightResult _light(vec3 viewDirection, vec3 normal, vec3 lightVector) {
...
+
+
+vec3 calculateImageDiffuse( vec3 vNormal, vec3 vViewPosition ){
+  // make 2 seperate builds 
+  vec3 worldCameraPosition =  vec3(0.0, 0.0, 0.0);  // hardcoded world camera position
+  vec3 worldNormal = normalize(vNormal);
+  vec3 lightDirection = normalize( vViewPosition - worldCameraPosition );
+  vec3 R = reflect(lightDirection, worldNormal);
+  // use worldNormal instead of R
+  vec2 newTexCoor = mapTextureToNormal( R );
```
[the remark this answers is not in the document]
AryanKoundal: Done, I have commented out the lines to test it. If the tests pass then I'll completely remove this section.

AryanKoundal:
> One last thing, do you think for the example image we use, we should use an actual sphere map image that wraps correctly so users know what kind of input this is intended for? e.g. something from https://polyhaven.com/hdris since the license for these is free and requires no attribution?

I agree with you. Those sphere map images would give a better understanding of the feature. 
But also would that create confusion among users? They might feel that they need a sphere map image for the background they want to use(But in reality even a simple image would work fine as well). What do you think?

davepagurek:
We talked about this on our call, but just adding it again here for anyone who might be reading: we've decided to use a sphere map for one example and a regular image for another.

davepagurek:
Maybe we can scale down the sphere map a bit? Right now it's 8000px wide, we probably don't need quite so many pixels 😅

PR #6259, opened 2023-07-08
In p5.Camera.slerp, change the eye-center distance interpolation to be logarithmically based

inaridarkfox4231:
Resolves: #6257

## Changes:
In the p5.Camera.slerp() logic, the interpolation of the distance between the viewpoint and the center used linear interpolation.
However, with this interpolation method, there are problems such as the camera breaking when it goes out of range.
Also, when applied to infinite loop animation, there is also the problem that the animation does not connect smoothly.
Furthermore, since orbitControl() changes the distance logarithmically, it is not very compatible with it.
From the above, I came to the conclusion that it is better to use logarithmic interpolation instead of linear interpolation for the distance interpolation method.
Also, along with that, added the content about distance interpolation in the unit test.

## Screenshots of the change:
[logarithmic interpolation](https://editor.p5js.org/dark_fox/sketches/ieOg7Lcy7)
```js
let cam,cam0,cam1;
function setup(){

  cam = createCamera();
  cam.camera(200,200,200,0,0,0,0,0,-1);
  cam.perspective(PI/3,width/height,0.1,1000);
  //cam.ortho(-200,200,-200,200,0.1,1000);

  cam0 = createCamera();
  cam0.set(cam);

  cam1 = createCamera();
  cam1.set(cam);
  cam1.camera(50,50,50,0,0,0,0,0,-1);
  //cam1.ortho(-50,50,-50,50,0.1,1000);
  
  setCamera(cam);
}

function draw(){
  noStroke();
  directionalLight(color("gray"),-9,-6,-1);
  ambientLight(color("gray"));
  fill("white");
  ambientMaterial("white");
  
  background(255);
  const p = frameCount % 60 / 60;
  cam.slerp(cam0, cam1, p);
  
  for(let i=0;i<8;i++){
    translate(0,-800,0);box(800);translate(0,800,0);
    translate(-800,0,0);box(800);translate(800,0,0);
    translate(0,0,-800);box(800);translate(0,0,800);
    scale(1/4);
  }
}
```

https://github.com/processing/p5.js/assets/39549290/0cbb6980-c8e3-4703-b04a-c7caba5f1f7d

Although it is a subtle difference, slerp2 (version with specification change) is smoother.
In addition, there is almost no difference in appearance in sketches such as resetting orbitControl().

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/p5.Camera.js
+++ b/src/webgl/p5.Camera.js
@@ -1648,7 +1648,8 @@ p5.Camera = class Camera {
  * When applying this function, all cameras involved must have exactly the same projection
  * settings. For example, if one is perspective, ortho, frustum, the other two must also be
  * perspective, ortho, frustum respectively. However, if all cameras have ortho settings,
- * only orbitControl's change of the projection settings is permissive.
+ * interpolation is possible if the ratios of left, right, top and bottom are equal to each other.
+ * For example, when it is changed by orbitControl().
  *
  * @method slerp
  * @param {p5.Camera} cam0 first p5.Camera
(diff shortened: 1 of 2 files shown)


inaridarkfox4231:
It seems fine... Please leave a review.

inaridarkfox4231:
I called logarithmic interpolation, but I don't know the official way to call it..., so I'm just calling it that way, and I don't know if it's appropriate. This means that for each value we take the logarithm in the appropriate base, linearly interpolate it, and return it in the same base.
This assumes that the ratio value is positive. For example, changes in components by orbitControl() satisfy that requirement.

inaridarkfox4231:
For final performance of p5.Camera.slerp(), in my environment it looked like this:

10000 times per frame: almost 40fps
5000 times per frame: 58～59fps

Although the performance has been degraded by rewriting it with matrix operations, I didn't think it would be a problem because it wasn't a function that would be called that many times. Readability is more important. Logarithmic interpolation slows down a little, but there is no problem because it is more convenient in various ways.

davepagurek (approved):

inaridarkfox4231:
Thank you for your review and merging! ('ω')
Thank you very much for reviewing many pull requests. I would like to look forward the release of next version...

PR #6276, opened 2023-07-23
Add support for webGL instancing

### LINKED ISSUE #6275: Instancing from WebGL2
### Increasing Access

This feature would allow for programs to run with better performance and for a lot more possibilities within the p5.js framework.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

### Feature request details

Add a method that allows you to instance a certain amount of objects and let the user handle the rest through their own shader code.

RandomGamingDev:
Resolves #6275

 Changes:

This change adds instancing from WebGL2 to `endShape()` allowing for more performant rendering and more possibilities overall.

 Screenshots of the change:

Here's an example of instancing. The shapes below the rotated cube were all drawn within a single draw call.
Instancing

#### PR Checklist

- [x] `npm run lint` passes

(I haven't had time to write a unit test yet nor am I sure what specific unit test to write)

This addition is based on a library that I wrote that works with the current latest release version of p5.js and that has its own demo.
https://github.com/RandomGamingDev/WebGLp5Instancing


aferriss:
This is an interesting proposal, it would be good to see the accessibility reasoning behind this feature as well in the issue that you opened. I think performance is a valid argument but maybe there's something else as well. 

Personally, I think that it might make more sense to add the count argument into beginShape() instead of endshape(), but I'm open to being wrong about that if there's a good reason.

Can you also share some examples in the p5 web editor using the feature that we can play with? If this feature is accepted it will need examples anyway,  and it would be good to get an idea of how this looks in the shader, and what will be required of the user.

Should instanced geometry also be provided some instanceId attribute so that the user can manipulate instances within the shader? Furthermore, wouldn't this feature be more useful if there was an interface to create generic attribute buffers for instanced geometry? Otherwise, how is a user intended to translate / rotate / scale individual instances?

For unit tests you could think about testing that instanced geometry looks the same as non instanced, and that all the different geometry modes work correctly. Also test things like fill and stroke.

src/core/shape/vertex.js
```diff
@@ -612,18 +613,22 @@ p5.prototype.endContour = function() {
...
-      shapeKind
+      this.isCurve,
+      this.isBezier,
+      this.isQuadratic,
+      this.isContour,
+      this.shapeKind,
+      count
     );
   } else {
+    if (count !== 1) {
```
aferriss, line 629: I think it could be good to catch if the user tries to use a count < 1 and either log an error or silently fall back to just using a count of 1

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -391,18 +392,34 @@ p5.RendererGL.prototype._drawImmediateFill = function() {
   // WebGL 1 doesn't support the QUADS and QUAD_STRIP modes, so we
   // need to convert them to a supported format. In `vertex()`, we reformat
   // the input data into the formats specified below.
-  if (this.immediateMode.shapeMode === constants.QUADS) {
-    this.immediateMode.shapeMode = constants.TRIANGLES;
-  } else if (this.immediateMode.shapeMode === constants.QUAD_STRIP) {
-    this.immediateMode.shapeMode = constants.TRIANGLE_STRIP;
+  if (this.immediateMode.shapeMode === QUADS) {
```
aferriss, line 395: Why remove the .constants?

src/core/shape/vertex.js
```diff
@@ -589,6 +589,7 @@ p5.prototype.endContour = function() {
  *
  * @method endShape
  * @param  {Constant} [mode] use CLOSE to close the shape
+ * @param  {Integer} [count] number of times you want to draw/instance the shape (for WebGL mode).
```
aferriss, line 592: Can this be renamed to "instanceCount" or something a little more descriptive?

src/webgl/p5.RendererGL.Immediate.js
RandomGamingDev: Sorry, that must've been a mistake from when I left them out for testing. I fixed it in the latest commit.

src/core/shape/vertex.js
RandomGamingDev: I was thinking of letting the WebGL implementation do the work since a count less than or equal to 0 could be valid, but prioritizing code safety does make more sense. I'll fix that in a new commit.

RandomGamingDev:
I think that it'd work better with `endShape()`. I find it a bit off that beginShape is used to change the shape type and endShape to determine whether or not it closes, but understand that that's how you'd naturally think when you draw a shape. Following this logic and the logic of WebGL/OpenGL, it'd still make more sense to do the instancing at the end when everything's actually being drawn, and when you'd think about making duplicates a lot of the time.

This feature doesn't have to be limited to custom shapes, but since you oftentimes have to deal with vertices themselves, and need shader and WebGL knowledge to get it working, not to mention the fact that it could adversely effect p5.js in other unforseen ways. That's why I don't believe it'd make sense to add this feature to other shapes just to increase the complexity on the p5.js side and the confusion on the end user's side.

Shaders aren't applied to strokes, so I don't see why instancing, which is a shader-centric idea would be applied to strokes either, especially since it wouldn't have any effect unless the strokes were transparent, and even then have very little of an effect. I feel like the strokes, when using instancing would rather be more for debugging, especially since an experienced user of shaders would be able to add their own strokes in, strokes that would most likely fit better than p5.js strokes.

For the demo, I have a basic one in the library, although I do have one here that I used for debugging: https://editor.p5js.org/PotatoBoy/sketches/bC6WITRDd

The instance id attribute is handled by WebGL 2 using gl_InstanceID inside of the shader.

It would indeed be more useful with those, however it would raise the bar to use instancing, which is a very important feature in my opinion. I think that we should focus on making it more accessible. For those who need attributes, they can use constants, uniforms, and textures to store their data, or modify the p5.js library themselves until instancing gets added with attribute buffers.

As for the performance increase, create a bunch of objects and compare that to drawing them without instancing and the performance can be seen within chrome.

src/core/shape/vertex.js
RandomGamingDev: The problem I have with the name `instanceCount` is that it could be taken the wrong way. Perhaps something like `numInstances` would work, but I was thinking of keeping its name simple and understandable since drawing smth a certain amount of times makes sense in my view. If that's a problem then I'd probably change it to `numInstances`, but it might be less understandable rather than more.

src/core/shape/vertex.js
RandomGamingDev: It's been added to the front of the function below the parameter check in the latest commit.

src/core/shape/vertex.js
RandomGamingDev: If u look at the rest of the documentation most of the parameter names are decently simple.

davepagurek:
> Personally, I think that it might make more sense to add the count argument into beginShape() instead of endshape(), but I'm open to being wrong about that if there's a good reason.

I don't have a strong opinion on this one, but maybe just think about what future extensions to this we might want to make (e.g. binding some per-instance attributes, or [having a per-instance transform matrix like Threejs does](https://threejs.org/docs/#api/en/objects/InstancedMesh.instanceMatrix)) and see if it would make it easier to do so in one location or the other (or if it makes no difference.) Any thoughts on what those might include @RandomGamingDev?

> Why would this feature need to be limited to custom shapes? Wouldn't it be more useful to instance things like lines or other basic shapes?

I think it might end up being easier to optimize performance for a feature like this if it's on a per shape level rather than something that happens inside of a primitive like a stroke. This way, we know for sure the user is intending to draw a bunch of the same thing, which should hopefully be a very clear optimization, whereas we've found it hard to get across-the-board performance improvements by adding instancing to all lines.

I do wonder, though, if we can do this in a way that works with `model()` too, as one of the things I'm going to be working on for the next release is a way to more easily make retained mode shapes, and pairing that with instancing could yield some great performance gains for particle system sorts of sketches.

> Shaders aren't applied to strokes, so I don't see why instancing, which is a shader-centric idea would be applied to strokes either

That's true for now while a shader + instance ID is the only way to make use of instanced rendering. If we think the most common use of instancing is to have a separate transformation matrix per instance, I can see us wanting to make a (non-shader) API specifically for this so that people don't have to write a shader to make use of it, and then we'd probably want the fills and strokes to both take that into account. That's probably the north star I'd want to work towards to balance usefulness and accessibility -- doesn't cover all cases, but would be much easier to write than making a custom shader, [especially while we don't have an easy way for users to augment our existing shaders without rewriting them.](https://github.com/processing/p5.js/issues/6144)

> As for the performance increase, create a bunch of objects and compare that to drawing them without instancing and the performance can be seen within chrome.

I think it would be good to have a minimum example of this both to use as a benchmark, and also to add to the p5 reference. As is, if a user adds a count to `endShape()`, all the instances would get drawn to the same spot and it would look like it's not working, so I think we need to provide a bit more documentation guidance on how the feature is intended to be used.

RandomGamingDev:
I think those would include a system for adding attributes using `beginShape()`, `vertex()`, and `endShape()` since those r decently limited rn with the rest being handled by unifroms and textures.
Whether it's on a shape level or not shouldn't matter too much as long as the person using instancing knows enough to optimize for instancing. It would be possible to do it with `model()`, but it'd require modifying that function.
It'd be useful, but I'm not too sure what the reaction would be to modifying so much of the p5.js framework, especially for smth that a lot of people won't know how to use. If we want to go with the general API approach, we'd have to deal with things like position, size, rotation, scale, among other things. If we were to go through that route we'd most likely want to store that data in a texture that can then later be queried, but I'm not sure that that'd be too effective.
Alr, I'll add documentation later when I can.

davepagurek:
> It would be possible to do it with model(), but it'd require modifying that function.

Makes sense, that could be a good feature to add to `model()`, but isn't required to be part of this PR.

> It'd be useful, but I'm not too sure what the reaction would be to modifying so much of the p5.js framework, especially for smth that a lot of people won't know how to use. If we want to go with the general API approach, we'd have to deal with things like position, size, rotation, scale, among other things. If we were to go through that route we'd most likely want to store that data in a texture that can then later be queried, but I'm not sure that that'd be too effective.

Solving for a more general case would maybe be better served by making an API for custom vertex attributes and allowing [the divisor](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/vertexAttribDivisor) to be changed so that it uses the same value for each instance. Looks like we could add that (and even add some nicer APIs for transformations specifically on top of that if we want) without running into issues here, which is good!

src/core/shape/vertex.js
```diff
@@ -559,6 +559,62 @@ p5.prototype.curveVertex = function(...args) {
  * </code>
  * </div>
  *
+ * @example
```
davepagurek, line 562: I think this is in the `endContour()` docs, did you want this example to be in `endShape()`?

davepagurek:
Looks like there are some trailing spaces in the code causing the linter test to fail:
image

src/core/shape/vertex.js
RandomGamingDev: Yeah, sorry about that. I fixed it in the latest commit.

src/core/shape/vertex.js
```diff
@@ -612,18 +669,27 @@ p5.prototype.endContour = function() {
...
+
   if (this._renderer.isP3D) {
     this._renderer.endShape(
       mode,
-      isCurve,
-      isBezier,
-      isQuadratic,
-      isContour,
-      shapeKind
+      this.isCurve,
```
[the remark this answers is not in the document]
RandomGamingDev: That was a mistake. I'll remove those.

davepagurek:
@aferriss I'm thinking of merging this in as-is, knowing that for now it requires custom shaders to take advantage of instancing, and that this is only currently feasible for fills. Going forward, I'd like to improve upon both of those points either by:
1. making a way for users to augment our existing shaders without completely rewriting them (https://github.com/processing/p5.js/issues/6144) so that one could add some per-instance positioning logic
2. taking the threejs approach of making it easy to have [per-instance transforms](https://threejs.org/docs/#api/en/objects/InstancedMesh.instanceMatrix) and adding support for that into our default shaders. Any other properties requiring per-instance variation would still have to be done with a custom shader, which will work the same as it currently does, and might also benefit from (1).

Do any of those approaches sound ok to you? This current PR would be a step towards either of them.

RandomGamingDev:
That sounds good.

src/core/shape/vertex.js
aferriss: I was mainly thinking about the renaming in the context of helping future maintainers understand what this is for. I still think `count` is too ambiguous in this context. `numInstances` sounds fine to me if you prefer that.

src/core/shape/vertex.js
```diff
@@ -609,21 +610,86 @@ p5.prototype.endContour = function() {
...
+ * in vec3 aPosition;
+ *
+ * uniform mat4 uModelViewMatrix;
+ * uniform mat4 uProjectionMatrix;
+ *
+ * void main() {
+ *
+ *   vec4 positionVec4 = vec4(aPosition, 1.0);
+ *   gl_Position = uProjectionMatrix * uModelViewMatrix * positionVec4 -
+ *   vec4(vec2(float(gl_InstanceID) * 100.0, 0.0), 0.0, 0.0);
```
aferriss, line 630: Can you add some comments to this shader to explain what's happening here? Maybe there's somewhere we can also add gl_InstanceID to the documentation?

src/core/shape/vertex.js
```diff
@@ -609,21 +610,86 @@ p5.prototype.endContour = function() {
...
+ * out vec4 outColor;
+ *
+ * void main() {
+ *   outColor = vec4(255, 0, 0, 255);
+ * }
+ * `;
+ *
+ * function setup() {
+ *   createCanvas(400, 400, WEBGL);
+ *   // Please note that it's best practice to put shaders in other files
```
aferriss, line 646: I think this is a stylistic choice, we can remove this comment.

src/core/shape/vertex.js
```diff
@@ -609,21 +610,86 @@ p5.prototype.endContour = function() {
...
       isBezier,
       isQuadratic,
       isContour,
-      shapeKind
+      shapeKind,
+      count
     );
   } else {
+    if (count !== 1) {
+      console.log('🌸 p5.js says: Instancing is only supported in WebGL2 mode');
```
aferriss, line 691: Are we actually checking for webGL2 mode anywhere? Wouldn't this just log if the user wasn't in webGL mode at all?

aferriss:
@davepagurek That sounds fine to me. Before we merge though we still need to have a few more things:
- complete documentation for this feature, including: explanation of instancing, feature support (strokes not supported for ex), commented examples. I left more comments throughout the PR.
- unit tests
- error check for webGL2

RandomGamingDev:
Yeah, I'll add more than just the example and also unit tests soon.
There's already a check for WebGL2 mode in `_drawImmediateFill` for if the instancing function fails since then we can be decently sure that they aren't using WebGL2.

src/core/shape/vertex.js
RandomGamingDev: This logs if the user isn't using WebGL mode at all, but there's already a check in place in `_drawImmediateFill()` which will trigger if trying to instance causes an error.

src/core/shape/vertex.js
RandomGamingDev: alr

src/core/shape/vertex.js
```diff
@@ -609,21 +610,86 @@ p5.prototype.endContour = function() {
...
+ * function setup() {
+ *   createCanvas(400, 400, WEBGL);
+ *   // Please note that it's best practice to put shaders in other files
+ *   fx = createShader(vs, fs);
+ * }
+ *
+ * function draw() {
+ *   background(220);
+ *
+ *   // strokes aren't instanced, and are rather used for debug purposes
```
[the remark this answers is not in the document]
RandomGamingDev: alr I documented it in the top portion of text

RandomGamingDev:
I added the documentation updates and I'll add the unit test when I get the time for it.

src/core/shape/vertex.js
```diff
@@ -626,8 +627,12 @@ p5.prototype.endContour = function() {
  * void main() {
  *
  *   vec4 positionVec4 = vec4(aPosition, 1.0);
+ * 
+ *   // gl_InstanceID represents a numeric value for each instance
+ *   // using gl_InstanceID allows us to move each instance separately
+ *   // here we move each instance horizontally by id * 100
  *   gl_Position = uProjectionMatrix * uModelViewMatrix * positionVec4 -
- *   vec4(vec2(float(gl_InstanceID) * 100.0, 0.0), 0.0, 0.0);
+ *     vec4(vec2(float(gl_InstanceID) * 100.0, 0.0), 0.0, 0.0);
```
aferriss, line 635: can you simplify this line to be `vec4(float(gl_InstanceID) * 100.0, 0.0, 0.0, 0.0);`. There's no need for the vec2 here.

RandomGamingDev: That looks good

RandomGamingDev:
Oh forgot to update the thing. The unit tests were added.

src/core/shape/vertex.js
```diff
@@ -594,9 +594,11 @@ p5.prototype.endContour = function() {
  * data defined since the previous call to <a href="#/p5/beginShape">beginShape()</a> is written into the image
  * buffer. The constant CLOSE as the value for the `mode` parameter to close
  * the shape (to connect the beginning and the end).
+ * When using instancing with <a href="#/p5/endShape">endShape()</a> the instancing will not apply to the strokes.
```
aferriss, line 597: This is where we still need documentation explaining what instancing is, what it does, and when to use it. 

"When the `count` parameter is used with a value greater than 1, it enables instancing for shapes built when in WEBGL mode. Instancing is a feature that allows the GPU to efficiently draw multiples of the same shape. It's often used for particle effects or other times when you need a lot of repetition. In order to take advantage of instancing, you will also need to write your own custom shader using the `gl_InstanceID` keyword. You can read more about instancing [here](https://webglfundamentals.org/webgl/lessons/webgl-instanced-drawing.html) or by working from the example on this page."

RandomGamingDev: Added the documentation u asked for.

aferriss (approved):
Thanks @RandomGamingDev I think it's looking good! Appreciate the patience :D

src/core/shape/vertex.js
```diff
@@ -651,9 +654,18 @@ p5.prototype.endContour = function() {
  * precision mediump float;
  *
  * out vec4 outColor;
+ * in int instanceID;
```
davepagurek, line 663: Not 100% sure if this is why this example is throwing an error compiling the shader, but it looks like the vertex shader uses a lowercase d in `Id` while the fragment shader capitalizes it

RandomGamingDev:
I made it so that the shader compiled by changing the input and output names to be the same for `instanceID` and by using `flat` instead.

src/core/shape/vertex.js
RandomGamingDev: That's on reason why, but int's also can't be interpolated like that so ur also supposed to use `flat`.

aferriss: Ah sorry, that's my fault and what I get for editing the code inline in browser. Thanks for catching!

davepagurek:
Great work on this @RandomGamingDev!

PR #6302, opened 2023-07-27
fix textbound invalid input

### LINKED ISSUE #6298: textBounds outputs nan if number is input instead of string

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [X] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

### p5.js version

1.7.0

### Web browser and version

FF 115.0.2

### Operating System

mac 12.6

### Steps:
1. input number to textBounds

### Snippet:

https://editor.p5js.org/micuat/sketches/sYWYkUTJv

```javascript
let f;
function preload() {
  f = loadFont("RobotoMono-VariableFont_wght.ttf")
}
function setup() {
  createCanvas(400, 400);
  console.log(f.textBounds("io", 0, 0, 10))
  console.log(f.textBounds(10, 0, 0, 10))
}

function draw() {
  background(220);
}
```

console:

```
{x: 0.9912109375, y: -7.2021484375, h: 7.2998046875, w: 10.41015625, advance: 0.9912109375}

{x: Infinity, y: Infinity, h: -Infinity, w: 0, advance: Infinity}
```

I'm not sure what is the expected result (number + "" would solve the issue). If number is not accepted, there should be an error message, I suppose?

adityagarg06:
Resolves #6298 

 Changes:
Added an additional check for invalid input. Also I have written test for the same

- [x] `npm run lint` passes

### DIFF
--- a/src/typography/p5.Font.js
+++ b/src/typography/p5.Font.js
@@ -76,6 +76,10 @@ p5.Font = class {
   // Check cache for existing bounds. Take into consideration the text alignment
   // settings. Default alignment should match opentype's origin: left-aligned &
   // alphabetic baseline.
+    if (typeof str !== 'string') {
+      console.error('Error: Text parameter must be a string.');
+      return null;
+    }
     const p = (opts && opts.renderer && opts.renderer._pInst) || this.parent;
 
     const ctx = p._renderer.drawingContext;
--- a/test/unit/typography/p5.Font.js
+++ b/test/unit/typography/p5.Font.js
@@ -34,6 +34,35 @@ suite('p5.Font', function() {
       assert.property(bbox, 'w');
       assert.property(bbox, 'h');
     });
+
+    test('returns null and logs an error when a number is passed instead of text', async function() {
+      const errorLog = await promisedSketch(function(sketch, resolve, reject){
+        let _font;
+        let invalidText = 10;
+        sketch.preload = function() {
+          _font = sketch.loadFont('path/to/valid/font.ttf', function() {}, reject);
+        };
+        sketch.setup = function() {
+          console.error = function(message) {
+            // Check if the error message contains the expected substring
+            if (message.includes('Error: Text parameter must be a string.')) {
+              resolve(true);
+            } else {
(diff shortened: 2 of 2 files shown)


src/typography/p5.Font.js
```diff
@@ -76,6 +76,10 @@ p5.Font = class {
   // Check cache for existing bounds. Take into consideration the text alignment
   // settings. Default alignment should match opentype's origin: left-aligned &
   // alphabetic baseline.
+    if (typeof str !== 'string') {
+      console.error('Error: Text parameter must be a string.');
+      return null;
+    }
```
limzykenneth, line 82: We don't usually handle incorrect parameter types this way as the FES is usually the one to take care of it. In terms of logic it should either do type coersion or return sensible default values, `null` is probably not useful in this case and can cause a property access error on the user's end.

limzykenneth (changes requested):

micuat:
to me, coercion into string makes sense in case of a number. but then what if the input is an object, for example?

munusshih:
I think this should be handled in FES. Making an exception for this function might be a workaround for now, but doesn't seem to provide a general solution for this types of problem.

Qianqianye:
> I think this should be handled in FES. Making an exception for this function might be a workaround for now, but doesn't seem to provide a general solution for this types of problem.

Thanks @munusshih. I'm wondering what the GSoC FES contributor and mentors think about this @almchung @nbriz @Ayush23Dash

Ayush23Dash:
+1 to what @limzykenneth said. Returning a null is not a good solution. Moreover, this might solve the issue temporarily but we need to have a general solution rather than a customised one.

dhowe:
status ?

limzykenneth:
@dhowe I'll close this for now as the implemented solution is not necessarily what we needed.

PR #6321, opened 2023-08-03
Clean up gruntfile release related steps

Qianqianye:


limzykenneth:
The [ci-lint.yml](https://github.com/processing/p5.js/blob/main/.github/workflows/ci-lint.yml) file will need to be modified so that `npm run list:source` is `npm run lint` instead, and the [ci-lint-examples.yml] file deleted. Since we are just running the whole lint as one step, no real need to split them up like that.

Qianqianye:
Thanks Ken.

PR #6324, opened 2023-08-03
Add shaders for filter() constants, and use them by default in P2D

### LINKED ISSUE #4820: Suggestion: provide shader for each p5js filter

wong-justin:
Resolves #4820 
Continuation of #6237 

 Changes (in progress):
- add fragment shader source files for BLUR, INVERT, etc
- replace original pixels `filter()` implementation with those shader filters
- add option to revert back to original pixels `filter()`

(This PR is a working draft that's part of the GSoC 2023 project 'Supporting shader-based filters in p5.js", mentored by @aferriss and @aceslowman)

---

#### PR Checklist

- [ ] `npm run lint` passes


wong-justin:
New `filter()` signature with this approach:
image

aceslowman:
I think this is looking good, and from what I can see, I don't think useWebGL is adding too much complexity, it looks like you've gone about it in a straightforward way.

src/webgl/shaders/filters/posterize.frag
wong-justin, line 1: It mostly works, but it looks different (and a little worse imo) compared to the original pixels POSTERIZE.
[example sketch with old pixels filter](https://editor.p5js.org/jwong/sketches/ZsWF6Gxd6)
[example sketch with this new shader](https://editor.p5js.org/jwong/sketches/r9fxPerPU)

Pixels filter code for comparison:
https://github.com/processing/p5.js/blob/6d3405fb08bef8673f9c5b7d2b11d39418f79cad/src/image/filters.js#L289-L308

- Not sure what the three bitshifts are doing this time; I've seen bitshifts of 8,16,24 and know that's for accessing different color channels, but these are different. Also there's no bitshifts with GLSL floats
- Not sure how to handle validating the extra filter parameter; here it throws a vanilla error when outside 2-255. But others like BLUR and THRESHOLD just do their own little validation and fail silently if the parameter is invalid

src/webgl/shaders/filters/erode.frag
```diff
@@ -0,0 +1,38 @@
...
+  vec4 neighbors[4];
+  neighbors[0] = texture2D(tex0, vTexCoord + vec2( texelSize.x, 0.0));
+  neighbors[1] = texture2D(tex0, vTexCoord + vec2(-texelSize.x, 0.0));
+  neighbors[2] = texture2D(tex0, vTexCoord + vec2(0.0,  texelSize.y));
+  neighbors[3] = texture2D(tex0, vTexCoord + vec2(0.0, -texelSize.y));
+
+  for (int i = 0; i < 4; i++) {
+    vec4 color = neighbors[i];
+    float lum = luma(color.rgb);
+    if (lum < curLuminance) {
```
wong-justin, line 31: Not sure if the `if` statement should be replaced by some sort of `step` logic. I know it's good to avoid branching if statements in shaders for speed. But a simple `if` statement seems easier to read and maintain

wong-justin:
I think I documented all the important features and behaviors between the `filter()` and `createFilterShader()` pages. I also tried to keep `filter()` a little less technical or webgl-centric, at least until the bottom, since that's more of an entry point page for beginners. The rest of the gotchas and explanations are in `createFilterShader()` for advanced users. Screenshots, with examples omitted for brevity:

image

image

Open to suggestions as always, especially @aceslowman - I remember you said it would be nice to have an earlier disclaimer about filter shaders not working on 3D geometries, for people coming from shader land?

src/webgl/p5.RendererGL.js
```diff
@@ -970,31 +985,42 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+        this.filterShader =
+          new p5.Shader(pg._renderer, userShader._vertSrc, userShader._fragSrc);
+        this.filterShader.parentShader = userShader;
+      }
     }
 
     // apply shader to pg
     pg.shader(this.filterShader);
     this.filterShader.setUniform('tex0', this);
+    this.filterShader.setUniform('texelSize', [1.0/this.width, 1.0/this.height]);
```
aferriss, line 1020: should we add `texelSize` to the documentation for `createFilterShader`?

src/webgl/shaders/filters/blur.frag
```diff
@@ -0,0 +1,50 @@
...
+  vec2 uv = vTexCoord;
+
+  // a single pass blur works by sampling all the neighbor pixels and averaging them up
+  // this is somewhat inefficient because we have to sample the texture 9 times -- texture2D calls are slow :( 
+  // check out the two-pass-blur example for a better blur approach
+  // get the webcam as a vec4 using texture2D
+
+  // spread controls how far away from the center we should pull a sample from
+  // you will start to see artifacts if you crank this up too high
+  float spread = 4.0;
```
aferriss, line 25: lets just simplify this whole block to `float spread = max(0.0, filterParameter);` and then set a default value for the filterParameter in the js that is overridden if explicitly set.

src/webgl/shaders/filters/blur.frag
```diff
@@ -0,0 +1,50 @@
+// Single-pass blur filter, taken from Adam Ferriss' repo of shader examples:
```
aferriss, line 1: How does the output of this blur compare to the cpu one?

src/webgl/shaders/filters/dilate.frag
```diff
@@ -0,0 +1,38 @@
...
+precision highp float;
+
+varying vec2 vTexCoord;
+
+uniform sampler2D tex0;
+uniform vec2 texelSize;
+
+float luma(vec3 color) {
+  // based on constants 77, 151, 28 from DILATE in filters.js,
+  // even though that's different than the luminance constants used in GRAY
```
aferriss, line 12: this is fine, haha. Probably just different implementers

src/webgl/shaders/filters/dilate.frag
```diff
@@ -0,0 +1,38 @@
+// Increase the bright areas in an image
```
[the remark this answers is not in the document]
aferriss: Same for erode, but it will be a min instead of max

src/webgl/shaders/filters/posterize.frag
aferriss: Interesting, I'll take a look and try to suss out whats going on here.

aferriss: I think your round was doing something weird. It should be a floor instead. See: https://editor.p5js.org/aferriss/sketches/mjQKudkIP

src/webgl/p5.RendererGL.js
```diff
@@ -970,31 +985,42 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     }
     let pg = this.filterGraphicsLayer;
 
+    // use internal shader for filter constants BLUR, INVERT, etc
+    let filterParameter = undefined;
```
aferriss, line 989: lets predefine some defaults for this value. 
```
if(operation === 'BLUR'){
  filterParameter = 4;
} else if(operation == 'POSTERIZE') ...
```
Then later you can check if there is a value in args[1] and set it equal to filterParameter if it exists.

src/webgl/shaders/filters/blur.frag
wong-justin: Here's a sketch for comparison, although bricks.jpg isn't the best example to compare blur:
https://editor.p5js.org/jwong/sketches/r9fxPerPU

aceslowman:
> I think I documented all the important features and behaviors between the `filter()` and `createFilterShader()` pages. I tried to keep `filter()` a little less technical or webgl-centric, at least until the bottom, since that's more of an entry point page for beginners. The rest of the gotchas and explanations and further reading are in `createFilterShader()` for advanced users. Screenshots, with examples omitted for brevity:
> 
> `filter()` image
> 
> `createFilterShader()` image
> 
> Open to suggestions as always, especially @aceslowman - I remember you said it would be nice to have an earlier disclaimer about filter shaders not working on 3D geometries, for people coming from shader land? Maybe swapping the second paragraph of `createFilterShader()` with the first?

I think you have a good organization of things for these reference pages. I think that it's important to keep filter() as easy to understand as possible, so putting more of that in createFilterShader() feels appropriate. In createFilterShader(), you mention 'entire contents of a canvas', and I would get rid of the word 'entire' to not confuse anyone around it's effect on 3D geometry. Personally, I think that the order of the first two paragraphs for createFilterShader() are okay, I would move the comment about it not applying to 3D geometry to the end of the second paragraph.

src/webgl/shaders/filters/dilate.frag
aceslowman: @jwong I think that having it match what is expected for the non-webgl filters feels best, glad to see you leave a comment about it here, I think it's helpful context.

src/webgl/shaders/filters/dilate.frag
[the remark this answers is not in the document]
wong-justin: `max()` seems to choose that greatest of each component, compared to the `luma()` way of choosing the greatest overall vector. Example:
```
a = (1, 1, 0)
b = (0, 0, 1)
max(a,b) = (1, 1, 1)  # instead of (1, 1, 0)
```

Not sure if that's more desired or not. The results look similar though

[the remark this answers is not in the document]
aferriss: I think either way is fine, if the luma calculation is more accurate to the CPU mode then disregard my comment!

src/webgl/p5.RendererGL.js
```diff
@@ -1033,16 +1034,46 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+
+      for (let i = 0; i < steps; i++) {
+        // first pass averaging horizontal neighbors
+        pg.shader(this.filterShader);
+        this.filterShader.setUniform('texelSize', [1/this.width, 1/this.height]);
+        this.filterShader.setUniform('tex0', pg);
+        this.filterShader.setUniform('direction', [2, 0]); // 2 is a decent
+        pg.rect(0,0,this.width,this.height);               //  default spread
+        // another pass, this time vertically
+        pg.shader(this.filterShader);
```
aferriss, line 1059: Do you actually need to call this again, isn't the shader already bound?

src/webgl/p5.RendererGL.js
```diff
@@ -1033,16 +1034,46 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+        -this.width/2, -this.height/2, this.width, this.height
+      );
+      // how much to blur, given by user
+      let steps = filterParameter;
+
+      for (let i = 0; i < steps; i++) {
+        // first pass averaging horizontal neighbors
+        pg.shader(this.filterShader);
+        this.filterShader.setUniform('texelSize', [1/this.width, 1/this.height]);
+        this.filterShader.setUniform('tex0', pg);
```
aferriss, line 1055: Does this work? Aren't you drawing to the same buffer that you're writing to?

src/webgl/p5.RendererGL.js
```diff
@@ -1033,16 +1034,46 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+        this.filterShader.setUniform('direction', [2, 0]); // 2 is a decent
+        pg.rect(0,0,this.width,this.height);               //  default spread
+        // another pass, this time vertically
+        pg.shader(this.filterShader);
+        // this.filterShader.setUniform('texelSize', [1/this.width, 1/this.height]);
+        this.filterShader.setUniform('tex0', pg);
```
aferriss, line 1061: On the second pass, would it make sense to just draw directly to the main renderer?
```
// somewhere earlier we need to make the 2nd pass shader in the main renderer's context
this.filterShaderMain = new Shader(blur.vert, blur.frag);
// later when we're about to draw to screen
    this._pInst.push();
    this._pInst.shader(this.filterShaderMain);
    this.filterShaderMain.setUniform('texelSize', [1/this.width, 1/this.height]);
    this.filterShaderMain.setUniform('tex0', pg);
    this.filterShaderMain.setUniform('direction', [0, 2]);
    this._pInst.rect(0, 0, this.width, this.height);
    this._pInst.pop();
    ```

File: src/webgl/shaders/filters/default.vert [hunk 11]
```diff
@@ -0,0 +1,17 @@
...
+varying vec2 vTexCoord;
+
+void main() {
+  // transferring texcoords for the frag shader
+  vTexCoord = aTexCoord;
+
+  // copy position with a fourth coordinate for projection (1.0 is normal)
+  vec4 positionVec4 = vec4(aPosition, 1.0);
+  // scale by two and center to achieve correct positioning
+  positionVec4.xy = positionVec4.xy * 2.0 - 1.0;
```
aferriss, line 14: I think we should change this shader to use the model view projection matrix instead. See my issue here #6367 where this shader is causing trouble because of the depth testing.

src/webgl/p5.RendererGL.js
wong-justin: That could work. I felt like it made more sense to do everything on the secondary renderer, and just leave the main renderer for the final step. Trying to keep things from being tangled together.

src/webgl/p5.RendererGL.js
wong-justin: It seems like it works now just using one renderer, but I was doing it wrong before. `setUniform('tex0', this)` seems to have helped.

wong-justin:
I seem to have fixed blur. Still testing and double checking things. It's promising that the parameter works as intended, and I fixed a flipping problem midway, and it works every frame, and there's no performance drops. It's much smoother than the single pass was. Some screenshots:

`filter(BLUR, 1)`
image

`filter(BLUR) # default 3`
image

and for reference, `filter(BLUR, 3, useWebGL=false)`
image

src/webgl/p5.RendererGL.js
```diff
@@ -970,32 +991,92 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+
+      // initial binding and setup
+      pg.shader(this.filterShader);
+      this.filterShader.setUniform('texelSize', [1/this.width, 1/this.height]);
+
+      // do initial horizontal and vertical pass,
+      // starting with parent renderer as tex0 uniform
+      this.filterShader.setUniform('tex0', this);     // vertically flips first
+      this.filterShader.setUniform('flipped', false); // so undo it
+      this.filterShader.setUniform('direction', [2, 0]);
```
aferriss, line 1050: I feel like the 2 here might be causing some slight pixellation, maybe we should change back to 1 for both horizontal and vertical directions

src/webgl/p5.RendererGL.js
```diff
@@ -970,32 +991,92 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    // apply blur shader with multiple passes
+    if (operation === constants.BLUR) {
+
+      // initial binding and setup
+      pg.shader(this.filterShader);
+      this.filterShader.setUniform('texelSize', [1/this.width, 1/this.height]);
+
+      // do initial horizontal and vertical pass,
+      // starting with parent renderer as tex0 uniform
+      this.filterShader.setUniform('tex0', this);     // vertically flips first
```
aferriss, line 1048: why do we need to do these initial passes? I think we could get away with just drawing the source canvas into the pg layer without a shader, and then running the loop.

src/webgl/p5.RendererGL.js
```diff
@@ -1015,8 +1015,18 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
           filterShaderVert,
           filterShaderFrags[operation]
         );
+
+        // two-pass blur filter needs another shader attached to main
+        if (operation === constants.BLUR) {
```
aferriss, line 1020: What if someone is trying to use the accelerated filters from p2d mode. Won't this cause an error?

Qianqianye:
Looks great! Thanks @wong-justin for working on it and @aferriss @aceslowman for reviewing. 🎉

davepagurek:
I think in the updated docs for `BLUR`, it mentions that it uses a box blur for the shader version as opposed to the Gaussian blur in CPU mode. We could plausibly still use a Gaussian kernel in the shader to get things to look more similar and maybe compare performance between the two to see if the tradeoff is worth it to match output? We could open an issue for that to at least investigate.

There was also some discussion in comments about whether or not to use luma-based dilate/erode, which I believe the CPU implementation uses, and the shaders now do not use (for performance?), which probably explains some of the difference for those two filters.

Also I wonder if pixel density handling affects it? I think the temp graphics layer doesn't specify pixel density so it'd use display density instead of matching the main canvas density, and `texelSize` may not factor pixel density in (although I don't know how the CPU version handles it, I'd need to look into that.) (Also maybe we should handle canvas resizes done after the temp layer has been created?)

Anyway we could maybe open follow-up issues to track and then iron out the differences?

PR #6413, opened 2023-09-15
Fixed mousePressed() Example Error

Utkarsh3128:
## Descriptiion 
The example code at: https://p5js.org/reference/#/p5/mousePressed does not function, no response when key is pressed. This is due to an error in the variable names. So I have changed the variable name ```value``` to ```colorValue``` in the function ```mousePressed()```

Resolves # [1415](https://github.com/processing/p5.js-website/issues/1415) : mousePressed() Example Error

 Changes:

Screenshot (222)

Screenshot of Changes:
Screenshot (223)

Additional Description:
I don't know why css is not working on local machine. That's why page is looking different

### DIFF
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -621,7 +621,7 @@ p5.prototype._onmousemove = function(e) {
  *   describe('black 50-by-50 rect turns white with mouse click/press.');
  * }
  * function mousePressed() {
- *   if (value === 0) {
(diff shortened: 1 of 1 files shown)


nickmcintyre:
Thanks for the fix @Utkarsh3128! Your fix and the CSS work fine on my machine and the tests pass, so we can go ahead and merge. What happens when you try a rebuild and a hard refresh?

Issue #6449, opened 2023-09-30
Edit typography docs to align with style guide


nickmcintyre:
### Increasing Access

Writing with plain, consistent language increases cognitive accessibility. It's also more welcoming to younger members of the community.

### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [X] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors
- [ ] Other (specify if possible)

### Feature enhancement details

I'll edit the inline documentation and code samples for each function in the [typography group](https://p5js.org/reference/#group-Typography).

EarthIsHeaven:
Hey @nickmcintyre can i work on this issue ?

nickmcintyre:
Sure @EarthIsHeaven! The more eyes we have on these edits the better. How about reviewing my most recent pull request to get started?

Here's how I build and test documentation locally. This workflow assumes you have [nvm](https://github.com/nvm-sh/nvm) installed on macOS or Linux.

```bash
# Install the LTS version of node and use it.
nvm install --lts
nvm use --lts

# Make a folder for docs and navigate to it.
mkdir docs && cd docs

# Clone Nick's fork of the p5.js repo.
git clone https://github.com/nickmcintyre/p5.js

# Clone the p5.js-website repo.
git clone https://github.com/processing/p5.js-website

# Navigate to the p5.js directory.
cd p5.js
# Check out the branch under review.
git checkout sod-typography-load-display
# Generate data for the API reference.
npm install
npx grunt lint:samples

# Navigate to the p5.js-website directory.
cd ../p5.js-website
# Copy data for the API reference.
cp ../p5.js/docs/reference/data* src/templates/pages/reference
# Build a draft of the website.
npm install
npm run build

# Navigate to the directory with the built website.
cd dist
# Serve the website on localhost:8000.
python3 -m http.server
```

skbhagat0502:
Hi @nickmcintyre I am working on this issue.

nickmcintyre:
@skbhagat0502 thanks for offering to help out! I'll make another pull request with edits soon and will tag you to review.

nickmcintyre:
[...]

PR #6450, opened 2023-09-30
Edit docs for typography load and display

nickmcintyre:
Addresses #6449 
I edited the inline documentation for the typography functions `loadFont()`, `text()`, and `textFont()` to bring them more in line with the style guide.

@davepagurek @dkessner @MsQCompSci @limzykenneth @Qianqianye @dhowe, @paulaxisabel, @SarveshLimaye, @SkylerW99, @hannahvy, @BamaCharanChhandogi, @singshris I'd love to incorporate any feedback you may have!

- [x] `npm run lint` passes

### DIFF
--- a/src/typography/loading_displaying.js
+++ b/src/typography/loading_displaying.js
@@ -226,6 +226,17 @@ p5.prototype.loadFont = function(path, onSuccess, onError) {
  *
  * <div>
  * <code>
+ * background('skyblue');
+ * textSize(100);
+ * // Emoji.
+ * text('🌈', 0, 100);
+ *
+ * describe('A rainbow in a blue sky.');
+ * </code>
+ * </div>
+ *
+ * <div>
+ * <code>
  * textSize(32);
  * fill(255);
  * stroke(0);
@@ -325,7 +336,7 @@ p5.prototype.text = function(str, x, y, maxWidth, maxHeight) {
  * <code>
  * textFont('Courier New');
  * textSize(24);
- * text('hi', 40, 55);
+ * text('hi', 35, 55);
  *
  * describe('The text "hi" written in a black, monospace font on a gray background.');
  * </code>
@@ -336,9 +347,10 @@ p5.prototype.text = function(str, x, y, maxWidth, maxHeight) {
  * background('black');
  * fill('palegreen');
  * textFont('Courier New', 10);
- * text('Wake up, Neo...', 6, 52);
+ * text('You turn to the left and see a door. Do you enter?', 5, 5, 90, 90);
+ * text('>', 5, 70);
  *
- * describe('The text "Wake up, Neo..." written in a green, monospace font on a black background.');
+ * describe('A text prompt from a game is written in a green, monospace font on a black background.');
  * </code>
  * </div>
  *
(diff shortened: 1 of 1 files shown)


nickmcintyre:
I noticed a quirk in the failure callback for `loadFont()`. When I tried logging the error `Event` object, I just got some text. This may be a bug.

dhowe:
Docs look good to me. When I log the object from the error callback, I do see that it is an Error object. However the text  component of that error (`"Error: Unsupported OpenType signature"`) is not visible in the p5js console, while it is in the javascript console. Not sure if this is intended behavior

davepagurek (approved):
The docs look good to me too!

dkessner:
In the text() examples, for historical (and unnecessary?!) accuracy, perhaps change to "The quick brown fox jumps over the lazy dog."    The sentence uses every letter in the alphabet, but misses 's' with 'jumped' instead of 'jumps' :)

nickmcintyre:
@dkessner nice catch with "jumps." Fixed!

@dkessner @MsQCompSci here's a [staging version](https://computiful.dev/sod-typography-load-display/reference/#group-Typography) of the docs for this pull request. Don't wander too far off the Reference pages or stuff will break–patched just enough of the site to get this working.

Please let me know if you'd like to see further revisions or if we should request a merge.

MsQCompSci:
Some thoughts considering newer or younger users:
- for `loadfont()` Maybe including the link to the font can make reproducing the code easier.  
- I wonder if there is a way to indicate that the code in the `text()` examples is located within setup. 
- In `text()` is there a rationale behind including the comment in the example (because there are no comments on any of the other examples)
- Note about comments: Newer users will struggle to understand the example code without the aid of comments to help direct their attention. For example, a comment that describes what a group of commands does may help newer users understand all examples better. 
- Is there a way to let users know the types of built in fonts in the `textFont()` section?

nickmcintyre:
@dkessner good point about linking to functions in the same category. What do you think about the "Related" section on Processing's [version of text](https://processing.org/reference/text_.html)?

@MsQCompSci great thoughts. Here's 
> for `loadfont()` Maybe including the link to the font can make reproducing the code easier.

Good call–done!

> I wonder if there is a way to indicate that the code in the `text()` examples is located within `setup()`.

My high school students regularly encounter bugs by calling p5 functions outside of `setup()` or `draw()`. On the one hand, the no-`setup()` style removes boilerplate as @davepagurek noted. On the other, it feels like this style reinforces a pretty common misconception. I revised examples in this section to use `setup()`.

> In `text()` is there a rationale behind including the comment in the example (because there are no comments on any of the other examples)

Not really. I think it's pretty clear that the rainbow is an emoji.

> Note about comments: Newer users will struggle to understand the example code without the aid of comments to help direct their attention. For example, a comment that describes what a group of commands does may help newer users understand all examples better.

Agreed. I've done so sparingly in some other references. A design goal is to make the examples as short and clear as possible, to the point that comments aren't necessary.

> Is there a way to let users know the types of built in fonts in the `textFont()` section?

Yep, added! The Parameters section at the bottom also lists types.

@Qianqianye @davepagurek @limzykenneth if these changes look good to you, please go ahead and merge them.

PR #6453, opened 2023-10-03
Edit docs for p5.Font

nickmcintyre:
Addresses #6449

Changes:
I edited the inline documentation for the `p5.Font` class to bring it more in line with the style guide.

@davepagurek @dkessner @MsQCompSci @limzykenneth @Qianqianye @dhowe, @paulaxisabel, @SarveshLimaye, @SkylerW99, @hannahvy, @BamaCharanChhandogi, @singshris, @EarthIsHeavin, @skbhagat0502 I'd love to incorporate any feedback you may have!

- [x] `npm run lint` passes
- [x] [Inline documentation] is included / updated

PR #6454, opened 2023-10-03
Edit docs for typography attributes

nickmcintyre:
Addresses #6449

Changes:
I edited the inline documentation for functions that set typography attributes to bring them more in line with the style guide.

@davepagurek @dkessner @MsQCompSci @limzykenneth @Qianqianye @dhowe, @paulaxisabel, @SarveshLimaye, @SkylerW99, @hannahvy, @BamaCharanChhandogi, @singshris, @EarthIsHeavin, @skbhagat0502 I'd love to incorporate any feedback you may have!

`textAscent()` and `textDescent()` still confuse me, so I'd appreciate your guidance there.

### DIFF
--- a/src/typography/attributes.js
+++ b/src/typography/attributes.js
@@ -227,7 +227,7 @@ p5.prototype.textStyle = function(theStyle) {
  *   text(s, 22, 55);
  *   line(22, 55, 22 + w, 55);
  *
- *   describe('The word "yo" written twice on separate lines. The words are divided by a horizontal line.');
+ *   describe('The word "yo" written twice, one copy beneath the other. The words are divided by a horizontal line.');
  * }
  * </code>
  * </div>
--- a/src/typography/attributes.js
+++ b/src/typography/attributes.js
@@ -199,26 +199,36 @@ p5.prototype.textStyle = function(theStyle) {
  * @example
  * <div>
  * <code>
- * textSize(28);
- * let str = 'hi';
- * let w = textWidth(str);
- * text(str, 35, 55);
- * line(35, 55, 35 + w, 55);
+ * function setup() {
+ *   background(200);
  *
- * describe('The text "hi" underlined.');
+ *   textSize(28);
+ *   strokeWeight(0.5);
+ *   let s = 'yoyo';
+ *   let w = textWidth(s);
+ *   text(s, 22, 55);
+ *   line(22, 55, 22 + w, 55);
+ *
+ *   describe('The word "yoyo" underlined.');
+ * }
  * </code>
  * </div>
  *
  * <div>
  * <code>
- * textSize(28);
- * // "\n" starts a new line.
- * let str = 'yo\nyo';
- * let w = textWidth(str);
- * text(str, 10, 40);
- * line(10, 40, 10 + w, 40);
+ * function setup() {
+ *   background(200);
  *
- * describe('The word "yo" written twice on separate lines. The words are divided by a horizontal line.');
+ *   textSize(28);
+ *   strokeWeight(0.5);
+ *   // "\n" starts a new line.
+ *   let s = 'yo\nyo';
+ *   let w = textWidth(s);
+ *   text(s, 22, 55);
+ *   line(22, 55, 22 + w, 55);
+ *
+ *   describe('The word "yo" written twice on separate lines. The words are divided by a horizontal line.');
+ * }
  * </code>
  * </div>
  */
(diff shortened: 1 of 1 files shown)


src/typography/attributes.js
```diff
@@ -81,30 +83,27 @@ p5.prototype.textAlign = function(horizAlign, vertAlign) {
 };
 
 /**
- * Sets/gets the spacing, in pixels, between lines of text. This setting will be
- * used in all subsequent calls to the <a href="#/p5/text">text()</a> function.
+ * Sets the spacing between lines of text when
+ * <a href="#/p5/text">text()</a> is called. Spacing is measured in pixels.
+ *
+ * Calling `textLeading()` without an arugment returns the current spacing.
```
dhowe, line 89: argument is mispelled

dhowe (commented):
I noticed one typo (see comment), but otherwise looks good, thanks!

src/typography/attributes.js
```diff
@@ -335,47 +330,14 @@ p5.prototype._updateTextMetrics = function() {
...
- * ellipse(11, 47, 1, 1);
- * ellipse(19, 47, 1, 1);
- * ellipse(15, 52, 5, 5);
- * line(15, 60, 15, 70);
- * line(15, 65, 5, 55);
- * line(15, 65, 25, 55);
- * line(15, 70, 10, 80);
- * line(15, 70, 20, 80);
- * </code>
- * </div>
```
limzykenneth, line 378: The additional examples here are to demonstrate a couple more edge cases for `textWrap()` (ok maybe the last one a bit less but I selfishly like it). Do you think it would make sense to keep at least the Chinese characters one or it would just be confusing?

limzykenneth:
In terms of `textAscent` and `textDescent` I'm unsure which aspects confused you so do clarify if you can but looking at the example for those two you created, I think you mostly got it. However I will probably suggest changing the examples to use instead of vertical line, to use horizontal lines instead, explained along with the following diagram

A diagram showing terms to do with letter height and positioning on the baseline. Alternative terms are italicised.

Where for `textAscent` we draw the baseline and the top ascent line only while for `textDescent` we draw baseline and the bottom descent line.

dkessner:
Overall readability looks good.  I like the suggestion from @limzykenneth to use horizontal lines in the diagrams to explain textAscent and textDescent.

src/typography/attributes.js
nickmcintyre: Good catch–fixed!

src/typography/attributes.js
nickmcintyre: The last example is a little busy. Good call to restore the Chinese characters. Fixed!

nickmcintyre:
@limzykenneth @dkessner I restored edited versions of the original examples for `textAscent()` and `textDescent()`. The originals were broken using the default system font on macOS, so I loaded Inconsolata for consistency.

Needing to multiply by `scalar` messes with my intuition. It feels like `textAscent()` and `textDescent()` should do this internally. We can table that for another time.

@dkessner @MsQCompSci here's a [staging version](https://computiful.dev/sod-typography-attributes/reference/#group-Typography) of the docs for this pull request. Don't wander too far off the Reference pages or stuff will break–patched just enough of the site to get this working.

Please let me know if you'd like to see further revisions or if we should request a merge.

dhowe:
Is there a problem with the image for textBounds()
Also, wondering if we should explain what 'tight bounds' means ?

dkessner:
I'm not sure what "scalar" is supposed to be doing in this example.  It's definitely confusing, and should be unnecessary based on my understanding of what textAscent() returns (the height of the 'd').

I looked at the Processing example, which has the same scalar.  But weirdly, the example doesn't work when I run it in Processing on Mac (the line is in the wrong place).  It does, however, work when I set the scalar to 1.0.  
https://processing.org/reference/textAscent_.html

MsQCompSci:
This looks good!

MsQCompSci:
- I wonder if there is a simpler example for `textWidth`, `textAscent`, `textDescent`. Younger users might find it difficult to understand what it is being used for in the examples included here.

nickmcintyre:
@dhowe revisions for the `font.textBounds()` documentation are part of #6453. You can preview them [here](https://computiful.dev/sod-p5-font/reference/#/p5.Font/textBounds).

@dkessner @MsQCompSci thanks for looking into this. Agreed that this should be simpler and will give the examples another go this evening.

nickmcintyre:
@dkessner @MsQCompSci the [staging version](https://computiful.dev/sod-typography-attributes/reference/#group-Typography) of the reference for this pull request has new examples. Please let me know if you have suggestions or would like to see further simplifications.

I've never taught these functions to beginners. They are precision tools for more advanced use cases. That said, `textAscent()` and `textDescent()` do seem overly complex. I'm probably missing some context for their design.

dhowe:
Indeed, the fontScale argument does seem less than intuitive. At very least we were trying to match the API for processing at the time, but I'm not sure if there were other factors at play. Taking a quick glance at the code below, it would _seem_ that this argument could be omitted, and thus could be marked as optional. Has anyone tested this?

image

nickmcintyre:
@dhowe thanks for looking into this. I just tested the following changes with the `textAscent()` example:
```javascript
// ...
textSize(24);
// ...
let a = textAscent(24);
// ...
textSize(48);
// ..
a = textAscent(48);
// ...
```
Passing the current font size to `textAscent()`, as in `textAscent(24)`, didn't affect the ascender's height.  It's still necessary to multiply `textAscent(24) * fontScale`. Calling `font._textAscent(24)` works the same way. It might be worth starting a separate discussion about `textAscent()` and `textDescent()`.

@dhowe @Qianqianye @limzykenneth @davepagurek if these changes look good to you, please go ahead an merge this pull request.

dhowe:
> Passing the current font size to `textAscent()`, as in `textAscent(24)`, didn't affect the ascender's height. It's still necessary to multiply `textAscent(24) * fontScale`. Calling `font._textAscent(24)` works the same way. It might be worth starting a separate discussion about `textAscent()` and `textDescent()`.

It _might_ be worth looking into whether we can include an appropriate `fontScale` as a default, though this would both break  backward compatibility and take us out of sync with processing

nickmcintyre:
@davepagurek @limzykenneth @Qianqianye if these changes look good to you, please go ahead and merge them.

Qianqianye:
Thank you all for reviewing. I will merge this PR for now and we can continue the conversation about `fontScale`.

PR #6464, opened 2023-10-09
Fix: Misleading error message when NaN passed

### LINKED ISSUE #6463: FES gives misleading error message when NaN passed to function

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [X] Friendly Errors

### p5.js version

1.7.0

### Web browser and version

Firefox 118.0.1

### Operating System

OS X

### Steps:
1. Run code below, see error message:
🌸 p5.js says: [sketch.js, line 9] line() was expecting Number for the first parameter, received number instead. (http://p5js.org/reference/#/p5/line)

```js
function setup() {
  createCanvas(400, 400);
  let i = 0, j = 0;
  point(i/j,10);
}
```

capGoblin:
Resolves #6463

 Changes:

```js
function setup() {
  createCanvas(400, 400);
  let i = 0, j = 0;
  point(i/j,10);
}
```
whenever a NaN is passed it gives error message:
> 🌸 p5.js says: [sketch.js, line 4] point() was expecting Number for the first parameter, received number instead. (http://p5js.org/reference/#/p5/point) 

where it is supposed to give:
> 🌸 p5.js says: [sketch.js, line 4] point() was expecting Number for the first parameter, received NaN instead. (http://p5js.org/reference/#/p5/point) 

I have updated the code in validate_params.js for `_friendlyParamError` function under the switch case "WRONG_TYPE" to include an additional check using `isNaN(arg)` to detect NaN values.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/friendly_errors/validate_params.js
+++ b/src/core/friendly_errors/validate_params.js
@@ -569,7 +569,7 @@ if (typeof IS_MINIFIED !== 'undefined') {
         const argType =
           arg instanceof Array
             ? 'array'
-            : arg === null ? 'null' : arg.name || typeof arg;
+            : arg === null ? 'null' : isNaN(arg) ? 'NaN' : arg.name || typeof arg;
 
         translationObj = {
           func,
--- a/src/core/friendly_errors/validate_params.js
+++ b/src/core/friendly_errors/validate_params.js
@@ -569,7 +569,7 @@ if (typeof IS_MINIFIED !== 'undefined') {
         const argType =
           arg instanceof Array
             ? 'array'
-            : arg === null ? 'null' : typeof arg === 'number' && isNaN(arg) ? 'NaN' : arg.name || typeof arg;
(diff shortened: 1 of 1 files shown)


limzykenneth:
The current implementation here does not quite work as it will give an incorrect value when the type given is a string. Ie. with `arg === null ? 'null' : isNaN(arg) ? 'NaN' : arg.name || typeof arg;` if the type of `arg` is a string, it will return `'NaN'` instead of `'string'` it previously did.

capGoblin:
hmm ya i totally forgot about strings existence😅.  Now, I've updated the code to check `isNaN(arg)` only if arg is of type 'number' like `typeof arg === 'number' && isNaN(arg) ? 'NaN'`.
Additionally, I've added an extra check after null for `undefined` just in case if an uninitialized variable is passed.

limzykenneth:
Looks good. Thanks!

PR #6512, opened 2023-10-29
Edit docs for p5.Element

nickmcintyre:
Addresses #6511 

Changes:
I edited the inline docs for the `p5.Element` methods from`parent()` through `dragLeave()` to bring them more in line with the style guide. These docs live in `src/core/p5.Element.js`. I'll edit the docs from `src/dom/dom.js` in another pull request.

@dkessner, @MsQCompSci, @davepagurek, @limzykenneth, @Qianqianye, @SarveshLimaye, @SoundaryaKoutharapu, @ramya202000, @BamaCharanChhandogi, @Obi-Engine10, @MarceloGoncalves, @hiddenenigma I'd love to incorporate any feedback you may have! Here's a [staging version](https://computiful.dev/sod-p5-element/reference/#/p5.Element) of the edits in this pull request.

#### PR Checklist

- [x] `npm run lint` passes
- [x] [Inline documentation] is included / updated

PR #6534, opened 2023-11-07
Revise docs for dom functions

nickmcintyre:
Addresses #6511 

Changes:

I edited the inline docs for the top-level functions in `src/dom/dom.js`. This includes everything in the [DOM group](https://computiful.dev/sod-dom-functions/reference/#group-DOM) except the following:

- `p5.Element` see #6512
- `changed()`
- `input()`
- `p5.MediaElement`
- `p5.File`

I'll make a separate pull request for the last four items.

@dkessner, @MsQCompSci, @davepagurek, @limzykenneth, @Qianqianye, @SarveshLimaye, @SoundaryaKoutharapu, @ramya202000, @BamaCharanChhandogi, @Obi-Engine10, @MarceloGoncalves, @hiddenenigma I'd love to incorporate any feedback you may have! Here's a [staging version](https://computiful.dev/sod-dom-functions/reference/#group-DOM) of the edits in this pull request.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -19,40 +19,75 @@
 import p5 from '../core/main';
 
 /**
- * Searches the page for the first element that matches the given CSS selector string (can be an
- * ID, class, tag name or a combination) and returns it as a <a href="#/p5.Element">p5.Element</a>.
- * The DOM node itself can be accessed with .elt.
- * Returns null if none found. You can also specify a container to search within.
+ * Searches the page for the first element that matches the given
+ * <a href="https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/CSS_basics#different_types_of_selectors" target="_blank">CSS selector string</a>.
+ * The string can be an ID, class, tag name, or a combination. `select()`
+ * returns a <a href="#/p5.Element">p5.Element</a> object if it finds a match
+ * and `null` otherwise.
+ *
+ * The second parameter, `container`, is optional. It specifies a container to
+ * search within. container can be CSS selector string, a
+ * <a href="#/p5.Element">p5.Element</a> object, or an
+ * <a href="https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement" target="_blank">HTMLElement</a> object.
  *
  * @method select
- * @param  {String} selectors CSS selector string of element to search for
+ * @param  {String} selectors CSS selector string of element to search for.
  * @param  {String|p5.Element|HTMLElement} [container] CSS selector string, <a href="#/p5.Element">p5.Element</a>, or
- *                                             HTML element to search within
- * @return {p5.Element|null} <a href="#/p5.Element">p5.Element</a> containing node found
+ *                                             <a href="https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement" target="_blank">HTMLElement</a> to search within.
+ * @return {p5.Element|null} <a href="#/p5.Element">p5.Element</a> containing the element.
  * @example
- * <div><code>
+ * <div>
+ * <code>
  * function setup() {
- *   createCanvas(50, 50);
- *   background(30);
- *   // move canvas down and right
- *   select('canvas').position(10, 30);
- * }
- * </code></div>
+ *   createCanvas(100, 100);
+ *   background(200);
  *
- * <div class="norender"><code>
- * // select using ID
- * let a = select('#container');
- * let b = select('#beep', '#container');
- * let c;
- * if (a) {
- *   // select using class
- *   c = select('.boop', a);
+ *   // Select the canvas by its tag.
+ *   let cnv = select('canvas');
+ *   cnv.style('border', '5px deeppink dashed');
+ *
+ *   describe('A gray square with a dashed pink border.');
  * }
- * // select using CSS selector string
- * let d = select('#container #bleep');
- * let e = select('#container p');
- * [a, b, c, d, e]; // unused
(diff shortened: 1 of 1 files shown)


MsQCompSci:
Edits look great!

src/dom/dom.js
```diff
@@ -339,18 +486,28 @@ p5.prototype.createDiv = function (html = '') {
 };
 
 /**
- * Creates a `&lt;p&gt;&lt;/p&gt;` element in the DOM with given inner HTML. Used
- * for paragraph length text.
+ * Creates a `&lt;p&gt;&lt;/p&gt;` element. It's commonly use for
```
davepagurek, line 489: Small typo:
```suggestion
 * Creates a `&lt;p&gt;&lt;/p&gt;` element. It's commonly used for
```

src/dom/dom.js
```diff
@@ -359,16 +516,42 @@ p5.prototype.createP = function (html = '') {
 };
 
 /**
- * Creates a `&lt;span&gt;&lt;/span&gt;` element in the DOM with given inner HTML.
+ * Creates a `&lt;span&gt;&lt;/span&gt;` element. It's commonly used as a
+ * container for other elements.
```
davepagurek, line 520: We also say this for `<div>`, maybe we can add something small about it being generally used as a container for inline things? Not sure what the best wording for that is. Things that might wrap along with text?

src/dom/dom.js
```diff
@@ -359,16 +516,42 @@ p5.prototype.createP = function (html = '') {
...
+ * </code>
+ * </div>
+ *
+ * <div>
+ * <code>
+ * function setup() {
+ *   background(200);
+ *
+ *   // Create an h3 element within the span.
+ *   let span = createSpan('<h3>p5*js</h3>');
```
davepagurek, line 548: This works but I also wonder if there's a way to illustrate the difference with a div here. Maybe if you have a div that already has some text, you could add a span with red text into the middle of its contents?

src/dom/dom.js
```diff
@@ -377,29 +560,54 @@ p5.prototype.createSpan = function (html = '') {
 };
 
 /**
- * Creates an `&lt;img&gt;` element in the DOM with given src and
- * alternate text.
+ * Creates an `&lt;img&gt;` element.
```
davepagurek, line 563: Maybe worth differentiating this from `p5.Image`: this exists *outside* of the canvas while `p5.Image` is only visible when you draw it on the canvas

src/dom/dom.js
```diff
@@ -505,30 +819,67 @@ p5.prototype.createSlider = function (min, max, value, step) {
...
- * Creates a `&lt;button&gt;&lt;/button&gt;` element in the DOM.
- * Use .size() to set the display size of the button.
- * Use .mousePressed() to specify behavior on press.
+ * Creates a `&lt;button&gt;&lt;/button&gt;` element.
+ *
+ * The first parameter, `label`, is a string that sets the label displayed on
+ * the button.
+ *
+ * The second parameter, `value`, is optional. It's a string that sets the
+ * button's value. See <a href="#/p5.Element/value">p5.Element.value()</a> for
```
davepagurek, line 828: Think it's worth linking to MDN here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#value for an explanation of why a button might need a value and how it gets used? It's definitely an edge case, I didn't even realize this was a thing until a few minutes ago, but I just had to go google it to see what it might mean 😅

src/dom/dom.js
```diff
@@ -1270,45 +1926,86 @@ function createMedia(pInst, type, src, callback) {
...
  * }
  *
- * // This function is called when the video loads
- * function vidLoad() {
- *   vid.loop();
- *   vid.volume(0);
+ * function handleVideo() {
+ *   video.size(100, 100);
+ *   video.loop();
+ *   video.volume(0);
```
davepagurek, line 2005: If I go directly to https://computiful.dev/sod-dom-functions/reference/#/p5/createVideo to simulate coming from a google search (or search for `createVideo`, go to the docs, then reload) the video doesn't immediately play, I suspect this is because we're muting after looping:
image

Maybe we should try muting first? If that does indeed fix it, it's probably worth mentioning explicitly in the docs that that ordering matters because of how the browser deals with autoplaying media.

src/dom/dom.js
```diff
@@ -1318,36 +2015,49 @@ p5.prototype.createVideo = function (src, callback) {
...
+ *
  * function setup() {
- *   ele = createAudio('assets/beat.mp3');
+ *   background(200);
+ *   text('Double-click to stop', 25, 55);
+ *
+ *   // Load the audio.
+ *   beat = createAudio('assets/beat.mp3', () => {
+ *     // Play the audio in a loop once it's loaded.
+ *     beat.loop();
```
davepagurek, line 2050: This one also doesn't work for me if I go directly to the URL for this doc. Maybe we can start it on click, or only load the audio on click? Probably worth mentioning for both audio and video that this might happen if there hasn't been a user interaction

davepagurek (commented):
This is looking good! Biggest questions for me are how we want to handle the media element ones and autoplaying and the browser's limitations of autoplaying with sound.

nickmcintyre:
Thanks @davepagurek! I fixed the typos, reworked the `<span>` example, and used `showControls()` to get around autoplay issues.

davepagurek (approved):
Nice, looks good!

PR #6549, opened 2023-11-11
Solves issue #6519

Garima3110:
Solves issue #6519
 Changes:
I have tried to incorporate all the changes suggested by @davepagurek . Please let me know if something else needs to be changed. Thank You..!
Link to the changes made:  [src/webgl/p5.Camera.js](https://github.com/processing/p5.js/compare/main...Garima3110:p5.js:perspective#diff-efc2926e978de2b4f241ea3f5251cd7572dacd3a58d87b357c9ecbbdcf342ced)

 Screenshots of the change:

Change in translate():
image

Documentation update:
image

#### PR Checklist

- [x] `npm run lint` passes


src/webgl/p5.Camera.js
```diff
@@ -127,9 +127,29 @@ p5.prototype.camera = function (...args) {
...
+ *
+ * aspect- The default aspect ratio is the ratio of renderer's width to renderer's height.
+ *
+ * near - The default value for the near clipping plane is 0.1 times the default distance from the camera to the point it is looking at i.e 800.
+ *
+ * far - The default value for the far clipping plane is 10 times the default distance from the camera to the point it is looking at i.e 800.
+ *
+ * If you prefer a fixed field of view, follow these steps:
+ *
+ * 1.Choose your desired field of view angle (`fovy`). This is how wide the camera can see.
```
davepagurek, line 142: Right now these are getting rendered as paragraphs. I think we want to make these appear as a proper list:
image

To do so, I believe we need a space after the . after each number, and I think we have to take out the blank lines between each list item.

src/webgl/p5.Camera.js
```diff
@@ -127,9 +127,29 @@ p5.prototype.camera = function (...args) {
...
+ *
+ * 2.To position the camera correctly, use the formula:
```
davepagurek, line 144: Rather than saying "position the camera correctly", maybe we could phrase this like:
```
To ensure that you can see the entire width across horizontally and height across
vertically, place the camera a distance of `(height / 2) / tan(fovy / 2)` back from its
subject.
```

Note that I also put backticks around the math bit, since that refers to code.

src/webgl/p5.Camera.js
```diff
@@ -127,9 +127,29 @@ p5.prototype.camera = function (...args) {
...
+ *   cameraDistance = (height / 2) / tan(fovy / 2);
+ *   This ensures that you can see the entire width across horizontally and height across vertically at the fixed field of view.
+ *
+ * 3.Set the near value to  cameraDistance / 10 and the far value to  cameraDistance * 10 .
```
davepagurek, line 148: I think we can probably omit this step, since there actually isn't a variable to set here, we'll just pass it right into `perspective`.

src/webgl/p5.Camera.js
```diff
@@ -127,9 +127,29 @@ p5.prototype.camera = function (...args) {
...
+ *
+ * 4.Simply, call perspective with the chosen field of view, canvas aspect ratio, and near/far values:
```
davepagurek, line 150: Super tiny tweak, but can we take out "simply" at the start? We generally try to keep things instructional without passing judgment on how easy or hard things are.

src/webgl/p5.Camera.js
```diff
@@ -127,9 +127,29 @@ p5.prototype.camera = function (...args) {
...
+ *    perspective(fovy, width / height, cameraDistance / 10, cameraDistance * 10);
```
davepagurek, line 151: We can put backticks around this too to format it like code:
```suggestion
 *    `perspective(fovy, width / height, cameraDistance / 10, cameraDistance * 10);`
```

src/webgl/p5.Camera.js
```diff
@@ -127,9 +127,29 @@ p5.prototype.camera = function (...args) {
  * vertical field of view, aspect ratio (usually width/height), and near and far
  * clipping planes.
  *
- * If no parameters are given, the following default is used:
- * perspective(PI/3, width/height, eyeZ/10, eyeZ*10),
- * where eyeZ is equal to ((height/2) / tan(PI/6)).
+ * If no parameters are given, the default values are used as:
+ *
+ * fov- The default field of view for the camera is such that the full height of renderer is visible when it is positioned at a default distance of 800 units from the camera.
```
davepagurek, line 132: I think we can maybe improve the readability of this list, perhaps by turning it into a bullet list?
image


To do so, we need to put a `- ` before each item, and take out the empty lines between them. I also think it might flow better once in a bullet list if we use a colon afterwards instead of a dash. Lastly, since each of these is a variable name, we can put them in backticks:
image

src/webgl/p5.Camera.js
```diff
@@ -127,9 +127,29 @@ p5.prototype.camera = function (...args) {
...
+ *
+ * aspect- The default aspect ratio is the ratio of renderer's width to renderer's height.
+ *
+ * near - The default value for the near clipping plane is 0.1 times the default distance from the camera to the point it is looking at i.e 800.
```
davepagurek, line 136:
```suggestion
 * near - The default value for the near clipping plane is 80, which is 0.1 times the default distance from the camera to its subject.
```

src/webgl/p5.Camera.js
```diff
@@ -127,9 +127,29 @@ p5.prototype.camera = function (...args) {
...
+ *
+ * far - The default value for the far clipping plane is 10 times the default distance from the camera to the point it is looking at i.e 800.
```
davepagurek, line 138:
```suggestion
 * far - The default value for the far clipping plane is 8000, which is 10 times the default distance from the camera to its subject.
```

src/webgl/p5.Camera.js
```diff
@@ -156,7 +176,7 @@ p5.prototype.camera = function (...args) {
  *
  *   rotateX(-0.3);
  *   rotateY(-0.2);
- *   translate(0, 0, -50);
+ *   translate(0, 0, -100);
```
davepagurek, line 179: We might need to play around with this some more. It looks like it's getting clipped the other way now:

https://github.com/processing/p5.js/assets/5315059/036cc39c-7927-4b60-9483-e12a7f35cc85

Garima3110:
Thanks @davepagurek  for these suggestions ..! I'll just look into these and get back to u soon.

Garima3110:
@davepagurek  I have tried to implement what you have suggested , But I am not sure about the translate distance setting to 85 in the negative z direction.

davepagurek:
I think the difference is that now, the camera is a fixed distance back (800), it was likely closer to the objects before. How do you these settings look to you:

Translating closer, and translating before rotating to keep things in frame:
```js
  translate(0, 0, 500);
  rotateX(-0.3);
  rotateY(-0.2);
```

Garima3110:
> I think the difference is that now, the camera is a fixed distance back (800), it was likely closer to the objects before. How do you these settings look to you:
> 
> Translating closer, and translating before rotating to keep things in frame:
> 
> ```js
>   translate(0, 0, 500);
>   rotateX(-0.3);
>   rotateY(-0.2);
> ```

Great! Thanks @davepagurek ..!! This absolutely works fine. Also, I would like to add that similar is the case for ortho also , should I replicate the same thing to that also and then commit the changes.

Without adding translate to 500:
image
image

After adding translate to 500
image
image

davepagurek:
That would be great if you could fix those too, thanks!

Garima3110:
Well after going through the camera methods written in camera.js file, I observed this issue with so many methods .

Rather than changing the existing code for the examples for individual methods, I think we should change the default camera settings which are causing this issue in in the local development environment.

What's your call on this @davepagurek ??

davepagurek:
I think the change that did this was the exact change that this PR is improving the documentation for, changing the camera behaviour to have a position that is independent of the field of view. Previously, the camera would move itself forward or backward when changing the window size in order to keep a fixed field of view, but that caused issues when using a custom camera, as it is unclear what should happen to cameras that have been manually positioned. Now that the camera positioning works a bit differently, it makes sense that these examples need to update their camera code a bit.

The alternate approach, where the behaviour would be completely consistent for the examples, we'd likely have to undo https://github.com/processing/p5.js/pull/6216, and rather than use a fixed field of view, move the camera closer to/farther from its subject as the window resizes in order to still fix the bugs related to the camera resetting. Those changes are also nontrivial, so I think it's ok in this case to just update the camera examples.

davepagurek:
Looks great! One last thing, it looks like the docs for `frustum` in both p5 and p5.Camera could also use that same `translate` call, mind adding that there too? then I'll go ahead and merge everything in!

Garima3110:
> Looks great! One last thing, it looks like the docs for `frustum` in both p5 and p5.Camera could also use that same `translate` call, mind adding that there too? then I'll go ahead and merge everything in!

Thanks a lot @davepagurek for your suggestions.! Learnt a lot solving this issue. Please let me know if some other changes are to be made to this one.

davepagurek (approved):
Thanks so much for your work on this!

inaridarkfox4231:
Many questions remain. It's one thing to not execute orbitControl(), but why didn't you check other examples as well?
There is no reason to leave the reference example incorrect. I can't understand why they tried to solve the problem with a simple band-aid solution, even though it could have been done by restoring the camera's specifications by camera function.

PR #6589, opened 2023-11-25
Fix rotation properties being in degrees irrespective of angleMode

### LINKED ISSUE #6565: Multiple reference to non-existant constant `POLYGON` in Renderer2D

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

### p5.js version

1.8.0

### Web browser and version

*

### Operating System

*

### Steps to reproduce this

There are multiple [references](https://github.com/processing/p5.js/blob/v1.8.0/src/core/p5.Renderer2D.js#L857) to `constants.POLYGON` in `src/core/p5.Renderer2D.js` but in `src/core/constants.js` where all constants came from, there is no `POLYGON` defined.

I don't know if it has affected anything (haven't looked into it) but if so it probably is very subtle as this has been around for a very very long time without notice. I would suggest anyone wanting to address this look into whether this problem has an effect or not first.

SlightlyEpic:
Resolves #6565

 Changes:  
 
1. Fixing the incorrectly lowercased reference to `constants.RADIANS` at
https://github.com/processing/p5.js/blob/773eb0d4aaee30238b15d53c955272ac6c53d806/src/events/acceleration.js#L622
This was also casuing rotation and pRotation properties being in degrees irrespective of the `_angleMode`.

2. Changes to `_handleMotion` function were made to make it work with both degree and radian values.

3. Changes in unit tests related to rotation were made as they expected degree values despite the default mode being radian.

4. The `angleMode` method in math/trigonometry.js was also changed to update pRotation properties when the mode is changed, this is needed because otherwise the units of rotation and pRotation could end up different, which may cause the `deviceTurned` event to fire incorrectly.

#### PR Checklist
- [x] `npm run lint` passes

### DIFF
--- a/src/events/acceleration.js
+++ b/src/events/acceleration.js
@@ -656,60 +656,83 @@ p5.prototype._handleMotion = function() {
   }
 
   if (typeof context.deviceTurned === 'function') {
-    // The angles given by rotationX etc is from range -180 to 180.
-    // The following will convert them to 0 to 360 for ease of calculation
+    // The angles given by rotationX etc is from range [-180 to 180] or [-PI to PI].
+    // The following will convert them to [0 to 360] or [0 to 2*PI] for ease of calculation
     // of cases when the angles wrapped around.
     // _startAngleX will be converted back at the end and updated.
-    const wRX = this.rotationX + 180;
-    const wPRX = this.pRotationX + 180;
-    let wSAX = startAngleX + 180;
-    if ((wRX - wPRX > 0 && wRX - wPRX < 270) || wRX - wPRX < -270) {
+
+    // This is multiplied to all constant rotation values used in calculations
+    // to make the equations agnostic to the angleMode
+    let rotationMultipler = 1;
+    if(this._angleMode === constants.RADIANS) {
+      rotationMultipler = constants.DEG_TO_RAD;
+    }
+
+    const wRX = this.rotationX + (180 * rotationMultipler);
+    const wPRX = this.pRotationX + (180 * rotationMultipler);
+    let wSAX = startAngleX + (180 * rotationMultipler);
+    if (
+      (wRX - wPRX > 0 && wRX - wPRX < (270 * rotationMultipler)) ||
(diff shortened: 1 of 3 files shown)


src/events/acceleration.js
```diff
@@ -619,14 +619,14 @@ p5.prototype.setShakeThreshold = function(val) {
...
+  let multiplier = 1;
+  if (this._angleMode === constants.RADIANS) {
+    multiplier = constants.PI / 180.0;
   }
-  this._setProperty('rotationX', e.beta);
-  this._setProperty('rotationY', e.gamma);
-  this._setProperty('rotationZ', e.alpha);
+  this._setProperty('rotationX', e.beta * multiplier);
+  this._setProperty('rotationY', e.gamma * multiplier);
+  this._setProperty('rotationZ', e.alpha * multiplier);
```
limzykenneth, line 629: There is an internal method to convert angles, can you see if you can use that to simplify the implementation instead? https://github.com/processing/p5.js/blob/main/src/math/trigonometry.js#L437

SlightlyEpic: Ive added a new internal method, similar to the one you pointed out which converts from degrees to the current angle mode unit (since that one was missing and was needed here).
Another simplification i made is that in the `_handleMotion` method, instead of switching the equations based on angle mode, rotation properties are now converted into degrees beforehand and all calculations are done in degrees.

SlightlyEpic:
Ive added a new internal method, similar to the one you pointed out which converts from degrees to the current angle mode unit (since that one was missing and was needed here).
Another simplification i made is that in the `_handleMotion` method, instead of switching the equations based on angle mode, rotation properties are now converted into degrees beforehand and all calculations are done in degrees.

limzykenneth:
Looks good. Thanks!

PR #6640, opened 2023-12-17
Fix syntax error re: `new p5.Vector.cross`

lindapaiste:
Fixes the failing tests explained in [this comment](https://github.com/processing/p5.js/issues/6633#issuecomment-1859300727_):
> All of the tests related to the camera `_orbit` method fail with `TypeError: _main.default.Vector.cross is not a constructor` because the code uses improper syntax `new p5.Vector.cross(up, front)`.  `cross` is a `static` method so there shouldn't be a `new` there.

 Changes:
* Replace syntax `new p5.Vector.cross(a, b)` with `p5.Vector.cross(a, b)` in a few places in the `p5.Camera` class.
* A bunch of tests which were failing before will pass now.

#### PR Checklist

- [x] `npm run lint` passes


limzykenneth:
Looks good. Thanks!

PR #6662, opened 2023-12-26
Fix the Array Comparison bug in outputs.js

nikhilkalburgi:
Signed-off-by: nikhilkalburgi <nikhilkalburgi19@gmail.com>

Resolves: #6660

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/accessibility/outputs.js
+++ b/src/accessibility/outputs.js
@@ -494,7 +494,9 @@ p5.prototype._accsOutput = function(f, args) {
   if (!this.ingredients.shapes[f]) {
     this.ingredients.shapes[f] = [include];
     //if other shapes of this type have been created
-  } else if (this.ingredients.shapes[f] !== [include]) {
+  } else if (
+    JSON.stringify(this.ingredients.shapes[f]) !== JSON.stringify([include])
+  ) {
     //for every shape of this type
     for (let y in this.ingredients.shapes[f]) {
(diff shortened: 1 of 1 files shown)


ayushanand308 (commented):
Looks good to me but this might fail if order of properties is different in two arrays even if  though both might be equal.

nikhilkalburgi:
Hi @ayushanand308,
What if we sort them and check?

` JSON.stringify(this.ingredients.shapes[f].slice().sort()) !== JSON.stringify([include].slice().sort()) `

This will create a **copy** , **sort** then **stringify** to compare

limzykenneth:
@nikhilkalburgi That will be very inefficient and considering this will run on every frame, may significantly impact performance if the drawings get complex enough.

nikhilkalburgi:
Hi @limzykenneth,
 Do you think that the initial condition is necessary as it always returns true? How can it get resolved?

limzykenneth:
In the subsequent for loop there is already a check for existence of the element in the list so the check before that is either not needed (ie. be an `else` instead of `else if` or be something that can be completed very quickly. I think for this case not having that earlier check should be fine and is how is has worked until now effectively. The loop can be optimized a bit by breaking early as well.

HusainModiwala:
maybe here we can use isEqual function of lodash library to compare them.

PR #6665, opened 2023-12-27
Addresses issue #6587

Garima3110:
Addresses issue: Use Cubemaps for Light Feature Performance Improvement #6587

Changes to this solution  are still in progress.
I would love to have suggestions to this . Thank you..!

#### PR Checklist

- [x] `npm run lint` passes


davepagurek:
@diyaayay looks like the PR is up here! I'm also going to give it a review soon when I get the chance, but it would be great to get your take on it too!

src/webgl/p5.RendererGL.js
```diff
@@ -1913,23 +1923,49 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
         defaultShaders.imageLightDiffusedFrag
       );
     }
+    // Create a shader for cubemap conversion
+    const cubemapShader = this._pInst.createShader(
```
davepagurek, line 1927: Just above this, we create and cache a shader so that we only ever need one copy. Can we add that caching here too?

src/webgl/p5.Texture.js
```diff
@@ -510,12 +510,64 @@ export function checkWebGLCapabilities({ GL, webglVersion }) {
     : gl.getExtension('OES_texture_half_float');
   const supportsHalfFloatLinear = supportsHalfFloat &&
     gl.getExtension('OES_texture_half_float_linear');
+  const supportsCubemap = gl.getExtension('OES_texture_cube_map');
```
davepagurek, line 513: I think this might only be necessary for WebGL 1, is that something you can double check? If it's built into WebGL2, then this line would be `supportsCubemap = webglVersion === constants.WEBGL2 || gl.getExtesion('OES_texture_cube_map')`

src/webgl/shaders/cubeFragment.glsl
```diff
@@ -0,0 +1,23 @@
+#version 330 core
+out vec4 FragColor;
+in vec3 localPos;
+
+uniform sampler2D equirectangularMap;
+
+const vec2 invAtan = vec2(0.1591, 0.3183);
+
+vec2 SampleSphericalMap(vec3 v)
```
davepagurek, line 9: Might be worth adding a comment mentioning https://learnopengl.com/PBR/IBL/Diffuse-irradiance as a source so that we can refer back to it in the future

src/webgl/shaders/cubeFragment.glsl
```diff
@@ -0,0 +1,23 @@
+#version 330 core
```
davepagurek, line 1: Right now this shader will only work in WebGL 2 mode. For shaders we want to work in both versions, we add these macros to the start https://github.com/processing/p5.js/blob/main/src/webgl/shaders/webgl2Compatibility.glsl by concatenating it here:
https://github.com/processing/p5.js/blob/bda63ff6de3b96d6b51e262221a578dea194759e/src/webgl/p5.RendererGL.js#L83-L85
...and by adding an appropriate prefix for the vertex/fragment shaders like this:
https://github.com/processing/p5.js/blob/bda63ff6de3b96d6b51e262221a578dea194759e/src/webgl/p5.RendererGL.js#L1745-L1748

To make this file use it, it would involve taking out the `#version` (that's added by the prefix we add), and replacing `out` and `in` with the macros `OUT` and `IN` which get `#define`d to the appropriate keyword depending on what WebGL version is being used. It also means not explicitly declaring `out vec4 FragColor` and instead using `OUT_COLOR`, which we also automatically add a definition for.

src/webgl/shaders/lighting.glsl
```diff
@@ -39,7 +39,7 @@ uniform float uQuadraticAttenuation;
 // boolean to initiate the calculateImageDiffuse and calculateImageSpecular
 uniform bool uUseImageLight;
 // texture for use in calculateImageDiffuse
-uniform sampler2D environmentMapDiffused;
+uniform sampler2D environmentMapDiffusedCubemap;
```
davepagurek, line 42: Should this be `samplerCube`?

src/webgl/shaders/lighting.glsl
```diff
@@ -111,7 +111,7 @@ vec3 calculateImageDiffuse( vec3 vNormal, vec3 vViewPosition ){
   vec3 worldCameraPosition =  vec3(0.0, 0.0, 0.0);  // hardcoded world camera position
   vec3 worldNormal = normalize(vNormal);
   vec2 newTexCoor = mapTextureToNormal( worldNormal );
-  vec4 texture = TEXTURE( environmentMapDiffused, newTexCoor );
+  vec4 texture = textureCube( environmentMapDiffusedCubemap, newTexCoor );
```
davepagurek, line 114: Previously, we were using the `TEXTURE`, which refers to a macro in https://github.com/processing/p5.js/blob/main/src/webgl/shaders/webgl2Compatibility.glsl to be either `texture2D` in WebGL1, or `texture` in WebGL2.

We need to do something similar for cubemap textures. In that file we'd have to `#define TEXTURE_CUBE` to either be `textureCube` in WebGL 1, or just `texture` in WebGL 2, and then use `TEXTURE_CUBE(...)` here.

davepagurek (commented):
Thanks for taking this on! I've added some comments here on some bits I think we'll need in order to make this work on WebGL 1 and WebGL 2.

Also once the pieces are all working, it would be great to make a p5 editor sketch using a newly built version of p5 so that we can test it out!

Garima3110:
Thanks a lot @davepagurek for these suggestions ! I'll just look into these and get back to you soon.

src/webgl/p5.Texture.js
Garima3110: yes , You are absolutely correct .
`WebGL2 supports cubemap textures natively, so you don't need to explicitly check for the OES_texture_cube_map extension in most cases.`
Mybad!
I'll just update this one too.

src/webgl/p5.RendererGL.js
Garima3110: I have applied caching mechanism to possibly all the shaders. I am not sure whether this aligns with what you were suggesting through this @davepagurek .PLease have a look into the changes I've made.
Thanks.!

src/webgl/p5.RendererGL.js
```diff
@@ -32,85 +34,114 @@ defineStrokeJoinEnum('ROUND', 0);
 defineStrokeJoinEnum('MITER', 1);
 defineStrokeJoinEnum('BEVEL', 2);
 
-const lightingShader = readFileSync(
+const getCachedShader= (shaderKey, shaderSource)=>{
```
davepagurek, line 37: I think this would need to be not just a `const`, but a property on `p5.RendererGL`, as each instance of a renderer will need its own shaders. We just want the ability to share those shaders within the renderer.

src/webgl/p5.RendererGL.js
davepagurek: I think we don't want to wrap all the existing shader sources in a caching function, since those are just the sources for one part of a shader (e.g. just the vertex or fragment source, and often not the full source either, since we need to add a prefix to it.) A caching function would be good for this `createShader` call (since you're still creating a new shader every time this line is hit, currently.) A few lines up, `this.diffusedShader` already does caching, but can be refactored to use the same caching system you use here if you're factoring that out into a common pattern.

davepagurek: oh actually I see that you've added that in `_getCubemapShader`. I think rather than making a new shader here, we can just use `const cubemapShader = this._getCubemapShader()` instead.

src/webgl/p5.RendererGL.js
```diff
@@ -32,85 +34,114 @@ defineStrokeJoinEnum('ROUND', 0);
...
+  return shaderCache[shaderKey];
+};
+
+const lightingShader = getCachedShader('lightingShader',readFileSync(
   join(__dirname, '/shaders/lighting.glsl'),
   'utf-8'
-);
-const webgl2CompatibilityShader = readFileSync(
+));
+const webgl2CompatibilityShader = getCachedShader('webgl2CompatibilityShader',readFileSync(
```
davepagurek, line 48: I don't think we want to wrap all of these in `getCachedShader` -- these are not full shaders, just pieces of shader source that will turn into a full shader.

src/webgl/p5.RendererGL.js
```diff
@@ -32,85 +34,114 @@ defineStrokeJoinEnum('ROUND', 0);
 defineStrokeJoinEnum('MITER', 1);
 defineStrokeJoinEnum('BEVEL', 2);
 
-const lightingShader = readFileSync(
+const getCachedShader= (shaderKey, shaderSource)=>{
+  if(!shaderCache[shaderKey]){
+    shaderCache[shaderKey]=p5.createShader(shaderSource);
```
davepagurek, line 39: Shaders also have both a vertex and a fragment shader, not a single shader source, so you'd need to add two parameters to the function and pass both into `createShader`.

That said, your `_getCubemapShader` implementation on its own does all the caching we need, so we might not need this in addition to that.

src/webgl/p5.Shader.js
```diff
@@ -548,6 +548,12 @@ p5.Shader = class {
           uniform.texture.src._animateGif(this._renderer._pInst);
         }
         break;
+      case gl.SAMPLER_CUBE:
```
davepagurek, line 551: Is this block the same as the one above? If so, we could do:
```js
case gl.SAMPLER_2D:
case gl.SAMPLER_CUBE:
        gl.activeTexture(gl.TEXTURE0 + uniform.samplerIndex);
        uniform.texture =
          data instanceof p5.Texture ? data : this._renderer.getTexture(data);
        // ...
```
...to handle both cases with the same code

davepagurek (commented):
Thanks for working on these changes! I left a few notes on the shader caching. I think if you were to try running this right now on a sketch using `imageLight()`, you'd run into some runtime errors. For next steps, I'd encourage you to try making a test sketch, running `grunt browserify` to generate `lib/p5.js`, then using that in the sketch to test out whether everything still runs. That will help ensure that you catch all the issues that one might run into when using the feature more easily than just looking at the code.

Once that's working, it would also be great to put that up in a p5 editor sketch so people looking at the PR can also test it out!

Garima3110:
>Once that's working, it would also be great to put that up in a p5 editor sketch so people looking at the PR can also test it out!

Yeah sure @davepagurek ..!Thankyou for the suggestions , I'll just make these changes and get back to you soon :-)

Garima3110:
> Once that's working, it would also be great to put that up in a p5 editor sketch so people looking at the PR can also test it out!

Hey @davepagurek ! I tried running this test sketch after building p5.js
```
let cubemap;
function preload() {
  // Load cubemap texture images
  let imgPaths = [
    '1/px.png',
    '1/nx.png',
    '1/py.png',
    '1/ny.png',
    '1/pz.png',
    '1/nz.png'
  ];

  cubemap = [];
  for (let i = 0; i < 6; i++) {
    cubemap[i] = loadImage(imgPaths[i]);
  }
}

function setup() {
}
function draw() {
  background(220);
  imageMode(CENTER);
  push();
  translate(0, 0, -200);
  scale(2);
  image(img, 0, 0, width, height);
  pop();
  ambientLight(50);
  const cubemapTexture=new CubemapTexture(WEBGL,cubemap,{});
  imageLight(cubemapTexture);
  specularMaterial(20);
  noStroke();
  scale(2);
  rotateX(frameCount * 0.005);
  rotateY(frameCount * 0.005);
  box(25);
}
```
And got this on the console window:
image

davepagurek:
I think you'll need to upload `lib/p5.js` to your sketch, and then edit `index.html` in your sketch to reference your uploaded file rather than the CDN's file.

Garima3110:
> I think you'll need to upload `lib/p5.js` to your sketch, and then edit `index.html` in your sketch to reference your uploaded file rather than the CDN's file.
> 
> Also, I think the idea is for the cubemap texture to only be used internally, so you'd still preload just one equirectangular image, and pass that into `imageLight()`. But then when it creates the diffused texture, it would be stored internally as a cubemap.

Oh yeah got it ..Thanks a lot for this..!

Garima3110:
> I think you'll need to upload `lib/p5.js` to your sketch, and then edit `index.html` in your sketch to reference your uploaded file rather than the CDN's file.
> 
> Also, I think the idea is for the cubemap texture to only be used internally, so you'd still preload just one equirectangular image, and pass that into `imageLight()`. But then when it creates the diffused texture, it would be stored internally as a cubemap.

Thanks @davepagurek for the suggestion,  sorry didn't know about this earlier.
This is the sketch now, please have a look , and let me know if there are some other suggestions.
https://editor.p5js.org/Garima3110/sketches/CeG_DQ1n5
Thanks once again..!

Garima3110:
> This is the sketch now, please have a look , and let me know if there are some other suggestions. https://editor.p5js.org/Garima3110/sketches/CeG_DQ1n5
> 
> hii @davepagurek :-) I am not able to figure out why exactly are the tests failing even after trying out this sketch could you please provide some insights on this. It would be very helpful for me ..Thanks a lot..!

I apologise directly pinging you again @davepagurek ! But could you please me out in this?!

src/webgl/p5.RendererGL.js
```diff
@@ -32,77 +33,135 @@ defineStrokeJoinEnum('ROUND', 0);
 defineStrokeJoinEnum('MITER', 1);
 defineStrokeJoinEnum('BEVEL', 2);
 
-const lightingShader = readFileSync(
+this._pInst.shaderCache={};
```
davepagurek, line 36: I think this is probably the cause, over here, `this` is undefined. If you want it to be initialized on a `RendererGL` instance, I think moving this line to the `RendererGL` constructor would work.

src/webgl/p5.RendererGL.js
```diff
@@ -32,77 +33,134 @@ defineStrokeJoinEnum('ROUND', 0);
...
+    const fragmentShader = this._pInst.createShader(
+      this._pInst.FRAGMENT_SHADER, fragmentShaderSource);
+
+    this._pInst.shaderCache[shaderKey] = this._pInst.createShaderProgram(
+      vertexShader, fragmentShader);
+  }
+  return this._pInst.shaderCache[shaderKey];
+};
+
+const lightingShader = p5.RendererGL.prototype.getCachedShader('lightingShader',readFileSync(
```
davepagurek, line 53: I think right now this is creating `p5.Shader` instances where before it was just loading shader *source code.* We still need to load the source code here, but we can't actually instantiate a `p5.Shader` until we have a p5 *instance*. Right now, you're calling the method directly on the class itself rather than an instance of the class, so `this` will be undefined.

I think the spot to call the `getCachedShader` method is inside of instance method like `getDefaultFillShader()`.

Garima3110:
Hey @davepagurek !
(I apologise if this approach did sound silly.)
I tried reordering some code in RendererGL file and did a few other changes to get rid of the error of `this` being undefined and this problem got solved too.
But while doing this , the `filterShaderVert` and `filterShaderFrags` not being used but declared , give the eslint errors.
Should I just disable ESLint for those specific lines , or do you have other suggestions in this regard ?!
image
image

Garima3110:
Also, 
this is what I'm getting right now :
https://editor.p5js.org/Garima3110/sketches/CeG_DQ1n5

image

src/webgl/p5.RendererGL.js
```diff
@@ -830,6 +780,122 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+      join(__dirname, '/shaders/lighting.glsl'),
+      'utf-8'
+    ));
+    const webgl2CompatibilityShader = readFileSync(
+      join(__dirname, '/shaders/webgl2Compatibility.glsl'),
+      'utf-8'
+    );
+
+    defaultShaders = {
+      immediateVert: renderer.getCachedShader('immediateVert',readFileSync(
```
davepagurek, line 795: One issue with this current structure is that we're calling `getCachedShader` in the `in `_resetContext`, so on load, it will try to make a bunch of shaders.

We maybe want to keep the shader source code in objects outside the constructor as it was before, and then we only call `getCachedShader` when we need to *use* the shader. e.g.:
```js
  _getCubemapShader() {
    return this.getCachedShader(
      'defaultCubemapShader', 
      this._webGL2CompatibilityPrefix('vert', 'mediump') +
        defaultShaders.cubemapVertexShader,
      this._webGL2CompatibilityPrefix('frag', 'mediump') +
        defaultShaders.cubemapFragmentShader
    )
  }
```

src/webgl/shaders/lighting.glsl
```diff
@@ -114,7 +114,7 @@ vec3 calculateImageDiffuse( vec3 vNormal, vec3 vViewPosition ){
   vec3 worldCameraPosition =  vec3(0.0, 0.0, 0.0);  // hardcoded world camera position
   vec3 worldNormal = normalize(vNormal);
   vec2 newTexCoor = mapTextureToNormal( worldNormal );
-  vec4 texture = TEXTURE( environmentMapDiffused, newTexCoor );
+  vec4 texture = TEXTURE_CUBE( environmentMapDiffusedCubemap, newTexCoor );
```
davepagurek, line 117: I wonder if the naming of `texture` is an issue here since in a WebGL 2 environment, `TEXTURE_CUBE` is also `#define`d as the builtin function `texture()`? Does the error go away if you rename this variable?

src/webgl/p5.RendererGL.js
```diff
@@ -1919,23 +1954,47 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+        this._pInst.rectMode(constants.CENTER);
+        this._pInst.noLights();
+        this._pInst.rect(0, 0, width, height);
+
+        // Capture the rendered face and store it in the faces array
+        faces[i] = this._pInst.get(0, 0, width, height);
+      });
+    }
+    // Use the diffusedShader for rendering
     newFramebuffer.draw(() => {
```
davepagurek, line 1978: Probably not the cause of the error you're seeing, but I think we probably no longer need this `newFramebuffer.draw(...)` block since we are no longer using this framebuffer as the texture. (I think that means we actually need to call `newFramebuffer.free()` here to avoid a small memory leak?)

src/webgl/p5.RendererGL.js
```diff
@@ -1919,23 +1954,47 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+        cubemapShader.setUniform('projection', captureProjection);
+        cubemapShader.setUniform('view', captureViews[i]);
+
+        this._pInst.noStroke();
+        this._pInst.rectMode(constants.CENTER);
+        this._pInst.noLights();
+        this._pInst.rect(0, 0, width, height);
+
+        // Capture the rendered face and store it in the faces array
+        faces[i] = this._pInst.get(0, 0, width, height);
```
davepagurek, line 1974: Also probably not the source of the error, but we might need to put this outside of the `draw(...)` block, and use `faces[i] = newFramebuffer.get()` instead of `this._pInst.get(...)`.

src/webgl/shaders/cubeVertex.vert
```diff
@@ -0,0 +1,13 @@
+// Adapted from:  https://learnopengl.com/PBR/IBL/Diffuse-irradiance
+layout (location = 0) in vec3 aPos;
```
davepagurek, line 2: This might also need to be changed into `IN vec3 aPos` in order to also work in WebGL 1 environments

src/webgl/shaders/lighting.glsl
davepagurek, line 117: OH you know what the issue might be? `TEXTURE_CUBE` takes in a `vec3` as the coordinate instead of a `vec2`. The compiler probably doesn't know what to do with the combination of inputs we're giving it here, and that could be causing some more errors further on.

I think we can pass the normal directly in, like: `TEXTURE_CUBE(environmentMapDiffusedCubemap, worldNormal)`

Garima3110: > OH you know what the issue might be? `TEXTURE_CUBE` takes in a `vec3` as the coordinate instead of a `vec2`. The compiler probably doesn't know what to do with the combination of inputs we're giving it here, and that could be causing some more errors further on.
> 
> I think we can pass the normal directly in, like: `TEXTURE_CUBE(environmentMapDiffusedCubemap, worldNormal)`

Yes, you're correct! In many GLSL implementations, the textureCube function takes a vec3 as the first argument, representing the 3D direction vector in the cube map space. So, using `TEXTURE_CUBE(environmentMapDiffusedCubemap, worldNormal)` is likely the correct way to sample from the cube map based on the normal vector.
Thanks a lot for looking into this @davepagurek !!

Garima3110:
@davepagurek I fixed some errors somehow, now I'm stuck with this one. Could you give some suggestion on this one too?!
Thankyou.
Sketch with the latest build:
https://editor.p5js.org/Garima3110/sketches/CeG_DQ1n5

image

davepagurek:
When I run the code I'm getting a slightly different error in the console:
image

Looks like it's coming from this:
```js
var captureProjection = matrixInstance.mat4.perspective(matrixInstance.mat4.get(), Math.PI / 2, 1, 0.1, 10);
```

It looks like that bit of the sketch isn't in your code in this PR, is the PR or the sketch possibly using out of date code?

Garima3110:
> It looks like that bit of the sketch isn't in your code in this PR, is the PR or the sketch possibly using out of date code?

@davepagurek  Thanks a lot for pointing this out..!

I apologise deeply for the confusion here, and for the delay too.. 
Everything seems to work fine here are the changes:

Sketch with the latest build:
https://editor.p5js.org/Garima3110/sketches/CeG_DQ1n5

Cubemap implementation:
image

Equirectangular map implementation:
image

The only change I could notice here was in the cubemap implementation , that the sphere had a sort of less brightness.

davepagurek:
Looks like it's hitting an error here because `gl` is undefined:
image

I think that's because it's expecting a renderer to be passed in, but we're passing in a p5 instance instead when we do `cubemapTexture=new CubemapTexture(this._pInst,faces, {});`. Maybe try `cubemapTexture=new CubemapTexture(this,faces, {});` instead (with just `this` as the first argument)?

Garima3110:
Sorry for the extreme delay @davepagurek !
No error now, But the desired result is missing, 
Sketch with the latest build : https://editor.p5js.org/Garima3110/sketches/CeG_DQ1n5
I am looking for the cause of the same, if you have any suggestions please do let me know, Thankyou!

src/webgl/p5.RendererGL.js
```diff
@@ -1899,44 +2011,100 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+      lookAt([0, 0, 0], [0, 1, 0], [0, 0, 1]), // Positive Y
+      lookAt([0, 0, 0], [0, -1, 0], [0, 0, -1]), // Negative Y
+      lookAt([0, 0, 0], [0, 0, 1], [0, -1, 0]), // Positive Z
+      lookAt([0, 0, 0], [0, 0, -1], [0, -1, 0]) // Negative Z
+    ];
+
+    // Get the cubemap shader
+    const cubemapShader = this._getCubemapShader();
+    this._pInst.shader(cubemapShader);
+    cubemapShader.setUniform('equirectangularMap', 0);
```
perminder-17, line 2057: This can't be hardcoded to 0. It must have a variable, a sampler2D variable specifically.

src/webgl/shaders/cubeFragment.frag
```diff
@@ -0,0 +1,22 @@
+// Adapted from: https://learnopengl.com/PBR/IBL/Diffuse-irradiance
+IN vec3 localPos;
```
perminder-17, line 2: In my opinion I feel for maximum compatibility, we can do this one in GL ES 100 instead of 300? Using attribute instead of IN, using the spacial variable gl_FragColor instead of defining an OUT_COLOR, and using texture2D() instead of texture()? What you feel?

perminder-17 (changes requested):

src/webgl/p5.RendererGL.js
```diff
@@ -2113,7 +2281,7 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
       // this.activeImageLight has image as a key
       // look up the texture from the diffusedTexture map
       let diffusedLight = this.getDiffusedTexture(this.activeImageLight);
-      shader.setUniform('environmentMapDiffused', diffusedLight);
+      shader.setUniform('environmentMapDiffusedCubemap', diffusedLight);
```
perminder-17, line 2284: I am not sure but I feel diffusedLight is also a sampler2D texture, could it be passed as a uniform with samplerCube texture?

src/webgl/p5.RendererGL.js
```diff
@@ -1899,44 +2011,100 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+    // Get the cubemap shader
+    const cubemapShader = this._getCubemapShader();
+    this._pInst.shader(cubemapShader);
+    cubemapShader.setUniform('equirectangularMap', 0);
+
+    // Define captureProjection as a 4x4 projection matrix
+    let captureProjection = new p5.Matrix();
+    captureProjection.perspective(Math.PI / 2.0, 1.0, 0.1, 10.0);
+
+    cubemapShader.setUniform('projection', captureProjection.mat4);
```
perminder-17, line 2063: We should not set uniforms and again create a new uniform. I think we already have our uniform named uProjectionMatrix.

src/webgl/p5.RendererGL.js
```diff
@@ -1899,44 +2011,100 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
...
+        this.GL.FRAMEBUFFER,
+        this.GL.COLOR_ATTACHMENT0,
+        this.GL.TEXTURE_CUBE_MAP_POSITIVE_X + i,
+        envCubemap,
+        0
+      );
+
+      this.GL.clear(this.GL.COLOR_BUFFER_BIT | this.GL.DEPTH_BUFFER_BIT);
+
+      cubemapShader.setUniform('view', captureViews[i].mat4);
```
perminder-17, line 2089: Same with this. We should not calculate the view matrix. We already had set up the modelVeiw matrix as "uModelViewMatrix".

src/webgl/p5.RendererGL.js
Garima3110: Here, diffusedLight is obtained from `getDiffusedTexture`, which returns a `CubemapTexture` object. Since `environmentMapDiffusedCubemap` in the shader expects a cubemap texture (samplerCube), so I think  `diffusedLight` is being correctly assigned to the `environmentMapDiffusedCubemap` uniform in the shader, which is of type `samplerCube`.

src/webgl/shaders/cubeFragment.frag
Garima3110: Well I started from that only , and was getting a lot of errors and from Dave's suggestion of using the `macros` instead resolved the errors , I again tried doing what u have suggested today, but strangely that gives in more errors! So I think we should stick to using these macros only :-)
You can refer to Dave's suggestion here in this comment:
https://github.com/processing/p5.js/pull/6665#discussion_r1437846579

src/webgl/p5.RendererGL.js
Garima3110: Actually I had this in mind that in WebGL (and OpenGL), when you set a texture uniform in a shader, you also need to specify which texture unit the texture is bound to.
If you see the next 2 lines:
``` js
this.GL.activeTexture(this.GL.TEXTURE0);
this.GL.bindTexture(this.GL.TEXTURE_2D, input);
```
Here, ` this.GL.activeTexture(this.GL.TEXTURE0)` activates texture unit 0, and `this.GL.bindTexture(this.GL.TEXTURE_2D, input)` binds the texture input to texture unit 0.
But lets just keep it simple and stick to what you have suggested. That seems to work fine.!

src/webgl/shaders/cubeFragment.frag
perminder-17: Hmmm...It looks like you need to use both webgl-1 and webgl-2 modes on shaders.

perminder-17:
From my honest opinion I feel like you don't need to work anymore on the shaders part. vertex shaders looks good, Also fragment shaders looks correct to me. The only thing I want to recheck is, are we getting the correct uniforms in fragment shaders from the CPU?

src/webgl/shaders/cubeFragment.frag
```diff
@@ -0,0 +1,22 @@
...
+    vec2 uv = vec2(atan(v.z, v.x), asin(v.y));
+    uv *= invAtan;
+    uv += 0.5;
+    return uv;
+}
+
+void main()
+{        
+    vec2 uv = SampleSphericalMap(normalize(localPos));
+    vec3 color = texture(equirectangularMap, uv).rgb;
```
perminder-17, line 19: I feel like, `texture()` is worked on webgl-2 and `texture2D()` is worked on webgl-1. You may use `TEXTURE` to run on both the modes.

src/webgl/shaders/cubeVertex.vert
```diff
@@ -0,0 +1,13 @@
+// Adapted from:  https://learnopengl.com/PBR/IBL/Diffuse-irradiance
+layout (location = 0) IN vec3 aPosition;
```
perminder-17, line 2: This looks good, but I feel like do we actually need this layout(). Can it be done just with `IN vec3 aPosition;`

src/webgl/p5.RendererGL.js
perminder-17: Oooh...sorry. I overlooked this, previously we had diffusedLight with `sampler2D` texture so I got confused. I see you have updated getDiffusedTexture()

src/webgl/shaders/cubeVertex.vert
Garima3110: Well honestly speaking , I'm also not sure about what exactly it signifies here, but as i searched on web , I found this
If your application code (typically written in a language like C++ or JavaScript) explicitly binds attribute locations using `glBindAttribLocation`, you can omit the `layout(location = 0)` in the shader. The application code would look something like this:
``` js
glBindAttribLocation(program, 0, "aPosition");
```
If your application does not do this, you should keep `layout(location = 0)` to ensure that the attribute location is explicitly defined.

Also , just to mention , I tried running the test sketch by omitting this, and the results were similar , So I think we should keep it for a safer side as its not causing any error.
What say?!

src/webgl/shaders/cubeFragment.frag
Garima3110: I had tried that earlier, tried that again today as per your suggestion but both give the same results.
https://github.com/processing/p5.js/pull/6665#discussion_r1461026949
 So i think we should go with `texture` only. :-)

src/webgl/shaders/cubeVertex.vert
perminder-17: Yes, this should not give any errors. Since, vertex shaders are just dealing with the positions, I think we don't have layout(...) at other files and they are working perfectly. 

But the main reason in my mind is, since we only use `layout` with `in` we can't use it with `attributes`, where I think is `layout` used only for webgl-2? If that the case, `layout` should be omitted.

src/webgl/shaders/cubeFragment.frag
perminder-17: hmm...what kind of errors? If you want your code to run on both environments I think `TEXTURE` shouldn't give errors. I don't know why? We can debug it by first writing the code for webgl-1 (changing it to GL ES 100 version I mentioned above, let me know if `texture2D` works there), similarly change it to version 300 and let me know if `texture()` works there? 

Sorry if it seems a long approach for you, But I seriously don't know why `TEXTURE()` is giving you errors. Also, can you share me the screenshots of the errors you are getting.

PR #6669, opened 2023-12-28
Enhanced loadModel() method signature with independent U and V flipping options.

### LINKED ISSUE #5021: Add independent loadModel() U and V flipping flags
Hello,

The feature suggests adding extra parameters to loadModel() so that OBJ files can have their U and/or V texture coordinates flipped, in order to fine-tune flipped texture displays.

- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [x] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [x] WebGL

In issue #4866 and related PR #5017, we fixed p5js's OBJ vertex UV texture coordinates parsing's behaviour to more closely match that of Processing, where the V coordinate is inverted, to anticipate WebGL's flipping of that V/Y coordinate. Still, this does not solve all texture horizontal or vertical flipping issues for people exporting .OBJ files for a p5js import using many kinds of 3d software.

Diverging slightly from Processing's loadShape() (which tolerates .OBJ files to make 3d geometries)'s signature, why not add 2 parameters to the p5js `loadModel(modelPath:String, normalize:boolean)`, giving one of those possibilities:
-  `loadModel(modelPath:String, normalize:boolean, flipU:boolean, flipV:boolean)`
-  `loadModel(modelPath:String, normalize:boolean, flipUV:list)` where flipUV would be a list of two 0-1 integers, eg. [0,1] for U and V respectively
- .. open to more ideas

Thanks @stalgiag for pushing me into adding this new issue

deveshidwivedi:
Resolves #5021 

 Changes:

The existing ```loadModel()``` method signatures are maintained, but a new, more flexible method is introduced as ```loadModel(path, [options]```. This new method allows users to specify optional parameters in an options object, providing greater control over the loading process without being constrained by parameter order. The addition of ```flipU()``` and ```flipV()``` options enables users to apply U and V flipping to models after they've been loaded.

 Screenshots of the change:

before:
Screenshot 2023-12-30 115103

after:
Screenshot 2023-12-28 131513

before:
Screenshot 2023-12-30 115132

after:
Screenshot 2023-12-30 115009

#### PR Checklist

- [x] `npm run lint` passes


deveshidwivedi:
I have made the mentioned changes, please take a look. @davepagurek Thank you!

src/webgl/loading.js
```diff
@@ -177,6 +177,94 @@ p5.prototype.loadModel = function(path) {
...
+ * @method loadModel
+ * @param  {String} path
+ * @param  {Object} [options]
+ * @param  {function(p5.Geometry)} [options.successCallback]
+ * @param  {function(Event)} [options.failureCallback]
+ * @param  {String} [options.fileType]
+ * @param  {boolean} [options.normalize]
+ * @return {p5.Geometry} the <a href="#/p5.Geometry">p5.Geometry</a> object
+ */
+p5.prototype.loadModel = function(path,options) {
```
davepagurek, line 190: It looks like we also still have the old implementation of `loadModel` above this. I think we would need to delete lines 106-176 so that the doc comments added here come directly after the last doc comment above the previous implementation.

src/webgl/p5.Geometry.js
```diff
@@ -142,6 +142,24 @@ p5.Geometry = class Geometry {
     this.vertexColors = [];
     return this;
   }
+  flipU() {
```
davepagurek, line 145: It would be good to add some doc comments for these methods too so we can show some examples of their usage.

src/webgl/loading.js
```diff
@@ -177,6 +177,94 @@ p5.prototype.loadModel = function(path) {
...
+  p5._validateParameters('loadModel', arguments);
+  let normalize= false;
+  let successCallback;
+  let failureCallback;
+  let fileType = path.slice(-4);
+  if (options && typeof options === 'object') {
```
davepagurek, line 196: This if/else structure looks good! Now that we have a version where we can pass in an object, did we also want to add optional `flipU` and `flipV` properties to the object? Where, if they are set to true, before calling the success callback, we would call `.flipU()` and/or `.flipV()` on the model?

davepagurek (commented):
Nice work on this so far! I've added a few more comments about filling out the documentation, and whether or not we want to also add flipU and flipV support to the options object for convenience.

src/webgl/loading.js
deveshidwivedi: I removed the mentioned lines but the tests are failing.

src/webgl/loading.js
deveshidwivedi: These properties are now added, do they look okay?

src/webgl/p5.Geometry.js
deveshidwivedi: Should it be like this?
```javascript
/**
 * Flip the U texture coordinates of the model.
 * @method flipU
 * @for p5.Geometry
 *
 * @returns {p5.Geometry} Returns the modified p5.Geometry object.
 *
 * @example
 * <div>
 * <code>
 * let model;
 * 
 * function preload() {
 *   model = loadModel('path/to/your/model.obj', modelLoaded);
 * }
 * 
 * function modelLoaded() {
 *   model.flipU();
 *   
 *   createCanvas(400, 400, WEBGL);
 *   background(200);
 *   
 *   // Original model (no texture coordinate flipping)
 *   fill('orange');
 *   translate(-100, 0, 0);
 *   model(model);
 *   
 *   // Flipped U texture coordinates model
 *   fill('blue');
 *   translate(200, 0, 0);
 *   model(model);
 * }
 * </code>
 * </div>
 */
```

davepagurek: Right, like that! Maybe instead of doing everything in a success callback, we can omit the callback, and just rename `modelLoaded` to `setup` (since it will wait for preload to finish before running it anyway.)

davepagurek:
Btw looks like tests are failing due to this linting error:

```
Error: unner/work/p5.js/p5.js/src/webgl/loading.js:106:3: Parsing error: 'return' outside of function [Error]
```

deveshidwivedi:
I tried adding this example:
```javascript
 /**
 * Flips the U texture coordinates of the model.
 * @method flipU
 * @for p5.Geometry
 *
 * @returns {p5.Geometry}
 *
 * @example
 * <div>
 * <code>
 * let modelU;
 *
 * function preload() {
 *   modelU = loadModel('path/to/your/model.obj');
 *
 *   modelU.flipU();
 * }
 *
 * function setup() {
 *   createCanvas(400, 400, WEBGL);
 *   background(200);
 *
 *   // Original model (no texture coordinate flipping)
 *   fill('orange');
 *   translate(-100, 0, 0);
 *   model(modelU);
 *
 *   // Flipped U texture coordinates model
 *   fill('blue');
 *   translate(200, 0, 0);
 *   model(modelU);
 * }
 * </code>
 * </div>
 */

  flipU() {
    this.uvs = this.uvs.flat().map((val, index) => {
      if (index % 2 === 0) {
        return 1 - val;
      } else {
        return val;
      }
    });
  }
  ```
  
But I'm getting a timeout error, how could it be solved?

```
 Error: Timeout of 4000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

**Comment by deveshidwivedi** [comment]
I tried adding the examples, how could these errors be solved?

**Comment by davepagurek** [comment]
image

Interesting that it's a timeout error. I wonder if there's another error but that isn't getting logged in the tests. If you launch a local docs server with `grunt yui:dev` and go to the docs for these methods on p5.Geometry, do the examples run successfully?

File: src/webgl/p5.Geometry.js [hunk 4]
```diff
@@ -143,6 +143,94 @@ p5.Geometry = class Geometry {
...
+ *
+ * @returns {p5.Geometry}
+ *
+ * @example
+ * <div>
+ * <code>
+ * let modelU;
+ *
+ * function preload() {
+ *   modelU = loadModel('path/to/your/model.obj');
```
davepagurek, line 158: oh wait it's probably because this isn't a "real" model. Maybe try borrowing a model path from e.g. the `loadModel` examples and reusing that here?

deveshidwivedi:
I tried borrowing model path from the ```loadModel``` examples and the tests do not fail anymore. I'm getting the same results for both models:
FlipU():
Screenshot 2023-12-29 120208


FlipV():
Screenshot 2023-12-29 120218

On logging results for flipV, this is what I'm getting:
Screenshot 2023-12-29 115828

davepagurek:
That's expected if you fill everything with a solid color. Since we're only flipping the texture coordinates, we will only see a difference if you load an image and use it as a texture. Maybe we can do that in the examples?

deveshidwivedi:
Right! 

On trying to load an image, the example doesn't seem to work.   

```javascript
* let img
 * let myModel;
 *
 *   function preload() {
 *   img = loadImage('assets/laDefense.jpg');
 *   myModel = loadModel('assets/octahedron.obj');
 * }
 *
 * function setup() {
 *   createCanvas(400, 400, WEBGL);
 *   background(200);
 * }
 * function draw() {
 *  background(0);
 *   translate(-150, 0, 0);
 *   texture(img);
 *   noStroke();
 *   plane(100);
 *
 *  translate(150, 0, 0);
 *  flipU(myModel);
 *  texture(img);
 *  noStroke();
 *  plane(100);
 *  model(myModel); 
 *
 * }
 * </code>
 * </div>
```
Could you suggest how I can make it better?

davepagurek:
Just took a look again! Some notes:
- It looks like you're calling `flipU(myModel);` instead of `myModel.flipU()`, since it's a method on the object.
- It also looks like the octahedron model does not have texture coordinates in it, so flipping them has no effect. Maybe rather than loading a model, it would be better to manually create a simple model with texture coordinates? e.g. in preload, doing:
   ```js
   myModel = buildGeometry(() => plane(50));
   ```
- Right now you're flipping the U values in `draw`. This should be done in setup, since once a model is drawn, updating its attributes (at least currently) will have no effect, since the resources are cached. Maybe we can make *two* models, that start off the same, but then we call `.flipU()` on one of them. Then you can draw them side-by-side in `draw()`.
- If you run `grunt yui:dev`, it will launch a local version of the p5 reference so you can test your examples. This will help catch a few other minor issues, e.g. the example canvas looks pretty big compared to the text, so maybe it can be made smaller, and also the two side-by-side models being offcenter.

deveshidwivedi:
Thank you very much, @davepagurek. Do you think any of these are okay to go ahead with, how can I make it work?
 
```javascript
let img;
let model1;
let model2;

function preload() {
  img = loadImage('assets/laDefense.jpg');
}

function setup() {
  background(200);

  model1 = buildGeometry(() => plane(50));

  model2 = buildGeometry(() => plane(50));
  model2.flipU();
}

function draw() {
  background(0);

  // original
  push();
  translate(-25, -25, 0);
  texture(img);
  noStroke();
  plane(50);
  model(model1);
  pop();

  //flipped
  push();
  translate(25, 25, 0);
  texture(img);
  noStroke();
  plane(50);
  model(model2);
  pop();
}
```

```javascript
let img;
let model1;
let model2;

function preload() {
  img = loadImage('assets/laDefense.jpg');
}

function setup() {
  background(200);

  model1 = createShape(50, 50);
  model1.texture(img);

  model2 = createShape(50, 50);
  model2.texture(img);
}

function draw() {
  console.log('Drawing frame...');
  background(0);

  //original
  console.log('Drawing original model...');
  push();
  translate(-25, -25, 0);
  texture(img);
  noStroke();
  plane(50);
  model(model1);
  pop();

  // flipped
  console.log('Drawing flipped model...');
  push();
  translate(25, 25, 0);
  texture(img);
  noStroke();
  plane(50);
  model(model2);
  pop();
}

function createMyShape(w, h) {
  let myShape = createGraphics(w, h);
  myShape.beginShape();
  myShape.vertex(-w / 2, -h / 2, 0, 0);
  myShape.vertex(w / 2, -h / 2, 1, 0);
  myShape.vertex(w / 2, h / 2, 1, 1);
  myShape.vertex(-w / 2, h / 2, 0, 1);
  myShape.endShape(CLOSE);
  return createModel(myShape);
}

```

davepagurek:
I think that first version will work! The second one could also work with some modifications (and would maybe actually be preferable since you can see the UV values being set!)

The secnd example is currently returning a `p5.Graphics` but trying to draw it with `model()`, which only works for `p5.Geometry`. If you want to turn your shape into geometry, it could look like this:
```js
function createMyShape(w, h) {
  return buildGeometry(() => {
    textureMode(NORMAL);
    beginShape();
    vertex(-w / 2, -h / 2, 0, 0);
    vertex(w / 2, -h / 2, 1, 0);
    vertex(w / 2, h / 2, 1, 1);
    vertex(-w / 2, h / 2, 0, 1);
    endShape(CLOSE);
  })
}
```

deveshidwivedi:
Hi @davepagurek please take a look, the examples work now.

src/webgl/loading.js
```diff
@@ -103,22 +103,42 @@ import './p5.Geometry';
...
  * @return {p5.Geometry} the <a href="#/p5.Geometry">p5.Geometry</a> object
  */
-p5.prototype.loadModel = function(path) {
+/**
+ * @method loadModel
+ * @param  {String} path
+ * @param  {Object} [options]
+ * @param  {function(p5.Geometry)} [options.successCallback]
+ * @param  {function(Event)} [options.failureCallback]
+ * @param  {String} [options.fileType]
```
davepagurek, line 112: When I run `grunt yui:dev`, it looks like our reference UI unfortunately doesn't currently display these nested parameters:

image

[For the `createFramebuffer` docs](https://p5js.org/reference/#/p5/createFramebuffer) we list the options just in the description of the function as a bullet list:
https://github.com/processing/p5.js/blob/37d3324b457b177319ed65468201f2806a66eff5/src/core/rendering.js#L306-L313

Maybe we should do that here too?

davepagurek (commented):
Awesome, the examples look great!

I think the last thing to do is make sure that the new options in `loadModel` get displayed in the p5 reference properly. I left a comment with a suggestion there. After that we should be good to go!

src/webgl/loading.js
deveshidwivedi: Sure, I just listed the options, do we need any kind of descriptions here?

davepagurek: I think a short description would be helpful, maybe just a short sentence for each? Then we're good to go!

deveshidwivedi: Done!

davepagurek (approved):
Thanks for updating the docs, I think this is good to merge now!

davepagurek:
@all-contributors please add @deveshidwivedi for code

PR #6703, opened 2024-01-06
Implement mirrored video feature for createCapture (#6441)

### LINKED ISSUE #6441: Support mirrored video for createCapture
### Increasing Access

For many creative coding projects that involve the webcam, we want to display the horizontally flipped webcam footage rather than the original footage. The flipped footage resembles a mirror and therefore provides a more natural interaction. This feature would give people with little or no prior experience in creative coding an intuitive way to mirror the webcam footage.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [X] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

### Feature request details

This feature would give users an option to mirror the webcam video. Some potential interfaces :
```javascript
function setup() {
  capture = createCapture(VIDEO);
  mirroredcCpture = mirrorCapture(capture);
}
```
or alternatively
```javascript
function setup() {
  mirroedCapture = createCapture(VIDEO, { flipped: true });
}
```

Unlike using functions like `scale(-1, 1)` which doesn't change the original video element, this feature would return a new video element. This would be easier to use than `scale`, which often results in things getting drawn to unexpected places and text being inverted. This feature would especially be helpful to the ml5 community since we could run prediction directly on the mirrored video element rather than trying to manually mirror both the video and the outputted data. 

Here is a link to the [discussion](https://github.com/ml5js/ml5-next-gen/issues/44) about this feature in the ml5 repo.

@shiffman Please feel free to add anything I missed :}

Vishal2002:
Resolves #6441

 Changes:

- Added a new feature to create a flipped video using `createCapture`.
- Modified the `createCapture` function to accept a boolean parameter for flipping the video.
 

 Screenshots of the change:
 
Screenshot 2024-01-06 162132
Screenshot 2024-01-05 234941

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -2135,7 +2135,11 @@ if (navigator.mediaDevices.getUserMedia === undefined) {
  * W3C documentation</a> for possible properties. Different browsers support different
  * properties.
  *
- * The second parameter, `callback`, is optional. It's a function to call once
+ * The 'flipped' property is an optional property which can be set to `{flipped:true}`
+ * to mirror the video output.If it is true then it means that video will be mirrored
+ * or flipped and if nothing is mentioned then by default it will be `false`.
+ *
+ * The second parameter,`callback`, is optional. It's a function to call once
  * the capture is ready for use. The callback function should have one
  * parameter, `stream`, that's a
  * <a href="https://developer.mozilla.org/en-US/docs/Web/API/MediaStream" target="_blank">MediaStream</a> object.
@@ -2148,6 +2152,8 @@ if (navigator.mediaDevices.getUserMedia === undefined) {
  * @param  {String|Constant|Object}  [type] type of capture, either AUDIO or VIDEO,
  *                                   or a constraints object. Both video and audio
  *                                   audio streams are captured by default.
+ * @param  {Object}                  [flipped] flip the capturing video and mirror the output with `{flipped:true}`. By
+ *                                   default it is false.
  * @param  {Function}                [callback] function to call once the stream
  *                                   has loaded.
  * @return {p5.Element} new <a href="#/p5.Element">p5.Element</a> object.
@@ -2181,6 +2187,19 @@ if (navigator.mediaDevices.getUserMedia === undefined) {
  *   // Invert the colors in the stream.
  *   filter(INVERT);
  * }
+ * </code>
+ * </div>
+ * <div class='notest'>
+ * <code>
+ * let capture;
+ *
+ * function setup() {
+ *   // Create the video capture with mirrored output.
+ *   capture = createCapture(VIDEO,{ flipped:true });
+ *   capture.size(100,100);
+ *   describe('A video stream from the webcam with flipped or mirrored output.');
+ * }
+ *
  * </code>
  * </div>
  *
@@ -2227,11 +2246,12 @@ p5.prototype.createCapture = function(...args) {
     if (arg === p5.prototype.VIDEO) useAudio = false;
     else if (arg === p5.prototype.AUDIO) useVideo = false;
     else if (typeof arg === 'object') {
-      constraints = arg;
-      flipped = constraints.flipped || false;
-    }
-    else if (typeof arg === 'boolean') {
-      flipped = arg;
+      // Check if the argument is an object with a 'flipped' property
+      if (arg.flipped !== undefined) {
+        flipped = arg.flipped;
+      } else {
+        constraints = arg;
+      }
     }
     else if (typeof arg === 'function') {
       callback = arg;
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -2275,7 +2275,7 @@ p5.prototype.createCapture = function(...args) {
 
     if (callback) callback(domElement.srcObject);
   });
-
+  videoEl.flipped=flipped;
   return videoEl;
 };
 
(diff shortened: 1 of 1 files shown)


src/dom/dom.js
```diff
@@ -2212,7 +2212,7 @@ if (navigator.mediaDevices.getUserMedia === undefined) {
 p5.prototype.createCapture = function(...args) {
```
davepagurek, line 2212: Can we update the doc comments above this to mention the `flipped` parameter, and maybe add an example that uses it?

src/dom/dom.js
```diff
@@ -2253,6 +2264,9 @@ p5.prototype.createCapture = function(...args) {
     if (domElement.width) {
       videoEl.width = domElement.width;
       videoEl.height = domElement.height;
+      if (flipped) {
+        videoEl.elt.style.transform = 'scaleX(-1)';
```
davepagurek, line 2268: If you call `yourCapture.hide()` and `image(yourCapture, 0, 0)` to draw it to the canvas, is it flipped there too? If not, we may need to update the part of the code where media elements draw to their internal canvas and draw it scaled there too (see this comment for details: https://github.com/processing/p5.js/issues/6441#issuecomment-1747681166)

src/dom/dom.js
```diff
@@ -2221,12 +2221,23 @@ p5.prototype.createCapture = function(...args) {
...
   for (const arg of args) {
     if (arg === p5.prototype.VIDEO) useAudio = false;
     else if (arg === p5.prototype.AUDIO) useVideo = false;
-    else if (typeof arg === 'object') constraints = arg;
-    else if (typeof arg === 'function') callback = arg;
+    else if (typeof arg === 'object') {
+      constraints = arg;
+      flipped = constraints.flipped || false;
+    }
+    else if (typeof arg === 'boolean') {
```
davepagurek, line 2233: I think it's more futureproof to force people to use the settings object (`createCapture(VIDEO, { flipped: true })`) than to add additional overloads with a boolean argument (`createCapture(VIDEO, true)`) in case we need to add any other additional boolean settings. How do you feel about taking out this branch?

src/dom/dom.js
```diff
@@ -2221,12 +2221,23 @@ p5.prototype.createCapture = function(...args) {
...
   let callback;
+  let flipped = false;
+
   for (const arg of args) {
     if (arg === p5.prototype.VIDEO) useAudio = false;
     else if (arg === p5.prototype.AUDIO) useVideo = false;
-    else if (typeof arg === 'object') constraints = arg;
-    else if (typeof arg === 'function') callback = arg;
+    else if (typeof arg === 'object') {
+      constraints = arg;
```
davepagurek, line 2230: This gets passed into [`getUserMedia`](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia) below. We might want to extract just the bits we need for `constraints` so we don't accidentally send extra info in in the future if we add anything to the settings object. Maybe we could do something like this?
```js
constraints = { video: arg.video, audio: arg.audio };
```

davepagurek (commented):
Thanks for taking this on! I've added a few comments, let me know what you think!

src/dom/dom.js
Vishal2002: I will soon update the inline documentation and give a simple good-to-go example.

src/dom/dom.js
Vishal2002: Ya that I missed

Vishal2002: @davepagurek I did it , now it is working with the capture.hide() and image(capture,0,0);

Screenshot 2024-01-10 232633
Screenshot 2024-01-10 232926

Vishal2002: This time i have changed the `_ensureCanvas()`  so that it works fine.

davepagurek: Nice, looks great!

src/dom/dom.js
```diff
@@ -2227,11 +2246,12 @@ p5.prototype.createCapture = function(...args) {
...
-      constraints = arg;
-      flipped = constraints.flipped || false;
-    }
-    else if (typeof arg === 'boolean') {
-      flipped = arg;
+      // Check if the argument is an object with a 'flipped' property
+      if (arg.flipped !== undefined) {
+        flipped = arg.flipped;
+      } else {
+        constraints = arg;
```
davepagurek, line 2253: Right now I think this means we either can provide `flipped`, or we can provide constraints on the capture size. I think we want to support both at once, but just with the `flipped` arg removed. Maybe something like:
```js
if (arg.flipped !== undefined) {
  flipped = arg.flipped;
}
constraints = { ...arg };
delete constraints.flipped;
```

davepagurek (approved):
I think this is good to go now! Since we've [started our release process with a beta build](https://github.com/processing/p5.js/releases/tag/v1.9.1-1), I'm going to hold off on merging this until after we finish the release so that we don't accidentally introduce new bugs during the testing period. I'll merge this right after that though!

src/dom/dom.js
```diff
@@ -2221,29 +2240,40 @@ p5.prototype.createCapture = function(...args) {
...
       } else {
         domElement.src = window.URL.createObjectURL(stream);
       }
-    } catch (err) {
+    }
+    catch(err) {
       domElement.src = stream;
     }
-  }, console.log);
+  }, console.err);
```
davepagurek, line 2276: Oh I just noticed, I don't think `console.err` is defined, I think this should be `console.error`:
```suggestion
  }, console.error);
```

PR #6710, opened 2024-01-07
.mtl files support for .obj files to render color in 3D custom geometry

### LINKED ISSUE #6670: The Support for colored STL and OBJ files is not present.
### Increasing Access

image
image

Went through the learn pages and found out that p5.js currently does not support colored STL files while creating custom geometry. 
Found the relevant code here:
```
/**
 * This function parses the Binary STL files.
 * https://en.wikipedia.org/wiki/STL_%28file_format%29#Binary_STL
 *
 * Currently there is no support for the colors provided in STL files.
 */
function parseBinarySTL(model, buffer) {
  const reader = new DataView(buffer);

  // Number of faces is present following the header
  const faces = reader.getUint32(80, true);
  let r,
    g,
    b,
    hasColors = false,
    colors;
  let defaultR, defaultG, defaultB;

  // Binary files contain 80-byte header, which is generally ignored.
  
```

Link to the code:
https://github.com/processing/p5.js/blob/8f2b47c2957fda064992ddffbceb62983a26103c/src/webgl/loading.js#L357C1-L375C1

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

### Feature request details

STL files can have color information in the form of per-face or per-vertex color attributes. We would need to modify the parser to extract and interpret this color data and then use it appropriately in the application.

Here's a high-level overview of the steps:

Parse Color Information:

For binary STL files, understand the format's structure to locate color information.
For ASCII STL files, adapt the parser to recognize and process color data in the file.
Integrate Color in Rendering:

Modify the rendering code to consider color information while rendering the 3D model.
Adjust shaders or rendering logic to incorporate color attributes.
Update File Upload and Display:

When handling file uploads, ensure that the code correctly interprets and processes color data from STL files.
Update the UI to display the 3D model with color.
Testing:

Test the implementation with STL files containing color information to verify that both geometry and color are rendered correctly.

I think this is where the functionality should be implemented:
  }
  if (hasColors) {
    // add support for colors here.
  }
  return model;
}

https://github.com/processing/p5.js/blob/8f2b47c2957fda064992ddffbceb62983a26103c/src/webgl/loading.js#L439-L445

diyaayay:
Resolves #6670 
This PR is a Work in Progress.
Review is requested.

 Changes:

 Screenshots of the change:
 The output after loading materials before the model is loaded and assigning diffuse colors to vertexColors
 The output contains parsedMaterials and model.vertexColors

image

image

#### PR Checklist

- [x] `npm run lint` passes


diyaayay:
@davepagurek The error above in the screenshots occur only when an obj file with an mtl file associated with it is loaded.
I think the `self._decrementPreload()` in this case is where I'm unable to catch the correct instance of ` this` to use and assign to `self ` as a lot of asynchronous calls were made implement loading of materials before model.

The sketch used is:
let octa;

function preload() {
   octa = loadModel('octa-color.obj');
}

function setup() {
}

function draw() {
  background(200);
  rotateX(frameCount * 0.01);
  rotateY(frameCount * 0.01);
  model(octa);
}

Requesting a review so that I can proceed further and also guidance in solving the current error.

diyaayay:
Sorry for the commit messages, I thought I had already committed some other changes earlier.

src/webgl/loading.js
```diff
@@ -178,6 +202,54 @@ p5.prototype.loadModel = function(path) {
   return model;
 };
 
+function parseMtl(mtlPath){ //accepts mtlPath to load file
+  return new Promise((resolve,reject)=>{
+    // console.log('parser is called');
+    let currentMaterial = null;
+    p5.prototype.loadStrings(
```
davepagurek, line 209: I think rather than calling this on `p5.prototype`, which will have no `this`, you'd want to call `loadStrings` on a p5 instance. That could mean calling `await parseMtl(this, mtlPath);` inside `loadModel` and then call `loadStrings` on the first argument, although it might be a nicer separation of concerns to separate loading the strings and the parsing, so it becomes:
```js
this.loadStrings(
  mtlPath,
  (mtlLines) => {
    const materials = parseMtl(mtlLines);
    parseObj(model, lines, materials)
  },
  () => {
    // handle failure
  }
)
```
Another benefit of this is that you maybe don't need `parseMtl` to resolve or reject, since the `loadStrings` part handles the case where it can't fetch the file, so `parseMtl` can look more like `parseObj`.

src/webgl/loading.js
```diff
@@ -103,6 +103,8 @@ import './p5.Geometry';
  * @param  {String} [fileType]
  * @return {p5.Geometry} the <a href="#/p5.Geometry">p5.Geometry</a> object
  */
+
+let materials = {}; //currently local
```
davepagurek, line 107: Do we use this btw? Looks like we always pass in materials as a parameter

src/webgl/loading.js
```diff
@@ -150,16 +152,38 @@ p5.prototype.loadModel = function(path) {
   } else if (fileType.match(/\.obj$/i)) {
     this.loadStrings(
       path,
-      strings => {
-        parseObj(model, strings);
-
-        if (normalize) {
-          model.normalize();
+      async lines => {
+        const mtlLine = lines.find(line => line.startsWith('mtllib '));
```
davepagurek, line 156: I think the obj spec supports multiple mtls, so we might need to `filter` to get all the lines that have `mtllib` and then parse all of them (maybe waiting for them all via `Promise.all(...)`?)

src/webgl/loading.js
```diff
@@ -178,6 +202,54 @@ p5.prototype.loadModel = function(path) {
...
+function parseMtl(mtlPath){ //accepts mtlPath to load file
+  return new Promise((resolve,reject)=>{
+    // console.log('parser is called');
+    let currentMaterial = null;
+    p5.prototype.loadStrings(
+      mtlPath,
+      lines => {
+        for (let line = 0; line < lines.length; ++line){
+          const tokens = lines[line].trim().split(/\s+/);
+          if(tokens[0] === 'newmtl') {
```
davepagurek, line 214: Nice work on this parser, looks great!

src/webgl/loading.js
```diff
@@ -213,7 +287,20 @@ function parseObj(model, lines) {
     const tokens = lines[line].trim().split(/\b\s+/);
 
     if (tokens.length > 0) {
-      if (tokens[0] === 'v' || tokens[0] === 'vn') {
+      if (tokens[0] === 'usemtl') {
+        // Switch to a new material
+        currentMaterial = tokens[1];
+        if (currentMaterial && materials[currentMaterial]) {
+          const diffuseColor = materials[currentMaterial].diffuseColor;
+          model.vertexColors.push([
```
davepagurek, line 295: I think we want to add to the diffuse colors *for each vertex*, where currently this is adding to the array once for each `usemtl` line. So maybe that means just setting the `currentMaterial` in this `if` branch, and in the branch below where we add to `loadedVerts`, add to `vertexColors` based on the material.

src/webgl/loading.js
```diff
@@ -213,7 +287,20 @@ function parseObj(model, lines) {
...
+            diffuseColor[0],
+            diffuseColor[1],
+            diffuseColor[2]
+          ]);
+        } else {
+          model.vertexColors.push([1, 1, 1]);
```
davepagurek, line 301: Just thinking this through: when a model has vertex colors, then `fill()` has no effect on the model, unless you call `.clearColors()` on it. Currently, all imported models have no colors, so they take on whatever fill is used. We want to keep that behaviour for objs that have no materials, so maybe that means if no materials are present in the obj, rather than pushing to `vertexColors`, we just leave it empty.

This would make for a good unit test btw: testing that if you import an obj that has materials, it uses the material colors when you draw it with `model()`, and when you import an obj that has no materials, it draws using whatever the fill color is.

test/unit/assets/octa-color.obj
```diff
@@ -0,0 +1,23 @@
+mtllib octa-color.mtl
```
davepagurek, line 1: The plant obj is 9mb, which might slow down tests. Maybe for the missing material test, we can make a duplicate of this colored obj, but change the path from `octa-color.mtl` to something that doesn't exist?

davepagurek (commented):
Great work so far! I've left a few comments, some about the error you were running into, and some about how we build up the model after that. Let me know what you think!

src/webgl/loading.js
diyaayay: Ah, yes I had to remove it , missed it though. I'll remove it.

test/unit/assets/octa-color.obj
diyaayay: yes, that's what I had thought of too, made the same changes already for `plant.obj` and will push them asap.

davepagurek:
Btw just a heads up, looks like this test is failing:
image

src/webgl/loading.js
```diff
@@ -150,18 +149,37 @@ p5.prototype.loadModel = function(path) {
...
+            // Replace the file name in the path with the MTL file name
+            //find a better way to resolve path
+            if(objFileDir){
+              mtlPath = objFileDir + '/' + tokens[1];
+            }
+            else{
+              mtlPath=tokens[1];
+            }
+            // Check if the MTL file has already been loaded
+            const existingMtl = mtlFiles.find(mtl => mtl.path === mtlPath);
```
davepagurek, line 174: I think this is a bit of a race condition, since we only push to the array after `await`ing an asynchronous function. Maybe you can synchronously push an object that just includes the path, and then after `await`ing, add the materials to it? something like:
```js
const file = { path: mtlPath };
mtlFiles.push(file);
const parsedMaterials = await parseMtl(self, mtlPath)
file.materials = parsedMaterials;
```

src/webgl/loading.js
```diff
@@ -202,52 +219,51 @@ p5.prototype.loadModel = function(path) {
...
-              parseFloat(tokens[1]),
-              parseFloat(tokens[2]),
-              parseFloat(tokens[3])
-            ];
-
-          }else if (tokens[0] === 'map_Kd') {
-            //Texture path
-            materials[currentMaterial].texturePath = tokens[1];
-          }
+async function parseMtl(self,mtlPath){ //accepts mtlPath to load file
```
davepagurek, line 222: Maybe we can call `self` `p5` instead so it's clearer what its type should be?

src/webgl/loading.js
```diff
@@ -202,52 +219,51 @@ p5.prototype.loadModel = function(path) {
...
+  let materials= {};
+  let currentMaterial = null;
+  self.loadStrings(
```
davepagurek, line 225: It looks like this function completes without waiting for the result of `loadStrings` now that it no longer returns a promise. So I think that means it will immediately return `{}`, and then asynchronously, properties will get added to it. I'm guessing it's important to have the complete materials object before continuing to parse .obj files later, so maybe it would make sense to add back in something like you had before?
```js
const materials = {};
const doneLoading = new Promise((resolve, reject) => {
  self.loadStrings(
    mtlPath,
    lines => {
      // do parsing and add to materials here
      resolve();
    },
    reject
  )
});
await doneLoading;
return materials;
```

davepagurek (commented):
Btw, we also recently merged in support for [visual tests](https://github.com/processing/p5.js/blob/main/contributor_docs/unit_testing.md#visual-tests). That might be a good fit here to check that vertex colors are visible on canvas when drawing.

src/webgl/loading.js
diyaayay: @davepagurek I changed my multiple mtl file support  functionality to this:
```
else if (fileType.match(/\.obj$/i)) {
    this.loadStrings(
      path,
      async lines => {
        const parsedMaterialPromises=[];
        try{
          for (let i = 0; i < lines.length; i++) {
            const mtllibMatch = lines[i].match(/^mtllib (.+)/);
            if (mtllibMatch) {
              let mtlPath='';
              const mtlFilename = mtllibMatch[1];
              const objPathParts = path.split('/');
              if(objPathParts.length > 1){
                objPathParts.pop();
                const objFolderPath = objPathParts.join('/');
                mtlPath = objFolderPath + '/' + mtlFilename;
              }else{
                mtlPath = mtlFilename;
              }
              const parsedMaterialsIndividual = await parseMtl(self, mtlPath);
              parsedMaterialPromises.push(parsedMaterialsIndividual);
            }
          }
          if(parsedMaterialPromises.length>0){
            const parsedMaterialsArray=
            await Promise.all(parsedMaterialPromises);
            const parsedMaterials = Object.assign({}, ...parsedMaterialsArray);
            await parseObj(model, lines, parsedMaterials);
          } }catch (error) {
          if (failureCallback) {
            failureCallback(error);
          } else {
            p5._friendlyError('Error during parsing: ' + error.message);
          }
        }
        finally{
          if (normalize) {
            model.normalize();
          }

          self._decrementPreload();
          if (typeof successCallback === 'function') {
            successCallback(model);
          }
        }
      },
      failureCallback
    );
  } else {
    p5._friendlyFileLoadError(3, path);
    if (failureCallback) {
      failureCallback();
    } else {
      p5._friendlyError(
        'Sorry, the file type is invalid. Only OBJ and STL files are supported.'
      );
    }
  }
  return model;
};
```

now single mtl files are loaded and then only passed to parseObj but multiple files still pass before fully loading all the materials thus the error: p5.js says: Error during parsing: Cannot read properties of undefined (reading 'diffuseColor')

In the earlier approach even after adding resolve, reject back and some other changes too, still the parseObj was called beforehand. 

where is the logic breaking in the current code that single mtl files are passed after materials fully loading but not multiple mtl files? I'm trying to still find since I used `Promise.all` also.

davepagurek: A few little things:
- Right now you do `const parsedMaterialsIndividual = await parseMtl(self, mtlPath);`. By `await`ing this, this means synchronously waiting for the material to be loaded/parsed, and the returned value is the actual result and not a promise. This works, but it means all mtl files are loaded one after another, and the `Promise.all` has no effect due to the results not being promises any more. If you remove the `await` on this line, the result is a promise, and the code only waits for them to finish at the `Promise.all`, meaning they all get loaded in parallel, which seems more like what you want.
- It sounds like right now, `await`ing `parseMtl` doesn't fully wait for it to be loaded. What does your `parseMtl` implementation look like right now? In the first comment in this chain, I had added some suggestions on how to update `parseMtl` to await for loading to complete, have you tried any of that out yet?

Let me know if I can help elaborate on any of those points!

diyaayay: @davepagurek below is the parseMtl function:
you had suggested using doneLoading and this is how I'm currently implementing the function:
```
function parseMtl(p5,mtlPath){ //accepts mtlPath to load file
  return new Promise((resolve, reject)=>{
    let currentMaterial = null;
    let materials= {};
    p5.loadStrings(
      mtlPath,
      lines => {
        for (let line = 0; line < lines.length; ++line){
          const tokens = lines[line].trim().split(/\s+/);
          if(tokens[0] === 'newmtl') {
            const materialName = tokens[1];
            currentMaterial = materialName;
            materials[currentMaterial] = {};
          }else if (tokens[0] === 'Kd'){
          //Diffuse color
            materials[currentMaterial].diffuseColor = [
              parseFloat(tokens[1]),
              parseFloat(tokens[2]),
              parseFloat(tokens[3])
            ];
          } else if (tokens[0] === 'Ka'){
          //Ambient Color
            materials[currentMaterial].ambientColor = [
              parseFloat(tokens[1]),
              parseFloat(tokens[2]),
              parseFloat(tokens[3])
            ];
          }else if (tokens[0] === 'Ks'){
          //Specular color
            materials[currentMaterial].specularColor = [
              parseFloat(tokens[1]),
              parseFloat(tokens[2]),
              parseFloat(tokens[3])
            ];

          }else if (tokens[0] === 'map_Kd') {
          //Texture path
            materials[currentMaterial].texturePath = tokens[1];
          }
        }
        resolve(materials);
      },reject
    );
  });
}
```

src/webgl/loading.js
```diff
@@ -312,8 +410,24 @@ function parseObj(model, lines) {
                 model.vertexNormals.push(loadedVerts.vn[vertParts[2]].copy());
               }
             }
+            if (usedVerts[vertIndex] && usedVerts[vertIndex]
```
davepagurek, line 413: Right now I think we just store the *last used material* for each vertex. I think this means that if we reuse the same vertex first with material1, then with material2, then a third time with material1 again, we'll end up with three duplicates of the vertex instead of just two, since we make a new duplicate if the current material is not the same as the last used material.

Maybe instead of storing a map of source index &rarr; material, we can store a map from source index &rarr; another map of material &rarr; destination index. In that world, this if statement would check if `usedVerts[vertIndex][currentMaterial]` is defined yet, and create it if not. (This would maybe have to also absorb the `else` branch right above it maybe.) Then, below, we would add `usedVerts[vertIndex][currentMaterial]` to the current face.

src/webgl/loading.js
```diff
@@ -331,7 +455,7 @@ function parseObj(model, lines) {
   if (model.vertexNormals.length === 0) {
     model.computeNormals();
   }
-
+  console.log(model.vertexColors);
```
davepagurek, line 458: I assume this is for debugging, although we should probably take this out before merging.

src/webgl/loading.js
```diff
@@ -322,6 +436,16 @@ function parseObj(model, lines) {
             face[1] !== face[2]
           ) {
             model.faces.push(face);
+            //same material for all vertices in a particular face
+            const materialDiffuseColor =
```
davepagurek, line 440: Do we need to add a check here if this exists, so that for objs without materials, we can leave the `vertexColors` array empty?

src/webgl/loading.js
```diff
@@ -322,6 +436,16 @@ function parseObj(model, lines) {
...
+             materials[currentMaterial].diffuseColor;
+            for (let i=0 ;i<face.length;i++) {
+              model.vertexColors.push([
```
davepagurek, line 443: I think elsewhere we treat `vertexColors` as a flat array: https://github.com/processing/p5.js/blob/80338bc715a9baa7d9b4f6fc3de54125a0c280b4/src/webgl/p5.RendererGL.Immediate.js#L117-L122

Maybe we can take out the `[...]` here to be consistent?

src/webgl/loading.js
diyaayay: Right, I missed out on how I could have optimized the logic here and thus have lesser duplication like in the example you mentioned above. Will work on it now. Thanks for pointing this out. @davepagurek

diyaayay:
@davepagurek Made a few changes in the duplication of vertices and the number of vertices in the `model.vertices ` is lesser , only copying the vertex a minimum number of times as per the requirement, I think?

src/webgl/loading.js
```diff
@@ -377,57 +378,38 @@ function parseObj(model, lines, materials= {}) {
...
-            if (usedVerts[vertIndex] && usedVerts[vertIndex]
-               !== currentMaterial) {
-              // Duplicate vertex, UV, and normal,faces to refer to new indices
-              vertParts = vertString.split('/');
-              for (let i = 0; i < vertParts.length; i++) {
-                vertParts[i] = parseInt(vertParts[i]) - 1;
-              }
-              const duplicatedVertIndex = model.vertices.length;
+
+            if (usedVerts[vertParts[0]][currentMaterial] === undefined) {
```
davepagurek, line 400: Nice!

src/webgl/loading.js
```diff
@@ -437,14 +419,14 @@ function parseObj(model, lines, materials= {}) {
...
             //same material for all vertices in a particular face
-            const materialDiffuseColor =
-             materials[currentMaterial].diffuseColor;
-            for (let i=0 ;i<face.length;i++) {
-              model.vertexColors.push([
-                materialDiffuseColor[0],
-                materialDiffuseColor[1],
-                materialDiffuseColor[2]
-              ]);
+            if (currentMaterial && materials[currentMaterial]) {
```
davepagurek, line 422: Right now this will break if we have some vertices that don't have a material but others that do (since it means the earlier vertices won't have pushed colors to the array.) Our data needs it to be all or nothing.

I don't think we need to actually support that case, but maybe we can try to detect it and throw an error if we encounter it? i.e. if we've added any vertices with no colors, and then find ourselves trying to add a vertex *with* color later on, throw an error letting people know they need colors for all their vertices or none.

test/unit/visual/cases/webgl.js
```diff
@@ -34,4 +34,20 @@ visualSuite('WebGL', function() {
       screenshot();
     });
   });
-});
+
+  visualSuite('3DModel', function() {
+    visualTest('OBJ model with MTL file displays diffuse colors correctly', function(p5, screenshot) {
+      return new Promise(resolve => {
+        p5.createCanvas(50, 50, p5.WEBGL);
+        p5.loadModel('unit/assets/octa-color.obj', model => {
```
davepagurek, line 42: Right now the screenshot looks like just a dot. Maybe to make this clearer we can call `model.normalize()` to normalize its size to, I think, 200px?

davepagurek (commented):
Nice work on this! I think we're getting really close.

test/unit/visual/cases/webgl.js
davepagurek: Maybe we can also add one more test importing an object with no colors, and then drawing it with a colored `fill()` to show that it can take on new colors.

src/webgl/loading.js
diyaayay: Thinking about it, there can be multiple ways of ensuring colors or no colors for all vertices, one being the first vertex added would dictate if all the further vertices to be added must have a color or not. i.e. if the first vertex has a color, all the vertices of the model must be colored and vice versa thus, all or nothing for colors of vertices.
Using a flag `colorRequirementSet` for this. Haven't thoroughly checked though if this too would break.

diyaayay:
```
function parseObj(model, lines, materials= {}) {
  // OBJ allows a face to specify an index for a vertex (in the above example),
  // but it also allows you to specify a custom combination of vertex, UV
  // coordinate, and vertex normal. So, "3/4/3" would mean, "use vertex 3 with
  // UV coordinate 4 and vertex normal 3". In WebGL, every vertex with different
  // parameters must be a different vertex, so loadedVerts is used to
  // temporarily store the parsed vertices, normals, etc., and indexedVerts is
  // used to map a specific combination (keyed on, for example, the string
  // "3/4/3"), to the actual index of the newly created vertex in the final
  // object.
  const loadedVerts = {
    v: [],
    vt: [],
    vn: []
  };


  // Map from source index → Map of material → destination index
  const usedVerts = {}; // Track colored vertices
  let currentMaterial = null;
  let uniqueVertices = 0; // Set to track unique vertices
  let coloredUniqueVertexCount = 0;
  let hasColoredVertices = false;
  for (let line = 0; line < lines.length; ++line) {
    // Each line is a separate object (vertex, face, vertex normal, etc)
    // For each line, split it into tokens on whitespace. The first token
    // describes the type.
    const tokens = lines[line].trim().split(/\b\s+/);

    if (tokens.length > 0) {
      if (tokens[0] === 'usemtl') {
        // Switch to a new material
        currentMaterial = tokens[1];
      }else if (tokens[0] === 'v' || tokens[0] === 'vn') {
        // Check if this line describes a vertex or vertex normal.
        // It will have three numeric parameters.
        const vertex = new p5.Vector(
          parseFloat(tokens[1]),
          parseFloat(tokens[2]),
          parseFloat(tokens[3])
        );
        if (tokens[0]==='v'){
          uniqueVertices++;
        }
        loadedVerts[tokens[0]].push(vertex);
      } else if (tokens[0] === 'vt') {
        // Check if this line describes a texture coordinate.
        // It will have two numeric parameters U and V (W is omitted).
        // Because of WebGL texture coordinates rendering behaviour, the V
        // coordinate is inversed.
        const texVertex = [parseFloat(tokens[1]), 1 - parseFloat(tokens[2])];
        loadedVerts[tokens[0]].push(texVertex);
      } else if (tokens[0] === 'f') {
        // Check if this line describes a face.
        // OBJ faces can have more than three points. Triangulate points.
        for (let tri = 3; tri < tokens.length; ++tri) {
          const face = [];
          const vertexTokens = [1, tri - 1, tri];

          for (let tokenInd = 0; tokenInd < vertexTokens.length; ++tokenInd) {
            // Now, convert the given token into an index
            const vertString = tokens[vertexTokens[tokenInd]];
            let vertParts=vertString.split('/');

            // TODO: Faces can technically use negative numbers to refer to the
            // previous nth vertex. I haven't seen this used in practice, but
            // it might be good to implement this in the future.

            for (let i = 0; i < vertParts.length; i++) {
              vertParts[i] = parseInt(vertParts[i]) - 1;
            }

            if (!usedVerts[vertParts[0]]) {
              usedVerts[vertParts[0]] = {};
            }

            if (usedVerts[vertParts[0]][currentMaterial] === undefined) {
              const vertIndex = model.vertices.length;
              model.vertices.push(loadedVerts.v[vertParts[0]].copy());
              model.uvs.push(loadedVerts.vt[vertParts[1]] ?
                loadedVerts.vt[vertParts[1]].slice() : [0, 0]);
              model.vertexNormals.push(loadedVerts.vn[vertParts[2]] ?
                loadedVerts.vn[vertParts[2]].copy() : new p5.Vector());

              usedVerts[vertParts[0]][currentMaterial] = vertIndex;
              face.push(vertIndex);
              //every unique vertex would have its current material undefined once
              coloredUniqueVertexCount++;
            } else {
              face.push(usedVerts[vertParts[0]][currentMaterial]);
            }
          }

          if (
            face[0] !== face[1] &&
            face[0] !== face[2] &&
            face[1] !== face[2]
          ) {
            model.faces.push(face);
            //same material for all vertices in a particular face
            if (currentMaterial
              && materials[currentMaterial]
              && materials[currentMaterial].diffuseColor) {
              const materialDiffuseColor =
              materials[currentMaterial].diffuseColor;
              hasColoredVertices = true;
              for (let i = 0; i < face.length; i++) {
                model.vertexColors.push(materialDiffuseColor[0]);
                model.vertexColors.push(materialDiffuseColor[1]);
                model.vertexColors.push(materialDiffuseColor[2]);
              }
            }
          }
        }
      }
    }
  }
  // If the model doesn't have normals, compute the normals
  if (model.vertexNormals.length === 0) {
    model.computeNormals();
  }
  if (hasColoredVertices) {
    // If any vertex has been colored, check if all have been
    if (coloredUniqueVertexCount !== uniqueVertices) {
      throw new Error('All vertices must be colored if any is colored, but found a mix.');
    }
  }
  return model;
}
```

@davepagurek Would something like this make sense for an all-or-nothing, i.e. a fully colored model or a colorless model. Assuming every unique vertex should have a color or no color for all vertices at all?

diyaayay:
@davepagurek Let me know what you think, what I've tried to do is if there are 6 vertices at 6 coordinate positions and all of them have color at least once, the model is valid. If none of them has color, the model is valid. Else, throw an error for consistency.

src/webgl/loading.js
```diff
@@ -419,9 +436,13 @@ function parseObj(model, lines, materials= {}) {
...
             model.faces.push(face);
             //same material for all vertices in a particular face
-            if (currentMaterial && materials[currentMaterial]) {
+            if (currentMaterial
+              && materials[currentMaterial]
+              && materials[currentMaterial].diffuseColor) {
               const materialDiffuseColor =
               materials[currentMaterial].diffuseColor;
+              //flag to track color or no color model
+              hasColoredVertices = true;
```
davepagurek, line 445: I wonder if we can simplify this a bit to not need the set. If you add a new variable called `hasColorlessVertices` and set it to true in an `else` branch, could we check that only one or the other is true but not both? i.e. `hasColoredVertices !== hasColorlessVertices`. I think we just want to make sure that we either always add colors, or never do, so this would be making sure we always go through the same branch of the if statement here.

src/webgl/loading.js
```diff
@@ -437,6 +458,19 @@ function parseObj(model, lines, materials= {}) {
   if (model.vertexNormals.length === 0) {
     model.computeNormals();
   }
+  if (hasColoredVertices) {
+  // Ensure every vertex added has a color
+    if (coloredVerts.size !== uniqueVertices) {
+      throw new Error('Not all vertices have a color.');
```
davepagurek, line 464: It would be great to add one more unit test with some invalid input to make sure it hits this error, then I think we're probably good to go!

davepagurek:
Great work on this so far! I think we're almost done -- I left some comments about a possible simplification, and a test for the error case. The last thing I can think of is, how should we handle it if e.g. you export an obj + mtl from Blender but only upload the obj? Right now we throw an error, but maybe it would be better to log a warning without actually throwing an error, and then treat it like it just has no materials? (Would that just involve parsing the `obj` in the `catch` block with no materials instead of rejecting the promise or would it be more complicated?)

src/webgl/loading.js
diyaayay: @davepagurek Yes, that would serve the purpose too. But this would not allow any duplicating vertex also not to have color, meaning if we have diffuse color even once , every vertex should have a diffuse color but what if vertex 1 is assigned color once but is repeated in another face, and should we allow it not having color later on further occurance? How I address it here in a set is all unique vertices of a model if they have color even once , allow them not to have assigned a color later. Because the model would still be fully colored. let me know which way are we trying to achieve a colored model?

diyaayay:
@davepagurek I think an obj file referencing mtl file but only obj uploaded can still be loaded and displayed on the canvas along with a warning. I'll look into it.

diyaayay:
> Great work on this so far! I think we're almost done -- I left some comments about a possible simplification, and a test for the error case. The last thing I can think of is, how should we handle it if e.g. you export an obj + mtl from Blender but only upload the obj? Right now we throw an error, but maybe it would be better to log a warning without actually throwing an error, and then treat it like it just has no materials? (Would that just involve parsing the `obj` in the `catch` block with no materials instead of rejecting the promise or would it be more complicated?)
@davepagurek  I tried displaying the obj model without colors if the associated mtl file is missing but it's more complicated than just logging the warning and then treat it like it has no materials as it would require multiple error handling and not just one. I tried but I get warnings and multiple errors still. I think p5.loadStrings would have to be updated to handle mtl file and return empty object {} and not an error when .mtl file 404 not found.Let me know what you think.

diyaayay:
@davepagurek Have a look when you find the time. Let me know what else should be done, Thanks.

src/webgl/loading.js
```diff
@@ -229,18 +229,16 @@ p5.prototype.loadModel = function(path,options) {
...
           if(parsedMaterials && typeof parsedMaterials==='object'){
-            try {
-              parseObj(model, lines, parsedMaterials);
-            } catch (error) {
-              console.error(error.message);
-            }
+            parseObj(model, lines, parsedMaterials);
           }
         }catch (error) {
+          console.error(error.message);
```
davepagurek, line 249: Do we need to `console.error` here when we `_friendlyError` below?

src/webgl/loading.js
```diff
@@ -158,6 +158,57 @@ p5.prototype.loadModel = function(path,options) {
...
+          if(objPathParts.length > 1){
+            objPathParts.pop();
+            const objFolderPath = objPathParts.join('/');
+            mtlPath = objFolderPath + '/' + mtlFilename;
+          }else{
+            mtlPath = mtlFilename;
+          }
+          try {
+            if(await fileExists(mtlPath)){
+              const parsedMaterialsIndividual = await parseMtl(self, mtlPath);
```
davepagurek, line 180: By adding `await` here, we're waiting for each material to load one at a time rather than waiting for all of them to load in parallel at the `Promise.all` below on line 195. Does this still work if we take out the `await` on this line, given that we already wait for all of them later?

src/webgl/loading.js
```diff
@@ -158,6 +158,57 @@ p5.prototype.loadModel = function(path,options) {
...
+          const objPathParts = path.split('/');
+          if(objPathParts.length > 1){
+            objPathParts.pop();
+            const objFolderPath = objPathParts.join('/');
+            mtlPath = objFolderPath + '/' + mtlFilename;
+          }else{
+            mtlPath = mtlFilename;
+          }
+          try {
+            if(await fileExists(mtlPath)){
```
davepagurek, line 179: This `await` will also cause us to do these `HEAD` requests one at a time instead of parallel. Maybe we can do something like:

```js
parsedMaterialPromises.push(
  fileExists(mtlPath).then(exists => {
    if (exists) {
      return parseMtl(self, mtlPath);
    } else {
      console.warn('MTL file not found or error in parsing; proceeding without materials.');
      return {};
    }
  }).catch(error => {
    console.warn('Error loading MTL file:', error);
    return {};
  })
);
```
(Also, while we're at it, maybe we can add the URL for the mtl file that we couldn't find to the warning we log?)

src/webgl/loading.js
```diff
@@ -158,6 +158,57 @@ p5.prototype.loadModel = function(path,options) {
...
+        resolve (materials);
+      } catch (error) {
+        resolve({});
+      }
+
+    });
+  }
+  async function fileExists(url) {
+    try {
+      const response = await fetch(url, { method: 'HEAD' });
```
davepagurek, line 206: Nice!

src/webgl/loading.js
diyaayay: Ah yes! I thought of it but was only checking all errors in console while making this work. Sorry for the oversight while pushing. I'll change these to Friendly Errors.

src/webgl/loading.js
diyaayay: Removing `await` here would result in parallel loading of all mtl filled and then `await` with `promise.all` which is more efficient in handling all files together , better than sequential parsing of each mtl file.

diyaayay:
@davepagurek Does this seem better? Thanks for the suggestions.

davepagurek (approved):
Thanks for your hard work on this and all the iterations you've made! I think this is good to go now!

Right now we're in the middle of releasing v1.9.1, and [have a beta build that we're bug testing](https://github.com/processing/p5.js/releases/tag/v1.9.1-1), so we're just merging in bug fixes until we finish the release. I've approved this PR though, so as soon as we finish the release, I'll merge this in!

diyaayay:
@davepagurek Had a great time working on this, as a university student this PR improved my vanilla javascript concepts a lot. Thank you for putting in all the time and reviewing it so many times✨

Qianqianye:
Looks great! Thanks @diyaayay and @davepagurek!

PR #6718, opened 2024-01-09
Remove unused grunt task.

IgnorantSapient:
When doing an `npm install` it always complained that `grunt-contrib-compress` wasn't installed, upon further inspection, it seems that `grunt-contrib-compress` is never used, so it is best if we remove this stale dependency.

#### PR Checklist

- [x] `npm run lint` passes


limzykenneth:
Looks good. Thanks!

PR #6731, opened 2024-01-12
fixed-Incorrect lines when applying a stroke to geometry built without initial stroke-#6318

Artimus100:
completely fixes the bug of issue https://github.com/processing/p5.js/issues/6318

Changes:
renderer._doStroke always true in  the constructor of GeometryBuilder Class within GeometryBuilder.js

Screenshot:

Screenshot 2024-01-12 at 3 38 30 PM

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/GeometryBuilder.js
+++ b/src/webgl/GeometryBuilder.js
@@ -9,6 +9,7 @@ import * as constants from '../core/constants';
 class GeometryBuilder {
   constructor(renderer) {
     this.renderer = renderer;
+    renderer._doStroke = true;
     renderer._pInst.push();
     this.identityMatrix = new p5.Matrix();
     renderer.uMVMatrix = new p5.Matrix();


Artimus100:
@davepagurek please review it!!

davepagurek:
Thanks for taking this on!

Have you tried doing any testing on how long it takes to before/after this change when creating geometry with lots of edges but with `noStroke()`? I think that's one of the risks of introducing this, where if the time difference is significant, we might need to also add a setting to ignore strokes for performance reasons.

Also, it would be great to add a test for this so we can be sure this doesn't break in the future. We also have just merged [visual tests](https://github.com/processing/p5.js/blob/main/contributor_docs/unit_testing.md#visual-tests) which might be a good fit here.

Artimus100:
@davepagurek ill get back to you with that data!!

Artimus100:
Hello @davepagurek should i mail you a doc with the time difference or just comment here so that you can look at it and tell me if we need to add the setting to ignore strokes for the performance reasons and ill start working on the tests right away!!

davepagurek:
I think commenting here would be great so that anyone else following the discussion can see the data too. We're also talking about adding a settings object as a parameter to `buildGeometry`/`beginGeoemtry` in https://github.com/processing/p5.js/issues/6722, so maybe if it turns out we need it, we can add a setting like `buildGeometry({ includeStroke: false })`.

src/webgl/GeometryBuilder.js
```diff
@@ -9,6 +9,7 @@ import * as constants from '../core/constants';
 class GeometryBuilder {
   constructor(renderer) {
     this.renderer = renderer;
+    renderer._doStroke = true;
```
davepagurek, line 12: Btw does this reset back to its previous value after the end of `buildGeometry`? Do we have to put it after the `renderer._pInst.push()` so that it gets automatically reset again after?

Artimus100: yup sorry for that!! i'll change it right away!

perminder-17:
Hi @Artimus100, are you still working on this PR? If you're busy, would it be okay if I or someone else continue the work? Let me know.

davepagurek:
oh after https://github.com/processing/p5.js/pull/7287 got merged, I wonder if that also happened to resolve the same issue this was addressing? Probably worth a check to see

Artimus100:
hey @perminder-17 sorry for taking this long,  ill get back to this asap!!

perminder-17:
Before (with no stroke) : 

Screenshot from 2024-10-02 23-40-10

After the PR #7287 (with no stroke) 

Screenshot from 2024-10-02 23-41-37

Sorry for missing this earlier. We no longer create edges for models that don't already have them, meaning if no strokes are made, we don't create edges. 

This was done previously by this block in model method. Now we have removed it which fixes this issue I guess. 

```
      if (model.edges.length === 0) {
        model._makeTriangleEdges();
      }

``` 

Thanks @davepagurek and also sorry for the oversight.

Artimus100:
hey @perminder-17  i think it will be better if you take on to this one 

-Thank you.

perminder-17:
> hey @perminder-17 i think it will be better if you take on to this one
> 
> -Thank you.

aah..No worries @Artimus100, I meant the issue is already been resolved by the PR @davepagurek mentioned above in the comment. But I really appropriate your efforts in this PR. I need to close this issue since it's already been resolved, so thanks so much for your efforts on this PR. :)

PR #6734, opened 2024-01-12
 Solves issue #4214

Garima3110:
Changes:
 Solves issue #4214 
 
 I have tried implementing the linePerspective() method to solve this issue.
 @davepagurek Please let me know is this the right approach for its implementation or not.
 I would love to have suggestions to this.
 Thanks a lot!

Sample sketch:
https://editor.p5js.org/Garima3110/sketches/p8jRwB5dw

#### PR Checklist

- [x] `npm run lint` passes


perminder-17:
Thank you for your efforts, Garima. I apologize if I misunderstood your approach, but I have a concern about the `linePerspective()` function. It seems that the issue with the misbehavior of strokeWeight occurs after the `perspective()` function is called, right? In your sketch, it seems to happen by default when we remove the `linePerspective(true/false)` method. Additionally, when we use `linePerspective(true/false)`, setting it to true seems to work well in your sketch. However, if someone sets `linePerspective` to false, it should use the default perspective camera. Perhaps we could implement something like this? Currently, when we use false, the stroke misbehaves in both cases, with and without the perspective call. 

### Actual 
when `linePerspective(false)` with and without perspective
stroke actual

### Maybe it should be like 

stroke expected



I think opting for this approach could be the most straightforward, as opposed to delving into intricate calculations for a custom perspective. However, I'm not entirely certain if this is the correct solution. I apologize if my approach doesn't prove effective.

Garima3110:
> I think opting for this approach could be the most straightforward, as opposed to delving into intricate calculations for a custom perspective. However, I'm not entirely certain if this is the correct solution. I apologize if my approach doesn't prove effective.

Thanks a lot @perminder-17 for your inputs. I'll just look into what's going wrong in here and get back to you soon!

src/webgl/p5.Camera.js
```diff
@@ -193,6 +193,23 @@ p5.prototype.perspective = function (...args) {
...
+ * When linePerspective is disabled, lines will have a uniform scale regardless of the camera's perspective.
+ *
+ * @method linePerspective
+ * @memberof p5.prototype
+ * @param {boolean} enable - Set to `true` to enable line perspective, `false` to disable.
+ *<br>
+ * @example
+ * @todo
+ */
+p5.prototype.linePerspective = function (enable) {
```
davepagurek, line 209: Can we add a second overload to this function that has no parameters and returns the current value? Similar to `frameRate`

src/webgl/p5.Shader.js
```diff
@@ -329,7 +329,7 @@ p5.Shader = class {
         this.setUniform('uPerspective', 1);
       } else {
         // strokes have uniform scale regardless of distance from camera
-        this.setUniform('uPerspective', 0);
+        this.setUniform('uPerspective', this._renderer._curCamera.useLinePerspective ? 1 : 0);
```
davepagurek, line 332: Nice!

src/webgl/p5.Camera.js
```diff
@@ -488,7 +505,7 @@ p5.Camera = class Camera {
     this._renderer = renderer;
 
     this.cameraType = 'default';
-
+    this.useLinePerspective = true;
```
davepagurek, line 508: Do you think calling `perspective()` should set this to true and calling `ortho()` or `frustum()` should set this to false? A user could then override it with `linePerspective()`, but still match the defaults we have now.

src/webgl/p5.Camera.js
Garima3110: @davepagurek do you want me to validate its parameters too or this would work fine?
```
p5.prototype.linePerspective = function (enable) {
  if (enable !== undefined) {
    // Set the line perspective if enable is provided
    this._renderer._curCamera.useLinePerspective = enable;
  } else {
    // If no argument is provided, return the current value
    return this._renderer._curCamera.useLinePerspective;
  }
};
```

src/webgl/p5.Camera.js
Garima3110: Yes maybe, I think  it makes sense to associate the behavior of linePerspective with the type of camera projection being used. If you're using `perspective()`, which simulates realistic perspective, having linePerspective set to true by default aligns well with that mode. Similarly, when using `ortho()` or `frustum()`, which do not simulate realistic perspective, setting linePerspective to false by default is appropriate.
This way, the default behavior aligns with the common expectations of how lines should behave based on the chosen camera projection. 
let me know what's your take on this.
Thanks!

davepagurek: I think that makes sense! we can mention in the documentation for `linePerspective` that `perspective()`, `ortho()`, and `frustum()` all change the value of `linePerspective`, so you can call it after those calls to override it.

src/webgl/p5.Camera.js
davepagurek: Validating parameters would be great too! Looks good so far.

Garima3110:
@davepagurek Sorry to directly ping you here, but its just to tell you that I had made the changes suggested by you. Please review them whenever you find time . No hurries at all !
Thanks a lot..!

src/webgl/p5.Camera.js
```diff
@@ -193,23 +193,81 @@ p5.prototype.perspective = function (...args) {
...
+ *  is set to `true` by default. This means that lines will be affected by the current
+ *  camera's perspective, resulting in a more natural appearance.
+ *
+ *  When using `ortho()` or `frustum()`, which do not simulate realistic perspective,
+ *  linePerspective is set to `false` by default. In this case, lines will have a uniform
+ *  scale regardless of the camera's perspective, providing a more predictable and
+ *  consistent appearance.
+ *
+ *  You can override the default behavior by explicitly calling `linePerspective()` after
+ *  using perspective(), ortho(), or frustum(). This allows you to customize the line
```
davepagurek: Maybe we can do these as inline code too for consistency:
```suggestion
 *  using `perspective()`, `ortho()`, or `frustum()`. This allows you to customize the line
```

src/webgl/p5.Camera.js
```diff
@@ -193,23 +193,81 @@ p5.prototype.perspective = function (...args) {
...
+ *  perspective based on your specific requirements.
  *
  * @method linePerspective
  * @memberof p5.prototype
  * @param {boolean} enable - Set to `true` to enable line perspective, `false` to disable.
+ * @return {boolean} The boolean value representing the current state of linePerspective().
```
davepagurek, line 216: For other p5 methods that could either be a setter or a getter, we just do one of them in the doc comment (e.g. here, having a `@param` and no `@return`) and then adding a second doc comment right after the first with the alternate overload (e.g. a `@return` with no `@param`.) An example of that is here: https://github.com/processing/p5.js/blob/37d3324b457b177319ed65468201f2806a66eff5/src/core/environment.js#L368-L371

Can we do the same thing here too?

src/webgl/p5.Camera.js
```diff
@@ -193,23 +193,81 @@ p5.prototype.perspective = function (...args) {
...
+ *
+ * @alt
+ * Demonstrates the dynamic control of line perspective in a 3D environment with rotating boxes.
  */
+
 p5.prototype.linePerspective = function (enable) {
-  this._renderer._curCamera.useLinePerspective = enable;
+  p5._validateParameters('linePerspective', arguments);
+
+  if (!this._renderer._curCamera) {
```
davepagurek, line 254: Did you encounter something in testing where where this wasn't defined out of curiosity?

davepagurek (commented):
Nice work! Just some minor comments on the docs formatting to be more consistent.

src/webgl/p5.Camera.js
Garima3110: Yeah, actually I got this on the console initially when i didn't write this if block : 
image
For the sketch: sketch1.js
Sketch link: https://editor.p5js.org/Garima3110/sketches/p8jRwB5dw

Garima3110:
@davepagurek I have made the changes as suggested. Please have a look , and let me know of any other change. Thank you!

src/webgl/p5.Shader.js
```diff
@@ -329,7 +329,7 @@ p5.Shader = class {
         this.setUniform('uPerspective', 1);
```
davepagurek, line 329: I was testing out the example and realized that we still force perspective when using a default camera. Maybe we can take out this if statement and just use the else branch?

src/webgl/p5.Camera.js
davepagurek: I believe that's just happening because your setup is:
```js
function setup() {
  createCanvas(400, 400);
  stroke(255);
  strokeWeight(2);
}
```
Doing `createCanvas(400,400,WEBGL)` causes the error to go away. Maybe rather than creating a camera in this if statement, we can throw an error saying that the method needs to be called in WebGL mode?

src/webgl/p5.Camera.js
```diff
@@ -193,6 +193,118 @@ p5.prototype.perspective = function (...args) {
...
+ * - When using `ortho()` or `frustum()`, which do not simulate realistic perspective,
+ *    linePerspective is set to `false` by default. In this case, lines will have a uniform
+ *    scale regardless of the camera's perspective, providing a more predictable and
+ *    consistent appearance.
+ * - You can override the default behavior by explicitly calling `linePerspective()` after
+ *    using `perspective()`, `ortho()`, or `frustum()`. This allows you to customize the line
+ *    perspective based on your specific requirements.
+ *
+ * @method linePerspective
+ * @memberof p5.prototype
```
davepagurek, line 214: I think elsewhere we tend to use he following syntax to say that it's a global p5 method:
```suggestion
 * @for p5
```

src/webgl/p5.Camera.js
```diff
@@ -193,6 +193,118 @@ p5.prototype.perspective = function (...args) {
...
+ *   rotateZ(PI/8);
+ *   translate(0, 0, 350);
+ *   for (let i = 0; i < 12; i++) {
+ *     translate(0, 0, -70);
+ *     box(30);
+ *   }
+ * }
+ *
+ * function mousePressed() {
+ *   perspective(PI/12, width/height, 1, 10000);
```
davepagurek, line 240: Since this isn't the default camera perspective any more, when you click, both the perspective in the lines and for the overall image changes. Maybe it would show the difference more clearly if we *just* call `linePerspective(false)` without also changing the perspective of the scene?

Garima3110:
Thanks @davepagurek .I’ll work on these suggestions  and get back to you soon!

Garima3110:
@davepagurek , Here's the sketch with the latest build including the changes you suggested.
Please have a look. Thankyou!

Garima3110:
>Doing createCanvas(400,400,WEBGL) causes the error to go away. Maybe rather than creating a camera in this if statement, we can throw an error saying that the method needs to be called in WebGL mode?

I am not exactly sure how to go about it , as I think directly writing this won't do the job:
```
  if (!this._renderer===constants.WEBGL) {
    throw new Error('linePerspective() must be called in WebGL mode.');
  }
```
It would be great if you could guide me through the process whenever you find time.
Thanks a lot..!

davepagurek:
Something pretty close to that! I think the way you would check would be:
```js
if (!(this._renderer instanceof p5.RendererGL)) {
  throw new Error('linePerspective() must be called in WebGL mode.');
}
```

Garima3110:
> ```js
> (this._renderer instanceof p5.RendererGL)
> ```

Great! It works as expected. Thanks @davepagurek 
And let me know for any further changes to this.

davepagurek:
Some last thoughts about the examples: right now it's a bit unclear what the difference is. How do you feel about:
- Changing the rotation a bit (I tried `rotateY(PI/24)`) so you can see the cubes getting smaller into the distance more clearly in the small canvas size
- Adding `setAttributes({ antialias: true })` after creating the canvas to get a bit better line quality
- Maybe making mouse click toggle back and forth rather than just turning off perspective? e.g. `linePerspective(!linePerspective())`


https://github.com/processing/p5.js/assets/5315059/9e47198f-c8f1-44b2-a4b4-88f31f15386e

Garima3110:
> * Changing the rotation a bit (I tried `rotateY(PI/24)`) so you can see the cubes getting smaller into the distance more clearly in the small canvas size
> * Adding `setAttributes({ antialias: true })` after creating the canvas to get a bit better line quality
> * Maybe making mouse click toggle back and forth rather than just turning off perspective? e.g. `linePerspective(!linePerspective())`

Great suggestions @davepagurek I have made the changes please have a look!

davepagurek (approved):
Awesome, thanks for all the changes! The examples look great, I think this is good to go.

PR #6761, opened 2024-01-21
Separate Model and View Matrices.

### LINKED ISSUE #5287: Separate the Model and View Matrices
- [x] WebGL

Currently in the source code, the model and view matrices are combined into one (`uMVMatrix`). Typically, these two matrices are kept distinct from each other. The model matrix used only for translating / rotating / scaling the model. The view matrix used only for translating / rotating the camera. 

One consequence of this, is that the model matrix is not available for use in shader programs. Currently, the view matrix and modelView matrix are passed to shaders as [follows][0]:
```js
p5.Shader.prototype._setMatrixUniforms = function() {
  ...
  this.setUniform('uViewMatrix', this._renderer._curCamera.cameraMatrix.mat4);
  this.setUniform('uModelViewMatrix', this._renderer.uMVMatrix.mat4);
  ...
};
```

If the two are kept separate in the code, then providing the model matrix to shader programs becomes trivial.
```js
p5.Shader.prototype._setMatrixUniforms = function() {
  ...
  const modelMatrix = this._renderer.modelMatrix;  // does not currently exist
  const viewMatrix = this._renderer._curCamera.cameraMatrix;
  const modelViewMatrix = (modelMatrix.copy()).mult(viewMatrix);

  this.setUniform('uModelMatrix', modelMatrix.mat4);
  this.setUniform('uViewMatrix', viewMatrix.mat4);
  this.setUniform('uModelViewMatrix', modelViewMatrix.mat4);
  ...
};
```

The separation of concerns will also make maintaining and/or expanding functionality in the source code simpler.

Exposing the model matrix as a uniform available for users to use in their shader programs can make p5 a better platform for people looking to learn about 3D graphics and shaders. It will make it possible to use p5 to follow along with the wider ecosystem of tutorials, books, and other educational material on computer graphics.

\---

This issue is an extension of the conversation [here][1].

[0]: https://github.com/processing/p5.js/blob/e7f2c143ae283c7983a890e4a2b914203f89c562/src/webgl/p5.Shader.js#L263
[1]: https://github.com/processing/p5.js-website/issues/1017#issuecomment-851041727

deveshidwivedi:
Resolves #5287

 Changes:

Introduced a distinct ```modelMatrix```. By separating these matrices (```modelViewMatrix```), it would become easier to manipulate each aspect of the transformation independently, allowing for more flexibility especially when working with shaders.

#### PR Checklist

- [ ] `npm run lint` passes


src/webgl/p5.Camera.js
```diff
@@ -1600,13 +1600,12 @@ p5.Camera = class Camera {
     this.cameraMatrix = cam.cameraMatrix.copy();
     this.projMatrix = cam.projMatrix.copy();
 
-    // If the target camera is active, update uMVMatrix and uPMatrix.
     if (this._isActive()) {
-      this._renderer.uMVMatrix.mat4 = this.cameraMatrix.mat4.slice();
+      this._renderer.uModelMatrix.mat4 = this.modelMatrix.mat4.slice();
```
lindapaiste, line 1604: This was something that I worked on in https://github.com/processing/p5.js/pull/6633#issuecomment-1859290684. We can do `this._renderer.uModelMatrix.set(this.modelMatrix);` instead of messing around with the "internals" of the matrix.

src/webgl/p5.RendererGL.Retained.js
```diff
@@ -189,12 +189,13 @@ p5.RendererGL.prototype.drawBuffersScaled = function(
   scaleY,
   scaleZ
 ) {
-  const uMVMatrix = this.uMVMatrix.copy();
+  const uModelMatrix = this.uModelMatrix.copy();
+  this.uModelMatrix.scale(scaleX, scaleY, scaleZ);
```
lindapaiste, line 193: Probably this should be inside the `try`?

src/webgl/p5.Camera.js
deveshidwivedi: yes, I updated these!

src/webgl/p5.Matrix.js
```diff
@@ -47,6 +47,15 @@ p5.Matrix = class {
     return this;
   }
 
+  reset() {
+    if (this.mat3) {
```
davepagurek, line 51: Nice!

src/webgl/p5.RendererGL.js
```diff
@@ -855,7 +856,7 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
   _update() {
     // reset model view and apply initial camera transform
     // (containing only look at info; no projection).
-    this.uMVMatrix.set(this._curCamera.cameraMatrix);
+    this.uViewMatrix.set(this._curCamera.cameraMatrix);
```
davepagurek, line 859: I think we also want to reset the model matrix here

davepagurek (commented):
Great work on this so far! I left some comments where it seems like the behaviour we had before wouldn't be completely the same afterwards -- normally when we reset `uMVMatrix` to just the camera matrix in the past, now that means we have to reset the model matrix and set the view matrix to the camera matrix.

package-lock.json
```diff
@@ -47,6 +47,7 @@
         "lint-staged": "^4.3.0",
         "marked": "^4.0.10",
         "mocha": "^10.2.0",
+        "mocha-chrome": "^2.2.0",
```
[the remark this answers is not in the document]
deveshidwivedi: Sorry for this!

src/webgl/3d_primitives.js
```diff
@@ -1246,7 +1246,7 @@ p5.RendererGL.prototype.triangle = function(args) {
   // this matrix multiplication transforms those two unit vectors
   // onto the required vector prior to rendering, and moves the
   // origin appropriately.
-  const uMVMatrix = this.uMVMatrix.copy();
+  const uModelMatrix = this._renderer.uModelMatrix.copy();
```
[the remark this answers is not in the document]
deveshidwivedi: Updated these, there's still errors..

davepagurek:
Looks like tests are failing because:
image


It seems like it's because of this line still setting `uMVMatrix` rather than the model and view matrices:
https://github.com/processing/p5.js/blob/2ceca4ec656f4d137390c3554b3bd085d698ed60/src/webgl/p5.Camera.js#L1393-L1395

Might be worth searching in the repo for `this.uMVMatrix` or `_renderer.uMVMatrix` to make sure we aren't leaving any other lingering references to it

deveshidwivedi:
@davepagurek I made a few changes, please let me know if they make sense..
I rechecked files to see where I missed initialization but couldn't really make any improvements..
I had a doubt, do we need to introduce or update any camera properties specifically to get things working?

davepagurek:
Looks like tests are still failing because it still tries to set `uMVMatrix` here too: https://github.com/processing/p5.js/blob/f5baf367c2fb413ed2f1923912b185971c52e717/src/webgl/p5.Camera.js#L1394

Maybe try searching this file to make sure it doesn't reference `uMVMatrix` anywhere any more?

> I had a doubt, do we need to introduce or update any camera properties specifically to get things working?

I think the camera matrices are OK as is! I think your code is making the right changes already, so it's just a matter of finding any bits that we missed where we need to apply the same pattern.

deveshidwivedi:
There are some different errors now @davepagurek, how could these be resolved?

src/webgl/GeometryBuilder.js
```diff
@@ -46,6 +46,8 @@ class GeometryBuilder {
    * transformations.
    */
   addGeometry(input) {
+    this.renderer.uMVMatrix =
```
davepagurek, line 49: Maybe rather than creating `uMVMatrix` here, we can remove this line and update the lines below to use `uModelMatrix` instead of `uMVMatrix`. In this case we can just get the inverse transpose of the model matrix, by ignoring the camera matrix it's like we're assuming it's an identity matrix and doesn't change anything.

davepagurek (commented):
Looks like now, the test failures might be at least in part because the tests need to be updated too. e.g. here, the test makes assertions on `uMVMatrix`, which no longer exists, so we'll probably need to look through the tests and replace these with assertions on `uModelMatrix` and `uViewMatrix`:
https://github.com/processing/p5.js/blob/faf5252a47befaf92fbc8b8afbe5ec0d964097c5/test/unit/webgl/p5.Camera.js#L393-L396

I also see some other tests failing that probably shouldn't fail. A `buildGeometry` test is trying to assert that the output looks the same when you draw a model directly to the canvas as if you first save it to geometry, and then draw it to the canvas with `model`. It's receiving the left image as output, but expecting the right:

<table>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

so it looks like maybe something about the camera isn't being reset correctly after `buildGeometry`. I left some comments, but I think we're accidentally taking the camera into account in `buildGeometry` right now, hopefully that error goes away after editing that one bit!

deveshidwivedi:
Made these changes, please have a look..

deveshidwivedi:
Hi @davepagurek! I tried updating a few other tests, please take a look

src/webgl/GeometryBuilder.js
```diff
@@ -46,13 +46,9 @@ class GeometryBuilder {
    * transformations.
    */
   addGeometry(input) {
-    this.renderer.uMVMatrix =
-     this.renderer.uModelMatrix.mult(this.renderer.uViewMatrix);
-    this.hasTransform = !this.renderer.uMVMatrix.mat4
```
davepagurek, line 49: I think we still want to set `this.hasTransform`, but just using the model matrix only rather than `uMVMatrix`.

test/unit/webgl/p5.Camera.js
```diff
@@ -386,13 +386,13 @@ suite('p5.Camera', function() {
 
       // Confirmation that the argument camera and the matrix of the camera
       // that received set() match
-      assert.deepEqual(copyCam.cameraMatrix.mat4, myCam.cameraMatrix.mat4);
```
davepagurek, line 389: I think we kept the matrices on cameras the same, so I think we can keep this one the same as it was

test/unit/webgl/p5.Camera.js
```diff
@@ -386,13 +386,13 @@ suite('p5.Camera', function() {
...
+      assert.deepEqual(copyCam.modelMatrix.mat4, myCam.modelMatrix.mat4);
       assert.deepEqual(copyCam.projMatrix.mat4, myCam.projMatrix.mat4);
       // If the set()ed camera is active,
       // the renderer's matrix will also change.
       assert.deepEqual(
-        copyCam.cameraMatrix.mat4,
-        myp5._renderer.uMVMatrix.mat4
+        copyCam.modelMatrix.mat4,
```
davepagurek, line 394: For these I think we want to compare camera's `cameraMatrix` to the renderer's view matrix

test/unit/webgl/p5.RendererGL.js
```diff
@@ -593,13 +593,23 @@ suite('p5.RendererGL', function() {
...
+
       myp5.push();
       myp5.rotateX(Math.random(0, 100));
       myp5.translate(20, 100, 5);
-      assert.notEqual(modelView.mat4, myp5._renderer.uMVMatrix.mat4);
+      // Check if the model matrix has changed
+      assert.notDeepEqual(modelMatrixBefore.mat4,
+        myp5._renderer.uModelMatrix.mat4);
+      // Check if the view matrix has changed
+      assert.deepEqual(viewMatrixBefore.mat4,
```
davepagurek, line 606: Should this one be `notDeepEqual` too like the one above?

test/unit/webgl/p5.Shader.js
```diff
@@ -67,7 +67,8 @@ suite('p5.Shader', function() {
       var expectedAttributes = ['aPosition', 'aNormal', 'aTexCoord'];
 
       var expectedUniforms = [
-        'uModelViewMatrix',
+        'uModelMatrix',
```
davepagurek, line 70: I think we can keep these the same too -- we still use `uModelViewMatrix` for shaders here: https://github.com/processing/p5.js/blob/b9828a71c63204c08e4e7c6435995f9603ed3044/src/webgl/p5.Shader.js#L339

...and although now we also pass in `uModelMatrix` and `uViewMatrix` for user shaders that need it, these tests only assert that uniforms are present in our existing shaders, which don't use those yet.

davepagurek (commented):
Thanks for your continued work on this! I noticed a few more things that might be throwing the tests off, let's see if those changes get the tests passing.

deveshidwivedi:
Should we update the matrix [here](https://github.com/processing/p5.js/blob/05ea0a4d96d8f5972f9aae9ff13862772610bf0c/src/accessibility/outputs.js#L547) also?

src/webgl/p5.Camera.js
```diff
@@ -1390,7 +1390,8 @@ p5.Camera = class Camera {
     this.cameraMatrix.translate([tx, ty, tz]);
 
     if (this._isActive()) {
-      this._renderer.uMVMatrix.set(this.cameraMatrix);
+      this._renderer.uModelMatrix.set(this.modelMatrix);
```
davepagurek, line 1393: `this` is a p5.Camera here, which doesn't have a `modelMatrix` or `viewMatrix`, just a camera matrix. The errors this is logging are I think the cause of one of the test failures. I think this line and the one below should be replaced by:
```js
this._renderer.uModelMatrix.reset();
this._renderer.uViewMatrix.set(this.cameraMatrix);
```

src/webgl/p5.Framebuffer.js
```diff
@@ -842,9 +842,10 @@ class Framebuffer {
     // RendererGL.reset() does, but this does not try to clear any buffers;
     // it only sets the camera.
     this.target.setCamera(this.defaultCamera);
-    this.target._renderer.uMVMatrix.set(
-      this.target._renderer._curCamera.cameraMatrix
-    );
+    this.target.resetMatrix();
+    this.target._renderer.uViewMatrix
+      .set(this.target._renderer._curCamera.cameraMatrix);
+    this.target._renderer.uModelMatrix.set(this.target.modelMatrix);
```
davepagurek, line 848: I think the target context doesn't have a `modelMatrix` property, so after setting the view matrix, we just need to reset the model matrix:
```suggestion
    this.target._renderer.uModelMatrix.reset();
```

src/webgl/p5.Shader.js
```diff
@@ -316,9 +316,10 @@ p5.Shader = class {
   }
 
   _setMatrixUniforms() {
+    const modelMatrix = this._renderer.uModelMatrix;
     const viewMatrix = this._renderer._curCamera.cameraMatrix;
     const projectionMatrix = this._renderer.uPMatrix;
-    const modelViewMatrix = this._renderer.uMVMatrix;
+    const modelViewMatrix = (modelMatrix.copy()).mult(viewMatrix);
```
davepagurek, line 322: I think the problem with the tests right now might have to do with the fact that we're still using `this._renderer.uMVMatrix` below to calculate the normal matrix, but that property is never updated. Can we add a line below this that does the following?
```js
this._renderer.uMVMatrix = modelViewMatrix;
```

src/webgl/p5.Shader.js
```diff
@@ -316,9 +316,10 @@ p5.Shader = class {
   }
 
   _setMatrixUniforms() {
+    const modelMatrix = this._renderer.uModelMatrix;
     const viewMatrix = this._renderer._curCamera.cameraMatrix;
```
davepagurek, line 320: Should be the same behaviour for now, but for consistency, can we read the new view matrix instead of directly from the camera?
```suggestion
    const viewMatrix = this._renderer.uViewMatrix;
```

davepagurek: Similarly, below, where it currently has
```js
    if (this.uniforms.uCameraRotation) {
      this._renderer.curMatrix.inverseTranspose(this._renderer.
_curCamera.cameraMatrix);
```
Can we change it to:
```js
    if (this.uniforms.uCameraRotation) {
      this._renderer.curMatrix.inverseTranspose(this._renderer.uViewMatrix);
```

davepagurek (commented):
I think we're almost done! We only have a few tests still failing, and I think the suggestions I've left will hopefully be enough to resolve them.

davepagurek:
> Should we update the matrix [here](https://github.com/processing/p5.js/blob/05ea0a4d96d8f5972f9aae9ff13862772610bf0c/src/accessibility/outputs.js#L547) also?

Ah yes good catch, I think so!

deveshidwivedi:
Hi @davepagurek! Sorry for the delay. I have made these changes and the tests do no fail anymore. Are any more updates required?

deveshidwivedi:
Sure @davepagurek! Looking forward!🎉

Qianqianye:
Thank you all. @davepagurek, are we good to merge this?

davepagurek:
good to go if we've done our last docs deploy!

Qianqianye:
Just to clarify, do you mean the updating all the reference doc? @davepagurek

davepagurek:
right, mostly we were just waiting to put this into a release where we would be able to have a standard beta testing period to catch any bugs. if we're going to be back into a regular release cadence again then this is good to merge!

davepagurek:
@all-contributors please add @deveshidwivedi for code

PR #6806, opened 2024-02-12
using typed array in noise.js

JayPokale:
## Changes:
src/math/noise.js:
Using Float64Array instead of regular Array to reduce memory and optimize efficiency.

#### PR Checklist
- [x] `npm run lint` passes

### DIFF
--- a/src/math/noise.js
+++ b/src/math/noise.js
@@ -198,8 +198,8 @@ let perlin; // will be initialized lazily by noise() or noiseSeed()
  */
 
 p5.prototype.noise = function(x, y = 0, z = 0) {
-  if (perlin == null) {
-    perlin = new Array(PERLIN_SIZE + 1);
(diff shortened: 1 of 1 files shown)


src/math/noise.js
```diff
@@ -198,8 +198,8 @@ let perlin; // will be initialized lazily by noise() or noiseSeed()
  */
 
 p5.prototype.noise = function(x, y = 0, z = 0) {
-  if (perlin == null) {
-    perlin = new Array(PERLIN_SIZE + 1);
+  if (perlin === null) {
```
limzykenneth, line 201: This causes the test to fail. `perlin` is never `null` but rather it is `undefined` at the start, the previous comparison is more loose so it passed.

limzykenneth (changes requested):

PR #6807, opened 2024-02-12
Default values of arguments && modernize some code

### LINKED ISSUE #6721: Default values of arguments
### Topic

I noticed that the default values for arguments are being set using if ladders. I think we can pass the defaults in the function's parameters itself to save lots of lines of code. But before jumping into making those changes I wanted to know the reason behind setting them like that. What do you think about changing the way of setting the default values?

```js
p5.prototype.cone = function(radius, height, detailX, detailY, cap) {
  this._assert3d('cone');
  p5._validateParameters('cone', arguments);
  if (typeof radius === 'undefined') {
    radius = 50;
  }
  if (typeof height === 'undefined') {
    height = radius;
  }
  if (typeof detailX === 'undefined') {
    detailX = 24;
  }
  if (typeof detailY === 'undefined') {
    detailY = 1;
  }
  if (typeof cap === 'undefined') {
    cap = true;
  }
```

asukaminato0721:
Resolves #6721

 Changes:

use default values of arguments, fix some missing `...args`, use some native apis.

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes


src/math/p5.Vector.js
```diff
@@ -365,76 +342,100 @@ p5.Vector = class {
...
+    }
+    if (yComponent !== 0) {
+      this.y = this.y % yComponent;
+    }
+    if (zComponent !== 0) {
+      this.z = this.z % zComponent;
+    }
+    return this;
+  }
+
```
limzykenneth, line 369: I'm not sure how well it will work or be supported but I wonder can these be marked as [private methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties)?

limzykenneth (changes requested):
One comment inline and one more thing which is that there are some inconsistency in terms of space before parenthesis after the `function` key word, ie

```js
function ()

// or

function()
```

For most of the rest of the code base, the consistent use is not having the space so I tend towards that. @nickmcintyre I can't find specific recommendation for this in the documentation style guide, do you have some thoughts on this?

src/math/p5.Vector.js
asukaminato0721: I am not sure the support of `#field` , so I use `@private`.

limzykenneth:
I'll do a final review later today to see if private methods will work or not. After that I will merge this.

limzykenneth:
The new syntax doesn't seem to work with browserify, we'll re-explore this as part of 2.0. I'll merge this for now. Thanks @asukaminato0721

adonis-jimenez:
> Resolves #6721
> 
> Changes:
> 
> use default values of arguments, fix some missing `...args`, use some native apis.
> 
> Screenshots of the change:
> 
> #### PR Checklist
> * [x]  `npm run lint` passes
> * [x]  [Inline documentation](https://github.com/processing/p5.js/blob/main/contributor_docs/inline_documentation.md) is included / updated
> * [x]  [Unit tests](https://github.com/processing/p5.js/tree/main/contributor_docs#unit-tests) are included / updated

PR #6823, opened 2024-03-02
Nat Decker and Bobby Joe Smith III updated access.md

natdecker:
Part of 2023 Processing Foundation Fellowship


Qianqianye:
@all-contributors please add @natdecker for a11y, doc

PR #6829, opened 2024-03-03
Contributor git workflow clarification

### LINKED ISSUE #6828: Documentation: Clarification of the necessity of `npm test` before `npm run build` for the Git Workflow portion of the Contributor Docs
### Increasing access

This would make it easier for those who aren't used to p5.js to understand what's going wrong when `npm run build` fails since https://github.com/processing/p5.js/blob/main/contributor_docs/contributor_guidelines.md#git-workflow can, to many, imply that `npm test` is unnecessary if you simply want to build the repo despite it needing to be ran at least once before `npm run build` is ran.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [X] Other (specify if possible)

Change the wording to make it very clear that running `npm test` once before running `npm run build` is required.

RandomGamingDev:
Resolves #6828

 Changes:

Changes the wording to clarify the fact that `npm test` has to be ran once before running `npm run build`.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/contributor_docs/contributor_guidelines.md
+++ b/contributor_docs/contributor_guidelines.md
@@ -327,7 +327,7 @@ Now, you are ready to make the changes you need to make; for more details about 
 npm test
 ```
 
-To try building p5.js from scratch and run all unit tests, this should complete with no errors. If you just want to build the library without running the tests, you can run:
+to build p5.js from scratch and run all unit tests (this should complete with no errors). After you've ran that once you'll be able to build p5.js without running the tests, by running:
 
 ```
 npm run build
--- a/contributor_docs/contributor_guidelines.md
+++ b/contributor_docs/contributor_guidelines.md
@@ -327,7 +327,7 @@ Now, you are ready to make the changes you need to make; for more details about 
 npm test
 ```
 
-to build p5.js from scratch and run all unit tests (this should complete with no errors). After you've ran that once you'll be able to build p5.js without running the tests, by running:
+to build p5.js from scratch and run all unit tests (this should complete with no errors). After you've ran that once you'll be able to build p5.js without running the tests by running:
 
 ```
 npm run build


RandomGamingDev:
> It might be possible to make `npm run build` work directly after install so we don't need to necessarily recommend running `npm test` first. I can look into it once I get some time.

It's been quite a bit of time from you're original comment. Have you already tested whether it's possible on certain setups?

On the setups I've tested it doesn't seem to work since `npm run build` requires a file generated by calling `npm test`.

limzykenneth:
Fixed with https://github.com/processing/p5.js/commit/f14a45e3aa7324329711506fdccf391c6b67d39e

@RandomGamingDev In the future it can help greatly if you can provide more information around replication and expected vs actual output, that way we don't need to spend time on replication. Eg. the relevant error message here is

```
Fatal error: Cannot find module '../../../docs/parameterData.json' from '/home/limzykenneth/Documents/p5.js/src/core/friendly_errors'
```
Just knowing this narrows down the fix faster. Thanks.

RandomGamingDev:
> Fixed with [f14a45e](https://github.com/processing/p5.js/commit/f14a45e3aa7324329711506fdccf391c6b67d39e)
> 
> @RandomGamingDev In the future it can help greatly if you can provide more information around replication and expected vs actual output, that way we don't need to spend time on replication. Eg. the relevant error message here is
> 
> ```
> Fatal error: Cannot find module '../../../docs/parameterData.json' from '/home/limzykenneth/Documents/p5.js/src/core/friendly_errors'
> ```
> 
> Just knowing this narrows down the fix faster. Thanks.

That's indeed the error. Sorry for not providing more information and thanks for the fix.

PR #6831, opened 2024-03-05
Update p5.Vector.js

asukaminato0721:
- use `...args`
- add missing `@return`

Resolves https://github.com/p5-types/p5.ts/issues/123

 Changes:

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -435,7 +435,8 @@ p5.Vector = class {
    * @param {p5.Vector | Number[]}  value  divisor vector.
    * @chainable
    */
-  rem (x, y, z) {
+  rem (...args) {
+    let [x, y, z] = args;
     if (x instanceof p5.Vector) {
       if ([x.x,x.y,x.z].every(Number.isFinite)) {
         const xComponent = parseFloat(x.x);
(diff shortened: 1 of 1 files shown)


src/math/p5.Vector.js
```diff
@@ -456,31 +457,31 @@ p5.Vector = class {
...
+        this.y = this.y % x;
+        this.z = this.z % x;
         return this;
       }
-    } else if (arguments.length === 2) {
-      const vectorComponents = [...arguments];
+    } else if (args.length === 2) {
+      const vectorComponents = args;
       if (vectorComponents.every(element => Number.isFinite(element))) {
         if (vectorComponents.length === 2) {
```
lindapaiste, line 470: Not related to your changes, but isn't this always `true`?

lindapaiste:
The `@return` statements look good.  So do the default arguments `length = 1`.  I'm not understanding the benefit of switching from named arguments to `...args`.

limzykenneth:
@lindapaiste There's a bit of a tradeoff here I think. Named arguments probably will make things slightly clearer but `...args` have the benefit of being actual array that array prototype methods can be called on, ie. it saves the need to convert `arguments` to an actual array when required.

lindapaiste:
> @lindapaiste There's a bit of a tradeoff here I think. Named arguments probably will make things slightly clearer but `...args` have the benefit of being actual array that array prototype methods can be called on, ie. it saves the need to convert `arguments` to an actual array when required.

It looks like only 3 of these methods do the `[...arguments]` casting and the rest don't.

Maybe I'd see the benefit more if the intermediate variable `vectorComponents` was removed, since it no longer serves any purpose when it's the same variable as `args` (`const vectorComponents = args;`).   You can now call `args.every(element =>` directly.

asukaminato0721:
> the benefit of switching from named arguments to ...args.

because the origin `x, y, z` is kind of misleading. It's not something like `Promise(resolve, reject)`, in which resolve is always resolve. The x can be not just "x" , but an array, which is called overload.

Besides, according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments#description

> Note: In modern code, [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) should be preferred.

limzykenneth:
Looks good. Thanks!

PR #6867, opened 2024-03-15
Update p5.Image references

nickmcintyre:
Preparing references in `src/image/p5.Image.js` for the new website.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/image/p5.Image.js
+++ b/src/image/p5.Image.js
@@ -118,6 +118,7 @@ p5.Image = class {
      * @property {Number} width
      * @name width
      * @readOnly
+     *
      * @example
      * <div>
      * <code>
@@ -154,6 +155,7 @@ p5.Image = class {
      * @property height
      * @name height
      * @readOnly
+     *
      * @example
      * <div>
      * <code>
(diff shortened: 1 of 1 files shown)


src/image/p5.Image.js
```diff
@@ -153,10 +194,10 @@ p5.Image = class {
     //For WebGL Texturing only: used to determine whether to reupload texture to GPU
     this._modified = false;
     /**
-     * An array containing the color of each pixel in the
-     * <a href="#/p5.Image">p5.Image</a> object. Colors are stored as numbers
-     * representing red, green, blue, and alpha (RGBA) values. `img.pixels` is a
-     * one-dimensional array for performance reasons.
+     * An array containing the color of each pixel in the image.
```
Qianqianye, line 197: Hey @nickmcintyre, could you share your thoughts on replacing `p5.Image` to 'image' in some of the descriptions while keeping `p5.Image` in others?

nickmcintyre: @Qianqianye my thoughts are "oops!" I suggest we use "image" in the one-line description and refer/link to `p5.Image` objects when it's helpful to clarify types in the long description.

Qianqianye: Thanks @nickmcintyre. 
I'd like to get @limzykenneth and @davepagurek's input on this. Since we are considering the possibility of surfacing the instance method references to the main reference page, would it be more or less confusing to include `p5.Image` in the references?

limzykenneth: For cases where it is referencing things like type of value passed as arguments, I think it is probably better to use `p5.Image` to avoid ambiguity (eg. There is an `Image` constructor that creates `HTMLImageElement` and a selected `<image>` element for example). If the use is just descriptive or self-describing, such as in this case, I think it should be fine.

PR #6877, opened 2024-03-16
Update environment.js

sphantom-code:
Simplify verbage of describe string for the framerate example.


Qianqianye:
Great catch, thanks @sphantom-code!

@all-contributors please add @sphantom-code for documentation

PR #6922, opened 2024-03-24
New Swedish translation

matsragnar:
Changes:
Added Swedish translation of error messages.


limzykenneth:
There are too many unrelated changes in this PR with most of it coming from out of date branch. We usually require an issue to be opened for discussions before a PR should be filed, so that we can work out any implementation details or to indicate that we are not interested in including any specific features in the library before any work is done. Please open a separate issue about what changes you are proposing. 

I'm closing this for now. Thanks.

PR #6984, opened 2024-04-20
fixed parameters for normal() method

### LINKED ISSUE #6982: Incorrect FES message for normal()

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [X] Friendly errors

### p5.js version

1.9.2

### Web browser and version

Chrome 124.0.6367.60

### Operating system

macOS 14.4.1

### Steps:
1. Create a `p5.Vector` object.
2. Pass the object to `normal()`.

### Snippet:

```js
// Create a p5.Vector object.
let n = createVector(-0.4, -0.4, 0.8);

// Draw the shape.
// Use normal() to set vertex normals.
beginShape();
normal(n);
vertex(-30, -30, 0);
// ...rest of the shape.
endShape();
```

I noticed this while working on the reference. Passing a `p5.Vector` to `normal()` generates the following FES message:

```
🌸 p5.js says: [sketch.js, line 29] normal() was expecting Vector
for the first parameter, received object instead. (http://p5js.org/reference/#/p5/normal)
```

See the [full sketch](https://editor.p5js.org/mcintyre/sketches/dVC_YegG5).

kaiserarg:
Resolves #6982

 Changes:

It seems that in `src\core\shape\vertex.js` for the `normal()` method the documentation was written as `@param  {Vector} vector `and since `Vector `isn't a native JS class when the FES couldn't use `instanceOf` to check and `typeOf `checked the e.g. example `let n = createVector(-0.4, -0.4, 0.8); ` which would return object since Vector is not a native JS class.

A quick and easy fix would be to change `@param  {Vector} vector` to `@param  {p5.Vector} vector` so FES can correctly check the parameter.

 Screenshots of the change:

image

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/shape/vertex.js
+++ b/src/core/shape/vertex.js
@@ -2117,7 +2117,7 @@ p5.prototype.vertex = function(x, y, moveTo, u, v) {
  * </code>
  *
  * @method normal
- * @param  {Vector} vector vertex normal as a <a href="#/p5.Vector">p5.Vector</a> object.
+ * @param  {p5.Vector} vector vertex normal as a <a href="#/p5.Vector">p5.Vector</a> object.
  * @chainable
  *
  * @example


davepagurek (approved):
Thanks for the fix!

PR #7036, opened 2024-05-13
Fix map example

PalumboN:
Fixing the example of `map` function. The bottom line was drown vertical instad of horizontal, as the example try to show.

```
'Two horizontal lines. The top line grows horizontally as the mouse moves to the right. The bottom line also grows horizontally but is scaled to stay on the left half of the canvas.'
```


nickmcintyre:
Good catch @PalumboN!

Qianqianye:
Thanks!
@allcontributors please add @PalumboN for bug and documentation

PR #7054, opened 2024-05-16
fixed nf to work with negative number

### LINKED ISSUE #7046: `nf()` produces problematic string-formatting of negative numbers

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [X] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

1.9.3

### Web browser and version

Chrome 124.0.6367.208

### Operating system

macOS 14.4.1

### Steps:
1. Call `nf()` with a negative number and extra left padding.
2. Display the formatted number.

### Snippet:

```js
function setup() {
  createCanvas(400, 400);

  background(220);

  // Format a negative number with
  // extra left padding.
  let num = nf(-123, 5);

  // Style the text.
  textAlign(CENTER, CENTER);
  textSize(32);

  // Display the text.
  text(num, 200, 200); // 0-123
}
```

Here's the [sketch](https://editor.p5js.org/mcintyre/sketches/XmxqHk65M) for reference.

After reviewing #5710, it may be worth revisiting the string utility functions as part of p5.js 2.0.

Akhilbisht798:

 Changes:

resolve #7046

 Screenshots of the change:

#### PR Checklist

Screenshot from 2024-05-16 21-49-43

- [x] `npm run lint` passes

### DIFF
--- a/src/utilities/string_functions.js
+++ b/src/utilities/string_functions.js
@@ -295,20 +295,21 @@ p5.prototype.nf = function(nums, left, right) {
 };
 
 function doNf(num, left, right) {
+  let isNegative = num < 0;
+  num = Math.abs(num);
   let [leftPart, rightPart] = num.toString().split('.');
+
 
   if (typeof right === 'undefined') {
     leftPart = leftPart.padStart(left, '0');
-    return rightPart ? leftPart + '.' + rightPart : leftPart;
+    let result = rightPart ? leftPart + '.' + rightPart : leftPart;
+    return isNegative ? '-' + result : result;
   } else {
     let roundedOff = num.toFixed(right);
     [leftPart, rightPart] = roundedOff.toString().split('.');
     leftPart = leftPart.padStart(left, '0');
-    if(typeof rightPart === 'undefined'){
-      return leftPart;
-    }else{
-      return leftPart + '.' + rightPart;
-    }
+    let result = typeof rightPart === 'undefined' ? leftPart : leftPart + '.' + rightPart;
+    return isNegative ? '-' + result : result;
   }
 }
 
--- a/src/utilities/string_functions.js
+++ b/src/utilities/string_functions.js
@@ -222,6 +222,9 @@ p5.prototype.matchAll = function(str, reg) {
  * then unused decimal places will be set to 0. For example, calling
  * `nf(123.45, 4, 3)` returns the string `'0123.450'`.
  *
+ * When the number is negative, for example, calling `nf(-123.45, 5, 2)`
+ * returns the string `'-00123.45'`.
+ *
  * @method nf
  * @param {Number|String} num number to format.
  * @param {Integer|String} [left] number of digits to include to the left of
--- a/src/utilities/string_functions.js
+++ b/src/utilities/string_functions.js
@@ -250,21 +250,21 @@ p5.prototype.matchAll = function(str, reg) {
  *
  *   // Display the number as a string.
  *   let formatted = nf(number);
- *   text(formatted, 20, 25);
+ *   text(formatted, 20, 20);
  *
  *   let negative = nf(-number, 4, 2);
- *   text(negative, 20, 37);
+ *   text(negative, 20, 40);
  *
  *   // Display the number with four digits
  *   // to the left of the decimal.
  *   let left = nf(number, 4);
- *   text(left, 20, 50);
+ *   text(left, 20, 60);
  *
  *   // Display the number with four digits
  *   // to the left of the decimal and one
  *   // to the right.
  *   let right = nf(number, 4, 1);
- *   text(right, 20, 75);
+ *   text(right, 20, 80);
  *
  *   describe(
  *     'The numbers "123.45", "0123.45", and "0123.5" written on three separate lines. The text is in black on a gray background.'
--- a/src/utilities/string_functions.js
+++ b/src/utilities/string_functions.js
@@ -252,6 +252,9 @@ p5.prototype.matchAll = function(str, reg) {
  *   let formatted = nf(number);
  *   text(formatted, 20, 25);
  *
+ *   let negative = nf(-number, 4, 2);
+ *   text(negative, 20, 37);
+ *
  *   // Display the number with four digits
  *   // to the left of the decimal.
  *   let left = nf(number, 4);
--- a/src/utilities/string_functions.js
+++ b/src/utilities/string_functions.js
@@ -267,7 +267,7 @@ p5.prototype.matchAll = function(str, reg) {
  *   text(right, 20, 80);
  *
  *   describe(
- *     'The numbers "123.45", "0123.45", and "0123.5" written on three separate lines. The text is in black on a gray background.'
+ *     'The numbers "123.45", "-0123.45", "0123.45", and "0123.5" written on four separate lines. The text is in black on a gray background.'
  *   );
  * }
  * </code>
--- a/test/unit/utilities/string_functions.js
+++ b/test/unit/utilities/string_functions.js
@@ -111,6 +111,12 @@ suite('String functions', function() {
       result = myp5.nf(num, 3, 0);
       assert.equal(result, '123');
     });
+
+    test('should return correct string', function() {
+      var num = -123;
+      result = myp5.nf(num, 5);
+      assert.equal(result, '-00123');
+    });
   });
 
   suite('p5.prototype.nfc', function() {


perminder-17 (changes requested):
Firstly, I want to apologize for the delayed review, @Akhilbisht798. Due to the maintainers being occupied with the version update, I’ll be taking over the review of your work.

I appreciate your patience and the effort you’ve put into this task. Your work is mostly accurate, but there is one issue with the rounding of numbers after the decimal point that may need to be addressed.

For example, if we use `let num = nf(123.58, 1,1);`, the expected output should be `123.6`, as it involves rounding the numbers after the decimal point. However, in your implementation, it currently results in `123.5` without rounding off correctly. Do you think we could adjust this?

Akhilbisht798:
hey @perminder-17 I can work on this.

perminder-17:
Thanks, @Akhilbisht798, for your great work so far, and I apologize again for the delay—I hope you don’t mind. I’d like to invite the utility stewards, @limzykenneth, @glopzel, and @davepagurek and @nickmcintyre  to share their thoughts as well.

I noticed that @limzykenneth has already opened a discussion on this issue (https://github.com/processing/p5.js/issues/7046#issuecomment-2200266696).

In short, should we be writing `let num = nf(-123, 4);` where 4 represents the number of digits to include to the left? Currently, the output produced after solving by @Akhilbisht798  is -0123, but if we count the minus sign (-) as a character, the output would be -123. The question is: should we consider the minus sign in the digit count? 

From my opinion I go with **not** counting minus sign(-). 

I’d appreciate everyone’s input on this discussion that @limzykenneth started. Also, if you have any other suggestions regarding nf(), please feel free to share. (To be honest, I haven’t used nf() much myself, so I’m a bit uncertain about it.)

limzykenneth:
I think we can possibly go with not counting the minus sign, which is also the behaviour of Processing. For those who needed the numbers to line up, `nfs()` can solve that.

src/utilities/string_functions.js
```diff
@@ -295,20 +295,23 @@ p5.prototype.nf = function(nums, left, right) {
...
   if (typeof right === 'undefined') {
     leftPart = leftPart.padStart(left, '0');
-    return rightPart ? leftPart + '.' + rightPart : leftPart;
+    let result = rightPart ? leftPart + '.' + rightPart : leftPart;
+    return isNegative ? '-' + result : result;
   } else {
-    let roundedOff = num.toFixed(right);
-    [leftPart, rightPart] = roundedOff.toString().split('.');
+    let roundedNum = Math.round(num * Math.pow(10, right));
+    roundedNum = roundedNum / Math.pow(10, right);
```
perminder-17, line 309: Can we roundooff the num by simply doing this:
```suggestion
    let roundedNum = num.toFixed(right);
```

Let me know if it works for you @Akhilbisht798 .

src/utilities/string_functions.js
```diff
@@ -295,20 +295,23 @@ p5.prototype.nf = function(nums, left, right) {
...
+    [leftPart, rightPart] = roundedNum.toString().split('.');
     leftPart = leftPart.padStart(left, '0');
-    if(typeof rightPart === 'undefined'){
-      return leftPart;
-    }else{
-      return leftPart + '.' + rightPart;
-    }
+    let result = typeof rightPart === 'undefined' ?
+      leftPart : leftPart + '.' + rightPart.padEnd(right, '0');
```
perminder-17, line 313: Was there a need to write `.padEnd`?
```suggestion
      leftPart : leftPart + '.' + rightPart;
```

can we simply add this? What you think?

src/utilities/string_functions.js
```diff
@@ -295,20 +295,23 @@ p5.prototype.nf = function(nums, left, right) {
...
+    let roundedNum = Math.round(num * Math.pow(10, right));
+    roundedNum = roundedNum / Math.pow(10, right);
+    [leftPart, rightPart] = roundedNum.toString().split('.');
     leftPart = leftPart.padStart(left, '0');
-    if(typeof rightPart === 'undefined'){
-      return leftPart;
-    }else{
-      return leftPart + '.' + rightPart;
-    }
+    let result = typeof rightPart === 'undefined' ?
```
perminder-17, line 312: Do you think @Akhilbisht798 that using ternary operator for result could be a bit complicated for the other contributors to understand in a first look. Can we use if and else block for this part only? Do you agree, or maybe you have any other thoughts? Please do let me know.

perminder-17 (commented):
Everything looks good, Thanks @limzykenneth for your response..just some minor cleanups then we are good to merge. Thanks @Akhilbisht798 for your hardwork and patience :)

src/utilities/string_functions.js
Akhilbisht798: yes we can, previous commit was using it. but as you have mentioned that 

> For example, if we use let num = nf(123.58, 1,1);, the expected output should be 123.6, as it involves rounding the numbers after the decimal point. However, in your implementation, it currently results in 123.5 without rounding off correctly. Do you think we could adjust this?


This is was not working with `num.toFixed(right);`. But if you want I can change it back.

ashish1729:
do we need to update some test + reference ?

src/utilities/string_functions.js
Akhilbisht798: So this add padding of 0 at the right decimal place if the right is more than the digit in the right. 
if we doNf with right of 2 in 123.5 it will pad 123.50.
I guess it would be great. But tell me known what you think about this. I can change if it is not required.

src/utilities/string_functions.js
Akhilbisht798: I have no problem in using if else block for this statement if that increase the look. Let me known what you think about the other changes and I will commit the changes of this with them.

ashish1729: just a thought...  

if we want to reduce the usage of ternary operator it can be mentioned in the contributor guideline somewhere

or if we have strong opinions about it, it should be part of some lint system to give warning / error

perminder-17:
Hmm..I realize I made an error earlier. When I reviewed the code again, I found that it wasn’t rounding the numbers as I initially thought, and the mistake was on my side. I sincerely apologize for any confusion caused. It seems your previous commit was indeed correct. I hadn’t thoroughly checked the code and was only focusing on the testing, which led me to overlook this. After testing it again now, everything appears to be working perfectly. 

here's the build of your previous commit:
https://editor.p5js.org/aman12345/sketches/Pq9D8HM0h

We can revert to the previous commit, as it looks all good to me. I’m really sorry for any inconvenience caused, and then we could merge this. Really thankful for your patience and hard work @Akhilbisht798.

src/utilities/string_functions.js
perminder-17: Right, when we will use `num.toFixed(right)` to roundoff the numbers, I don't think we would actually need to add padding of 0 to the right.

perminder-17:
> do we need to update some test + reference ?

Yes, adding tests would be a great idea. I believe we already have a good reference sketch with a detailed explanation in the code. If @Akhilbisht798 has some time, we could also add a test for negative numbers, which is currently causing issues. You can check it here: 
here: https://github.com/processing/p5.js/blob/7d393ec4215ce20f7bc3178b779e0ec67689f928/test/unit/utilities/string_functions.js#L79 

we could probably add the same way as it is written here ;)

src/utilities/string_functions.js
perminder-17: Were you also able to reproduce the same error of rounding off with `num.toFixed(right);`. I am currently testing and here's the build and it works well.

https://editor.p5js.org/aman12345/sketches/Pq9D8HM0h

src/utilities/string_functions.js
perminder-17: > just a thought...
> 
> if we want to reduce the usage of ternary operator it can be mentioned in the contributor guideline somewhere
> 
> or if we have strong opinions about it, it should be part of some lint system to give warning / error

It's not about reducing the use of ternary operators. In this case, I was just considering whether using if and else blocks might make the code more readable, especially when checking the type. If it's undefined, we perform one operation, and for everything else, we do something different—all in a single line. I'm open to discussion and would love to hear your thoughts on this. Otherwise ternary operator also works for me. Any feedback @Akhilbisht798 and @ashish1729 ? 

```
let result = typeof rightPart === 'undefined' ? leftPart : leftPart + '.' + rightPart;
```

Akhilbisht798:
@perminder-17 It is not a problem. I will revert back to the previous commit and will add the test case for this one also.

Akhilbisht798:
@perminder-17 all done you can let me known.

perminder-17 (approved):
Looks good @Akhilbisht798 . Thanks for your hard work:)

ashish1729:
should update the reference for negative numbers

I can see myself using this function for having a fixed-width display... but the way it is implemented negative numbers have more width due to the extra minus sign.

can be taken up as a separate task to unblock the merging of this PR

perminder-17:
> should update the reference for negative numbers

Yes, we could consider adding support for negative numbers as well. However, since the issue specifically addresses the bug 'nf() produces problematic string formatting for negative numbers,' this PR resolves that issue correctly. It might be a good idea to handle any additional improvements in a separate task.

ashish1729:
oh by reference I meant documentation, the reference page of the website ..

perminder-17:
> oh by reference I meant documentation, the reference page of the website ..

Yep, I was imagining the same. Sorry... I will make the above comment more clearer. Haha😅

ashish1729:
then why is documentation considered additional improvement ? shouldn't it be in sync with any update in the code?


the PR on top has 3 checkboxes, which is included in every PR
- npm run lint passes
- [Inline documentation](https://github.com/processing/p5.js/blob/main/contributor_docs/inline_documentation.md) is included / updated
- [Unit tests](https://github.com/processing/p5.js/tree/main/contributor_docs#unit-tests) are included / updated


It was my understanding that documentation should be updated to check the 2nd point

perminder-17:
That's a good point. However, since the PR has been open for five months and @Akhilbisht798 has been very patient and kind in making updates and adding tests, requesting more changes one by one might not be ideal. @Akhilbisht798, if you're interested, it would be great if you could add the documentation for negative numbers, but if you're busy, I'd be happy to work on it and submit a PR. Again thanks a lot for your hard work and patience.

ashish1729:
I am also trying to learn how thing works, so thank you for explaining.

Akhilbisht798:
@perminder-17 No problem, I can add documentation changes. 
but I personally think that the inline documentation for this function is fine and explain things perfectly and to just a changes for negative number won't be that big of deal and won't be changing the functionality in a big way

perminder-17:
Right, the current documentation is also very good no doubt. But as @ashish1729 said, 

> I can see myself using this function for having a fixed-width display... but the way it is implemented negative numbers have more width due to the extra minus sign.

Also, one more reason could be, we are not counting the minus sign, so we could also add a line of comment describing it in the code with an example sketch where it shows a negative  number. 

What you think?

Akhilbisht798:
ya no problem i can do that.

src/utilities/string_functions.js
```diff
@@ -222,6 +222,9 @@ p5.prototype.matchAll = function(str, reg) {
  * then unused decimal places will be set to 0. For example, calling
  * `nf(123.45, 4, 3)` returns the string `'0123.450'`.
  *
+ * When the number is negative, for example, calling `nf(-123.45, 5, 2)`
```
perminder-17, line 225: Thank you for adding that. Just one last thought: could you also modify the reference example so that it appears like this?

image

Since we are not counting the minus sign as an extra digit, negative numbers may have greater width due to the additional minus sign. Therefore, we could show users that negative numbers occupy more space as we are not counting it.

Akhilbisht798: ya no problem.

Akhilbisht798:
hey @perminder-17 let me known

perminder-17 (approved):
Looks good to me. Thanks for your work and your patience.

PR #7087, opened 2024-06-06
Fix nf function to handle undefined left parameter , giving a default value and update documen…

### LINKED ISSUE #6994: `nf` seems to have wrong prototypes

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [X] Utilities
- [ ] WebGL
- [ ] Build process
- [X] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

main branch

### Web browser and version

_No response_

### Operating system

_No response_

### Steps to reproduce this

This is generally a report about the documentation and test cases.

From the [reference](https://p5js.org/reference/#/p5/nf), the prototypes for `nf` are:

* `nf(num, [left], [right])`
* `nf(nums, [left], [right])`

However, from the source code, especially in `doNf(num, left, right)`, there isn't check of `type left === 'undefined'`, nor a default value for `left`, so when calling `leftPart = leftPart.padStart(left, '0')`, `left` cannot be `undefined`.

Also in the test suite of `p5.prototype.nf`, test cases are missing:
- `result = nf(1234)`, if `left` _is_ optional
- `result = nf([1234, 3.141516, 3.141516e-2], 2)` for an array of numbers to format

VANSH3104:
Resolves #6994

Changes:
- Set default value for left parameter in doNf function and also update it for undefined.
- documentation for nf function.

Screenshots of the change:
Screenshot from 2024-06-06 22-55-59
Screenshot from 2024-06-06 22-55-38

#### PR Checklist

- [ ✓] `npm run lint` passes
- [ x ] [Unit tests] are included / updated

### DIFF
--- a/src/utilities/string_functions.js
+++ b/src/utilities/string_functions.js
@@ -224,8 +224,8 @@ p5.prototype.matchAll = function(str, reg) {
  *
  * @method nf
  * @param {Number|String} num number to format.
- * @param {Integer|String} [left] number of digits to include to the left of
- *                                the decimal point.
+ * @param {Integer|String} [left=0] number of digits to include to the left of
+ *                                the decimal point. Defaults to 0 if undefined.
  * @param {Integer|String} [right] number of digits to include to the right
  *                                 of the decimal point.
  * @return {String} formatted string.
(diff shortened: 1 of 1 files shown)


limzykenneth:
Hi @VANSH3104, thanks for the PR. Unfortunately I don't think this is necessarily the right fix. For `nf()` it may not make sense to not include the `left` parameter and if that is the case the FES should try to catch it instead. In addition @Akhilbisht798 has already indicated in the original issue interest to tackle this problem so I would like to give them a chance first.

I would suggest continuing discussion in the original issue to find the right way to solve this, either as a code implementation, documentation, or other fixes if necessary before opening a PR when a solution is reached. I will close this PR for now. Thanks.

PR #7093, opened 2024-06-13
Implemented roll function

### LINKED ISSUE #6760: Implementing a `roll()` function in p5.Camera

rohanjulka19:
Resolves #6760

 Changes:
Addressed the changes suggested by @SableRaf and @davepagurek  mentioned in the issue #6819 for implementation of roll function highlighted below. Now quaternions are used to calculate the new up vector after rotation of camera on the z-axis. 

https://github.com/processing/p5.js/pull/6819#issuecomment-1961624727
> A good approach would be to compute the camera rotation using quaternions, then compute the up and forward vectors from the resulting rotation and apply them to the camera. This demo by scudly is a good example of how to do it: https://infinitefunspace.com/p5/fly/
> 
> I made a minimal example based on it here: https://editor.p5js.org/SableRaf/sketches/WPYb6PRMu

https://github.com/processing/p5.js/pull/6819#issuecomment-1960584109.
> I think the correct output of `roll()` would leave the center and eye the same, but just rotate the up vector. @SableRaf does that make sense, given how much you've been looking into cameras recently?
>

Right now not really sure about the code structure and correctness (still don't completely understand quaternions and their working) that's why haven't added unit tests.

 Screenshots of the change:

https://github.com/processing/p5.js/assets/19673968/4e7cce58-c37d-4c31-82fe-bf4c6c83f860

#### PR Checklist

- [x] `npm run lint` passes


davepagurek (commented):
Thanks for taking this on! I'm also not super familiar with quaternions yet, so to start off I just have some questions about how the method works.

For tests: maybe we could try adding one that does a roll with numbers we can verify by hand, like a 90 degree roll? I think what we would expect to see afterwards is that if you start with a [0, 1, 0] up vector, then after the roll, you'd get something like [1, 0, 0] (maybe with a minus sign in there, the handedness of our vectors always messes me up, haha.)

src/webgl/p5.Camera.js
```diff
@@ -2460,6 +2460,83 @@ p5.Camera = class Camera {
...
+ *
+ * @alt
+ * camera view rotates in counter clockwise direction with vertically stacked boxes in front of it.
+ */
+  roll(amount) {
+    const local = this._getLocalAxes();
+    const axisQuaternion = p5.Quat.fromAxisAngle(
+      this._renderer._pInst._toRadians(amount),
+      local.z[0], local.z[1], local.z[2]);
+    const upQuat = new p5.Quat(0, this.upX, this.upY, this.upZ);
```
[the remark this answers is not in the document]
rohanjulka19: > Would you mind explaining a bit how this works? Previously when I've seen a vector multiplied by a quaternion, it has some slightly different math. In the demo @SableRaf linked, they do it like this:


Yeah, I actually referred to the below link.
https://danceswithcode.net/engineeringnotes/quaternions/quaternions.html

So Initially we have the axis over which the vector has to be rotated and the angle with which the vector has to be rotated. 

Based on this information I used the below formula to get a quaternion for the axis and angle 

Screenshot 2024-06-17 at 5 05 04 PM


And then for rotation first I converted the vector to quaternion and then used the rotation formula (15b) to get the new vector as mentioned in the below figure. 


Screenshot 2024-06-17 at 5 07 39 PM

This is the code in which I have applied the rotation formula (15b)
https://github.com/rohanjulka19/p5.js/blob/fb5289a485bad54046efbdb389bd9d9c5f52fdcc/src/webgl/p5.Quat.js#L46
 
Now I also didn't understood @SableRaf's logic but based on the below stackexchange link it seems like that is a more simplified version of the formula mentioned in the above figure (15b). 

https://gamedev.stackexchange.com/questions/28395/rotating-vector3-by-a-quaternion

So I guess will change the code and use the simplified version instead as it requires less operations. Also will add the test and the above stack link in comments for future reference.

rohanjulka19:
Hi @davepagurek ,
Apologies for the delay in response. 
Added unit tests for roll function and a new function which rotates a vector over a quaternion axis, the code I picked from @sableraf's sketch, the one you mentioned. 
```
rotate( p ) {
    // ( w*w - dot( v, v )) * p + 2 * dot( p, v ) * v + 2 * w * cross( v, p )
    return p5.Vector.mult( p, this.w*this.w - this.v.dot(this.v) )
      .add( p5.Vector.mult( this.v, 2 * p.dot(this.v) ) )
      .add( p5.Vector.mult( this.v, 2 * this.w ).cross( p ) );
  }
```

Also as you highlighted have added the links to the stackexchange answer and the doc I referred to, in the comments.  

Additionaly, I had to add a new function called `clampToZero` in `p5.Vector` class because the result of subtracting floating point numbers is often not exactly zero but very close to zero. I rounded them to zero following the guidelines from 
 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/EPSILON

src/webgl/p5.Camera.js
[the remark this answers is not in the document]
SableRaf: Don't let my "logic" confuse things for you 😅 I really only have very surface knowledge of quaternions.

src/math/p5.Vector.js
```diff
@@ -3887,5 +3887,34 @@ p5.Vector = class {
     }
     return v.equals(v2);
   }
-};
-export default p5.Vector;
+
+
+  /**
+ * ClampToZero check against a <a href="#/p5.Vector">p5.Vector</a>
```
davepagurek, line 3893: Can I suggest:

> Replaces the components of a `p5.Vector` that are <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/EPSILON">very close to zero</a> with zero.
>
> Sometimes, math in JavaScript can give in slightly imprecise answers due to the way numbers are represented. This can make it hard to check if a number is zero, as it may be close but not exactly zero. This method rounds very close numbers to zero to make those checks easier.

src/webgl/p5.Quat.js
```diff
@@ -0,0 +1,90 @@
+import p5 from '../core/main';
+
```
davepagurek, line 2: If this file should be part of the public-facing documentation, can we also add module/submodule information? Example here for `p5.Vector`:

https://github.com/processing/p5.js/blob/eb61f7a260531d17930de3c987949558ce242d35/src/math/p5.Vector.js#L1-L5

davepagurek (commented):
Thanks for adding the tests and comments @rohanjulka19! I think this is looking good.

The last thing I want to maybe iterate on is the documentation. I left a comment on the method for `p5.Vector`, but also can we check is whether, when you run `grunt yui:dev` to launch a local version of the p5.js reference, the new quaternion class is visible in the public reference? If so, we may want to either mark it as private for now or iterate on some of the documentation. (I think it's also fine for now if we want to make it temporarily private, and then open a separate issue for documentation.)

src/math/p5.Vector.js
rohanjulka19: Wow this is so much more clearer

Can I change this part 

> Sometimes, math in JavaScript can give in

to this 

> In computers handling numbers with decimals can give

So the new text will become -

> In computers handling numbers with decimals can give slightly imprecise answers due to the way those numbers are represented. This can make it hard to check if a number is zero, as it may be close but not exactly zero. This method rounds very close numbers to zero to make those checks easier

davepagurek: Sounds good! I'd maybe just add a comma after "In computers".

rohanjulka19:
Ok have made quaternion class private for the time being.

davepagurek (approved):

davepagurek:
@all-contributors please add @rohanjulka19 for code

davepagurek:
@all-contributors please add @haroon10725 for code

PR #7103, opened 2024-06-23
Fix vector reflect() mutating surface normal arg (#7088)

### LINKED ISSUE #7088: p5.Vector reflect() unexpectedly modifies surface normal argument

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [X] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

1.9.4

### Web browser and version

all

### Operating system

all

### Snippet:

```js
function setup() {
	instanceDemo();
	staticDemo();
}

function instanceDemo() {
	const surfaceNormal = createVector(1, 0);
	const incidentVector = createVector(1, 1);
	
	incidentVector.reflect(surfaceNormal);

	//We'd hope surfaceNormal remains [1, 0, 0], but it'll be [2, 0, 0];
	console.log(surfaceNormal.toString());
}

function staticDemo() {
	const surfaceNormal = createVector(1, 0);
	const incidentVector = createVector(1, 1);
	
	p5.Vector.reflect(incidentVector, surfaceNormal);

	//We'd hope surfaceNormal remains [1, 0, 0], but it'll be [2, 0, 0];
	console.log(surfaceNormal.toString());
}

```

### Expected behaviour: 

I'd expect the surface normal argument to be unchanged by the calculation, in both instance and static cases.

### Actual behaviour:

The surface normal argument is mutated during the calculation, in both instance and static cases.

### Misc:

Here's a (simplification of  a) real usage where the bug caught me out: https://openprocessing.org/sketch/2295422

nbogie:
Resolves #7088

 Changes:

Added unit tests to check the reflect functions (static and instance) don't alter the given surfaceNormal vector.

Changed the common p5.Vector reflect function so that the given surfaceNormal vector is not mutated unexpectedly.

Did so by making a local copy of the input vector:

```js
reflect(surfaceNormal) {
    const surfaceNormalCopy = p5.Vector.normalize(surfaceNormal);
    return this.sub(surfaceNormalCopy.mult(2 * this.dot(surfaceNormalCopy)));
}
```

 Screenshots of the change:
 n/a

#### PR Checklist

- [x] `npm run lint` passes
- [ ] [Inline documentation] is included / updated - n/a (bugfix)

Note: eslint passes but emits one warning.
```
p5.js/src/core/reference.js:0:0: File ignored because of a matching ignore pattern. Use "--no-ignore" to override. [Warning]
```

#### misc implementation notes

Also considered this alternative implementation using the static methods to avoid a local copy of the vector, but I think it turns out harder to read, and requires the normalisation calculation happen twice.
 
```js
reflect(surfaceNormal) {
    return this.sub(
      p5.Vector.mult(
        p5.Vector.normalize(surfaceNormal),
        2 * this.dot(p5.Vector.normalize(surfaceNormal))
      )
    );
  }
```

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -2972,8 +2972,8 @@ p5.Vector = class {
  * </div>
  */
   reflect(surfaceNormal) {
-    surfaceNormal.normalize();
-    return this.sub(surfaceNormal.mult(2 * this.dot(surfaceNormal)));
+    const surfaceNormalCopy = p5.Vector.normalize(surfaceNormal);
+    return this.sub(surfaceNormalCopy.mult(2 * this.dot(surfaceNormalCopy)));
   }
 
   /**
--- a/test/unit/math/p5.Vector.js
+++ b/test/unit/math/p5.Vector.js
@@ -1646,6 +1646,21 @@ suite('p5.Vector', function() {
           0.01
         );
       });
+      test('should not update surface normal', function() {
+        const tolerance = 0.001;
+        assert.closeTo(x_normal.x, 3, tolerance);
+        assert.closeTo(x_normal.y, 0, tolerance);
+        assert.closeTo(x_normal.z, 0, tolerance);
+
+        assert.closeTo(y_normal.x, 0, tolerance);
+        assert.closeTo(y_normal.y, 3, tolerance);
+        assert.closeTo(y_normal.z, 0, tolerance);
+
+        assert.closeTo(z_normal.x, 0, tolerance);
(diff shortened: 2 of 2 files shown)


limzykenneth:
Thanks @nbogie. I think having a local copy of the vector is the better approach as it is also more memory efficient.

PR #7110, opened 2024-07-02
Corrects the circle() description

JulioGitLab:
Changes:

 This clarifies that each point on the edge of the circle is half its diameter.

#### PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/src/core/shape/2d_primitives.js
+++ b/src/core/shape/2d_primitives.js
@@ -492,7 +492,8 @@ p5.prototype.ellipse = function(x, y, w, h, detailX) {
  *
  * A circle is a round shape defined by the `x`, `y`, and `d` parameters.
  * `x` and `y` set the location of its center. `d` sets its width and height (diameter).
- * Every point on the circle's edge is half the distance, `d`, from its center (radius = diameter/2).
+ * Every point on the circle's edge is the same distance, `0.5 * d`, from its center.
+ * `0.5 * d` (half the diameter) is the circle's radius.
  * See <a href="#/p5/ellipseMode">ellipseMode()</a> for other ways to set its position.
  *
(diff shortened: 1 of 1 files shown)


src/core/shape/2d_primitives.js
```diff
@@ -490,12 +490,10 @@ p5.prototype.ellipse = function(x, y, w, h, detailX) {
...
  *
- * A circle is a round shape defined by the `x`, `y`, and `d`
- * parameters. `x` and `y` set the location of its center. `d` sets its
- * width and height (diameter). Every point on the circle's edge is the
- * same distance, `d`, from its center. See
- * <a href="#/p5/ellipseMode">ellipseMode()</a> for other ways to set
- * its position.
+ * A circle is a round shape defined by the `x`, `y`, and `d` parameters.
+ * `x` and `y` set the location of its center. `d` sets its width and height (diameter).
+ * Every point on the circle's edge is half the distance, `d`, from its center (radius = diameter/2).
```
nickmcintyre, line 495: Good catch @JulioGitLab! What do you think about the following?

> Every point on the circle's edge is the same distance, `0.5 * d`, from its center. `0.5 * d` (half the diameter) is the circle's radius.

nickmcintyre (changes requested):

src/core/shape/2d_primitives.js
JulioGitLab: Thanks @nickmcintyre! I think your way of explaining it is easier to understand. Let's use it!

nickmcintyre (approved):
Looks good to me!

Qianqianye:
Thank you both! @JulioGitLab @nickmcintyre

PR #7113, opened 2024-07-05
Solves issue #7059

Garima3110:
Resolves https://github.com/processing/p5.js/issues/7059

Changes:
Here's a combined sketch that demonstrates the `worldToScreen` function in both 2D and WebGL contexts.
https://editor.p5js.org/Garima3110/sketches/ilpG7v9Cw

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes


Garima3110:
Please review the code , whenever some maintainer finds the time, Also if you have any idea or some sort of a suggestion, for what example code should i add  for the inline documentation of `worldToScreen()` it would be great!
Thankyou :-)

davepagurek:
Thanks for taking this on, the code looks good!

- Do you think we can add a unit test for 2D and a unit test for 3D? Maybe both a simple case (e.g. no transforms have been applied) and also another that we can verify manually (like a 90 degree rotation)?
- For an example, maybe you could have something like a rotating square (or cube in 3D) where we draw the screen coordinates next to each vertex?

Garima3110:
> Thanks for taking this on, the code looks good!
> 
> * Do you think we can add a unit test for 2D and a unit test for 3D? Maybe both a simple case (e.g. no transforms have been applied) and also another that we can verify manually (like a 90 degree rotation)?
> * For an example, maybe you could have something like a rotating square (or cube in 3D) where we draw the screen coordinates next to each vertex?

Yeah sure! @davepagurek  
I'll try doing that by the end of this week. 
Thankyou.

Garima3110:
@davepagurek Just wanted to know , I tried `npm run build` command for building p5 and run some tests, but for dev2.0 branch the command doesn't work. Is there a different way to build  p5 for dev2.0 branch, unlike the main branch which  builds p5 just with `npm run build`.

limzykenneth:
@Garima3110 `npm run build` should work, what error did you get for it?

Garima3110:
> @Garima3110 `npm run build` should work, what error did you get for it?

image

limzykenneth:
Right I see, currently FES is still relying on the `parameterData.json` file for parameter validation but that file is dynamically generated by the 1.x build via grunt. @Garima3110 Can you switch back to the `main` branch, run the build, then switch to the `dev-2.0` branch to try again? This should generate the `parameterData.json` file that can be packaged by the rollup build.

limzykenneth:
Actually just running `npm run docs` in the `dev-2..0` branch will also generate the parameterData file as @davepagurek have set it up to run the generation script manually.

test/unit/core/environment.js
```diff
@@ -236,4 +236,41 @@ suite('Environment', function() {
...
+
+  suite('3D context test', function() {
+    beforeEach(function() {
+      myp5.createCanvas(100, 100, myp5.WEBGL);
+    });
+
+    test('worldToScreen for 3D context', function() {
+      let worldPos = myp5.createVector(0, 0, 0);
+      let screenPos = myp5.worldToScreen(worldPos);
+      assert.isTrue(screenPos.x >= 0 && screenPos.x <= 100);
```
davepagurek, line 262: Do you think we can assert exact coordinates for the 3D tests like we do for the 2D tests instead of having a range like this? e.g. 50, 50 in this case, and I think 50, 100 in the second case?

test/unit/core/environment.js
```diff
@@ -236,4 +236,41 @@ suite('Environment', function() {
...
+
+  suite('2D context test', function() {
+    beforeEach(function() {
+      myp5.createCanvas(100, 100);
+    });
+
+    test('worldToScreen for 2D context', function() {
+      let worldPos = myp5.createVector(50, 50);
+      let screenPos = myp5.worldToScreen(worldPos);
+      assert.closeTo(screenPos.x, 50, 0.1);
```
davepagurek, line 248: This one is great! Can we add a test of rotation in 2D too?

Garima3110:
I have added more tests , and made some changes according to the suggestions please have a look!

test/unit/core/environment.js
```diff
@@ -236,4 +236,91 @@ suite('Environment', function() {
...
+      myp5.rotateY(myp5.PI / 2);
+      let worldPos = myp5.createVector(50, 0, 0);
+      let screenPos = myp5.worldToScreen(worldPos);
+      myp5.pop();
+      assert.closeTo(screenPos.x, 50, 0.1);
+      assert.closeTo(screenPos.y, 50, 0.1);
+    });
+
+    test('worldToScreen for a rotating cube in 3D', function() {
+      myp5.push();
```
davepagurek, line 305: This one doesn't assert anything so it might not work super well as a unit test, but this would be a cool example to have in the reference for this item! Maybe we could move it into the reference comments instead? (Same for the 2D case, having those as two examples would be great!)

test/unit/core/environment.js
```diff
@@ -236,4 +236,91 @@ suite('Environment', function() {
...
+      let worldPos = myp5.createVector(50, 50);
+      let screenPos = myp5.worldToScreen(worldPos);
+      assert.closeTo(screenPos.x, 50, 0.1);
+      assert.closeTo(screenPos.y, 50, 0.1);
+    });
+
+    test('worldToScreen with rotation in 2D', function() {
+      myp5.push();
+      myp5.translate(50, 50);
+      myp5.rotate(myp5.PI / 2);
```
davepagurek, line 255: I think since we translate before we rotate, the rotation doesn't end up actually affecting the position. If we rotate first, we should see an effect (I would expect (x, y) to become (y, -x) I think?)

test/unit/core/environment.js
davepagurek: @limzykenneth in the 2.0 branch, is there an easy way to get a preview of how the docs will look, similar to `grunt yui:dev` on the main branch?

test/unit/core/environment.js
```diff
@@ -236,4 +236,91 @@ suite('Environment', function() {
...
+    test('worldToScreen for 3D context', function() {
+      let worldPos = myp5.createVector(0, 0, 0);
+      let screenPos = myp5.worldToScreen(worldPos);
+      assert.closeTo(screenPos.x, 50, 0.1);
+      assert.closeTo(screenPos.y, 50, 0.1);
+    });
+
+    test('worldToScreen with rotation in 3D', function() {
+      myp5.push();
+      myp5.rotateY(myp5.PI / 2);
```
davepagurek, line 296: Similar comment here: just rotating won't affect the coordinates, so could we do a translation after this?

davepagurek (commented):
Thanks for making changes, and sorry for the delay! Got sick and then recovered in the mean time, but we're back now!

test/unit/core/environment.js
limzykenneth: Not at the moment. The offline reference (which I also forked out [here](https://github.com/processing/p5.js-offline-reference)) can potentially help but it is not setup to work automatically yet.

test/unit/core/environment.js
davepagurek: Wait, actually, ignore this -- it's only true if you're using (0,0) as your local coordinate. I see the offset by 10 does actually switch axes here!

test/unit/core/environment.js
davepagurek: actually, ignore that, as mentioned above, it should be ok as long as you're converting a nonzero coordinate. But it feels odd that the resulting coordinate would still be 50,50 since the untransformed example above also ends up at that. should this value be something else?

davepagurek: oh, I definitely didn't catch that this was a `rotateY` -- I read it as just `rotate`, which would have been around the z axis. That now explains why the local offset doesn't update the resulting screen x and y values. The 50,50 result now makes sense, since I think the canvas is only 100x100: 
https://github.com/processing/p5.js/blob/420067b182a3f77b564dda4be7b5092e7f3e5970/test/unit/core/environment.js#L242

So I think this result does actually make sense, and maybe we should just add one more test of rotation about the Z axis so that we can confirm that an example like the one you use in the 2D tests also works in WebGL.

src/core/environment.js
```diff
@@ -1152,8 +1152,61 @@ function exitFullscreen() {
...
+ *     createVector(50, -50, -50),
+ *     createVector(50, 50, -50),
+ *     createVector(-50, 50, -50),
+ *     createVector(-50, -50, 50),
+ *     createVector(50, -50, 50),
+ *     createVector(50, 50, 50),
+ *     createVector(-50, 50, 50)
+ *   ];
+ *
+ *   push();
```
davepagurek, line 1195: How do you feel about moving this to `draw()` and making the rotation change slowly over time?

Garima3110: A great idea  , I must say!

src/core/environment.js
```diff
@@ -1137,6 +1137,119 @@ function exitFullscreen() {
...
+ * }
+ *
+ * function draw() {
+ *   background(200);
+ *
+ *   // Animate rotation
+ *   let rotationX = millis() / 1000;
+ *   let rotationY = millis() / 1200;
+ *
+ *   rotateX(rotationX);
```
davepagurek, line 1209: If we want to make sure the text itself isn't rotated, I think we might need to wrap the rotation + worldToScreen calls in a push/pop like you've got in the 2D mode example.

src/core/environment.js
```diff
@@ -1137,6 +1137,119 @@ function exitFullscreen() {
...
+ *   // Convert world coordinates to screen coordinates
+ *   let screenPos = vertices.map(v => worldToScreen(v));
+ *
+ *   // Display screen coordinates
+ *   screenPos.forEach((pos, i) => {
+ *     fill(255);
+ *     noStroke();
+ *     ellipse(pos.x, pos.y, 3, 3); // Draw points as small ellipses
+ *     fill(0);
+ *     text(`(${pos.x.toFixed(1)}, ${pos.y.toFixed(1)})`, pos.x, pos.y);
```
davepagurek, line 1221: Are the screen coordinates relative to the top left? if so we may need to subtract width/2 and height/2 to keep these centered when drawing them in WebGL mode

src/core/environment.js
```diff
@@ -1137,6 +1137,119 @@ function exitFullscreen() {
...
+ *   rotate(PI / 4);
+ *
+ *   // Convert each vertex to screen coordinates
+ *   let screenPos = vertices.map(v => worldToScreen(v));
+ *   pop(); // Restore original drawing state
+ *
+ *   background(200);
+ *
+ *   screenPos.forEach((pos, i) => {
+ *     text(`(${pos.x.toFixed(1)}, ${pos.y.toFixed(1)})`, pos.x, pos.y);
```
davepagurek, line 1177: Looks great!

davepagurek:
Sorry for the delay on this! I recently got the tests passing on the dev-2.0 branch, so if you pull that branch in your fork and merge it into this branch again, we can double-check that we get a green checkmark from the test runner and then merge if it looks good.

Garima3110:
While I continue troubleshooting the failed tests, I'm open to any suggestions you might have, @davepagurek!

src/core/environment.js
```diff
@@ -1146,6 +1146,144 @@ function exitFullscreen() {
...
+    const transformMatrix = new DOMMatrix()
+      .scale(1 / renderer._pInst.pixelDensity())
+      .multiply(renderer.drawingContext.getTransform());
+    const screenCoordinates = transformMatrix.transformPoint(
+      new DOMPoint(worldPosition.x, worldPosition.y)
+    );
+    return new p5.Vector(screenCoordinates.x, screenCoordinates.y);
+  } else {
+    // Handle WebGL context (3D)
+    const cameraCoordinates = renderer.uMVMatrix.multiplyPoint(worldPosition);
```
davepagurek, line 1258: oh, maybe it's this -- we split up `uMVMatrix` into `uModelMatrix` and `uViewMatrix`, and only update `uMVMatrix` right before we draw a shape here: https://github.com/processing/p5.js/blob/4fbbaf514856ec11b771f701dab5dea7e7f14267/src/webgl/p5.Shader.js#L579-L584

Maybe we can refactor the bit that sets `uMVMatrix` into a function on the renderer, like `calculateCombinedMatrix()`, which gets called in `_setMatrixUniforms` above, and then you can also call it here right before you make `cameraCoordinates`?

Garima3110:
Strange, all tests pass as I try this for the main branch but not here in the 2.0 branch.

src/webgl/p5.Shader.js
```diff
@@ -937,7 +937,7 @@ function shader(p5, fn){
       const viewMatrix = this._renderer.states.uViewMatrix;
       const projectionMatrix = this._renderer.states.uPMatrix;
       const modelViewMatrix = (modelMatrix.copy()).mult(viewMatrix);
-      this._renderer.states.uMVMatrix = modelViewMatrix;
+      this._renderer.states.uMVMatrix = this._renderer.states.calculateCombinedMatrix();
```
davepagurek, line 940: I think this method needs to be on the renderer and not the states:
```suggestion
      this._renderer.states.uMVMatrix = this._renderer.calculateCombinedMatrix();
```

src/webgl/p5.RendererGL.js
```diff
@@ -935,6 +935,15 @@ p5.RendererGL = class RendererGL extends Renderer {
     this.clear(_r, _g, _b, _a);
   }
 
+  // Combines the model and view matrices to get the uMVMatrix
+  // This method will be reusable wherever you need to update the combined matrix.
+  calculateCombinedMatrix() {
+    const modelMatrix = this.uModelMatrix;
```
davepagurek, line 941: I think this and the line below would need to access `states` now:
```suggestion
    const modelMatrix = this.states.uModelMatrix;
```

src/webgl/p5.RendererGL.js
```diff
@@ -935,6 +935,15 @@ p5.RendererGL = class RendererGL extends Renderer {
...
+    const viewMatrix = this.uViewMatrix;
```
davepagurek, line 942:
```suggestion
    const viewMatrix = this.states.uViewMatrix;
```

davepagurek (commented):
I noticed a few places where it either wasn't accessing `states` where it should, or was accessing it where it shouldn't, we can see if it works after this. The `states` thing is new in the 2.0 branch so that's probably where the discrepancy is coming from, hopefully once that's fixed itll work again!

Garima3110:
Thanks for the clarification @davepagurek
I’ll work on these suggestions and get back to you soon!

src/core/environment.js
```diff
@@ -1146,6 +1146,128 @@ function exitFullscreen() {
...
+    const screenCoordinates = transformMatrix.transformPoint(
+      new DOMPoint(worldPosition.x, worldPosition.y)
+    );
+    return new p5.Vector(screenCoordinates.x, screenCoordinates.y);
+  } else {
+        // Handle WebGL context (3D)
+        const modelViewMatrix = renderer.calculateCombinedMatrix();
+        const cameraCoordinates = modelViewMatrix.multiplyPoint(worldPosition);
+        const normalizedDeviceCoordinates =
+          renderer.uPMatrix.multiplyAndNormalizePoint(cameraCoordinates);
```
davepagurek, line 1261: Looks like this one is causing the test failures now:
image

I think this one should be in `states` too:
```suggestion
          renderer.states.uPMatrix.multiplyAndNormalizePoint(cameraCoordinates);
```

Garima3110:
@davepagurek I have made some changes in the doc examples too, everything seems to work fine, please let me know if any other change needs to be done.
Thankyou!

PR #7162, opened 2024-08-08
Fix: Added unique group names for radio buttions

### LINKED ISSUE #6763: Radio buttons interact with each other

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [X] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

1.9.0

### Web browser and version

Brave	1.61.116 Chromium: 120.0.6099.217

### Operating system

Windows 11

### Steps:
1. Create two radio buttons (with multiple option)
2. Set a selected value for each
3. Only the last selected value will be selected
4. Selecting a value on the first radio will unselect all options from the second radio

### Snippet:

```js
function setup() {
  createCanvas(100, 100);

  const radio1 = createRadio();
  radio1.option('car');
  radio1.option('bicycle');
  radio1.option('airplane');
  radio1.selected('bicycle');
  
  const radio2 = createRadio();
  radio2.option('tiger');
  radio2.option('elephant');
  radio2.option('parrot');
  radio2.selected('tiger');
}
```

Forchapeatl:
This PR ensures unique names for each radio button group in createRadio function
Resolves #6763 

Resolves #6763 

 Changes:

- Added a Counter for Unique Radio button Names
- Modified the `self._name` Assignment

 Screenshots of the change:

https://github.com/user-attachments/assets/9da27258-29b3-4f2d-9515-0288d4be1305

#### PR Checklist

- [x]  `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -1656,7 +1656,7 @@ p5.prototype.createRadio = function(...args) {
       labelElement = optionEl.parentElement;
     }
 
-    // Check if span element exists, else create it    
+    // Check if span element exists, else create it
     let spanElement;
     if (!isSpanElement(labelElement.lastElementChild)) {
       spanElement = document.createElement('span');
@@ -1665,7 +1665,7 @@ p5.prototype.createRadio = function(...args) {
       spanElement = labelElement.lastElementChild;
     }
 
-    // Set the innerHTML of span element as the label text    
+    // Set the innerHTML of span element as the label text
     spanElement.innerHTML = label === undefined ? value : label;
 
     // Append the label element, which includes option element and
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -1580,6 +1580,9 @@ p5.prototype.createSelect = function(...args) {
  * @method createRadio
  * @return {p5.Element} new <a href="#/p5.Element">p5.Element</a> object.
  */
+
+//counter for unique names on radio button
+let counter = 0;
 p5.prototype.createRadio = function(...args) {
   // Creates a div, adds each option as an individual input inside it.
   // If already given with a containerEl, will search for all input[radio]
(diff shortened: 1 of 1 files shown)


src/dom/dom.js
```diff
@@ -1738,6 +1723,10 @@ p5.prototype.createRadio = function(...args) {
   return self;
 };
 
+// Initialize counter for unique radio group names
+p5.prototype.createRadio.counter = 0;
```
ashish1729, line 1727: I wonder if p5's prototype is a good place to keep this counter, although I am not much aware of the code base to provide an alternative.

Forchapeatl: thank you for the review ,  please where is the prototype class . I will be happy to keep the counter there. I am still learning the structure of the codebase.

ashish1729: I see that we are not adding anything new to the `prototype` of p5 ... we are adding to `createRadio` which is already  there, so I guess this can be OK, you can ignore my previous comment

limzykenneth:
I see many of the existing code comments are removed, is there a reason for that?

src/dom/dom.js
limzykenneth: I'm not sure attaching to `createRadio` as a property is a good idea here. If it is just something that is kept tracked of internally, it can just be a scoped variable, ie. `let counter = 0`.

Forchapeatl: When I use  `let counter = 0`  within the function’s scope, the counter  resets with each call to `createRadio`, and causes duplicate names like `radioOption_0` for multiple radio groups.  When I attach the counter to `p5.prototype.createRadio`, each group gets a unique identifier (e.g., `radioOption_1`, `radioOption_2`), avoiding potential conflicts between radio groups.

limzykenneth: The `let` statement should be outside of the function scope, it can be in the file scope as the bundler won't leak file level local variable to the larger environment.

Forchapeatl:
>I see many of the existing code comments are removed, is there a reason for that?
I have placed  back the comments

limzykenneth:
Looks good. Thanks!

PR #7164, opened 2024-08-08
Fix : removed parent div from check box element

### LINKED ISSUE #5422: createCheckbox() inserts <div>
Using createCheckbox() without specifying a label creates an <input> element wrapped in a <div>. By comparison, createInput() simply inserts an <input> element. For consistency and ease of having to style checkboxes with CSS it would be desireable if createCheckbox() with no label specified behaved like createInput().

- [ ] Color
- [ ] Data
- [X] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [ ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL

#### Details about the bug:

- p5.js version: 1.4.0
- Web browser and version: Google Chrome 93.0.4577.82 (Official version) (x86_64)
- Operating System: MacOSX 10.14.6
- Steps to reproduce this:

function setup() {
  createCanvas(400, 400);
  let s = createCheckbox();
  let f = createInput();
}

function draw() {
}

(examine created <input> elements and their associated HTML code with the browser's developer inspector)

Forchapeatl:
**Before**

image

**After**

image

Resolves #5422 

 Changes:

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -1107,26 +1107,24 @@ p5.prototype.createCheckbox = function(...args) {
   // Create checkbox type input element
   const checkbox = document.createElement('input');
   checkbox.type = 'checkbox';
-
   // Create label element and wrap it around checkbox
   const label = document.createElement('label');
-  label.appendChild(checkbox);
-
-  //checkbox must be wrapped in p5.Element before label so that label appears after
+  //checkbox must be wrapped in label
   const self = addElement(label, this);
-
   // Append the checkbox to the label
   label.appendChild(checkbox);
 
   self.checked = function(...args) {
-    if (args.length === 0) {
-      return cb.checked;
-    } else if (args[0]) {
-      cb.checked = true;
-    } else {
-      cb.checked = false;
+    const cb = self.elt.getElementsByTagName('input')[0];
+    if (cb) {
+      if (args.length === 0) {
+        return cb.checked;
+      } else if (args[0]) {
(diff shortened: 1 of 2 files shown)


limzykenneth:
Sorry for taking so long to get to this and thanks for the PR.

I think this does solve the original problem presented in the issue nicely although I'm thinking whether there should be a larger review of other DOM elements. Specifically the use of `<label>` which in most cases would be beneficial to have, eg. for text input fields.

The other thing is that `createCheckbox()` currently behaves in the same manner as `createRadio()` in that they both wrap the created elements in a `<div>` creating block elements. Should `createRadio()` also have its surrounding `<div>` removed for consistency sake? ie. all DOM elements unless inherently being a block element, should have their default display property. `createRadio()`'s options are also by default inline which may not be what people needed. 

@Qianqianye @davepagurek I think there are some potential breaking changes to the overall DOM API that may be needed here that we may do for 2.0 to get things like form input creation more intuitive and accessible with proper use of `<label>`. 

Overall I think this PR is a breaking change as the `<div>` will never be created so those who were expecting `createCheckbox()` including with a label to create a block display checkbox will have the behavior differ. I'm still unsure how to resolve this with minimal complexity. :thinking:

PR #7199, opened 2024-09-01
Added example and Unit tests for ClampToZero method on p5.Vector.

### LINKED ISSUE #6750: Add missing examples (reference).
### Increasing Access

Adding these examples would result in better understanding of implementation of pre-defined methods in p5.js for beginners.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [X] DOM
- [ ] Events
- [X] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Build Process
- [ ] Unit Testing
- [ ] Internalization
- [ ] Friendly Errors

The added examples are straightforward and easy to follow. These, like other already existing examples for various methods would help in understanding how the p5 methods could be implemented in sketches.

Forchapeatl:
Resolves #6750

 Changes:

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -3901,6 +3901,30 @@ p5.Vector = class {
  * @method clampToZero
  * @return {p5.Vector} with components very close to zero replaced with zero.
  * @chainable
+ * @example
+ * <div">
+ * <code>
+ * function setup() {
+ *   // Create a p5.Vector object.
+ *   let v = p5.Vector.random2D();
+ *
+ *   // Prints "p5.Vector Object : [x, y, 0]" to the console
+ *   // where x and y are small random numbers.
+ *   print(v.toString());
+ * }
+ * function setup() {
+ *   // Create a new vector
+ *   let v = createVector(0.0000000000000002220446049250313, 5 );
+ *
+ *   console.log('Before:', v.x , v.y);
+ *
+ *   // Clamp components close to zero to zero
+ *   v.clampToZero();
+ *   console.log('After:', v.x , v.y);
+ *   describe('Round down very small numbers to zero');
+ * }
+ * </code>
+ * </div>
  */
   clampToZero() {
     this.x = this._clampToZero(this.x);
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -3905,14 +3905,6 @@ p5.Vector = class {
  * <div">
  * <code>
  * function setup() {
- *   // Create a p5.Vector object.
- *   let v = p5.Vector.random2D();
- *
- *   // Prints "p5.Vector Object : [x, y, 0]" to the console
- *   // where x and y are small random numbers.
- *   print(v.toString());
- * }
- * function setup() {
  *   // Create a new vector
  *   let v = createVector(0.0000000000000002220446049250313, 5 );
  *
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -3902,18 +3902,18 @@ p5.Vector = class {
  * @return {p5.Vector} with components very close to zero replaced with zero.
  * @chainable
  * @example
- * <div">
+ * <div>
  * <code>
  * function setup() {
- *   // Create a new vector
- *   let v = createVector(0.0000000000000002220446049250313, 5 );
+ *   // Create a 2D vector where the x-component is near-zero
+ *   let v = createVector(0.00000000000000005, 5 );
  *
  *   console.log('Before:', v.x , v.y);
  *
- *   // Clamp components close to zero to zero
+ *   // Clamp negigabe value of x-component to zero
  *   v.clampToZero();
  *   console.log('After:', v.x , v.y);
- *   describe('Round down very small numbers to zero');
+ *   describe('Round down very small numbers of vector components to zero');
  * }
  * </code>
  * </div>
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -3910,7 +3910,7 @@ p5.Vector = class {
  *
  *   console.log('Before:', v.x , v.y);
  *
- *   // Clamp negigabe value of x-component to zero
+ *   // Clamp negligible value of x-component to zero
  *   v.clampToZero();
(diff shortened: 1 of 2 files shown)


src/math/p5.Vector.js
```diff
@@ -3901,6 +3901,22 @@ p5.Vector = class {
  * @method clampToZero
  * @return {p5.Vector} with components very close to zero replaced with zero.
  * @chainable
+ * @example
+ * <div">
```
ashish1729, line 3905: do we want div> instead of div"> .... in the second one we have extra quotation marks

src/math/p5.Vector.js
```diff
@@ -3901,6 +3901,22 @@ p5.Vector = class {
...
+ * <code>
+ * function setup() {
+ *   // Create a new vector
```
ashish1729, line 3908: this comment can be made more useful,  by saying that we are creating a vector with components having small numerical value

src/math/p5.Vector.js
```diff
@@ -3901,6 +3901,22 @@ p5.Vector = class {
...
+ *   let v = createVector(0.0000000000000002220446049250313, 5 );
```
ashish1729, line 3909: why are we using such a detailed number can't we use 0.000....02 

do we need all those digits ?

src/math/p5.Vector.js
```diff
@@ -3901,6 +3901,22 @@ p5.Vector = class {
...
+ *
+ *   console.log('Before:', v.x , v.y);
+ *
+ *   // Clamp components close to zero to zero
+ *   v.clampToZero();
+ *   console.log('After:', v.x , v.y);
+ *   describe('Round down very small numbers to zero');
```
ashish1729, line 3916: I think this can be improved by saying that we round down the value of components of a vector, 

you can use your own words, basically this function doesn't round down small number, this is for a vector, we can make that clearer

test/unit/math/p5.Vector.js
```diff
@@ -1882,5 +1882,48 @@ suite('p5.Vector', function() {
...
+      });
+
+      test('should leave zero components unchanged', function() {
+        v = new p5.Vector(0, 0);
+        v.clampToZero();
+        expect(v.x).to.equal(0);
+        expect(v.y).to.equal(0);
+      });
+
+      test('should clamp very small numbers in all components of a 3D vector to zero', function() {
```
ashish1729, line 1917: I like how you mentioned that we are clamping "components of a vector", I think it would be awesome if other tests followed the same wording 


ie instead of saying that we are rounding down the number we can say rounding down the component of a vector

src/math/p5.Vector.js
ashish1729: same value is used in tests as well, 

it is ok if this value is kept, you can make that decision

ashish1729:
thank you for fixing all these, although I am not a maintainer, I don't think I can/should approve changes.

Forchapeatl:
No worries @ashish1729 , I really needed someone to put me on track. Thank you.

src/math/p5.Vector.js
```diff
@@ -3901,6 +3901,22 @@ p5.Vector = class {
...
+ * <code>
+ * function setup() {
+ *   // Create a 2D vector where the x-component is near-zero
+ *   let v = createVector(0.00000000000000005, 5 );
+ *
+ *   console.log('Before:', v.x , v.y);
+ *
+ *   // Clamp negligible value of x-component to zero
+ *   v.clampToZero();
+ *   console.log('After:', v.x , v.y);
```
perminder-17, line 3915: If you run `grunt yui:dev`, the user won't be able to see the log values on the reference page. I think they have to open their browser developer tools to see that.

test/unit/math/p5.Vector.js
```diff
@@ -1882,5 +1882,48 @@ suite('p5.Vector', function() {
...
+        expect(v.x).to.equal(0);
+        expect(v.y).to.equal(5);
+      });
+
+      test('should clamp very small negative number of vector components to zero', function() {
+        v = new p5.Vector(-0.0000000000000002, 5);
+        v.clampToZero();
+        expect(v.x).to.equal(0);
+        expect(v.y).to.equal(5);
+      });
```
perminder-17, line 1901: I believe we use ```Math.abs((val || 0) - 0) <= Number.EPSILON ? 0 : val ``` in our clampToZero function, so there's no need to check for negative values. It seems like we're testing whether Math.abs() works correctly rather than testing our p5.js function.

test/unit/math/p5.Vector.js
```diff
@@ -1882,5 +1882,48 @@ suite('p5.Vector', function() {
...
+
+      test('should not clamp regular numbers of vector components', function() {
+        v = new p5.Vector(0.01, 5);
+        v.clampToZero();
+        expect(v.x).to.equal(0.01);
+        expect(v.y).to.equal(5);
+      });
```
perminder-17, line 1908: Here as well, it looks like we are checking for the same test and we are specifically testing for `Number.EPSILON` . What I think we can remove all the tests and can just keep a single big test as you have a test at last. 

something with :
```
        v = new p5.Vector(
          0.00000000000000005,
          -0.0000000000000002220446049250313,
          0.0000000000000002220446049250313);
```
With these three values, we can create different combinations to verify if clampToZero() works correctly. Since unit tests can be time-consuming to run, this would be a kind of optimization.

Let me know what you think ;)

perminder-17:
Keep up the great work. Some thoughts, let me know what you feel. :)

Here how it shows when you run `grunt yui:dev` in your code-editor, you need to open the developer tools in your browser to see the log values.

image


This could be resolved by two solution.

1. If you write the code just as a function, for ex:
```
```js
your sketch....```
```
and by writing some description for eg;
Here's the after result (maybe in your words):
```
```js
before : values
after : values```
```


2. The second option is to use `text()` function and can write the log values into the canvas by making a canvas of 100X100.  

Let me know what you think by these thoughts :)

Forchapeatl:
thanks @perminder-17 , I will work on this over the weekend

PR #7256, opened 2024-09-09
Make shaders define what they get used for.

Garima3110:
This is a work in progress addressing the issue #https://github.com/processing/p5.js/issues/6759

#### PR Checklist

- [x] `npm run lint` passes


perminder-17:
Great job so far, Garima! The idea is that when you call `shader(yourShader)`, it should only apply to fills, regardless of other settings, like the `uStrokeShader` uniform or whether `lights()` is used in the sketch.

Currently, methods like `isLightShader()` 
> https://github.com/processing/p5.js/blob/349b7d482d4fd2270fc19e7737a86292f9c48d59/src/webgl/p5.Shader.js#L983

check which uniforms are defined, and if any are defined, the function returns true. This allows us to apply fill shaders when `lights()` is used in sketches.

One possible solution would be to remove these methods that check whether uniforms and attributes are defined. Instead, we can directly set the appropriate user shader based on the method the user calls, such as `strokeShader(shader)` for stroke shaders.

Also, we should modify the function `_getRetainedFillShader()` written here:

> https://github.com/processing/p5.js/blob/349b7d482d4fd2270fc19e7737a86292f9c48d59/src/webgl/p5.RendererGL.js#L1768

We could probably return the user-defined fill shader if present, or light shader if no fill is set. I guess this should work for fill shaders since `_getLightShader()` is returned when lighting is enabled or a texture is present. That's it for fill shaders, I can leave more reviews for `strokeShaders` and `imageShaders` as we proceed further. Thankyou so much for working on it.

src/webgl/p5.Shader.js
```diff
@@ -1012,6 +1012,10 @@ p5.Shader = class Shader {
     return this.uniforms.uStrokeWeight !== undefined;
   }
 
+  isImageShader() {
```
perminder-17, line 1015: I think we don't need this method because we don't want our `imageShader(yourShader)` method to work depending on some uniforms (`uImageSampler` for your case). We could probably take this one out.

src/webgl/material.js
```diff
@@ -760,21 +756,185 @@ p5.prototype.shader = function (s) {
...
   s.ensureCompiledOnContext(this);
 
-  if (s.isStrokeShader()) {
-    this._renderer.userStrokeShader = s;
-  } else {
-    this._renderer.userFillShader = s;
-    this._renderer._useNormalMaterial = false;
-  }
+  // Always set the shader as a fill shader
+  this._renderer.userFillShader = s;
```
perminder-17, line 760: nice!

Garima3110:
Thanks  for reviewing the code @perminder-17 !
I have made the suggested changes, and the fill shaders seem to be fixed. Please have a look , if everything seems fine I'll work further with the docs and other shaders too.
Thankyou!

src/webgl/p5.RendererGL.js
```diff
@@ -1755,18 +1757,16 @@ p5.RendererGL = class RendererGL extends Renderer {
...
-    } else if (this._tex) {
-      if (!fill || !fill.isTextureShader()) {
-        return this._getLightShader();
-      }
-    } else if (!fill /* || !fill.isColorShader()*/) {
-      return this._getColorShader();
+      return this._getLightShader();
     }
-    return fill;
+    if (this._tex) {
```
perminder-17, line 1766: Can we use the OR operator here and combine it into a single if statement? Something like:
```
if(this._enableLighting || this._tex) 
return this._getLightShader();
```
Can this make the code cleaner instead of using two separate if blocks?

perminder-17:
Great work @Garima3110 so far. I tested it, and it seems to be working well. But to be double sure, I'll test it again to make sure nothing is breaking. In the meantime, you should focus on completing the documentation for the fill shader to meet the first week's target.

The documentation should include: (these are just my thoughts, feel free to add your own explanations if you have any). 

1. A description stating that the fill shader applies only to fills. (explain it in your way)
2. Reference links to other shaders (`strokeShader`, `imageShader`) with a brief explanation, just one line for each of them.
3. 2-3 example sketches for the tutorial, showing how when `lights()` are enabled, it returns `this._getLightShader()` (which is a light shader). When `lights()` are not enabled, it uses the fill shader and returns the fill. Don't go too deep into p5.js's internal system, but give users a basic idea of what’s happening.

Also, talking about the `strokeShaders` here's the method created 
> https://github.com/processing/p5.js/blob/4a2f704e07c4c653661ba420a958d960b9889773/src/webgl/p5.Shader.js#L589

You can remove the method and update the function here as well: (just you did for fill shaders)   
> https://github.com/processing/p5.js/blob/4a2f704e07c4c653661ba420a958d960b9889773/src/webgl/p5.RendererGL.js#L1710

and can move to the docs of `strokeShader()` as well. 

Let me know if `strokeShader()` works for you. 

I think we also need to alter some tests created for shaders, will update you for the tests as we completes the docs and code of this project.

Garima3110:
Thanks for the follow up @perminder-17 
I'll work on these and get back to you soon!

perminder-17:
> Great work @Garima3110 so far. I tested it, and it seems to be working well. But to be double sure, I'll test it again to make sure nothing is breaking. In the meantime, you should focus on completing the documentation for the fill shader to meet the first week's target.

Just an update, previously i mentioned the wrong block of code. I just saw when you pushed it you need to remove the isStrokeShader() code.

src/webgl/p5.Shader.js
```diff
@@ -579,9 +579,6 @@ p5.Shader = class Shader {
     const modelViewProjectionMatrix = modelViewMatrix.copy();
     modelViewProjectionMatrix.mult(projectionMatrix);
 
-    if (this.isStrokeShader()) {
-      this.setUniform('uPerspective', this._renderer._curCamera.useLinePerspective ? 1 : 0);
```
perminder-17, line 583: I think we use `uPerspective` uniform in used in our `line.vert` file here:
> https://github.com/processing/p5.js/blob/4fbbaf514856ec11b771f701dab5dea7e7f14267/src/webgl/shaders/line.vert#L31

So, we could keep them without the if block as well. Otherwise we would not be able to use `uPerspective` unfiorm and could throw errors.

src/webgl/material.js
```diff
@@ -582,11 +582,11 @@ p5.prototype.createFilterShader = function (fragSrc) {
...
  * apply. For example, calling `shader(myShader)` applies `myShader` to
- * process each pixel on the canvas. The shader will be used for:
- * - Fills when a texture is enabled if it includes a uniform `sampler2D`.
- * - Fills when lights are enabled if it includes the attribute `aNormal`, or if it has any of the following uniforms: `uUseLighting`, `uAmbientLightCount`, `uDirectionalLightCount`, `uPointLightCount`, `uAmbientColor`, `uDirectionalDiffuseColors`, `uDirectionalSpecularColors`, `uPointLightLocation`, `uPointLightDiffuseColors`, `uPointLightSpecularColors`, `uLightingDirection`, or `uSpecular`.
- * - Fills whenever there are no lights or textures.
- * - Strokes if it includes the uniform `uStrokeWeight`.
+ * process each pixel on the canvas. This only affects the fill (the inside part of shapes),
+ * not the strokes (outlines).
+ * If you want to apply shaders to strokes or images, use the following methods:
+ * - **[strokeShader()](#/p5/strokeShader)**: Applies a shader to the stroke (outline) of shapes, allowing independent control over the stroke rendering using shaders.
```
perminder-17, line 588: Okay, I was thinking to add these references at the last. 
Like at first we could have our fill shader explanation.
At the middle we could have our example sketches how it works.
After the sketches we could have the references of `strokeShader()` and `imageShader()`. Let me know if that sound's okay to you?

perminder-17:
As we're making breaking changes in this PR, we’ll end up with two distinct methods for applying shaders: `strokeShader()` and `imageShader()`. I'm considering whether we should also rename `shader()` to `fillShader()` for consistency. I'd like to get the WebGL stewards' thoughts on this. Tagging: @davepagurek @lukeplowden @stalgiag @aferriss @aceslowman, @ShenpaiSharma,  @teragramgius, @JazerUCSB, @richardegil, @itsjoopark, @Gaurav-1306, @jeanetteandrews

The advantage of renaming it would be creating a clearer, more defined structure: `fillShader()` for fills, `strokeShader()` for strokes, and imageShader() for images. However, a potential downside is that users have grown accustomed to using `shader()` over the years, especially since fill shaders are used most often. This change might cause some confusion for long-time users. What are your thoughts on this?

Here's how it looks as of now:

[Screencast from 15-09-24 01:13:59 AM IST.webm](https://github.com/user-attachments/assets/0fbbd201-7edc-485b-b9e9-642e8192223a)

src/webgl/material.js
Garima3110: Yeah that too makes sense, I'll do that no worries!

Garima3110:
Thank you for raising this discussion @perminder-17 ! 

Well, I would also like to add something to your suggestion, I see the potential value in renaming `shader()` to `fillShader()` for consistency alongside `strokeShader()` and `imageShader()`. However, I believe it would be better to keep the name `shader()` as it is. My reasoning is that in p5.js, we often retain a feature's basic name based on its default behavior, without explicitly naming it according to that behavior.
The `shader()` function naturally applies to fills by default, and users are familiar with this behavior. Renaming it could introduce unnecessary complexity or confusion, especially for those who have been using p5.js for a long time. By keeping `shader()` intact, we maintain simplicity while introducing `strokeShader()` and `imageShader()` for more specific cases. 

This approach aligns with how we’ve handled similar features, where the core functionality remains intuitively named, while specialized methods are introduced for distinct use cases.
I’d love to hear your thoughts on this approach and whether this strikes a good balance between clarity and consistency.
Thankyou!

lukeplowden:
Looks good @perminder-17 !

I agree with @Garima3110, I don't think `shader()` should be changed to `fillShader()`. It's breaking, but I also think it's better to leave the name general because '`shader()` can often be used for things which aren't really 'fills'. For example you don't have to draw the results of a 'shader()' call directly to the screen inside of 'p5.Framebuffer.draw()`, it could be a post processing pass, or it could be related to instancing using `endShape()`. Eventually they might affect fill colour but I think the general function name is more intuitive.

With `imageShader()` is the idea to affect `image()` calls? I'm curious to see how that one will work!

davepagurek:
I think I also agree with @lukeplowden and @Garima3110. Maybe the way to think about it is not in terms of fills and strokes, but in terms of *materials* and strokes. I suppose fill is actually just one component of an object's material, in addition to other properties like texture or shininess. Strokes are a separate system which adds something on top of the base material. In that view of everything, the material system is the main one you interact with and includes the most stuff, so treating it like the default in its naming feels right to me.

One thing that's not super clean about the model I just described is that images are kind of just a separate system, and regular objects that have a material and stroke can't go through the image system, and likewise, images don't get materials and strokes. I think that's fine, since we want images to always draw an image, so making overriding of image() very separate and explicit probably will help avoid accidental overrides and confusion. So I see image shaders as maybe more of a convenient way for developers to change small bits about image drawing and still take advantage of the fit and fill options of image() rather than a core part of the material/stroke system I guess.

perminder-17:
Thankyou @lukeplowden @davepagurek and @Garima3110 for your thoughts.

> With imageShader() is the idea to affect image() calls? I'm curious to see how that one will work!

We are now planning to create a method that will only affect image calls. In simple terms, we are separating our methods because in many scenarios, we have seen that sometimes fill gets applied to images and sometimes it doesn't, which causes problems and confusion see here: https://github.com/processing/p5.js/issues/6327 , https://github.com/processing/p5.js/issues/6564 . To avoid accidental shaders being applied and also to avoid confusion, we will create a new method called `imageShader()` that will specifically handle shaders for images called through the `image()` function.

I’d also like your thoughts on something related to the question, which may seem obvious but I want to confirm. Do you think the `image()` function should be influenced by `shader()`? Or, put simply, should `shader()` ignore `image()` calls?

perminder-17:
Okay, @Garima3110 currently, the `imageShader` is not working. As we discussed on the call, we need to troubleshoot some functions.             

For example,  if you see the code
https://github.com/processing/p5.js/blob/5fbde13db72c894941db5c92eba919af08ed8132/src/webgl/p5.Shader.js#L1374

we check if `this.samplers.length > 0` so, when it returns true that means we have our image in our shader and we are currently using it (by calling `uniform sampler2D uSampler` and also using it). 

For now, it’s worth disabling the fill shader effect for images, as @davepagurek's inclination was also towards not using fill shader for effecting image. For doing that,  We can update the function https://github.com/processing/p5.js/blob/5fbde13db72c894941db5c92eba919af08ed8132/src/webgl/p5.RendererGL.js#L1768
by adding an extra condition which could handle `isTextureShader()`.

Then, for `imageShader()` we can create a new function to return `_getLightShader()` when `isTextureShader()` is true.

Also, we could add example of by shaderHooks for only `strokeShader()` by studying this interesting tutorial. https://github.com/processing/p5.js/pull/7149#issuecomment-2336404893 .

Garima3110:
Thank you for bringing this up! @perminder-17
> I’d also like your thoughts on something related to the question, which may seem obvious but I want to confirm. Do you think the `image()` function should be influenced by `shader()`? Or, put simply, should `shader()` ignore `image()` calls?

From my perspective, the motivation behind introducing a separate `imageShader()` method is to bring greater clarity and control to the way shaders are applied in p5.js. Specifically, this new method would target images drawn using the `image()` function, thereby avoiding the confusion and unintended behavior that can arise when `shader()` affects image rendering.

Considering the issues mentioned (#6327, #6564), I would say that the `shader()` function should not influence `image()` calls. Instead, `shader()` should be reserved for fills, and `strokeShader()` should handle strokes. Meanwhile, the newly proposed `imageShader()` method, which me and @perminder-17 are currently working on, would exclusively manage shaders, for images drawn with the `image()` function.
I would also like to know what others have to say on this one!
Thankyou.

davepagurek:
Ahh @perminder-17 you're right, those methods are going to clash, so we'll have to rename something. Maybe the hooks ones should be renamed to `defaultStrokeShader()` instead of just `strokeShader()`? It's a little verbose, maybe someone else has a better name idea?

src/webgl/p5.RendererGL.js
```diff
@@ -1781,6 +1780,15 @@ p5.RendererGL = class RendererGL extends Renderer {
...
   }
 
+  _getRetainedImageShader() {
+    const imageShader = this.userImageShader;
+    if (imageShader && imageShader.isTextureShader()) {
+      return this._getLightShader(); // Return light shader if the shader is a texture shader
+    }
+    return imageShader || this._getColorShader(); // Default to color shader if no image shader is defined
+  }
+
```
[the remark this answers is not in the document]
perminder-17: Also, you should rename it to `getImmidiateImageShader`

perminder-17:
> `defaultStrokeShader`

How about `lineShader()` @davepagurek ?

davepagurek:
oops good catch @lukeplowden, I guess if I can confuse the two already then maybe that's a sign they'd be too similar.



> However couldn't `strokeShader()` both set the active stroke shader and return it? Given that the hooks version doesn't take any parameters

Normally when we do that, it's for something that acts as both a setter and a getter, so if we do that here, I'd expect it to return whatever the active stroke shader is, which might be something set by users. We probably will still want a reliable way to get the default one though. Maybe that could mean `defaultStrokeShader()` always returns the default,  `strokeShader()` with no args returns the active stroke shader (or the default at first, when a custom one isn't set), and `strokeShader(yourShader)` would set it?

lukeplowden:
That’s what I was imagining, but then again its also not ideal as it doesn’t align well with the other shader hook functions `materialShader()`, normal etc which can’t be set in the same way. 

`lineShader()` may just be better

davepagurek:
yeah I think if we rename this one we'd probably also want to rename the others to have the same naming convention, `defaultColorShader()` and `defaultMaterialShader()` maybe? Or to be less verbose, `base` instead of `default`, e.g. `baseMaterialShader`?

Garima3110:
Well. I find all three `lineShader()` , `defaultStrokeShader()` and `baseStrokeShader()` suitable for the names, but my  suggestion would be, maybe using `baseStrokeShader` would be a good name here!

perminder-17 (commented):
Maybe I am not technically 100% correct but what I feel is we are separating our shader in two parts (one with without image and one that will only handle image). 

Since there's no method defined for images which you are using in the codebase and it throws errors. Also, there's a need to pass the function which you have made for images here:
> https://github.com/processing/p5.js/blob/e62430728ad5dc89b85eece31ac16415127fec23/src/webgl/p5.RendererGL.Immediate.js#L195

I am not 100% sure what could be the condition for that, but I think you can figure it out. That's the main reason why `imageShader` is still not working for you.

Garima3110:
Thanks @perminder-17 for the insights,I’m working on these changes ,will push the code super soon..!

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -580,4 +580,46 @@ p5.RendererGL.prototype._drawImmediateStroke = function() {
...
+  let shader = this._getImmediateImageShader();
+  this._setImageUniforms(shader);
+
+  for (const buff of this.immediateMode.buffers.image) {
+    buff._prepareBuffer(this.immediateMode.geometry, shader);
+  }
+  shader.disableRemainingAttributes();
+  this._setTexUniforms(shader);
+  this._applyColorBlend(
+    this.curImageColor,
```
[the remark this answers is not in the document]
Garima3110: well I think  If we want separate control over the color of filled shapes and the color (or tint) of images, then keeping `curImageColor` might be necessary.

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -580,4 +580,46 @@ p5.RendererGL.prototype._drawImmediateStroke = function() {
...
+ * enabling all appropriate buffers, binding textures, applying color blend,
+ * and drawing the image geometry.
+ * @private
+ */
+p5.RendererGL.prototype._drawImmediateImage = function(count = 1) {
+  const gl = this.GL;
+  let shader = this._getImmediateImageShader();
+  this._setImageUniforms(shader);
+
+  for (const buff of this.immediateMode.buffers.image) {
```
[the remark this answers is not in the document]
Garima3110: Nice catch it wasn't, I'll define it in the next code push. Thanks!

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -580,4 +580,46 @@ p5.RendererGL.prototype._drawImmediateStroke = function() {
...
+    buff._prepareBuffer(this.immediateMode.geometry, shader);
+  }
+  shader.disableRemainingAttributes();
+  this._setTexUniforms(shader);
+  this._applyColorBlend(
+    this.curImageColor,
+    this.immediateMode.geometry.hasImageTransparency()
```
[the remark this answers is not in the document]
Garima3110: shouldn't we have a separate `hasImageTransparency()`  and  `_setImageUniforms()`  defined to have more control for imageShaders unlike just using the predefined fill methods here as you suggested?!
Or may be for the time being we can just test the working of imageShaders with predefined `hasFillTransparency()` and `_setFillUniforms`...
Let me know your thoughts on this @perminder-17 ?!!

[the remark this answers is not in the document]
perminder-17: Okay, i was thinking that if we are seperating a certian method from the fill shader, we could use the same properties but with different conditions i.e. when we have `imageShader` called.

src/webgl/material.js
```diff
@@ -693,12 +693,8 @@ p5.prototype.createFilterShader = function (fragSrc) {
...
  * The parameter, `s`, is the <a href="#/p5.Shader">p5.Shader</a> object to
  * apply. For example, calling `shader(myShader)` applies `myShader` to
- * process each pixel on the canvas. The shader will be used for:
- * - Fills when a texture is enabled if it includes a uniform `sampler2D`.
- * - Fills when lights are enabled if it includes the attribute `aNormal`, or if it has any of the following uniforms: `uUseLighting`, `uAmbientLightCount`, `uDirectionalLightCount`, `uPointLightCount`, `uAmbientColor`, `uDirectionalDiffuseColors`, `uDirectionalSpecularColors`, `uPointLightLocation`, `uPointLightDiffuseColors`, `uPointLightSpecularColors`, `uLightingDirection`, or `uSpecular`.
- * - Fills whenever there are no lights or textures.
- * - Strokes if it includes the uniform `uStrokeWeight`.
- *
+ * process each pixel on the canvas. This only affects the fill (the inside part of shapes),
+ * not the strokes (outlines).
```
perminder-17, line 697: do you think we can also add "not for `images()`" as well?

src/webgl/material.js
```diff
@@ -718,171 +725,521 @@ p5.prototype.createFilterShader = function (fragSrc) {
...
+ * }
+ * </code>
+ * </div>
  *
- *   // Apply the p5.Shader object.
- *   shader(shaderProgram);
+ * @example
+ * <div modernizr='webgl'>
+ * <code>
+ * // Example 2: Fill Shader with Lights Affected by Mouse
```
perminder-17, line 767: okay, when I say `lights()` I don't mean to say lightPos or something you have made
```
let lightPos = [(mouseX - width / 2) / width, (mouseY - height / 2) / height, 1.0];
 *   fillShader.setUniform('uLightPos', lightPos); // Mouse changes light position
 ```
 Since `shader()` function was not running when we had `lights()` called in our sketch, we need to have certian uniforms or attributes to make that run. So you can use lights() function instead of that.

Comment (perminder-17) [reply src/webgl/material.js:767]: Since this example is mostly the same than the above, you may use lights() in the above example and with a comment that lights() getting ignored as shader is controlling the fill part. For this one, maybe we could have a sketch where we can see `texture` is being effected by shader().

File: src/webgl/material.js [hunk 12]
```diff
@@ -718,171 +725,521 @@ p5.prototype.createFilterShader = function (fragSrc) {
...
  *   plane(100, 100);
- *
- *   describe('A yellow square.');
  * }
  * </code>
  * </div>
  *
- * <div>
+ * @example
+ * <div modernizr='webgl'>
```
perminder-17, line 814: For this one, from my idea, I think you can draw a sphere(with strokes) and with a panorama(img) and you can show that the image on the panorama is not effected, also the strokes of the sphere is also not effected by shader. The only thing effected by `shader()` is the color of the sphere. Let me know what you think.

src/webgl/material.js
perminder-17: Also, I noticed a line "They’re written in a language called GLSL" can we mention  about shaderHooks there as well?

src/webgl/p5.Shader.js
```diff
@@ -1379,6 +1358,7 @@ p5.Shader = class Shader {
     return this.uniforms.uStrokeWeight !== undefined;
```
perminder-17, line 1358: strokeShader was not working in the docs. This was because of two reasons:

1. As I mentioned you need to remove this method created here https://github.com/processing/p5.js/pull/7256#issuecomment-2345767755
2. 2nd reason is since you are using the same name for strokeShader as we have in shaderHooks, it will clash. We will manage it afterwords but for testing purpose it would be nice if you change your method name from strokeShader to something else.

src/webgl/material.js
```diff
@@ -718,171 +725,521 @@ p5.prototype.createFilterShader = function (fragSrc) {
...
- *   // Set the shader uniform p to an array.
- *   mandelbrot.setUniform('p', [-0.74364388703, 0.13182590421]);
+ * This method applies the given shader to strokes, allowing customization of
+ * how lines and outlines are drawn in 3D space. The shader will be used for
+ * strokes until <a href="#/p5/resetShader">resetShader()</a> is called or another
+ * strokeShader is applied.
  *
- *   describe('A fractal image zooms in and out of focus.');
+ * The shader will be used for:
+ * - Strokes only, regardless of whether the uniform `uStrokeWeight` is present.
```
perminder-17, line 885: mmm....that sounds a bit odd. Specially for the new user learning about `strokeShader`, they will think why this is written if they have no idea working with older `shader()` function for strokes. Can we ommit this line with `uStrokeWeight` uniform?

Also, I see you have three examples of `strokeShader`, as I mentioned earlier, we could have one with shaderHooks (you can see the refrence here : https://github.com/processing/p5.js/pull/7149#issuecomment-2336404893 made by @davepagurek )  and one could be the last example you have drawn currently.  Otherwise looks good.

src/webgl/material.js
```diff
@@ -718,171 +725,521 @@ p5.prototype.createFilterShader = function (fragSrc) {
...
+ *
+ * @method strokeShader
```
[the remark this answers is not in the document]
davepagurek: We've renamed the shader hooks method now to be `baseStrokeShader`, but I'm just commenting here to mention that we'll probably need to update the examples for that after this is merged. Currently, the shaders made with it are applied via `shader()`. I think after this change, they will need to be applied with `strokeShader()`.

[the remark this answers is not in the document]
perminder-17: Sure thing! @davepagurek , thanks for the reminder

davepagurek:
btw @Garima3110 feel free to ignore the merge conflicts with the dev-2.0 branch for now. If things are working on this branch, once the project is done I can handle the merge 🙂

src/webgl/material.js
```diff
@@ -718,171 +725,422 @@ p5.prototype.createFilterShader = function (fragSrc) {
...
+p5.prototype.shader = function (s) {
+  this._assert3d('shader');
+  p5._validateParameters('shader', arguments);
+
+  s.ensureCompiledOnContext(this);
+
+  // Always set the shader as a fill shader
+  this._renderer.userFillShader = s;
+  this._renderer._useNormalMaterial = false;
+
```
perminder-17, line 872: I think you can still keep this line 
```js
s.setDefaultUniforms();
```
since it's used by `shaderHooks`. for every entry of `this.hooks.uniforms` it sets the default values. So, we could keep it.

src/webgl/p5.RendererGL.js
```diff
@@ -1756,20 +1758,17 @@ p5.RendererGL = class RendererGL extends Renderer {
...
-      if (!fill || !fill.isLightShader()) {
-        return this._getLightShader();
-      }
-    } else if (this._tex) {
-      if (!fill || !fill.isTextureShader()) {
-        return this._getLightShader();
-      }
-    } else if (!fill /* || !fill.isColorShader()*/) {
-      return this._getColorShader();
+    if(fill && !fill.isTextureShader()){
```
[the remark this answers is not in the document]
perminder-17: We could change the condition to :
```suggestion
    if(fill ){
```
since, `isTextureShader()` is not checking for image. it's done in `_getImmediateFillShader()`

src/webgl/p5.RendererGL.js
```diff
@@ -1781,6 +1776,15 @@ p5.RendererGL = class RendererGL extends Renderer {
...
 
+  _getImmediateImageShader() {
+    const imageShader = this.userImageShader;
+    if(imageShader) {
+      return imageShader;
+    }
+    return this._getImmediateModeShader() ; 
+  }
+  
+
```
perminder-17, line 1787: we could probably make just one method for all three, `_getImmediateImageShader()`, `_getImmediateFillShader()` and `_getRetainedFillShader()` and could name as `_getFillShader()` something like that. 

The only problem could be with the overriding of `imageShader()` with the beginShape/endShape part, so we could probably add a logic here: https://github.com/processing/p5.js/blob/581075435d6cb1e818f19102dc087e2068e75554/src/webgl/3d_primitives.js#L3437
something like:
```js
this._drawingImage = true;
this.beginShape();
// ...
this.endShape();
this._drawingImage = false;
```

and for the `_getFillShader()` I am talking about, it could look something like:
```js
getFillShader() {
  if (this._drawingImage) {
    return this.userImageShader;
  } else if (this.userFillShader) {
    return this.userFillShader;
  } else if ...
```

Really thanks to @davepagurek for his suggestions on this.:)

src/webgl/p5.RendererGL.js
```diff
@@ -627,7 +628,11 @@ p5.RendererGL = class RendererGL extends Renderer {
           new p5.RenderBuffer(3, 'lineTangentsOut', 'lineTangentsOutBuffer', 'aTangentOut', this),
           new p5.RenderBuffer(1, 'lineSides', 'lineSidesBuffer', 'aSide', this)
         ],
-        point: this.GL.createBuffer()
+        point: this.GL.createBuffer(),
+        image: [
+          new p5.RenderBuffer(3, 'imageVertices', 'imageVertexBuffer', 'aPosition', this, this._vToNArray),
+          new p5.RenderBuffer(2, 'imageUVs', 'imageUVBuffer', 'aTexCoord', this, this._flatten)
+        ]
```
perminder-17, line 635: could be reverted, I don't think we need this.

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -579,4 +595,46 @@ p5.RendererGL.prototype._drawImmediateStroke = function() {
   shader.unbindShader();
 };
 
+/**
+ * Called from image rendering functions. Responsible for setting shader uniforms,
+ * enabling all appropriate buffers, binding textures, applying color blend,
+ * and drawing the image geometry.
+ * @private
+ */
+p5.RendererGL.prototype._drawImmediateImage = function(count = 1) {
```
perminder-17, line 604: Could probably remove the whole function. Since the logic with merging it, `this._drawingImage` will take care of that.

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -246,7 +251,18 @@ p5.RendererGL.prototype.endShape = function(
...
+
+      if (this.textureMode === constants.IMAGE && imgShader) {
+        usimageShader(imgShader);
+        this._drawImmediateImage(count);
+      }
+      else {
+        shader(fillShader);
+        this._drawImmediateFill(count);
+      }
+
```
perminder-17, line 265: Could be reverted.

src/webgl/p5.RendererGL.Immediate.js
```diff
@@ -129,6 +129,9 @@ p5.RendererGL.prototype.vertex = function(x, y) {
   );
 
   if (this.textureMode === constants.IMAGE && !this.isProcessingVertices) {
+    const imageShader = this._getImmediateImageShader();
+    if(imageShader){
+      // We have an image shader, normalize UV coordinates based on texture dimensions
     if (this._tex !== null) {
```
perminder-17, line 135: I don't think we need any extra checks, Could be reverted.

perminder-17 (commented):
Thanks to everyone, @Garima3110 and @davepagurek . Things have started coming at a good form. Will request @Garima3110 to add visual tests for all the shaders and could remove/update the previous test where it used the function `isTextureShader()` or `isLightShader()` which we currently don't have, that's the main reason of our tests to fail. Also, we could

Garima3110:
Thanks for all the suggestions to you @perminder-17 
Special thanks to @davepagurek too for his valuable insights !
I'll work on these and aim to finish it super soon!

src/webgl/p5.RendererGL.js
```diff
@@ -1724,52 +1726,26 @@ p5.RendererGL = class RendererGL extends Renderer {
...
    */
-  _getImmediateFillShader() {
-    const fill = this.userFillShader;
-    if (this._useNormalMaterial) {
-      if (!fill || !fill.isNormalShader()) {
-        return this._getNormalShader();
-      }
+   _getFillShader() {
+    if (this._drawingImage) {
+      return this.userImageShader;
```
perminder-17, line 1735: we could probably check if this.userImageShader is defined. If it's not defined we could return this._getLightShader() 

something like, 
```suggestion
if(this.userImageShader){
      return this.userImageShader;
      } else{
      return this._getLightShader();
      }
```

src/webgl/p5.RendererGL.js
```diff
@@ -1724,52 +1726,26 @@ p5.RendererGL = class RendererGL extends Renderer {
...
-      }
-    } else if (this._tex) {
-      if (!fill || !fill.isTextureShader()) {
-        return this._getLightShader();
-      }
-    } else if (!fill /* || !fill.isColorShader()*/) {
+    else if (this._enableLighting || this._tex) {
+      return this._getLightShader();
+    }
+    else {
```
perminder-17, line 1746: could probably remove this else block, it's obvious that it could return `this._getColorShader();` .

perminder-17 (commented):
Looks great so far, we are just on the last bits.

src/webgl/p5.RendererGL.js
perminder-17: we could just write return `this._getColorShader();` without the else block.

test/unit/visual/cases/webgl.js
```diff
@@ -130,4 +130,88 @@ visualSuite('WebGL', function() {
...
+          gl_FragColor = texColor;
+        }
+        `
+      );
+      p5.imageShader(imgShader);
+      p5.texture(img);
+      p5.noStroke();
+      p5.rect(-25, -25, 50, 50);
+      screenshot();
+    });
```
perminder-17, line 214: imageShader() applies to `image()` function and not with texture(img) followed with rect. So, it could be a better check to just see weather imageShader() is not applying to these. But we could also add a test for image(img,0,0) to test weather imageShader probably applies to image or not.

src/webgl/material.js
```diff
@@ -718,171 +725,493 @@ p5.prototype.createFilterShader = function (fragSrc) {
...
+ *   brightness = clamp(brightness, 0.4, 1.0);
+ *   vec3 color = vec3(0.3, 0.5, 1.0);
+ *   color = color * brightness * 3.0;
+ *   gl_FragColor = vec4(color, 1.0);
  * }
  * `;
  *
  * function setup() {
  *   createCanvas(100, 100, WEBGL);
+ *   fillShader = createShader(vertSrc, fragSrc);
```
perminder-17, line 797: I think, we could include a shaderHooks example for fillShader, by removing one example.

perminder-17:
Just a question: Suppose we render something, like a rectangle with a red color, and then apply the `filter(BLUR)` function. Naturally, the result is a slightly blurred red rectangle. Now, if we also use a shader where `gl_fragColor = vec4(green color)`, it would produce a blurred green rectangle.

But here's where it gets interesting: If we use `imageShader()`, it would render a completely red canvas. This happens because we're creating a framebuffer that matches the canvas size here : https://github.com/processing/p5.js/blob/c377b5358ef277949236799512e6274c0bfad678/src/webgl/p5.RendererGL.js#L1057 and then using the `image()` function to render the contents of the framebuffer (FBO) back onto the main canvas or target .  https://github.com/processing/p5.js/blob/c377b5358ef277949236799512e6274c0bfad678/src/webgl/p5.RendererGL.js#L1170 This behavior also impacts `panorama(img)`, likely because it uses filter shaders. So, I am thinking if we could disable `imageShader()` when applying filters? 

 @davepagurek, can you give a thumbs up if you agree?

Garima3110:
Alright @perminder-17 , thankyou for the suggestions, working on these!

Garima3110:
I completely agree with the suggestion made above by @perminder-17 . The key issue lies in how `imageShader()` interacts with the framebuffer, especially when filters like `filter(BLUR)` are applied.

So according to me also, introducing a condition to disable `imageShader()` when filters are being applied, or adding a boolean like `this._drawingFilter` within the filter block in `_getFillShader()`, would help address this issue. It would allow filters to function as expected without interference from shaders, and maintain the correct interaction when both are used together.

src/webgl/p5.RendererGL.js
```diff
@@ -1724,54 +1726,41 @@ p5.RendererGL = class RendererGL extends Renderer {
...
+   * depending on the current context (image or shape).
    */
-  _getImmediateFillShader() {
-    const fill = this.userFillShader;
-    if (this._useNormalMaterial) {
-      if (!fill || !fill.isNormalShader()) {
-        return this._getNormalShader();
+  _getFillShader() {
+    // If a filter like BLUR is applied, disable imageShader
+    if (this._drawingFilter) {
```
perminder-17, line 1735: Have you tested it? I think you need to include `this._drawingFilter` = `false` initally and then in `filter(...args){` you need to make it true so it could catch the filter() calls.

src/webgl/p5.RendererGL.Retained.js
```diff
@@ -133,8 +133,10 @@ p5.RendererGL.prototype.drawBuffers = function(gId) {
     this.retainedMode.geometry[gId].vertexCount > 0
   ) {
     this._useVertexColor = (geometry.model.vertexColors.length > 0);
-    const fillShader = this._getRetainedFillShader();
+
```
perminder-17, line 136: add an extra condition here, 
```js
if (this._drawingFilter){
const fillShader = this.userFillShader;
} else {
 const fillShader = this._getFillShader();
}
```

src/webgl/p5.RendererGL.js
```diff
@@ -1724,54 +1726,41 @@ p5.RendererGL = class RendererGL extends Renderer {
...
+      return this._getLightShader(); // Ignore user-defined imageShader
+    }
+    // If drawing an image, use the image shader or light shader
+    else if (this._drawingImage) {
+      if(this.userImageShader){
```
perminder-17, line 1740: like, instead of adding an extra condition can we just do something like

```js
if(this.userImageShader && !this._drawingFIlter){
// your code
} else{
return this._getLightShader();
}
```

src/webgl/p5.RendererGL.js
```diff
@@ -575,6 +575,7 @@ p5.RendererGL = class RendererGL extends Renderer {
     this.userFillShader = undefined;
     this.userStrokeShader = undefined;
     this.userPointShader = undefined;
+    this.userImageShader = undefined;
```
perminder-17, line 578: above this, where every boolean variables are defined and set to false, you need to set there `this._drawingFilter = false;`

and at the function 
```js
filter(...args) {
// where the image call is present, above that set the 
this._drawingFilter = true;
this._pInst.image(fbo, ....)

Garima3110: Alright @perminder-17 working on these changes.!

Garima3110:
@perminder-17 I have made all the changes, please have a look and let me know if any other thing needs to be updated.
Thankyou!

perminder-17 (approved):
Thankyou so much @Garima3110 for working through everything. I request @davepagurek to give a round of review. 
here's the build : https://editor.p5js.org/aman12345/sketches/fzhPwc8Hr (from the main branch)

PR #7270, opened 2024-09-15
[p5.js 2.0] State machines and renderer refactoring

limzykenneth:
## Renderer state machine
The base `p5.Renderer` class will provide a `states` property that will keep track of all states within the renderer that are affected by `push()` and `pop()`. The base version looks like the following:
```js
this.states = {
      doStroke: true,
      strokeSet: false,
      doFill: true,
      fillSet: false,
      tint: null,
      imageMode: constants.CORNER,
      rectMode: constants.CORNER,
      ellipseMode: constants.CENTER,
      textFont: 'sans-serif',
      textLeading: 15,
      leadingSet: false,
      textSize: 12,
      textAlign: constants.LEFT,
      textBaseline: constants.BASELINE,
      textStyle: constants.NORMAL,
      textWrap: constants.WORD
};
```
Each renderer that extends `p5.Renderer` should add any additional states to the `states` properties when it should be affected by `push()` and `pop()`, eg. `this.states.someState = 'someValue'`. The base implementation of `push()` and `pop()` will keep track of the states with `push()` pushing a copy of the current state into an array then `pop()` restores the last saved state in the array to the renderer's current state. (See line comments for a bit more details below)

The individual renderer should call `super.push()` and `super.pop()` in their own implementation of `push()` and `pop()` in addition to acting on the necessary changes in states. 

## OOP hierachy
The previous hierachy has `p5.Element` as the base class which `p5.Renderer` extends and `p5.Renderer2D/GL` again extends. This means all renderer are assumed to have a `p5.Element` backing. Part of the changes made here is to remove this assumption so `p5.Renderer` can be extended into renderers that does not render to a HTML element.

Eventually, the individual renderers `p5.Renderer2D` and `p5.RendererGL` will have their own reference to `p5.Element` that they can operate through.

Overall event handling is still something that needs to be think through for this case though.

## Global mode
`_setProperty()` is no longer used. Global variables will now all be using getters that refers to the value in the p5 instance.

Pending global functions to be attached in the same way.


src/core/p5.Renderer.js
```diff
@@ -35,69 +40,107 @@ p5.Renderer = class Renderer extends p5.Element {
...
+  id(id) {
+    if (typeof id === 'undefined') {
+      return this.elt.id;
+    }
+
+    this.elt.id = id;
+    this.width = this.elt.offsetWidth;
+    this.height = this.elt.offsetHeight;
+    return this;
+  }
```
limzykenneth, line 103: This will need to be refactored, possibly into individual renderer. It is copied from p5.Element because accessibility features are expecting it.

src/core/p5.Renderer.js
```diff
@@ -35,69 +40,107 @@ p5.Renderer = class Renderer extends p5.Element {
...
-        _textLeading: this._textLeading,
-        _leadingSet: this._leadingSet,
-        _textSize: this._textSize,
-        _textAlign: this._textAlign,
-        _textBaseline: this._textBaseline,
-        _textStyle: this._textStyle,
-        _textWrap: this._textWrap
-      }
-    };
+    const currentStates = Object.assign({}, this.states);
```
limzykenneth, line 109: This should probably be using `structuredClone()` because states may have objects as properties that won't be fully preserved if copied in this way. It is currently not `structuredClone()` because somewhere in the WebGL renderer states is a reference to `window` which cannot be cloned.

limzykenneth:
@davepagurek Following on what we discussed regarding the renderer, I've set up an implementation of the renderer state machine here. For 2D most of the things are working since it is relatively simple, for WebGL however it seems there are some states not being tracked correctly, causing tests to fail.

src/core/p5.Renderer.js
davepagurek: I believe only [some base types](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#javascript_types) work in structured clone, and I'm not certain all the things we put into this state conform to that (e.g. in WebGL, currently we have a camera object.) We have a few options here, do you have any preference between:
- making an interface like `clone()` that state properties can optionally implement so that we can have custom object types in the state
- forcing all state to be a base type and not storing custom classes

I think the former might be a smaller refactor, but that extra flexibility might also mean a higher surface area for bugs. If we go for the latter, for WebGL, I think that would mean storing camera properties rather than the camera itself, and adding a custom handler on `pop()` to reapply those properties to the active camera.

davepagurek: I think for performance we might want the option of not deep copying everything if it isn't necessary, so I maybe slightly lean towards the former?

limzykenneth: Some of the key things not cloneable for our use will be (including nested in objects) symbols, functions, and DOM nodes.

> making an interface like clone() that state properties can optionally implement so that we can have custom object types in the state

Does this mean we will loop through each direct properties of the `state` object and if there is a method `clone()` call it and store that?

> forcing all state to be a base type and not storing custom classes

I'm slightly more in favor of this and I think it is similar to the above option (if I understand it correctly) but with the renderer itself doing the `clone()` step before setting it in the state object.

For something like `p5.Camera` it was never suited for deep cloning anyway as it would be duplicating the underlying p5 instance and what not. Saving the camera properties then reapplying them to the active camera feels more semantic but I'm not sure if there will be any performance impact (through repeated instantiation of heavy object for example)?

Another option is that the base `p5.Renderer.states` object is essentially a convenience in that the push and pop stack will be implemented already in the base class so the derived class don't need to implement it, but if the derived class needs more complex behavior including using non-structuredCloneable states, it can perhaps be handled independently in the derived class's `push()` `pop()` implementation, if that made sense.

davepagurek: > Does this mean we will loop through each direct properties of the `state` object and if there is a method `clone()`, call it and store that?

Yep!


> Another option is that the base `p5.Renderer.states` object is essentially a convenience in that the push and pop stack will be implemented already in the base class so the derived class don't need to implement it, but if the derived class needs more complex behavior including using non-structuredCloneable states, it can perhaps be handled independently in the derived class's `push()` `pop()` implementation, if that made sense.


My comment about perf was mostly about avoiding automatic deep copying of everything, so I'm good with this too if the more expensive clones are done explicitly when needed. So maybe this method would mean that by default, everything is a simple type that should be replaced instead of mutated in place so we can keep doing `Object.assign(...)` instead of a `structuredClone(...)`, and then a renderer can do whatever copy it needs manually in its `push` override? And it should be just a few things needing that custom behaviour.

limzykenneth: If we can further limit state properties to be basically primitives only (copy types only), `Object.assign()` should be enough, it is just when the state properties are objects or arrays then `structuredClone()` would be needed. I'm fine going with the primitives only approach for an overall lighter state management by default if there's no obvious need for reference types to be used as state properties?

If that sounds good, I can try to have a look at WebGL renderer states or since there are other WebGL stuff going on as well, maybe you'd have a better idea of when to review them? I'm going to move back onto the hierachy refactor and instance states next so it's not immediately blocking at this point.

davepagurek: hmm right, I guess we have a lot of array states currently in WebGL as that's the format colors are stored in, and those are all treated as reference types right now. To avoid deep copies on all those, we'd have to wrap them in some other class to avoid the structured clone, so it feels like maybe that'd add more overhead than it's worth. Are there many 2D properties that need a clone? Would relying on object reference for everything by default be a possibility?

I tried doing some quick find-and-replacing this morning to start converting state into WebGL here: https://github.com/limzykenneth/p5.js/compare/2.0-modules...davepagurek:p5.js:2.0-webgl?expand=1 This is currently using the `.copy()` method on things if it exists, but we can change how that's done based on what we decide here, and currently this changeover has broken more tests that will need looking into. But anyway since this is just changing how some properties are accessed, I think it's ok to make these changes whenever, and it won't be too big of a merge for the other contributors.

limzykenneth: For 2D the only one using reference type is `tint` using an array of colors as far as I remember, but it is actually currently on the base p5.Renderer class as WebGL can tint as well. If there is either a guarantee or just enforcement through documentation that if reference types are set as state properties, it must not be separately modified, eg. additional tints creates new array and never reuses it, it would still be fine to just `Object.assign()`.

You should be able to push to this PR directly if you want to do the conversion and test fixing as necessary. I'll merge from dev-2.0 regularly as I work on this too.

src/core/p5.Renderer2D.js
```diff
@@ -17,6 +17,19 @@ class Renderer2D extends Renderer {
...
+    this.wrappedElt = new p5.Element(elt, pInst);
+    for (const p of Object.getOwnPropertyNames(p5.Element.prototype)) {
+      if (p !== 'constructor' && p[0] !== '_') {
+        Object.defineProperty(this, p, {
+          get() {
+            return this.wrappedElt[p];
+          }
+        })
+      }
+    }
```
limzykenneth, line 32: For changing the hierachy of the OOP structure, `p5.Element` is created as a property `wrappedElt` of `p5.Renderer2D` as `p5.Renderer` may not need any DOM nodes. This creates a problem which is that p5.Element methods are expected to be present in the renderer object (returned by `createCanvas()` for example) so that one may call `.id()` or `.mousePressed()` on it.

This change above more or less preserve the existing behavior outwardly by aliasing `p5.Renderer.id()` to `p5.Renderer.wrappedElt.id()` using getter (may not be necessary and can just be directly assigned). This gives `p5.Renderer2D` the appearance of being `p5.Element` but not technically an instance of it.

Another less magical option is to make a breaking change so that the renderer's underlying `p5.Element` can only be accessed through a property of the renderer, and so getting the id of the element backing a renderer would be more like `canvas.element.id()` instead of the current `canvas.id()`. This option has potential to break quite a bit of existing sketches though as it is quite common to do `canvas.parent("#some-div")` to place a canvas in the DOM.

davepagurek: I'm ok with either of those (maybe we could make a `mixin()` function so that other renderers can easily do the same thing if we go with the first option?)

Another thought: do users expect a renderer to come out of `createCanvas` or could that return the element of the renderer? Off the top of my head I can't think of (public) APIs on a renderer itself that aren't just on `p5.Element` that people might be using. I suppose not all renderers have an element to return though (maybe `undefined` is a valid return type here?)

limzykenneth: Hmm...perhaps returning p5.Element for those using it could be something `createCanvas` returns as it aligns with the other `createX()` functions in the DOM module as well. However in this case I guess the created renderer will not be available publicly necessarily then.

davepagurek: Right, I think that's OK if renderers are private if renderers have a way of adding public methods to p5. So things like sphere() could exist just in the WebGL renderer, added to p5

limzykenneth: Yeah, that one is somewhere on my list for this PR, might go with the existing method (defining `p5.prototype.sphere()` that calls `this._renderer.sphere()` or might go with the getter technique here as well. We can discuss when I get to that implementation.

davepagurek:
oops I resolved the merge conflicts with the dev-2.0 updates, but that also brought in some more bits that I need to convert to use states. I'll push another commit soon

src/webgl/p5.RendererGL.js
limzykenneth, line 1: @davepagurek Probably shouldn't update right away but just to let you know in advanced. The refactor for Renderer2D moves all DOM related operation and references to be handled by the renderer's `createCanvas` method along with reshuffling of the initialization logic around default canvas and others. As such WebGL renderer will also need updating accordingly.

I'm working on getting p5.Graphics to work first before I do a round of clean up refactoring.

src/core/p5.Renderer.js
davepagurek: Btw for now I've got it still slicing arrays and calling `.clone()` on classes that support that because there were a lot more instances than I remembered. Still potentially feasible to change, but maybe not the top priority either. The updates to the math library will also affect how exactly cloning matrix properties works, so maybe that would be a good time to revisit and maybe refactor.

limzykenneth:
@davepagurek I don't want to change too much of WebGL at the moment seeing that you and others are also working on it so I can't fix all the WebGL tests. The main failure is likely linked to p5.Framebuffer and possibly also the WebGL filters.

The overall code is still a bit too spaghetti, I'll try to get things a bit more streamlined but have a busy period coming up so I'll see how much I can complete.

I've added some notes below regarding things that eventually should be implemented, let me know if they don't make sense.

* WebGL attributes should be set per renderer instead of per p5 instance: Fixes `p5.RendererGL -> webglVersion -> works on p5.Graphics` unit test
* WebGL filter for 2D canvas not working because the copy helper using the RendererGL `image()` method expects `this._pInst` to be in WebGL mode but since the renderer for the real p5 instance is 2D, it cannot call WebGL methods (eg. `noLights()`. Possible solution is to move these methods to be implemented within p5.RendererGL, with the instance method acting as wrapper of renderer method.
* `p5.Graphics` should not be treated as a valid p5 instance, ie. it should not be used where internal `this._pInst` is expected. Either use its `_renderer` property where necessary or the actual `_pInst` instead.

limzykenneth:
I just thought of a possible idea for p5.Graphics, perhaps I can leverage the new module API where it is a function taking in the signature of `function(p5, fn)` where `fn` is `p5.prototype`, instead I can attach modules and methods to p5.Graphics so that it can become a pseudo p5 instance where necessary without elements that don't make sense.

limzykenneth:
`p5.Graphics` has almost all the needed methods now, with the exception of `blendMode()` and `clearDepth()` (`clearDepth()` may make sense to move to the webgl module, `blendMode()` need a new home that is also included in `p5.Graphics` already).

Also WebGL methods are not attached to p5.Graphics yet as they need to use the new module syntax, @davepagurek do you have an idea of around when would be a good point to do that conversion for WebGL? Probably need to align with other things going on so we don't convert in the middle of someone's work. 

Another thing is that not everything make sense to attach to p5.Graphics, with a general guideline being only renderer related stuff should be attached, other environment related stuff should use the p5 instance version (eg. constants, framerates, data loading, interactions, objects such as p5.Color so probably also p5.Texture etc as well).

davepagurek:
@limzykenneth Luke's project should be wrapping up this week, so that could potentially be a good point to start? But also if we're mostly renaming properties and maybe moving where they're defined, I can resolve any conflicts that come up with Luke's and Garima's projects if this is in the way of more refactor work.

davepagurek:
For what to attach to the graphics prototype, what would be the best way for a module to conditionally attach something? A check if `fn` is equal to `p5.prototype`? We could maybe pass an additional options object with `isGraphics` or something.

limzykenneth:
`isGraphics` or `this instanceof p5.Graphics` can both work depending on what's needed, in some cases it can be refactor into getting members from `p5.Graphics._renderer` or from the `p5` instance instead, but if there are properties that should be in p5.Graphics that are not attached yet, it can be looked into for addition.

davepagurek:
Also once @lukeplowden's and @Garima3110's / @perminder-17's projects are merged, in addition to the renderer changes we're already planning, I'm hoping to do some restructuring because there's a lot of features whose functionality gets spread out over different files. The refactors I'm planning are:
- At a high level, I want all entrypoints into WebGL mode to be in `RendererGL`
  - It will delegate to subsystems in other files as necessary.
  - Contributors should always be able to start in `RendererGL` and then follow the information flow from there rather than trying to find the entrypoint in one of many files.
- Consolidate the rendering code in the retained/immediate files into one system.
  - e.g. in `p5.RendererGL.Immediate.js`, the `_drawImmediateStroke` function has basically the same code as a section of `drawBuffers` in `p5.RendererGL.Retained.js`, and these should really share an implementation because both draw `p5.Geometry` objects at the end of the day, just from different sources.
  - I'm thinking of doing something like `p5.GeometryBuilder`, making a little class that knows about one piece of functionality that the main renderer can call upon. In this case, it would know how to manage and draw buffers.
- Consolidate setting shader uniforms from `p5.RendererGL` and `p5.Shader` into one system
  - e.g. in `p5.RendererGL`, `setFillUniforms` reads renderer state and sets values on a shader. In `p5.Shader`, `_setMatrixUniforms` also reads renderer state and sets uniforms.
  - Like before, this could be encapsulated in an object. I'm thinking of putting all the uniform-setting helpers in one place there, and the renderer would call `shaderManager.prepareFillShader(theShader)` or `prepareStrokeShader` or `prepareImageShader`, and it would handle matrix uniforms and whatever else is needed based on the type.
- Move all shape shape drawing handling to one file
  - Apart from the rendering stuff that I'm already going to move elsewhere, `p5.RendererGL.Immediate.js` handles converting arbitrary polygons and curve calls into triangles, and part of that implementation is in `p5.RendererGL` when it sets up `tessy`, the tessellator library.
  - I think this all can be moved into a single shape manager object
  - @GregStanton's shape drawing refactor will also remove a lot of the implementation details into a common shape drawing module, so the base `Renderer` will be able to build up a shape using this common code. In `endShape`, `RendererGL`'s shape manager will just query the result for polylines of vertices, pass that through the tessellator, and then hand the resulting `p5.Geometry` off to the buffer manager to draw
    - There's also a fair amount of duplicate code right now just interpolating colors along curves, for different types of curves. Luke's project also adds arbitrary numeric attributes in addition to colors. I'm hoping Greg's shape drawing refactor will make it easier for us to do all that interpolation in one place instead of the current 3x duplication of it that we have.
- Move WebGL text rendering in `text.js` into a manager object like the others above

One other thing that I'm not 100% sure about yet is how to fit in a minimal renderer that can be used by 2D mode to run shader filters. Currently 2D mode uses a WebGL `p5.Graphics`, which means importing the entire WebGL module, preventing us from unbundling it for size without also losing the new filter implementations. Ideally, we can make a minimal WebGL renderer that imports just a subset of the subsystems, e.g. skipping the shape drawing subsystem because it will always just be drawing a fullscreen rectangle, skipping the text subsystem, etc.

limzykenneth:
Sounds like a very good plan overall, a few extra points:

> At a high level, I want all entrypoints into WebGL mode to be in RendererGL

One thing that would still be external is the extra p5 instance binding (eg. `box()`) that still needs its own definition attached to `p5.prototype`. These will be the ones that `p5.Graphics` attach to itself to gain possibilities of drawing in 3D.

> Consolidate the rendering code in the retained/immediate files into one system.

I like the idea of a geometry builder class, one thing to consider is whether to attach it to `p5` or not (same with other classes). The case to not attach it would be when it is not expected to be available to the user, ie. only the WebGL renderer use it internally.

> One other thing that I'm not 100% sure about yet is how to fit in a minimal renderer that can be used by 2D mode to run shader filters.

I don't know enough about the system so might be wrong but I guess the parts that are needed are essentially textures and fragment shaders? Since the plan is to split things up into smaller subsystems, if these subsystem can be reused (similar to how the module syntax is reused by `p5.Graphics`) then the filter code or p5.Graphics code or some new internal class (with minimal WebGL context setup) can just import these subsystem independently and leave everything else alone. That way on a build without WebGL, unused code like geometry won't be included; while on full build the same module will only be included once and imported in two places.

davepagurek:
> One thing that would still be external is the extra p5 instance binding (eg. `box()`) that still needs its own definition attached to `p5.prototype`. These will be the ones that `p5.Graphics` attach to itself to gain possibilities of drawing in 3D.

Makes sense! Maybe that can be done in the same file, even if it's done in a different way, so the entrypoints are easier to find.



> I don't know enough about the system so might be wrong but I guess the parts that are needed are essentially textures and fragment shaders? Since the plan is to split things up into smaller subsystems, if these subsystem can be reused (similar to how the module syntax is reused by `p5.Graphics`) then the filter code or p5.Graphics code or some new internal class (with minimal WebGL context setup) can just import these subsystem independently and leave everything else alone. That way on a build without WebGL, unused code like geometry won't be included; while on full build the same module will only be included once and imported in two places.

I think that'll be the plan! I'm not sure yet just how minimal its dependencies can be -- we'll have to strike a balance between using as little as possible, while trying to avoid having duplicate but more minimal implementations of stuff also in other subsystems, which just means more stuff to update and test when we make changes -- so I'll maybe figure that out as we go. As an example, I think it's fine to hardcode a single rectangle into it without relying on the whole shape drawing subsystem, but probably we want the whole shader subsystem to handle all the different ways one might use the filter shader.

limzykenneth:
> Makes sense! Maybe that can be done in the same file, even if it's done in a different way, so the entrypoints are easier to find.

Probably don't recommend having them in the same file as the renderer itself as the primitives are the ones that will be attached to `p5.Graphics` using the module syntax and the renderer itself don't need to be attached.

limzykenneth:
@davepagurek I'm trying to fix framebuffer generated from a webgl p5.Graphics at the moment and since what I'm planning now is for p5.Graphics to be a p5 instance lite but without many of the runtime related stuff, the `p5.Framebuffer` class is trying to access many of the runtime related stuff from the `target` which will be the `p5.Graphics` at the moment and so it won't exist.

I'm wondering if it would be possible to split out the renderer (which `p5.Graphics` will provide) and the runtime (provided with a reference to the current `p5` instance, similar to `pInst` in many classes) to get this to work in the signature of the `p5.Framebuffer` class?

davepagurek:
It might be that p5.Framebuffer only needs the renderer and not the runtime. Right now it looks like it uses `target` directly (as opposed to `target._renderer`) for:
- Getting `webglVersion`. This is specific to the WebGL renderer, so rather than having the runtime own it, we could maybe expose it on the runtime via a getter that accesses something on the renderer so that the renderer doesn't have to reach "up" to the runtime to get it
- In `Framebuffer.begin`/`end`, calling `push`, `setCamera`, `resetMatrix`, and `pop`.
  - `resetMatrix` and `setCamera`, like the above, could be shifted around so that they can be called directly on the renderer
  - `push` and `pop` previously had return values that mattered so they couldn't be called directly on the renderer, but I think that's no longer the case. Is there any difference between calling those methods on the runtime vs on the renderer now? If not, maybe this is also ok to just do on the renderer

This also might be related to something that was the source of a number of bugs in the filter shader implementation, which is that filters were applied using end-user functions like `image()` and `shader()` and such, similar to what framebuffers are doing in `begin`. This lead to some issues where we had to make sure that user state such as `imageMode()` or translations didn't trickle down and affect the filter runner. I think we've currently got it resetting all the state it needs to, but it feels odd to have to reach "up" to the runtime to call these, and it seems brittle. Maybe ideally there are stateless renderer functions that one could call, and runtime functions internally pass whatever necessary state down to the stateless versions on the renderer?

limzykenneth:
I think the two fixes should fix most of the things but I'll try and make it just refers to the renderer, although there may be places where `p5.Framebuffer` itself or something else is looking for a runtime function (I think `color()` is called somewhere but not sure where). `push()` and `pop()`'s implementation are all done in renderer, the runtime just passes it on.

I generally want to keep renderer states with the renderer, I think in the case of something like the filter shader, what we were briefly thinking about before around a minimal WebGL renderer for 2D and minimal 2D renderer for WebGL may be how we solve this? Alternatively and possible simpler, without tanking performance, we can also use the regular renderers we currently have and we instantiate them at the first call then caching them for later use?

limzykenneth:
There's too many blocking stuff because of it so I've refactored all classes to be by default exported. It turns out the vast majority of the don't need reference to `p5` or `p5.prototype` meaning they don't need to be defined in the new library syntax function, which also means they can be exported.

Going forward, all classes will be defined outside of the new library syntax scope, attached to `p5` within that scope, exported and can be imported by other modules.

@davepagurek I've got no errors with testing frame buffer code but it is showing nothing and the tests are rather broken at the moment. I may need to refactor the tests as part of this PR as the sketches I've tested are actually working quite well other than frame buffers.

davepagurek:
How are you testing framebuffers currently? e.g. using them in the sketch in the `preview` directory with `npm run dev`? I can debug that a bit if so, although currently I'm getting some unrelated issues with some lingering references to `p5` in `p5.Geometry` to go through first.

limzykenneth:
Yes I'm using them in `preview` with the `p5.Graphics` `createFramebuffer()` reference example adapted to instance mode.

```js
const sketch = function(p){
  let myBuffer, pg, torusLayer, boxLayer;
  p.setup = function(){
    p.createCanvas(200, 200);
    pg = p.createGraphics(100, 100, p.WEBGL);

    torusLayer = pg.createFramebuffer();
  };

  p.draw = function(){
    drawTorus();
    p.background(200);
    pg.background(50);
    p.image(pg, 0, 0, 100, 100);
  };

  function drawTorus() {
    // Start drawing to the torus p5.Framebuffer.
    torusLayer.begin();

    // Clear the drawing surface.
    pg.clear();

    // Turn on the lights.
    pg.lights();

    // Rotate the coordinate system.
    pg.rotateX(p.frameCount * 0.01);
    pg.rotateY(p.frameCount * 0.01);

    // Style the torus.
    pg.noStroke();

    // Draw the torus.
    pg.torus(20);

    // Start drawing to the torus p5.Framebuffer.
    torusLayer.end();
  }
};
```

davepagurek:
Oh at a glance it looks like you need to draw the framebuffer to the graphic before drawing the graphic back to the main canvas.

I tried adding in `pg.image(torusLayer, 0, 0)` although it looks like I'm now hitting this error:
```
Uncaught (in promise) Error: noLights() is only supported in WEBGL mode. If you'd like to use 3D graphics and WebGL, see  https://p5js.org/examples/form-3d-primitives.html for more information.
    _assert3d p5.RendererGL.js:2414
    noLights light.js:1743
    image 3d_primitives.js:3537
    draw index.html:39
    redraw structure.js:381
    _draw main.js:280
    _draw main.js:302
    _draw main.js:302
    _draw main.js:302
    #_start main.js:208
    p5 main.js:182
    <anonymous> index.html:69
p5.RendererGL.js:2414:12
```
Looks like `this._renderer.isP3D` is undefined when calling `pg.image()` instead of on the main instance

davepagurek:
Also, going back to this:

> I generally want to keep renderer states with the renderer, I think in the case of something like the filter shader, what we were briefly thinking about before around a minimal WebGL renderer for 2D and minimal 2D renderer for WebGL may be how we solve this? Alternatively and possible simpler, without tanking performance, we can also use the regular renderers we currently have and we instantiate them at the first call then caching them for later use?

I was initially suggesting still having all state within the renderer, but doing something like this:
```js
class MyRenderer {
  constructor() {
    this.states = {
      imageMode: 'corner'
    }
  }

  // TODO: better naming
  statelessImage(img, x, y, width, height) {
    // draw the image without modifying internals based on state
  }

  image(img, x, y, width, height) {
    if (imageMode === 'center') {
      x -= width/2
      y -= height/2
    }
    this.statelessImage(img, x, y, width, height)
  }
}
```
...and then we would encourage internals to only use the stateless version.

While a minimal renderer could address this issue, I also wonder if addons are going to have similar issues. e.g. if something is doing some math internally and relies on p5 math functions, they would have to remember to change the angle mode back or the internals will break. I imagine it being easy to run into this problem from [Bret Victor's article on learnable programming](https://worrydream.com/LearnableProgramming/) if you were to try to package a helper function as an addon:
image


Another possible way to address the same problem without this separation could be a way for addons or internals to quickly reset all state except what they care about, e.g.:
```js
class MyRenderer {
  defaultState() {
    return {
      // ...
    }
  }

  constructor() {
    this.states = this.defaultState()
  }

  resetStates(omit) {
    const newStates = this.defaultState()
    for (const key in omit) {
      newStates[key] = this.states[key]
    }
    this.states = newStates;
  }
}
```

limzykenneth:
@davepagurek The `noLights()` error will need a bigger fix in that ideally the implementation for renderer stays in the renderer, eg. `noLights()` only changes the internal states of the renderer so when the renderer wants to turn all lights off, it should just call `this.noLights()` instead of `this._pInst.noLights()`, that way things can be a bit more portable as well espcially if we need to reuse the `noLights()` function for a different renderer. Although to do this quite a lot of `RendererGL` may need to be refactored, I don't want to introduce breaking changes everywhere so want to check when would be a good time to start, also if you have plans for `RendererGL` refactor as well? We can discuss over a call if it's easier.

---

For state management solutions, each have their own trade off I think. I like the stateless version of a function idea but it introduces double implementation for one functionality which can be a maintenance overhead. Now that I think about it, going back to your original idea, what if all the rendering functions in the renderers are stateless, then the p5 instance will handle all state management and also state translation?

For that to work we need a few requirements to be met:
* Renderers (present and future) can all be implemented stateless
  * Not sure if that works for 2D renderer where its states are handled by the 2D drawing context (`this.drawingContext.save()`)
* All renderers can use the same state translation result, eg. `rectMode(CENTER)` translate arguments passed to `rect()` always to the same set of four numbers and all renderers must work with the same set of numbers
* Non renderer states (states not affected by `push/pop()` may also cause the same thing with potential addons for example

Just thinking out loud at the moment and just looking at what I came up with above, it might not be very feasible? It is probably desirerable for addons to call the p5 instance `elllipse()` rather than the renderer's `ellipse()` for example.

Reset states feels like a wrapper around what we currently do as it will need to be wrapped in `push/pop()` as well and don't feel as efficient. 

I probably need to think a bit (while resisting the urge to rewrite...) 🤔

davepagurek:
I was thinking of tackling the things in https://github.com/processing/p5.js/pull/7270#issuecomment-2380656539 after this PR and Garima's PR are merged in.

For the `this._pInst.something` --> `this.something` change, I wonder if that's something we could do with a find-and-replace just in the webgl folder? I ran `find src/webgl -name "*.js" | xargs egrep 'this\._pInst\.\w+\(\)'` and it looks like we call `createFramebuffer`, `push`, `pop`, `clear`, `clearDepth`, `noStroke`, `noLights`, `resetMatrix`, `resetShader`, and `pixelDensity`. Are all of those things that we would be (or should be) able to call directly on the renderer?

<hr />

For renderer statelessness, it does seem like it'd be a big refactor to fully pull things apart, and you're right that addon authors and contributors will probably be more familiar with the public APIs, so it'd be good to let them just use those.

Generally not all methods need it anyway (e.g. for `sphere` theres usually no need to explicitly pass in all the material properties, its pretty unlikely that you actually want a default-state sphere.) It might be lighter to do a way to reset specific states within a push/pop, and shallow copying just the state you need in? e.g.

```js
defaultStates() {
  return { ... }
}

constructor() {
  this._defaultStates = this.defaultStates()
  this.states = this.defaultStates()
}

runStateless(keys, cb) {
  this.push()
  for (const key in keys) {
    this.states[key] = this._defaultStates[key] // No deep copy here because it won't modify it anyway
  }
  cb()
  this.pop()
}

filter(filterType) {
  this.runStateless(['imageMode', 'rectMode', 'userFillShader'], () => {
    // do the filter
  })
}
```

limzykenneth:
> For the this._pInst.something --> this.something change, I wonder if that's something we could do with a find-and-replace just in the webgl folder? I ran find src/webgl -name "*.js" | xargs egrep 'this\._pInst\.\w+\(\)' and it looks like we call createFramebuffer, push, pop, clear, clearDepth, noStroke, noLights, resetMatrix, resetShader, and pixelDensity. Are all of those things that we would be (or should be) able to call directly on the renderer?

Many of them don't have the implementation on the renderer so we basically need to move the implementation over so just a find and replace might not work fully. I'll do a bit of clean up and refactor first to see how far I can take it.

---

I think a selective resetable state might be the way, although I want to think a bit more about the API, eg. the default states might make more sense statically attached to the renderer (or not).

limzykenneth:
@davepagurek I think I got graphics frame buffer working! Have a try and see if it works for you as well?

davepagurek:
Looks like it's mostly working now!

One thing I notice is that something might be a bit off with the cameras. I'm using a webgl main canvas and this test sketch:
```js
const sketch = function(p){
  let myBuffer, pg, torusLayer, boxLayer;
  p.setup = function(){
    p.createCanvas(200, 200, p.WEBGL);
    pg = p.createGraphics(100, 100, p.WEBGL);

    torusLayer = pg.createFramebuffer();
  };

  p.draw = function(){
    drawTorus();
    p.background(200);
    pg.background(50);
    pg.imageMode(pg.CENTER);
    pg.image(torusLayer, 0, 0);
    p.imageMode(p.CENTER)
    p.image(pg, 0, 0, 100, 100);
  };

  function drawTorus() {
    // Start drawing to the torus p5.Framebuffer.
    torusLayer.begin();

    // Clear the drawing surface.
    pg.clear();

    // Turn on the lights.
    pg.lights();

    // Rotate the coordinate system.
    pg.rotateX(p.frameCount * 0.01);
    pg.rotateY(p.frameCount * 0.01);

    // Style the torus.
    pg.noStroke();

    pg.box(20);

    // Start drawing to the torus p5.Framebuffer.
    torusLayer.end();
  }
};

new p5(sketch);
```

The 2.0 branch looks like this:
image

While [on 1.11.0](https://editor.p5js.org/davepagurek/sketches/G1Ni9l2rd) it looks like this, which is expected:
image

I'm not 100% sure what's up yet. If the graphic is the same size as the main canvas, the box is still in the lower right corner, so it's not that the graphic is using the main canvas's viewport size. Drawing directly to the graphic is fine without the intermediate framebuffer. Drawing to a framebuffer on the main canvas is fine too. So it seems like there's something about framebuffers on graphics that still has an issue.

limzykenneth:
Might be a pixel density mismatch, I'll look into it.

src/webgl/p5.Framebuffer.js
```diff
@@ -6,1622 +6,1636 @@
...
     }
-  };
+
+    this._recreateTextures();
+
+    const prevCam = this.target._renderer.states.curCamera;
+    this.defaultCamera = this.createCamera();
+    this.filterCamera = this.createCamera();
+    this.target._renderer.states.curCamera = prevCam;
+
```
limzykenneth, line 130: @davepagurek I think the problem we are having is likely around here along with the `createCamera` method below. Notably the camera that is created for the graphics after its creation is different from what it was after the framebuffer is created, my guess is that its a reference issue, perhaps the `FramebufferCamera` class needs a `.clone()` implementation?

limzykenneth:
Actually I think it might be `imageMode` not working.

davepagurek:
Ahhhh that makes sense, good catch!

I think everything's fine then. I've updated https://github.com/processing/p5.js/issues/7230 to add a line about adding constants to `p5.Graphics` in the backwards compatibility addon.

limzykenneth:
@davepagurek I think this is about as much as I can do to fix tests without doing the larger WebGL refactor, I'll see if there are other parts to work on first but do let me know when we are ready to start refactoring RendererGL.

davepagurek:
Garima's project is merged into `dev-2.0` now so I think once those changes are merged into this branch too then I'm good to get started on the refactoring! I can also look into doing that merge manually if things have already moved around too much for git to recognize where things should go.

limzykenneth:
I had a look at the merge conflicts and I'm not immediately confident about resolving it, not sure if it will be too complicated if you were to do the conflict resolution along with some of the refactoring? We can do a session together at the start so that we cover both set of changes in case either of us missed anything if it helps?

davepagurek:
I think most of the merge conflicts were because the indentation changed (e.g. moving some things into a function, or moving some things *out* of a function) so I tried just adjusting the indentation of the conflicting files to match, merging, and then putting the indentation back. That made the vast majority of the conflicts go away! So I'll probably do that again if we have more upstream changes in the future.

Anyway, I think this should be good to go now.

limzykenneth:
Most tests are passing now and the ones that are failing are all either visual test or pixel comparison test, I'm not entirely sure if they are false negative or there are some visual edge cases I haven't tried in a test sketch though.

davepagurek:
All the visual tests used to pass on CI here so unless we change how we do the rendering, I think it should be possible to get them green. Looks like something is still up, e.g. one of the visual tests expected this but received a blank canvas:
image

Another visual test seems to be mostly the same, except the expected result has an opaque background, but the received one has a transparent background:
image

vs 

image

Maybe something about `background()` changed for WebGL?

limzykenneth:
@davepagurek For the first case it seems like the `GRAY` filter is not working correctly, that part did get a lot of refactor around moving things to be handled by the renderer directly, not sure if I missed something?

For the second case it is weirder as when I copied the code from the visual test to run locally I don't see a diff there, ie. the background is gray.

```
const sketch = function (p) {
      p.setup = async function () {
        // p.createCanvas(200, 200, p.WEBGL);
        p.noLoop();
        p.createCanvas(50, 50, p.WEBGL);
        // p.noStroke();
        // p.fill('red');
        // p.circle(0, 0, 20);
        // p.filter(p.GRAY);
        // p.circle(10, 10, 20);

        const tex = await p.loadImage('./cat.jpg');

        p.translate(-p.width / 2, -p.height / 2);
        p.background(200);
        p.fill(255);
        p.stroke(0);

        p.texture(tex);
        p.beginShape(p.QUAD_STRIP);
        p.vertex(10, 10, 0, 0, 0);
        p.vertex(45, 5, 0, tex.width, 0);
        p.vertex(15, 35, 0, 0, tex.height);
        p.vertex(40, 45, 0, tex.width, tex.height);
        p.endShape();
      };
};
```

src/webgl/p5.RendererGL.js
```diff
@@ -1110,58 +841,63 @@ p5.RendererGL = class RendererGL extends Renderer {
...
     this._drawingFilter = true;
-    this._pInst.image(fbo, -target.width / 2, -target.height / 2,
-      target.width, target.height);
+    this.image(
+      fbo,
+      0, 0,
+      this.width, this.height,
+      -target.width / 2, -target.height / 2,
+      target.width, target.height
+    );
```
limzykenneth, line 896: @davepagurek Perhaps I got this signature wrong?

limzykenneth: Thanks, I had a look but still couldn't figure out where the problem is. It might be in how the filter is applied.

davepagurek: I'll take a look tonight and see if I can figure out what's going on!

davepagurek:
I fixed a few of the WebGL issues:
- `ensureCompiledOnContext()` was part of the user-facing `shader()` but not the renderer's internal `shader()` method so the filter shaders didn't have all the expected metadata. I've moved that call into the internal version, which seemed to fix the filter shader tests.
- The issue with the background being transparent instead of opaque seems to have something to do with `loadImage` being async, so the `background()` from earlier no longer seems to have applied. Still not sure yet exactly what was clearing it, but just loading images before setting up the canvas fixed that test.

There are still some more test failures I'm going to look at, but one interesting one is this:
https://github.com/processing/p5.js/blob/0f444136e36f68e0a747d519177ff4cd0890431d/test/unit/webgl/p5.RendererGL.js#L324-L335

When I was debugging this, I found that `pg._renderer._pixelsState === myp5._renderer._pixelsState`. I think the `pInst` we pass in is now always (as the name implies) the p5 instance everything is part of, which makes sense, although I think that means we need to change a bit about how `_pixelsState` works. For the renderer of a `p5.Graphics`, I think the `_pixelsState` needs to refer to the `Graphics` object instead. Not sure the best way to pass that in just yet.

test/unit/webgl/p5.RendererGL.js
```diff
@@ -34,7 +34,8 @@ suite('p5.RendererGL', function() {
       assert.equal(myp5.webglVersion, myp5.WEBGL);
     });
 
-    test('works on p5.Graphics', function() {
+    // NOTE: should graphics always create WebGL2 canvas?
```
davepagurek, line 37: I think the intended behaviour originally was like, graphics have their own `setAttributes()`, so you would make them be webgl 1 manually if you want it to be webgl 1, otherwise it'll try to be webgl 2 by default if it's available. I'd also be ok defaulting it to whatever the main canvas does too, since this is already very much an edge case.

davepagurek:
took a while to find the cause of the masking bugs, but the cause of that one was the ordering of `_setAttributeDefaults` and `_initContext`. The default attributes have to be set *before* initializing the context in order for us to pass the right values into `canvas.getContext()`. Without that, it falls back on the browser defaults, which does not include a stencil buffer (while p5 asks for one by default to make clipping work.) I don't think moving it back to be before `_initContext` failed any new tests, but let me know if there are other reasons for having it lower down in initialization and I can think about how to move around the necessary defaults another way.

It turned out to not be the underlying cause, but I also refactored `fill` and `stroke` a bit to have them also set `states.doFill` and such. No tests were broken because of that, but it might be a cause of a future subtle bug if internal `fill` doesn't enable filling while external `fill` does.

test/unit/webgl/p5.RendererGL.js
limzykenneth: Ok, I see. I was a bit confused because it seemed like the main canvas and the graphics have different behavior dealing with the constants but I think I just missed the setAttributes. I think it's good to have it be different from the main canvas for more flexibility.

The test was failing for this case though so may need to check it later.

limzykenneth:
I don't think `_initContext` was moved specifically but if it needs to be somewhere else feel free to move it.

davepagurek:
Sounds good, I've moved it earlier so I'll just keep it there.

What do you think about having a separate `pixelsState` between the main canvas and graphics? In addition to `pInst`, should we add an argument to the `Renderer` constructor to also give it a pixel state object (the same as `pInst` for the main canvas; the graphic for a graphic?)

davepagurek:
Right now it acts kind of like an interface that anything that supports loadPixels conforms to to be able to share a 2D loadPixels implementation. So right now it's kind of like this:
```typescript
interface SupportsPixels {
  pixelsState: {
    _pixelDensity: number
    width: number
    height: number
    drawingContext: CanvasRenderingContext2D
    pixels: Uint8ClampedArray

    imageData?: ImageData // Cached image data after loadPixels called
  }
}
```

And then things that have this reuse the 2D renderer's `loadPixels`, `updatePixels`, `set()`, and `get()`. e.g. `p5.Image`:
https://github.com/processing/p5.js/blob/60dbf864614137c03c7826f065f7078affa3344b/src/image/p5.Image.js#L457-L460

So if we want to keep using that pattern, `pixelsState` will be different between the main canvas and a graphic. Previously this was also the case, `pInst` in a graphic confusingly refers to something other than the main instance, which is what necessitated the weird gymnastics in the filter rendering:
https://github.com/processing/p5.js/blob/60dbf864614137c03c7826f065f7078affa3344b/src/core/p5.Renderer2D.js#L21-L28

...but now, I think `pInst` always refers to the main instance? Which I think makes more sense for most things overall, but does break the assumptions used in `pixelsState`. It might be time for a new design for that anyway since 2D and WebGL things implement it differently anyway.

Since graphics and the main canvas both have a `Renderer`, these could just all be on the renderer and not in a separate `pixelsState` object and that would be simpler. The slight complication is that `p5.Image` also wants to have these methods, and it doesn't have a renderer. Maybe we can split that functionality out into a class `PixelsState2D`?

limzykenneth:
The `pixelState` interface basically are all covered by the 2D renderer and calling it there will be easier than using the instance. For anything called `pInst` going forward, it should refer to the actual p5 instance for consistency and simplicity, and also it centralizes the instance states like color mode.

For p5.Image it has most of the `pixelState` interface implemented as well, if the implementation of things like `loadPixel()` don't differ with the 2D renderer's implementation, we can consider making them mixins for the different class to share code as long as they themselves have `pixelState` interface.

For p5 instance, it won't need to implement `pixelState` as it will call the renderer's `loadPixel()` etc when called on the instance, the same as how we handle the rest of the renderer's interface.

davepagurek:
I think one thing we'll need is a way for the renderer to "reach out" to the thing that owns it to set its `pixels` property. I see that p5.Graphics adds getters for all non-private renderer properties, which means we can just have the renderer own `pixels` if the main instance makes getters too.

I've added that, and caught a few cases where the renderer had a property with the same name as a main instance function (e.g. `textureMode`) that I had to change (`textureMode` should really be in `states` anyway), and a few cases in dom.js where `this.elt` was being assigned to the last created element in `createSelect()` and a few others (I don't think it was being used anyway and has been in the codebase forever it seems, so I just removed it.)

There are a few tests still failing that have to do with `updatePixels` in WebGL not working so I'll take a look at those next.

davepagurek:
Actually, this approach works for now but it means a lot more things are now leaked into the global scope that probably shouldn't be, e.g. `this.zoomVelocity`, `this.prevTouches`, etc in `RendererGL`. I guess that was already true on `p5.Graphics` but it matters less there since people aren't usually defining their own properties on those, whereas stuff on `p5` becomes global.

I'm going to just make it expose `pixels` for now.

limzykenneth:
I don't know how much of the renderer we need to expose to the instance beyond `pixels`, but if possible and there aren't too many, we should manually expose them.

Basically what I think would be nice is that global variables like `width`, `height`, `pixels` are all getters to the renderer's own `width`, `height`, `pixels` etc but manually expose them instead of looping though all the renderer's properties. ie. Remove the need for the renderer to need to reach to the instance as much as possible and let the renderer handle changes to its own state, which will then reflect into the global state through the getters.

src/webgl/p5.RendererGL.js
```diff
@@ -1389,27 +1122,27 @@ p5.RendererGL = class RendererGL extends Renderer {
...
-    this._pInst.push();
-    this._pInst.resetMatrix();
-    this._pInst.clear();
-    this._pInst.imageMode(constants.CENTER);
-    this._pInst.image(fbo, 0, 0);
-    this._pInst.pop();
+    this.push();
+    this.resetMatrix();
+    this.clear();
+    this.states.imageMode = constants.CORNER;
```
davepagurek, line 1130: Turns out the updatePixels problems were because setting the imageMode to CENTER does nothing when calling `this.image` as opposed to the user-exposed `image()`. I think that's fine, and actually probably for the best that our internal implementations depend less on state, but we should watch out for more cases like this left over from before.

davepagurek: Technically we also don't need this line since it'll use corner mode regardless. Might be worth keeping for clarity, although then it's misleading because it looks like something that you can change when you actually can't. Maybe I should remove it?

limzykenneth: Yeah, that should be removed if it's doing nothing, I kept it around because I didn't know if it was actually needed or not and I just translated from what was before.

davepagurek:
Ok last webgl test was because it was accessing a constant from a graphic, which is now undefined. Tests are all green on my end now!

davepagurek:
> Basically what I think would be nice is that global variables like `width`, `height`, `pixels` are all getters to the renderer's own `width`, `height`, `pixels` etc but manually expose them instead of looping though all the renderer's properties. ie. Remove the need for the renderer to need to reach to the instance as much as possible and let the renderer handle changes to its own state, which will then reflect into the global state through the getters.

Do you think there would be a case where a new renderer, e.g. SVG, would want to add new state visible to the end user? If so, it might make sense to make a way for the renderer itself to define what gets exposed (could be as simple as doing `this.exposedProperties = ['width', 'height', 'pixels']` and then having the renderer loop over those.)

limzykenneth:
Two failed on my end that looks like single pixel and off by one error
```
p5.RendererGL
default stroke shader
coplanar strokes match 2D
AssertionError: expected [ +0, +0, +0, 255 ] to deeply equal [ +0, +0, 255, 255 ]
 - /test/unit/webgl/p5.RendererGL.js:123:14
- Expected
+ Received

  Array [
    0,
    0,
-   255,
+   0,
    255,
  ]
color interpolation
bezierVertex() should interpolate curFillColor
AssertionError: expected [ 255, 128, 128, 255 ] to deeply equal [ 255, 129, 129, 255 ]
 - /test/unit/webgl/p5.RendererGL.js:1900:14
- Expected
+ Received

  Array [
    255,
-   129,
-   129,
+   128,
+   128,
    255,
  ]
```

If you can do a rebase and push it should run here on CI then we can check here.

limzykenneth:
> Do you think there would be a case where a new renderer, e.g. SVG, would want to add new state visible to the end user? If so, it might make sense to make a way for the renderer itself to define what gets exposed (could be as simple as doing this.exposedProperties = ['width', 'height', 'pixels'] and then having the renderer loop over those.)

That could be a thing but probably need to be a slightly more elaborate API as the renderer implementer might want to set both the getter and setter.

limzykenneth:
Rebasing now

limzykenneth:
@davepagurek All tests passed in CI!

Shall we merge this in then you can start working on additional WebGL refactoring while I work on some more details stuff to make internal API clearer?

davepagurek:
Awesome! Sounds good, thanks for doing that rebase!

PR #7277, opened 2024-09-20
[p5.js 2.0] Vector nD getter and setter, added values.

holomorfo:
Resolves ##6765 

 Changes:
Modify Vector class to include a `values` list to make it n-dimensional while mantaining the existing API of `x`, `y`, `z` by setting up accessors, getter and setter.
Pending Update documentation and add new Unit tests

 Screenshots of the change:

#### PR Checklist

- [X] `npm run lint` passes
- [x] [Unit tests] are included / updated **(Didn't modify any existing test, added new tests to increase coverage of the Vector file from 94.59% to 97.12%)**

CC: @limzykenneth @davepagurek @Qianqianye


src/math/p5.Vector.js
```diff
@@ -118,23 +127,63 @@ function vector(p5, fn) {
...
+
+    get values() {
+      return this._values
+    }
+
+    set values(newValues) {
+      this._values = newValues.slice()
+    }
+
+    get x() {
```
holomorfo, line 156: this allows us to still consume the vector x, y and z without having an explicit class attributes with that name, this making it more flexible
@limzykenneth

src/math/p5.Vector.js
```diff
@@ -118,23 +127,63 @@ function vector(p5, fn) {
...
+        );
       }
-      this.x = x;
-      this.y = y;
-      this.z = z;
+    }
+    get y() {
+      return this._values[1] || 0
+    }
+    get z() {
```
holomorfo, line 173: should we add a `w` value for 4D? or for nD do users will just search values by number?
@limzykenneth

src/math/p5.Vector.js
```diff
@@ -118,23 +127,63 @@ function vector(p5, fn) {
...
+
+    set values(newValues) {
+      this._values = newValues.slice()
+    }
+
+    get x() {
+      return this._values[0] || 0
+    }
+
+    setValue(index, value) {
```
holomorfo, line 160: probably I need to create also a `get value(index)` so users can get entries of the vector by ID
@limzykenneth

src/math/p5.Vector.js
davepagurek: I'm ok adding a w personally, since math for 3D stuff uses 4D vectors/matrices quite often. it seems fairly common in other apis too, e.g. native js DOMPoint. And then anything above 4D would use indices

limzykenneth: Yeah, adding a `w` is fine for me as well. When working on the documentation we need to balance this convention of using `w` and it potentially being unintuitive for those very new to linear algebra, possibly by de-emphaiszing `w` in some way or other solution.

src/math/p5.Vector.js
```diff
@@ -158,7 +207,7 @@ function vector(p5, fn) {
      * </div>
      */
     toString() {
-      return `p5.Vector Object : [${this.x}, ${this.y}, ${this.z}]`;
+      return `p5.Vector Object : [${this.values.join(', ')}]`;
```
limzykenneth, line 210: One of the things I'm considering is to have this be serialized into something more standard/machine readable, so maybe just the JSON part at the end.

holomorfo: do you mean something like this?
```
    toString() {
      return `${this.values.join(', ')}]`;
    }
```

holomorfo:
I have added new inline documentation for the new methods. Here I share the analysis of the pending methods to update and some questions/challenges I'm facing:

### Vector methods
- [x] constructor
- [x] § (get) values - DONE
- [ ] § (set) values -> Check dimentions?
```
    set values(newValues) {
      if (newValues.length === this.dimensions) {
        this._values = newValues.slice()
      } else {
        p5._friendlyError(
          'The new values array does not matches the dimention of the current vector',
          'p5.Vector.values'
        );
      }
    }
```
- [x] (get) x
- [x] getValue 
- [x] setValue
- [x] § (get) y 
- [x] (get) z
- [x] (get) w
- [x] § (set) x 
- [x] (set) y
- [x] § (set) z 
- [x] § (set) w
- [ ] toString 
	Comment from Ken regarding serializtion
- [ ] set
	Need to set dimension also?
- [x] сору
- [x] add
- [ ] rem
	Needs ND change
- [x] sub
- [x] mult 
- [ ] div
	Needs ND change
- [x] mag 
- [x] magSq
- [x] dot
- [ ] cross
	Needs determinant
- [x] dist 
- [x] normalize
- [x] limit 
- [x] setMag
	why is it call setMag?
	maybe because it does it in place
- [x] heading
	Check `_fromRadians`
- [ ] setHeading 
	Only for 2D, how to handle for ND?
- [ ] rotate
	Only for 2D, how to handle for ND?
- [ ] angleBetween 
	needs `cross` to be nD
- [ ] lerp
	Problem with inference of last value being amount for ND
- [ ] slerp
	Problem with inference of last value being amount for ND
- [x] reflect
- [ ] array 
		Needs ND, how to handle default being 3D?
- [x] equals 
- [x] clampToZero 
- [x] `_clampToZero`

Static methods
- [ ] fromAngle 
	only 2D
- [ ] fromAngles 
	only 3D
- [ ] random2D
- [ ] random3D
- [x] сору
- [x] add
- [ ] rem
	Needs module ND
- [x] sub
- [x] mult
- [ ] rotate 
needs rotate Nd
- [ ] div 
- [x] dot
- [ ] cross 
- [x] dist
- [ ] lerp
- [ ] slerp
- [x] mag
- [x] magSq  
- [x] normalize
- [x] limit 
- [x] setMag  
- [x] heading 
- [ ] angleBetween 
	needs corss
- [x] reflect  
- [ ] array
- [x] equals

@davepagurek @limzykenneth @Qianqianye

src/math/p5.Vector.js
limzykenneth: yeah perhaps, might be worth considering how other libraries serialize their vector objects if they do at all

limzykenneth:
For the inline documentation, it is not great at inferring the right name and parent class/module with the code only so we may need to add additional hints around these, you can have a look at the other already documented entries and follow their convention. We can revisit the documentation at a later date though so no need to solve this immediately.

For `set()` and `(set) values`, is the idea here that they can be used to change the dimension of the vector or would the dimension of a vector stay the same throughout its lifetime?

For `setMag()` and others, the name basically all come from Processing and we'll keep them as is for compatibility reasons.

`setHeading()` and `rotate()` currently is documented to work only in 2D. For 3D there probably need to be an axis defined or things like xyz/quartenion values, for 4D I'm not sure what it would mean. I think it can be fine to make them 2D only methods but if you have specific ideas or other libraries handle these in some way, do put something together for consideration.

holomorfo:
> For the inline documentation, it is not great at inferring the right name and parent class/module with the code only so we may need to add additional hints around these, you can have a look at the other already documented entries and follow their convention. We can revisit the documentation at a later date though so no need to solve this immediately.

Sounds good, will do my best now and will revisit the docs later

> For `set()` and `(set) values`, is the idea here that they can be used to change the dimension of the vector or would the dimension of a vector stay the same throughout its lifetime?

 Actually this is my exact question that I'm facing with operations that have multiple dimensions like div and mult, I have added some comments to the PR, I believe that dimension needs to change also in the setters when the values are changed, increased or decrease in components 

> For `setMag()` and others, the name basically all come from Processing and we'll keep them as is for compatibility reasons.

Understood, thanks

> `setHeading()` and `rotate()` currently is documented to work only in 2D. For 3D there probably need to be an axis defined or things like xyz/quartenion values, for 4D I'm not sure what it would mean. I think it can be fine to make them 2D only methods but if you have specific ideas or other libraries handle these in some way, do put something together for consideration.

Sounds right, I will indicate explicitly what functions will stay in fixed dimentions, I will think also of possible generalisations for nD for the defined axis

holomorfo:
I have added the required test for my changes, I increased the cover for the Vector file from:
Screenshot 2024-09-26 at 10 16 01 PM
To
Screenshot 2024-09-26 at 10 17 48 PM

More tests should and will be added to make sure the nD versions of the functions are correct.

CC: @limzykenneth @davepagurek @Qianqianye

holomorfo:
This PR is ready for review.
@limzykenneth @Qianqianye @davepagurek

limzykenneth:
I'd like to test the documentation generation first to see how it works, will get to it when I have a bit more time, you can go ahead with next steps branching off of this branch.

src/math/p5.Vector.js
```diff
@@ -118,23 +127,225 @@ function vector(p5, fn) {
...
         this._toRadians = args[1];
-        x = args[2] || 0;
-        y = args[3] || 0;
-        z = args[4] || 0;
-        // This is what we'll get with new p5.Vector()
+        values = args.slice(2).map(arg => arg || 0);
+      }
+      if (dimensions === 0) {
+        this.dimensions = 2;
+        this._values = [0, 0, 0];
```
davepagurek, line 140: Should the dimensions here match the number of items in `_values`?

src/math/p5.Vector.js
```diff
@@ -118,23 +127,225 @@ function vector(p5, fn) {
...
+     * dimensions of the vector. If the array is shorter, the missing components will be
+     * set to 0. If the array is longer, the extra values will be ignored.
+     *
+     * @param {Array<number>} newValues - An array of numbers representing the new values for the vector.
+     *
+     */
+    set values(newValues) {
+      let dimensions = newValues.length; 
+      if (dimensions === 0) {
+        this.dimensions = dimensions;
```
davepagurek, line 176: I notice this is a little different than in the constructor, where we set `dimensions` to 2 in this case. Do we want this method to work the same way?

davepagurek (commented):
This is looking great! I just flagged some bits that might be inconsistent in case it's unintentional.

src/math/p5.Vector.js
limzykenneth: One thing I'm just now thinking about, should the default vector when no argument is passed be 2D instead of 3D? Seeing as most people start learning to create sketches with 2D before moving into 3D.

davepagurek: I think that makes sense. Even in webgl mode, you may very well still be using 2D vectors for a number of things, so its not a bad idea to have to be explicit to make 3D vectors

src/math/p5.Vector.js
holomorfo: yeah it was a mis, will fix, thanks

holomorfo:
I have added my proposal for the adapter for toggling between `MatrixLegacy` which is what we currently have that is based on the `gl-matrix` implementation and `MatrixNumJs` classs.
To toggle between this I currently setup a static constant
`export const MATRIX_ENGINE = 'legacy';`
`export const MATRIX_ENGINE = 'numjs';`

All existing tests pass with both versions of the matrix, please let me know your thoughts @limzykenneth @davepagurek.

holomorfo:
Since we might have two bundles of `p5.js` with different matrix engines, maybe we should do a configuration to export differently for example `p5.numjs.js` or something similar, and on build time pass a specific flag.
I wanted to ask what do you suggest can be a next step to integrate these changes

limzykenneth:
@holomorfo I can work on the build a bit later, for now we default to the older glMatrix implementation and keep the numjs class around.

holomorfo:
This PR is not needed anymore since changes are in #7405

PR #7314, opened 2024-10-15
Fixes language reference and Markdown syntax

danicruz0415:
Resolves [#106](https://github.com/processing/p5.js-website/issues/106) of repo [p5.js-website](https://github.com/processing/p5.js-website) https://github.com/processing/p5.js-website/issues/106

### Changes:
Fixes the language reference which in some cases was defined as `JavaScript` instead of `javascript` (all lower case). Also In some cases four backticks were used instead of 3 for code blocks


davepagurek (approved):
Thanks for making all these updates!

PR #7326, opened 2024-10-21
Update sketch verifier to check for redefinitions and print friendly messages

sproutleaf:
Addresses https://github.com/processing/p5.js/issues/7269

Changes:

- Added a new function `checkForConstsAndFuncs()` that checks for redefinitions and prints friendly error messages
- Updated unit test accordingly

PR Checklist

- [x] npm run lint passes
- [x] [Inline documentation](https://github.com/processing/p5.js/blob/main/contributor_docs/inline_documentation.md) is included / updated
- [x] [Unit tests](https://github.com/processing/p5.js/tree/main/contributor_docs#unit-tests) are included / updated


src/core/friendly_errors/sketch_verifier.js
```diff
@@ -111,16 +164,116 @@ function sketchVerifier(p5, fn) {
...
+      ...userDefinitions.functions
+    ];
+
+    // Helper function that generates a friendly error message that contains
+    // the type of redefinition (constant or function), the name of the
+    // redefinition, the line number in user's code, and a link to its
+    // reference on the p5.js website.
+    function generateFriendlyError(errorType, name, line) {
+      const url = `https://p5js.org/reference/#/p5/${name}`;
+      const message = `${errorType} "${name}" on line ${line} is being redeclared and conflicts with a p5.js ${errorType.toLowerCase()}. JavaScript does not declaring a ${errorType.toLowerCase()} more than once. p5.js reference: ${url}.`;
```
davepagurek, line 196: Looks like we maybe lost a verb in here:
```suggestion
      const message = `${errorType} "${name}" on line ${line} is being redeclared and conflicts with a p5.js ${errorType.toLowerCase()}. JavaScript does not support declaring a ${errorType.toLowerCase()} more than once. p5.js reference: ${url}.`;
```

davepagurek: Also, I guess JavaScript does technically allow it in different scopes? We could consider saying "your code might not behave correctly" or "your code might behave unexpectedly" or something like that.

src/core/friendly_errors/sketch_verifier.js
```diff
@@ -111,16 +164,116 @@ function sketchVerifier(p5, fn) {
...
+      return;
+    }
   }
 }
 
 export default sketchVerifier;
 
 if (typeof p5 !== 'undefined') {
   sketchVerifier(p5, p5.prototype);
+  p5.prototype.loadP5Constructors();
```
davepagurek, line 278: If we ever want to support loading this module separately, where `p5` may not be globally defined and the exported `sketchVerifier` function is run manually later, it looks like this `loadConstructors` call wouldn't get run. Maybe if we run this as the last thing inside of `sketchVerifier` then it'll get run either way?

src/core/friendly_errors/sketch_verifier.js
```diff
@@ -1,11 +1,64 @@
...
+    'keyPressed',
+    'keyReleased',
+    'keyTyped',
+    'windowResized',
+    'name',
+    'parent',
+    'toString',
+    'print',
+    'stop',
+    'onended'
```
davepagurek, line 38: The functions meant to be redefined by the user make sense. Out of curiosity where do the other false positive entries come from? Is that from looking at `classesWithGlobalFunctions`?

src/core/friendly_errors/sketch_verifier.js
```diff
@@ -111,16 +164,116 @@ function sketchVerifier(p5, fn) {
...
+      const libDefinition = constants[name];
+      if (libDefinition !== undefined) {
+        if (checkForRedefinition(name, libDefinition, line, "Constant")) {
+          return true;
+        }
+      }
+    }
+
+    // Get the names of all p5.js functions which are available globally
+    const classesWithGlobalFns = ['Renderer', 'Renderer2D', 'RendererGL'];
```
davepagurek, line 228: I'm actually not sure we need this any more in 2.0. I don't have the full context on why we had this initially but I think now, the only globals are things that have been added directly to `p5.prototype`. @limzykenneth does that sound right?

davepagurek (commented):
Nice work on this! The one thing I'd want to get a bit of clarification on is some of the false positive cases -- I wonder if that's from treating some p5 classes as globals when they no longer need to be?

src/core/friendly_errors/sketch_verifier.js
limzykenneth: It is relatively unlikely this block will be run unless someone is importing this module as an addon library (which is what it's for). It should be moved into the `sketchVerifier` function instead or if deferred execution is desired, be set to run in the `presetup` lifecycle hook.

src/core/friendly_errors/sketch_verifier.js
```diff
@@ -1,11 +1,64 @@
...
+    // This must be done only when everything has loaded otherwise we get
+    // an empty array
+    for (let key of Object.keys(p5)) {
+      // Get a list of all constructors in p5. They are functions whose names
+      // start with a capital letter
+      if (typeof p5[key] === 'function' && key[0] !== key[0].toLowerCase()) {
+        p5Constructors[key] = p5[key];
+      }
+    }
+  }
```
limzykenneth, line 60: If this function is internal to this module only, ie. not called by other parts of the p5.js code base via `this.loadP5Constructors()` this should probably be defined simply as a scoped function:

```js
const loadP5Constructors = () => {
}
```

src/core/friendly_errors/sketch_verifier.js
```diff
@@ -111,16 +164,116 @@ function sketchVerifier(p5, fn) {
...
+      const url = `https://p5js.org/reference/#/p5/${name}`;
+      const message = `${errorType} "${name}" on line ${line} is being redeclared and conflicts with a p5.js ${errorType.toLowerCase()}. JavaScript does not declaring a ${errorType.toLowerCase()} more than once. p5.js reference: ${url}.`;
+      return message;
+    }
+
+    // Helper function that checks if a user definition has already been defined
+    // in the p5.js library, either as a constant or as a function.
+    function checkForRedefinition(name, libValue, line, type) {
+      try {
+        const userValue = eval("name");
```
limzykenneth, line 204: I'm not sure what this line is meant to be doing but we want to avoid `eval()` as the bundler generally does work well with it.

src/core/friendly_errors/sketch_verifier.js
limzykenneth: Yes, these classes will not have methods attached to global. The rules for attaching anything to global are (if true for both of the following):
1. It is a member of `p5.prototype`
2. Its name does not start with `_`

limzykenneth (commented):
I think we should be able to leverage the AST more as it gives more context and information about where and what things are defined

limzykenneth:
I've also just checked the dependency sizes, it seems acorn and to a slightly lesser extend zod adds quite a bit to the bundle size. Are they importable as treeshakable modules so we don't include code we aren't using?
Screenshot 2024-10-24 at 15 03 26

src/core/friendly_errors/sketch_verifier.js
sproutleaf: Moved `loadP5Constructors` to a utils file!

limzykenneth:
@sproutleaf I kinda forgot about this one but is the changes in here still relevant for merge?

PR #7390, opened 2024-11-23
Adding strokeMode() to access SIMPLE lines. 

perminder-17:
fixes : #7278 

Created a method `strokeMode()` to access SIMPLE lines for faster fps.

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -60,6 +60,18 @@ export const WEBGL2 = Symbol('webgl2');
  * @final
  */
 export const ARROW = 'default';
+
+/**
+ * @property {String} SIMPLE
+ * @final
+ */
+export const SIMPLE = 'simple';
+/**
+ * @property {String} FULLY
+ * @final
+ */
+export const FULLY = 'fully';
+
 /**
  * @typedef {'crosshair'} CROSS
  * @property {CROSS} CROSS
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -67,10 +67,10 @@ export const ARROW = 'default';
  */
 export const SIMPLE = 'simple';
 /**
- * @property {String} FULLY
+ * @property {String} FULL
  * @final
  */
-export const FULLY = 'fully';
+export const FULL = 'full';
 
 /**
  * @typedef {'crosshair'} CROSS
--- a/src/webgl/3d_primitives.js
+++ b/src/webgl/3d_primitives.js
@@ -528,7 +528,7 @@ function primitives3D(p5, fn){
  *
  * - `SIMPLE`: Optimizes for speed by disabling caps, joins, and stroke color features.
  *   Use this mode for faster line rendering when these visual details are unnecessary.
- * - `FULLY`: Enables caps, joins, and stroke color for lines.
+ * - `FULL`: Enables caps, joins, and stroke color for lines.
  *   This mode provides enhanced visuals but may reduce performance due to additional processing.
  *
  * Choose the mode that best suits your application's needs to either improve rendering speed or enhance visual quality.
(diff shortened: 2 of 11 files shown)


perminder-17:
The test which are currently failing are 

1. https://github.com/processing/p5.js/blob/d4809cc648177cf3502d15a2f52b4b0e1ff13c7e/test/unit/visual/cases/webgl.js#L256 
2. https://github.com/processing/p5.js/blob/d4809cc648177cf3502d15a2f52b4b0e1ff13c7e/test/unit/visual/cases/typography.js#L13

I have observed that these tests are also failing in the current Dev 2.0 branch. Therefore, I believe that this pull request is not related to the issues affecting these tests.

davepagurek:
It looks like they're hitting an error in the simple lines code, we might need to double check that it's accessing the properties correctly:
image

src/core/constants.js
```diff
@@ -60,6 +60,18 @@ export const WEBGL2 = Symbol('webgl2');
...
+/**
+ * @property {String} SIMPLE
+ * @final
+ */
+export const SIMPLE = 'simple';
+/**
+ * @property {String} FULLY
+ * @final
+ */
+export const FULLY = 'fully';
```
davepagurek, line 73: A minor note, could we call this FULL instead of FULLY to use the same adjective form as SIMPLE?

src/webgl/3d_primitives.js
```diff
@@ -520,6 +520,94 @@ function primitives3D(p5, fn){
...
+ *   vertex(centerX - lineLength / 2, centerY, 0);
+ *   vertex(centerX + lineLength / 2, centerY, 0); 
+ *   endShape();
+ * }
+ * </code>
+ * </div>
+ */
+  
+fn.strokeMode = function(mode){
+  if(mode === constants.SIMPLE){
```
davepagurek, line 603: Could we make this return the current mode if we call it with no arguments?

src/webgl/p5.Geometry.js
```diff
@@ -1398,7 +1399,7 @@ class Geometry {
       if (dirOK) {
         this._addSegment(begin, end, fromColor, toColor, dir);
       }
-
+      if(!this.renderer._simpleLines){
```
davepagurek, line 1402: Since `p5.Geometry` created manually won't have a renderer passed in, I think we need to handle the case where there is no renderer, and mayube default to the full line behaviour:
```suggestion
      if(!this.renderer?._simpleLines){
```

davepagurek: Also, do you think we could intent the block inside this to keep the indentation consistent everywhere?

src/webgl/p5.Geometry.js
```diff
@@ -1520,6 +1522,7 @@ class Geometry {
       }
     }
     this.lineVertices.push(...a, ...b, ...a, ...b, ...b, ...a);
+    if(!this.renderer._simpleLines){
```
davepagurek, line 1525: Similarly here, we might need to check for the existence of the renderer first:
```suggestion
    if(!this.renderer?._simpleLines){
```

src/webgl/shaders/line.vert
```diff
@@ -67,6 +68,8 @@ vec2 lineIntersection(vec2 aPoint, vec2 aDir, vec2 bPoint, vec2 bDir) {
 
 void main() {
   HOOK_beforeVertex();
+
+  if(!uSimpleLines){
```
davepagurek, line 72: Could we indent the block below?

src/webgl/3d_primitives.js
```diff
@@ -520,6 +520,94 @@ function primitives3D(p5, fn){
...
+ *   background(128);
+ *   
+ *   let centerX = 0;
+ *   let centerY = 0;
+ *
+ *   // Length of the small centered line
+ *   let lineLength = 50; 
+ * 
+ *   beginShape(LINES);
+ *   vertex(centerX - lineLength / 2, centerY, 0);
```
davepagurek, line 565: How do you feel about having a 3D shape like a default `sphere()`, and maybe a single Bezier curve in each of these examples? I feel like those are two cases where the difference is clear: the sphere probably looks the same regardless, and the curve (especially with a higher stroke weight) will look very disjoint without full curve decorations.

We can also make the examples more than 100x100 if that makes it easier to see since I think the new site's layout accommodates that decently.

davepagurek (commented):
Thanks @perminder-17, this is looking good, great explanation of the use of both modes!

perminder-17:
Adjusting the indentation caused unexpected spacing issues in the code, making it difficult to review. I apologize for any inconvenience this may have caused. Also, the checks were failing because we need to check the existence of the renderer first. Thanks for the suggestions:)

davepagurek (approved):
Thanks for the updates, looks great!

PR #7429, opened 2024-12-16
[dev-2.0] visual tests for Type

koundinyad:
Visual tests for typography module 

 Changes:

- New visual tests for:
  - Font Properties
    - Font kerning
    - Font stretch values
    - Font variant caps
  - Text Alignment
    - Multi-line and single-line alignments
    - Wrap character and word tests
  - Text Baseline
  - Text Bounds
    - RTL text bounds
    - Rect mode variations
  - Text Font
    - Default and monospace font tests
  - Text Leading
  - Text Size
  - Text Spacing
    - Letter and word spacing
  - Text Style
  - Text Width

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes


test/unit/visual/cases/type.js
```diff
@@ -18,100 +18,10 @@ visualSuite("Typography", function () {
       p5.text("test", 0, 0);
       screenshot();
     });
-
-    visualTest('with a Google Font URL', async function(p5, screenshot) {
```
davepagurek, line 22: It looks like this deleted some earlier tests. Would you be able to restore these?

test/unit/visual/cases/type.js
```diff
@@ -1,6 +1,6 @@
 import { visualSuite, visualTest } from "../visualTest";
 
-visualSuite("Typography", function () {
```
davepagurek, line 3: Maybe let's keep the title "Typograhy" for now so we won't have to move all the existing screenshots to a new folder

davepagurek (commented):
Thanks for adding more tests! Before merging:
- The existing screenshots for other visual tests have been removed. Could you restore them by calling `git checkout dev-2.0 -- test/unit/screenshots` ?
- It looks like the new tests don't yet have screenshots. Running `npm run test` locally will generate new screenshots for tests that don't have any yet. Could I get you to do that and then commit the results?

koundinyad:
Thank you @davepagurek! Have updated the Typography tests along with all the tests' screenshots

test/unit/visual/cases/typography.js
```diff
@@ -122,22 +124,25 @@ visualSuite("Typography", function () {
...
         { alignX: p5.CENTER, alignY: p5.BOTTOM },
         { alignX: p5.RIGHT, alignY: p5.BOTTOM },
+        { alignX: p5.LEFT, alignY: p5.BASELINE },
+        { alignX: p5.CENTER, alignY: p5.BASELINE },
+        { alignX: p5.RIGHT, alignY: p5.BASELINE },
       ];
 
-      p5.createCanvas(300, 80);
-      p5.textSize(60);
+      p5.createCanvas(300, 200);
```
davepagurek, line 132: For speed, each screenshot is resized to fit in 50x50 before comparison. Since this image (and the alignment test too) is so large compared to that, differences in antialiasing between devices could cause this test to fail.

You can call `screenshot()` multiple times per test and it will generate a separate image each time. Maybe we can make the canvas size smaller for these, and instead of placing each case somewhere else in one canvas, we can clear the canvas for each case and `screenshot()` them before moving on to the next?

davepagurek: For added context, the test failure on CI right now is saying it expected this:
image

but got this:
image

PR #7449, opened 2024-12-27
[dev-2.0] WOFF font support

dhowe:
- Add WOFF support via upgraded Typr.js
- Fixes null argument bug in fontWidth/textWidth
- Fixes direct-font-set triggering on non-string fonts


test/unit/visual/screenshots/Typography/textWeight/can control variable fonts from files/000.png
davepagurek, line 1: Looks like the test for this still exists, were these meant to be deleted? If not, we can restore by running this then committing the results:
```sh
git checkout dev-2.0 -- "test/unit/visual/screenshots/Typography/textWeight/can control variable fonts from files"
```

package-lock.json
```diff
@@ -17,7 +17,8 @@
         "gifenc": "^1.0.3",
         "libtess": "^1.2.2",
         "omggif": "^1.0.10",
-        "opentype.js": "^1.3.1"
+        "opentype.js": "^1.3.1",
```
davepagurek, line 20: oh this reminds me, we can probably `npm uninstall opentype.js` too if we no longer use it

src/type/text2d.js
```diff
@@ -286,22 +279,22 @@ function text2d(p5, fn) {
         family += '-' + font.name.fontSubfamily;
       }
     }
-
-    if (typeof family !== 'string') {
-      throw Error('null font passed to textFont', font);
+    else if (typeof font === 'string') {
```
davepagurek, line 282: Good catch for this one! Do we have a test yet anywhere that checks for a raw canvas string that includes the font size? If not, it could be a good thing to add to make sure we don't accidentally break it going forward

test/unit/visual/screenshots/Typography/textWeight/can control variable fonts from files/000.png
dhowe: tests still failing after I restored those files, but now with a pixel mismatch -- I regenerated the screenshots but still needs a visual check

davepagurek: It looks like the font has changed now, so I think pixel mismatch failures were actually correct in this instance, and something about font loading has changed:
image

davepagurek:
@limzykenneth looks like we just merged some FES changes into dev-2.0, I think they might be causing some test failures 
image

limzykenneth:
Yeah still fixing it, the test for FES may need to be rewritten a bit and some outdated documentation around parameters may also need to be updated. I'll push fixes some time next week.

davepagurek:
I just merged dev-2.0 in again. Everything almost looks good. The raw string test is currently failing, with this expected / received pair:
image
image

I think that's because the server CI is running on doesn't have the same default fonts here. I'm going to see if using a google font manually added to the page via CSS and then using a raw font string works.

davepagurek:
The single line alignment tests are also failing but due to what appears to just be antialiasing issues from the small images. I'm going to see what I can do about that too

davepagurek:
ok I made a bunch of changes to the diffing to try to better handle text antialiasing differences, I think it's good now! I'm going to merge

PR #7462, opened 2025-01-10
Fix typo in image.js docs

reidab:
Changes:
Fixes a typo in the docs from `filenmame` to `filename`

#### PR Checklist


nickmcintyre:
Good catch @reidab!

Issue #6798, opened 2024-02-07
[p5.js 2.0 RFC Proposal]: expand keyIsDown() to work with characters as arguments


kjhollen:
### Increasing access

Having some symmetry where you can also do:

```
keyIsDown('a');
```

would be useful for teaching beginners.

### Which types of changes would be made?

- [ ] Breaking change (Add-on libraries or sketches will work differently even if their code stays the same.)
- [ ] Systemic change (Many features or contributor workflows will be affected.)
- [ ] Overdue change (Modifications will be made that have been desirable for a long time.)
- [X] Unsure (The community can help to determine the type of change.)

### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [X] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [X] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [ ] Other (specify if possible)

### What's the problem?

`keyIsDown()` currently only works with keycodes. Sometimes this is confusing for my students, because they can use this for things like:

```
keyIsDown(LEFT_ARROW);
keyIsDown(ENTER);
```

but if they want to check if an alphanumeric key is down, the code is different:

```
keyIsPressed && key == 'a'
```

### What's the solution?

`keyIsDown()` would need an addition that checks the type of the argument and allows passing text/characters. This means p5.js would have some symmetry where you can also do:

```
keyIsDown('a');
```

### Pros (updated based on community comments)

Example list:
* **Consistency:** This proposal increases API consistency by giving users a way to enter the name of an alphanumeric key for `keyIsDown()`;
* can use `keyIsDown()` without having to look up keycodes

### Cons (updated based on community comments)

* would this be confusing for cases like `keyIsDown(4)` where the user might mean the number 4? (note that 3, 8, and 9 are valid keycodes, so we can't just allow all numbers < 10)

### Proposal status

Under review

RuimingShen:
Hi, I'd like to work on this, thank you! I think I solve it.

wwwld1:
Sorry I don't know if this is a right pin @limzykenneth, I saw that you are under I/O and I am currently working on the implementation for this. It seems to be a conflict for the const "LEFT_ARROW" with the purposed solutions of using String as a input. I can propose 2 solutions.

[...]

PR #7472, opened 2025-01-17
Replaced deprecated keyCode functionality and docs with KeyboardEvent.code & KeyboardEvent.key also updates the keyIsDown function to accept alphanumerics as parameters 

### LINKED ISSUE #7436: [p5.js 2.0 RFC Proposal]: Replace deprecated `keyCode` functionality and docs with `KeyboardEvent.code` & `KeyboardEvent.key` web standard replacements
### Increasing access

Due to the deprecation of the `keyCode` in web standards as shown [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and p5.js's reliance on it for keyboard input and all the doc's examples (at least those I've come in contact with) using them it's time to replace them with their `KeyboardEvent.code` & `KeyboardEvent.key` web standard replacements in order to guarantee the continued functioning of p5.js without having to.

- [X] Breaking change (Add-on libraries or sketches will work differently even if their code stays the same.)

- [X] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [X] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### What's the problem?

`keyCode` is still the primary method of detecting key presses despite it being deprecated for a while now.

### What's the solution?

Replace it with the `KeyboardEvent.code` & `KeyboardEvent.key` web standard replacements in order to guarantee the continued functioning of p5.js without having to.

* **Continued Support:** Continued support by browsers who might remove support in the future
* **Teaching Usage of Web Standards:** Since a large part of p5.js is aimed at beginners we want to avoid teaching outdated and deprecated concepts like the `keyCode`.
* * **Integration:** Although we can keep `keyCode`s to maintain support with older libraries, not moving forward by adding official support and docs everywhere for the modern alternatives might mean losing the ability to intuitively integrate with libraries.

* **Work:** It'll take some work to replace it when `keyCode`s shouldn't get removed from browsers for at least a couple years if not more than a decade.

### Proposal status

Under review

Vaivaswat2244:
Resolves #7436 and #6798

 Changes:

This PR modernizes keyboard event handling by transitioning from the deprecated `keyCode` property to the modern `KeyboardEvent.code` and `KeyboardEvent.key` properties. This change improves keyboard input reliability and brings p5.js in line with current web standards.

* Updated keyboard event handling to use `KeyboardEvent.code` instead of `e.which`
* Added proper key tracking using `_code` property
* Updated keyboard `constants` to use modern string values
* Updated keyIsDown function to use alphanumeric as parameters

 Screenshots of the change:

Screenshot from 2025-01-18 23-45-02

/preview/index.html test sketch is:

```javascript
const sketch = function(p) {
      p.setup = function() {
        p.createCanvas(800, 600);
        p.textSize(16);
        p.textFont('monospace');
      };

      p.draw = function() {
        p.background(240);
        let y = 30;

        // Title
        p.fill(0);
        p.text("keyIsDown() Test", 20, y);
        y += 40;

        // Current key state
        p.text(Current key: ${p.key}, 20, y);
        y += 25;
        p.text(Current code: ${p._code}, 20, y);
        y += 25;

        // Test keyIsDown with direct codes
        p.text("Code String Tests:", 20, y);
        y += 25;
        p.text('ArrowLeft': ${p.keyIsDown('ArrowLeft')}, 40, y);
        y += 20;
        p.text('KeyA': ${p.keyIsDown('KeyA')}, 40, y);
        y += 40;

        // Test keyIsDown with single characters
        p.text("Single Character Tests:", 20, y);
        y += 25;
        p.text('a': ${p.keyIsDown('a')}, 40, y);
        y += 20;
        p.text('A': ${p.keyIsDown('A')}, 40, y);
        y += 20;
        p.text('1': ${p.keyIsDown('1')}, 40, y);
        y += 40;

        // Show current _downKeys object
        p.text("Current _downKeys object:", 20, y);
        y += 25;
        p.text(JSON.stringify(p._downKeys, null, 2), 40, y);
      };

```

#### PR Checklist

- [x] `npm run lint` passes


src/events/keyboard.js
```diff
@@ -902,11 +906,32 @@ function keyboard(p5, fn){
...
+    
+    // For backward compatibility - if code is a number
+    if (typeof code === 'number') {
+      return this._downKeys[code] || false;
+    }
+    
+    // For string inputs (new functionality)
+    if (typeof code === 'string') {
+      // Handle single character inputs
+      if (code.length === 1) {
```
davepagurek, line 921: Maybe we should also add an else if branch to handle the multi-character possibilities for `key`, for things like Enter and the arrow keys: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values If it's one of those, we would need to check against down keys as opposed to down key codes.

davepagurek (commented):
Thanks for taking this on, this is a great start!

I think we might need to keep track of two downkey objects, one for `code`s and one for `key`s, in order to fully support querying of both.

src/events/keyboard.js
```diff
@@ -902,11 +906,32 @@ function keyboard(p5, fn){
...
    */
-  fn.keyIsDown = function(code) {
-    // p5._validateParameters('keyIsDown', arguments);
-    return this._downKeys[code] || false;
+    p5.prototype.keyIsDown = function(code) {
+    console.log('Current _downKeys:', this._downKeys);
+    console.log('Current key:', this.key);
+    
+    // For backward compatibility - if code is a number
+    if (typeof code === 'number') {
```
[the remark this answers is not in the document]
davepagurek: Also @limzykenneth do you think we need backwards compatibility? We might need to put a lookup table like this https://stackoverflow.com/a/66180965 into our code to map old key codes to keys, although it looks like it's hard to get *complete* backwards compatibility. That's probably enough though.

src/events/keyboard.js
```diff
@@ -97,6 +97,8 @@ function keyboard(p5, fn){
    */
   fn.isKeyPressed = false;
   fn.keyIsPressed = false; // khan
+  fn._code = null;
```
limzykenneth, line 100: If this is only used in this module and not used by other modules through the instance, it should be a local variable and not attached to `fn` if possible.

src/events/keyboard.js
```diff
@@ -902,11 +906,32 @@ function keyboard(p5, fn){
    * </code>
    * </div>
    */
-  fn.keyIsDown = function(code) {
-    // p5._validateParameters('keyIsDown', arguments);
-    return this._downKeys[code] || false;
+    p5.prototype.keyIsDown = function(code) {
```
limzykenneth, line 909: This should still be `fn.keyIsDown` and the indentation fixed.

src/events/keyboard.js
limzykenneth: There are some overlap between `key` and `code`, we'll need to decide which to prefer in case of overlap I think.

src/events/keyboard.js
```diff
@@ -902,11 +906,32 @@ function keyboard(p5, fn){
...
+      return this._downKeys[code] || false;
+    }
+    
+    // For string inputs (new functionality)
+    if (typeof code === 'string') {
+      // Handle single character inputs
+      if (code.length === 1) {
+        if (/[A-Za-z]/.test(code)) {
+          // For letters, we need to check the actual key value
+          return this.key === code;
```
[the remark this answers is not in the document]
Vaivaswat2244: okay, got it

src/events/keyboard.js
Vaivaswat2244: here, we are using ```key``` only for the single characters(because of same code for characters i.e. keyA for 'a' and 'A') and for rest of the inputs we are using ```code```

src/events/keyboard.js
```diff
@@ -902,11 +906,32 @@ function keyboard(p5, fn){
...
+        } else if (/[0-9]/.test(code)) {
+          return this._downKeys[`Digit${code}`] || false;
```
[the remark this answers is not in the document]
Vaivaswat2244: yes, we are currently checking with `key` if `code.length === 1`. and checking for `code` otherwise. I'm assuming that we have to completely replace the usage the keyCode ig

src/events/keyboard.js
Vaivaswat2244: Well, currently user is able to do
`text(key, 50, 50);`
giving the key of the key pressed. maybe the same can be done for `code` with added documentation.

src/events/keyboard.js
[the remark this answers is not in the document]
davepagurek: Fully replacing `event.keyCode` is correct! I think ideally we could tell users that if they pass in a code (e.g. `'Digits3'`), then it will check if that physical key is pressed, and if the user passes in a character (e.g. `'3'`), it checks whether whatever key produces that character is pressed. I think that would mean we have to detect whether they passed in an `event.key` or an `event.code`, and then handling each branch separately.

The slight inconsistency right now is if the user passes in the string `'3'`, it is checking for the physical key via the `code` still.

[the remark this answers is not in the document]
Vaivaswat2244: Okay, got it.
I think this logic might work,

```javascript
fn.keyIsDown = function(input){
    if (typeof input !== 'string') {
      return false;
    }
    
    if (typeof input ==='string'){
      if (input.length === 1) {                 
        return this._downKeys[input] ||
               (/[0-9]/.test(input) && this._downKeyCodes[`Digit${input}`])   
                || false; 
      }
      return this._downKeyCodes[input] || this._downKeys[input] || false;
    }
  
    return false;
  };
```
here _downKeyCodes is storing e.code and _downKeys is storing e.key.
now if user passes `'a'` or `'3'`, this will be checked by key and returned true, 
and if passes `'Digit3'`, this will be checked by code and returned true. and for special keys such as Enter,  ' ', Shift, ShiftLeft,  this will give true as they are getting checked for key or code.
Hence, user can use either key or code for the keyisDown function,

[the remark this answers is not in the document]
davepagurek: Sounds good! It might makes sense to make a `isCode()` function that we can call, so you can do:
```js
fn.keyIsDown = function(input) {
  if (isCode(input)) {
    return this._downCodes[input];
  } else {
    return this._downKeys[input];
  }
}
```
...and possibly also unit test `isCode` separately too.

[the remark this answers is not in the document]
Vaivaswat2244: Sure, got it!

src/events/keyboard.js
```diff
@@ -906,32 +906,27 @@ function keyboard(p5, fn){
...
-    console.log('Current key:', this.key);
+  function isCode(input) {
+    if (typeof input !== 'string') {
+      return false;
+    }
     
-    // For backward compatibility - if code is a number
-    if (typeof code === 'number') {
-      return this._downKeys[code] || false;
+    // If it's a single digit, it should be treated as a code (with "Digit" prefix)
```
davepagurek, line 914: Is there a reason why we should treat single digits as codes rather than keys, like we would for alphabetic single characters?

src/events/keyboard.js
```diff
@@ -906,32 +906,27 @@ function keyboard(p5, fn){
...
-          return this._downKeys[`Digit${code}`] || false;
-        }
-      }
-      // Handle direct code inputs (e.g., 'KeyA', 'ArrowLeft', etc.)
-      return this._downKeys[code] || false;
+    // If it's longer than 1 character, it's a code
+    return input.length > 1;
+  }
+  fn.keyIsDown = function(input) {
+    if (isCode(input)) {
```
davepagurek, line 923: nice!

src/events/keyboard.js
Vaivaswat2244: Well, no specefic reason for that. for single character such as `a` or `A`, we cant use `code` because of overlapping value `KeyA`, while for numeric, it doesn't matter.

davepagurek: If a user remaps their keyboard to a different layout, the codes will be the same as before, but the keys will be different, so there could be a difference still between handling the key for a digit and the code for a digit. If we automatically convert digits to codes, then there would be no way for a user to handle the keys for the digits, only the codes (e.g. if you pass in '1', it would turn into the code `'Digit1'`). If we don't special-case digits, though, then users can handle both, either by passing in `'1'` if they want the key, or `Digit1` if they want the code.

Vaivaswat2244: Also, when I run the test script in preview/index.html. 
It gives friendly error in console that `keyisDown` is expecting number as parameter, not string.
I looked into the `param_valdator` and `validate_params` file but couldn't any call to keyboard.js file.
Can you guide me to how to fix this.

davepagurek: oh good catch! It gets the expected type from the comment above the function here: https://github.com/processing/p5.js/blob/5709aac6ab427809b8f1359669fd8fecaac603ea/src/events/keyboard.js#L113 but it only updates the values used by FES when you run `npm run docs`. Try giving that command a run and then testing it out again, does that fix it?

davepagurek: thanks! no rush 🙂

Vaivaswat2244: > oh good catch! It gets the expected type from the comment above the function here:
> 
> https://github.com/processing/p5.js/blob/5709aac6ab427809b8f1359669fd8fecaac603ea/src/events/keyboard.js#L113
> 
> but it only updates the values used by FES when you run `npm run docs`. Try giving that command a run and then testing it out again, does that fix it?

hey tried this after changing the `@param with String` but got this error,
param_validator.js:173 Uncaught (in promise) Error: Unsupported type 'P2D' in parameter validation. Please report this issue.

also reverted the changes in the docs folder and changed String back to Number, and then built it, still got the same.
Also being paranoid, thought to make another clone and tried there, still the same.
What should I do for this?

davepagurek: Looks like this is an error in FES. For now just ignore parameter validation, I can handle that in a separate PR. For your own testing, you can set `p5.disableFriendlyErrors = true` to avoid the error.

Vaivaswat2244: Sure

davepagurek: I'm working on a fix in https://github.com/processing/p5.js/pull/7497, once that's merged you can update your fork to include new changes from dev-2.0 and it should be ok again!

Vaivaswat2244: Sure,
Also I had a problem while writing the tests, like after writing the tests for the new isCode function,
I am getting the `isCode` not defined error while it is defined in the keyboard.js and also works when I test its usage in the preview file.
these are the tests,
```javascript
suite('p5.prototype.isCode', function() {
    test('returns false for non-string inputs', function() {
      assert.isFalse(isCode(null));
      assert.isFalse(isCode(undefined));
      assert.isFalse(isCode(123));
      assert.isFalse(isCode({}));
      assert.isFalse(isCode([]));
    });
  
    test('returns false for single non-digit characters', function() {
      assert.isFalse(isCode('a'));
      assert.isFalse(isCode('Z'));
      assert.isFalse(isCode('!'));
      assert.isFalse(isCode(' '));
    });
  
    test('returns true for multi-character strings', function() {
      assert.isTrue(isCode('Enter'));
      assert.isTrue(isCode('ArrowUp'));
      assert.isTrue(isCode('Shift'));
      assert.isTrue(isCode('Control'));
      assert.isTrue(isCode('ab'));
    });
  
    test('handles edge cases correctly', function() {
      assert.isFalse(isCode(''));  // empty string
      assert.isTrue(isCode('11')); // multi-digit number
      assert.isTrue(isCode('1a')); // digit + letter
    });
  });
```

i wanted to change this as per the updated logic but stuck here because of this. Can you recommend something for this.
The isCode function is like this,
`fn.isCode = function(input){...}`

davepagurek: I'd recommend moving `isCode` outside of `function keyboard` and exporting it, i.e.:
```js
export function isCode(input) {
  // ...
}

function keyboard(p5, fn) {
  // ...
}
```

Then, in your test file, you can import it:
```js
import { isCode } from '../../../src/events/keyboard.js';
```

Vaivaswat2244: yupp this worked. Thanks!

Vaivaswat2244: I have added the tests. I think this should be good to go. Can you have a look at this, @davepagurek

src/events/keyboard.js
```diff
@@ -899,11 +915,23 @@ function keyboard(p5, fn){
...
    */
-  fn.keyIsDown = function(code) {
-    // p5._validateParameters('keyIsDown', arguments);
-    return this._downKeys[code] || false;
-  };
+  function isCode(input) {
+    if (typeof input !== 'string') {
+      return false;
+    }
+    if (input.length === 1 && /[0-9]/.test(input)) {
```
davepagurek, line 922: Based on https://github.com/processing/p5.js/pull/7472#discussion_r1927445260, I think we'd still want inputs like `'9'` to be registered as keys, as `'Digit9'` would be the corresponding code. I think that would just mean removing this if statement, and maybe adding a unit test to double check?

src/events/keyboard.js
```diff
@@ -612,17 +625,20 @@ function keyboard(p5, fn){
...
+    delete this._downKeys[e.key];
+
 
     if (!this._areDownKeys()) {
       this.keyIsPressed = false;
+      this.key = '';
+      this.code = null;
+    } else {
+      // If other keys are still pressed, update code to the last pressed key
+      const lastPressedKey = Object.keys(this._downKeys).pop();
```
davepagurek, line 638: Should this grab from `_downKeyCodes` if we're using it for `this.code` below?

src/core/constants.js
```diff
@@ -879,92 +879,104 @@ export const MITER = 'miter';
...
+export const BACKSPACE = 'Backspace';
+
 /**
- * @typedef {17} CONTROL
+ * @typedef {'ControlLeft' | 'ControlRight'} CONTROL
  * @property {CONTROL} CONTROL
  * @final
  */
-export const CONTROL = 17;
+export const CONTROL = 'ControlLeft';
```
davepagurek, line 902: Each constant can only be defined as one thing, so we might need to make two constants instead of a single one

davepagurek (commented):
I think this is almost there! I left a few things to clean up, and then as a last thing we can delete the `keyCode` property from here: https://github.com/processing/p5.js/blob/9cfef5ad21ed588fc816bd572e20ac4e566a4edc/src/events/keyboard.js#L292

src/events/keyboard.js
Vaivaswat2244: Sure, got it

src/events/keyboard.js
Vaivaswat2244: Yess. I'll also add another variable for handling keys on keyup() as well, same as code

src/core/constants.js
Vaivaswat2244: So, I'm not entirely sure what to do for this.
Should I make two constants one for key and code, like
```javascript
export const CONTROL_CODE = 'Control';
export const CONTROL_KEY = 'ControlLeft';
```

Vaivaswat2244:
hello, there has been a commit in the dev-2.0 branch for using meta key. That has been made assuming the usage of keycode. In that implementation, a new object of `_metaKeys` has been created for storing keyCode of meta key. But this implementation of using key and code already handles the case of meta as well. So should I go ahead and resolve those conflicts by removing the metaKey object?

davepagurek:
That's fine as long as the original issue is resolved! Try using the instructions in this comment https://github.com/processing/p5.js/issues/7282#issuecomment-2379835245 to see what the original issue was, and then if that works on your branch already, we can take out the special casing that was added in main.

Vaivaswat2244:
Yeah, its working. In the issue #7282, you tested for event.metaKey && event.keyCode == 90,
with changes by this pr, it handles for the event.metaKey && event.key === 'z' also for code and keyCode as well.

src/core/constants.js
Vaivaswat2244: Can you resolve this doubt of mine?

davepagurek: I was thinking originally having a const `CONTROL_LEFT` and `CONTROL_RIGHT`, but on second thought, the use case before of having a single `CONTROL` implies that it doesn't matter which. So it probably makes most sense to just `export const CONTROL = 'Control` so that it ends up using the `key` instead of the `keyCode` so that it would check for both.

I think this might mean we need to update `isCode` so that `isCode('Control')` returns false though. And for everything we do for `Control`, we would also do for `Alt` and `Shift`, since they have a left/right version too.

davepagurek (commented):
Thanks for doing the merge, it's great that we can simplify the meta key situation with your code!

src/events/keyboard.js
```diff
@@ -916,6 +925,15 @@ function keyboard(p5, fn){
    * </div>
    */
   function isCode(input) {
+    const leftRightKeys = [
```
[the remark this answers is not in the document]
Vaivaswat2244: Oh yeah, sure.

test/unit/events/keyboard.js
```diff
@@ -189,6 +189,11 @@ suite('Keyboard Events', function() {
       assert.isTrue(isCode('Control'));
       assert.isTrue(isCode('ab'));
     });
+
+    test('returns false for strings for letright keys', function() {
+      assert.isFalse(isCode('AltLeft'));
```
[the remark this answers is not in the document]
Vaivaswat2244: Yeah, that's right.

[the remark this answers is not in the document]
Vaivaswat2244: fixed this, returning false for keys and true for codes

davepagurek (approved):
Thanks for your patience making all these changes, great work! I think we're good to go!

davepagurek:
@all-contributors please add @Vaivaswat2244 for code

PR #7544, opened 2025-02-13
Improve constrain

luizbills:
Changes:

- Make `constrain()` more readable: https://github.com/processing/p5.js/commit/c73e39638b2ff45eea9df512ae23b7fee48222c9

- Reduce some bytes of the minified bundle.

```js
// current constrain() minified
constrain=function(e,t,r){return i.default._validateParameters("constrain",arguments),Math.max(Math.min(e,r),t)

// proposed constrain() minified
constrain=function(e,t,r){return i.default._validateParameters("constrain",arguments),e<t?t:r<e?r:e}
```

#### PR Checklist

- [x] `npm run lint` passes


ksen0:
Hi @luizbills !  I'll close this issue for now as it's updating something based only on style, and does not address any open or approved issue. In the future, please note that according to [contributor guidelines](https://p5js.org/contribute/contributor_guidelines/) PRs without issues will be closed:

> You should not file a pull request (or start working on code changes) without a corresponding issue or before an issue has been approved for implementation; that is because the proposed fix may not be accepted, need a different approach entirely, or the actual problem is somewhere else. Any pull requests filed before the issue has been approved for fixing will be closed until approval is given to the issue.

PR #7561, opened 2025-02-17
Add screenToWorld function mirroring worldToScreen

### LINKED ISSUE #7342: Automatically keep track of transformations
### Increasing access

This would mean that transformations can be fully managed by p5.js, which not only means increased ease of control for the user, but also not having to have 2 sets of data for the same thing, one on p5.js's side and one on the user's side.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [X] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### Feature request details

While this is already possible in `P2D` mode with `drawingContext.getTransform();` it doesn't apply to WebGL which doesn't use `CanvasRenderingContext2D`. Also, the transform returned with `drawingContext.getTransform();` is a transformation matrix, which many people won't know how to use.

The way this would be done is through an official p5.js functions that return the transformation matrix, transform or untransform points, or return attributes that are decided as essential (the rest can be taken from the transformation matrix).

bojidar-bg:
Resolves #7342

Changes:
This PR adds a `screenToWorld` function, which is the inverse of the `worldToScreen` function added by @Garima3110 in #7113. (In fact, this PR closely mirrors the changes made in Garima's PR.)

In 2D, this function should allow users to make transformed shapes interactive, and e.g. detect mouse clicks inside rotated rectangles without extra trigonometry.
Sadly, the function is not too useful for 3D as-is, since in 3D one would need actual ray-casting (or at least plane-ray intersection) to calculate the correct depth of the point.

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -1272,7 +1272,9 @@ function environment(p5, fn){
    * of 3D objects.
    *
    * @method worldToScreen
-   * @param {p5.Vector} worldPosition The 3D coordinates in the world space.
+   * @param {Number|p5.Vector} x The x coordinate in world space. (Or a vector for all three coordinates.)
+   * @param {Number} y The y coordinate in world space.
+   * @param {Number} [z] The z coordinate in world space.
    * @return {p5.Vector} A vector containing the 2D screen coordinates.
    * @example
    * <div>
@@ -1384,6 +1386,11 @@ function environment(p5, fn){
    *
    */
   fn.worldToScreen = function(worldPosition) {
+    if (typeof worldPosition === "number") {
+      // We got passed numbers, convert to vector
+      worldPosition = this.createVector(...arguments);
+    }
+
     const matrix = this._renderer.getWorldToScreenMatrix();
     const screenPosition = matrix.multiplyAndNormalizePoint(worldPosition);
     return screenPosition;
@@ -1391,15 +1398,17 @@ function environment(p5, fn){
   /**
    * Converts 2D screen coordinates to 3D world coordinates.
    *
-   * This function takes a vector and converts its coordinates
-   * from the world space to screen space. This can be useful for determining
-   * the mouse position relative to a 2D or 3D object.
+   * This function takes a vector and converts its coordinates from coordinates
+   * on the screen to coordinates in the currently drawn object. This can be
+   * useful for determining the mouse position relative to a 2D or 3D object.
    *
-   * The Z component of the screen coordinates is treated as "depth", or
-   * distance from the camera, when converting to 3D world space.
+   * If given, the Z component of the input coordinates is treated as "depth",
+   * or distance from the camera.
    *
    * @method screenToWorld
-   * @param {p5.Vector} screenPosition The coordinates in screen space.
+   * @param {Number|p5.Vector} x The x coordinate in screen space. (Or a vector for all three coordinates.)
+   * @param {Number} y The y coordinate in screen space.
+   * @param {Number} [z] The z coordinate in screen space.
    * @return {p5.Vector} A vector containing the 3D world space coordinates.
    * @example
    * <div>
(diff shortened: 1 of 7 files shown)


src/core/environment.js
```diff
@@ -1406,6 +1406,102 @@ function environment(p5, fn){
...
+   * </div>
+   *
+   */
+  fn.screenToWorld = function(screenPosition) {
+    const renderer = this._renderer;
+    if (renderer.drawingContext instanceof CanvasRenderingContext2D) {
+      // Handle 2D context
+      const inverseTransformMatrix = new DOMMatrix()
+        .multiply(renderer.drawingContext.getTransform().inverse())
+        .scale(renderer._pInst.pixelDensity());
```
bojidar-bg, line 1485: This code could potentially be moved to a `Renderer2D.calculateCombinedMatrix()` function, which would let us get rid of the `if` above.
Thoughts on such a refactoring (I would do the same to `worldToScreen`)?

test/unit/core/environment.js
```diff
@@ -292,5 +323,47 @@ suite('Environment', function() {
...
+      let screenPos = myp5.createVector(50, 50, 10/11);
+      let worldPos = myp5.screenToWorld(screenPos);
+      myp5.pop();
+      assert.closeTo(worldPos.y, 0, 0.1);
+    });
+
+    test('screenToWorld with rotation in 3D around Z-axis', function() {
+      myp5.push();
+      myp5.rotateZ(myp5.PI / 2);
+      let screenPos = myp5.createVector(50, 60, 10/11);
```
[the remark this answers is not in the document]
davepagurek: To confirm my understanding, is `worldToScreen(createVector(0,0,0)).z` equal to 10/11?

src/core/environment.js
davepagurek: That makes sense to me if we want to extract these into functions on the respective renderers. `calculateCombinedMatrix` is specifically for model-view currently, so maybe we could make a new function for it that returns a matrix going right from p5 screen coordinates --> world coordinates?

test/unit/core/environment.js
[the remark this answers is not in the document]
bojidar-bg: @davepagurek Correct, yes. I used that to avoid hardcoding the value in the code example for `screenToWorld`.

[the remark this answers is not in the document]
davepagurek: ah ok got it. Actually, do you think it makes sense to always rewrite the `z` value coming into `screenToWorld`, assuming the screen is 2D? So regardless of what you pass in, we replace the z value with the camera plane, which I guess would be `worldToScreen(createVector(currentCam.centerX, currentCam.centerY, currentCam.centerZ)).z`?

src/core/environment.js
bojidar-bg: `calculateScreenToWorldMatrix`? :sweat_smile: 

Also, do note that `calculateCombinedMatrix` is used in only 3 places right now. One is internally in RendererGL, one is in screenToWorld/worldToScreen, and one is in accessibility/outputs.js's `_getPos` function (since your 631a14d157b3a066be2e941dc1248f8ab858e829), which... looks like it's almost exactly the same thing as `worldToScreen` and, if I'm not wrong, will be fine (better!) with the projection matrix applied as well?

davepagurek: Right, there's probably more use for the fully combined one, but there are some specific spots (passing it into `uMVMatrix` for shaders mostly) where we need just model + view and nothing else, so as long as we keep that the same, it's all good! we can rename the existing one to something else if we need.

test/unit/core/environment.js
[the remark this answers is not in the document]
bojidar-bg: Hmm... yes and no. I think it would make sense to make a 0 z value (or, failing that, a 800 z value) be equivalent to the normal, unrotated plane, instead of the 10/11 value. But I'd like to avoid assuming the screen is 2D if possible.
(..unless one can manually get the world-to-screen matrix in some documented way, in which case having screenToWorld take only x and y would be perfectly fine)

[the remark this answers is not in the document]
davepagurek: right, I wouldn't expect most users to assume they'd have to grab a default z. If we add an overload so in addition to `screenToWorld(vec)` you could call `screenToWorld(x, y, [z])`, we could see if z is undefined and then stick a default value in? And for vectors, we can check if `vec.dimensions === 2`, and provide a default z, and otherwise leave it as is.

src/core/environment.js
bojidar-bg: @davepagurek I switched it to a `getWorldToScreenMatrix` function that carries out the whole calculation, and changed the `_getPos` in accessibility/outputs.js to use it instead (in fact, going through worldToScreen directly).
Let me know if you feel like a `RendererGL.getModelViewMatrix()` function would be useful.

src/math/p5.Vector.js
```diff
@@ -34,14 +34,14 @@ class Vector {
   // This is how it comes in with createVector()
   // This check if the first argument is a function
   constructor(...args) {
-    let dimensions = args.length; // TODO: make default 3 if no arguments
```
bojidar-bg, line 37: With `createVector`, this would be e.g. 5 for a 3-dimensional vector. As it is, there are no uses of the `dimensions` property, which is likely why it went by unnoticed.

src/webgl/p5.RendererGL.js
```diff
@@ -1024,12 +1024,23 @@ class RendererGL extends Renderer {
...
+    const projectionMatrix = this.states.uPMatrix;
+    const projectedToScreenMatrix = new Matrix(4);
+    projectedToScreenMatrix.scale(this.width, this.height, 1);
+    projectedToScreenMatrix.translate([0.5, 0.5, 0.5]);
+    projectedToScreenMatrix.scale(0.5, -0.5, 0.5);
+
+    const modelViewMatrix = modelMatrix.copy().mult(viewMatrix);
+    const modelViewProjectionMatrix = modelViewMatrix.mult(projectionMatrix);
+    const worldToScreenMatrix = modelViewProjectionMatrix.mult(projectedToScreenMatrix);
+    return worldToScreenMatrix;
```
bojidar-bg, line 1043: Tips on avoiding allocations here welcome :sweat_smile: I guess the `copy()` is unavoidable, but allocating `projectedToScreenMatrix` feels avoidable :thinking: (Or should I just cache it on resize?)

src/math/p5.Vector.js
davepagurek: Good catch!

test/unit/math/p5.Vector.js
```diff
@@ -64,6 +68,10 @@ suite("p5.Vector", function () {
       assert.equal(v.y, 2);
       assert.equal(v.z, 3);
     });
+
+    test("should have dimensions initialized to 3", function () {
+      assert.equal(v.dimensions, 3);
```
davepagurek, line 73: Thanks for adding these!

src/webgl/p5.RendererGL.js
bojidar-bg: Yeah, that's what I ended up discovering as well :sweat_smile: 
Let's just leave it as is for now; we need more sketches that use it before it's clear how impactful that one function is. And given that `modelMatrix.copy().mult(viewMatrix)` already happens multiple times per shape, I'm pretty sure there are slower things around anyway.

davepagurek: ok sounds good, I think we're all set to merge this then! thanks for your work!

davepagurek (approved):

davepagurek:
@all-contributors please add @bojidar-bg for code, test

PR #7636, opened 2025-03-15
loadFilterShader fixes

perminder-17:
While working on the docs with loadFilterShaders, I noticed a bug saying 

`t._validateParameters is not a function`.

Not adding a parameters validation fixes the issue.


src/webgl/material.js
```diff
@@ -533,7 +533,7 @@ function material(p5, fn){
    * let myShader;
    *
    * async function setup() {
-   *   myShader = await loadFilterShader('assets/shader.frag');
+   *   myShader = await loadFilterShader('assets/basic.frag');
```
perminder-17, line 536: there were some linking issue with vertex shader from shader.frag, so using basic.frag

Screenshot from 2025-03-16 03-50-04

davepagurek (approved):
Thanks for catching this, I thought we removed all instances of it but I guess we missed this one!

PR #7638, opened 2025-03-16
loadbytes, loadTable and some minor fixes with other modified functions.

perminder-17:


src/io/p5.Table.js
```diff
@@ -1083,32 +1166,43 @@ class Table {
...
+   *   table = await loadTable('assets/mammals.csv', ',', 'header');
+   *
+   *   // Set text properties
+   *   fill(0);      // Text color: black
+   *   textSize(12); // Adjust text size as needed
+   *
+   *   // Display each table cell value on the canvas one below the other.
+   *   // We use a variable 'y' to increment the vertical position.
+   *   let y = 20;
+   *   text(table.getString(0, 0), 10, y); // 0
```
holomorfo, line 1188: is there a pedagogical or didactic reason to have these additions unfolded and not in a for loop like in the rest of the previous examples? It's correct of course, but I found it strange and different than the pattern you were following up until now.
I see that each line has the comment of the Animal type, but since its not being rendered, was wondering if it would not be best to show the list after the a for loop.

holomorfo (commented):
@perminder-17 Changes look good, only please resolve the conflicts in the PR and I had a small question regarding a potential for loop that might be needed instead of explicitly unfolding it.

src/io/p5.Table.js
perminder-17: yes, Cristian, but these are just modified functions, I have just updated it so it should display on the canvas instead of my console. Here's the previous implementation which again didn't uses a for loop in it.

```md
   *   print(table.getString(0, 0)); // 0
   *   print(table.getString(0, 1)); // Capra hircus
   *   print(table.getString(0, 2)); // Goat
   *   print(table.getString(1, 0)); // 1
   *   print(table.getString(1, 1)); // Panthera pardus
   *   print(table.getString(1, 2)); // Leopard
   *   print(table.getString(2, 0)); // 2
   *   print(table.getString(2, 1)); // Equus zebra
   *   print(table.getString(2, 2)); // Zebra
   ```

so, I didn't change it completely using a for loop.

holomorfo: looks good, only wanted to understand better, thanks

holomorfo (approved):

holomorfo:
@perminder-17 Looks good, lets test how it renders maybe we can add also screenshots in the PRs, I'll do the same with mine

PR #7649, opened 2025-03-21
fixing-strokeShader() and some minor fixes in shaderHooks-docs

perminder-17:
I know we should hold-off on the docs with the shaderHooks but they were very minor so fixing little stuffs.


holomorfo (approved):
Looks good, lets test how it renders maybe we can add also screenshots in the PRs, I'll do the same with mine

PR #7676, opened 2025-03-27
[p5.js 2.0] Restore single color value define grayscale color

### LINKED ISSUE #7652: [p5.js 2.0 Beta Bug Report]: Ambiguity in how grayscale value should be evaluated in non-RGB color mode

- [ ] Accessibility
- [x] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.0.0-beta.4 but applies generally

### Web browser and version

_No response_

### Operating system

_No response_

### Steps to reproduce this

@perminder-17 Brought this up with me yesterday around difference in how `color()` (and in effect any function that accepts a color such as `fill()`) deals with a single numerical argument is given. The reference says that if only one numerical argument was given, it will be interpreted as grayscale value. However, there are ambiguity around how it should be interpreted precisely, especially in regards to maxes. See the following cases:

In `colorMode(RGB)` it is by default assumed that `color(100)` is equivalent to `color(200, 200, 200)` ie the value is effectively used for all RGB values, this makes sense. However, we can also set `colorMode(RGB, 255, 200, 200)` which creates an ambiguity when `color(100)` is called: what is the max value for this argument here? Is it 255, 200, or somewhere in between 200 and 255? (In 1.x, rather unintuitively, the blue max is used, not sure why)

In `colorMode(HSB)` or any other non-RGB color mode it is also not clear which max will be used if it were to create a grayscale color. In 1.x, it seems it is [meant to use brightness/lightness max](https://github.com/processing/p5.js/blob/main/src/color/p5.Color.js#L1048-L1052). We can possibly to the same for 2.0 for colors, so using lightness where available and brightness otherwise, although it can be a bit ambiguous still for HWB color which needs two values to get the full range of black to white color and for future potential color spaces where grayscale is not on a single dimension.

@ksen0 @davepagurek Do you have any thoughts on this?

limzykenneth:
Resolves #7652

 Changes:

Defining colors with single value now always gives grayscale color. Behavior of RGB, HSB, and HSL mode should be identical to 1.x (ie. RGB still uses Blue max, not sure if that is still what we want but it can be changed if needed). 

Lab, LCH, OKLab, and OKLCH will use the lightness value and max.

HWB is the trickiest one and I don't think have a good solution. The reason is the full range of grayscale value in HWB requires both the whiteness and blackness values (ie. W and B) so it may be required to parse a single max value from potentially 2 different max values. What I have done in this case is to get both maxes and calculate an average max and use that instead (eg. W max is 50 and B max is 100: (50 + 100) / 2 = 75, 75 is the max used). Unlike the other single value max however, it cannot take into account more complex max ranges such as [-100, 100] since that is even more ambiguous (defining max as a range is a new not yet documented feature of the color mode function).

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -44,18 +44,23 @@ class Color {
 
   static colorMap = {};
   static #colorjsMaxes = {};
+  static #grayscaleMap = {};
 
   // Used to add additional color modes to p5.js
   // Uses underlying library's definition
   static addColorMode(mode, definition){
     ColorSpace.register(definition);
     Color.colorMap[mode] = definition.id;
+
     // Get colorjs maxes
     Color.#colorjsMaxes[mode] = Object.values(definition.coords).reduce((acc, v) => {
         acc.push(v.refRange || v.range);
         return acc;
       }, []);
     Color.#colorjsMaxes[mode].push([0, 1]);
+
+    // Get grayscale mapping
+    Color.#grayscaleMap[mode] = definition.fromGray;
   }
 
   constructor(vals, colorMode, colorMaxes) {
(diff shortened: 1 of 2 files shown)


ksen0:
This seems great! Thank for the clarification too.

@perminder-17 I see the new colorMode arguments are on your documentation list, please prioritize that task one from the remaining "Added" documentation, and include these grayscale updates. I'm adding it to the spreadsheet as well. Thank you!

perminder-17:
> This seems great! Thank for the clarification too.
> 
> @perminder-17 I see the new colorMode arguments are on your documentation list, please prioritize that task one from the remaining "Added" documentation, and include these grayscale updates. I'm adding it to the spreadsheet as well. Thank you!

Sure, thanks

PR #7690, opened 2025-03-30
Updating Gray-scale in colorMode()

perminder-17:


src/color/setting.js
```diff
@@ -773,14 +773,31 @@ function setting(p5, fn){
    *
    * `OKLCH`  - An easier-to-use representation of OKLAB, expressing colors in terms of Lightness, Chroma, and Hue.
    *          This mode retains the perceptual benefits of OKLAB while offering a more intuitive format for color manipulation.
-   * 
+   *
    * <a href="#/p5.Color">p5.Color</a> objects remember the mode that they were
    * created in. Changing modes doesn't affect their appearance.
    *
+   *  `Single-value (Grayscale) Colors` :
```
holomorfo, line 780: The colon in "Single-value (Grayscale) Colors :" has an extra space before it. small thing but wanted to check for concistency.

src/color/setting.js
```diff
@@ -773,14 +773,31 @@ function setting(p5, fn){
...
    * <a href="#/p5.Color">p5.Color</a> objects remember the mode that they were
    * created in. Changing modes doesn't affect their appearance.
    *
+   *  `Single-value (Grayscale) Colors`:    
+   *  When a color is specified with only one parameter (e.g., `color(g)`), p5.js will interpret it
+   *  as a grayscale color. However, how that single parameter translates into a grayscale value
+   *  depends on the color mode:
+   *
+   * - `RGB, HSB, and HSL`: The single value will be used consistently for the gray equivalent 
+   *    in each of these modes (for example, mapping to red/green/blue channels equally in RGB).
```
limzykenneth, line 786: RGB uses blue max (not sure if it is worth to change it).

HSB and HSL uses Brightness and Lightness max respectively with hue and saturation at 0, at this mode, both color spaces are equivalent.

src/color/setting.js
```diff
@@ -1092,7 +1111,55 @@ function setting(p5, fn){
...
    * </div>
+   * 
+   * @example
+   * <div>
+   * <code>
+   * 
+   *  // Example: Single-value (Grayscale) colors in different color modes. 
+   *  // Each rectangle is filled with one parameter, but its final color depends
+   *  // on how that parameter is interpreted by the current color mode.
+   *
```
perminder-17, line 1122: Hi @limzykenneth can you look at this sketch? can we keep this one or should I change it?

limzykenneth: I'm thinking whether to include LAB and HWB instead of the other color spaces, as although LAB and HWB may potentially be the most confusing to work with and can benefit from examples, it may be counter productive to include it here because of the same reason with grayscale usage possibly something that people actively choosing LAB or HWB color space would not be aiming to do.

Eg. For Lab color space, LCH is much easier to use while the lightness value (L) works the same for both. While for HWB, the benefit is to be able to choose a hue and adjust its brightness and darkness with two separate handle like mixing color directly, as opposed to thinking about saturation then brightness/lightness which can be less intuitive, as such it can be reasoned that grayscale for HWB is more of an edge case.

perminder-17:
Hi @limzykenneth I removed Lab and hwb, let me know if still there's any changes required.

holomorfo (approved):

limzykenneth (approved):

PR #7721, opened 2025-04-10
Updating foundation-section for async/await

perminder-17:
need to add more examples and polish minor docs

Testing still left

### DIFF
--- a/src/core/reference.js
+++ b/src/core/reference.js
@@ -1141,36 +1141,48 @@
 /**
  * Asynchronous Asset Loading with Async/Await.
  *
- * Using async/await in your p5.js sketch allows you to load assets, such as images and JSON data,
- * in a straightforward, linear fashion without resorting to nested callbacks. Declaring a function
- * with the async keyword tells JavaScript that this function will perform asynchronous operations and
- * will always return a promise. Within an async function, the await keyword pauses the execution until
- * the promise is resolved, making the code appear synchronous and easier to read and maintain.
+ * The keywords `async` and `await` let you write asynchronous code in a more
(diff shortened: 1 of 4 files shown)


src/core/reference.js
```diff
@@ -1137,6 +1137,105 @@
...
+ * When you mark a function with the `async` keyword—like `async function setup() {...}`—it
+ * signals that the function contains asynchronous operations and will return a
+ * promise. Any time you use the `await` keyword inside this function, JavaScript
+ * will pause the function’s execution until the awaited promise settles.
+ *
+ * In p5.js, you can use `async/await` to handle media loading functions such as
+ * `loadImage()`, `loadJSON()`, `loadSound()`, and so on. This allows you to:
+ *   - load files in a more readable, top-to-bottom manner
+ *   - decide when the assets are fully available before proceeding
+ *   - avoid nested callbacks often referred to as "callback hell"
```
ksen0, line 1159: This is great and much needed documentation! Two notes here:

1. I believe it would be best not to use the phrase "callback hell" here. This is because the people familiar with this term are not likely to be the majority of the documentation readers. Instead, spell out in 1 or 2 additional sentences:  that nested callbacks require managing additional information and behavior, and lazy loading can be helpful but it also requires you to design your sketch around that.

2. Please me sure to have a separate paragraph here of 1-3 sentences explaining that callbacks can still be used as they were before, since this has come up a few times in past Q&A

Thank you so much!

perminder-17: Sure, i can update this ASAP. Thanks for the review @ksen0

ksen0:
Updates look great @perminder-17 ! Thank you

PR #7728, opened 2025-04-12
P2HDR docs      

perminder-17:


perminder-17:
Looks like tests aren't passing, will look into it.

src/core/constants.js
```diff
@@ -15,14 +15,28 @@ export const VERSION = 'VERSION_WILL_BE_REPLACED_BY_BUILD';
 
 // GRAPHICS RENDERER
 /**
- * The default, two-dimensional renderer.
- * @typedef {unique symbol} P2D
+ * The default, two-dimensional renderer in p5.js.
+ *
+ * Use this when calling {@link p5.createCanvas|createCanvas} (for example,
```
davepagurek, line 20: I'm not sure that we support `@link` yet, this is the result of `npm run docs`:
image

It looks like we don't use this syntax elsewhere in the codebase currently, could we keep using the old `<a href="#/p5/createCanvas">` for consistency?

src/core/constants.js
```diff
@@ -15,14 +15,28 @@ export const VERSION = 'VERSION_WILL_BE_REPLACED_BY_BUILD';
...
+ * `createCanvas(400, 400, P2D)`) to specify a 2D context.
+ *
+ * @typedef {'p2d'} P2D
  * @property {P2D} P2D
  * @final
  */
 export const P2D = 'p2d';
 
-export const P2DHDR = 'p2d-hdr';
```
davepagurek, line 25: I think you'll need to add this line back after your comment block to fix the tests -- looks like the error is that this constant is no longer defined.

perminder-17:
Sure @davepagurek , sorry for the oversight.

davepagurek:
No worries! `@link` is part of the jsdoc spec so it's not obvious that it wouldn't be supported.

PR #7780, opened 2025-04-26
Match the Text Wrapping Behavior to p5.js Documentation

### LINKED ISSUE #7163: Text Wrapping Behavior in p5.js doesn't match the documentation

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [X] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

1.10.0

### Web browser and version

127.0.6533.89 (Official Build) (arm64) 

### Operating system

macOS 13.4

### Description
The text wrapping behavior in p5.js does not match the [documentation for `text()`](https://p5js.org/reference/p5/text/) (or the documentation is ambiguous). The documentation states that specifying `maxWidth` and `maxHeight` in the `text()` function should automatically wrap text within those dimensions:

> "The fourth and fifth parameters, maxWidth and maxHeight, are optional. They set the dimensions of the invisible rectangle containing the text...Text will wrap to fit within the text box."

While word wrapping occurs when boundaries are set, character-wise wrapping requires the additional `textWrap(CHAR)` call, which is not mentioned in the documentation. 

This is also different from the way the `text()` function works in Processing, where text character-wise wrapping is enabled by default when setting the fourth and fifth parameters of the `text()` function (see: https://github.com/processing/processing-website/issues/544)

### Steps to Resolve
I can think of two ways to resolve this:

A) Set `text()` to automatically wrap characters when setting the fourth and fifth parameters. This would have the advantage of matching the behavior of the function by the same name in Processing.

or

B) Update the documentation to clarify that "Text will wrap to fit within the text box" on _words_ only and that an additional `textWrap(CHAR)` will be required to wrap on _characters_.

Mamatha1718:
Resolves #7163 

 Changes:

1. Added documentation for the `textWrap()` function in `src/typography/loading_displaying.js`.
 
3. Included examples showing how to use `textWrap(CHAR)` and `textWrap(WORD)`.
 

 

#### PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/src/typography/loading_displaying.js
+++ b/src/typography/loading_displaying.js
@@ -180,8 +180,29 @@ p5.prototype.loadFont = function(path, onSuccess, onError) {
 };
 
 /**
- * Draws text to the canvas.
+ * Specifies how text should wrap when displayed inside a text box using the text() function.
+ * 
+ * When you use text() with width and height parameters, the text will wrap automatically.
+ * By default, p5.js wraps text based on whole words — it moves entire words to the next line
+ * when they don’t fit in the given width.
+ * 
+ * If your text contains very long words or no spaces (like long URLs or technical terms),
+ * and you want it to wrap at the character level instead, use textWrap(CHAR).
+ * 
+ * You can switch between word-based or character-based wrapping by passing either WORD or CHAR.
+ * 
+ * @method textWrap
+ * @param {Constant} mode either WORD or CHAR
  *
+ * @example 
+ * textWrap(CHAR);
+ * text("TheQuickBrownFoxJumpsOverTheLazyDogWithoutAnySpaces", 10, 30, 150, 100);
+ *
+ * @example 
+ * textWrap(WORD);
+ * text("This is a sentence that wraps based on word boundaries.", 10, 30, 150, 100);
+ */
+ /** 
  * The first parameter, `str`, is the text to be drawn. The second and third
  * parameters, `x` and `y`, set the coordinates of the text's bottom-left
  * corner. See <a href="#/p5/textAlign">textAlign()</a> for other ways to


perminder-17 (changes requested):
Hi @Mamatha1718 , Really thanks for working on this and I am really sorry for getting back to you this late. From now, you can tag me on the documentation PR's and I'll review/merge it within a day or two. 

I see, now the tests are failing. Because `textWrap()` is already [documented in the code](https://p5js.org/reference/p5/textWrap/), adding a second `@method textWrap` block here would only create a duplicate entry in the generated reference. The cleaner fix for the 1.x branch is therefore:

 ### Replace the single-sentence note inside `text()`, 
 You should replace the below line in `text()` function,  
```
  Text will wrap to fit within the text box. 
```
 TO

```
By default p5.js wraps text **at word boundaries**.  
If you need exceptionally long words (e.g. URLs, CJK text without spaces) to fit
inside the same box, call `textWrap(CHAR)` first to wrap on every character.
Call `textWrap(WORD)`—or simply omit the call—to restore the default behaviour.
```

Feel free to write this in your own words, It's just the idea what needed to be done. That fulfils the request in issue #7163 without touching the rest of the description.


### Add two extra example that demonstrates textWrap(CHAR) and textWrap(WORD).

If you see the examples, something like
```
<code>
your code....
</code>
```
So, you need to add examples which uses `textWrap(CHAR)` and `textWrap(WORD)` this will. Put this at the end of the existing `@example` series inside the same `text()` block. 

Let me know if I am clear with the explanation, you should not need to add any additional JSDoc tags.

PR #7795, opened 2025-05-06
Consistent `movedX` and `movedY` behaviour across zoom levels

### LINKED ISSUE #7790: incorrect `movedX` and `movedY` behavior on zoomed sketches in Chrome

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [x] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

1.11.5

### Web browser and version

135.0.7049.115 (Official Build) (64-bit) (cohort: Stable) 

### Operating system

Windows 10

### Steps to reproduce this

1. Open this sketch in Chrome: https://editor.p5js.org/SableRaf/sketches/zAXl3tNm5
3. Zoom the page in or out (e.g., Ctrl+ or Ctrl- / Cmd+ or Cmd-).
5. Click and drag the circle across the canvas.

```js
let x = 200;
let y = 200;

function setup() {
  createCanvas(400, 400);
}

function draw() {
  background(220);
  
  // The code below causes unexpected behavior in Chrome if you zoom in/out
  
  if(mouseIsPressed){
    x += movedX;
    y += movedY; 
  }
  
  // The code below works correctly regardless of zoom level
  
  // if(mouseIsPressed){
  //   x += mouseX - pmouseX;
  //   y += mouseY - pmouseY; 
  // }
  
  circle(x,y,10);
}
```

**Additional context**

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX#browser_compatibility):

> Warning: Browsers use different units for movementX and screenX than what the specification defines. Depending on the browser and operating system, the movementX units may be a physical pixel, a logical pixel, or a CSS pixel. You may want to avoid the movement properties, and instead calculate the delta between the current client values (screenX, screenY) and the previous client values.

Related W3C issue: [https://github.com/w3c/pointerlock/issues/42](https://github.com/w3c/pointerlock/issues/42)

**Suggestion**
`movedX` and `movedY` should be updated internally to calculate based on `mouseX - pmouseX` and `mouseY - pmouseY`, rather than relying on `movementX` and `movementY`.

IIITM-Jay:
Resolves #7790 

### Approach Used:
- `_updateMouseCoords()` is called at the beginning of the existing function to store the previous frame’s positions.
- After computing the new `mouseX`, `mouseY`, the `movedX` and `movedY` values are derived in terms of `deltaX` and `deltaY` using `pmouseX` and `pmouseY` values and the current values of X and Y.

### Alternative Approach:
Instead of calling `_updateMouseCoords()` at the beginning , we could use variables to store previous values which could then be used to find delta.

```js
// Store previous mouseX and mouseY for calculating movedX/Y
    const prevMouseX = this.mouseX;
    const prevMouseY = this.mouseY;

// Calculate movedX/Y based on delta between current and previous positions
    const deltaX = mousePos.x - prevMouseX;
    const deltaY = mousePos.y - prevMouseY;

```

But, In my opinion there is no point in using extra variable to track the previous values. So, to avoid this manual tracking and leveraging the existing function, used the approach given earlier.

### Observations:
_**Previously, the behaviour was:**_

[Screencast from 07-05-25 12:12:39 AM IST.webm](https://github.com/user-attachments/assets/cccdb2bf-9a9f-422e-9e64-ab26cf55de4a)

**_Now, after fixing,_**

[Screencast from 07-05-25 12:08:48 AM IST.webm](https://github.com/user-attachments/assets/ace08fb4-db1b-415e-b3b3-83f96fc2661c)

### Browser Tested:
- Chrome
- Firefox

- [x] `npm run lint` passes

### DIFF
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -819,8 +819,6 @@ p5.prototype.mouseIsPressed = false;
 
 p5.prototype._updateNextMouseCoords = function(e) {
   if (this._curElement !== null && (!e.touches || e.touches.length > 0)) {
-    this._updateMouseCoords();
-
     const mousePos = getMousePos(
       this._curElement.elt,
       this.width,
--- a/src/events/mouse.js
+++ b/src/events/mouse.js
@@ -819,21 +819,27 @@ p5.prototype.mouseIsPressed = false;
 
 p5.prototype._updateNextMouseCoords = function(e) {
   if (this._curElement !== null && (!e.touches || e.touches.length > 0)) {
+    this._updateMouseCoords();
+
     const mousePos = getMousePos(
       this._curElement.elt,
       this.width,
       this.height,
       e
     );
-    this._setProperty('movedX', e.movementX);
-    this._setProperty('movedY', e.movementY);
+
     this._setProperty('mouseX', mousePos.x);
     this._setProperty('mouseY', mousePos.y);
     this._setProperty('winMouseX', mousePos.winX);
     this._setProperty('winMouseY', mousePos.winY);
+
+    const deltaX = this.mouseX - this.pmouseX;
+    const deltaY = this.mouseY - this.pmouseY;
+    this._setProperty('movedX', deltaX);
+    this._setProperty('movedY', deltaY);
   }
+
   if (!this._hasMouseInteracted) {
-    // For first draw, make previous and next equal
     this._updateMouseCoords();
     this._setProperty('_hasMouseInteracted', true);
   }


IIITM-Jay:
@SableRaf May I request you to have a look at this PR if this is what actually expected and the solution is aligned to the expected behaviour. Let me know if some changes are required, or needed some more modifications.

CC: @ksen0 , @limzykenneth , requesting some view points and feedback on the same.

SableRaf:
Hi @IIITM-Jay, and thanks for taking the time to work on this!

Just to clarify: I opened https://github.com/processing/p5.js/issues/7790 to invite discussion and give maintainers a chance to confirm an approach before any work began. Following the [contributor guidelines](https://github.com/processing/p5.js/tree/main/contributor_docs), we usually ask contributors to wait for the issue to be confirmed, an implementation is agreed on, and the issue is assigned before opening a PR.

In any case, I’m personally not the best person to review this PR, so I’ll leave space for others who are more familiar with this part of the project to chime in.

Thanks again for your interest and energy in contributing!

IIITM-Jay:
Thanks @SableRaf for the response!

Yeah, I felt that this approach might be a bit more readable and maintainable, and it seemed to integrate well with the existing p5.js structure — especially since it avoids extra tracking variables by reusing `_updateMouseCoords()` which already handles previous mouse positions.

I thought it could be helpful to open a PR just to propose one of possible solution and make the discussion a bit more concrete. That said, yes, it's always a good practice to share multiple idea among maintainers and collaborators — I’ll share a few alternate approaches too in the issue so we can compare and decide on the most suitable one.

Thanks a lot again!

IIITM-Jay:
Hi @ksen0 , @limzykenneth May I request for feedback and review on this PR to know whether I am on a right rack to solve this issue. I have discussed few points on  the issue as well. Would love to incorporate the feedback changes too.
Thanks!

ksen0:
Hi @IIITM-Jay thank you for your detailed work on this! A couple of things we saw in review:

* Because your proposed code is updating coords without checking for interaction flag, could you check if [this regression](https://github.com/processing/p5.js/pull/1471) has occurred? Because
* Right now the code logic seems to be: update pmouseX/Y (in the update coords call); then update mouseX/Y; then update pmouseX/Y a second time. So Could you confirm that pmouseX/Y are not getting overwritten here?

Because testing this involves interaction and time, it's not something that is covered well by 1.x unite tests. Ift would be really helpful if you include your manual tests (with visuals, preferably) in the PR, so that in the future it's more possible to manually spot regressions if they happen!

Please let me know if I can clarify anything.

src/events/mouse.js
```diff
@@ -819,21 +819,27 @@ p5.prototype.mouseIsPressed = false;
 
 p5.prototype._updateNextMouseCoords = function(e) {
   if (this._curElement !== null && (!e.touches || e.touches.length > 0)) {
+    this._updateMouseCoords();
```
limzykenneth, line 822: Just to add to above from @ksen0, I have tested this locally and I can't find any difference between having having this line and not having this line. Is there something specific this is aiming to solve with this line?

limzykenneth (commented):
In addition to the inline comment, I find that there is another issue with Chrome, regardless of zoom issue which is when the example sketch used by @SableRaf in the original issue is view in Chrome, refreshing the page while not moving the mouse cursor then clicking on the page caused unintended movement on the circle in the middle.


https://github.com/user-attachments/assets/18f4c560-ba64-4c2c-9f8b-4693defc9f27

It may be unrelated to this issue though so we don't need to solve it in this PR. However both the above issue and the original issue for this PR does not present in 2.x from my testing, we are using `pointerevents` instead of `mouseevents` in 2.x so that may be the reason.

src/events/mouse.js
IIITM-Jay: Thanks @limzykenneth and @ksen0 for reviewing this PR.

I would like to address the concerns as stated above:

- Yes the regression for #1471 does not show up. The reason is that after the mouse is already being interacted, the latter if block  won't get executed as the variable became true and the concern of overwritten got  cleared, though I checked using console logs and it worked as expected without overwritten. So, I guess, @ksen0 concerns of overwritten values for regression of #1471 is being resolved with this.
- Secondly, the concern from @limzykenneth of calling `this._updateMouseCoords()` because of two reasons, firstly, The `if (!this._hasMouseInteracted)` block runs only once; It is a first-time setup. After first interaction, `this._hasMouseInteracted = true`, and the block will never run again. So in all subsequent frames, that block is skipped. But even after the first interaction, `mouseX` keeps changing — so we still need to update `pmouseX` before setting new `mouseX` on every frame, not just the first. The key reason is - to ensure that `movedX = mouseX - pmouseX` (and `movedY`) is accurate, `pmouseX` must first be updated to the previous value of `mouseX` before we assign a new value to `mouseX`.
   
Secondly, when we carefully observe the written codes, the `_updateMouseCoords()` inside `_updateNextMouseCoords` block is very much late. We have the following order of execution when interaction happens:
```js
if (this._curElement !== null && (!e.touches || e.touches.length > 0)) {
  // setting mouseX directly
  this._setProperty('mouseX', newValue);
  ...
}

// now we check first interaction
if (!this._hasMouseInteracted) {
  this._updateMouseCoords();  // too late! mouseX has already changed
  ...
}
```
In this flow, we set `mouseX` first _(e.g., to 100)_. Then later call `_updateMouseCoords()` which sets `pmouseX = 100` — the new value! So `pmouseX === mouseX`, making `movedX = 0`. **This is wrong**. What we indeed wanted is 
**Before overwriting `mouseX`, store the old value in `pmouseX`:**

```js
this._updateMouseCoords(); // saves current mouseX -> pmouseX
this._setProperty('mouseX', newValue);
```
So, even if `_updateMouseCoords()` is called afterwards in the `!_hasMouseInteracted` block, it’s too late to help with delta calculation, and with this way, it always gurantees the correct movedX

ksen0:
Thanks so much for your thorough work here @IIITM-Jay !

> would you want me to write an example test under /test/manual-test-examples/mouse-events possibly with index.html and sketch.js

I don't think that's strictly necessary, since this is in 1.x only. Because this PR is already very detailed, and can be found with `git blame`, it is enough. @limzykenneth do you think this needs formal manual test addition?

src/events/mouse.js
limzykenneth: > But even after the first interaction, mouseX keeps changing — so we still need to update pmouseX before setting new mouseX on every frame, not just the first. The key reason is - to ensure that movedX = mouseX - pmouseX (and movedY) is accurate, pmouseX must first be updated to the previous value of mouseX before we assign a new value to mouseX.

`pmouseX` is already updated in the main draw loop [here](https://github.com/processing/p5.js/blob/a990d4f153e7fd5cb489c2d4507f175ec7a2fc8e/src/core/main.js#L536). Calling it in this function here repeatedly caused the behavior of `pmouseX` to change, namely `pmouseX` is supposed to be the position of the mouse in the **previous frame** not the previous firing of the event. If this property makes `pmouseX` insufficient to solve the issue here, we shouldn't use it but to keep track separately.

limzykenneth:
@ksen0 @IIITM-Jay For now it's fine, the example in the original issue seems enough to test this case.

src/events/mouse.js
limzykenneth: In the first example in #1471, if we reduce the frame rate to say 10, without this addition we get this:
Screenshot 2025-05-13 at 11-55-45 P5 test

With this addition, we get this:
Screenshot 2025-05-13 at 11-56-01 P5 test

IIITM-Jay: Ah I see!

> pmouseX is already updated in the main draw loop [here](https://github.com/processing/p5.js/blob/a990d4f153e7fd5cb489c2d4507f175ec7a2fc8e/src/core/main.js#L536). Calling it in this function here repeatedly caused the behavior of pmouseX to change, namely pmouseX is supposed to be the position of the mouse in the previous frame not the previous firing of the event. 

I agree with above @limzykenneth , but I also checked consoling it is not repeatedly changing but it's ok to remove that line (perhaps to remove redundancy), so yes, removing that line and committing again.

IIITM-Jay:
Thanks a lot once again @limzykenneth and @ksen0 for giving time in reviewing it and giving feedback.

As per our detailed discussion as above, I have removed the the first calling of this._updateMouseCoords() as it does not add much to any noticeable change and other parts of the code already does that. I have incorporated the feedback and requested a review again.

Let me know if something else needs to be done for this PR.

Thanks once again!

PR #7808, opened 2025-05-13
Fix: keyTyped() can now accept the same char consecutively

AhmedMagedC:
Resolves #7805

 Changes:
 instead of completely ignore the key when it equals to the previous key, ensure first it's repeated (holding down) then you can ignore it.
 ```javascript
if (e.which === this._lastKeyCodeTyped && this.isKeyRepeated) {
    // prevent multiple firings
    return;
  }
  ```
why not just using `e.repeat` instead of setting `isKeyRepeated` property?
because the `repeat` property in the `keypress` event always equals to false.

#### PR Checklist

- [x] `npm run lint` passes
- [x] `npm test` passed
- [ ] [Unit tests] are included / updated

PR #7809, opened 2025-05-13
Fixing keyTyped() for dev-2.0 branch 

### LINKED ISSUE #7805: keyTyped function ignores last key typed

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [x] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

1.11.5 & 2.0.1

### Web browser and version

_No response_

### Operating system

Mac OSX 15.4.1

### Steps:
Run the code below and type several characters. They will be displayed but it will ignore a key if it has the same ASCII code as the previous  key typed.
For instance type `abcdddddefgGdh` will display
`a b c d e f g G d h`

Note that only the first `d` is displayed, repeats are ignored but can be displayed  later if a different key is typed first.

### Snippet:
```js
let s = '>';

function setup() {
  createCanvas(600, 300);
  textSize(16);
}

function keyTyped(){
  s = s + ' ' + key;
}

function draw() {
  background(240); noStroke();  fill(0, 0, 160);
  text(s, 10, 10, width - 20, height - 20);
}

```

perminder-17:
Resolves #7805 

 Changes:

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/events/keyboard.js
+++ b/src/events/keyboard.js
@@ -663,7 +663,6 @@ function keyboard(p5, fn){
 
   };
 
-
   /**
    * A function that's called once when keys with printable characters are pressed.
    *
@@ -795,7 +794,7 @@ function keyboard(p5, fn){
    * </div>
    */
   fn._onkeypress = function(e) {
-    if (e.which === this._lastKeyCodeTyped) {
+    if (e.which === this._lastKeyCodeTyped && e.repeat) {
       // prevent multiple firings
       return;
     }


AhmedMagedC:
Hi, [@perminder-17](https://github.com/perminder-17)
may i know why did you choose to use `e.repeat`? as i said in my [PR](https://github.com/processing/p5.js/pull/7808) `e.repeat` always results in False value, the main solution was to allow for the same char to be pressed consecutively more than once while preventing from printing the same char if the user holds down the key.

AhmedMagedC:
Hi [@perminder-17](https://github.com/perminder-17), the issue is that `e.repeat` results in false even if you hold down the key,

but when i tried to print the `e.repeat` value in ```fn._onkeydown = function(e) {}``` `e.repeat` gives true if you hold down, but it gives always false for ```fn._onkeypress = function(e) {}```

i tested on chrome & avast & Microsoft edge

perminder-17:
>  e.repeat value in fn._onkeydown = function(e) {} e.repeat gives true if you hold down, but it gives always false for fn._onkeypress = function(e) {}

Okay, thanks for reporting. 

But I also checked on firefox and chrome (at linux) and everything was working correctly as expected. This looks like could be a browser specific issue. Someone previously had a similar issue with this.  : https://stackoverflow.com/questions/53200698/onkeypress-repeat-always-returns-false-in-some-browsers

Workaround could be, maybe we will move `e.repeat` in `fn.onkeyDown` and then probably use that in `fn.onkeyPress`. 
Do you agree with this @davepagurek or could be a different workaround for it. I think its a browser specific issue, are you also able to reproduce the same in your system as well?

Issue #7813, opened 2025-05-14
FES should fetch translations within the same minor issue


ksen0:
### Increasing access

FES internationalization improvement to match 1.x/2.x difference

### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [ ] Other (specify if possible)

### Feature enhancement details

Currently, FES fetches translations using latest: https://github.com/processing/p5.js/blob/5f0fcfac9d03a5ec2fee271a98fa1d1edb70ec68/src/core/internationalization.js#L153

Which results in a. 1.x sketch getting 2.x FES translations, which may be inaccurate. Should fetch within the minor version, but latest patch, for example: if the sketch is 1.11.1, it should get language for 1.11.7 (or whatever is latest)

This should be patched first in 1.x, and then a matching 2.x

Satish-Medar:
@ksen0 I’d like to work on this issue. Could you please assign it to me?

PR #7878, opened 2025-06-04
Use FES internationalization within the minor version

ksen0:
Addresses #7813

- [x] `npm run lint` passes

```js
const version = '1.11.7';
const loadPath = 'https://cdn.jsdelivr.net/npm/p5@' + version.replace(/^(\d+\.\d+)\.\d+.*$/, '$1') + '/translations/{{lng}}/{{ns}}.json';
console.log(loadPath);
// https://cdn.jsdelivr.net/npm/p5@1.11/translations/{{lng}}/{{ns}}.json
// will become https://cdn.jsdelivr.net/npm/p5@1.11/translations/es/translation.json which exists
```

### DIFF
--- a/src/core/internationalization.js
+++ b/src/core/internationalization.js
@@ -1,6 +1,6 @@
 import i18next from 'i18next';
 import LanguageDetector from 'i18next-browser-languagedetector';
-import * as constants from './constants';
+import { VERSION } from './constants';
 
 
 let fallbackResources, languages;
@@ -129,7 +129,6 @@ export let translator = (key, values) => {
  * Set up our translation function, with loaded languages
  */
 export const initialize = () => {
-  let latestMinorVersionPath = 'https://cdn.jsdelivr.net/npm/p5@' + constants.VERSION.replace(/^(\d+\.\d+)\.\d+.*$/, '$1');
   let i18init = i18next
     .use(LanguageDetector)
     .use(FetchResources)
@@ -152,7 +151,12 @@ export const initialize = () => {
       },
       backend: {
         fallback: 'en',
-        loadPath: latestMinorVersionPath + '/translations/{{lng}}/{{ns}}.json'
+
+        // ensure that the FES internationalization strings are loaded
+        // from the latest patch of the current minor version of p5.js
+        loadPath: `https://cdn.jsdelivr.net/npm/p5@${
+          VERSION.replace(/^(\d+\.\d+)\.\d+.*$/, '$1')
+        }/translations/{{lng}}/{{ns}}.json`
       },
       partialBundledLanguages: true,
       resources: fallbackResources


src/core/internationalization.js
```diff
@@ -1,5 +1,7 @@
 import i18next from 'i18next';
 import LanguageDetector from 'i18next-browser-languagedetector';
+import { version } from '../../package.json';
+
```
limzykenneth: Browserify used by the 1.x branch does not treeshake JSON which means the whole `package.json` file will be included in the build with this. We have the version number exported by [`core/constants.js`](https://github.com/processing/p5.js/blob/main/src/core/constants.js#L14-L15) so we can import that and use that value directly instead.

limzykenneth (commented):
Sorry this took awhile to get to. One suggestion on alternative approach.

src/core/internationalization.js
ksen0: Thanks! Updated.

src/core/internationalization.js
```diff
@@ -127,6 +129,7 @@ export let translator = (key, values) => {
  * Set up our translation function, with loaded languages
  */
 export const initialize = () => {
+  let latestMinorVersionPath = 'https://cdn.jsdelivr.net/npm/p5@' + constants.VERSION.replace(/^(\d+\.\d+)\.\d+.*$/, '$1');
```
IIITM-Jay: Hi @ksen0 , @limzykenneth just a thought here.
Will it be a better idea  to separate out the base `CDN` URL here and the version parsing logic for better readability and reusability too, and then compose the final `loadPath`. I was thinking to encapsulate it like below:
```js

const getCDNPath = (() => {
  const base = 'https://cdn.jsdelivr.net/npm/p5@';
  const version = constants.VERSION.replace(/^(\d+\.\d+)\.\d+.*$/, '$1');
  return `${base}${version}`;
})();
```

src/core/internationalization.js
```diff
@@ -149,8 +152,7 @@ export const initialize = () => {
       },
       backend: {
         fallback: 'en',
-        loadPath:
-          'https://cdn.jsdelivr.net/npm/p5/translations/{{lng}}/{{ns}}.json'
+        loadPath: latestMinorVersionPath + '/translations/{{lng}}/{{ns}}.json'
```
IIITM-Jay: and then here,  we could compose it like
```js
loadPath: `${getCDNPath}/translations/{{lng}}/{{ns}}.json`
```

src/core/internationalization.js
limzykenneth: I'm not sure it's worth to split it out in this way, especially also encapsulating it in a function, if needed, just a string interpolation should suffice as well.
```js
`https://cdn.jsdelivr.net/npm/p5@${constants.VERSION.replace(/^(\d+\.\d+)\.\d+.*$/, '$1')}`
```

IIITM-Jay: Yes, I see that there's no broader scope for reusability in this case, so I agree — using direct string interpolation is sufficient and keeps it clean.

IIITM-Jay: My initial thought was that separating the base URL and version logic might help with readability and clarity, especially if more dynamic paths were ever added later. But given the current usage, simplicity could be more prioritized. 

Thanks for sharing this!

src/core/internationalization.js
IIITM-Jay: and then I think with string interpolation we could simply write  with one liner change:
```js
loadPath: `https://cdn.jsdelivr.net/npm/p5@${constants.VERSION.replace(/^(\d+\.\d+)\.\d+.*$/, '$1')}/translations/{{lng}}/{{ns}}.json`
```

PR #7902, opened 2025-06-11
Adding docs for `code` in the refrence.

### LINKED ISSUE #7881: [p5.js 2.0 Bug Report]:  Add documentation for the `code` system variable in key handling.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [x] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.x

### Web browser and version

Firefox

### Operating system

Linux

### Steps to reproduce this

In p5.js, we currently support some ways to detect which key a user presses: `keyCode`, `key`, and `code` are some of it. While `keyCode` and `key` are documented in the reference, the `code` variable has not yet been documented, which can lead to confusion for users.

https://github.com/processing/p5.js/blob/c39ad5c452d773a18de4c03067a7315422442eb4/src/events/keyboard.js#L463-L465

- `key` returns the character or value associated with the key press, regardless of the keyboard layout.

- `code` returns the Constants on the keyboard (e.g., ENTER, ArrowUP), independent of what character is printed on that key.

Because only `key` is documented, users do not know that `code` exists or how it behaves.

perminder-17:
Resolves #7881 

 Changes:
Added documentation and examples for `code` function. 

#### PR Checklist

- [ ] `npm run lint` passes


src/events/keyboard.js
```diff
@@ -195,6 +194,117 @@ function keyboard(p5, fn){
...
+   *
+   * Unlike <a href="#/p5/key">key</a>, the `code` property differentiates between 
+   * physical keys that generate the same character—for example, `CtrlLeft` and 
+   * `CtrlRight`—so each can be handled independently.
+   * 
+   * Pressing the key physically labeled “A” always yields `KeyA`, regardless 
+   * of the current keyboard layout (QWERTY, Dvorak, AZERTY, etc.) or the character 
+   * that appears in a text field.
+   * 
+   * The code property returns a plain string (e.g. 'ArrowRight'), you can
```
IIITM-Jay: Hi @perminder-17 ,
just a small thing - `(e.g. 'ArrowRight')`  to be `(e.g., 'ArrowRight')` a comma `(,)` after the abbreviation, for stating eaxmple and probably a full-stop after it `(e.g. 'ArrowRight'),` --> `(e.g., 'ArrowRight'). You can....`

perminder-17: Thanks for catching this. Do you think the PR looks good to you now?

src/events/keyboard.js
```diff
@@ -195,6 +194,117 @@ function keyboard(p5, fn){
...
+   *
+   *   // Display the last key pressed.
+   *   text(code, 50, 50);
+   * }
+   * </code>
+   * </div>
+   *
+   * <div>
+   * <code>
+   * // Click on the canvas to begin detecting key presses.
```
ksen0, line 335: Hi @perminder-17 ! Apologies for the delay on this. What do you think about linking to the code docs (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) at the end of the description - and also adding an example that's inspired by their interactive one? (Or by this one https://www.toptal.com/developers/keycode that we have linked before). Something that shows each of the values of `key` and `code`? Here's a sketch I made: https://editor.p5js.org/ksen0/sketches/oxgLK9HIV you're welcoem to use it as a starting point if you think it's a good idea. That sketch is actually completely backwards compatible except for the line with `code` so maybe it could also be added to the 1.x reference? 

Lastly, while reviewing the transition guide I suggested an overview table https://github.com/processing/p5.js-compatibility/pull/27/files#r2156196242 - maybe the 2.x reference could use the info in just the second column, if you think it improves clarity?

src/events/keyboard.js
```diff
@@ -195,6 +194,117 @@ function keyboard(p5, fn){
...
+   * property returns a value that isn't altered by keyboard layout or the state 
+   * of the modifier keys.
+   * 
+   * This property is useful when you want to handle keys based on their 
+   * physical positions on the input device rather than the characters associated 
+   * with those keys;
+   *
+   * Unlike <a href="#/p5/key">key</a>, the `code` property differentiates between 
+   * physical keys that generate the same character—for example, `CtrlLeft` and 
+   * `CtrlRight`—so each can be handled independently.
```
ksen0, line 209: I think linking to the mozilla code docs at the end of this would be useful?

src/events/keyboard.js
```diff
@@ -195,6 +194,117 @@ function keyboard(p5, fn){
...
+   * 
+   * Pressing the key physically labeled “A” always yields `KeyA`, regardless 
+   * of the current keyboard layout (QWERTY, Dvorak, AZERTY, etc.) or the character 
+   * that appears in a text field.
+   * 
+   * The code property returns a plain string (e.g., 'ArrowRight'). You can 
+   * compare it directly with string literals:
+   * ```js
+   *  if (code === 'ArrowRight') {
```
ksen0: I think because `keyIsDown` is a backwards-compatible pattern with the system variables, we should only show that in the example code, what do you think?

And explain that it's the same as `code === 'ArrowRight'` and `key === 'ArrowRight'`

src/events/keyboard.js
```diff
@@ -195,6 +194,117 @@ function keyboard(p5, fn){
...
+   *  // …
+   * }
+   * ```
+   * 
+   * For extra clarity, you can instead use the predefined key-code constants
```
ksen0: I think the way these are introduced and described in the previous docs (https://p5js.org/reference/p5/keyCode/) is really concise, maybe this can be used instead? And not make it sound like an optional/extra, but rather the recommended way to use keyboard events?

src/events/keyboard.js
```diff
@@ -195,6 +194,117 @@ function keyboard(p5, fn){
...
+   *
+   *   background(200);
+   *
+   *   describe(
+   *     'A gray square with a black circle at its center. The circle moves when the user presses an arrow key. It leaves a trail as it moves.'
+   *   );
+   * }
+   *
+   * function draw() {
+   *   // Update x and y if an arrow key is pressed.
```
ksen0, line 351: What do you think about using `keyIsDown` here as well? https://editor.p5js.org/ksen0/sketches/vrMANhunx Main benefit: backwards compatible snippet. Main drawback: makes this reference page less standalone / use another function. Up to you if you think it's a good idea!

ksen0 (approved):
I've added some comments, please feel free to update based on the feedback where you think it's appropriate. Thank you for adding this!

src/events/keyboard.js
perminder-17: Can you just confirm that I need to add this table

| variable | p5.js 1.x | p5.js 2.x |
|--|--|--|
| `key` | Text String (e.g.: "ArrowUp") | Text String (e.g.: "ArrowUp", "f" or "F") |
| `code` | Not supported | Text String (e.g.: "ArrowUp", "KeyF")  |
| `keyCode` | Number | Number (e.g., 70) |
|  System variables | Number | Text String (e.g: "ArrowUp") | 

in our reference of `code`?

ksen0: @perminder-17 if you agree that it would be a helpful overview, then yes please, unless something speaks against it?

ksen0 (approved):
Hi @perminder-17, sorry for the delay on this. Please feel free to merge; you can add the table or not depending on your judgment if you think it's useful!

PR #7930, opened 2025-06-22
[p5.js 2.x] chore: enable eslint rules

error-four-o-four:
Hey there,

this PR is a follow-up to #7853 to discuss the usage of linter rules.

It does not include the fixed/applied changes of the rules because it would block other PRs or would result in a lot of rebasing and resolving merge conflicts. I'd argue that the fixes should be applied in separate PRs.

(Discussed rules were marked as completed)

## [eslint recommended](https://eslint.org/docs/latest/rules/)

* [ ] [constructor-super](https://eslint.org/docs/latest/rules/constructor-super) error
* [ ] [for-direction](https://eslint.org/docs/latest/rules/for-direction) error
* [ ] [getter-return](https://eslint.org/docs/latest/rules/getter-return) error
* [ ] ~~[no-async-promise-executor](https://eslint.org/docs/latest/rules/no-async-promise-executor) error~~
* [ ] [no-class-assign](https://eslint.org/docs/latest/rules/no-class-assign) error
* [ ] [no-compare-neg-zero](https://eslint.org/docs/latest/rules/no-compare-neg-zero) error
* [ ] ~~[no-cond-assign](https://eslint.org/docs/latest/rules/no-cond-assign) error~~
* [ ] [no-const-assign](https://eslint.org/docs/latest/rules/no-const-assign) error
* [ ] [no-constant-binary-expression](https://eslint.org/docs/latest/rules/no-constant-binary-expression) error
* [ ] [no-constant-condition](https://eslint.org/docs/latest/rules/no-constant-condition) error
* [ ] [no-control-regex](https://eslint.org/docs/latest/rules/no-control-regex) error
* [ ] [no-debugger](https://eslint.org/docs/latest/rules/no-debugger) error
* [ ] [no-dupe-args](https://eslint.org/docs/latest/rules/no-dupe-args) error
* [ ] [no-dupe-class-members](https://eslint.org/docs/latest/rules/no-dupe-class-members) error
* [ ] [no-dupe-else-if](https://eslint.org/docs/latest/rules/no-dupe-else-if) error
* [ ] [no-dupe-keys](https://eslint.org/docs/latest/rules/no-dupe-keys) error
* [ ] [no-duplicate-case](https://eslint.org/docs/latest/rules/no-duplicate-case) error
* [ ] [no-empty-character-class](https://eslint.org/docs/latest/rules/no-empty-character-class) error
* [ ] [no-empty-pattern](https://eslint.org/docs/latest/rules/no-empty-pattern) error
[...]

PR #7933, opened 2025-06-22
[p5.js 2.x] chore: enable `pre-commit` hooks

error-four-o-four:
adresses #7930 

This PR enables the pre-commit hooks to make use of `lint-staged`.

As a one time fix this PR uses `npx sort-package-json` to sort the fields in the  package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,10 +53,12 @@
     "test": "vitest"
   },
   "husky": {
-    "hooks": {}
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
   },
   "lint-staged": {
-    "src/**/*.js": "eslint",
+    "src/**/*.js": "eslint --max-warnings=0",
     "test/**/*.js": "eslint",
     "utils/**/*.{js,mjs}": "eslint"
   },
--- a/package.json
+++ b/package.json
@@ -1,24 +1,70 @@
 {
   "name": "p5",
+  "version": "2.0.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,24 +1,72 @@
...
   "name": "p5",
+  "version": "2.0.3",
+  "description": "[![npm version](https://badge.fury.io/js/p5.svg)](https://www.npmjs.com/package/p5)",
+  "homepage": "https://p5js.org",
+  "bugs": {
+    "url": "https://github.com/processing/p5.js/issues"
+  },
   "repository": "processing/p5.js",
+  "license": "LGPL-2.1",
+  "author": "",
```
ksen0, line 11: As far as I understand, the "author" field is a best practice to include but optional. Because this project follows the All Contributors specification, I think it could either be omitted or using something like "p5.js contributors"? Any thoughts on this? Minor preference against adding this as a blank field.

ksen0 (commented):
Thanks for working on this @error-four-o-four , sorry for delay in review. Only 1 minor comment re:authorship. Tagging also other DevOps stewards: @Vaivaswat2244 @limzykenneth @lirenjie95

package.json
limzykenneth: The author field here is not quite added but just rearranged from elsewhere in this file, ie. it is there already before as a blank string. I'm not sure this field is read by anything specific (that I know of) to populate something so I think something like "p5.js contributors" is fine either way.

ksen0: Ah Ok I saw other rearranged fields but missed this one. Sorry!

PR #7967, opened 2025-07-13
Individual flags to disable part of FES by the user

limzykenneth:
Addresses https://github.com/ffd8/P5LIVE/issues/117

 Changes:

Adds two new flags to selectively disable specific components of FES.

* `p5.disableParameterValidator` - disables parameter validation check
* `p5.disableSketchChecker` - disables sketch overwrite global variables check

The implementation is not very sophisticated at this point but should be functional and we can refactor later. The names of the flags are still up for suggestion if any.


ffd8:
That's great! In the meantime, I see that this message about overwriting global variables seems to have completely gone away (I don't see anything in 2.2.3 and x versions earlier). Thanks for adding this temp fix before it ultimately got removed!

PR #7969, opened 2025-07-13
Add integration tests for ESM build (typescript/react/vuejs)

### LINKED ISSUE #7938: [p5.js 2.x] add integration tests
### Topic

Hey there,

in order to push the state of the esm build forward I'd suggest to add some kind of integration tests. The main issue I'm trying to tackle is to be able to check correct type declarations. At a later stage one could also add separate tests for each addon/build which is defined in the `exports` field of the package.json. Beyond that one could test the global or instance mode in other frameworks such as react or vuejs if neccessary.

To be more specific I'm thinking of running a child process like `tsc --noEmit` within a test folder/project with different compiler options and setups (unless there are better options).

```
├── test
│   ├── integrations
│   │   ├── typescript
│   │   │   ├── package.json
│   │   │   ├── index.ts
│   │   │   └── tsconfig.json
│   │   ├── react
│   │   └── vuejs
│   ├── unit
│   └── visual
├── package.json
└── <root>
```

```json
// ./test/integrations/typescript/package.json
{
  "name": "ts-test",
  "scripts": {
    "test": "tsc --noEmit",
  },
  "devDependencies": {
    "p5": "file:../../../dist/app.js",
    "typescript": "^5.0.0"
  }
}
```

```ts
// ./test/integrations/typescript/index.ts
import p5 from 'p5';

new p5((p) =>{
  p.setup = () => {
    // ...
  }
})
```

```js
import { execSync } from 'node:child_process';

process.chdir('<root>/test/integration/typescript');
execSync('npm i');
execSync('npm run test');
```

I'd recommend to run such tests only locally and not integrate them to the CI.

---

Another approach would be to use a library like [mockFs](https://www.npmjs.com/package/mock-fs) or [fs-fixture](https://github.com/privatenumber/fs-fixture). They programmatically create test files in a temp folder and delete them afterwards.

The general benefit would be a streamlined test experience.

A prerequisite would be to migrate to esm completely in order to settle the issues mentioned in #7863 

Any of this is open for discussion.

pratham-radadiya:
Add integration test

resolves #7938

Changes:

Introduced integration test setup for TypeScript, React, and Vuejs using ESM build

TypeScript tests validate declaration accuracy using tsc --noEmit

React tests check usage in a TSX component with hooks and class-based sketches

Vue tests use vue-tsc to verify type compatibility in .vue files

- [ ] `npm run lint` passes


test/integrationRunner.js
```diff
@@ -0,0 +1,23 @@
+const { execSync } = require('node:child_process');
+const { join } = require('node:path');
+const { existsSync } = require('node:fs');
+
+const testDirs = ['typescript', 'react', 'vuejs'];
```
lirenjie95, line 5: I think we can consider reading all directories under `test/integrations` dynamically instead of hard code.

ksen0:
Hi @pratham-radadiya thanks for working on this! I see you've marked a review comment resolved, do you plan to implement that suggestion in a future commit? In terms of scope: it looks like the top-level package and package-lock files were touched, could you confirm why this was necessary please?

Also tagging the other DevOps stewards on this one @Vaivaswat2244 @error-four-o-four @limzykenneth

limzykenneth:
I personally don't quite like the setup here introducing individual standalone project which adds a lot of complexity and the `exec` stuff can be very brittle as well. 

I don't think there is a specific need to test React and Vue, the main thing from the original issue is looking at is around type correctness, React and Vue don't necessarily introduces different ways of working with types here. The original issue mention Vitest's [type testing](https://vitest.dev/guide/testing-types) I think is a better direction. I would like our tests to remain within Vitest's ecosystem as much as possible so we don't build out toolings that are somewhat brittle and over time didn't get maintained and used, which is a problem we had with the 1.x devops setup.

pratham-radadiya:
Hi @ksen0  and @limzykenneth, thank you both for the detailed feedback!

On the package.json and package-lock.json changes:
These changes were required to include dependencies like vue-tsc, react, and typescript for the integration tests to run correctly in each ESM project. However, I understand now that this introduces unnecessary complexity at the top level. I’ll move these dependencies into local package.json files inside the test subfolders instead, or clean them up if we decide to drop Vue/React tests.

 On the scope of React/Vue testing:
You're right — the original issue (#7938) was focused on verifying type correctness, and I see now that React/Vue may not bring extra value here. I can remove the React and Vue test setups and instead focus on tsc + Vitest’s type-only testing approach as suggested. That would keep things consistent with the existing testing strategy and avoid brittle tooling.

I'll push an updated commit shortly reflecting these changes. Please let me know if you'd like me to go in a slightly different direction — happy to adjust!

PR #7980, opened 2025-07-20
Fix mediaElement.copy

pearmini:
Resolves https://github.com/ml5js/ml5-next-gen/pull/258

 Changes:

I'm currently working on updating ml5 to support p5.js version 2.0. However, the following example breaks during inspection:

```js
let bodySegmentation;
let video;
let segmentation;

let options = {
  maskType: "background",
};

function preload() {
  bodySegmentation = ml5.bodySegmentation("SelfieSegmentation", options);
}

function setup() {
  createCanvas(640, 480);
  // Create the video
  video = createCapture(VIDEO);
  video.size(640, 480);
  video.hide();

  bodySegmentation.detectStart(video, gotResults);
}

function draw() {
  background(0, 0, 255);
  if (segmentation) {
    video.mask(segmentation.mask);
    image(video, 0, 0);
  }
}

// callback function for body segmentation
function gotResults(result) {
  segmentation = result;
}

```

After debugging, I found that ·video.mask(segmentation.mask)` throws the following error:

```
ReferenceError: fn is not defined
```

image

So I updated `fn` to `p5.prototype` to fix it, based on the reference from the source code in the main branch.

 Screenshots of the change:

image

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/dom/p5.MediaElement.js
+++ b/src/dom/p5.MediaElement.js
@@ -816,7 +816,7 @@ class MediaElement extends Element {
   }
   copy(...args) {
     this._ensureCanvas();
-    fn.copy.apply(this, args);
+    p5.prototype.copy.apply(this, args);
   }
   mask(...args) {
     this.loadPixels();
--- a/test/unit/dom/p5.MediaElement.js
+++ b/test/unit/dom/p5.MediaElement.js
@@ -2,10 +2,12 @@ import { vi } from 'vitest';
 import { mockP5, mockP5Prototype } from '../../js/mocks';
 import { default as media, MediaElement } from '../../../src/dom/p5.MediaElement';
 import { Element } from '../../../src/dom/p5.Element';
+import { default as pixels } from '../../../src/image/pixels';
 
 suite('p5.MediaElement', () => {
   beforeAll(() => {
     media(mockP5, mockP5Prototype);
+    pixels(mockP5, mockP5Prototype);
     navigator.mediaDevices.getUserMedia = vi.fn()
       .mockResolvedValue("stream-value");
   });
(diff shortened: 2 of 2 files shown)


limzykenneth:
A reference to global `p5` is not always guaranteed (eg. when using ESM) so this is not quite general enough. Two possible options are either to duplicate the implementation in `src/image/pixels.js` or to make the implementation exportable and import into relevant place. The second option is preferable but we need to be careful to see if it introduces cyclic dependencies and if so extract that out into its own helper module.

pearmini:
@limzykenneth Thanks for the review! 

Just out of curiosity — I noticed that other MediaElement methods, such as mask, also reference the global p5 object:

```js
mask(...args) {
  this.loadPixels();
  this.setModified(true);
  p5.Image.prototype.mask.apply(this, args);
}
```

Do these methods have the same issue, or are they handled differently?

limzykenneth:
@pearmini Good spot, I think they should have the same issue as well.

pearmini:
@limzykenneth Since the other methods have the same issue, would it make sense to address them all in a separate PR? I suspect similar problems might exist elsewhere in the p5 codebase as well.

For this PR specifically, would it be possible to merge it as-is? At least, this change ensures that `MediaElement.copy` works correctly when using the UMD format. Although I can still import and use the `copy` method to fix this one, I’m not sure I’ll have time to fix all the related issues right now. What approach would you prefer?

limzykenneth:
@pearmini If you need a release soon with a temp fix we can merge this for now and see when we'll cut a patch release? @ksen0

pearmini:
@limzykenneth That would be great! We're planning to release support for p5's async setup soon!

PR #8033, opened 2025-08-17
Adding korean translation steward. 

perminder-17:


ksen0:
Thanks for your review on https://github.com/processing/p5.js-website/pull/931 and volunteering to support future Korean translation reviews @hana-cho!

PR #8047, opened 2025-08-24
Fix: call failureCallback if omggif fails to parse GIF in loadImage

### LINKED ISSUE #8021: loadImage with GIFs that can't be parsed throw an error that breaks the sketch

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [x] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] p5.strands
- [ ] WebGL
- [ ] Friendly Errors

### p5.js version

1.11.8

### Web browser and version

Firefox 141.0, Chromium 139.0.7258.66

### Operating system

Linux, macOS

### Steps:

When using `loadImage`, GIFs that can't be parsed by omggif throw an unhandleable exception that breaks further execution of the sketch.

Here's a sketch demonstrating the issue: https://editor.p5js.org/alexpls/sketches/81k-Dpv1-

I would expect that exceptions thrown while parsing the GIF would instead result in the `failureCallback` of `loadImage` to be called with their details.

I'd approach a fix for this by wrapping `new omggif.GifReader(arrayBuffer);` [here](https://github.com/processing/p5.js/blob/bde84ea01ae9355dbd66c8802c248ca95c548b15/src/image/loading_displaying.js#L649) with a try/catch, and handle the error gracefully.

If this sounds reasonable let me know, and I'd be happy to contribute a patch!

### Snippet:

The [sketch above](https://editor.p5js.org/alexpls/sketches/81k-Dpv1-) reproduces the issue reliably.

---

Thanks for your work on p5.js!

Satish-Medar:
Resolves #8021

Changes:

Wrapped the omggif.GifReader constructor in a try/catch block in loading_displaying.js.
If a GIF cannot be parsed, the failureCallback of loadImage is now called with the error, instead of throwing an uncatchable exception.
This allows sketches to handle image loading errors gracefully and prevents the sketch from crashing.
Tested locally with a broken GIF using a simple HTML test file (test-gif-error.html).
Screenshots of the change:
Screenshot 2025-08-24 111218

When loading a broken GIF, the failure callback is called and the sketch continues running (see console and page output in the test file).

PR Checklist
 npm run lint passes
 Inline reference is included / updated (not applicable)
 Unit tests are included / updated (manual test provided)

test-gif-error.html  ( p5.js  --> create [ test-gif-error.html  ] )

code in [[ test-gif-error.html ]]

<!DOCTYPE html>
<html>
  <head>
    <script src="lib/p5.js"></script>
    
    <script>
      function preload() {
        loadImage(
          "broken.gif", // Use a GIF that omggif cannot parse
          (img) => console.log("Loaded!", img),
          (err) => {
            console.error("Properly handled error:", err);
            document.body.innerHTML += "<p>Failure callback called!</p>";
          }
        );
      }
      function setup() {
        createCanvas(100, 100);
        background(200);
      }
    </script>
  </head>
  <body></body>
</html>

### DIFF
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -646,7 +646,20 @@ function _createGif(
   failureCallback,
   finishCallback
 ) {
-  const gifReader = new omggif.GifReader(arrayBuffer);
+  let gifReader;
+  try {
+    gifReader = new omggif.GifReader(arrayBuffer);
+  } catch (e) {
+    p5._friendlyFileLoadError(8, pImg.src);
+    if (typeof failureCallback === 'function') {
+      failureCallback(e);
+    } else {
+      console.error(e);
+    }
+    finishCallback();
+    return;
+  }
+
   pImg.width = pImg.canvas.width = gifReader.width;
   pImg.height = pImg.canvas.height = gifReader.height;
   const frames = [];


ThomasFinnern:
is calling "failureCallback(e); on failing gif ... safe ? Or does it enable a possible call of a not wanted function.
Just an offside comment with good intention. I am sorry if  it is on the wrong place here.

Satish-Medar:
@ThomasFinnern Thanks for raising this! failureCallback here is user-supplied (similar to the success callback), so calling it with the error object is consistent with how other loading functions behave in p5.js. It shouldn’t allow unwanted function calls since the user has to explicitly pass the callback. Still, I appreciate the point — good to double-check!

src/image/loading_displaying.js
```diff
@@ -646,7 +646,20 @@ function _createGif(
...
+  try {
+    gifReader = new omggif.GifReader(arrayBuffer);
+  } catch (e) {
+    p5._friendlyFileLoadError(8, pImg.src);
+    if (typeof failureCallback === 'function') {
+      failureCallback(e);
+    } else {
+      console.error(e);
+    }
+    finishCallback();
```
limzykenneth, line 659: I think for consistency sake with other loading functions we don't call `finishCallback()` here which will resolve the `preload` even in the case of an error.

PR #8122, opened 2025-10-03
Fix Chrome window.innerWidth priority for correct canvas resizing

### LINKED ISSUE #8121: Rotating device to landscape and back gives wrong windowWidth after resizeCanvas

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] p5.strands
- [ ] WebGL
- [ ] Friendly Errors

### p5.js version

1.11.10 and 2.0.5

### Web browser and version

Chrome 140 (is OK in Firefox)

### Operating system

Android

### Steps:
1. Run the code below in portrait orientation in Chrome Android. Observe e.g. windowWidth=392 clientWidth=392
2. Rotate to landscape. Observe e.g. windowWidth=788 clientWidth=788
3. Rotate back to portrait. Observe e.g. windowWidth=788 clientWidth=392

### Code:
```
<!DOCTYPE html>
<html>
<head>
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <script src="https://cdn.jsdelivr.net/npm/p5@1.11.10/lib/p5.min.js"></script>
</head>
<style>
    body {
        margin: 0;
    }
    canvas {
        display: block;
    }
</style>
<body>
    <h1 id="h1"></h1>
    <script>
        function setup() {
            createCanvas(windowWidth, windowHeight)
            h1.textContent = `windowWidth=${windowWidth} clientWidth=${document.body.clientWidth}`
        }
        function windowResized() {
            resizeCanvas(windowWidth, windowHeight)
            h1.textContent = `windowWidth=${windowWidth} clientWidth=${document.body.clientWidth}`
        }
    </script>
</body>
</html>
```

ayushman1210:
Resolves #8121 

Changes:
- Modified the `getWindowWidth()` and `getWindowHeight()` functions in `environment.js` to prioritize `document.documentElement.clientWidth` and `document.documentElement.clientHeight` over `window.innerWidth` and `window.innerHeight`.
- This resolves an issue in Chrome on Android where `windowWidth` and `windowHeight` values do not update correctly when the device is rotated between portrait and landscape orientation.
- Updated logic ensures consistent behavior across mobile browsers while maintaining backward compatibility with desktop browsers.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -66,345 +66,36 @@ p5.prototype.print = function(...args) {
   }
 };
 
-/**
- * A `Number` variable that tracks the number of frames drawn since the sketch
- * started.
- *
- * `frameCount`'s value is 0 inside <a href="#/p5/setup">setup()</a>. It
- * increments by 1 each time the code in <a href="#/p5/draw">draw()</a>
- * finishes executing.
- *
- * @property {Integer} frameCount
- * @readOnly
- * @example
- * <div>
- * <code>
- * function setup() {
- *   createCanvas(100, 100);
- *
- *   background(200);
- *
- *   // Display the value of
- *   // frameCount.
- *   textSize(30);
- *   textAlign(CENTER, CENTER);
- *   text(frameCount, 50, 50);
- *
- *   describe('The number 0 written in black in the middle of a gray square.');
- * }
- * </code>
- * </div>
- *
- * <div>
- * <code>
- * function setup() {
- *   createCanvas(100, 100);
- *
- *   // Set the frameRate to 30.
- *   frameRate(30);
- *
- *   textSize(30);
- *   textAlign(CENTER, CENTER);
- *
- *   describe('A number written in black in the middle of a gray square. Its value increases rapidly.');
- * }
- *
- * function draw() {
- *   background(200);
- *
- *   // Display the value of
- *   // frameCount.
- *   text(frameCount, 50, 50);
- * }
- * </code>
- * </div>
- */
 p5.prototype.frameCount = 0;
-
-/**
- * A `Number` variable that tracks the number of milliseconds it took to draw
- * the last frame.
- *
- * `deltaTime` contains the amount of time it took
- * <a href="#/p5/draw">draw()</a> to execute during the previous frame. It's
- * useful for simulating physics.
- *
- * @property {Integer} deltaTime
- * @readOnly
- * @example
- * <div>
- * <code>
- * let x = 0;
- * let speed = 0.05;
- *
- * function setup()  {
- *   createCanvas(100, 100);
- *
- *   // Set the frameRate to 30.
- *   frameRate(30);
- *
- *   describe('A white circle moves from left to right on a gray background. It reappears on the left side when it reaches the right side.');
- * }
- *
- * function draw() {
- *   background(200);
- *
- *   // Use deltaTime to calculate
- *   // a change in position.
- *   let deltaX = speed * deltaTime;
- *
- *   // Update the x variable.
- *   x += deltaX;
- *
- *   // Reset x to 0 if it's
- *   // greater than 100.
- *   if (x > 100)  {
- *     x = 0;
- *   }
- *
- *   // Use x to set the circle's
- *   // position.
- *   circle(x, 50, 20);
- * }
- * </code>
- * </div>
- */
 p5.prototype.deltaTime = 0;
-
-/**
- * A `Boolean` variable that's `true` if the browser is focused and `false` if
- * not.
- *
- * Note: The browser window can only receive input if it's focused.
- *
- * @property {Boolean} focused
- * @readOnly
- * @example
- * <div>
- * <code>
- * // Open this example in two separate browser
- * // windows placed side-by-side to demonstrate.
- *
- * function setup() {
- *   createCanvas(100, 100);
- *
- *   describe('A square changes color from green to red when the browser window is out of focus.');
- * }
- *
- * function draw() {
- *   // Change the background color
- *   // when the browser window
- *   // goes in/out of focus.
- *   if (focused === true) {
- *     background(0, 255, 0);
- *   } else {
- *     background(255, 0, 0);
- *   }
- * }
- * </code>
- * </div>
- */
 p5.prototype.focused = document.hasFocus();
 
-/**
- * Changes the cursor's appearance.
- *
- * The first parameter, `type`, sets the type of cursor to display. The
- * built-in options are `ARROW`, `CROSS`, `HAND`, `MOVE`, `TEXT`, and `WAIT`.
- * `cursor()` also recognizes standard CSS cursor properties passed as
- * strings: `'help'`, `'wait'`, `'crosshair'`, `'not-allowed'`, `'zoom-in'`,
- * and `'grab'`. If the path to an image is passed, as in
- * `cursor('assets/target.png')`, then the image will be used as the cursor.
- * Images must be in .cur, .gif, .jpg, .jpeg, or .png format and should be <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#icon_size_limits">at most 32 by 32 pixels large.</a>
- *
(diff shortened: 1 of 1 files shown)


eyaler:
@ayushman1210 Thanks! I think `innerWidth` should be before `body.clientWidth` as the body dimensions reflect the content and not the viewport. Also, the suggested change may have issue when scrollbars are present (in Chrome they are part of the `innerWidth` but not of the `documentElement.clientWidth`) - I did not look into this yet.

ayushman1210:
Hi  @eyaler ,

Thanks for the feedback! I’ve updated the code so that window.innerWidth and window.innerHeight are checked first, before document.body.clientWidth and document.documentElement.clientWidth/Height. This ensures the viewport dimensions are captured correctly and also handles scrollbars properly in Chrome.

Thanks
Ayushman

perminder-17 (changes requested):
Can you explain what's the reason behind removing all the docs? I think you need to fix the PR with the minimal changes whatever needed.

eyaler:
@ayushman1210 i don't get it. this was the state of things before with the issue that innerWidth may be wrong, no? my suggestion was that document.documentElement.clientWidth will be used first (but i think we need more feedback before making this change)

ayushman1210:
@eyaler 
 good catch and thanks for the feedback.
Quick recap of what I changed so far

- I fixed the timing issue by deferring the resize handler to the next animation frame so layout has settled after rotation (so windowResized() sees the post-rotation viewport).

- I also updated getWindowWidth() / getWindowHeight() to prefer visualViewport (when available) and then fall back to other viewport measures

About your suggestion (use [document.documentElement.clientWidth] first ) 

- I agree that preferring [document.documentElement.clientWidth] avoids surprising inclusion of scrollbar width in many desktop Chrome cases — that was exactly your point.

The tradeoffs:

- [documentElement.clientWidth]  tends to reflect the CSS/layout viewport and avoids scrollbars, which is often what sketches expect.

- [window.innerWidth] includes scrollbars on some browsers; that can make sketch sizing look different when scrollbars appear.

- [window.visualViewport] (when available) is still the best for mobile cases (on-screen keyboard, pinch-zoom) so I kept it as the top preference.

Because timing of which value updates when can vary across browsers, I left the resize handler deferred to rAF so whichever source we pick will be read after layout has updated.

Proposed conservative change (what I can implement now)

Use the following order in [getWindowWidth()] / [getWindowHeight()] :
[window.visualViewport (when available)
[document.documentElement.clientWidth]/ [clientHeight]
[window.innerWidth] / [innerHeight]
[document.body.clientWidth]/ [clientHeight]
This matches your suggestion while preserving [visualViewport] for mobile edge cases.

Questions / next steps

Is that ordering acceptable to you (visualViewport -> documentElement -> innerWidth -> body)?
If yes I’ll update the PR to use this ordering and push the change.
If you prefer a different rule (e.g., a heuristic that switches based on scrollbar presence, or exposing a small opt-in API to choose the preference), tell me and I’ll implement that instead.

Thanks 
Ayushman

eyaler:
thanks @ayushman1210 

>     * I fixed the timing issue by deferring the resize handler to the next animation frame so layout has settled after rotation (so windowResized() sees the post-rotation viewport).

I coundn't find the added requestAnimationFrame in your PR, and in my own testing requestAnimationFrame did not solve the issue. Can you point me to the commit/line?

>     * I also updated getWindowWidth() / getWindowHeight() to prefer visualViewport (when available) and then fall back to other viewport measures

In my testing visualViewport does not include scrollbars just as documentElement.clientWidth, and also is wrong when rotating back just as innerWidth. I did find viewport.segments[0].width that seems to both include scrollbars and be correct after orientation change but that is still experimental and has partial support (and i don't have a good enough understanding of it)

>     * I agree that preferring [document.documentElement.clientWidth] avoids surprising inclusion of scrollbar width in many desktop Chrome cases — that was exactly your point.

On the contrary, I actully think scrollbar should be included if possible

ayushman1210:
@eyaler Hi maintainer !! 
I pushed a change on branch fix/chrome-window-innerwidth (commit 4b5ee8b82) that adjusts how p5 computes viewport size in [environment.js]: it now prefers [visualViewport.segments[0]]  (when available), then [window.visualViewport] then [window.inner*] (so scrollbars are included when possible), and finally falls back to [document.documentElement.client*]and [document.body.client*]). Experimental accesses are guarded with try/catch.
 I ran the full test suite locally (all tests passed). Please let me know if you'd prefer excluding the experimental [segments] usage or a different fallback order — I can update quickly.

Thanks
ayushman

eyaler:
@ayushman1210 sorry but i am not a maintainer... btw, i still couldn't find where you used `requestAnimationFrame`

ayushman1210:
@eyaler
You're absolutely right about the scrollbar issue. I've removed `visualViewport.width` from the fallback chain to avoid the inconsistency.

**Updated implementation:**
- Priority 1: `visualViewport.segments[0]` (experimental foldable/dual-screen API)
- Priority 2: `window.innerWidth` (includes scrollbars, matches historical p5.js behavior)
- Priority 3: `document.documentElement.clientWidth` (fallback when innerWidth unavailable)
- Priority 4: `document.body.clientWidth` (last resort)

The segments API is only used when explicitly available (for foldable devices), then we fall back directly to `innerWidth`. This avoids the Chrome scrollbar issue you mentioned while still supporting dual-screen devices.

Regarding `requestAnimationFrame`: I apologize for any confusion - I did NOT add or modify any RAF code. My changes are limited to the `getWindowWidth()` and `getWindowHeight()` helper functions, which are only called:
1. On initialization in `_updateWindowSize()` 
2. In the `_onresize` event handler

These functions are NOT called in the animation loop, so there's no per-frame performance impact.

Thanks 
Ayushman

ayushman1210:
Hi @perminder-17 
can you please Review this PR #8122  for the issue #8121

ayushman1210:
Hi @ksen0,
When you have a moment, could you please review PR #8122 ? 
Thank you to all the maintainers for your time and for keeping this project running smoothly! 🙏

perminder-17 (approved):
Sorry for the delay in review, I'll be testing the changes once.

src/core/environment.js
```diff
@@ -782,7 +782,39 @@ p5.prototype._onresize = function(e) {
...
+      // if (typeof window.visualViewport.width === 'number' && window.visualViewport.width > 0) {
+      //   return window.visualViewport.width;
+      // }
+    }
+  } catch (e) {
+    // experimental access may throw; ignore and continue with fallbacks
+  }
+
   return (
     window.innerWidth ||
```
perminder-17, line 819: The extra `visualViewport.segments` part and the `try/catch` around it aren't really needed for this, which isn't what this bug is about. There are also a few commented-out lines left in `getWindowWidth/getWindowHeight` that we can remove to keep things tidy.

So it could come down to just this one line:

```js
(document.documentElement && document.documentElement.clientWidth) || window.innerWidth || (document.body && document.body.clientWidth) || 0
```

perminder-17 (changes requested):
Hi @ayushman1210 @eyaler , really sorry for the delay in reviewing this one. 

Btw, are you still able to reproduce this issue with the latest chrome. Looks like I am unable to reproduce this issue on my android chrome. 

If the issue is still reproduceable, @eyaler gave a very nice finding;

> the issue seems to be with chrome window.innerWidth, so maybe in https://github.com/processing/p5.js/blob/main/src/core/environment.js#L787 the order should be changes so that innerWidth gets lower priority?

The main thing the issue needs is just changing the order in which we check the window size,  checking `document.documentElement.clientWidth` first, before `window.innerWidth`, since `innerWidth` is the one that gives the wrong value on Chrome Android after rotating.

perminder-17 (commented):
For now, I am closing this PR since I am unable to reproduce the issue. Will be reopening if we again face the same issue, but looks like the latest chrome doesn't have that issue.

Issue #6992, opened 2024-04-22
Accessibility Features Proposal - STF 2


calebfoss:
### Topic

Written for the p5.js Documentation Organization & Accessibility project, this issue proposes a plan for upcoming improvements to p5's accessibility features.

## Background on Existing Features

p5's gridOutput() and textOutput() methods are very cleverly designed tools that aim to automatically describe the content on the canvas.

Current limitations and issues for these methods include:

- descriptions are exclusively in English
- text is not described
- 3D shapes are not described
- 2D primitives rendered in webGL are described incorrectly due to not factoring in the camera's perspective
- 2D primitives with similar features are combined in descriptions, resulting in inaccurate counts of shapes
- 2D primitives positioned outside the canvas are described as though they are visible
- Custom 2D shapes made with beginShape() etc. are not described

So currently, these functions can accurately describe sketches featuring exclusively 2D primitives, each with somewhat distinct features, all on the canvas for an English-reading audience. They describe the color, approximate location, and approximate percentage of space occupied of each shape.

I would argue that this improves accessibility for a very limited selection of p5 sketches. Even those that can be accurately described.

### Example

![Yellow smiley face on a light gray background](https://github.com/processing/p5.js/assets/16294664/5be252e9-7195-4cc6-a670-7981c346d5b8)

textOutput() produces the following description

```
Your output is a, 400 by 400 pixels, white canvas containing the following 4 shapes:

    yellow circle, at middle, covering 79% of the canvas.
    black circle, at top left, covering 5% of the canvas.
    black circle, at top right, covering 5% of the canvas.
    black arc, at middle, covering 22% of the canvas.

yellow circle	location = middle	area = 79%
black circle	location = top left	area = 5%
black circle	location = top right	area = 5%
black arc	location = middle	area = 22%
```

The description is technically accurate, but I would argue that it is significantly less useful at conveying the content than a concise, manually written description such as

```
A yellow smiley face on a light gray background.
```

### Side Note on AI

An idea I anticipate coming up in relation to the issue I presented above is to use AI-generated descriptions. I would like to take a moment to strongly advise against that because

- A feature like that would most likely rely on a Large Language Model, which could put p5 in a position of depending on [exploitative and inhuman labor practices](https://time.com/6247678/openai-chatgpt-kenya-workers/).
- Even the most cutting edge AI models have an ongoing issue with conveying false information as fact (often referred to as "hallucinating").
[...]

PR #8125, opened 2025-10-05
Fix Broken examples for textOutput() and gridOutput() 

hxrshxz:
Addresses #6992 

 Changes:

## Before
https://github.com/user-attachments/assets/2bbb0483-71c7-49ca-aa89-e7b8cb2d8307

https://github.com/user-attachments/assets/0b806139-9fc8-40a6-96cd-d0347b0396c3

## After 

https://github.com/user-attachments/assets/712b2b2e-59a9-466a-8bc6-2d0e5900d852

https://github.com/user-attachments/assets/2c9ab1f2-5c51-40f6-9da6-a8962c928be4

- Fixed examples https://p5js.org/reference/p5/textOutput/ 3rd and 4th (suggested by @perminder-17 )that were previously not working
 
#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/docs/yuidoc-p5-theme/assets/js/reference.js
+++ b/docs/yuidoc-p5-theme/assets/js/reference.js
@@ -2349,7 +2349,7 @@ define('listView',[
             var hash = App.router.getHash(item);
 
             // fixes broken links for #/p5/> and #/p5/>=
-            item.hash = item.hash.replace('>', '&gt;');
+            item.hash = item.hash.replace(/>/g, '&gt;');
 
             // Create a group list
             if (!self.groups[group]) {
@@ -3395,7 +3395,7 @@ var prettyPrint;
       // only when not following [|&;<>].
       '^.[^\\s\\w.$@\'"`/\\\\]*';
     if (options['regexLiterals']) {
-      punctuation += '(?!\s*\/)';
+      punctuation += '(?!\\s*\/)';
     }
 
     fallthroughStylePatterns.push(
--- a/src/typography/p5.Font.js
+++ b/src/typography/p5.Font.js
@@ -362,7 +362,7 @@ p5.Font = class {
       x = xOriginal;
       let line = lines[i];
 
-      line = line.replace('\t', '  ');
+      line = line.replace(/\t/g, '  ');
       const glyphs = this._getGlyphs(line);
 
       for (let j = 0; j < glyphs.length; j++) {


src/accessibility/outputs.js
```diff
@@ -37,6 +37,20 @@ import p5 from '../core/main';
...
+ * camera's perspective is not accounted for. Shapes with similar features may
+ * be combined in descriptions, resulting in inaccurate counts. Shapes
+ * positioned outside the canvas boundaries are described as though they're
+ * visible. Custom 2D shapes created with
+ * <a href="#/p5/beginShape">beginShape()</a> are not described.
+ *
+ * <a href="#/p5/describe">describe()</a> and
+ * <a href="#/p5/describeElement">describeElement()</a> provide more accurate
+ * and flexible alternatives for writing canvas descriptions.
+ *
```
perminder-17: Avoid absolute or negative phrasing like “does not work” and “inaccurate counts.” Prefer “not supported” and “may be imprecise.” Also, Keep a neutral, helpful tone. Also, the sketches in `textOutput`if you see 3rd and 4th are not working.  https://p5js.org/reference/p5/textOutput/ Can this be fixed in your PR as well?

perminder-17 (changes requested):
Hi, I think the points you mentioned are technically very correct but I think it sounds little odd as compared to other reference. Please keep a neutral, helpful tone for docs.

src/accessibility/outputs.js
hxrshxz: Sure , On it !

hxrshxz:
@perminder-17 Please have a look

src/accessibility/outputs.js
```diff
@@ -102,11 +117,13 @@ import p5 from '../core/main';
  *
  * <div>
  * <code>
- * function draw() {
+ * function setup() {
  *   // Add the text description and
  *   // display it for debugging.
  *   textOutput(LABEL);
```
perminder-17, line 116: Actually, you may just add `createCanvas(200,200)` to the `setup()` and move `textOutput` to the `draw()` call only. Same with the below sketch. That will work too I guess.

src/accessibility/outputs.js
```diff
@@ -37,6 +37,19 @@ import p5 from '../core/main';
  * <a href="https://p5js.org/tutorials/writing-accessible-canvas-descriptions/">Writing accessible canvas descriptions</a>
  * to learn more about making sketches accessible.
  *
+ * `textOutput()` currently generates descriptions in English only.
```
ksen0: Could you update this documentation please to reflect [the documentation style guide](https://p5js.org/contribute/documentation_style_guide/)? Mainly, significantly simplifying the sentences. Alternatively, this can be removed from this PR, because it's not quite in the scope of fixing the broken exampels.

src/accessibility/outputs.js
ksen0: +1, please make sure the `textOutput` is in `draw` not `setup`, and that `draw` contains the correctly sized canvas creation (`createCanvas(100,100);`). The purpose of the example is to demonstrate how `textOutput` updates to reflect the animation. (Same for the other example)

hxrshxz:
I have moved the documentation part into a separate PR #8143

src/accessibility/outputs.js
hxrshxz: done #8143

src/accessibility/outputs.js
hxrshxz: done

PR #8129, opened 2025-10-05
Use correct type for event param of windowResized

### LINKED ISSUE #8128: Zod error "expected UIEvent" in windowResized parameter validation

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] p5.strands
- [ ] WebGL
- [x] Friendly Errors

### p5.js version

2.0.5

### Web browser and version

Microsoft Edge Version 140.0.3485.94 (Official build) (64-bit), Firefox 143.0 (64-bit)

### Operating system

Windows

### Steps:
1. Have any sketch with a `windowResized` method
2. Resize the window
3. Zod prints this error on the console:
```
        Zod error object {code: "custom", message: "Expected a UIEvent", …}
        p5.js says:  in windowResized().
```

### Snippet:

```js
	p.windowResized = (event) => {
		console.log({ event })
		p.resizeCanvas(p.windowWidth, p.windowHeight)
	}
```

The reason is that the event param is documented to be of type `UIEvent`. However, in most browser, it will just be of type `Event`.

See: https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event#event_type

yogan:
See: https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event#event_type

Using the correct type should get rid of Zod printing errors in the console when window resize events are triggered:

    Zod error object {code: "custom", message: "Expected a UIEvent", …}
    p5.js says:  in windowResized().

Resolves #8128

 Changes:
Put correct event type in docstring of `windowResized`

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -717,7 +717,7 @@ p5.prototype.windowHeight = 0;
  * can be used for debugging or other purposes.
  *
  * @method windowResized
- * @param {UIEvent} [event] optional resize Event.
+ * @param {Event} [event] optional resize Event.
  * @example
  * <div class="norender">
  * <code>


perminder-17 (approved):
Btw, I was unable to reproduce this issue. Can you provide me the full sketch what you were trying to use which shows zod errors? 

Also, in the issue you mentioned it gives errors in `2.0.5` version but for 2.x versions we are working on `dev-2.0` branch. If the issue persist in both versions, it would be nice to open-up a followup PR for dev-2.0 branch as well.

Apart from all, taking reference from MDN docs, resize-window generic type is `Event` hence it should be `Event` instead of `UIEvent` which looks correct. Thanks for working on this.

aidanwyber:
@perminder-17 I've also been getting this error
image
Thank you @yogan for fixing it!

yogan:
@perminder-17 here is an example sketch that throws the error on resize: https://github.com/yogan/p5js-creative-coding/blob/main/src/sketches/sketch1.ts#L41 - `setup()` and `draw()` are not really relevant, it is sufficient to have this very common pattern:

```ts
	p.windowResized = () => {
		p.resizeCanvas(p.windowWidth, p.windowHeight)
	}
```

The error looks exactly like the screenshot from @aidanwyber.

Thanks for pointing out the `dev-2.0` branch, I'll create a separate PR there as well.

yogan:
I've just noticed that the type is already correct on the `dev-2.0` branch:
https://github.com/processing/p5.js/blob/dev-2.0/src/core/environment.js#L731
This might explain why @perminder-17 can't reproduce the error.

Note that when I specified that I'm using `v2.0.5`, I was referring to the npm version, I'm using p5js as a package in a Vite project.
Not sure how the npm release process works, but looking into the code in my `node_modules` there is definitely the wrong type there.

perminder-17 (commented):
Idk, i was testing it in `dev-2.0` since in the issue it was written: 

> p5.js version

> 2.0.5

Anyways, thanks for fixing this.

PR #8138, opened 2025-10-11
refactor: Move setViewport to standalone addon in lib/addons

ombalgude:
Hi! @perminder-17 I've just pushed the requested changes.

Based on your feedback, I've refactored the setViewport feature. It has been removed from the core and is now a standalone addon located at lib/addons/p5.viewport.js, following the convention of other addons in the project.

Thank you for the guidance! Let me know if there's anything else.

### DIFF
--- a/lib/addons/p5.viewport.js
+++ b/lib/addons/p5.viewport.js
@@ -1,0 +1,50 @@
+/**
+ * @module p5.viewport
+ * @submodule p5.viewport
+ * @for p5
+ * @version 0.0.1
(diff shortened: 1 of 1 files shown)


limzykenneth:
Addon libraries should be self hosted and in this case is a test bed for the feature. We can't merge this as we won't be distributing this addon as an official addon. Please see the final section of [creating libraries](https://p5js.org/contribute/creating_libraries/) on the p5.js website if you wish to hae the addon listed on the website as a third party addon.

Issue #8117, opened 2025-09-29
[p5.js 2.0 Bug Report]: If you create a p5.Vector without arguments, the 2d vector subtracted by it will be strange.


inaridarkfox4231:
### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [ ] Other (specify if possible)

### p5.js version

2.0.5

### Web browser and version

Chrome

### Operating system

Windows

### Steps:
1. Create a vector without arguments.
2. Create a 2D vector and subtract the vector you prepared earlier.
3. The third component will be NaN. Furthermore, If you apply rotate() to it, all components will become NaN.

### Snippet:

```js
function setup() {
  createCanvas(400,400);
  const v=createVector(100,0);
  const w=createVector();
  v.sub(w);
  console.log(v.toString());
  v.rotate(1);
  console.log(v.toString());
}
```

<img width="881" height="678" alt="Image" src="https://github.com/user-attachments/assets/3332c5f2-7474-41b7-8343-72f3c38293f3" />

If you explicitly make it two-dimensional with two zeros, there is no problem.

<img width="498" height="381" alt="Image" src="https://github.com/user-attachments/assets/51acf11c-c60c-4cda-98ad-6ed9bd9d11a3" />

This issue has been occurring since version 2.0.0. It has not occurred in version 1.x.

<img width="799" height="757" alt="Image" src="https://github.com/user-attachments/assets/f56d7d2a-c725-481b-9846-36b48bfd9b21" />

<img width="779" height="732" alt="Image" src="https://github.com/user-attachments/assets/1e6454b3-95be-49be-992a-2e58c0767898" />

Specifying a value without arguments is a common practice when creating a default vector, whether in two or three dimensions, so I thought this could be inconvenient in some cases.

inaridarkfox4231:
Subtracting a zero vector is a rare occurrence, but it does also occur when any vector applies to it by add() (even a 2D vector!), so it's a problem.
[...]

PR #8142, opened 2025-10-13
fromAngle returns a 2D Vector

ksen0:
Addresses https://github.com/processing/p5.js/issues/8118 - this thread has much more detail, but the short version is as follows.

In 2.x, there are n-dimensional vectors, rather than 3D vectors, or 2D vectors that are underneath actually 3D with a 0 z-value. While this change is stabilizing, there are a couple of consequences.

First, the bug linked above: the `random2D()` function which uses `fromAngle()` should both be using 2D vectors, as they both are defined in 2D space. See the above thread for the whole discussion, but rather than a breaking change, this has been recognized as an unintentional bug.

Second, operations on vectors of non-matching sizes have to be handled. To make the whole topic more actionable, this is considered to be the scope of this issue: https://github.com/processing/p5.js/issues/8117

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -3241,7 +3241,7 @@ class Vector {
    *   // Create a p5.Vector object.
    *   let v = p5.Vector.random2D();
    *
-   *   // Prints "p5.Vector Object : [x, y, 0]" to the console
+   *   // Prints "p5.Vector Object : [x, y]" to the console
    *   // where x and y are small random numbers.
    *   print(v.toString());
    * }


sidwellr:
Can we please make a new issue (or sub-issue) for "operations on vectors of non-matching sizes have to be handled"? Issue #8117 is focused on eliminating the `NaN` in the result when subtracting a 3D vector from a 2D vector, which breaks many version 1 sketches when moved to version 2. This pull request will fix many of the problems since `random2D()` is a common source of the unintentional 3D vectors. But `createVector()` with no parameters will also create a 3D vector, which will result in `NaN` when subtracted from a 2D vector.

limzykenneth:
@sidwellr The dimension mismatch case we'll use #8117 to discuss and resolve.

PR #8163, opened 2025-10-18
Add noiseDetail() support for p5.strands

### LINKED ISSUE #8160: Support noiseDetail in p5.strands

hxrshxz:
Resolves #8160 

 Changes:

This PR adds support for `noiseDetail()` in p5.strands by exposing noise parameters to shaders via uniforms.

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes


src/strands/strands_api.js
```diff
@@ -121,6 +121,20 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
       return originalNoise.apply(this, args); // fallback to regular p5.js noise
     }
 
+    const hasNoiseUniforms = strandsContext.uniforms.some(u => u.name === 'uNoiseOctaves');
```
davepagurek: There are some subtleties around where these values come from. With this implementation, the shader will grab the values at the time the shader is used, not at the time `noise()` is called. So in this scenario:
```js
const sh = baseMaterialShader().modify(() => {
  const t = uniformFloat(() => millis() * 0.001)
  getWorldInputs((inputs) => {
    noiseDetail(2, 0.5)
    inputs.position.x += noise(t)
    return inputs
  })
})

noiseDetail(4, 0.5)
shader(sh)
sphere(30)
```
In this scenario, the sphere's position would be offset using 4 octaves of noise instead of 2 like we'd expect.

To address this, I think we may need to also override `noiseDetail` while inside of a p5.strands context (similar to how we override `noise`) to keep track of what the last noise detail value set was, and then pass that in as an argument to the GLSL `noise` function rather than using a uniform. So in the example above, the last value would be 2 and 0.5, so it like cements those values in. It also means that if the user does something dynamic, like taking a uniform value and adding something to it, then the noise detail values may be other strands nodes, but that still works if we pass them into the function.

hxrshxz: Thanks  I have updated the PR  to pass noiseDetail() values as function parameters (not uniforms), capturing them at call time inside modify() to fix the timing issue.

davepagurek:
This approach looks like it should work! I haven't yet had the chance to run this myself, but it would help with review if you tried making a test sketch that uses noise + noiseDetail in a strands shader. This can be in the p5 web editor, although there you'll have to manually upload your `lib/p5.js` after running `npm run build` to the web editor assets and edit `index.html` to use that instead of the p5 CDN in order to have your sketch use the updated library code.

It also looks like there's a merge conflict that needs resolution. Is that something you're comfortable doing?

hxrshxz:
https://editor.p5js.org/harshmastic/sketches/JAcdZZZNo 
Hey I have tested the sketch as u suggeted and it's working fine ,also fixed the merge conflict

davepagurek:
Hi @hxrshxz! It looks like that example isn't using noise in p5.strands, it's using it in beginShape/endShape. A test of p5.strands would use noise to offset vertex positions in a shader, and then draw content using that shader. I also see that it's logging warnings about the usage of `text()`, and the text you're rendering doesn't show up. Please make sure you take a look at the output if you're using AI tools to assist you.

hxrshxz:
@davepagurek Sorry for the delay I was actually busy with my university exams . I Will update the Pull request in some days

hxrshxz:
image
 it's down will share the example link as soon as it's up

hxrshxz:
https://editor.p5js.org/harshmastic/sketches/F4sFSd0wP - Here's the new sketch

davepagurek (approved):
Sorry for the delay, this is looking good now!

PR #8167, opened 2025-10-20
Updated 1.x manual tests

ksen0:
Part of the release process for p5.js is verifying manual tests. This PR updates some external API call (needed by tests on JSON fetching) which no longer work (because the original API used requires a key). The calls now rely on [wttr.in](https://github.com/chubin/wttr.in).


davepagurek (approved):
thanks for keeping tabs on the manual tests!

PR #8240, opened 2025-11-08
Fix(vector): restrict setHeading() to 2D vectors and add friendly error for others

### LINKED ISSUE #8215: [2.x] Fix dimensional quirk in the `setHeading()` method of `p5.Vector`
### Problem

### Solution

reshma045:
Resolves #8215

 Changes:

This PR updates the p5.Vector.setHeading() method to support only 2D vectors in alignment with the proposed 2.x design.

- Added a dimension check in `setHeading()`:
- If the vector’s z component is nonzero (i.e., not 2D), the method now emits a clear `p5._friendlyError()` explaining that `setHeading()` is 2D-only in p5.js 2.x.
- For 2D vectors (`z === 0`), heading behavior remains the same, preserving magnitude and rotating in the xy-plane.
- Keeps compatibility with existing sketches that use `setHeading()` on true 2D vectors.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -2229,7 +2229,21 @@ p5.Vector = class {
 
   setHeading(a) {
     if (this.isPInst) a = this._toRadians(a);
-    let m = this.mag();
+    if (this.z !== 0) {
+      p5._friendlyError(
+        'p5.Vector.setHeading() only supports 2D vectors (z === 0). ' +
+        'For 3D or higher-dimensional vectors, use rotate() or another ' +
+        'appropriate method instead.',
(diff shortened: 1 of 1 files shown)


perminder-17 (changes requested):
Hi @reshma045 , thanks for your work and time on this. This PR should target the dev-2.0 branch since the issue here is only with the 2.x versions. Can you please retarget your PR or open up a new PR on targetting dev-2.0 branch?

reshma045:
Thanks @perminder-17! I will open a new PR and target it to the dev-2.0 branch as requested.

PR #8244, opened 2025-11-08
Fix: prevent print() with no arguments from triggering browser print dialog

sdass1918:
Resolves [processing/p5.js-web-editor#3678](https://github.com/processing/p5.js-web-editor/issues/3678)

 Changes:

- Updated src/core/environment.js
- Modified p5.prototype.print to check for empty arguments before calling window.print().
- Added a console warning when print() is called without any arguments instead of triggering the browser print dialog.

 Behavior Before:
```js
print(); // opened browser print popup
print("Hello"); // logged correctly
```
Behavior After:
```js
print(); // logs a warning, no popup
print("Hello"); // logs correctly
```

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/core/environment.js
+++ b/src/core/environment.js
@@ -48,7 +48,7 @@ p5.prototype._targetFrameRate = 60;
  */
 p5.prototype.print = function(...args) {
   if (!args.length) {
-    console.warn('p5.js: print() called with no arguments. Nothing to print.');
+    console.log('');
     return;
   } else {
     console.log(...args);


src/core/environment.js
```diff
@@ -51,16 +48,8 @@ let windowPrintDisabled = false;
...
-      _windowPrint();
-      if (
-        window.confirm(
-          'You just tried to print the webpage. Do you want to prevent this from running again?'
-        )
-      ) {
-        windowPrintDisabled = true;
-      }
-    }
+    console.warn('p5.js: print() called with no arguments. Nothing to print.');
```
oneplusiota: shouldn't we print a newline character if no argument is passed.
It is a convention that is followed in java `println` or js `console.log`.

sdass1918: Thanks for the suggestion, @oneplusiota! 🙏

That’s a great point — following the convention of `console.log()` and Java’s `println` makes sense for consistency.

I can update the implementation so that when `print()` is called without any arguments, it simply prints a newline (i.e., `console.log("")`) instead of showing a warning.

Would you prefer that behaviour instead of the warning message? I can make the adjustment and push an update once confirmed.

oneplusiota:
@perminder-17 I disagree. Why should `print` work differently for different arguments. It's work is to write to the output stream. And more importantly, there is no use of the print window functionality in the p5 editor. And in case the user does want to print the page, they can do so by directly using the browser's print screen functionality.

src/core/environment.js
oneplusiota: yeah this will work.

sdass1918:
Thanks both for your feedback — I appreciate the different perspectives here 🙏

My intent with this fix was to prevent `print()` with no arguments from unintentionally triggering the browser’s print dialog, since that can interrupt the sketch execution and isn’t consistent with how `print()` behaves elsewhere in p5.js or in typical console output functions.

I agree that consistency is important. From a usability point of view, having `print()` behave like `console.log("")` (i.e., output a blank line) seems clearer and safer than opening the print window, especially since users can explicitly call `window.print()` if they want that behavior.

src/core/environment.js
sdass1918: Great! I've updated the implementation so that `print()` with no arguments now outputs a blank line `(console.log(""))`.
Let me know if any further tweaks are needed. 👍

perminder-17 (commented):
I'm so sorry, but this is the intended behavior of the `print()` functions and these are the discussions related to it: https://github.com/processing/p5.js/issues/4272 and https://github.com/processing/p5.js/issues/2765. 

Thanks for all your work on this. Keep looking at the issue section: https://github.com/processing/p5.js/issues and would be very great to ask maintainers confirmation on the issues before working. 

Really thanks for your work and patience on this one.

PR #8245, opened 2025-11-09
Removing grunt from contributor guidelines

perminder-17:


perminder-17:
Hi @sidwellr can you please have a look, if that's looks correct to you?

sidwellr:
Yes, I don't think anything needs changing except that one line

Issue #8212, opened 2025-10-27
Friendly Error points to executing line in the p5.js file, not the line causing the issue


Namrad:
### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] p5.strands
- [ ] WebGL
- [ ] DevOps, Build process, Unit testing
- [ ] Internationalization (i18n)
- [x] Friendly Errors
- [ ] Other (specify if possible)

### p5.js version

2.0.5.

### Web browser and version

Vivaldi 7.6.3797.63 (Stable channel) (arm64) 

### Operating system

Apple M4 Max, Sequoia 15.6

### Steps to reproduce this

No real 'steps' involved. Just ran into this while transferring to p5.js 2.0.5:

**While using VS Code: When reading the console the Friendly Error message points to the line in the p5.js main-file that executes the Friendly Error message itself, rather than what caused the Friendly Error message in my own .js file.**

Because of this I had to manually go over all instances of `createVector()` and `float()`, to see where the error came from.

The friendly error messages are really helpful as opposed to the 'unfriendly ones', especially for autodidact coders like me. Maybe therefor this is not necessarily a bug, but more a feature request. I hope this is helpful either way. Sorry if this is an incorrect bug report.

<img width="625" height="50" alt="Image" src="https://github.com/user-attachments/assets/2486bfb4-5d1c-4566-a8ff-3e3bdb5866cf" />
<img width="382" height="119" alt="Image" src="https://github.com/user-attachments/assets/5d303dd2-3c0a-4f6a-8cd5-ab9760b78b14" />

### Instances:
1. When passing `NaN` into a `createVector()` method.
2. When using `float()` method incorrectly.

### Snippet:

```js
function setup() {
  let vectorA = createVector(10, NaN);
  let vectorB = createVector(vectorA.x + vectorA.y, vectorA.y);
}
```
```js
function setup() {
  let divisor = 2;
  let time = float(frameCount / divisor);
}

```

Sahil5H4RD4:
Hey, I’d like to work on this issue and help get it fixed.
I’ve gone through the problem and I think I can contribute a solid solution.
Please assign this to me if it’s available.

Mario5T:
# The Fix

[...]

PR #8246, opened 2025-11-10
fixed the Friendly Error message issue in p5.js where error stack tra…

Mario5T:
# The Problem (Issue #8212)

When using functions like `createVector(10, NaN)` or `float()` incorrectly, the **Friendly Error message** would point to the **line in the p5.js library file** that executes the error message, rather than the **line in the user's sketch** that caused the error.  
This made debugging difficult for users.

---

# Root Cause

In **`validate_params.js`**, the code used a **hardcoded stack frame index** `parsed[3]` to extract the location of the error.  
This assumed a **fixed call stack depth**, which didn't account for errors originating from within **p5.js library functions**.

---

###  Typical Stack Trace

| Index | Description |
|--------|--------------|
| [0] | Error creation in `_friendlyParamError` |
| [1] | `_friendlyParamError` itself |
| [2] | `_validateParameters` |
| [3] | The p5.js function (e.g., `createVector`, `float`) ← *was pointing here* |
| [4] | User's code ← *should point here* |

---

# The Fix

Modified **`validate_params.js`** (lines **608–672**) to:

- Detect the **p5.js** library file name by throwing a test error and parsing its stack trace  
- Filter out all frames from the **p5.js** library to find the first frame from **user code**  
- Use the **user's code frame** for error location reporting  
- Maintain backward compatibility by **falling back to `parsed[3]`** if no user frame is found  

This approach mirrors the logic already used in the **`processStack()`** function in **`fes_core.js`** for handling global errors.

---

# Testing

The fix can be tested with the examples from the bug report:

```javascript
function setup() {
  let vectorA = createVector(10, NaN);  // Error will now point to this line
  let vectorB = createVector(vectorA.x + vectorA.y, vectorA.y);
}

### DIFF
--- a/src/core/friendly_errors/validate_params.js
+++ b/src/core/friendly_errors/validate_params.js
@@ -609,6 +609,22 @@ if (typeof IS_MINIFIED !== 'undefined') {
         // const re = /Function\.validateParameters.*[\r\n].*[\r\n].*\(([^)]*)/;
         const myError = new Error();
         let parsed = p5._getErrorStackParser().parse(myError);
+        let p5FileName;
+        try {
+          throw new Error();
+        } catch (testError) {
+          const testStacktrace = p5._getErrorStackParser().parse(testError);
+          p5FileName = testStacktrace[0].fileName;
+        }
+        
+        let userFrame = null;
+        for (let i = 0; i < parsed.length; i++) {
+          if (parsed[i].fileName && parsed[i].fileName !== p5FileName) {
+            userFrame = parsed[i];
+            break;
+          }
+        }
+        
         if (
           parsed[3] &&
           parsed[3].functionName &&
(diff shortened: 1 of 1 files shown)


Mario5T:
@perminder-17 plz review it.

perminder-17:
Hi, thanks for your work and your patience. https://github.com/processing/p5.js/issues/8212#issuecomment-3520598906 I just wrote a comment, and could wait for the approvals?

Mario5T:
@perminder-17  just asking if the issue is approved to be worked on will my pr be considered?

perminder-17:
> @perminder-17 just asking if the issue is approved to be worked on will my pr be considered?

Actually, @sahilsharda commented first on the issue. If they still want's to work, I think I should wait for their PR to open. If not, I'll surely review your PR.

PR #8291, opened 2025-11-25
Updated point() examples for visible points.

### LINKED ISSUE #8266: `point()` should default to `1px` and respect `strokeWeight(1)` in WEBGL mode; update `point()` examples for visible points

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] p5.strands
- [x] WebGL
- [ ] Friendly Errors

### p5.js version

1.x and 2.x

### Web browser and version

Firefox

### Operating system

linux

### Steps to reproduce this

In webgl: `strokeWeight(1)` ignored for `point()`, always uses larger default size
### Steps:
```js
// WEBGL
function setup() {
  background(200);

  stroke(0);
  strokeWeight(1);
  point(0, 0);
}
```

Compare that with: 

```js
function setup() {
  background(200);

  stroke(0);
  strokeWeight(2);
  point(0, 0);
}
```

Docs say a point’s default width is 1 pixel, and `strokeWeight()` sets the width in pixels for points. So in WEBGL, `strokeWeight(1)` should give a 1px point, and increasing to 2 should visibly double it.

**Actual behavior:**

With `strokeWeight(1)` or without strokeWeight() called, the point appears ~5-6px wide.
So `strokeWeight(1)` appears to be ignored and `WebGL` falls back to a larger default size, also without calling strokeWeight(), it still not defaults to 1px.

The idea is that `point()` should work the same way as described in the docs. If the default is 1px (and that works correctly in P2D), then WEBGL mode should also use 1px by default.

Also, currently in the docs, when we see the examples, it uses the default 1px points. We can add `strokeWeight()` to make it more visible.

Image

So, the goal is to update the docs for adding `strokeWeight()` in the example, and also make the `point()` being consistent in both WEBGL and 2D mode.

menacingly-coded:
Resolves #8266

 Changes:
There were three examples of point() in the docs which I updated and made them visible as compared to before.

 Screenshots of the change:
before:
image

after:
image

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/shape/2d_primitives.js
+++ b/src/shape/2d_primitives.js
@@ -761,7 +761,7 @@ function primitives(p5, fn){
    *
    *   background(200);
    *   
-   *   // Making point to 5 pixels.
+   *   // Making point to 5 pixels
    *   strokeWeight(5);
    * 
    *   // Top-left.


perminder-17 (approved):
Looks good to me!

perminder-17:
@all-contributors please add @menacingly-coded for doc

menacingly-coded:
Hi @perminder-17 , Thank you for the update! Sure, I’ll make the necessary changes for the main branch as well. Let me know if there’s anything else you'd like me to adjust.

PR #8305, opened 2025-11-29
Updated point() examples for visible points in main branch. 

### LINKED ISSUE #8266: `point()` should default to `1px` and respect `strokeWeight(1)` in WEBGL mode; update `point()` examples for visible points

menacingly-coded:
Resolves #8266

 Changes:
There were three examples of point() in the docs which I updated and made them visible as compared to before for the main branch. (similar that has done for dev-2.0 branch)

 Screenshots of the change:
before:
image

after:
image

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -634,6 +634,7 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
       }
     };
 
+    this.curStrokeWeight = 1;
     this.pointSize = this.curStrokeWeight;
     this.curStrokeCap = constants.ROUND;
     this.curStrokeJoin = constants.ROUND;
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -7248,6 +7248,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "Itsrajsk",
+      "name": "Raj Kashid",
+      "avatar_url": "https://avatars.githubusercontent.com/u/117977675?v=4",
+      "profile": "https://github.com/Itsrajsk",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "repoType": "github",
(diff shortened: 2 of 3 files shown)


menacingly-coded:
Hi @perminder-17 , As suggested, I’ve created this PR for the main branch.
I’ll continue working on this and push further changes in the same branch, so I’m converting this PR into a draft for now.

menacingly-coded:
Hey @perminder-17, I have fixed the issues. Please review and let me know if any changes are needed.

src/webgl/p5.RendererGL.js
```diff
@@ -634,8 +634,8 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
       }
     };
 
-    this.pointSize = 5.0; //default point size
     this.curStrokeWeight = 1;
+    this.pointSize = this.curStrokeWeight;
```
perminder-17, line 638: Thanks. This change makes the behavior match the docs, but the concern is that users who rely on the old default point size in their sketches will see a visual change and will now need to add `strokeWeight(5)` to keep the old look.

PR #8328, opened 2025-12-10
Resolves #8278

### LINKED ISSUE #8278: [2.0] Proposal: A consistent mechanism for reading renderer properties (2d)
As discussed with @limzykenneth, there is not currently a consistent usage pattern for reading renderer state properties. In some cases (e.g. `textWeight()`) calling the function with no argument returns the current setting, while in other cases (`rectMode()`, for example) it results in an error message.

There are many use-cases for this information (for example, writing a function which needs to adjust to the various rect-modes) and in v1.x, people have used non-API access like `_renderer._rectMode`. In 2.x, this info is available in `_renderer.states`, also not part of the API, and also not available in global mode (for example, in the editor, where one would have to do ` p5.instance._renderer.states.rectMode`.

The proposal here is to standardise the API so that calling no-argument versions of the state-setting functions return the current values. This will require evaluating whether there are existing uses of any of the existing no-argument versions of these methods (for example, `randomSeed()`) as well as checking for any other problematic cases

dhowe:
Added the various getters, plus tests, as detailed in ticket #8278

- [x] `npm run lint` passes

Includes the following changes:
- updated `paramaterData.json` for getter functions to accept no-argument calls
- for getters that return an object (`textAlign`, `scale`, `translate`, etc.), the setter now accepts that object 
- added `background` (either a color or image) to `renderer.state`
- all color-related getters return `p5.Color`
- added tests for all of these in `test/unit/core/properties.js`
- no change to `randomSeed()` or `noiseSeed()`

### DIFF
--- a/src/core/p5.Renderer3D.js
+++ b/src/core/p5.Renderer3D.js
@@ -1486,7 +1486,7 @@ export class Renderer3D extends Renderer {
     // works differently and is global p5 state. If the p5 state has
     // been cleared, we also need to clear the value in uSampler to match.
     fillShader.setUniform("uSampler", this.states._tex || empty);
-    fillShader.setUniform("uTint", this.states.tint.array().map(v => v * 255));
+    fillShader.setUniform("uTint", this.states.tint._getRGBA([255, 255, 255, 255]));
 
     fillShader.setUniform("uHasSetAmbient", this.states._hasSetAmbient);
     fillShader.setUniform("uAmbientMatColor", this.states.curAmbientColor);


src/core/p5.Renderer.js
```diff
@@ -35,6 +35,7 @@ class ClonableObject {
 
 class Renderer {
   static states = {
+    background: null,
```
davepagurek: We didn't end up using this right? Safe to remove?

src/core/p5.Renderer2D.js
```diff
@@ -200,8 +203,12 @@ class Renderer2D extends Renderer {
       if (this._isErasing) {
         this._pInst.erase();
       }
+      bgForState = color; // save for getter (#8278)
     }
     this.pop();
+
+    this.states.setValue('background', bgForState); // set state (#8278)
```
davepagurek: Similarly, if we don't return this anywhere, can we clean this up?

src/core/p5.Renderer3D.js
```diff
@@ -1486,7 +1486,7 @@ export class Renderer3D extends Renderer {
     // works differently and is global p5 state. If the p5 state has
     // been cleared, we also need to clear the value in uSampler to match.
     fillShader.setUniform("uSampler", this.states._tex || empty);
-    fillShader.setUniform("uTint", this.states.tint);
+    fillShader.setUniform("uTint", this.states.tint.array().map(v => v * 255));
```
davepagurek: Looking below at how `tint()` used to work, it looks like we used to set the state to `c._getRGBA([255, 255, 255, 255])`. Maybe we should be using the same here, which explicitly includes the color mode and the maxes, which could potentially be different and not reflected in `array()`?

src/image/loading_displaying.js
```diff
@@ -1138,6 +1138,12 @@ function loadingDisplaying(p5, fn){
    * sets the alpha value. For example, `tint(255, 0, 0, 100)` will give images
    * a red tint and make them transparent.
    *
+   * Calling `tint()` without an argument returns the current tint as a <a href="#/p5.Color">p5.Color</a> object.
+   *
+   * @method frameRate
```
davepagurek: This looks like maybe it was copy and pasted by accident?

src/image/loading_displaying.js
```diff
@@ -1242,10 +1248,18 @@ function loadingDisplaying(p5, fn){
...
+   * @return {p5.Color}      the current tint color
+   */
   fn.tint = function(...args) {
-    // p5._validateParameters('tint', args);
-    const c = this.color(...args);
-    this._renderer.states.setValue('tint', c._getRGBA([255, 255, 255, 255]));
+    if (args.length === 0) {
+      return this._renderer.states.tint; // getter
+    }
+    if (args && args.length) {
```
davepagurek: super tiny but I think with rest params like `args`, we don't have to worry about it being falsy, so `if (args.length)` would work here

src/type/textCore.js
```diff
@@ -1461,7 +1461,14 @@ function textCore(p5, fn) {
 
     // the setter
     if (typeof h !== 'undefined') {
+      // accept what is returned from the getter
+      if (h.hasOwnProperty('horizontal')) {
+        h = h.horizontal;
+      }
       this.states.setValue('textAlign', h);
+      if (h.hasOwnProperty('vertical') && typeof v === 'undefined') {
```
davepagurek: Does this condition work if you pass in an object with both `horizontal` and `vertical` as keys? Looks like if it has `horizontal` then `h` gets reassigned to a number above, and would no longer go through this if statement.

test/unit/webgl/p5.RendererGL.js
```diff
@@ -1533,7 +1548,7 @@ suite('p5.RendererGL', function() {
           };
         });
       }).then(function(_tint) {
-        assert.deepEqual(_tint, [255, 255, 255, 255]);
+        assert.deepEqual(_tint.array(), [1, 1, 1, 1]);
```
davepagurek: It's a little confusing that `.array()` here is in a 0-1 range but the others are in a 0-255 range. I think the color constructor `new Color([1, 1, 1, 1])` is correct -- it correctly outputs `#ffffff` as `.toString()` -- but this shows us that `.array()` actually uses the channel maxes of the input and isn't always normalized. Maybe for all the tests in this set of tint tests we should use `._getRGBA([255, 255, 255, 255])` to be consistent about 0-255 range?

src/core/p5.Renderer3D.js
dhowe: I did this to be consistent with the other color properties in renderer.states, all of which are stored as p5.Color

src/image/loading_displaying.js
dhowe: good catch

src/type/textCore.js
dhowe: please recheck

test/unit/webgl/p5.RendererGL.js
dhowe: was trying to avoid using private methods, but can do

dhowe: please check

src/core/p5.Renderer3D.js
davepagurek: Right, I think we need to do the same `_getRGBA(...)` call here though, since if any of the colors were created with different max values, then it looks like `.array().map(v => v * 255)` will be different from what we expect (e.g. `.array()` might be in 0-1 range, as the default tint is, or it might already be in 0-255 range, as the other values in the tint test file were, so multiplying by 255 again would make the values too large.)

dhowe: done

davepagurek (approved):
Awesome, thanks for all your work on this! I think we're good to go!

PR #8330, opened 2025-12-15
Fix setHeading behavior for non-2D p5.Vector instances

### LINKED ISSUE #8215: [2.x] Fix dimensional quirk in the `setHeading()` method of `p5.Vector`

aashu2006:
Resolves #8215

Changes:
- Restrict `p5.Vector.prototype.setHeading()` to vectors with a zero or undefined z component.
- Throw an error when `setHeading()` is called on vectors with a non-zero z value.
- Added unit tests to cover invalid-dimension behavior.

Screenshots of the change:
N/A (non-visual change)

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -2228,12 +2228,18 @@ p5.Vector = class {
  */
 
   setHeading(a) {
+    if (this.z !== undefined && this.z !== 0) {
+      throw new Error('setHeading() is only supported for 2D vectors.');
+    }
+
     if (this.isPInst) a = this._toRadians(a);
     let m = this.mag();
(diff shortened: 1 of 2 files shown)


aashu2006:
Happy to adjust this if a different error behavior or messaging is preferred.

perminder-17 (commented):
Hi, thanks for your work on this. The issue is actually assigned to someone else and already had an opened PR for it, so I need to close your PR. Thanks for your work on this.

aashu2006:
Thanks for letting me know! Appreciate the clarification and the review. Happy to contribute again.

PR #8351, opened 2025-12-20
Fix Camera.slerp for ortho by cloning projMatrix into uPMatrix when active

nakednous:
**Resolves #8295**

This PR fixes the crash that occurs when calling `Camera.slerp()` between orthographic cameras.

The solution updates `uPMatrix` using `states.setValue()`—as recommended by @davepagurek—instead of assigning directly to `mat4`, which avoids the crash while keeping the renderer state consistent.

**Change:**

* Replace direct `mat4` assignment with a `states.setValue('uPMatrix', …)` update inside `Camera.slerp()`.

### DIFF
--- a/src/webgl/p5.Camera.js
+++ b/src/webgl/p5.Camera.js
@@ -1602,8 +1602,7 @@ class Camera {
       );
       // If the camera is active, make uPMatrix reflect changes in projMatrix.
       if (this._isActive()) {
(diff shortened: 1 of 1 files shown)


davepagurek:
Thanks for making the changes! It looks like an autoformatter ran on the file, is it possible to get it to just have the slerp update?

nakednous:
Formatting reverted; only the minimal slerp fix remains now. Ready for review—thanks!

PR #8357, opened 2025-12-24
Fixes for server side usage

limzykenneth:
Changes:
Mainly putting some of the references to `window` and `document` behind checks that they exist so that they can run on a node.js runtime. Only two modules doesn't work so far which are FES and events. Events module is not expected to work since it relies on DOM event listeners, FES will need some changes for it to work.

This is still a bit of a work in progress and should be merged at least after a round of dependency updates.


limzykenneth:
Put something simple together to test this as a proof of concept: https://github.com/limzykenneth/p5.term.js

limzykenneth:
@ksen0 @davepagurek Do have a look when you have the time. This is all still a bit experimental in nature and probably need a bit more iteration in future PR.

src/core/environment.js
```diff
@@ -724,21 +727,11 @@ function environment(p5, fn, lifecycles){
   };
 
   function getWindowWidth() {
-    return (
-      window.innerWidth ||
-      (document.documentElement && document.documentElement.clientWidth) ||
-      (document.body && document.body.clientWidth) ||
-      0
-    );
+    return isBrowser ? document.documentElement.clientWidth : 0;
```
davepagurek, line 738: Are there any implications to using `document.documentElement.clientWidth` rather than `window.innerWidth`?

limzykenneth: `window.innerWidth` includes scrollbar while `document.documentElement.clientWidth` does not, which means the former will include areas that will be covered up by the scrollbar while the later will only be the visible area.

limzykenneth: It is kind of a different behavior although the previous behavior feels a bit undefined because it fallback on these different definition, at the same time I think `document.documentElement.clientWidth` is more likely to be the desired value. Can change it back though if there are edge cases that will be messed up by this.

davepagurek: I think that makes sense then, we can just make sure we test this a bit when it's in an RC to see if there are any edge cases it breaks

limzykenneth: One thing I can think of is that if the sketch is created when the page does not have scroll bars but later on in the page lifetime it gained a scroll bar, that may be a bit unexpected since the value of `document.documentElement.clientWidth` changed in this case while `window.innerWidth` would stay the same. Not sure if it would be a problem though.

L1lith:
https://github.com/processing/p5.js/issues/6830#issuecomment-4188501655 My tests show that this PR does not fix the server side import issues yet. See this comment for more, thanks.

limzykenneth:
@L1lith Just updated, can you try https://pkg.pr.new/p5@c071281 and see if it works. It is working for me locally with the only caveat being that there won't be top level function usage check which probably don't apply in this case when using instance mode.

L1lith:
> @L1lith Just updated, can you try https://pkg.pr.new/p5@c071281 and see if it works. It is working for me locally with the only caveat being that there won't be top level function usage check which probably don't apply in this case when using instance mode.

The import issues seem to be fixed, thanks! I haven't tried using it to any serious capacity beyond just doing imports however.

Also I'm curious when these changes will be published on NPM on the main p5 package?

limzykenneth:
@L1lith I've tried some basic instance mode sketch and it seems to work, the sketch would need to be put behind something like `createEffect()` to ensure it initialize on the client though.

Once this is merged and a new version is released, this will be available directly on NPM as usual.

PR #8359, opened 2025-12-25
Fix LogicalExpression (&&, ||) handling in p5.strands

### LINKED ISSUE #8358: [p5.js 2.0 Bug Report]: [p5.strands] && and || in if conditions not working correctly

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.1.2

### Web browser and version

Any

### Operating system

Any

### Steps:
1. Create a p5.strands shader
2. Include an if statement with a compound condition

Take this `if`:
```js
let c = [1, 1, 1, 1]
// Doesn't work
if (abs(inputs.texCoord.x - 0.5) > 0.2 || abs(inputs.texCoord.y - 0.5) > 0.2) {
  c = [1, 0, 0, 1]
}
```

If I put a debugger statement in there, it's getting transpiled to this:
```js
let c = __p5.strandsNode([
    1,
    1,
    1,
    1
])
{
    c = __p5.strandsNode(c)
    const __block_2 = __p5.strandsIf(abs(inputs.texCoord.x.sub(0.5)).greaterThan(0.2) || abs(inputs.texCoord.y.sub(0.5)).greaterThan(0.2), () => {
        let __copy_c_0 = c.copy()
        __copy_c_0 = __p5.strandsNode([
            1,
            0,
            0,
            1
        ])
        return { c: __copy_c_0 }
    }).Else(() => {
        let __copy_c_1 = c.copy()
        return { c: __copy_c_1 }
    })
    c = __block_2.c
}
```

Note that there's still an `||`, which doesn't work because these are not actual booleans. It should be getting transpiled to a `.or(...)`.

I *thought* this bit of the transpiler code would do that, but apparently it's not:
https://github.com/processing/p5.js/blob/7fa997eea0a0a4eb4a18d69f70cc5f47a1170758/src/strands/strands_transpiler.js#L247-L277

### Snippet:

Live: https://editor.p5js.org/davepagurek/sketches/5MtHbKB-A

```js
let sh
async function setup() {
  createCanvas(windowWidth, windowHeight, WEBGL)
  sh = baseFilterShader().modify(() => {
    getColor((inputs) => {
      let c = [1, 1, 1, 1]
      // Doesn't work
      if (abs(inputs.texCoord.x - 0.5) > 0.2 || abs(inputs.texCoord.y - 0.5) > 0.2) {
        c = [1, 0, 0, 1]
      }
      
      // Works
      // if (abs(inputs.texCoord.x - 0.5) > 0.2 ) {
      //   c = [1, 0, 0, 1]
      // }
      // if (abs(inputs.texCoord.y - 0.5) > 0.2 ) {
      //   c = [1, 0, 0, 1]
      // }
      
      return c
    })
  })
}

function draw() {
  filter(sh)
}
```

Aayushdev18:
### Summary
Adds support for `LogicalExpression` nodes in the p5.strands transpiler so that `&&` and `||` conditions are correctly handled in `if` statements.

### Changes
- Added `LogicalExpression` handling alongside existing `BinaryExpression` logic
- Ensures both operands are transpiled and logical operators are preserved
- Added a unit test covering `&&` and `||` conditions

### Tests
- Added a unit test that fails on `dev-2.0` before this change and passes after

Resolves #8358

### DIFF
--- a/test/unit/webgl/p5.Shader.js
+++ b/test/unit/webgl/p5.Shader.js
@@ -733,19 +733,21 @@ suite('p5.Shader', function() {
         assert.approximately(centerPixel[0], 255, 5);
         assert.approximately(centerPixel[1], 255, 5);
         assert.approximately(centerPixel[2], 255, 5);
-        const edgePixel = myp5.get(5, 25);
-        assert.approximately(edgePixel[0], 255, 5);
-        assert.approximately(edgePixel[1], 0, 5);
-        assert.approximately(edgePixel[2], 0, 5);
+        const leftEdgePixel = myp5.get(5, 25);
+        assert.approximately(leftEdgePixel[0], 255, 5);
+        assert.approximately(leftEdgePixel[1], 0, 5);
+        assert.approximately(leftEdgePixel[2], 0, 5);
+        const topEdgePixel = myp5.get(25, 5);
+        assert.approximately(topEdgePixel[0], 255, 5);
+        assert.approximately(topEdgePixel[1], 0, 5);
+        assert.approximately(topEdgePixel[2], 0, 5);
       });
       test('handle if statement with && (AND) operator', () => {
         myp5.createCanvas(50, 50, myp5.WEBGL);
         const testShader = myp5.baseMaterialShader().modify(() => {
-          const condition1 = myp5.uniformFloat(() => 1.0);
-          const condition2 = myp5.uniformFloat(() => 1.0);
           myp5.getPixelInputs(inputs => {
             let color = myp5.float(0.0);
-            if (condition1 > 0.5 && condition2 > 0.5) {
+            if (inputs.texCoord.x > 0.5 && inputs.texCoord.y > 0.5) {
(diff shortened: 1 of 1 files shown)


davepagurek:
Thanks for taking this on! For the OR test, can we test both the left and the top? in the original buggy example, the left was already showing up red, so the rest as written would pass. But the buggy version of the code was not correctly making the top red because only part of the of condition was being evaluated.

For the AND test, rather than tearing right in the middle, which is where thr color should switch in this example, can we test a sample in each of the corners? just the bottom right should be filled in I think.

davepagurek:
@all-contributors please add @Aayushdev18 for code

PR #8360, opened 2025-12-25
Dependencies updates

limzykenneth:
Changes:

Updating dependencies where relevant and corresponding build/test updates as well.

WIP, this probably don't need to make it into 2.2 and can make 2.3.


limzykenneth:
@davepagurek CI failed on type tests

> Error: test/types/typography.ts(50,23): error TS2550: Property 'at' does not exist on type 'Vector[]'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2022' or later.

Is there meant to be a tsconfig for us to define these things?

davepagurek:
@limzykenneth not yet, but we could update the script in package.json to be `npx tsc --project test/types/tsconfig.json --noEmit test/types/*.ts`, and then create `test/types/tsconfig.json` with:
```json
{
  "compilerOptions": {
    "lib": "es2022"
  }
}
```

src/webgl/material.js
limzykenneth, line 3716: @davepagurek I'm not entirely sure what the error here means, do you have some idea where it could have occurred?

limzykenneth:
@ksen0 @davepagurek Once the test error is sorted I'd like to get this merged or there will be more and more merge conflict here and in other PRs after this has been merged. There are still some maintenance stuff I'd still like to do but I can do them as later PRs.

src/webgl/material.js
davepagurek:
```
TypeError: Cannot read properties of undefined (reading 'ensureCompiledOnContext')
 ❯ Renderer3D.shader src/webgl/material.js:3716:7
    3714|     this.states.setValue("userFillShader", s);
    3715|     this.states.setValue("_useNormalMaterial", false);
    3716|     s.ensureCompiledOnContext(this);
       |       ^
    3717|     s.setDefaultUniforms();
    3718|   };
```
I think that test uses a custom shader, but possibly after some changes, it's no longer returning that shader. Looks like `s`, the shader variable, is undefined here for some reason. I can take a closer look later if its not obvious how that could have happened

limzykenneth: No worries, I found the problem and just pushed a fix

limzykenneth:
@ksen0 @davepagurek All working now. Do have a look and merge if all looks good. The main breakage that may happen with existing PR is when tests load files, the root path starts from project root and not the `test` folder, ie. all paths should go from `/unit/assets/[...]` to `/test/unit/assets/[...]`.

PR #8363, opened 2025-12-27
fix  createColorPicker ignores initial colour arg on chrome

### LINKED ISSUE #8284: [p5.js 2.0 Bug Report]: createColorPicker ignores initial colour arg on chrome, mac.

- [ ] Accessibility
- [x] Color
- [ ] Data
- [x] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.1.1

### Web browser and version

142.0.7444.176 (Official Build) (arm64) 

### Operating system

macOS 14.6.1

### Steps:

A: with above platform run this sketch https://openprocessing.org/sketch/2797749 or [same on p5 web editor](https://editor.p5js.org/neill0/sketches/KCRGKGInY).  
OR 
B: run the following snippet.

Observed: three colour pickers created, all defaulting to black instead of expected blue, magenta, yellow.

### Snippet:

```js
function setup() {
  createCanvas(200, 200);
  background(200)
  createColorPicker(color(50, 0, 240))
  createColorPicker("magenta")
  createColorPicker("#F0FF00")
}
```

### results on various browsers:

Browsers yielding black:

* Chrome 142.0.7444.176 on macOS 14.6.1
* Safari 17.6 (very old!) on macOS 14.6.1

Browsers correctly yielding colours: 

* Firefox (145.0.1 (aarch64) on macOS 14.6.1

### Affected p5 versions

* The problem occurs in at least these versions of p5: v2.0.0, v2.0.5, v2.1.1
* Changing p5 version back to v1.11.11 I get the expected colours on all browsers tested.

Piyushrathoree:
Resolves #8284 

 Changes:

I updated **toString()** in **p5.Color.js** to handle the **#rrggbb** format manually. It now calculates the values directly and ensures we always return a full 6-digit hex string. This fixes the issue for initial color breaking on createColorPicker.

 Screenshots of the change:

image

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -384,15 +384,14 @@ class Color {
       
       if (format === '#rrggbb') {
         colorString = String(colorString);
-        
-        if (colorString.length <= 5) { 
-          const match = colorString.match(/^#?([0-9a-fA-F])([0-9a-fA-F])([0-9a-fA-F])/);
-          if (match) {
-            colorString = `#${match[1]}${match[1]}${match[2]}${match[2]}${match[3]}${match[3]}`;
-          }
+        if (colorString.length === 4) { 
+            const r = colorString[1];
+            const g = colorString[2];
+            const b = colorString[3];
+            colorString = `#${r}${r}${g}${g}${b}${b}`;
         }
         if (colorString.length > 7) {
-           colorString = colorString.substring(0, 7);
+            colorString = colorString.slice(0, 7);
         }
         colorString = colorString.toLowerCase();
       }
@@ -404,7 +403,6 @@ class Color {
     }
     return colorString;
   }
-
   /**
    * Checks the contrast between two colors. This method returns a boolean
    * value to indicate if the two color has enough contrast. `true` means that


Piyushrathoree:
@davepagurek could you please take a look

src/color/p5.Color.js
```diff
@@ -318,14 +318,35 @@ class Color {
...
     let colorString = serializationMap[key];
 
-    if(!colorString){
+    if (!colorString) {
       colorString = serialize(this._color, {
-        format
+        format: outputFormat
       });
+      
+      if (format === '#rrggbb') {
```
davepagurek, line 385: Is this part of the code necessary? What does `colorString` look like after we pass `hex` into `serialize`?

Piyushrathoree: just checked it, seems like it is not necessary removing it right now.

Piyushrathoree: updated @davepagurek

Piyushrathoree: > just checked it, seems like it is not necessary removing it right now.

when I try to remove it multiple tests are breaking , I'm not sure but it is converting even colors like 'rgb(100% 0% 40% / 0.8)' to string which is causing the problem and that's why this check it important.

Piyushrathoree:
@davepagurek please take a look .

src/color/p5.Color.js
davepagurek: reading this more closely, is this just turning strings like #FFF into #FFFFFF? if so, could we do this without the regex and just use the indices of characters in the string since color serialization happens so frequently in p5?

davepagurek: also I just checked this commit https://github.com/processing/p5.js/pull/8363/commits/f465feb3f802d0c0e752486432f7faaa04f69d19 and it looks like that's applying the regex length thing to ALL color formats. I originally was asking if it was necessary to adjust the output of colorjs at all, e.g. removing the whole if statement including its contents.

davepagurek:
Hey @Piyushrathoree!  Just checking in, are you still working on this one? I left some more comments on the thread above to try to clarify some things.

davepagurek:
Also there were some merge conflicts -- in `dev-2.0`, now it returns the input string to `color(...)` if it is present and if the format is not being converted -- let me know if I can explain any of those updates.

Piyushrathoree:
> Also there were some merge conflicts -- in `dev-2.0`, now it returns the input string to `color(...)` if it is present and if the format is not being converted -- let me know if I can explain any of those updates.

got it will resolve soon

src/color/p5.Color.js
Piyushrathoree: Hi @davepagurek ! I've updated the code to use string indices instead of regex. Now it directly accesses characters by their position (e.g., colorString[1], colorString[2]) and checks the length (=== 4 for short hex) rather than using regex pattern matching.
I 've tried and tested it working fine , please take look.

Piyushrathoree:
> Also there were some merge conflicts -- in `dev-2.0`, now it returns the input string to `color(...)` if it is present and if the format is not being converted -- let me know if I can explain any of those updates.

I pulled the changes first and then pushed the new changes; I didn't encounter any issue.

Piyushrathoree:
> Thanks for making those changes!

actually I was busy in my exams sorry for the delay

Piyushrathoree:
> > Thanks for making those changes!
> 
> actually I was busy in my exams sorry for the delay

PR #8366, opened 2025-12-27
fix(dom/events): ensure mouseReleased fires after DOM interactions

### LINKED ISSUE #8293: Missing mouseReleased event with createSelect DOM object in Safari

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [x] DOM
- [x] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] p5.strands
- [ ] WebGL
- [ ] Friendly Errors

### p5.js version

1.11.11

### Web browser and version

Safari Version 26.0.1 (20622.1.22.118.4).

### Operating system

MacOSX

### Steps to reproduce this

Likely a bug with Safari/DOM, but worth seeing if something could be resolved from within p5js code. When creating a dropdown DOM with createSelect() function in p5js, it seems that on Safari, interactions with that dropdown are triggering a p5js mousePressed event but not a corresponding p5js mouseReleased event. This leaves internal variable mouseIsPressed set to true, and subsequent mouse move events are being categorized as mouse draw events from within src/events/mouse.js as a result.  Problem was reproduced on two machines using MacOSX + Safari, but the problem does not occur when using Chrome or Firefox on those same machines, nor when using Safari on an iPad.

### Steps:
1.   Click on the dropdown widget to make selection
2.  Release mouse after making selection
3.  Mouse over the p5js Canvas.

### Snippet:

```js

// Demo at https://editor.p5js.org/michaelgoldwasser/sketches/Dbdib65Qp
function setup() {
  createCanvas(400, 400);
  background(220);
  noStroke();
  selector = createSelect();
  selector.option("red", color(255, 0, 0));
  selector.option("blue", color(0, 0, 255));
}

function mousePressed() {
  console.log("mousePressed fired");
}

function mouseReleased() {
  console.log("mouseReleased fired");
}

function mouseDragged() {
  fill(selector.value());
  circle(mouseX, mouseY, 5);
}

```

sidhansu10:
### Summary
Fixes an issue where interacting with p5 DOM elements (e.g. `createSelect`)
in Safari triggers `mousePressed` but not `mouseReleased`, leaving
`mouseIsPressed` stuck in a true state.

### Details
- Safari does not always propagate `mouseup` back to the canvas after DOM interactions
- This patch forwards DOM `mouseup` events to p5’s internal mouse handler
- Guarded to only trigger when `mouseIsPressed` is true
- Implemented centrally in `addElement` to cover all p5 DOM elements

Resolves #8293

### DIFF
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -481,9 +481,15 @@ p5.Element.prototype.input = function (fxn) {
 function addElement(elt, pInst, media) {
   const node = pInst._userNode ? pInst._userNode : document.body;
   node.appendChild(elt);
+  elt.addEventListener('mouseup', e => {
+    if (pInst && pInst.mouseIsPressed && typeof pInst._onmouseup === 'function') {
+      pInst._onmouseup(e);
+    }
+  });
   const c = media
     ? new p5.MediaElement(elt, pInst)
     : new p5.Element(elt, pInst);
+
   pInst._elements.push(c);
   return c;
 }
--- a/src/dom/dom.js
+++ b/src/dom/dom.js
@@ -481,11 +481,6 @@ p5.Element.prototype.input = function (fxn) {
 function addElement(elt, pInst, media) {
   const node = pInst._userNode ? pInst._userNode : document.body;
   node.appendChild(elt);
-  elt.addEventListener('mouseup', e => {
-    if (pInst && pInst.mouseIsPressed && typeof pInst._onmouseup === 'function') {
-      pInst._onmouseup(e);
-    }
-  });
   const c = media
     ? new p5.MediaElement(elt, pInst)
     : new p5.Element(elt, pInst);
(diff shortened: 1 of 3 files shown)


src/dom/dom.js
```diff
@@ -481,9 +481,15 @@ p5.Element.prototype.input = function (fxn) {
 function addElement(elt, pInst, media) {
   const node = pInst._userNode ? pInst._userNode : document.body;
   node.appendChild(elt);
+  elt.addEventListener('mouseup', e => {
+    if (pInst && pInst.mouseIsPressed && typeof pInst._onmouseup === 'function') {
+      pInst._onmouseup(e);
```
davepagurek: For non-Safari and for non-`select` elements, would this cause the event to be dispatched twice?

sidhansu10: The guard on mouseIsPressed should prevent double dispatch in browsers where mouseup already propagates correctly. If the canvas mouseup has already fired, mouseIsPressed will be false, so _onmouseup won’t be called again. This mainly addresses Safari, where mouseup often doesn’t reach the canvas after DOM interactions, leaving mouseIsPressed stuck true.

perminder-17 (approved):
Looks good to me! thanks for your work on this

Issue #8364, opened 2025-12-27
[p5.js 2.0 Bug Report]: filters don't work in .min.js builds


davepagurek:
### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [ ] Other (specify if possible)

### p5.js version

2.2-rc.3

### Web browser and version

All

### Operating system

All

### Steps:
1. Load p5 from a .min.js file
2. Create a WebGL or WebGPU canvas
3. Use a filter

You get errors like: `ReferenceError: r is not defined`.

Here's an example of the issue when calling `panorama(img)`. [Here's](https://github.com/processing/p5.js/blob/9512db6747942ce878a18ce9be9f37068c4e463d/src/core/p5.Renderer3D.js#L1914-L1942) the original code:

```js
this.sphereMapping = this.baseFilterShader().modify(() => {
  const uEnvMap = p5.uniformTexture();
  const uFovY = p5.uniformFloat();
  const uAspect = p5.uniformFloat();
  // Hack: we don't have matrix uniforms yet; use three vectors
  const uN1 = p5.uniformVec3();
  const uN2 = p5.uniformVec3();
  const uN3 = p5.uniformVec3();
  p5.getColor((inputs) => {
    const uFovX = uFovY * uAspect;
    const angleY = p5.mix(uFovY/2.0,  -uFovY/2.0, inputs.texCoord.y);
    const angleX = p5.mix(uFovX/2.0, -uFovX/2.0, inputs.texCoord.x);
    let rotatedNormal = p5.normalize([angleX, angleY, 1]);
    rotatedNormal = [
      // Don't mind me, just doing matrix vector multiplication...
      p5.dot(rotatedNormal, uN1),
      p5.dot(rotatedNormal, uN2),
      p5.dot(rotatedNormal, uN3),
    ];
    const temp = rotatedNormal.z;
    rotatedNormal.z = rotatedNormal.x;
    rotatedNormal.x = -temp;
    const suv = [
      p5.atan(rotatedNormal.z, rotatedNormal.x) / (2.0 * p5.PI) + 0.5,
[...]

PR #8367, opened 2025-12-27
Fix p5.strands filters not working on minified code

davepagurek:
Resolves https://github.com/processing/p5.js/issues/8364

### Changes

So this stems from the fact that p5.strands code gets transpiled, and when we run the result, it's not in the same runtime context as before. It has no access to outside variables. So we had previously made a way to inject those variables: pass them in as an object in a second parameter after your function, and we'll automatically feed them in as parameters to your transpiled function with the original names so that you have access again:

```js
const myp5 = something;
baseFilterShader.modify(() => {
  myp5.someCall(...)
}, { myp5 })

// Turns into this after transpilation:

(myp5) => {
  myp5.someCall(...)
}
```

The problem was that minified js and ts code also change the variable names, *but not the keys of that object you pass in.* Previously it worked because those were always the same, but minification makes that no longer be the case. The above code could turn into something like this:

```js
const e = something;
baseFilterShader.modify(() => {
  e.someCall(...)
}, { myp5: e })

// Turns into this after transpilation:

(myp5) => {
  e.someCall(...) // e is not defined!
}
```

To address that, there are two possible routes:
- Update the function so it doesn't get minified
- Update the context object so that the minifier uses the right name

The first option is easier (don't write a function, write a source code string), and that would work for our library, but would make js/ts builds a lot more cumbersome. No one really likes writing code in a string. So I moved on to the second option.

There's already a solution along the lines of the second option in the wild. In Puppeteer, when you write a function in nodejs to run in a Chrome window, it's also recreating a new function. The way it handles it is that it doesn't automatically inject variables for you, you have to manually grab them from function arguments. So that's what I've opted for here:

```js
// Original code:

const myp5 = something;
baseFilterShader.modify(({ myp5 }) => {
  myp5.someCall(...)
}, { myp5 })

// After minification:

const e = something;
baseFilterShader.modify(({ myp5: f }) => {
  f.someCall(...)
}, { myp5: e })

// Turns into this after transpilation:

({ myp5: f }) => {
  f.someCall(...) // all good now!
```

This is more verbose than before, but I think it's the only way we can avoid this issue.

I also noticed that sometimes the minifier turns for loops like this:
```js
for (let i = 0; i < 5; i++) {
  a += i;
  b += i;
}
```
...into something like this for compactness, with comma separated statements:
```js
for (let i = 0; i < 5; i++) {
  a += i, b += i;
}
```

This wasn't getting handled by our transpiler because we were trying to whitelist the statements we need to traverse into to find variable updates in a loop. I've updated that code to now use the default Acord tree traversal, and manually skip the ones we don't want to traverse into.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/strands/strands_transpiler.js
+++ b/src/strands/strands_transpiler.js
@@ -271,6 +271,37 @@ const ASTCallbacks = {
       };
       node.arguments = [node.right];
     },
+    LogicalExpression(node, _state, ancestors) {
+      // Don't convert uniform default values to node methods, as
+      // they should be evaluated at runtime, not compiled.
+      if (ancestors.some(nodeIsUniform)) { return; }
+      // If the left hand side of an expression is one of these types,
+      // we should construct a node from it.
+      const unsafeTypes = ['Literal', 'ArrayExpression', 'Identifier'];
+      if (unsafeTypes.includes(node.left.type)) {
+        const leftReplacementNode = {
+          type: 'CallExpression',
+          callee: {
+            type: 'Identifier',
+            name: '__p5.strandsNode',
+          },
+          arguments: [node.left]
+        }
+        node.left = leftReplacementNode;
+      }
+      // Replace the logical operator with a call expression
+      // in other words a call to BaseNode.or(), .and() etc.
+      node.type = 'CallExpression';
+      node.callee = {
+        type: 'MemberExpression',
+        object: node.left,
+        property: {
+          type: 'Identifier',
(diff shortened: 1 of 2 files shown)


Aayushdev18:
Thanks for the detailed explanation! I read through the PR and learned a lot about how minification interacts with transpilation. The destructuring approach is really clever, I wouldn't have thought of that. Appreciate you taking the time to document the process.

PR #8383, opened 2026-01-04
Fix FES double error when friendlyStack is null (#8381)

### LINKED ISSUE #8381: FES error masks actual undefined variable ReferenceError

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] p5.strands
- [ ] WebGL
- [x] Friendly Errors

### p5.js version

2.1.2

### Web browser and version

Chrome

### Operating system

macOS

### Steps:
1. Open the example sketch: https://editor.p5js.org/davepagurek/sketches/aVaqe0sNn
2. Run the sketch (or create a new sketch with the code below)
3. Open the browser console
4. Observe that two errors appear instead of one

**Problem:**
When a variable is undefined (e.g., due to a typo or forgetting to define it), p5.js shows two errors:
1) An error from FES trying to make an error
2) The actual ReferenceError (e.g., "s is not defined")

The first FES error masks the real error message, making debugging harder.

**Expected behavior:**
Users should see a single, clear error message indicating the undefined variable.

**Actual behavior:**
Two errors appear — the FES error and the actual ReferenceError.

**Additional context:**
This is not p5.strands-specific; it also happens in regular p5.js.

### Snippet:
```js
warp = baseStrokeShader().modify(() => {
  const frame = uniformFloat(() => frameCount)

  getObjectInputs((inputs) => {
    inputs.position.xy += 20 * noise(
      inputs.position.x * s,
      inputs.position.y * s,
      frame * 0.01
    )
  })

  return inputs
})


Note:
The variable s is intentionally undefined in this example.

Aayushdev18:
Resolves #8381

## Changes:
Fixes a bug where the Friendly Error System (FES) would throw its own error when handling a `ReferenceError` (e.g., undefined variable), causing a double error that masks the actual error message.

**Root Cause:**
When `fesErrorMonitor()` processes a `ReferenceError`, it attempts to access `friendlyStack[0].lineNumber` to provide location information. However, `processStack()` can return `null` for `friendlyStack` in certain cases, causing FES itself to throw a `TypeError` when accessing `friendlyStack[0]`.

**Solution:**
Added defensive null checks before accessing `friendlyStack[0].lineNumber` in `fesErrorMonitor()` to prevent FES from throwing its own error.

**Files Changed:**
- `src/core/friendly_errors/fes_core.js`: Added null checks for `friendlyStack` and `friendlyStack[0]` before accessing properties
- `test/unit/core/fes_error_monitoring.js`: Added unit test to verify FES handles null `friendlyStack` gracefully

**Testing:**
- All existing tests pass (1892/1892)
- New test verifies FES does not throw when `friendlyStack` is null
- Fix prevents the double error issue described in #8381

#### PR Checklist
- [x] `npm run lint` passes
- [ ] [Inline reference] is included / updated (N/A - internal bug fix)


davepagurek:
Would it be possible to create a test that includes a reference error like in the original example rather than manually creating a `ReferenceError`? This was the stack trace in the original issue:

image

It looks like it goes through the `handleMisspelling` function, and it only does that in here based on the `.type` property:
https://github.com/processing/p5.js/blob/9799eada451dea7f2b50fe5f52ffd725463b343a/src/core/friendly_errors/fes_core.js#L813-L823

I'm not certain that the manually created error object has the right property for that, and so the test here might not be hitting the same issue as in the example. It also seems like a different problem than the one mentioned in your PR description:

> When fesErrorMonitor() processes a ReferenceError, it attempts to access friendlyStack[0].lineNumber to provide location information. However, processStack() can return null for friendlyStack in certain cases, causing FES itself to throw a TypeError when accessing friendlyStack[0].

Did you have an example that produces the above problem? Or is this something that an LLM decided based on the issue and the codebase?

Rather than letting an LLM find the issue, I'd encourage you to run the example sketch https://editor.p5js.org/davepagurek/sketches/aVaqe0sNn with your browser's dev tools open. It should pause on the error:

image

Then you can navigate through the stack trace to see how it got there.

Aayushdev18:
Thanks for pointing me to the example sketch! I ran it with dev tools open and found the actual root cause.

**What I Found:**
When I ran the example sketch, I saw the error stack trace you mentioned:
- TypeError: can't access property "pixels", this._renderer is undefined
- This happens in: get pixels → getSymbols → defineMisusedAtTopLevelCode → handleMisspelling

**Root Cause:**
The issue is that getSymbols(fn) is called with fn (the p5 prototype), not a p5 instance. When it checks typeof obj[name] for properties like 'pixels', it triggers the getter, which tries to access this._renderer.pixels, but this._renderer is undefined on the prototype.

**The Fix:**
I've updated the fix to use Object.getOwnPropertyDescriptor to check if a property is a getter before accessing it. If it's a getter, we skip it to avoid triggering instance property access when obj is the prototype.

I tested it locally with the example sketch and confirmed it now shows only the original ReferenceError, no FES TypeError. The fix is working correctly.

Aayushdev18:
and also the fix checks `descriptor.get` to skip any getter property, not just 'pixels'. 
This makes it comprehensive,, it will handle all getters that access instance properties (like drawingContext, width, height, canvas, etc.) since getSymbols is called with the prototype, not an instance.

davepagurek:
Would that make it not catch variables that are misspellings of getter properties like `width`? I think rather than skipping getter properties, we just need to make sure that we are using the current p5 instance instead of `fn`.

Aayushdev18:
Thanks for the feedback! I've updated the fix to use `p5.instance` instead of the prototype when checking property types, as you suggested. This allows getters like `width` and `pixels` to be accessed safely on an instance where instance properties exist.
When no instance is available (instance mode or before initialization), it falls back to using `Object.getOwnPropertyDescriptor()` to avoid crashes. This preserves misspelling detection for getter properties while preventing the TypeError from getters that access instance properties.
I tested it locally with the example sketch and the unit test passes. The fix now shows only the original ReferenceError without any FES crash.

davepagurek:
I think this approach works, thanks. The test file might still need a bit of cleaning up, as currently it sets up something to track logs sent to FES, but then doesn't use it, instead doing a custom error handler in the test.

Does it work using the `logs` variable instead of a window error handler? If so, maybe we don't need the custom window error handler.

Otherwise, I think we can remove the `logs` setup at the top of the test file, since it's not getting looked at right now. Also, we maybe want to resolve the test in the error handler if we're going this route, rather than using a 200ms `setTimeout`. (If some code changes and the error stops triggering, the test will time out anyway because vitest has test time limits already, so it will still fail.) We can put the setup and cleanup of the `window` handler in `beforeEach` and `afterEach` so that things get cleaned up even if the test fails and times out.

perminder-17 (commented):
Closing this PR since we can't reproduce this bug anymore.  Thanks for your work @Aayushdev18

PR #8384, opened 2026-01-05
Auto return shader hooks

### LINKED ISSUE #7994: [p5.strands] Auto-return input objects if omitted

vietnguyen2358:
Resolves #7994 
Addresses #7992 

 Changes:

- Added auto-return behavior for shader hooks with matching input and return types, allows users to omit `return inputs` when modifying struct properties in place
- Implemented type checking to ensure auto-return only applies to struct-returning hooks with a single parameter of matching type

#### PR Checklist

- [X] `npm run lint` passes


vietnguyen2358:
Sure, I will look into it.

davepagurek:
Hi @vietnguyen2358! We're working on some updates to the p5.strands API for the next release in https://github.com/processing/p5.js/pull/8314 and we don't want to put any time pressure on you related to that, so I'm going to integrate your PR into that branch and fix the tests there, since other parts of the code are moving around too. Thanks for your help on this so far, I'm excited to have this functionality in the next version 🙂

davepagurek:
@all-contributors please add @vietnguyen2358 for code

Issue #8387, opened 2026-01-07
[p5.js 2.0 Bug Report]: Warnings about omitting strokes on high-detail 3D primitives don't use FES


davepagurek:
### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [ ] Other (specify if possible)

### p5.js version

All

### Web browser and version

All

### Operating system

All

### Steps:
1. Turn off FES
2. Draw a cylinder with a high level of detail
3. It still logs a warning

That's because these sorts of warnings are raw `console.log`s in the codebase:
https://github.com/processing/p5.js/blob/9799eada451dea7f2b50fe5f52ffd725463b343a/src/webgl/3d_primitives.js#L2582-L2585

### Snippet:

```js
p5.disableFriendlyErrors = true
function setup() {
  createCanvas(400, 400, WEBGL);
  cylinder(20, 100, 40, 40)
}
```

Additionally, this also triggers when you are drawing to a p5.Geometry, i.e.:
```js
function setup() {
  createCanvas(400, 400, WEBGL);
  const myGeom = buildGeometry(() => cylinder(20, 100, 40, 40))
  noStroke()
  model(geom)
}
```
Arguably we should just silence this error if inside of `buildGeometry` since we don't know if it will eventually be drawn with strokes or not, to avoid misleading false positives.

Aayushdev18:
I’d like to work on this. The warning appears to bypass FES due to a raw console.log. I’ll investigate fixing or silencing it appropriately

shuklaaryan367-byte:
This does look like a clear FES consistency issue.

From reading the code, it seems the warning for high-detail cylinders is currently emitted via a raw console.log, so it still appears even when p5.disableFriendlyErrors = true. This also seems to trigger inside buildGeometry(), where stroke usage isn’t yet known, which could lead to misleading warnings.

I’d be interested in helping fix this and align the behaviour with the Friendly Error System if that sounds reasonable.
Thank you 😇

skyyash:
i agree this can be misleading for beginners. i’m happy to take this if no one’s on it?

[...]

PR #8411, opened 2026-01-13
Suppress stroke warnings during buildGeometry

skyash-dev:
Addresses #8387 

Changes:
Suppresses high-detail stroke warnings while geometry is being constructed via `buildGeometry()` / `beginGeometry()`.
During geometry construction, p5.js cannot yet know whether the resulting geometry will be rendered with strokes, so emitting warnings at that stage can be misleading. The warning behavior during normal drawing remains unchanged.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/3d_primitives.js
+++ b/src/webgl/3d_primitives.js
@@ -993,7 +993,7 @@ p5.prototype.plane = function(
     planeGeom.computeFaces().computeNormals();
     if (detailX <= 1 && detailY <= 1) {
       planeGeom._makeTriangleEdges()._edgesToVertices();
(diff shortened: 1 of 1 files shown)


davepagurek:
Thanks for making this! Right now it's still using console.log so this will show up even though `p5.disableFriendlyErrors = true` is set. Do you think we could update it to go through FES's logger?

davepagurek:
Discussed some more in the issue, there isnt actually a warning logger yet, so we're going to leave this as is for `main` but set up something for that in `dev-2.0`.

PR #8417, opened 2026-01-14
Fix `noiseDetail()` in p5.js 2.x so that the `falloff` parameter defaults to `0.5` when omitted.

### LINKED ISSUE #8402: [p5.js 2.0 Bug Report]: noiseDetail(1) triggers Friendly Error even though second parameter is optional

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [x] Friendly errors

### p5.js version

2.1.2

### Web browser and version

Google Chrome Version 143.0.7499.170

### Operating system

Windows 11

### Steps:

1. Open the p5.js Web Editor

2. Ensure Friendly Errors are enabled (default)

3. Run the sketch below

4. Observe a Friendly Error warning in the console

### Snippet:

```js

function setup() {
  noCanvas();
  noiseDetail(1);
  console.log(noise(0.5));
}


```
**Expected behaviour:**

According to the documentation, the second argument to noiseDetail() (falloff) is optional because a default value (0.5) is defined. Calling noiseDetail(1) should therefore be valid and should not trigger a Friendly Error warning.

**Actual behaviour:**

p5.js displays a Friendly Error warning:
_“Expected number at the second parameter in noiseDetail()”_
even though the call is valid and the sketch runs correctly.

**Additional confirmation:**

When p5.disableFriendlyErrors = true is set, the warning disappears, confirming this is an FES validation bug rather than a problem with noiseDetail() itself.

_I’m happy to help implement a fix if this is approved._

shuklaaryan367-byte:
Resolves #8402

 Changes:
 - Add default value for `falloff` in runtime implementation
- Update Friendly Error System validation to allow a single-argument overload

 Screenshots of the change:
 Not applicable (non-visual change) 

#### PR Checklist

- [ ] `npm run lint` passes


davepagurek:
Thanks for taking this on! This looks good so far, although we also override `noiseDetail` here when p5.strands is active, so I think we should add the default parameter value there too: https://github.com/processing/p5.js/blob/a491a9ee9e83703579b37955f7e48e2b22c56608/src/strands/strands_api.js#L168-L175

davepagurek:
Oh one other thing, you don't have to manually update parameterData.json, running `npm run docs` will generate that.

shuklaaryan367-byte:
Added default parameter handling to the noiseDetail override in strands_api.js as suggested.
Left parameterData.json unchanged since it’s auto-generated via npm run docs.

davepagurek:
@all-contributors please add @shuklaaryan367-byte for code

PR #8418, opened 2026-01-15
Fix WebGL shader and texture memory leak on sketch removal

Sanchit2662:
## Summary

This PR fixes a critical WebGL lifecycle issue where shader programs and textures were never explicitly released from GPU memory when a sketch was destroyed. As a result, applications that repeatedly create and remove WebGL sketches (e.g. hot-reload editors, instance-mode apps, long-running installations) experienced unbounded GPU memory growth.

The fix introduces explicit `dispose()` methods for WebGL-backed resources and registers a renderer-level cleanup hook that is automatically invoked when `sketch.remove()` is called.

---

## Impact

**Before**

* WebGL shaders, programs, and textures remain allocated after `remove()`
* GPU memory grows unboundedly in multi-sketch or hot-reload scenarios
* No user-level workaround is possible

**After**

* All WebGL shader and texture resources are explicitly released
* GPU memory remains stable across repeated sketch creation/destruction
* Instance-mode and long-running WebGL sketches behave reliably

---

### 1. `p5.Shader.dispose()`

Adds an explicit teardown path for shader GPU resources.

```js
dispose() {
  if (!this._gl || !this._glProgram) return;

  this._glProgram = null;
}
```

* Properly detaches shaders before deletion
* Deletes both shader objects and the linked program
* Guards against double deletion or uninitialized shaders

---

### 2. `p5.Texture.dispose()`

Ensures GPU textures are freed when no longer needed.

```js
dispose() {
  if (this.glTex && !this.isFramebufferTexture) {
  }
}
```

* Releases GPU texture memory
* Skips framebuffer-managed textures to avoid double-freeing

---

### 3. Renderer-level cleanup hook (`p5.RendererGL`)

Introduces a centralized cleanup method that disposes all WebGL resources owned by the renderer.

Key resources cleaned:

* Default and user shaders
* Cached textures (`this.textures`)
* Filter shaders and filter layers
* Framebuffers and retained-mode buffers
* Empty texture singleton

The cleanup is automatically invoked on sketch teardown:

```js
this._pInst.registerMethod(
  'remove',
  this._cleanupWebGLResources.bind(this)
);
```

---

#### PR Checklist

* [x] `npm run lint` passes

### DIFF
--- a/src/webgl/p5.RendererGL.js
+++ b/src/webgl/p5.RendererGL.js
@@ -699,8 +699,8 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
    * @private
    */
   remove() {
-    // Dispose all cached shaders
-    const shadersToDispose = [
+    // Remove all cached shaders
+    const shadersToRemove = [
       this._defaultLightShader,
       this._defaultImmediateModeShader,
       this._defaultNormalShader,
@@ -715,25 +715,25 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
       this.filterShader
     ];
 
-    // Also dispose filter shaders
+    // Also add filter shaders
     if (this.defaultFilterShaders) {
       for (const key in this.defaultFilterShaders) {
-        shadersToDispose.push(this.defaultFilterShaders[key]);
+        shadersToRemove.push(this.defaultFilterShaders[key]);
       }
     }
 
-    // Dispose each shader
-    for (const shader of shadersToDispose) {
-      if (shader && typeof shader.dispose === 'function') {
-        shader.dispose();
+    // Remove each shader
+    for (const shader of shadersToRemove) {
+      if (shader && typeof shader.remove === 'function') {
+        shader.remove();
       }
     }
 
-    // Dispose all cached textures
+    // Remove all cached textures
     if (this.textures) {
       for (const texture of this.textures.values()) {
-        if (texture && typeof texture.dispose === 'function') {
-          texture.dispose();
+        if (texture && typeof texture.remove === 'function') {
+          texture.remove();
         }
       }
       this.textures.clear();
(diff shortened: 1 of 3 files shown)


Sanchit2662:
Hi @davepagurek,
This PR fixes a WebGL lifecycle issue where shader programs and textures weren’t being released when a sketch was removed, which could lead to unbounded GPU memory growth in long-running or hot-reload scenarios. The update adds explicit cleanup for WebGL resources and ensures everything is properly disposed on sketch.remove().

Whenever you have time, I’d appreciate a review. Thanks!

davepagurek:
Thanks for taking this on! Some small thoughts:
- Could we use the `remove()` naming everywhere for consistency rather than introducing `dispose()` for some things?
- This PR is into the `main` branch, which is just for 1.x code. The `dev-2.0` branch has our work on p5 2.x, which will become default sometime this year. If you're up for it, before calling the issue resolved, we should make similar changes there, although it's a bit more complicated because we have both WebGL and WebGPU support for these objects. It's likely complicated enough that we'd want to make some unit tests there too.

src/webgl/p5.RendererGL.js
```diff
@@ -680,6 +680,127 @@ p5.RendererGL = class RendererGL extends p5.Renderer {
     this.fontInfos = {};
 
     this._curShader = undefined;
+
+    // Register cleanup hook to free WebGL resources when sketch is removed
+    this._pInst.registerMethod('remove', this._cleanupWebGLResources.bind(this));
```
davepagurek: The tests are also showing errors that this method is not defined when inside of `createGraphics`

Sanchit2662:
Hi @davepagurek ,
Thanks for the feedback! I’ve updated the fix so it avoids breaking changes and ensures proper WebGL cleanup for createGraphics() via p5.Graphics.remove(). All tests are passing now.

src/webgl/p5.Texture.js
```diff
@@ -453,6 +453,26 @@ p5.Texture = class Texture {
...
+  /**
+   * Frees the GPU resources associated with this texture.
+   *
+   * This method deletes the WebGL texture from GPU memory. Call this when
+   * you no longer need the texture to prevent memory leaks.
+   *
+   * @method dispose
+   * @private
+   */
+  dispose() {
```
davepagurek: Hi! Could we turn this into `remove()` as well for consistency?

Sanchit2662: Hi @davepagurek ,
Fixed — switched this to remove() for consistency. Thanks for pointing it out!

davepagurek:
Thanks again for implementing these methods. Once merged, would you be interested in doing something similar in the `dev-2.0` branch next?

davepagurek:
@all-contributors please add @Sanchit2662 for code

PR #8430, opened 2026-01-21
Normalize strands noise output range to match JS noise()

### LINKED ISSUE #8386: [p5.js 2.0 Bug Report]: p5.strands noise has a different range than non-strands noise

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.1.2

### Web browser and version

Firefox

### Operating system

MacOS

### Steps:
1. Create a p5.strands shader that offsets position with noise
2. Draw something with the shader

It moves both above and below the initial value. Here's an example showing three lines:
- In white, the initial line at y=0
- In red, the line offset by a strands noise function
- In blue, the line drawn with noise added outside of strands

The blue line only ever adds to `y`, pushing it down. The red line moves it up half the time and down half the time.

Image

(This maybe also explains why the shader noise visual test was failing, since it was outputting negative colours.)

While it's not necessarily important for the shader implementations to match exactly (there are performance reasons why it probably should be different), it'll probably lead to confusion if the output range for noise is different. It confused me, at least!

I think there's definitely a use for noise centered at 0, but that should maybe be a different noise mode option if we add noiseMode.

### Snippet:

```js
let lineGeom;
let wobble;

function setup() {

  lineGeom = buildGeometry(() => {
    noFill();
    beginShape();
    const n = 400;
    for (let i = 0; i <= n; i++) {
      vertex(map(i, 0, n, -width / 2, width / 2), 0);
    }
    endShape();
  });

  wobble = baseStrokeShader().modify(() => {
    const t = uniformFloat(() => millis());
    const size = uniformVec2(() => [width, height]);
    getWorldInputs((inputs) => {
      noiseDetail(1, 0.5);
      inputs.position.y +=
        noise((inputs.position.x / size.x) * 3, t * 0.0001) * size.y * 0.5;
      return inputs;
    });
  });
}

function draw() {
  background(0);
  push();
  strokeWeight(5);
  push();
  stroke("red");
  strokeShader(wobble);
  model(lineGeom, 50);
  pop();
  stroke("blue");
  beginShape();
  const n = 100;
  const t = millis()
  noFill()
  noiseDetail(1, 0.5)
  for (let i = 0; i <= n; i++) {
    vertex(map(i, 0, n, -width / 2, width / 2), height*0.5*noise(i/n * 3, t*0.0001));
  }
  endShape();
  stroke("white");
  model(lineGeom);
  pop();
}
```

Live: https://editor.p5js.org/davepagurek/sketches/lXuGsdnK6

rakesh2OO5:
This PR normalizes the output range of the strands noise implementation to match the JavaScript `noise()` function.

Changes:
- Normalize GLSL backend output from [-1,1] to [0,1]
- Normalize WGSL backend output from [-1,1] to [0,1]

Tested:
- Verified GPU (WEBGL and WebGPU) noise output remains within [0,1]
- Matches CPU `noise()` output range

Fixes #8386

### DIFF
--- a/vitest.workspace.mjs
+++ b/vitest.workspace.mjs
@@ -46,13 +46,12 @@ export default defineWorkspace([
           capabilities: process.env.CI ? {
             'goog:chromeOptions': {
               args: [
-                '--no-sandbox',
                 '--headless=new',
-                '--enable-unsafe-webgpu',
-                '--use-vulkan=swiftshader',
-                '--use-webgpu-adapter=swiftshader',
-                '--use-angle=vulkan',
                 '--no-sandbox',
+                '--enable-webgl',
+                '--use-gl=angle',
+                '--use-angle=swiftshader-webgl',
+                '--enable-unsafe-swiftshader',
               ]
             }
           } : undefined
@@ -105,7 +104,6 @@ export default defineWorkspace([
                 '--use-vulkan=swiftshader',
                 '--use-webgpu-adapter=swiftshader',
                 '--use-angle=vulkan',
-                '--no-sandbox',
               ]
             }
           } : undefined


rakesh2OO5:
CI is awaiting maintainer approval since this is from a fork — happy to update the branch if needed.

davepagurek:
I think this works, although is there a reason for doing the normalization differently in the two files? It could be easier for people in the future comparing the implementations if we use the same math in both places.

rakesh2OO5:
Good point — there’s no strong reason for them to differ. I used the equivalent forms out of habit, but I agree it’s better for readability and future comparisons if both backends use the same expression.

I have updated both GLSL and WGSL to use the same normalization math (e.g. (result + 1.0) * 0.5) and pushed a small follow-up commit.

Now , both backends use the same normalization expression for consistency — thanks for the suggestion!

davepagurek:
Thanks, code looks good! I think the last step before merging is just to test out a build of this doing something similar to the sketch in the original issue. I can get to that eventually, but if you have time, that would also be great. If you run `npm run build`, then you can upload `lib/p5.js` to a p5 web editor sketch, and edit the html to have `<script src="p5.js"></script>` rather than the usual longer CDN link to p5.

rakesh2OO5:
Tested using the original issue sketch in the p5 Web Editor with a custom built p5.js (loaded via local p5.js file).
Tested on Windows 11 / Chrome (WebGL2 enabled).
Both paths stay normalized and match numerically:

CPU noise range: 0.0119 → 0.4960

GPU (strands/shader) noise range: 0.0119 → 0.4960

Screenshot attached showing red (GPU), blue (CPU), and console output.
image

davepagurek:
That looks great, thanks @rakesh2OO5!

rakesh2OO5:
It looks like the failing job is due to a headless Chrome WebGL context initialization issue rather than the shader change itself. The change only normalizes the noise return value and doesn’t touch renderer setup. Since this is a fork PR, I don’t have permission to re-run CI — would someone with access be able to re-run the test (ubuntu-latest, chrome) job?

davepagurek:
I've been taking a look into it, and have rerun it a few times so far. Doesn't seem to change the result. The tests do work for me locally, and nothing weird has changed in your PR, so this seems like a CI issue. I did notice that between the last PR merged into dev-2.0 and this one, the chrome version used by vitest changed:
image

image

...but also chrome 144 is what i run locally and that seems to work, so I'm still not sure what the issue is just yet, the chrome version thing could be a coincidence. Let me now if you have other ideas for what might be the issue!

davepagurek:
I wonder if it's this: https://chromium.googlesource.com/chromium/src/%2B/main/docs/gpu/swiftshader.md?utm_source=chatgpt.com

Going to push something to your branch to see what happens!

davepagurek:
ahh yep ok that was it. Good to merge now!

PR #8459, opened 2026-01-27
docs(p5.strands): document smoothstep

### LINKED ISSUE #8442: [p5.strands] document smoothstep method
### Increasing access

This is a function from GLSL that is available in p5.strands, but is not documented anywhere yet.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

This should exist in a p5.strands section of the reference and should include parameters and usage examples.

jjnawaaz:
Resolves #8442

### Changes
- Added reference documentation for the `smoothstep()` function in the p5.strands API
- Documented parameters, return value, and behavior consistent with GLSL `smoothstep`
- Included a usage example demonstrating smooth opacity transitions via `combineColors`

### Screenshots of the change
Not applicable (documentation-only change).

#### PR Checklist
- [x] `npm run lint` passes


jjnawaaz:
I’ve updated the `smoothstep` docs to:
- start with a simple coordinate-based example (no uniforms) to introduce the concept clearly,
- followed by a second example that uses a uniform for animation, as a progression.

Both examples now follow the p5.strands patterns from the tutorial and should run as-is.

I also ran `npm run build:reference` on the p5.js-website repo to make sure the reference builds cleanly with these changes.

Happy to iterate further if you have more suggestions

nbogie:
Hi @jjnawaaz!
I like the simplification of the idea, but that first code example does not run for me.  I am new to p5.strands but I don't think `pixelInputs` is available in [`baseColorShader`](https://beta.p5js.org/reference/p5/basecolorshader/).

(I only checked your first example.)

I would recommend always developing, testing, and saving all examples on the p5 web editor first, finally pasting them into the reference comment only once you're happy that they run without error.   If you do need to make later modifications to the example code, I would absolutely make the change back in the p5 web editor, rather than directly in the comment-block of the .js source file, so that you can again ensure your example code runs. The back-and-forth process can be a little laborious, I would agree - that's all the more reason to get your examples as close to perfect as possible before integrating them into the comment block.

You can also include, in this conversation, links to your intended examples in the web editor while they're under discussion - that can make it a lot more convenient for you, and for reviewers, during early feedback.

Let me know if I can help!

nbogie:
Also, I could be wrong but I think the usefulness of smoothstep will be demonstrated better by a narrower transition region between the colours.  If it's just being used to make a linear transition across essentially the entire width of the canvas, then we may as well just use the normalized x value directly, without smoothstep.  I hope that makes sense!

jjnawaaz:
Hi @nbogie - thanks a lot for the detailed feedback, this is really helpful.

You’re absolutely right: pixelInputs isn’t available in baseColorShader, which explains why the first example doesn’t run. I’ll revise that example to use only inputs that are valid for baseColorShader, and I’ll make sure to test it in the p5 Web Editor before updating the reference.

Good point as well about the transition region - I agree that using smoothstep over the entire canvas doesn’t really demonstrate its purpose. I’ll tighten the transition window so the benefit of smoothstep (soft edges vs a linear ramp) is clearer.

I’ll post an updated example shortly and, if helpful, include a link to the Web Editor sketch while it’s under discussion. 
Thanks again for taking the time to review and explain!

jjnawaaz:
I’ve pushed an update addressing the review comments:

Replaced the earlier examples with tested, working p5.strands filter examples

First example now demonstrates smoothstep using only screen coordinates (no uniforms), with a narrow transition band so its effect is clear

Second example builds on that by introducing a uniform for animation

Both examples were tested in the p5 Web Editor (2.x / p5.strands) and verified via npm run build:reference in the website repo

nbogie:
Hi!  That first example is working for me now - well done.
Going back to my first point - introducing the concept of smoothstep without the need for the reader to know about a lot of other shader topics - perhaps you can think of some ways to simplify it still.  Think about all the mechanisms the short example is using and whether any can be cut out and replaced with something simpler.

jjnawaaz:
Hi Thanks, and I’m glad the example is running now.
I see your point about minimizing how many other concepts a reader has to hold in their head just to understand smoothstep. Even though the example is short, it still introduces several p5.strands / shader mechanisms alongside the function itself.
I’ll take another pass to see how far the example can be reduced so that smoothstep is doing almost all of the explanatory work, and the surrounding machinery is as lightweight as possible. I’ll focus on simplifying first, and only keep additional pieces where they directly help clarify what smoothstep does. I’ll follow up with an updated version once I’ve explored a simpler formulation. 
Appreciate the guidance.

jjnawaaz:
I’ve updated the examples to further reduce the number of concepts introduced alongside smoothstep:

The first example now shows a simple grayscale fade across the canvas using only smoothstep and a coordinate (no uniforms, no texture sampling).

The second example builds on the first by introducing a single uniform to animate the transition.

Both examples were tested in the p5 Web Editor using the p5.js 2.0 / p5.strands environment and should run as-is.

Happy to iterate further if there’s an even simpler formulation you’d like to see.

src/strands/p5.strands.js
```diff
@@ -275,6 +280,97 @@ if (typeof p5 !== "undefined") {
  * </div>
  */
 
+/**
+ * @method smoothstep
+ * @description
+ * A shader function that performs smooth Hermite interpolation between `0.0`
```
perminder-17, line 286: Everything looks good to me. I was going through the docs of [smoothstep](https://thebookofshaders.com/glossary/?search=smoothstep). 

How about adding a line : "Smoothstep() is useful in cases where a threshold function with a smooth transition is desired" here as well?

You can probably rephrase it by yourself, then it would be good to merge

nbogie:
A _very_ minor suggestion:
How about offsetting the transition position away from the centre to better show the control smoothstep gives, 
e.g. 
`let t = smoothstep(0.25, 0.35, x);`

(Otherwise, there's still the small risk that it looks to a newcomer to smoothstep that the function just transitions between values and they're not so aware they can choose _when_ it transitions!)

jjnawaaz:
Thanks both for the thoughtful feedback!

I’ve made a small follow-up update to address the points raised:

Added a short line in the description clarifying that smoothstep() is especially useful when a smooth threshold or cutoff is desired (per @perminder-17 suggestion).

Adjusted the first example so the transition region is offset away from the center, to better highlight that edge0 / edge1 control where the transition happens (per @nbogie 's note).

Kept the first example as minimal as possible (no uniforms, direct use of the result), with a second example that introduces animation separately.

Both examples have been tested in the p5.js web editor and should run as-is.

Thanks again - happy to make any final tweaks if needed!

src/strands/p5.strands.js
jjnawaaz: Thanks for the suggestion!

I’ve added a line to the description explaining that smoothstep() is useful when a threshold or cutoff is needed with a smooth, gradual transition instead of a hard edge, and rephrased it to fit the surrounding docs.

Please let me know if the wording looks good to you - happy to tweak it further, but otherwise this should be ready to merge.

perminder-17 (approved):
Looks good to me!

Issue #878, opened 2015-08-29
offline interactive reference


futuremarc:
So I'm using FileSaver.js to grab the generated reference page as an .xhtml file and using jszip.js to wrap up all the corresponding dependencies. This gives us a wonderful downloadable interactive reference. 

Any idea where to implement this feature within the reference?

lmccart:
ah this is so exciting and desperately needed! in terms of build, I think it would be amazing if it were somehow built into the docs generation process, so each time the docs are updated, theres a new static version is created as well. https://github.com/processing/p5.js/blob/master/Gruntfile.js#L107 I'm not sure offhand the best way to do this though.

in terms of placement on the website, maybe we add a link at the bottom of the [reference](http://p5js.org/reference/) page to the full download zip?

shiffman:
It would be lovely to package the reference with the p5.js editor and implement a "find in reference" option a la Processing (select a function name --> right-click to "find in reference").  @antiboredom, reasonable to implement?

peteruithoven:
Could a inclusion into devdocs.io be useful? This could give you the offline support. 
I have suggested adding p5.js on: https://trello.com/c/PVnfdeaN/26-suggest-new-docs-here

lmccart:
hey @futuremarc I think it's really important to finish this one up since it could be a significant hurdle for those with slow internet connections. could you give me an update on where you're at and i can take it from there? i know you got pretty far and i don't want to duplicate your effort. thanks!

toolness:
I'm excited about supporting this feature too, and would love to help out if needed.

futuremarc:
update on this: so without changing how our backend works I was able to zip the libraries and dependencies along with the html dump of the reference, but no linked pages would be saved. If we were using npm theres a package called wget that would do exactly what we need.  

My next step is to look into building it with the grunt build process. However this doesnt address the issue of how slow the Reference loads at p5js.org; we should be thinking about a different way to generate the docs or having them generated beforehand as static files.

Anyone have any other advice or thoughts on changing the way we build the reference?

lmccart:
thanks for looking into this @futuremarc. I agree, we should look into building the reference as static files. i'm going to close this issue, and we can continue discussion here: https://github.com/processing/p5.js-website/issues/143

PR #8462, opened 2026-01-28
Update contributing_to_the_p5js_reference for p5 v2

### LINKED ISSUE #8454: [p5.js 2.0 Bug Report]: guide "contributing to the p5js reference" is out of date for v2.x

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.2.0

### Web browser and version

n/a

### Operating system

n/a

### Steps:

1. visit https://beta.p5js.org/contribute/contributing_to_the_p5js_reference/
2. search for JSDoc and YUIDoc.  It talks about YUIDoc which is not used in v2.x
3. view [these v1.x instructions](https://beta.p5js.org/contribute/contributing_to_the_p5js_reference/#additional-info-constants) for typing function params which take constants
4. view section ["generating and previewing the reference"](https://beta.p5js.org/contribute/contributing_to_the_p5js_reference/#generating-and-previewing-the-reference) which refers to npm scripts non-existent in v2.x

I'd like to work on doing a full update of this document for v2.x

Update: This should include an update of https://beta.p5js.org/contribute/documentation_style_guide/  for 2.0, too.

nbogie:
Resolves #8454

## Changes in summary: 
1. update inline reference guide - contributor_docs/contributing_to_the_p5js_reference.md
  * correct for v2.0
  * cover some missing basics
  * add lots more examples
  * add a quick tag reference lookup
  * re-write and reformat some content for readability
2. update style guide for v2.0 correctness - /contributor_docs/documentation_style_guide.md

### regarding main document: /contributor_docs/contributing_to_the_p5js_reference, 

* comprehensive overhaul for v2.0 correctness, detail, and (hopefully) readability, as follows:

* update first sections
    * add a high-level ToC
    * improve first para
    * add an abstracted overview of a comment block

* update description section
    * add importance of first sentence of desc, inc many examples
    * add how to link to other function/variable/class docs
    * examples of markdown, link to markdown syntax guide

* update the params section
  * add space
  * constants different in v2
  * added documenting callback fns
  * added documenting arrays
  * specifying callback functions
  * on patching types in emergency (link to reference-generation-process.md)
  
* update return section
    * add space
    * improve return type discussion
    * add examples of returning promise, or array.

* on code examples:
    * wrote on complexity of examples, and self-contained examples
    * on providing multiple examples
    * mention optional text between examples (needs checked)
    * Remove `<div><code>` wrapper instructions, remove `notest`, update `norender` labelling.
    * Replaced partial examples with complete ones, and downplayed the ability to have partial examples auto-wrapped in a setup function.
    

* on assets:
    * document use of assets in description and in code examples
    * linked to there from the style-guide's coverage of same

 * document case: instance and static methods have same name
 * on `@module` and `@submodule` - clarified this is a documentation concept, gave examples
 * document `@beta` and `@deprecated`

* on how to generate and preview the reference:
    * rewrote the section on generating and previewing the reference
    * has some repetition now with [/docs/scripts.mjs](mostly pointing at https://github.com/processing/p5.js-website/blob/2.0/docs/scripts.md#testing-the-docs-of-your-fork)
   * add section on limitations of the preview website (no search, broken links)
   
* add section on linting doc comments

* on `@class`: add dave's section on creating and documenting classes, from his "jsdoc best practices"

* on the formatting of the guide itself:
  * add an example max line width for code examples
  * space out method, param, return (allowing examples for each)
  * add more sub-headings and fix level of some headings
  * mark all legal jsdoc comment blocks as js for markdown
  * reduce width of prose descriptions in example comment blocks for readability, hard-wrapping them to fit.  (They become very hard to read if they get wrapped.)
  * break some other long paras for readability.

* on the sin() example used throughout
    * update the sin() doc comment example
    * add link to sin() source in repo

* add section: tag quick reference table
  * currently an appendix
  * each tag links to its main section in the page
  * there's an argument this could go first (or be a separate document).
 
* add various hyperlinks
    * link to the relevant official jsdoc tag docs as they come up (@method, @param, @return, etc)
    * link to our documentation style guide
    * link to more entries in the reference
    * link to p5 source code in some cases

* v1-v2 differences
  * add an initial aside about p5 versions and this document
  * Add section on yuidoc being replaced with documentation.js and yuidoc
  * covered assets, hotlinking, other v1-v2 differences
  * also add param constants to v1-v2 differences appendix
  * add an appendix of differences for documenting v1 and v2

### regarding document: style guide
* finished
* Replaced yuidoc mention with documentation.js + jsdoc
* linked to the inline reference guide for generation and preview instructions
* rewrote section on assets (and again linked to inline reference guide's coverage of same)
* some external links (e.g. to wikipedia) remain broken and can't be fixed here - [#878](https://github.com/processing/p5.js-website/issues/878).  (I have a fix if wanted)
* (links to specific sections of inline reference guide are necessarily fudged to work in prod. `foo.md#section` becomes `foo/#section`)
    
### regarding document: jsdoc best practices
   * incorporated these details into the main guide
   * jsdoc.md is now empty except for a link to the main guide

#### PR Checklist

[x] checked for bad links
[x] build:contributor-docs runs
[x] spelling check - (american english)
[ ] grammar check


contributor_docs/contributing_to_the_p5js_reference.md
```diff
@@ -25,65 +54,233 @@ When you look at the source code of p5.js, you will see many lines in the librar
...
+#### <a id="linking"></a>Linking to another function or variable
 
-Anything in a block in this manner will be interpreted as reference documentation. You may be familiar with this style of code comments through [JSDoc](https://jsdoc.app/). While p5.js does not use JSDoc, it uses a very similar tool called [YUIDoc](https://yui.github.io/yuidoc/), which has a very similar reference syntax. In this style of reference comments, each comment block is further divided into individual elements, which we will have a look at next.
+A hyperlink HTML element - with the `a` tag) can be used within a function description to link to the documentation of a relevant function or variable.
 
+The url should be of this "fragment" or "hash" form: `#/p5/ + function_or_variable_name`, e.g. `#/p5/circle`, `#/p5/DEGREES`. 
+   It should _not_ include a direct link to the website.
 
-## Reference comments block
+<!-- TODO: correct this rule.  what about p5.Vector, p5.Graphics, etc.  what about methods of those classes ? p5.Vector/setMag?  static methods? -->
```
davepagurek: The syntax is generally `#/class/prop`, where the class for globals is just `p5`. 

We made an exception for constants because I think there was a naming clash between `blend()` and `BLEND` (the deployment doesnt support case dependent urls), so we put constants in `p5/constant/name-of-constant`

davepagurek: I believe static methods are treated like regular methods in that they dont have a different URL scheme. so if you have both an instance and static method with the same name, you'd just document one of them currently, and mention the existence of the static pattern in the prose of the documentation

contributor_docs/contributing_to_the_p5js_reference.md
```diff
@@ -93,44 +290,174 @@ At the very top of the comment is the text description of the function. This des
...
 ```
 @param {type} name Description here.
 ```
 
+For example, the `sin()` function has: 
+```
+@param  {Number} angle  the angle.
+```
+
+1. Following the keyword `@param`, the _type_ of the parameter is mentioned in curly brackets `{}`. (in this case, `{Number}`)
```
davepagurek, line 301: Possibly worth mentioning that this is encouraged but not required, so we don't use it yet for p5.strands where the types are weird

contributor_docs/contributing_to_the_p5js_reference.md
```diff
@@ -93,44 +290,174 @@ At the very top of the comment is the text description of the function. This des
...
+instead of:
+```
+@param {Array<Number>} colors an array of colors.
+```
+
+**"Number" not "number".  "String" not "string".**
+
+If you happen to be familiar with TypeScript, you might be more used to using lowercase `number` or `string` instead of the capitalized versions you see in the examples above.  Throughout the p5.js codebase the primitive types are all capitalized.
+
+<!-- TODO: reason here?  historic with yuidoc?  closure compiler? -->
```
davepagurek: Historic yuidoc (see examples on https://yui.github.io/yuidoc/), and also the choice to not care about the distinction between `string` and `String` like TS does, since instances of those classes and primitives are technically different (see https://stackoverflow.com/questions/203739/why-does-instanceof-return-false-for-some-literals). I think the simplification of "strings are instances of the String class" is the right abstraction level for p5 readers anyway!

contributor_docs/contributing_to_the_p5js_reference.md
```diff
@@ -269,15 +643,67 @@ If you do not want the example to be run as part of the automated tests (for exa
 * }
 * </code></div>
 ```
+### Requiring browser features for example code
+
+In some cases, the example code may be intended to demonstrate  p5.js functionality that depends on a feature that is not universally available on all browsers or devices (most notably, WebGL).  Rather than having this code run and break on incompatible browsers, use the "modernizr" attribute along with the name of the required feature.  At runtime, if that feature is not detected on the user's browser, the code will not be executed.  ([Features are listed here](https://modernizr.com/docs/#features).)
```
davepagurek: I'm not sure we use these any more, I think this probably could be removed

davepagurek: oh also looks like we're removing the wrapping divs in https://github.com/processing/p5.js/pull/8360

contributor_docs/contributing_to_the_p5js_reference.md
```diff
@@ -329,16 +760,42 @@ You can use the `@private` if a property or variable is a private function or va
...
+The ["Color" module](https://beta.p5js.org/reference/#Color) in `src/color` has the following submodules: 
+ "Creating & Reading", "Setting", and "Color Conversion".
+
+The _conceptual_ ["3D" module](https://beta.p5js.org/reference/#3D) documents code  mostly from `src/webgl` (there is no `src/3D`), and has the following submodules: "Camera", "Interaction", "Lights", "Material", "strands", "p5.Camera", "p5.Shader".
+
+#### The `@for` tag
 
 The `@for` tag defines the relationship between this module and the overall `p5` class, effectively saying this module is a part of the `p5` class.
 
+<!-- TODO: clarify the nature of this relationship documented by the @for tag.  Where might this be used - still only in docs?  also in type-gen? elsewhere? -->
```
davepagurek: Sometimes I think we add it when documentation.js inference fails. The main one I think is in properties of a class that we want to document -- since it doesnt correctly parse when we put the comment block in the class itself, we have to document it outside, and then we need to use `@for` to attach the docs to the right class.

contributor_docs/contributing_to_the_p5js_reference.md
```diff
@@ -329,16 +760,42 @@ You can use the `@private` if a property or variable is a private function or va
...
+
+#### The @requires tag
+
 The `@requires` tag defines the required imported modules that the current module depends on.
 
-```
+<!-- TODO: clarify where this @requires info is made use of.  What responsibilities does the author have to state these correctly and comprehensively? -->
```
davepagurek: I'm also not sure we really need this, as we use `import` now rather than assuming other files will be loaded.

contributor_docs/contributing_to_the_p5js_reference.md
```diff
@@ -348,69 +805,188 @@ The `@requires` tag defines the required imported modules that the current modul
...
+npm run custom:dev https://github.com/yourUsername/p5.js.git#yourBranch
+```
 
-- The main command to generate the reference from the reference comments in the source code is to run the following command.
+#### Limitations
 
+The website won't be _fully_ functional when partially prepared in this way.  Notably:
+
+* links between pages may be broken
+  * You'll need to manually add a trailing slash '/'
```
davepagurek: tbh we should probably fix that in the website code when we encounter it and not rely on auto adding of the `/` on deployments, so we can encourage people to file and fix these if they see them!

davepagurek: (we dont make non-trailing-slash urls work on dev because they dont exist on deploys, and it means relative links work differently than when there's a trailing slash, which means something can work locally but break on the live site.)

contributor_docs/contributing_to_the_p5js_reference.md
nbogie: Thanks for the spot!  I've updated to...
* remove the `<div><code>` wrappers and instructions (ensuring an `@example` tag for each)
* remove notest  and modernizr
* reflect the new way to do norender -  // META:norender

contributor_docs/contributing_to_the_p5js_reference.md
nbogie: Sorry, I'm not clear on this.  Do you mean all links in the reference doc comments should be changed to include a trailing slash (in the source code)?  Like this:
```
* <a href="#/p5.Image/">p5.Image</a>
```

(So far, I think there are zero instances of that in the codebase.)

(low-priority - even if this is so, I'm guessing it doesn't have to make this first v2 overhaul of the guide.)

contributor_docs/contributing_to_the_p5js_reference.md
nbogie: updated with those linking rules, thanks.

contributor_docs/contributing_to_the_p5js_reference.md
nbogie: Ty.  On reflection, I simplified to just state the convention, thinking a historic note might just be noise.

contributor_docs/contributing_to_the_p5js_reference.md
ksen0: @nbogie was this clarified at some other point? AFAIK we don't want to add trailing slashes in the code. But rather testers tend to need to add a trailing slash to pages, and this is probably something we should fix in the future, but for now it's a peculiarity of the dev environment

nbogie: Hi, @ksen0.  Thanks! No, I didn't get clarity on what was meant here.  As you can see, I went with the following as link format with lots of examples:

```
#/class/property
```

Example URLs linking to functions and variables:

```
#/p5/circle
#/p5/createCanvas
#/p5/fill
#/p5/mouseX
#/p5.Vector/lerp
#/p5.Vector/dist
#/p5.Image/width
```

Example URLs linking to classes:

```
#/p5.Color
#/p5.Image
#/p5.Vector
```

davepagurek: oops sorry for missing this -- yeah we don't need to change how we link to thinks in the reference, what you've got is good! ideally we update all the spots *in the website repo* where there are hardcoded links without trailing slashes (often in sidebars), and we already have some code that auto-appends trailing slashes for in-content links like this.

Ideally we wouldn't need a note in the docs about having to manually add a trailing slash in your browser when a link takes you to a broken page, but it is something important to note right now while we haven't fixed all of those occurrences.

contributor_docs/contributing_to_the_p5js_reference.md
ksen0: If unused, I'm ok with this being removed

nbogie: Hi @limzykenneth!  Are you aware of any need to keep the @requires tags in the  JSDoc comments for the v2 codebase?  (I'm documenting the various tags).

ksen0:
Thanks for all the hard work on this @nbogie !

contributor_docs/contributing_to_the_p5js_reference.md
limzykenneth: The tag specifies module dependencies, it doesn't really do anything for us since we don't author or publish the library in this kind of required dependency way so safe to remove.

nbogie: Thanks all.  Raised as #8628 (which I'll be happy to do or support)

PR #8473, opened 2026-01-31
Fix friendly error handling in color APIs to prevent runtime crashes

### LINKED ISSUE #8458: [p5.js 2.0 Bug Report]: Friendly errors do not prevent runtime exceptions in color APIs

- [ ] Accessibility
- [x] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [x] Friendly errors

### p5.js version

2.2.0

### Web browser and version

Google Chrome Version 144.0.7559.61

### Operating system

Windows 11

### Steps to reproduce this

**Description:**
While testing p5.js 2.0 color-related functions, I noticed that several APIs emit a Friendly Error but still throw runtime exceptions (e.g. TypeError or uncaught Error), causing sketches to crash.

In p5.js, a Friendly Error is generally expected to safely handle invalid input without crashing execution. However, this behaviour is inconsistent across color APIs.

### Steps:
1. Run the sketch given below  
2. Observe that it throws an FES error but still sketch crashes 

### Snippet:

```js

function setup() {
  noCanvas();

  let c1 = color(255, 0, 0);
  let c2 = color(0, 0, 255);

  // Case 1: lerpColor crashes after friendly error
  console.log(lerpColor(null, c2, 0.5));

  // Case 2: invalid color string crashes
  background("");
}


```

**Observed Behaviour**

1. Friendly Error is logged
2. Sketch then crashes with:

- TypeError: Cannot read properties of null
- or Error: Invalid color string

**Expected Behaviour**

_Friendly Error should:_
1. clearly report invalid input
2. prevent runtime exceptions
3. allow sketch execution to continue safely

**Additional Notes**
This pattern occurs in multiple color-related APIs (background, fill, stroke, color, alpha, lerpColor). Some functions correctly handle invalid input after emitting a Friendly Error, while others still throw.

_I'm willing to implement a fix if this gets approved_

Satish11012007:
Resolves #8458

While testing color-related APIs in p5.js 2.x, I noticed that several functions emit a Friendly Error but still throw runtime exceptions (e.g. TypeError, uncaught Error), causing sketches to crash.
image
image

Friendly Errors are generally expected to report invalid input and fail gracefully without terminating execution, but this behavior was inconsistent across color APIs.

This PR improves error handling in
 color “creating & reading” functions to ensure that after a Friendly Error is logged, execution exits safely instead of throwing.

### DIFF
--- a/src/color/creating_reading.js
+++ b/src/color/creating_reading.js
@@ -111,8 +111,12 @@ import '../core/friendly_errors/fes_core';
  */
 p5.prototype.alpha = function(c) {
   p5._validateParameters('alpha', arguments);
-  return this.color(c)._getAlpha();
+  if (!(c instanceof p5.Color)) {
+    c = this.color(c);
+  }
+  return c._getAlpha();
 };
+
 
 /**
  * Gets the blue value of a color.
@@ -1008,11 +1012,8 @@ p5.prototype.hue = function(c) {
 p5.prototype.lerpColor = function(c1, c2, amt) {
   p5._validateParameters('lerpColor', arguments);
 
-  if (!(c1 instanceof p5.Color)) {
-    c1 = color(c1);
-  }
-  if (!(c2 instanceof p5.Color)) {
-    c2 = color(c2);
+  if (!(c1 instanceof p5.Color) || !(c2 instanceof p5.Color)) {
+    return;
   }
 
   const mode = this._colorMode;


limzykenneth:
The issue is related to 2.x, so any changes should be filed against the `dev-2.0` branch. The issue also has not yet been approved for fix with a proposed solution, so you shouldn't file a PR before that. Thanks

PR #8494, opened 2026-02-04
Improve WebGL font error message to suggest textFont() usage

yugalkaushik:
Improves the error message when rendering text in WebGL mode without properly setting a font.

**Problem:**
The current error "only Opentype (.otf) and Truetype (.ttf) fonts are supported" is misleading when users have loaded a valid font but forgot to call textFont().

**Solution:**
Added clarification: "Make sure to set the font using textFont() before drawing text."

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/text.js
+++ b/src/webgl/text.js
@@ -654,7 +654,7 @@ p5.RendererGL.prototype._renderText = function(p, line, x, y, maxY) {
 
   if (!this._isOpenType()) {
     console.log(
-      'WEBGL: only Opentype (.otf) and Truetype (.ttf) fonts with glyph data are supported. Make sure to set the font using textFont() before drawing text.'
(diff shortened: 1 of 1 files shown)


ksen0:
If the test runs, then this sounds reasonable and helpful! I checked the failing test and I'm not sure what was up, I reran. If it fails, then code would need to be updated so tests do pass.

Minor wording revision to keep the message same length:

`WEBGL: requires Opentype (.otf) or Truetype (.ttf) font, loaded using textFont() before drawing text.`

cc @davepagurek @perminder-17

yugalkaushik:
@ksen0 all tests are passing now.

davepagurek (approved):
Thanks! Would you be interested in making a similar update to the `dev-2.0` branch, which p5 2.x builds are made from?

PR #8504, opened 2026-02-07
fix: use `@chainable` per-overload

### LINKED ISSUE #8356: [p5.js 2.0 Bug Report] (TypeScript): `@chainable` methods can't have overrides with different return types

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Build process
- [ ] Unit testing
- [ ] Internationalization
- [x] Friendly errors

### p5.js version

2.1.2

### Web browser and version

_No response_

### Operating system

_No response_

### `Element.size()` return value is incorrect

The generated types from the [element `size` method](https://github.com/processing/p5.js/blob/dev-2.0/src/dom/p5.Element.js#L1012) look like this:
```ts
  size(): Element;
  size(w?: number | typeof p5.AUTO, h?: number | typeof p5.AUTO): Element;
```
but really they should be
```ts
  size(): { width: number; height: number }; // (or `Object`)
  size(w?: number | typeof p5.AUTO, h?: number | typeof p5.AUTO): Element;
```

### `@chainable` code

I haven't narrowed down the _exact_ problem, but when `@return {Element}` is used directly instead of `@chainable` the type compilation is more correct.

The problem source is likely [this line](https://github.com/processing/p5.js/blob/dev-2.0/utils/typescript.mjs#L507), where`chainable` is applied per-method instead of per-overload.

### Proposed solutions

**Quick fix**: Use `@return {Element}` instead of `@chainable` when there's an override

**Right fix**: Figure out how to get `@chainable` working per-overload instead of per-method

LuLaValva:
- Resolves #8356 

 Changes:

- `@chainable` in JSDoc applies to _one overload_ and not the whole function
- To test, view updates to `p5.d.ts` after `npm run generate-types`

code screenshot showing that the type signature for a size method changed from returning Element for both overloads to object and element in the correct places

Updated methods: `volume`, `speed`, `position`, `size`, `style`, `attribute`, `value`

#### PR Checklist

I don't think any of these apply 😅... `lint` has a bunch of warnings but none related to my code, and I don't see unit/snapshot tests for the `p5.d.ts` file.

- [ ] `npm run lint` passes

### DIFF
--- a/src/webgl/p5.Shader.js
+++ b/src/webgl/p5.Shader.js
@@ -6,6 +6,7 @@
  * @requires core
  */
 
+const TypedArray = Object.getPrototypeOf(Uint8Array);
 class Shader {
   constructor(renderer, vertSrc, fragSrc, options = {}) {
     this._renderer = renderer;
@@ -1003,17 +1004,13 @@ class Shader {
    *   plane(100, 100);
    * }
    */
-  setUniform(uniformName, rawData) {
+  setUniform(uniformName, data) {
     this.init();
 
     const uniform = this.uniforms[uniformName];
     if (!uniform) {
       return;
     }
-
-    const data = this._renderer._mapUniformData
-      ? this._renderer._mapUniformData(uniform, rawData)
-      : rawData;
 
     if (uniform.isArray) {
       if (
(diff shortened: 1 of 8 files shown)


davepagurek:
@all-contributors please add @LuLaValva for bug, code

PR #8505, opened 2026-02-08
Fix Show actual type in strands hook error messages

### LINKED ISSUE #8444: [p5.strands] Better error messages when you return the wrong type in a hook
### Increasing access

If I call `getWorldInputs(() => [1, 2, 3, 4])`, returning a vec4 instead of an inputs struct, I get: `You have returned a undefined from getWorldInputs when a Vertex was expected.`

I did not return undefined, I returned a vec4! So that could already be clearer. Also although the data type expected here is `Vertex`, this is not really something that is understandable to users. They should either be modifying and returning the inputs that are passed in to the function, or as of the next release, *just* modifying the inputs and returning nothing at all for cases where a struct is passed in as a param and the same struct is expected to be returned.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

In other cases where a different return is expected, the type name is still meaningless to users, we'd need to tell them what properties to expect on the object.

Anshumancanrock:
Resolves #8444

### Changes:

- Fixed bug: `retNode.baseType` → `returnedNode.baseType` (was causing "undefined" instead of actual type)
- Error messages now show the actual returned type (e.g., "vec4" instead of "undefined")
- Struct mismatch error now lists expected properties so users know what the struct contains
- Added guidance explaining users should modify and return the input struct
- Improved incomplete struct error to show expected vs received properties

#### PR Checklist

- [x] `npm run lint` passes


davepagurek (commented):
Thanks for taking this on! Would you be able to add some unit tests for this? A strategy could be to spy on `FES.userError` (search for `spy` in the tests directory for examples) and assert that it was called with the expected message.

Anshumancanrock:
@davepagurek Done! Added unit tests with `vi.mock` on `strands_FES` , covers all three error paths (wrong type in struct hook, vector dimension mismatch, incomplete struct).

Please take a look and let me know if you would like any changes. Thanks !

PR #8513, opened 2026-02-09
Improve WebGL font error message to suggest textFont() usage

yugalkaushik:
Improves the error message when rendering text in WebGL mode without properly setting a font.

Problem:
The current error "only Opentype (.otf) and Truetype (.ttf) fonts are supported" is misleading when users have loaded a valid font but forgot to call textFont().

Solution:
Added clarification: "Make sure to set the font using textFont() before drawing text."
#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/text.js
+++ b/src/webgl/text.js
@@ -685,7 +685,7 @@ function text(p5, fn) {
 
     if (!p5.Font.hasGlyphData(this.states.textFont)) {
       console.log(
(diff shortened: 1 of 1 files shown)


davepagurek (approved):

yugalkaushik:
@all-contributors please add @yugalkaushik  for bug, code

PR #8514, opened 2026-02-09
docs: add inline reference for uniform* methods in p5.strands

### LINKED ISSUE #8439: [p5.strands] Document `uniform*` methods
### Increasing access

These are used in other p5.strands docs for passing data in, but do not have reference pages themselves.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

These should be present in a new p5.strands section of the docs, and should include parameters and usage examples.

Anshumancanrock:
Resolves #8439

### Changes:

Added inline reference documentation for the `uniform*` methods in `p5.strands`. These methods are used in strands shader callbacks to pass data from JavaScript into shader code, but previously had no reference pages.

Documented methods:
- `uniformFloat()` - single float value (includes 2 examples: with callback + manual setUniform)
- `uniformInt()` - integer value
- `uniformBool()` - boolean toggle
- `uniformVec2()` - 2D vector
- `uniformVec3()` - 3D vector
- `uniformVec4()` - 4D vector (RGBA colors, etc.)
- `uniformTexture()` - texture sampling via `getTexture()`

Each entry includes a description, parameter/return info, usage examples, and notes about aliases (e.g. `uniformVector2()`, `uniformSampler2D()`).

- [x] `npm run lint` passes

### DIFF
--- a/src/strands/p5.strands.js
+++ b/src/strands/p5.strands.js
@@ -915,135 +915,69 @@ if (typeof p5 !== "undefined") {
 /**
  * Declares a float uniform variable for use in a p5.strands shader.
  *
- * Uniforms are values that are passed from JavaScript into shader code. They
- * stay the same for every pixel and vertex in a single draw call but can
- * change between frames. `uniformFloat()` creates a uniform that holds a
- * single number.
+ * Uniforms are values passed from JavaScript into shader code. They stay the
+ * same for each pixel and vertex during one draw call, but can change between
+ * frames.
  *
- * `uniformFloat()` can only be called inside a shader
- * <a href="#/p5.Shader/modify">`modify()`</a> callback or a build function
- * such as <a href="#/p5/buildMaterialShader">`buildMaterialShader()`</a>.
+ * `uniformFloat()` creates a uniform that holds one Number. It can only be
+ * called inside a shader <a href="#/p5.Shader/modify">`modify()`</a>
+ * callback or a build function such as
+ * <a href="#/p5/buildMaterialShader">`buildMaterialShader()`</a>.
  *
- * The variable name used in the shader is automatically taken from the
- * JavaScript variable name the result is assigned to.
+ * The most common usage is to pass a callback so the value is refreshed each
+ * frame:
  *
- * An optional callback function can be passed as an argument. If provided,
- * the shader will call it every frame to get an updated value. If no
- * callback is provided, the uniform must be set manually using
- * <a href="#/p5.Shader/setUniform">`setUniform()`</a>.
+ * ```js example
+ * let myShader;
+ * function setup() {
+ *   createCanvas(200, 200, WEBGL);
+ *   myShader = buildMaterialShader(material);
+ * }
+ *
+ * function material() {
+ *   let time = uniformFloat(() => millis() * 0.001);
+ *
+ *   worldInputs.begin();
+ *   worldInputs.position.y += 20 * sin(time + worldInputs.position.x * 0.05);
+ *   worldInputs.end();
+ * }
+ *
+ * function draw() {
+ *   background(255);
+ *   shader(myShader);
+ *   lights();
+ *   noStroke();
+ *   fill('red');
+ *   sphere(50);
+ *   describe('A red sphere that undulates over time.');
+ * }
+ * ```
+ *
+ * Advanced usage: pass a name to make the uniform stable even if your build
+ * process minifies variable names.
+ *
+ * ```js
+ * let time = uniformFloat('time');
+ * // later:
+ * myShader.setUniform('time', millis());
+ * ```
  *
  * @method uniformFloat
  * @beta
- *
- * @param {Function} [callback]
- *        a function that returns a Number. If provided, the uniform will
- *        automatically update each frame with the return value.
- *
- * @returns {Number}
- *          a shader value representing a float that can be used in shader
- *          calculations.
- *
- * @example
- * <div modernizr="webgl">
- * <code>
- * // Use uniformFloat with a callback for automatic updates
- *
- * let myShader;
- * function setup() {
- *   createCanvas(200, 200, WEBGL);
- *   myShader = buildMaterialShader(material);
- * }
- *
- * function material() {
- *   // The uniform updates automatically each frame
- *   const time = uniformFloat(() => millis() * 0.001);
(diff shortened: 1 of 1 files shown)


davepagurek (commented):
Thanks for taking this on, looking good so far! Some notes:
- In the examples, can we use the new flatter API for hooks where instead of doing `getWorldInputs((inputs) => { ... })` we use `worldInputs.begin()` and `worldInputs.end()`? The examples on the beta reference site do this now too.
- For each of these, we can optionally also add a string as the parameter to manually name the uniform. So if you do `uniformFloat('something')`, then you'd later do `myShader.setUniform('something', val)`. If omitted, you use the name of the variable, but this can be useful for environments where your code is getting minified and the variable names might change. Definitely advanced usage, so we probably want to only mention / show examples of this after the simpler and more common cases. The way to do that is to use **inline examples** rather than `@example` blocks at the end. I explained a bit about that here https://github.com/processing/p5.js/issues/8454#issuecomment-3865343780 and you can also look at the linked PR for that issue to see some WIP updated docs on how to use it.
- Do you think I could get you to provide screenshots of what the examples look like in your PR or a comment?

Anshumancanrock:
Hey @davepagurek, thanks for the review! I've made all the changes you suggested.

- Switched to the flat hooks API.
- Added the optional string naming param as a separate advanced example after the simple one.
- Using inline ```js example blocks instead of @example

Here are some screenshots from the website:

image

image


Please let me know if there's anything else!

src/strands/p5.strands.js
```diff
@@ -911,3 +911,400 @@ if (typeof p5 !== "undefined") {
...
+ *
+ * ```js
+ * let time = uniformFloat('time');
+ * // later:
+ * myShader.setUniform('time', millis());
+ * ```
+ *
+ * @method uniformFloat
+ * @beta
+ * @param {(String|Function)} [nameOrCallback] optional uniform name, or a
```
davepagurek, line 967: Rather than packing these into one method, these should likely be two different overloads. See how p5.Image does that by having two doc comment blocks for the two overloads: https://github.com/processing/p5.js/blob/v2.2.2/src/image/loading_displaying.js#L1020

src/strands/p5.strands.js
```diff
@@ -911,3 +911,400 @@ if (typeof p5 !== "undefined") {
...
+ *   shader(myShader);
+ *   lights();
+ *   noStroke();
+ *   fill('red');
+ *   sphere(50);
+ *   describe('A red sphere that undulates over time.');
+ * }
+ * ```
+ *
+ * Advanced usage: pass a name to make the uniform stable even if your build
```
davepagurek, line 956: Instead of just calling it advanced usage, we should maybe explain more why you would want to do this. The main use case is for using p5 in a minified project, where your variable names aren't guaranteed to stay the same.

src/strands/p5.strands.js
```diff
@@ -911,3 +911,400 @@ if (typeof p5 !== "undefined") {
...
+ * <a href="#/p5.Graphics">`p5.Graphics`</a>, or
+ * <a href="#/p5.Framebuffer">`p5.Framebuffer`</a>. You can sample it with
+ * <a href="#/p5/getTexture">`getTexture()`</a>.
+ *
+ * ```js example
+ * let myShader;
+ * let pg;
+ * function setup() {
+ *   createCanvas(200, 200, WEBGL);
+ *   pg = createGraphics(200, 200);
```
davepagurek, line 1267: Best practice in WebGL is to use framebuffers over p5.Graphics because performance will be significantly better, can we switch to using a framebuffer here?

src/strands/p5.strands.js
```diff
@@ -911,3 +911,400 @@ if (typeof p5 !== "undefined") {
...
+ *
+ * `uniformFloat()` creates a uniform that holds one Number. It can only be
+ * called inside a shader <a href="#/p5.Shader/modify">`modify()`</a>
+ * callback or a build function such as
+ * <a href="#/p5/buildMaterialShader">`buildMaterialShader()`</a>.
+ *
+ * The most common usage is to pass a callback so the value is refreshed each
+ * frame:
+ *
+ * ```js example
```
davepagurek, line 930: For each of these, we may want to copy-and-paste each the example twice: the first time, without using any parameter (`uniformFloat()` + `setUniform` in `draw`), and then after that, an explanation of how you can pass a callback to automatically pass in a value each frame.

src/strands/p5.strands.js
```diff
@@ -911,3 +911,400 @@ if (typeof p5 !== "undefined") {
...
+ * let myShader;
+ * function setup() {
+ *   createCanvas(200, 200, WEBGL);
+ *   myShader = buildMaterialShader(material);
+ * }
+ *
+ * function material() {
+ *   let time = uniformFloat(() => millis() * 0.001);
```
davepagurek, line 938: We now have the ability to just use `millis()` without passing it in as a uniform. Maybe now the reason you'd want to use one of these could be to pass in the value of a `createSlider()`. Also just a note, those initially go below the canvas and the p5 website doesnt automatically make room for them, so we may want to manually place it over the canvas with `.position(0, 0)`.

src/strands/p5.strands.js
```diff
@@ -911,3 +911,400 @@ if (typeof p5 !== "undefined") {
...
+ *
+ * ```js example
+ * let myShader;
+ * function setup() {
+ *   createCanvas(200, 200, WEBGL);
+ *   myShader = buildMaterialShader(material);
+ * }
+ *
+ * function material() {
+ *   let waves = uniformInt(() => 5);
```
davepagurek, line 988: Similar to before, should we use a slider here to make it more obvious why you'd need to make it a uniform at all?

davepagurek (commented):
Sorry for the delay on this, thanks for the updates! I've left some more feedback.

perminder-17 (commented):
Hi @Anshumancanrock, thanks for working on this. I noticed there are still some review comments from @davepagurek pending. Are you planning to work on those? If possible, could you please address them?

If not, I can help finish this PR.

PR #8523, opened 2026-02-11
Fix build*Shader methods in instance mode by forwarding optional scope

### LINKED ISSUE #8520: [p5.js 2.0 Bug Report]: build*Shader methods don't work yet in instance mode

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.2.1

### Web browser and version

All

### Operating system

All

### Steps:

To use `.modify()` in instance mode, you have to call it like this:

```js
baseMaterialShader.modify(({ p5 }) => {
  p5.getWorldInputs((inputs) => {
    inputs.position.x += 5
    return inputs
  })
}, { p5 })
```

Currently, the new `build*Shader` methods don't pass along the second context object:
https://github.com/processing/p5.js/blob/276a96aa9ecb8fc2c0d3005e1936042dabb80758/src/webgl/material.js#L708-L710

It needs to do so to be able to support that syntax, e.g.:
```js
buildMaterialShader(({ p5 }) => {
  p5.getWorldInputs((inputs) => {
    inputs.position.x += 5
    return inputs
  })
}, { p5 })
```

aashu2006:
Resolves #8520

## Changes

The five build*Shader methods in `src/webgl/material.js` were calling `.modify(callback)` without forwarding the optional `scope` parameter.

In instance mode, users pass { p5 } or { sketch } as scope so the callback receives the correct instance. Since this argument was not forwarded, the callback didn’t receive the expected context.

This PR updates these methods to accept an optional `scope` parameter and pass it to `.modify(callback, scope)`

No changes are made to .modify() itself, and existing code without scope continues to work.

I have also added a unit test under test/unit/webgl/p5.Shader.js (inside the existing p5.strands suite) to verify that buildMaterialShader correctly forwards the optional scope to .modify(). JSDoc comments were updated to document the optional scope parameter.

```js
// Before
fn.buildFilterShader = function (callback) {
  return this.baseFilterShader().modify(callback);
};

// After
fn.buildFilterShader = function (callback, scope) {
  return this.baseFilterShader().modify(callback, scope);
};
```

#### PR Checklist

- [x] `npm run lint` passes


davepagurek:
Thanks for taking this on! It's looking good so far, but can we add a unit test for one of these methods to make sure it continues to work going forward? Also, can we add the optional context parameter to the doc comments, mirroring what is found in the `p5.Shader.modify` doc comments?

aashu2006:
Hii @davepagurek I have added the unit test for scope forwarding and updated the JSDoc comments as suggested. All tests pass locally. Thanks ❤️

test/unit/webgl/p5.Shader.scope.js
```diff
@@ -0,0 +1,25 @@
+import p5 from '../../../src/app.js';
+
+suite('p5.Shader', function() {
```
davepagurek: Hi! The test looks good, is there a reason why this is in its own file as opposed to the existing p5.Shader unit tests file? It might be better in there.

Also, generally the tests in that file don't create a new instance, they reuse a `myp5` instance. To have less moving parts, they also omit a setup function, they just create a canvas, initialize whatever they need to, and then make their assertions. This means you wouldn't need a `done` function.

test/unit/webgl/p5.Shader.scope.js
```diff
@@ -0,0 +1,25 @@
...
+    test('forwards scope to modify in instance mode', function(done) {
+      new p5((sketch) => {
+        sketch.setup = function() {
+          sketch.createCanvas(10, 10, sketch.WEBGL);
+
+          const scope = { sketch };
+          let receivedSketch = null;
+
+          sketch.buildMaterialShader(function({ sketch: s }) {
+            receivedSketch = s;
```
davepagurek: p5.strands shaders don't have the ability to assign to variables outside of their functions -- since we transpile them, they lose all external scope, which is the reason why we have to manually pass scope in.

Rather than doing something like this in the test, we can maybe just use scope the way we use it in the `.modify` calls in the other tests, e.g. this call, before your changes, wouldn't work with `buildMaterialShader`: https://github.com/processing/p5.js/blob/3a7992c53eb422ec73469ec8e0eede960b639091/test/unit/webgl/p5.Shader.js#L555-L565

We can just wrap the creation of the shader in an `expect(() => { /* ... */ }).not.toThrow()` call maybe?

test/unit/webgl/p5.Shader.scope.js
aashu2006: Thanks, that's right! I will move the test into p5.Shader.js and update it to reuse myp5 like the other tests, without creating a new instance or using done.

test/unit/webgl/p5.Shader.scope.js
aashu2006: Thanks for explaining, I will update the test to follow the existing `.modify()` pattern and wrap the shader creation in an `expect(() => {}).not.toThrowError()` call instead of assigning to external variables.

aashu2006:
Hi @davepagurek . I have moved the scope test into p5.Shader.js under the existing p5.strands suite and aligned it with the surrounding test style. All tests have passed locally. Thanks 🙂

davepagurek (approved):
Awesome! thanks again!

aashu2006:
> Awesome! thanks again!

Thanks a lot for the review and guidance, I really appreciate the time you took to explain the test structure & strands behavior. Learned a lot from this one 🙂

PR #8530, opened 2026-02-15
Refactor global node handling and add millis function

### LINKED ISSUE #8474: Adding support for `millis()` in strands
### Increasing access

In p5.js, `millis()` returns the number of milliseconds since the sketch started running. This feature request proposes adding `millis()` support in p5.strands in the same way we currently handle functions like noise().

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### Feature request details

When p5.strands is active, p5.js functions such as `noise()` are routed to a strands-specific implementation. Similarly, while strands is active, calling `millis()` should return a strand node instead of an immediate numeric value, so it can be composed inside the strands graph.

perminder-17:
Resolves #8474 

 Changes:

Added support for millis(),

Here's the result. 
 https://editor.p5js.org/aman12345/sketches/YPHvvipq7

In the code side we are doing nothing but 

`getOrCreateUniformNode`: This function checks if a StrandsNode for a given uniform already exists. If it exists, it returns the cached node. If it doesn’t, it registers the uniform in `strandsContext.uniforms` (only once), builds the IR variable node, wraps it using `createStrandsNode` to make a StrandsNode, stores it in the cache, and returns it.

Using the same function as an helper for `getBuiltinGlobalNode` as we do same stuffs there for global property like mouseX etc. 

#### PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/src/strands/strands_api.js
+++ b/src/strands/strands_api.js
@@ -79,14 +79,14 @@ function getBuiltinGlobalNode(strandsContext, name) {
   if (!spec) return null;
 
   const uniformName = `_p5_global_${name}`;
+  const instance = strandsContext.renderer?._pInst || strandsContext.p5?.instance;
 
   const node = getOrCreateUniformNode(
     strandsContext,
     uniformName,
     spec.typeInfo,
     () => {
-      const p5Instance = strandsContext.renderer?._pInst || strandsContext.p5?.instance;
-      return p5Instance ? spec.get(p5Instance) : undefined;
+      return instance ? spec.get(instance) : undefined;
     }
   );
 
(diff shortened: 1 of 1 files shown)


perminder-17:
Adding reviewer: @Satish11012007 and @Anshumancanrock since you were interested in contributing. Thanks.

src/strands/strands_api.js
```diff
@@ -303,6 +317,21 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
...
+  fn.millis = function (...args) {
+    if (!strandsContext.active) {
+      return originalMillis.apply(this, args);
+    }
+    return getOrCreateUniformNode(
+      strandsContext,
+      '_p5_global_millis',
+      DataType.float1,
+      () => {
+        const p5Instance = strandsContext.renderer?._pInst || strandsContext.p5?.instance;
```
davepagurek: I know this isn't introduced in this PR, but I'm realizing there may be a bug here -- if you use p5.strands in multiple contexts, the global `strandsContext` may have been reset to a different context. Rather than trying to extract the instance here, could we maybe cache a reference to `this` outside of this callback (e.g. `const instance = this`) to capture the current instance and then use it here? And similarly for `getBuiltinGlobalNode`?

davepagurek (approved):
Looks good!

PR #8531, opened 2026-02-15
WebGPU compute shaders

davepagurek:
Resolves https://github.com/processing/p5.js/issues/7345 (for WebGPU mode only -- we can open a new issue later to see what, if anything, is possible to backport to WebGL)

### Changes

This adds the concept of a compute shader, which is a separate sort of thing to a regular shader, which has vertex/fragment components. Here are some differences:
- You run compute shaders via `compute(shader, x, [y], [z])` to specify how to run it
- `buildComputeShader` only has one hook, so you don't need to do any `.begin()` or `.end()` calls within it -- we automatically do that for you. Currently that one hook is called `iteration`, so you access per-iteration info like your index via `iteration` like you would access the inputs to any other hook.
  - You can still call `.modify()` on a compute shader, so if a library author makes a base compute shader with multiple hooks, you can still specify all the hooks separately with `modify`.
- You create storage that compute shaders can read/write with `createStorage()`. Those are always arrays of floats. You can pass in either the length, to start it off empty, or pass in an array of initial data.
- You can access storage within a shader by using `uniformStorage`.
- Within a shader, you can use `myStorageUniform[index]` to get or set contents of that storage.

To do:
- [x] Make `uniformStorage` work with default values -- currently you have to manually call `setUniform()`
- [x] Add a lot of tests around storage access and writes within shaders, I assume there are some bugs around branching/looping currently
- [x] Iterations on the API?
- [x] Write docs

### Screenshots of the change

https://github.com/user-attachments/assets/25eba8a5-6a92-4510-a8c2-f7a60c56b858

### Test sketch

https://editor.p5js.org/davepagurek/sketches/MVc84tjJw

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/webgl/p5.Shader.js
+++ b/src/webgl/p5.Shader.js
@@ -1099,6 +1099,10 @@ class Shader {
       return;
     }
 
+    // In p5.strands-related code, where some of the code may be in
+    // p5.webgpu.js instead of the main p5.js build, we generally use
+    // duck typing instead of instanceof to avoid accidentally importing
+    // and comparing against a separate copy of p5 classes
     if (data?.isVector) {
       data = data.values.length !== data.dimensions ? data.values.slice(0, data.dimensions) : data.values;
     } else if (data?.isColor) {
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -61,7 +61,11 @@ class Color {
   static #colorjsMaxes = {};
   static #grayscaleMap = {};
 
-  // For duck typing
+  // This property is here where duck typing (checking if obj.isColor) needs
+  // to be used over more standard type checking (obj instanceof Color). This
+  // needs to happen where we are building multiple files, such as in p5.webgpu.js,
+  // where if we `import { Color }` directly, it will be a separate copy of the
+  // Color class from the one imported in the main p5.js bundle.
   isColor = true;
 
   // Used to add additional color modes to p5.js
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -48,6 +48,12 @@ class Vector {
       this.dimensions = dimensions;
       this._values = values;
     }
+
+    // This property is here where duck typing (checking if obj.isVector) needs
+    // to be used over more standard type checking (obj instanceof Vector). This
+    // needs to happen where we are building multiple files, such as in p5.webgpu.js,
+    // where if we `import { Vector }` directly, it will be a separate copy of the
+    // Vector class from the one imported in the main p5.js bundle.
     this.isVector = true;
   }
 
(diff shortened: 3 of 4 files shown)


davepagurek:
Update: now you can create objects in your storage buffer and access their properties in your shaders! Demo: https://editor.p5js.org/davepagurek/sketches/y_e7f4JZ-

davepagurek:
Started adding some docs! One thing that I noticed is that docs for WebGPU things will need the WebGPU addon script included for the examples to run, and should probably include a warning saying so for users too. I've added a `@webgpu` tag you can add to JSDoc to indicate some reference examples will need the script, and `@webgpuOnly` indicates that the method itself is only available in the WebGPU addon, which displays a warning at the top of the reference page.

image

src/webgl/p5.Shader.js
```diff
@@ -1020,6 +1099,12 @@ class Shader {
       return;
     }
 
+    if (data?.isVector) {
```
ksen0, line 1106: Why not instanceof?

davepagurek: Because the WebGPU addon gets bundled as a separate file, if we do `instanceof SomeClass` within it, it will always fail, because it actually gets a whole separate copy of any class you import. It works ok if you do `instanceof p5.SomeClass` to avoid importing it directly, which duplicates the code, and it also works ok without that if you're editing a file that's exclusively used within the main repo. But if in the future we distribute core addons as separate files, they'll have that issue too. So I figured there are enough footguns that it might be easier to just not do `instanceof` across the board and use duck typing instead, as a pattern that will work everywhere (with the caveat that it could be a false positive if something else has `isVector: true`, but arguably that's a feature for extensibility?)

davepagurek: Open to other opinions though!

ksen0: I appreciate duck typing in general but I think it's unusual in the codebase (I could be wrong!) and if so could you add a very short inline comment here that it's intentional? It stands out

davepagurek:
Added comments everywhere duck typing is used, with a longer one in p5.Shader where it's not strictly necessary but just easy to get mixed up among the other p5.strands files, and similar explanation around the isColor/isVector properties in their respective classes.

src/webgl/p5.Shader.js
limzykenneth: Just to mention I plan to have most if not all of the type check transition to duck typing check instead of `instanceof` check. The reasons are partly mentioned by @davepagurek already but also to add that currently we have some problem with duplicates in the custom module build as some modules import the entire class from a different module when it only does an `instanceof` check which is not very efficient and not treeshakeable. Another reason is that using duck typing we allow more alternative usage of the interface we expose, as long as something has the right shape, it can be used rather than needing to be explicitly of a specific class instance, ie. addons can provide objects of matching shape to our interfaces rather than needing it to be specific classes.

Issue #8469, opened 2026-01-30
text() with maxWidth and maxHeight removes '\t' tabs


ffd8:
### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [x] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [x] Typography
- [ ] Utilities
- [ ] p5.strands
- [ ] WebGL
- [ ] DevOps, Build process, Unit testing
- [ ] Internationalization (i18n)
- [ ] Friendly Errors
- [ ] Other (specify if possible)

### p5.js version

1.11.11

### Web browser and version

Chrome 143.0.7499.193

### Operating system

MacOSX 13.6.5

### Steps:
1. Create `text()` with text containing tabs `\t`
2. Give textbox a maxWidth and maxHeight
3. Tabs are stripped from text

### Snippet:

```js

function setup() {
	createCanvas(windowWidth, windowHeight)
}

function draw() {
	background(0)
	fill(255)
	let txt = 'here is \n\tsome text'
	textSize(50)
	text(txt, 0, 0) // has tabs
	// text(txt, 0, 0, width, height) // removes tabs
}
```

rakesh2OO5:
I tested the exact snippet on main (1.11.12-rc.0) using text(txt, 0, 0, width, height). Tabs render correctly in both bounded and unbounded cases now.

This seems to be fixed since 1.11.11. I can try to locate the commit that changed this and add a regression test if that would be helpful.

Screenshot :

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ebb25fde-9986-4e49-8a57-1c7d5ed7a859" />

rakesh2OO5:
I tested the same snippet on Windows + Brave (Chromium) with p5.js 1.11.11 and couldn’t reproduce the issue — tabs render correctly in both bounded and unbounded cases.

This may be macOS- or browser-specific (original report was macOS + Chrome). It might be worth confirming with another macOS user or testing across Safari/Firefox on macOS.

Screenshot :

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/9cc65a87-b4b0-4beb-ba14-21d693237a34" />

rakesh2OO5:
Quick summary of my testing so far:

-macOS + Chrome (original report): bug reported in 1.11.11
[...]

PR #8533, opened 2026-02-16
Issue#8469

dhowe:
Fixes issue #8469 with tabs at start of line by only trimming from ends of line during lineation

### DIFF
--- a/src/type/textCore.js
+++ b/src/type/textCore.js
@@ -1790,6 +1790,11 @@ function textCore(p5, fn) {
     return { bounds, lines };
   };
 
+  Renderer.prototype._trimEnd = function (str) {
+    // trim trailing linebreaks and whitespace only
+    return str.replace(/[\s\n]+$/, '');
+  };
+
   /*
     Adjust width, height of bounds based on current rectMode
    * @private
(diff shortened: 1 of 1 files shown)


src/type/textCore.js
```diff
@@ -2247,13 +2252,13 @@ function textCore(p5, fn) {
         testLine = `${line + words[widx]}` + splitter;
         testWidth = this._textWidthSingle(testLine);
         if (line.length > 0 && testWidth > maxWidth) {
-          newLines.push(line.trim());
+          newLines.push(this._trimEnd(line));
```
davepagurek, line 2255: Would this have the same problem as before but for right aligned text? Maybe we have to trim dependent on the alignment?

dhowe: good point... though it is a question of both alignment and text-direction, no?

davepagurek: Ah yes I think you're right. Not sure what that means for mixed LTR and RTL text, like if you throw some English letters into an Arabic sentence... gets a little complicated knowing what to check

davepagurek: actually, taking a step back for a sec, how important is it that we trim the lines anyway? is just not trimming at all an option?

dhowe: Yes, I was just thinking the same thing...

This has to do with multi-line text that is broken based on width; so we break between words, and we don't want that space on which we broke to be visible (on either line), if we are following standard typographic conventions

but we could allow for the specific case of tabs

davepagurek: ah, got it, that makes sense. maybe we just need regexes to match leading/trailing space characters that are not the tab character then? also thankfully that would mean not having to handle RTL/LTR differences I think?

PR #8534, opened 2026-02-16
Add visual regression test for leading tabs in text

Omar-2718:
Addresses #8469

 Changes:
Added screen shots for the required result in \test\unit\visual\screenshots\Typography\textFont\Preserve leading tabs
and added regression test to test if the tap renders in a text with text('test\n\ttest', 0, 0, p5.width, p5.height)

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "p5",
-  "version": "2.2.2",
+  "version": "2.2.3-rc.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "p5",
-      "version": "2.2.2",
+      "version": "2.2.3-rc.0",
       "license": "LGPL-2.1",
       "dependencies": {
         "@davepagurek/bezier-path": "^0.0.7",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "test/**/*.js": "eslint",
     "utils/**/*.{js,mjs}": "eslint"
   },
-  "version": "2.2.2",
+  "version": "2.2.3-rc.0",
   "dependencies": {
     "@davepagurek/bezier-path": "^0.0.7",
     "@japont/unicode-range": "^1.0.0",


test/unit/visual/cases/typography.js
```diff
@@ -89,6 +89,13 @@ visualSuite('Typography', function () {
       p5.text('p5*js', -p5.width / 2, -p5.height / 2 + 10, p5.width);
       screenshot();
     });
+
+    visualTest('Preserve leading tabs',function (p5, screenshot){
+      p5.createCanvas(100, 100);
+      p5.textSize(35);
+      p5.text('test\n\ttest', 0, 0, p5.width, p5.height);
```
davepagurek: Hi! This wasn't generated with @dhowe's changes in, right? I'm wondering why the screenshot shows the tab having an effect, when it's the buggy code still generating the visual. @Omar-2718 in your browser this was still showing incorrect behaviour right? If you mirror that test more closely here (e.g. with a space before the `\n`? not sure if that's significant) do you see the screenshot starting to show the incorrect behaviour?

In general I'd expect to see the screenshot show *incorrect* results, and then only after @dhowe's code is merged, would we be able to lock in the correct behaviour in a screenshot.

Omar-2718: I used @dhowe's code to generate the screenshot (as that is the supposed correct answer for the test case) so that my version (the one that isnt working) fails that test. My bad,
So I'm supposed to push the screenshot of my failing code(it fails with 'test\n\ttest' ) before @dhowe's changes, right? And only after @dhowe's pull is merged, update the screenshot

davepagurek: ah ok, that makes sense then! although I'm also curious why CI says that the tests are paying even though this PR doesn't include the fixes yet... in any case, we should merge this after the fix is in.

Omar-2718: image
Yeah thats weird, this image is from the CI visual-test-report and somehow passes. When I run it locally, it fails normally. I'm not sure if this behaviour is because of font differences between Ubuntu on the CI and Windows that generated the test image. Or the difference threshold to make the test pass is too big. I can look into it

PR #8568, opened 2026-02-25
Fix filter() crash on createGraphics(WEBGL) by mirroring strands API …

### LINKED ISSUE #8549: filter() crashes on createGraphics(WEBGL) buffers

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] p5.strands
- [x] WebGL
- [ ] Friendly Errors

### p5.js version

v2.2.1 and v2.2.2-rc.2 (reproduces in both)

### Web browser and version

Google Chrome 145.0.7632.75 (Official Build, 64-bit)

### Operating system

Linux Ubuntu 24.04 LTS

### Steps to reproduce this

1. Create a WebGL canvas.
2. Create a secondary createGraphics() buffer in WEBGL mode.
3. Draw a shape on the graphics buffer.
4. Call filter() on the graphics buffer.

The sketch crashes with a `TypeError`

### Snippet:

```js

function setup() {

  const g = createGraphics(100, 100, WEBGL);
  g.noStroke();
  g.fill('red');
  g.circle(0, 0, 40);

  g.filter(INVERT); // crashes
}

```
### Console Error:

```
Uncaught (in promise) TypeError: e.getColor is not a function
    at _makeFilterShader (p5.min.js)
    at filter (p5.min.js)
    at Graphics.filter
```

Calling filter() on a createGraphics(WEBGL) renderer causes a runtime TypeError originating inside the WebGL filter shader pipeline.

aashu2006:
Resolves #8549

### Changes

Fixes `filter()` crash when called on `createGraphics(WEBGL)` buffers.

Strands previously augmented only `p5.prototype`, but `p5.Graphics` does not inherit from it. When `filter()` executed on a `p5.Graphics` instance, strands callbacks (e.g. `getTexture`, `vec*`, `dot`, etc.) were missing, causing runtime errors.

This PR:

- Introduces two helpers:
  - `augmentFn` (permanent, unconditional overwrite)
  - `augmentFnTemporary` (temporary hook with restoration via `graphicsOverrides`)
- Mirrors all permanent strands API augmentations from `p5.prototype` onto `p5.Graphics.prototype`
- Mirrors temporary shader hooks during `modify()` and restores them in `deinitStrandsContext`
- Ensures `createGraphics(WEBGL).filter()` has access to the full strands API layer
- Prevents temporary hook leakage after `modify()` completes

WebGL unit tests and visual tests pass locally. Manual testing confirms the crash is resolved.

Screenshot from 2026-02-25 21-10-48

Manual test confirming no crash and correct behavior:

- `createGraphics(WEBGL).filter(INVERT)` runs without error
- Temporary hooks are cleaned up after execution

#### PR Checklist
- [x] `npm run lint` passes


aashu2006:
@ksen0, @davepagurek Thanks for taking a look!

This implements the strands => `p5.Graphics.prototype` mirroring we discussed in #8549. Permanent strands API functions are mirrored via `augmentFn`. Temporary shader hooks are mirrored via `augmentFnTemporary` and cleaned up properly. All tests pass locally (WebGL unit & visual), and manual testing confirms `createGraphics(WEBGL).filter()` no longer crashes. Let me know if you would like any clarifications or adjustments before merge, happy to refine.

src/strands/strands_api.js
```diff
@@ -114,6 +114,39 @@ function installBuiltinGlobalAccessors(strandsContext) {
...
+/**
+ * Temporarily augment both p5.prototype (fn) and p5.Graphics.prototype
+ * with a hook function. Saves the previous own property value in
+ * overridesStore so deinitStrandsContext can restore it.
+ */
+function augmentFnTemporary(fn, p5, name, value, overridesStore) {
+  fn[name] = value;
+  const GraphicsProto = p5?.Graphics?.prototype;
+  if (GraphicsProto) {
+    overridesStore[name] = Object.prototype.hasOwnProperty.call(GraphicsProto, name)
```
davepagurek: Maybe we should just pass `strandsContext` in so that we can always make sure we assign to `strandsContext.graphicsOverrides` here? Saves a bit of indirection. We could also then update this to assign to `strandsContext.fnOverrides`, `window`, and `windowOverrides` so that we're handling all of those uniformly.

src/strands/strands_api.js
```diff
@@ -114,6 +114,39 @@ function installBuiltinGlobalAccessors(strandsContext) {
...
+ */
+function augmentFn(fn, p5, name, value) {
+  fn[name] = value;
+  const GraphicsProto = p5?.Graphics?.prototype;
+  if (GraphicsProto) {
+    GraphicsProto[name] = value;
+  }
+}
+
+/**
```
davepagurek:
```suggestion
/*
```
If we use two asterisks at the top, it's a doc comment, which gets put in the public p5 reference, so we want to save that just for public APIs.

src/strands/strands_api.js
aashu2006: I have refactored `augmentFnTemporary` to accept `strandsContext` directly and assign to `strandsContext.graphicsOverrides` internally, aligning it with how `fnOverrides` and `windowOverrides` are handled. No behavioral changes, all WebGL unit and visual tests pass locally.

src/strands/strands_api.js
aashu2006: Got it! I have done the change.

aashu2006:
Hi @davepagurek, I have addressed the review feedback:

- Refactored `augmentFnTemporary` to accept `strandsContext` directly and assign to `strandsContext.graphicsOverrides` internally, aligning override handling with `fnOverrides` and `windowOverrides`.
- Added a visual regression test inside the existing `filter` suite that creates a `createGraphics(WEBGL)` buffer, runs `filter(INVERT)`, draws it back to the main canvas, and screenshots the result.
- Confirmed WebGL unit tests and visual tests pass locally.

Let me know if anything else should be adjusted, thanks!

src/strands/strands_api.js
```diff
@@ -726,7 +759,7 @@ export function createShaderHooksFunctions(strandsContext, fn, shader) {
       strandsContext.windowOverrides[name] = window[name];
       strandsContext.fnOverrides[name] = fn[name];
       window[name] = hook;
-      fn[name] = hook;
+      augmentFnTemporary(fn, strandsContext, name, hook);
```
davepagurek, line 762: It looks like this method assigns to `fn` and still without assigning to `fnOverrides`, my suggestion earlier is to make sure we're always assigning to a target + the overrides map for that target at the same time. Since we also always assign to window, fn, and Graphics, I think we should basically move the three lines above this one into `augmentFnTemporary` to have those all co-located.

aashu2006: I have refactored `augmentFnTemporary` to centralize override handling for `window`, `fn`, and `Graphics`, so assignment and override bookkeeping are now co-located. WebGL unit and visual tests are passing locally.

aashu2006:
@davepagurek Done 👍

davepagurek (approved):
Thanks, looks good to go!

PR #8601, opened 2026-03-02
Fix touches being empty in touchEnded() by using changedTouches

suhr25:
## SUMMARY

Fixes `touchEnded()` getting an empty `touches[]`.
Now uses `changedTouches` when `touches` is empty (per spec).
Changes in `src/events/touch.js` + a small unit test.

---

## FIX

* **Issue:** `e.touches` is empty on `touchend`, so loop never runs
* **Fix:** fallback to `e.changedTouches`

**Before**

```js
for (let i = 0; i < e.touches.length; i++) {
```

**After**

```js
const touchList = e.touches.length ? e.touches : e.changedTouches;
for (let i = 0; i < touchList.length; i++) {
```
---

## VERIFICATION

* Run tests → should pass
* Manual test:

```js
function touchEnded() {
  console.log(touches.length);
}
```

**Before:** 0
**After:** 1 (correct final touch)

### DIFF
--- a/src/events/touch.js
+++ b/src/events/touch.js
@@ -95,7 +95,11 @@ p5.prototype.touches = [];
 p5.prototype._updateTouchCoords = function(e) {
   if (this._curElement !== null) {
     const touches = [];
-    for (let i = 0; i < e.touches.length; i++) {
+    // During a touchend event, the lifted finger is removed from e.touches
+    // (W3C spec) and is only present in e.changedTouches. Fall back to
+    // e.changedTouches so that touches[] is populated inside touchEnded().
+    const touchList = e.touches.length > 0 ? e.touches : e.changedTouches;
+    for (let i = 0; i < touchList.length; i++) {
       touches[i] = getTouchInfo(
         this._curElement.elt,
         this.width,
--- a/test/unit/events/touch.js
+++ b/test/unit/events/touch.js
@@ -138,6 +138,24 @@ suite('Touch Events', function() {
   });
 
   suite('touchEnded', function() {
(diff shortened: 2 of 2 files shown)


suhr25:
Hi @davepagurek @ksen0,
Can you please take a look at this.
Thanks!

ksen0:
@suhr25 is there a bug report or a sketch that shows the problem behavior that this is fixing? Thanks!

suhr25:
Hi @ksen0,
Yes ,here is a minimal sketch that reproduces the issue:

```js
function setup() {
  createCanvas(200, 200);
}

function touchEnded() {
  console.log(touches.length);
}
```
On mobile browsers , lifting the last finger logs `0`, even though a touch just ended. This happens because on `touchend`, `event.touches` is often empty, while the ended touch is only present in `event.changedTouches`.
This PR falls back to `changedTouches` when `touches` is empty so `touchEnded()` receives the final touch correctly.

Thanks!

perminder-17 (commented):
Event | e.touches | e.changedTouches | Before (touches.length) | After this PR (touches.length)
-- | -- | -- | -- | --
1st finger lifted | [F2, F3] | [F1] | 2 | 2
2nd finger lifted | [F3] | [F2] | 1 | 1
3rd (last) finger lifted | [] | [F3] | 0 | 1 (Since you added a fallback)


For me, this is not a bug but an API design question, where you could probably open up an issue and stewards or other contributor could discuss on the behaviour. 

So, hence closing this PR, and you could probably open an issue for that and if the issue got approved, we can then go ahead with this PR. Thanks for your work and patience.

PR #8603, opened 2026-03-03
Continuous release with pkg.pr.new and esm.sh

### LINKED ISSUE #8570: Continuous Release with pkg.pr.new and esm.sh
### Increasing access

This does not affect end user necessarily. With continuous release, we can create testable CDN hosted build of the library per push/PR so that features that are inactive development (eg. p5.strands) can be more easily tested without needing to wait for a beta release or some kind of local build and distribution.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### Feature request details

Continuous release for p5.js will involve two services, [pkg.pr.new](https://pkg.pr.new) and [esm.sh](https://esm.sh/). The way that it works is by first installing the GitHub app for pkg.pr.new and setting up the relevant CI hook, this will then run the build and publish the same set of artifacts we publish to NPM to pkg.pr.new, this includes the CDN files that can be used in `<script>` tags. However, pkg.pr.new itself only replicate the behavior of NPM and does not host the files as a CDN, in this case we will need esm.sh which serves the files from pkg.pr.new as a CDN to actually serve the built p5.js library files, this happens automatically.

A few things to consider: 
* To ensure only approved versions of the library are built, the CI workflow should be set to require maintainer approval so that maintainers can review the changes before allowing the build to proceed. 
* The automatic comment left by pkg.pr.new on PR upon successful deployment will need to be customized to provide link to the esm.sh version as well.
* The esm.sh link will need to use the "[Escape hatch: raw source files](https://esm.sh/#docs)" link of `https://raw.esm.sh` for p5.js files to enable it to be used as `<script>` tag `src` property value.

One potential problem this can have and we want to avoid is for these WIP builds to be used as some kind of default version somewhere, if we publish these, they must be used for contributor testing purposes only and shouldn't be considered officially supported versions. They must not be used in any production environment.

---

_This issue is a request for comment, please do leave feedbacks or questions if there are any, but do not ask to be assigned._

limzykenneth:
Resolves #8570

 Changes:

Add dev dependency and workflow required to enable continuous release. Still need to verify that the deployment and custom message is working as intended.


limzykenneth:
@ksen0 @davepagurek This is done now and ready for review. The build and release are triggered on pull request event so each push update will run it, similar to the regular test CI; it will also run on push to `dev-2.0` branch which should mean that there will always be a head commit build.

Once built, the package will be published to pkg.pr.new in the same manner as publishing on NPM except it will be tagged with the specific commit rather than version number. Once publish, esm.sh will automatically pick it up and the relevant CDN link will be available. We should change the CI build run to require approval for all external contributors before merging this.

Also involved is the comment bot @p5-bot, which will comment on PR like the above with relevant links ready to be used. "CDN link" is the link ready to be used in a script tag; "Published Packages" is the link that can be `npm install`ed. As additional push are made to the PR and new version is built and published, the previous link will be recorded in the collapsible section under the relevant commit hash that the build is generated from.

Please review the implementation to check I didn't miss any specific edge cases especially in terms of any unintended publishing. Once it all looks good we can squash and merge since the history here is mostly trial and error stuff from me.

.github/workflows/continuous-release.yml
```diff
@@ -0,0 +1,55 @@
+name: Publish approved pull requests and latest commit to pkg.pr.new
+on:
+  pull_request:
+    branches:
+      - 'dev-2.0'
+  push:
+    branches:
+      - 'dev-2.0'
+    tags:
+      - '!**'
```
ksen0, line 10: Extremely minor: or keep RCs?

.github/workflows/continuous-release.yml
```diff
@@ -0,0 +1,55 @@
...
+          node-version: 22
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build library
+        run: npm run build
+
+      - name: Publish library
+        run: npx pkg-pr-new publish --no-template --json output.json --comment=off
```
ksen0, line 34: also minor: pin pkg-pr-new version?

.github/workflows/continuous-release.yml
limzykenneth: RC will get published on NPM so it is available through jsdelivr already

.github/workflows/continuous-release.yml
limzykenneth: `pkg-pr-new` is pin as a dev dependency in the repo and not through an action so should be fixed by `package-lock.json`

Issue #8551, opened 2026-02-22
[p5.js 2.0 Bug Report]: saveTable() doesn't escape double quotes in CSV, round-trip corrupts data


avinxshKD:
### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [x] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [ ] Other (specify if possible)

### p5.js version

2.x (main)

### Web browser and version

any

### Operating system

any

### Steps to reproduce this

[saveTable()](vscode-file://vscode-app/c:/Users/avina/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) wraps values containing commas in quotes (that got fixed in an old PR), but it never escapes " characters inside those values. So any cell with both a comma and a quote produces malformed CSV.

```
function setup() {
  let t = new p5.Table();
  t.addColumn('name');
  t.addColumn('note');

  let r = t.addRow();
  r.setString('name', 'thing');
  r.setString('note', 'He said "yes, go ahead"');

  saveTable(t, 'out.csv');
}
```

Output:

```
name,note
thing,"He said "yes, go ahead""
```

That middle " is unescaped. Most CSV parsers (Excel, Python csv, loadTable itself) will choke on it or misread the row.

RFC 4180 says inner quotes must be doubled. Should be:

```
name,note
thing,"He said ""yes, go ahead"""

```

Also: headers with commas or quotes are never escaped either. And a newline inside a value will silently break row boundaries with no quoting at all.

[loadTable](vscode-file://vscode-app/c:/Users/avina/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) handles all of this correctly on the way in. So load a valid CSV → save it back → file is now broken. Classic round-trip corruption.

The fix is one line per write call: escape " as "" before wrapping, and trigger quoting on " or [\n](vscode-file://vscode-app/c:/Users/avina/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) too, not just ,.

[...]

PR #8194, opened 2025-10-24
Add `AGENTS.md`

SableRaf:
This PR introduces an `AGENTS.md` file, which defines clear guidelines for how AI-assisted tools (like Claude, ChatGPT, Copilot, etc.) should interact with this repository and its contributors. 

### Motivation
We’ve noticed more contributors using coding assistants when opening PRs, sometimes without taking time to understand the corresponding issue or read our [contribution guidelines](https://github.com/processing/p5.js/blob/main/CONTRIBUTING.md).

As automated coding assistant are increasingly included by default at every step of the contribution process, new contributors are bound to encounter them and use them in good faith. While these tools can help newcomers navigate the codebase, they often generate well-meaning but unhelpful submissions. For example, large unreviewed code changes or fixes that contributors cannot fully explain or take responsibility for.

While other projects might simply close AI-generated PRs without a review, we usually try to offer feedback and guidance instead. However, the growing number of these submissions makes that approach increasingly difficult and time-consuming for our stewards, reviewers, and maintainers.

The goal of this PR is to add clear instructions that **help AI coding tools act more like guides than code generators**. When someone uses an assistant to contribute, the tool will be prompted to explain the code, point to our documentation, and suggest asking questions in the community.

### Approach
This PR uses the open `AGENTS.md` format (see https://agents.md/) as a prompting layer to shape how coding agents behave when set to work in this repository.

When used in development environments like VS Code, Cursor, or Aider, AI agents usually begin by scanning the project files to understand how the codebase works and what rules to follow. Because reading all files repeatedly would be slow and expensive, they perform this analysis once and save a summary in a separate file (for example, `CLAUDE.md`, `.cursorrules`, or `.aider.conf`). This “analysis file” helps the agent remember key details and reuse them in future sessions without re-scanning the entire repository.

We leverage this behavior by asking AI tools to include our AI guidelines directly in their analysis file. Since these files are often the first and only summary an agent refers to in later sessions, embedding our rules there ensures they remain visible and active whenever the tool is used again.

For more context on agentic coding, this (promotional) [short video](https://www.youtube.com/watch?v=AJpK3YTTKZ4) does a good job of explaning the workflow.

> [!NOTE]
> Including an `AGENTS.md` file does not mean we endorse any specific AI tool or service, or encourage their use.

### Limitations
- This approach will only work for agentic workflows. 
- AI tools may still misunderstand the file or ignore it. 
- Contributors may delete the file or prompt around it. 

[...]

PR #8604, opened 2026-03-05
[TESTERS NEEDED] Rewrite AGENTS.md with community-first agent guidelines

SableRaf:
> [!NOTE]
> See more context in the similar PR in https://github.com/processing/processing4/pull/1453

This PR is a rewrite of the `AGENTS.md` first introduced in [processing/p5.js\#8194](https://github.com/processing/p5.js/pull/8194).

### What's new?

- Agents should now point contributors toward Discord, CONTRIBUTING.md, CODE_OF_CONDUCT.md, and the contributor guidelines before doing anything else.
- Agents should now try to understand where their interlocutor is coming from (e.g.: Are they a first-time contributor? Have they used p5.js before? Do they have JavaScript experience?) and adapt their responses accordingly.
- Responses should be shorter and more conversational.
- Agents should redirect contributors who are looking for a different project (Processing Java, p5.js Web Editor, p5.js Website) to the right repo.
- Agents should now be aware of the 1.x/2.0 branch split and should point contributors to the 2.x project board.
- Agents should now check that an issue is unassigned and approved before encouraging someone to work on it.
- GSoC candidates should be directed to the Processing Foundation's GSoC guidelines.
- Agents should now suggest non-code contributions like translation, documentation, and testing.

### How to test this PR

The best way to test this is to check out this PR an agentic coding environment (Claude Code, Copilot, Cursor, etc.) and try a few contributor scenarios:

- A first-time contributor with no specific issue in mind (should be onboarded and helped to find an open issue)
- Someone asking about Processing Java (should be redirected to the Processing repo)
- A GSoC candidate wanting to contribute a PR (should be redirected to GSoC Guidelines)
- An inexperienced contributor who wants to jump straight to writing code (should be redirected)
- An experienced contributor wanting to skip the onboarding (should be allowed)

Test the same input multiple times and check whether the responses are consistently helpful for the particular scenario.

Note any misaligned behavior and share it in the comments below. Please specify which model, version, and environment was used.

Note that this is largely subjective so use your judgment when you evaluate agent behavior. Discussion in the comments is encouraged.

### Common failure-cases for agent behavior

- Not reading the `AGENTS.md` file
- Recommending issues that are already assigned or not ready for contribution
- Being suspicious of or interrogating the contributor's motivations
- Jumping straight to code or codebase analysis without first orienting the contributor
- Responding to a p5.js question with Processing contribution guidelines
- Doing the thinking for the contributor, instead of redirecting to the community
- Asking multiple questions at once in a way that feels like an interrogation
- Being preachy or reciting the guidelines back to the contributor
- Replying with an overwhelming wall-of-text

### Other changes
- Added `copilot-instructions.md` to direct copilot to read AGENTS.md, since it tended to skip it.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,1 +1,1 @@
-ALWAYS refer to the AGENTS.md document for complete guidelines.
+BEFORE you do ANYTHING ELSE, ALWAYS refer to the AGENTS.md document for complete guidelines.


ksen0:
Thanks @SableRaf ! This update seems like a positive improvement.

AGENTS.md
```diff
@@ -1,152 +1,280 @@
...
+
+The best time to help is before any code is written. 
+
+That is when the contributor is most open to guidance, and when your guidance can have the most impact.
+
+Do not fetch issue details, scan the codebase, check assignments, or start writing code until you have had a conversation with the contributor. The goal is to help them understand the project and find their way into it, not to get them to a PR as quickly as possible. Make sure they are oriented to the community and the issue before you do anything else. That is how good contributions begin. If they are in the wrong place entirely, redirect them warmly to where they need to go.
+
+## How to help
+
+### 0. Welcome them and find out where they are
```
davepagurek, line 112: Not super important but both this and the next point are labelled as `0`, is that intentional?

SableRaf: Ah good catch! I fixed it

SableRaf:
Note that Claude CLI tends to ignore `AGENTS.md` if the initial user request is specific enough (tested with: "I want to fix issue https://github.com/processing/p5.js/issues/8551")

I have managed to correct this behavior locally by adding a `CLAUDE.md` file that includes the line: 

`BEFORE you do ANYTHING ELSE, ALWAYS refer to the AGENTS.md document for complete guidelines.`

However, I have resisted including vendor-specific prompt files in the root of the project so far, as it feels like free advertising or even endorsement.

Would it be worth it in this case? Curious what others think.

SableRaf:
Failure cases when asking to explore the codebase

## Claude acts suspicious of the contributor: 

image


## Claude skips reading AGENTS.md:

image

SableRaf:
I got better results with GPT

image

image

image

davepagurek:
I'm also a little hesitant, at least at first, to add instructions for specific tools, for the reasons you describe and also because it seems like we might be playing whack-a-mole as new tools come up. But it's also true that new tools, and even commonly used ones like Claude, may just ignore our instructions. To a certain extent, I feel like maybe we do what we can via methods like this, but then beyond a reasonable amount of effort on our part, fall back on our AI usage policy to close PRs that ignore them?

Megabyte135:
Based on experience from several test runs. I didn’t have Claude at hand, so the experiments were carried out with Codex 5.3 in xhigh mode. Overall, the result matches the desired behavior, except for three scenarios:
- GSoC. With or without AGENTS.md, Codex decided not to remind me about the GSoC Guidelines.
- It stubbornly refuses to say anything about Processing Java.
- If you ask it to perform a specific task, about half the time it jumps straight into doing it without even trying to read AGENTS.md.

I had the idea of trying a kind of prompt injection to force the agent to read the instructions file, but that did not produce much of a result: even when the warning about the need to read AGENTS.md was surfaced as ESLint warnings, it could still ignore them without issue, while saying something like "it's default warning, nothing to worry about". It seems that advanced agents are smart enough to ignore commands coming from tools (ls/dir with instructions in file names, warnings from eslint, and so on).

I'm agree, I think it’s not worth trying to force an AI agent to familiarize itself with the instructions when this is its owner responsibility. Instead of adding support for every vendor’s file format, maybe it would be better to make that process a bit easier for AI users. For example, it may be worth adding a special section under “AI Usage Policy” in README.md (agents sometimes tend to read this before starting a job, even while ignoring AGENTS.md) with an onboarding prompt, since it seems that the initial user command cannot be reliably overridden. Something like:

> Read the AGENTS.md file located in the project root and follow the instructions and guidance it contains. It is your responsibility to maintain compliance with these instructions within the project, both by following them yourself and by ensuring that all sub-agents within your area of responsibility follow them as well. Once you have finished reviewing it, proceed to carry out my next request (below).
> {_**here goes request**_}

It'll make things easier for human contributors. A bit.

PR #8618, opened 2026-03-06
improve wording in setBlue documentation.

Eritrix:
Changes:
Improved wording in the setBlue() documentation to make the value range clearer.

Screenshots:
Not applicable (documentation change).


limzykenneth:
`setRed` and `setGreen` uses similar phrasing, why is only `setBlue` changed in this case?

PR #8620, opened 2026-03-07
harden GitHub Actions workflows

perminder-17:
This PR improves the security of our GitHub Actions workflows.

### DIFF
--- a/.github/workflows/auto-close-issues.yml
+++ b/.github/workflows/auto-close-issues.yml
@@ -7,13 +7,18 @@ on:
       - 2.0
       - dev-2.0
 
+permissions:
+  contents: read
+  issues: write
+  pull-requests: read
+
 jobs:
   close_issues:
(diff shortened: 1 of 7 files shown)


.github/workflows/auto-close-issues.yml
```diff
@@ -7,13 +7,18 @@ on:
       - 2.0
       - dev-2.0
 
+permissions:
+  contents: read
+  issues: write
```
limzykenneth, line 12: @ksen0 Do you know this and the write permission in `stewards-update.yml` below is working?

ulgens:
Is there a discussion about this implementation? It looks like it's implementing recommendations from zizmor, but zizmor itself is not in CI so any changes in future won't be validated against the same rule set.

ksen0:
@ulgens I think would be great to discuss in the contribute discord channel or in https://github.com/processing/p5.js/issues/8674

PR #8637, opened 2026-03-12
Fix saveTable() not escaping double quotes in CSV output

### LINKED ISSUE #8551: [p5.js 2.0 Bug Report]: saveTable() doesn't escape double quotes in CSV, round-trip corrupts data

avinxshKD:
Resolves #8551

Changes:
`saveTable()` only quoted CSV fields containing commas, but never escaped `"` inside them, producing malformed output. Added `csvEscape()` next to the existing `escapeHelper`, which doubles inner quotes and triggers quoting on `"` and `\n` too, not just `,`. Also fixes headers, which were never quoted at all. Updated the CSV download test to assert against the actual expected RFC 4180 output instead of rebuilding the same broken string.

### DIFF
--- a/src/io/files.js
+++ b/src/io/files.js
@@ -2272,6 +2272,16 @@ function escapeHelper(content) {
     .replace(/'/g, '&#039;');
 }
 
+// RFC 4180 escape: double inner quotes, wrap when value has comma/quote/newline
+function csvEscape(val, ext) {
+  if (ext !== 'csv') return String(val);
+  let s = String(val);
+  if (s.includes(',') || s.includes('"') || s.includes('\n')) {
+    return '"' + s.replace(/"/g, '""') + '"';
+  }
+  return s;
+}
+
 /**
  *  Writes the contents of a <a href="#/p5.Table">Table</a> object to a file. Defaults to a
  *  text file with comma-separated-values ('csv') but can also
@@ -2331,9 +2341,9 @@ p5.prototype.saveTable = function(table, filename, options) {
     if (header[0] !== '0') {
       for (let h = 0; h < header.length; h++) {
         if (h < header.length - 1) {
-          pWriter.write(header[h] + sep);
+          pWriter.write(csvEscape(header[h], ext) + sep);
         } else {
-          pWriter.write(header[h]);
+          pWriter.write(csvEscape(header[h], ext));
         }
       }
       pWriter.write('\n');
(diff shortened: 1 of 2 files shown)


avinxshKD:
Hey @davepagurek @ksen0 pls take a look, added csvEscape() beside the existing escapeHelper, doubles inner quotes and triggers quoting on " and \n, not just ,. Headers get the same treatment now too.

perminder-17 (commented):
Hi, I am closing this one since in the issue it's mentioned a 2.0 bug `[p5.js 2.0 Bug Report]` and you've raised a PR targetting the `main` branch. I think this should be targetting `dev-2.0` branch?

PR #8644, opened 2026-03-15
Fix TypeScript typing for filterColor shader hook

### LINKED ISSUE #8574: [p5.js 2.0 Bug Report]: p5.strands | methods and properties of filterColor not known in instance mode

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.2.2.

### Web browser and version

Safari 26.3

### Operating system

MacOSX

### Problem

Even though they are definetly usable, in instance mode the methods and properties of filterColor don't show up on autocomplete and typescript says they don't exist on type object (ts(2339)). 

### Snippet:

```js

import p5 from "p5";

let sketch = (p5: p5) => {
  let filterShader: p5.Shader;

  function testShader({ p5 }: { p5: p5 }) {
    p5.filterColor.begin();
    let coord = p5.vec2(p5.filterColor.texCoord);
    p5.filterColor.set([coord, 0, 1]);
    p5.filterColor.end();
  }

  p5.setup = () => {
    p5.createCanvas(800, 800, p5.WEBGL);
    p5.fill(255);

    filterShader = p5.buildFilterShader(testShader, { p5 });
  };

  p5.draw = () => {
    p5.background(0);
    p5.filter(filterShader);
  };
};

new p5(sketch);

```

Kathrina-dev:
Resolves #8574

## Changes

Fixes incorrect TypeScript typing for the `filterColor` shader hook.

Previously, the generated TypeScript definitions declared:

declare const filterColor: object;

Because of this, accessing documented properties such as:

- `filterColor.texCoord`
- `filterColor.canvasSize`
- `filterColor.begin()`
- `filterColor.set(...)`

resulted in TypeScript errors like:

Property 'texCoord' does not exist on type 'object'.

## Other Methods attempted at resolving the error

Initially I attempted to resolve this through JSDoc annotations in the source by using:

- `@property`
- `@typedef`

to explicitly define the structure of `filterColor`.

However, the documentation generator used in the build pipeline still produced the type as `object` in the generated TypeScript definitions. Because of this limitation in the JSDoc → TypeScript generation step, the correct structure was not preserved.

## Solution

To ensure the generated typings match the documented API, I added a patch in:

utils/patch.mjs

This patch updates the generated declaration to:
```
declare const filterColor: {
texCoord: any;
canvasSize: any;
texelSize: any;
canvasContent: any;
begin(): void;
end(): void;
set(color: any): void;
};
```

This aligns the TypeScript definitions with the documented properties of the `filterColor` shader hook and prevents TypeScript errors when accessing them.

## Screenshots of the change

N/A (TypeScript typing fix)

## Suggested Reviewers
- @christophseibel
- @davepagurek

#### PR Checklist

- [x] `npm run lint` passes


utils/patch.mjs
```diff
@@ -3,18 +3,18 @@ import fs from 'fs';
...
-  
+
   const replace = (path, src, dest) => {
     if (Array.isArray(path)) {
       path.forEach(path => replace(path, src, dest));
       return;
     }
     try {
-      if (!path.startsWith("types/"))
-        path = "types/" + path;
```
davepagurek, line 14: heads up, looks like you have an autoformatter running on the file. You'll want to turn that off when working on the p5 codebase because it makes diffs in PRs harder to read since a nunch of unrelated changes get bundled in with your change.

davepagurek (commented):
Thanks for taking this on @Kathrina-dev! While the original issue was for `filterColor`, I think this actually applies to all the hook objects documented around `filterColor` in the code. While we *could* add a patch method for each one, this seems likely to drift from the documented properties, especially as we add new hooks. It might be worth instead taking the approach of updating the typescript generation script to work with JSDoc typedefs (or something else, like documenting a pretend class for each hook object or something?) so that we can have the jsdoc be the sole source of truth for both documentation and types.

Kathrina-dev:
Thanks for the suggestion, I've already removed the formatted part of the code to make it easier to identify the diff. 

My initial approach was to use Typescript generation script (@property, @typedef) but it kept collapsing the JSDoc typedef into an object. I agree that adding a patch would cause a drift from the documented properties, so I'll try the alternative way you've suggested and let you know the results.

davepagurek:
Hi! If you run `npm run test:types`, it gets a number of errors, e.g.:
```
Error: types/global.d.ts(1387,21): error TS2749: 'CENTER' refers to a value, but is being used as a type here. Did you mean 'typeof CENTER'?
Error: types/global.d.ts(1387,30): error TS2749: 'RADIUS' refers to a value, but is being used as a type here. Did you mean 'typeof RADIUS'?
Error: types/global.d.ts(1387,39): error TS2749: 'CORNER' refers to a value, but is being used as a type here. Did you mean 'typeof CORNER'?
Error: types/global.d.ts(1387,48): error TS2749: 'CORNERS' refers to a value, but is being used as a type here. Did you mean 'typeof CORNERS'?
Error: types/global.d.ts(1574,36): error TS2749: 'FALLBACK' refers to a value, but is being used as a type here. Did you mean 'typeof FALLBACK'?
Error: types/global.d.ts(1574,47): error TS2749: 'LABEL' refers to a value, but is being used as a type here. Did you mean 'typeof LABEL'?
Error: types/global.d.ts(1636,60): error TS2749: 'P2D' refers to a value, but is being used as a type here. Did you mean 'typeof P2D'?
Error: types/global.d.ts(1636,66): error TS2749: 'WEBGL' refers to a value, but is being used as a type here. Did you mean 'typeof WEBGL'?
Error: types/global.d.ts(1636,74): error TS2749: 'P2DHDR' refers to a value, but is being used as a type here. Did you mean 'typeof P2DHDR'?
Error: types/global.d.ts(1670,24): error TS2749: 'FALLBACK' refers to a value, but is being used as a type here. Did you mean 'typeof FALLBACK'?
Error: types/global.d.ts(1670,35): error TS2749: 'LABEL' refers to a value, but is being used as a type here. Did you mean 'typeof LABEL'?
Error: types/global.d.ts(1704,136): error TS2749: 'BLEND' refers to a value, but is being used as a type here. Did you mean 'typeof BLEND'?
Error: types/global.d.ts(1704,144): error TS2749: 'DARKEST' refers to a value, but is being used as a type here. Did you mean 'typeof DARKEST'?
Error: types/global.d.ts(1704,154): error TS2749: 'LIGHTEST' refers to a value, but is being used as a type here. Did you mean 'typeof LIGHTEST'?
Error: types/global.d.ts(1704,165): error TS2749: 'DIFFERENCE' refers to a value, but is being used as a type here. Did you mean 'typeof DIFFERENCE'?
Error: types/global.d.ts(1704,178): error TS2749: 'MULTIPLY' refers to a value, but is being used as a type here. Did you mean 'typeof MULTIPLY'?
Error: types/global.d.ts(1704,189): error TS2749: 'EXCLUSION' refers to a value, but is being used as a type here. Did you mean 'typeof EXCLUSION'?
```

Looks like the way the existing typedefs are being exported has changed -- the typedef itself should be a type, but also there should be a const of the same name from the rest of the jsdoc.

Kathrina-dev:
I think the issue was that some constants (like BLEND, ADD, etc.) weren’t being picked up properly. I updated the lookup to include them and adjusted how they’re handled in the type conversion depending on where they’re used. That seems to fix the errors. 

Thanks for being patient with me, I'm new to contributing to Open Source and forgot to run the test cases before committing. I'll keep it in mind next time.

utils/typescript.mjs
```diff
@@ -595,6 +617,22 @@ function generateClassDeclaration(classData) {
 function generateTypeDefinitions() {
   let output = '// This file is auto-generated from JSDoc documentation\n\n';
 
+  const strandsMethods = processStrandsFunctions();
```
davepagurek: Any reason why this was moved up here from below?

utils/typescript.mjs
```diff
@@ -618,12 +659,26 @@ function generateTypeDefinitions() {
...
     }
-    const type = convertTypeToTypeScript(constant.type, { isInsideNamespace: false, isConstantDef: true });
+    let type;
+    // Avoid invalid self-referential types like `typeof FOO`
+    if (
+      constant.type?.type === 'NameExpression' &&
+      constant.type.name === constant.name
+    ) {
+      // Self-referential constant → fallback
+      type = 'number';
```
davepagurek: I think this is not sufficient yet. When I run `npm run generate-types`, and then open `types/p5.d.ts`, I see this:

```js
declare const P2D: number;

declare const __P2D: typeof P2D;
```

Previously, it looked like this:

```js
declare const P2D: 'p2d';

declare const __P2D: typeof P2D;
```

I think a next step should be not just to get the tests to pass -- first, I would `git checkout dev-2.0`, run `npm run generate-types`, and then make a copy of `types/p5.d.ts`.  Then, if you `git checkout fix-filterColor-object-type` again, the goal will be to make it so that when you run `npm run generated-types`, and you compare the newly generated file with the old one via `diff previousTypesFile.d.ty types/p5.d.ts`, you should only see the addition of the `FilterColorHook` type.

src/strands/p5.strands.js
```diff
@@ -612,7 +612,18 @@ if (typeof p5 !== "undefined") {
  */
 
 /**
- * @property {Object} filterColor
+ * @typedef {Object} FilterColorHook
+ * @property {any} texCoord
+ * @property {any} canvasSize
+ * @property {any} texelSize
+ * @property {any} canvasContent
+ * @property {function(): undefined} begin
```
davepagurek, line 824: Should these be `: void` too or is there a reason for them to be `: undefined`?

davepagurek (commented):
Thanks for making some updates, good to see the tests passing now! It looks like although they past, the generated type files still aren't 100% correct. I left some comments, one of which includes some steps I'd follow when working on this to help verify the change.

src/strands/p5.strands.js
Kathrina-dev: When I tried using `void`, the generated `p5.d.ts` ended up showing `any` instead. I tried using different ways to define a function but it still kept showing `any`. So I've defined it as `undefined` which showed up correctly while generating `p5.js.d.ts`

utils/typescript.mjs
Kathrina-dev: Yeah sure, I'll try comparing the diffs and try to find out what corrections need to be made to get the original structure + filterColorHook type

Kathrina-dev:
Thanks for taking out the time to leave those reviews, I'll look further into everything you've mentioned. This issue is a lot bigger than I expected but I'm very thankful for your help in this matter.

utils/typescript.mjs
Kathrina-dev: I had moved it earlier because it was being used too late in the generation flow and wasn’t available in time, which was causing some errors. But after the recent changes I made, I don't think it's needed to initialise it at the start of the function

Kathrina-dev:
Considering that this issue requires alot more modifications to the `convertTypeToTypeScript` and `generateTypeDefinitions` functions rather than just a small type fix issue. 

I'm considering to revert the changes back to how it was before and then creating a seperate function to handle Object types parsing. Since we already have separate functions for classes and methods, I think doing the same for typedefs will make it easier to handle things like function properties and avoid affecting other type conversions. This will make the `convertTypeToTypeScript` and `generateTypeDefinitions` less cluttered and make it modular. 

What are your thoughts on this approach @davepagurek ?

davepagurek:
That sounds ok to me @Kathrina-dev! Agreed that this gets a little hairy, thanks for your patience working on it!

Kathrina-dev:
@davepagurek I've added the new function to handle typedef properties. It passes the `npm run test:types` locally and I've also compared it to the generated types from `types/p5.d.ts` in the dev-2.0 branch and everything matches up exactly + the new introduced type interface. 

The only concern I have is about leaving the `begin()` and `end()` function as "undefined". I've tried to type it as void but after generating, it shows up as "any". I've made a temporary fix by parsing "undefined" and converting it to "void" in the `hasTypedefProperties` function but I'm not sure that's the correct approach. How could we best fix this issue?

davepagurek:
I think it's ok to leave `undefined` as the return type for `begin`/`end` if that's what it takes to get it to work, I just wanted to understand the reasoning first.

Thanks for making the updates! It looks like the constants have the right types now, although it looks like `FilterColorHook` is just showing up as `any` now:

```typescript
declare const FilterColorHook: any;

declare const __FilterColorHook: typeof FilterColorHook;

/**
 * A shader hook block that sets the color for each pixel in a filter shader. This hook can be used inside <a href="#/p5/buildFilterShader">`buildFilterShader()`</a> to control the output color for each pixel.
 * 
 * `filterColor` has the following properties:
 * 
 * - `texCoord`: a two-component vector representing the texture coordinates (u, v).
 * 
 * - `canvasSize`: a two-component vector representing the canvas size in pixels (width, height).
 * 
 * - `texelSize`: a two-component vector representing the size of a single texel in texture space.
 * 
 * - `canvasContent`: a texture containing the sketch's contents before the filter is applied.
 * 
 * Call `.set()` on the hook with a vector with four components (red, green, blue, alpha) to update the final color.
 * 
 * This hook is available in:
 * 
 * - <a href="#/p5/buildFilterShader">`buildFilterShader()`</a>
 */
declare const filterColor: object;

declare const __filterColor: typeof filterColor;
```

Kathrina-dev:
Totally forgot about that, thanks for pointing it out. Let me add a condition for it.

davepagurek:
Awesome, the types are looking good now! So I think we're good to move ahead with this method, and document the types of other filter hook blocks that are currently just typed as `Object`.

Kathrina-dev:
Yes! It looks good to me too, it 
- passes `npm run test:types`
- doesn't mutate any other types in the doc
- generates the filterColorHook along with it's `@property` definitions
- `filterColor` refers `filterColorHook`

I think it's good to go :)

Kathrina-dev:
@davepagurek , am I missing some other implementations? I think this PR is good to be merged, unless there's some work going on for an upcoming software release. Please let me know.

davepagurek:
Hi @Kathrina-dev! So you've added the types for `filterColor`, but also right below it is `objectInputs`, another hook that also needs types. There are a number of documentation blocks like that in the same file. Now that types get correctly passed through to TypeScript, we want to do what you did for `filterColor` to the rest of those remaining blocks.

Kathrina-dev:
@davepagurek that sounds like a nice idea. I will change all the required blocks and make another commit.

Kathrina-dev:
@davepagurek let me know if I missed anything there. It looks good from my end. Thanks for working with me on this issue.

src/strands/p5.strands.js
```diff
@@ -214,7 +214,10 @@ if (typeof p5 !== "undefined") {
 
 /* ------------------------------------------------------------- */
 /**
- * @property {Object} worldInputs
+ * @typedef {Object} WorldInputsHook
```
davepagurek, line 216: Can we add properties to all of these, similar to what we've done for `FilterColorHook`? So they'd have begin/end methods, some have .set(), and then they have whatever other properties are described in their descriptions.

Kathrina-dev: I've added them after your review. Please take a look.

src/strands/p5.strands.js
```diff
@@ -689,7 +724,15 @@ if (typeof p5 !== "undefined") {
  */
 
 /**
- * @property finalColor
+ * @typedef {Object} FinalColorHook
+ * @property {any} color
```
davepagurek, line 727: In https://github.com/processing/p5.js/pull/8706 we added `texCoord` as a property to this one, although I realize it also isn't mentioned in the bullet list of properties in the block below this. Do you think we could add it to both that and the type here now? I think that's the last thing before we're good to go!

Kathrina-dev: Ok I'll add it as a property as well as in the bulleted list in the doc.

davepagurek (approved):
I think this is good to go now, thanks for all your work on this!

davepagurek:
@all-contributors please add @Kathrina-dev for code

PR #8648, opened 2026-03-16
change strands examples to use millis() not uniform

### LINKED ISSUE #8645: Update p5.strands docs to not use a uniform for time
### Increasing access

Now that `millis()` works automatically in p5.strands, many animated shader examples in the reference will work without having to introduce the concept of uniforms to users, leaving that for more advanced usage, e.g. passing in the value of a slider or something like that.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

There are examples in many (all?) of the `build*Shader` methods, e.g. https://beta.p5js.org/reference/p5/buildfiltershader/, all of which can be updated.

nbogie:
Resolves #8645

### Changes:
1. Don't use a uniform to pass time in strands examples in p5.strands.js, material.js.  Instead use millis() built-in.

2. In the example of how to pass a uniform for modify(), don't pass time as this now is not motivating.  Instead, use an example where we'd NEED to use a uniform.  (I chose a count of mouse movements - is there a simpler, easily demonstrated value that is not also covered by a built-in?)

3. Snuck in a couple of minor tidy-ups of other strands examples:
3.1. removed old `<div><code>` element wrappers around examples in getTexture and smoothstep (the final remnants in the code-base).
3.2 had smoothstep examples use `buildFilterShader` instead of `baseFilterShader().modify`

#### `p5.strands.js`
one replacement in each of the following:
  - ✅ worldInputs: /reference/p5/worldInputs/ 
  - ✅ smoothstep: /reference/p5/smoothstep/
  - ✅ pixelInputs: /reference/p5/pixelInputs/ 
  - ✅ objectInputs: /reference/p5/objectInputs/
  - ✅ cameraInputs: /reference/p5/cameraInputs/

#### `material.js`
  - ✅ createShader: /reference/p5/createShader/ 
    - ✅ near text: "Make a version of the shader with a hook overridden"
  - buildFilterShader: /reference/p5/buildFilterShader/
    - ✅ near text: generate a color for each pixel
    - ✅ near text: You can also animate your filters over time using the
  - buildMaterialShader: /reference/p5/buildMaterialShader/
    - ✅ near text: One thing you can do with a material shader is animate the positions of vertices
  - loadMaterialShader: 
    - ✅ near text: Inside your shader file, you can call
  - buildNormalShader: /reference/p5/buildNormalShader/
    - ✅ near text: One thing you may want to do is update the position of all the vertices in an object over time
  - loadNormalShader: /reference/p5/loadNormalShader/
    - ✅ near text: might use the `finalColor` hook to change
  - buildColorShader: /reference/p5/buildColorShader/
    - ✅ near text: One thing you might want to do is modify the position
  - loadColorShader: /reference/p5/loadColorShader/
    - ✅ near text: Since this function loads data from another file
  - buildStrokeShader: /reference/p5/buildStrokeShader/
    - ✅ near text: You might also want to update some properties per vertex

### what I searched for

* `\*.*uniformFloat\(`
* `\*.*setUniform\(`
* `\*.*millis\(\)`

#### PR Checklist
- [x] `npm run lint` passes
- [x] [Unit tests] are included / updated - 
  some unrelated unit test failures (mostly time-outs): noise, shape modes, typography.
- [x] - Checked each modified example was still working

### AI Usage

No AI was used in this PR or this write-up

### DIFF
--- a/src/strands/p5.strands.js
+++ b/src/strands/p5.strands.js
@@ -329,7 +329,7 @@ if (typeof p5 !== "undefined") {
  *
  * function setup() {
  *   createCanvas(300, 200, WEBGL);
- *   fadeShader = baseFilterShader().modify(fadeCallback);
+ *   fadeShader = buildFilterShader(fadeCallback);
  * }
  *
  * function draw() {
@@ -356,7 +356,7 @@ if (typeof p5 !== "undefined") {
  *
  * function setup() {
  *   createCanvas(300, 200, WEBGL);
- *   animatedShader = baseFilterShader().modify(animatedFadeCallback);
+ *   animatedShader = buildFilterShader(animatedFadeCallback);
  * }
  *
  * function draw() {


src/webgl/p5.Shader.js
```diff
@@ -206,31 +206,39 @@ class Shader {
...
-   *
+   * //count of frames in which mouse has been moved
+   * let moveCounter = 0;
+   * 
    * function setup() {
    *   createCanvas(200, 200, WEBGL);
    *   myShader = baseMaterialShader().modify(() => {
-   *     // Get the current time from p5.js
-   *     let t = uniformFloat(() => millis());
+   *     // Get the move counter from our sketch
```
ksen0, line 224: ah thanks for the catch

ksen0:
> In the example of how to pass a uniform for modify(), don't pass time as this now is not motivating. Instead, use an example where we'd NEED to use a uniform. (I chose a count of mouse movements - is there a simpler, easily demonstrated value that is not also covered by a built-in?)

Mouse interaction makes sense! I think keyboard interaction is a more complex example that will need to rely on uniform. Ideally using `keyIsPressed` or `keyIsDown(..)`. I'll merge this in to include in updated RC but if the above makes sense and if you want to make a follow up to use keyboard interaction, please feel free!

PR #8666, opened 2026-03-22
fix: gracefully handle mixed-material OBJ models instead of crashing

Nixxx19:
## Summary

Fixes a crash in `parseObj()` (loading.js:655-658) that broke loading of real-world OBJ files exported from Blender, Sketchfab, and similar tools.

### Root cause

The check used to detect inconsistent vertex coloring was:

```js
// BEFORE (broken)
if (hasColoredVertices === hasColorlessVertices) {
  throw new Error('Model coloring is inconsistent...');
}
```

The `===` comparison fires in two completely different situations:

| State | Meaning | Should crash? | Did crash? |
|---|---|---|---|
| both `false` | Model has no faces at all | no | yes - bug |
| both `true` | Some faces have material colors, some don't | no | yes - bug |

Many standard OBJ exports have only some mesh groups with an explicit `usemtl` entry - this is valid OBJ syntax and common in practice.

### Fix

Changed the condition to only fire when both flags are `true` (the genuinely mixed case), and replaced the thrown error with a `console.warn` + vertex color strip so the model still loads:

```js
// AFTER (fixed)
if (hasColoredVertices && hasColorlessVertices) {
  console.warn('p5.js: This OBJ model has mixed material coloring...');
  model.vertexColors = [];
}
```

### Accessibility impact

Beginners loading common 3D assets from Sketchfab or Blender would get an opaque crash message with no path forward. This fix makes those models load and render with the default fill color, keeping the `loadModel()` experience frustration-free.

### Test changes

The existing test `'inconsistent vertex coloring throws error'` is updated to assert that a mixed-material model loads successfully and returns a `Geometry` with `vertexColors.length === 0`.

All 13 loadModel unit tests pass.

## Test plan

- [x] `npx vitest run test/unit/io/loadModel.js` - all 13 tests pass
- [x] Mixed-material model (`eg1.obj`) now loads successfully and emits a `console.warn`
- [x] Fully-colored models and no-material models are unaffected
- [x] No changes to public API


davepagurek:
In 2.x, I think we actually *can* support mixed colors. In `buildGeometry`, [we set the fill to -1 ](https://github.com/processing/p5.js/blob/1f0816dad13b16eb8fc1e61445ba50b934850ae1/src/core/p5.Renderer3D.js#L443)for anything where the fill isn't explicitly set, and at runtime it takes on whatever the global fill is. Rather than adjusting whether this is an error or a warning, maybe we should just handle it?

Nixxx19:
thank you @davepagurek for pointing out the buildGeometry pattern! updated the loader to push [-1, -1, -1, -1] for uncolored vertices so mixed models now render with the global fill for colorless faces and their material colors for colored ones. removed the warning since mixing is natively supported now. updated the test to verify both sentinel and real colors end up in vertexColors

davepagurek (approved):
Nice, thanks for the update!

PR #8676, opened 2026-03-25
Refactor p5.Vector to prioritize smaller dimension in binary operations on mismatched-size vectors

ksen0:
Because it's a major change, this needs more work and feedback. It is shared as a draft open for comment - even once it is merged into the 2.3 Release Candidate, please feel free to comment here or file new issues.

In 2.x, vectors can be of different dimensions. Binary operations should not be done on vectors  with mismatched dimensions; however, for backwards compatibility and to avoid breaking older code that might have edge cases (resulting from how 1.x handled 2D vectors), we decided to apply the logic of "always use the smaller dimension" when operating on two vectors with mismatched size. Note that solo numeric arguments (such as `createVector(1,2,3).mult(2)`) are still supported.

To discuss on [discord](http://discord.p5js.org), please see the "vector-discussion" channel in "Code" category. Prior community discussion which led to this is documented [in these meeting notes](https://docs.google.com/document/d/1WwFTFfyX6-lDsBcHeQJcHb6qQwrFfLeholZr33u4tlE/edit?tab=t.0). One important consideration here was that switching from 2D to 3D can benefit from a bit of intentional scaffolded friction: code should not completely break, but it's okay if it looks a bit wrong, because doing binary operations on mismatched vectors is not actually supported, and should not be done.

### Room for improvement

In one of the iterations, the `prioritizeSmallerDimension` helper also emitted a friendly error related to dimension mismatch, which **doesn't break execution** but is also **not supported**. However, there are multiple valid uses of 1-dimensional vector operating with 2 or 3 dimensional vector (mult, rem, div), and the logic below made false-positive warnings then. So it has been removed in the PR, but if there's a way to only emit warning when it's relevant, it can be added in the future.

```
const prioritizeSmallerDimension = function(currentVectorDimension, args) {
  return Math.min(currentVectorDimension, args.length);

  // if (args.length !== currentVectorDimension && args.length !== 1) {
  // TODO how to suppress for valid solo arguments?
  // this._friendlyError(
  //  `Operating on two vectors of different sizes, the smaller dimension is used. In this operation, both vector will be treated as ${minDimension}D vectors, and any additional values of the longer vector will be ignored.`, 'p5.Vector'
  //);
  //}
  //return minDimension;
};
```

#### PR Checklist

- [ ] `npm run lint` passes


src/math/math.js
```diff
@@ -92,21 +92,8 @@ function math(p5, fn) {
...
    */
-  fn.createVector = function (x, y, z) {
-    if (arguments.length === 0) {
-      p5._friendlyError(
-        'In 1.x, createVector() was a shortcut for createVector(0, 0, 0). In 2.x, p5.js has vectors of any dimension, so you must provide your desired number of zeros. Use createVector(0, 0) for a 2D vector and createVector(0, 0, 0) for a 3D vector.'
-      );
-    }
-    if (this instanceof p5) {
-      return new p5.Vector(
-        this._fromRadians.bind(this),
```
ksen0, line 102: This is done only in `p5.Vector`.

test/unit/webgl/p5.Shader.js
```diff
@@ -2242,7 +2242,9 @@ test('returns numbers for builtin globals outside hooks and a strandNode when ca
         }, { myp5 });
       } catch (e) { /* expected */ }
 
-      assert.isAbove(mockUserError.mock.calls.length, 0, 'FES.userError should have been called');
+
+
+      assert.isAbove(mockUserError.mock.calls.length, 0, 'FES.userError should have been called, btw: '+globalThis.FESCalled);
```
ksen0, line 2545: Hi @davepagurek ! I don't know why this is happening but some tests that as far as I can tell should not be affected are not passing:

Image

I checked the behavior and errors are emitted correctly.

I did do an FES mocking in `p5.Vector.js` unit tests like this:

```js
    globalThis.FESCalled = false;
    globalThis.p5 = {
      _friendlyError: function(msg, func) {
        globalThis.FESCalled = true;
        console.warn(msg);
      }
    };
```

Though I don't think it's related.

test/unit/math/p5.Vector.js
```diff
@@ -101,9 +150,9 @@ suite('p5.Vector', function () {
     });
   });
 
-  suite('new p5.Vector(1,2,undefined)', function () {
+  suite('new p5.Vector(1,2)', function () {
     beforeEach(function () {
-      v = new mockP5.Vector(1, 2, undefined);
+      v = new Vector(1, 2);
```
ksen0, line 156: @limzykenneth another substantive test case change; fails because I've added number validation to `Vector` based on my understanding of our earlier discussion

test/unit/math/p5.Vector.js
```diff
@@ -1885,137 +2000,139 @@ suite('p5.Vector', function () {
...
 
   suite('set values', function () {
     beforeEach(function () {
-      v = new mockP5.Vector();
+      v = new Vector();
     });
 
-    test('should set values to [0,0,0] if values array is empty', function () {
+    test('should NOT set values to [0,0,0] if values array is empty', function () {
       v.values = [];
```
ksen0, line 2076: This is a substantive change reflecting intent of the update

test/unit/math/p5.Vector.js
```diff
@@ -1885,137 +2000,139 @@ suite('p5.Vector', function () {
...
     });
 
-    test.fails(
-      'should throw friendly error if attempting to get element outside lenght',
+    test('should throw friendly error if attempting to get element outside length',
       function () {
-        let vect = new mockP5.Vector(1, 2, 3, 4);
-        assert.equal(vect.getValue(5), 1);
+        let vect = new Vector(1, 2, 3, 4);
+        globalThis.FESCalled = false;
```
ksen0: @limzykenneth this and another `fails` test stopped passing when I mocked FES, I think maybe they were not failing for the right reason originally? Anyway, here's a proposed update to reflect my understanding of the intent of the setValue/getValue tests

test/unit/webgl/p5.Shader.js
davepagurek: Are those getting reset correctly when the tests finish?

davepagurek: as a quick test, if you skip all those vector tests, do you see the strands tests start to pass?

ksen0: @davepagurek hmmm good question, no must be the code. Same errors with:

```js
          include: [
            //'./test/unit/math/p5.Vector.js',
            './test/unit/webgl/*.js',
          ],
```

test/unit/math/p5.Vector.js
limzykenneth: I'm not getting a fail here specifically but the fail I see locally are when the `p5.Vector` class is trying to use `p5._friendlyError`. It does not necessarily have a reference to `p5` to it should not really use that, I need to complete the FES rewrite stuff I'm still working on but in this kind of case, it can use a regular console.log instead. There isn't anything special this is doing.

limzykenneth:
I believe the CI test is failing on only the use of `p5._friendlyError` and not for p5.strands, so if CI all pass and @davepagurek can confirm whether the test pass overall for him locally too, it should be ok.

davepagurek:
I made some updates to inject a reference to `p5` into decorators if they need it, and also did some more dependency injection for the `Vector` class where a stub `_friendlyError` gets replaced with the real one when the addon function is called. @limzykenneth let me know what you think of the approach I took, it's definitely a little sloppy, and there are plenty more places in the codebase that also assume a global `p5` to exist and be accessible that we'd maybe also want to change once we've got a settled pattern. Calling a stub and replacing it sorta works but feels like maybe something better could exist.

limzykenneth:
@davepagurek If possible, I'd like to avoid adding reference to `p5` to the `decorateClass` function as that is supposed to be generic, ie. it should be usable for any classes that is not `p5`, for example exported as a utility for someone wanting to implement runtime decoration. The decorator function returned from say `_validatedVectorOperation` should also be a TC39 decorator function if possible so that if someone else has a decorator function fitting that standard, they can drop it in and use it directly without changes, the reference to `p5` in there would be a problem.

If we shift the `_friendlyError` function to be on the instance instead of as a static property, we can access it through `this` in the decorator function and not need to pass `p5` in. 

The overall approach to printing messages will change in the upcoming FES stuff I'm working through though if I manage to get the idea working, it will be more granular with console styling and translations built in.

davepagurek:
makes sense, pushed out an update!

davepagurek:
lgtm @ksen0 if you want to merge!

PR #8677, opened 2026-03-25
fix: [2.x] Fix dimensional quirk in the `setHeading()` method of 

### LINKED ISSUE #8215: [2.x] Fix dimensional quirk in the `setHeading()` method of `p5.Vector`

localhost-legend:
Summary

      - **src/math/p5.Vector.js**: Added input validation for the heading parameter and maintained existing functionality

      What changed
      Modify the `setHeading()` method to support only 2D vectors and add error messages for vectors of other dimensions.

       Testing
      I tested this locally and the changes work as expected. Happy to make any adjustments if needed!

      Closes #8215


localhost-legend:
Thank you for the warm welcome and for providing the contributor guidelines and resources. I'll make sure to link this PR to an approved issue going forward, and I'll review the discussion to ensure I'm not duplicating effort. I'll also add myself to the list of contributors once this PR is merged, thanks for the reminder.

limzykenneth:
The issue has not been approved for fix and the fix here does not address the issue, where it essentially deleted a whole file. Closing for now. Please avoid opening PR before working out the actual problem and solution in the issue.

PR #8695, opened 2026-04-01
Make instanceID() work in both vertex and fragment shaders

### LINKED ISSUE #8691: Make instanceID() work in both vertex and fragment shaders
### Increasing access

Currently, `instanceID()` only works in p5.strands vertex shader hooks because the GLSL `gl_instanceID` is only available globally in vertex shaders. However, the distinction between vertex and fragment shaders is intentionally not made upfront, so it can be confusing when you try to use `instanceID()` in, for example, a `finalColor` block and it doesn't work. Arguably we should do some work under the hood to *make* that work to add some more scaffolding to our teaching of vertex/fragment shaders, letting their explanation be deferred further without unblocking learners of, for example, compute shaders, who already have a lot of new things to be focusing on.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

Basically, to make this work, we'd have to notice when `instanceID()` has been called in fragment shader hooks, and bridge its value from the vertex shader automatically via a varying variable. This would need to happen in a way that works for both GLSL and WGSL.

aashu2006:
Resolves #8691 

Changes:
- Made instanceID() work in fragment hooks (like finalColor)
- Passes the instance ID from vertex to fragment using a flat varying
- Works for both WebGL and WebGPU
- Fixed WGSL type issue and multiple return handling
- also added an visual test to confirm it works

Screenshots of the change:
- Instances now render with different colors when using `instanceID()` in fragment shaders.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-p5.js-2.0-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/1-p5.js-2.0-bug-report.yml
@@ -1,18 +1,16 @@
-name: 📃 p5.js 2.0 Beta Bug Report
-description: This template is for submitting a bug report for bugs found in the p5.js 2.0 beta releases.
-title: "[p5.js 2.0 Beta Bug Report]: "
-labels: [p5.js 2.0]
+name: 📃 p5.js 2.0+ Bug Report
+description: This template is for submitting a bug report for bugs found in p5v2!
+title: "[p5.js 2.0+ Bug Report]: "
+labels: [p5.js 2.0+]
 body:
   - type: markdown
     attributes:
       value: |
         ### What falls under this category?
-        There has been many changes to p5.js in 2.0 that is currently released as beta versions. If you suspect there may be a bug, please follow the below steps before opening a bug report using this template:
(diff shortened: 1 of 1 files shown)


aashu2006:
@davepagurek would love a quick review when you have time 🙂

davepagurek:
Thanks for working on this @aashu2006! I'm going to take a closer look tonight but in the mean time, can we add testing for the webgpu implementation similar to what you've done for webgl?

aashu2006:
Got it, I’ll add a similar test for the WebGPU path as well 👍

aashu2006:
@davepagurek  added the WebGPU visual test too.

davepagurek (approved):
Looks great, thanks!

PR #8696, opened 2026-04-02
Auto-spread large WebGPU compute dispatches

### LINKED ISSUE #8690: Improve how WebGPU compute shaders spread iterations across threads
### Increasing access

Currently, if you make a compute shader that runs, say, one million iterations, and you do `compute(myShader, 1000000)`, it will likely still be laggy because it is not fully utilizing the GPU. If you do `compute(myShader, 1000, 1000)`, spreading it across both x and y, it generally runs a lot better. (There is also a z axis we can spread across, too!) This is somewhat unintuitive, and requires you to manually compute the array index you want in your compute shader from your x and y (and maybe z) coordinates. I've done this 1D-problem-spread-across-2D technique here: https://editor.p5js.org/davepagurek/sketches/n1q_WYNrE

Ideally, this is something we can do under the hood for you so that you can focus on the core pedagogical problem of compute shaders, which is thinking about your loop in parallel.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

We currently somewhat arbitrarily pick the number of work groups to run the compute shader on. We can do something a little smarter there, maybe trying to make it as square or cubelike as possible.

The main requirement is that we also need to present the shader hook with an index **in the original coordinates**, not our modified one. So if the user calls `compute(myShader, 1000000)`, even if we convert this into a 1000x1000 square, the `index` we give to the shader should be a vector of size `[1000000, 1, 1]`, not `[1000, 1000, 1]`.

aashu2006:
Resolves #8690

## Changes
Auto spreads large compute dispatches across 2D/3D to stay within GPU limits

## Problem
`compute(shader, 1000000)` would fail (GPU limit), users had to manually spread and handle indices

## Solution
- auto spreads large dispatches (square <1M, cube >1M)
- users still see simple 1D indices
- fixed strands void return bugs

## Modified files
- `p5.RendererWebGPU.js`: Spreading logic
- `compute.js`: Index reconstruct
- Strands files: for void return handling

## Testing
Tested with 1M+ particle simulations:
- Before: `compute(shader, 1000000)` - black screen (failed)
- After: `compute(shader, 1000000)` - gives 60 fps
- Edge case: `compute(shader, 1000003)` - gives 48 fps

## Screenshots
image
image

#### PR Checklist
- [ ] `npm run lint` passes


src/webgpu/p5.RendererWebGPU.js
```diff
@@ -3813,10 +3813,45 @@ ${hookUniformFields}}
...
+      let px = x;
+      let py = y;
+      let pz = z;
+
+      // we spread if we exceed GPU limits OR if it involves a large 1D dispatch
+      const exceedsLimits = x > MAX_THREADS_PER_DIM || y > MAX_THREADS_PER_DIM || z > MAX_THREADS_PER_DIM;
+      const isLarge1D = totalIterations > 1024 && y === 1 && z === 1;
+
+      if (exceedsLimits || isLarge1D) {
+        if (totalIterations > 1000000) {
```
davepagurek: Out of curiosity is there any benefit to spreading across dimensions like this for lower iteration counts too? e.g. if you're doing a big for loop inside of each iteration, with a smaller number of iterations, is there any difference?

src/strands/strands_api.js
```diff
@@ -219,7 +219,7 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
     const nodeData = DAG.createNodeData({
       nodeType: NodeType.STATEMENT,
       statementType: StatementType.EARLY_RETURN,
-      dependsOn: [valueNode.id]
+      dependsOn: value !== undefined ? [valueNode.id] : []
```
davepagurek, line 222: Mind elaborating on what these changes are there to handle? Anything we should have more test cases for in the tests?

src/webgpu/p5.RendererWebGPU.js
aashu2006: Good question! Currently I only auto-spread when count > 1024 to avoid overhead for small dispatches. For lower counts with heavy per iteration work, manual spreading might still help but I kept it simple for now. We could test this if you think it's worth optimizing?

src/strands/strands_api.js
aashu2006: These fix void return types in compute shaders. Without them, doing `return;` in a compute hook would crash with "Missing dataType". Most compute shaders use void (side-effects only), so the auto-spread wouldn't work without this fix.

For tests - should I add cases for void hooks with early returns? The main compute functionality already has test coverage.

davepagurek: ah, got it. Right, let's add a test for early returns, since this wasn't a case covered by any tests before. Thanks!

aashu2006: I've added the test cases for void compute hooks with early returns. Both tests are passing. 
Thanks!

src/webgpu/p5.RendererWebGPU.js
davepagurek: I think it's worth testing at least to know what kind of difference it makes, and similarly if it's better to spread across 3 dimensions earlier too. A sort of table of performance tests would help us just be a bit more confident about our optimizations.

aashu2006: I can run some quick tests comparing different spreading approaches across small, medium, and large counts. I’ll check 1D, 2D (square/rectangular), and 3D, and share a simple performance table with the results. Should be interesting to see where things start to slow down.

let me know if there’s anything specific you’d like me to test, or if you want to try something on your machine as well 👍

aashu2006:
@davepagurek I've updated the auto spreading strategy based on benchmark results. Now it uses 2D square for all cases instead of switching to 3D cube at >1M iterations.

Benchmarks showed 2D square equals or outperforms 3D cube at every scale tested, with simpler shader code. Full results: [Benchmark](https://gist.github.com/aashu2006/9c41556c6c38ee0644c765a37a816367)

Issue #8667, opened 2026-03-22
[2.0] [p5.strands] Add texCoord to finalColor block


nbogie:
### Increasing access

Newcomers to shaders would be able to make simple unlit material shaders sooner, with _only_ [finalColor](https://beta.p5js.org/reference/p5/finalColor/) (and, say, noise() on `texCoord`).
In contrast, currently, I think they need to understand how to use shared variables to propagate the `texCoord` from [pixelInputs](https://beta.p5js.org/reference/p5/pixelInputs/), and need to know that it even exists there.

### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [x] Other (specify if possible)

### Feature enhancement details

Instead of this for a material shader:

### before
```js
let coord = sharedVec2()
pixelInputs.begin();    
coord = pixelInputs.texCoord;
pixelInputs.end();

finalColor.begin();
let coord = finalColor.texCoord;
finalColor.set(noise(coord.x, coord.y));
finalColor.end();
```
...we could instead write this:

### after
```js
finalColor.begin();
let coord = finalColor.texCoord;
finalColor.set(noise(coord.x, coord.y));
finalColor.end();
```

This assumes there is no situation where it texCoord for finalColor would be meaningless or ambiguous - I don't know.

### Conversation in discord: 

https://discord.com/channels/836700474425475088/1384610702467272785/1485021371607154749

YuktiNandwana:
Hi @nbogie and @davepagurek! I've been exploring the new p5.strands architecture for my GSoC proposal and I'd love to take this on. Making shaders more accessible for beginners aligns perfectly with what I want to contribute to p5.js.

I'm currently looking into the WebGL module to see where the default properties for the finalColor block are defined. Could I be assigned to this issue?

davepagurek:
Hi @YuktiNandwana! Sure, I can assign it to you. In p5.strands, these hook blocks start at the shader level. Currently, a base shader does two things:

(1) it calls a hook function somewhere in its source code, like this one for the final color:
[...]

PR #8698, opened 2026-04-02
WIP: feat(webgl): Add texCoord parameter to getFinalColor shader hook

YuktiNandwana:
### Description
This draft PR implements the changes required to pass texture coordinates (`texCoord`) to the `getFinalColor` hook in WebGL, as discussed in Issue #8667. This makes it easier for users to create unlit material shaders using noise or textures directly within the `finalColor` block.

**Changes made:**
* **Fragment Shaders:** Updated `basic.frag`, `line.frag`, `normal.frag`, `phong.frag`, and `point.frag` to declare `IN vec2 vTexCoord;` and pass it as the second argument to `HOOK_getFinalColor()`.
* **Vertex Shaders:** Updated `line.vert`, `point.vert`, `normal.vert`, and `vertexColor.vert` to properly declare and provide `vTexCoord` to the fragment shaders. This ensures that the shaders compile correctly and that attributes are not missing.
* **Renderer GL:** Updated the default hook strings in `p5.RendererGL.js` to accept the new parameter format: `(vec4 color, vec2 texCoord)`.
* **Documentation:** Updated JSDoc example strings in `src/webgl/material.js` and `src/webgl/p5.Shader.js`.

### Current Status
* **Unit Tests:** All 1848 unit tests are now **passing** locally after fixing the initial shader compilation errors.
* **Next Steps:** I am looking for feedback on the core implementation. Once the architecture looks good, I will add specific visual test cases to verify the new `texCoord` property in action within `p5.strands`.

cc: @davepagurek - The implementation is now stable and tests are green. Would love a review whenever you have time!

### DIFF
--- a/src/webgl/material.js
+++ b/src/webgl/material.js
@@ -1081,7 +1081,7 @@ p5.prototype.shader = function (s) {
  *       myNormal = inputs.normal;
  *       return inputs;
  *     }`,
- *     'vec4 getFinalColor': `(vec4 color) {
+ *     'vec4 getFinalColor': `(vec4 color, vec2 texCoord) {
  *       return mix(
  *         vec4(1.0, 1.0, 1.0, 1.0),
  *         color,
@@ -1260,7 +1260,7 @@ p5.prototype.baseMaterialShader = function() {
  *   createCanvas(200, 200, WEBGL);
  *   myShader = baseNormalShader().modify({
  *     'vec3 getWorldNormal': '(vec3 normal) { return abs(normal); }',
- *     'vec4 getFinalColor': `(vec4 color) {
+ *     'vec4 getFinalColor': `(vec4 color, vec2 texCoord) {
  *       // Map the r, g, and b values of the old normal to new colors
  *       // instead of just red, green, and blue:
  *       vec3 newColor =
@@ -1466,7 +1466,7 @@ p5.prototype.baseColorShader = function() {
  * </td></tr>
  * <tr><td>
  *
- * `vec4 getFinalColor`
+ * 'vec4 getFinalColor': '(vec4 color, vec2 texCoord) {
  *
  * </td><td>
  *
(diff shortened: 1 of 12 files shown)


davepagurek:
Thanks for working on this so far! So I think your approach is just about correct. However, the development of p5.strands is just in p5.js 2.x, which lives on the `dev-2.0` branch, and p5 1.x, which is on `main` until this summer when we make 2.x the default, is just receiving bug fixes. Unfortunately it's pretty hard to drop the changes onto the `dev-2.0` branch, so you may need to start from there and manually replay your changes. In the dev-2.0 branch, you'll also see that all the `buildMaterialShader` etc methods exist, and `RendererWebGPU` exist, where they don't exist on the `main` branch. (In 1.x, you could only modify base shaders in GLSL, not javascript.)

Could I get you to make probably a separate PR into `dev-2.0` and close this one? I think your approach will be basically the same over there. I do have some other minor comments:
- Thanks for noticing that the line shader doesn't really have texture coordinates -- just using a zero vector makes sense there. For efficiency we can leave the vertex shader unmodified though, and just pass `vec2(0.0, 0.0)` in as a parameter directly in the fragment shader.
- We like to use visual tests to help ensure features don't break in the future. Take a look at `test/unit/visual/cases/webgl.js` (and `webgpu.js` in the same folder) to see some examples of those. We should likely make some tests for this hook too.
- In `dev-2.0`, you'll also find the WebGPU and 2D filter runner, which will need similar changes to the ones you've made in the WebGL renderer here.

YuktiNandwana:
Hi @davepagurek, thanks for the review and for catching the branch mix-up! Ah, that makes perfect sense—I will switch over to the dev-2.0 branch.

I also noted your optimization for the line shader (passing vec2(0.0, 0.0) directly in the fragment shader instead of touching the vertex shader) to save efficiency. I will implement that, apply the equivalent changes to WebGPU/2D filter runners, and add the visual tests in a fresh PR targeting dev-2.0.

Closing this one. I'll ping you on the new PR soon!

PR #8726, opened 2026-04-16
feat(webgpu): add read() to p5.StorageBuffer with tests

### LINKED ISSUE #8722: Add a way to read data from a WebGPU storage buffer
### Increasing access

One of the benefits of WebGPU compute shaders is that it can be a faster way to do things like simulations or calculations. Often that's paired with rendering, but sometimes it'd be helpful to be able to get the raw numbers out as a result. Currently, there's not a good way to do that, so we're limiting the use of compute shaders to rendering applications only.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

`p5.StorageBuffer` could get a method like `.get()` or maybe `.read()` that you'd have to `await` and would pull its data down from the GPU and unpack it into an array of the same format that it was initialized with.

aashu2006:
Resolves #8722

### Changes:
I’ve added an async `.read()` method to `p5.StorageBuffer` so we can read data back from GPU buffers into JavaScript.

- It uses a staging buffer + `mapAsync` to copy data from GPU -> CPU
- For number buffers, it returns a `Float32Array`
- For struct buffers, it returns plain JS objects based on the schema
- I also added a private `_unpackStructArray()` helper to rebuild struct data from the packed buffer
- This follows the same pattern already used in `readFramebufferPixels`

### Tests
I've added unit tests for:
  1. reading float buffers
  2. reading struct buffers
  3. reading after `.update()`
  4. reading after GPU compute verifiying actual GPU -> CPU flow
  5. struct buffers with vector fields

### Some Notes
- Struct data is returned as plain objects for e.g. arrays for vector fields, not `p5.Vector` or `p5.Color`
- `.read()` introduces a GPU - CPU sync point, so frequent use maybe can impact the performance

#### PR Checklist
- [x] `npm run lint` passes


aashu2006:
@davepagurek  Hey, I’ve added the `.read()` implementation and tests for this, have a quick look at it whenever you have time 🙂

src/webgpu/p5.RendererWebGPU.js
```diff
@@ -173,6 +173,124 @@ function rendererWebGPU(p5, fn) {
...
+
+      if (this._schema !== null) {
+        return this._unpackStructArray(rawCopy, this._schema);
+      }
+      return rawCopy;
+    }
+
+    // Inverse of _packStructArray reads packed buffer back into plain JS objects
+    // using the same schema layout - fields, stride and offsets
+    _unpackStructArray(floatView, schema) {
```
davepagurek: Do you think we can move this to the renderer next to `_packStructArray` for consistency, and defer to the renderer from this class?

test/unit/webgpu/p5.RendererWebGPU.js
```diff
@@ -160,4 +160,89 @@ suite('WebGPU p5.RendererWebGPU', function() {
...
+        { position: myp5.createVector(1, 2), speed: 5.0 },
+        { position: myp5.createVector(3, 4), speed: 10.0 },
+      ];
+      const buf = myp5.createStorage(input);
+
+      const result = await buf.read();
+
+      expect(result).to.be.an('array');
+      expect(result.length).to.equal(2);
+      // Vector fields come back as plain arrays
```
davepagurek, line 218: ahh good catch. I guess we don't currently capture in the schema if these values came from arrays, vectors, or colors. I think we might want to record that so we can deserialize into the same format it came in as. Do you think something like that is feasible?

davepagurek:
this is looking good, and thanks for the detailed tests!

src/webgpu/p5.RendererWebGPU.js
aashu2006: Yes, that makes sense. I will move `_unpackStructArray` next to `_packStructArray` in the renderer and call it from here for consistency.

test/unit/webgpu/p5.RendererWebGPU.js
aashu2006: yeah that’s a good point, right now we’re just returning plain data, but preserving the original types like vectors or colors would be useful.

I think it’s doable by extending the schema to store the original input type, but it might add some complexity. I can look into that as a follow-up issue if that sounds good.

aashu2006:
@davepagurek I've pushed the changes moved `_unpackStructArray` to the renderer and updated usage. Happy to make any other tweaks if needed!

test/unit/webgpu/p5.RendererWebGPU.js
davepagurek: I think it might be worth exploring upfront, just because it changes the API a bit -- e.g. if we were to release just this then change the structure of return values, it could break users' code. (I don't *think* we're going to do a release in between PRs, but that's generally how I think about API changes -- we generally want to avoid, if we can, merging code with an API that we think we'll change.)

aashu2006: If we think the return format might change, it’s better to settle that before merging. Would you prefer that we preserve original types like vectors or colors in this PR, or keep the current plain data approach and commit to that API for now?

aashu2006:
@davepagurek Just wanted to follow up on the return format question, happy to proceed either way, just wanted to check what direction you'd prefer here

test/unit/webgpu/p5.RendererWebGPU.js
davepagurek: Let's see what preserving the original types involves. If it requires a lot of changes let me know and we can talk about whether it might be better to omit it, but if we can make it work, it feels like that'd be the preferable user experience.

aashu2006: Got it, I’ll explore preserving the original types and see how much it impacts the current schema logic. I’ll try to keep it minimal and share an update.

aashu2006:
@davepagurek I've now pushed an update to preserve the original types. For colors, I’m now scaling back using `colorMaxes` to match how `_getRGBA` normalizes values, so it works correctly with different `colorMode` settings. Let me know what you think!

aashu2006:
@davepagurek just a ping!

src/webgpu/p5.RendererWebGPU.js
```diff
@@ -3398,6 +3357,65 @@ ${hookUniformFields}}
...
+            if (n === 1) {
+              item[field.name] = floatView[idx];
+            } else {
+              const values = Array.from(floatView.slice(idx, idx + n));
+              if (field.kind === 'vector') {
+                item[field.name] = this._pInst.createVector(...values);
+              } else if (field.kind === 'color') {
+                // Color was packed as normalized RGBA [0-1] via _getRGBA([1,1,1,1])
+                // Scale back to the current colorMode range
+                const maxes = this.states.colorMaxes[this.states.colorMode];
```
davepagurek, line 3404: nice!

davepagurek (approved):
thanks for the updates @aashu2006, this looks solid!

PR #8730, opened 2026-04-18
implementing random function for strands

perminder-17:


perminder-17:
Hi @davepagurek and @ksen0 , sorry for the delay, its open for the review, Can you please check once?

src/strands/strands_api.js
```diff
@@ -382,6 +384,89 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
...
+        DataType.float1,
+        userSeed !== null
+          ? () => userSeed
+          : () => performance.now(),
+      );
+    }
+
+    const callIndex = strandsContext._randomCallCount++;
+
+    const nodeArgs = [seedNode, callIndex];
```
davepagurek: I think that means that if you were to call `random()` in a for loop, e.g.:
```js
let data = uniformStorage(someStorage)

for (let i = 0; i < 5; i++) {
  data[i] += random()
}
```
...then they'd all get the same random value added to them because `random` here is just one AST node and would have one fixed call index, translating into something like:

```wgsl
for (var i = 0; i < 5; i++) {
  data[i] += random(seed, 0);
}
```

Do we have the option of making the call index be a global variable in the shader code so that it's not fixed? e.g something like:

```wgsl
var<private> invocation: i32 = 0;
fn random(seed: f32) -> f32 {
  let val = ...; // Do random gen here, using `invocation`
  invocation++;
  return val;
}

// So then the call site is just:
for (var i = 0; i < 5; i++) {
  data[i] += random(seed);
}
```

src/webgpu/p5.RendererWebGPU.js
```diff
@@ -2557,10 +2558,86 @@ function rendererWebGPU(p5, fn) {
...
+        return `${params.slice(0, -1)}${hasParams ? ', ' : ''}${additionalParams.join(', ')})`;
+      };
+
+      let hookExtraParams = [];
+      let hookExtraArgs = [];
+
       if (shaderType === 'vertex') {
-        if (!main.match(/\@builtin\s*\(\s*instance_index\s*\)/)) {
-          main = main.replace(/\)\s*(->|\{)/, ', @builtin(instance_index) instanceID: u32) $1');
+        const ensuredInstance = ensureBuiltinParam(main, 'instance_index', 'instanceID', 'u32');
```
davepagurek, line 2677: nice work generalizing this!

davepagurek (commented):
Thanks for taking this on, the structure is looking good! I left one thing about possibly making the call index a runtime thing instead of compile time to better support for loops, and maybe we could add a test case for vertex + fragment shaders too?

src/webgpu/p5.RendererWebGPU.js
ksen0: oh wow yes, nice!

davepagurek (approved):
Nice, this looks good! Great work @perminder-17!

PR #8753, opened 2026-04-28
support map() inside p5.strands shaders

### LINKED ISSUE #8750: support map() inside p5.strands shaders
right now p5.strands doesn't support the regular p5 map() function inside shader callbacks. since map is one of the most used helpers in normal p5 sketches, not having it in strands means people end up rewriting it manually as something like start2 + (n - start1) / (stop1 - start1) * (stop2 - start2) every time.

would be nice to add it to the strands builtins so it just works the same way it does in regular p5 code.

something like:
map(n, start1, stop1, start2, stop2)

with the same signature as the cpu version. the withinBounds variant can probably come later if needed.

came up in a recent call, opening this so we have a place to track it.

Nixxx19:
### Summary
- adds support for the regular p5 map() function inside p5.strands modify() callbacks, so it works the same way as in normal p5 sketches.
- when strands is not active, falls back to the original p5 map.
- internally rewrites map(n, s1, e1, s2, e2) to mix(s2, e2, (n - s1) / (e1 - s1)) using the existing strands math operators and the mix builtin.
- the optional withinBounds argument is handled with clamp(result, min(s2, e2), max(s2, e2)) so it works regardless of which bound is larger.

### Tests
- adds two unit tests in test/unit/webgl/p5.Shader.js covering the basic case and the withinBounds case.
- both pass locally with npx vitest.

### PR Checklist
- [x] npm run lint passes
- [ ] Inline reference is included / updated (not applicable, no new public api)


src/strands/strands_api.js
```diff
@@ -294,6 +294,23 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
...
+  augmentFn(fn, p5, 'map', function (...args) {
+    if (!strandsContext.active) {
+      return originalMap.apply(this, args);
+    }
+    const [n, start1, stop1, start2, stop2, withinBounds] = args;
+    const nNode = p5.strandsNode(n);
+    const start1Node = p5.strandsNode(start1);
+    const stop1Node = p5.strandsNode(stop1);
+    const t = nNode.sub(start1Node).div(stop1Node.sub(start1Node));
+    const result = fn.mix(start2, stop2, t);
```
davepagurek: I think this doesn't make a difference here since this part of the code happens just while the strands context isn't active, but maybe let's use `this.` instead of `fn.` to better match expected usage, since the intention is for these functions to not be called statically.

davepagurek (commented):
Thanks, this is looking good! Do you think we could add some unit tests that map from one range to a different range to test the math? Possibly a visual test that maps the texture coordinate to something?

src/strands/strands_api.js
Nixxx19: thanks for the review @davepagurek!

swapped fn. to this. for mix/clamp/min/max in strands_api.js. you're right that it doesn't change behavior here but it's nicer to keep consistent with the rest of the file (like getTexture).

also added a few more tests in p5.Shader.js since the original two were both identity mappings (0..1 to 0..1), so they weren't really exercising the math:
- range shrink: map(5, 0, 10, 0, 1) -> 0.5
- offset output: map(0.5, 0, 1, 0.2, 0.8) -> 0.5 mapped to 128
- negative input range: map(0, -1, 1, 0, 1) -> 0.5
- a horizontal gradient test that pipes inputs.texCoord.x through map() and checks left/middle/right pixels (covers the visual side you mentioned)

all 6 pass locally with vitest. lmk if you want anything else changed!

PR #8768, opened 2026-05-06
Fix #8756: Allow computed array indexing on non-storage vectors

### LINKED ISSUE #8756: [p5.js 2.0+ Bug Report]: `get()` can only be used on storage buffers

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] WebGPU
- [x] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.2.3

### Web browser and version

Chrome 147.0.7727.137

### Operating system

MacOSX 26.4.1

### Steps:
1. Create a WebGPU sketch that uses `buildComputeShader()`.
2. Inside the compute callback, define a helper function that returns an array, for example `return [1, 2]`.
3. Read from that returned value using computed indexing like `arr[0]`.
4. Call `buildComputeShader(errorCallback)` and then try `errorShader.inspectHooks()`.

### Snippet:

```js
let shader;
let errorShader;
let storage;
const COUNT = 10;

async function setup() {
  await createCanvas(400, 400, WEBGPU);
  noLoop();
  background(100);
  storage = createStorage(new Float32Array(COUNT));

  shader = buildComputeShader(callback);
  shader.inspectHooks();
  console.log("======Compute shader source:======");
  console.log(shader.computeSrc());
  console.log("================================");
  
  errorShader = buildComputeShader(errorCallback);
  errorShader.inspectHooks();
  console.log("======Error shader source:======");
  console.log(errorShader.computeSrc());
  console.log("================================");
}

function errorCallback() {
  let data = uniformStorage(storage);
  function getArray() {
    return [1, 2];
  }
  let arr = getArray();
  data[index.x] = arr[0] + arr[1] + index.x;
}

function callback() {
  let data = uniformStorage(storage);
  function getNumber() {
    return float(index.x);
  }
  let num = getNumber();
  data[index.x] = num;
}

async function draw() {
  compute(shader, COUNT);
  const data = await storage.read();
  console.log(data);
}
```

### Console error:
`Uncaught (in promise) Error: get() can only be used on storage buffers`

### Expected behavior:
`buildComputeShader(errorCallback)` should either:
- support indexing into a helper-returned array/vector-like value, or
- fail with a clearer compile/transpile error explaining that this pattern is unsupported.

It should also ideally still be possible to inspect the generated hooks/source for debugging.

### Actual behavior:
`buildComputeShader(errorCallback)` throws before `errorShader.inspectHooks()` can run, so the shader is never returned and its generated source cannot be inspected.
The scalar-return version works:
```js
function getNumber() {
  return float(index.x);
}
```
but the array-return version fails when indexing the returned value with `arr[0]`.

### Additional notes:
From local repo inspection, this looks like a `p5.strands` transpilation issue rather than a WGSL compile error:
- array literals are transformed into strands values
- computed indexing like `arr[0]` is rewritten into `.get(0)`
- `.get()` currently appears to only work for storage buffers

So the bug seems to be specifically that non-storage array/vector indexing is being routed through the storage-buffer access path.

SOUMITRO-SAHA:
Resolves #8756 

 Changes:

This PR fixes the bug where `get()` could only be used on storage buffers in `p5.strands`, preventing array indexing (`arr[0]`) on non-storage vectors like those returned from helper functions.

##### Root Cause: 
The transpiler converted ALL computed member access (`arr[0]`) to `.get()` calls, but the `get()` method in `strands_node.js` had a guard that only allowed it for storage buffers, throwing **"get() can only be used on storage buffers"** for non-storage vectors.

- `src/strands/strands_node.js`
- ` src/strands/strands_transpiler.js`
- `src/webgl/strands_glslBackend.js` (Added `ARRAY_ACCESS` handler, missing from GLSL backend - flagged by @Nixxx19 ).
- `test/unit/webgpu/p5.Shader.js`

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/strands/strands_transpiler.js
+++ b/src/strands/strands_transpiler.js
@@ -571,7 +571,8 @@ const ASTCallbacks = {
     }
 
     if (node.elements.length < 2 || node.elements.length > 4) {
-      throw new Error(
+      FES.userError(
+        'type error',
         `Array literals in shader functions are transpiled to vectors and must have 2-4 elements (got ${node.elements.length}).`
       );
     }
--- a/src/strands/strands_transpiler.js
+++ b/src/strands/strands_transpiler.js
@@ -569,6 +569,13 @@ const ASTCallbacks = {
     if (ancestors.some(a => nodeIsUniform(a) || nodeIsUniformCallbackFn(a, state.uniformCallbackNames))) {
       return;
     }
+
+    if (node.elements.length < 2 || node.elements.length > 4) {
+      throw new Error(
+        `Array literals in shader functions are transpiled to vectors and must have 2-4 elements (got ${node.elements.length}).`
+      );
+    }
+
     const original = JSON.parse(JSON.stringify(node));
     node.type = 'CallExpression';
     node.callee = {
@@ -1240,8 +1247,8 @@ const ASTCallbacks = {
     delete node.update;
   },
 
-  
-  
+
+
 }
 
 // Helper function to check if a function body contains return statements in control flow
--- a/src/webgl/strands_glslBackend.js
+++ b/src/webgl/strands_glslBackend.js
@@ -330,6 +330,12 @@ export const glslBackend = {
         const parentExpr = this.generateExpression(generationContext, dag, parentID);
         return `${parentExpr}.${node.swizzle}`;
       }
+      if (node.opCode === OpCode.Binary.ARRAY_ACCESS) {
+        const [bufferID, indexID] = node.dependsOn;
+        const bufferExpr = this.generateExpression(generationContext, dag, bufferID);
+        const indexExpr = this.generateExpression(generationContext, dag, indexID);
+        return `${bufferExpr}[${indexExpr}]`;
+      }
       if (node.dependsOn.length === 2) {
         const [lID, rID] = node.dependsOn;
         const left  = this.generateExpression(generationContext, dag, lID);
--- a/src/strands/strands_node.js
+++ b/src/strands/strands_node.js
@@ -166,18 +166,15 @@ export class StrandsNode {
   }
 
   get(index) {
-    // Validate baseType is 'storage'
     const nodeData = getNodeDataFromID(this.strandsContext.dag, this.id);
-    if (nodeData.baseType !== 'storage') {
-      throw new Error('get() can only be used on storage buffers');
-    }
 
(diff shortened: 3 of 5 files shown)


src/strands/strands_transpiler.js
```diff
@@ -569,6 +569,13 @@ const ASTCallbacks = {
     if (ancestors.some(a => nodeIsUniform(a) || nodeIsUniformCallbackFn(a, state.uniformCallbackNames))) {
       return;
     }
+
+    if (node.elements.length < 2 || node.elements.length > 4) {
+      throw new Error(
```
davepagurek: Maybe we can use this helper for consistency? https://github.com/processing/p5.js/blob/d3bb21c4e8be438045ccac3f7b89233fc112251d/src/strands/strands_FES.js#L6

src/webgl/strands_glslBackend.js
```diff
@@ -330,6 +330,12 @@ export const glslBackend = {
         const parentExpr = this.generateExpression(generationContext, dag, parentID);
         return `${parentExpr}.${node.swizzle}`;
       }
+      if (node.opCode === OpCode.Binary.ARRAY_ACCESS) {
+        const [bufferID, indexID] = node.dependsOn;
```
davepagurek, line 334: Do you think we could add a test for the WebGL functionality too, in addition to the WebGPU tests?

davepagurek (commented):
Thanks for working on this, the code is looking good!

SOUMITRO-SAHA:
Thanks for the review, @davepagurek  Great suggestions — I'll address both:

**1. `userError` helper** — Good catch. I'll update the error in `strands_transpiler.js` to use `userError('type error', ...)` from `strands_FES.js` for consistency with other strands error messages.

**2. WebGL tests** — Absolutely.

I'll push the updates shortly.

SOUMITRO-SAHA:
Hi @davepagurek , I've pushed the changes addressing both of your review comments

davepagurek (approved):
Thanks for the updates, this looks good!

PR #8772, opened 2026-05-09
Add StorageBuffer.set(index, value) for single-element GPU updates

### LINKED ISSUE #8749: Add method to update a specific index in p5.StorageBuffer
### Increasing access

Unsure

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### Current behavior

`p5.StorageBuffer.update()` always rewrites the entire buffer from index 0. If a user wants to change a single element for example, one particle out of 10,000, they currently have to:

1. `await buf.read()` - full GPU→CPU sync
2. modify the element in the JS array
3. `buf.update(data)` - rewrite the whole buffer

This is slow and awkward for simple updates.

### Proposed feature

Add a `set(index, value)` method to `p5.StorageBuffer` that updates a single element at a given index without affecting the rest of the buffer.

```js
// Float buffer
let buf = createStorage(new Float32Array([1, 2, 3, 4]));
buf.set(2, 9.5); // only updates index 2

// Struct buffer
let particles = createStorage(makeParticles());
particles.set(42, { position: createVector(0, 0), velocity: createVector(1, 0) });
````
### Why this helps

WebGPU's `GPUQueue.writeBuffer()` natively supports a byte offset, so a partial write doesn’t need a readback, staging buffer, or GPU sync. The cost is proportional to one element instead of the whole buffer.

### Implementation notes

The byte offset for index `i` in a struct buffer is `i * schema.stride`. For a float buffer it's `i * 4`. The existing `_packField` logic already handles all field types, so packing a single element just reuses it with `baseOffset = 0`, then calls:

```js
device.queue.writeBuffer(this.buffer, byteOffset, packedElement);
```

**Scope for v1:**
- Full element replacement only (no partial field updates)
- Works for both float and struct buffers
- Bounds check with a friendly error

This complements the `.read()` method added recently.

Related to #8722 
This idea came up in discussion with @davepagurek after the `.read()` implementation.

aashu2006:
Resolves #8749

### Changes

I've added a `set(index, value)` method to `p5.StorageBuffer` that updates a single element in a buffer without rewriting the whole thing. 

Before this, updating one particle out of thousands required a full GPU -> CPU read, a JS edit, then a full CPU -> GPU write back. Now you can do:

```js
// float buffer
buf.set(2, 9.5);

// struct buffer
particles.set(42, { position: createVector(0, 0), velocity: createVector(1, 0) });
```

This works by using `GPUQueue.writeBuffer()` with a byte offset, so only one element's bytes are sent to the GPU.

**What changed:**
- `src/webgpu/p5.RendererWebGPU.js`
added set(index, value) to StorageBuffer
added JSDoc and examples
- `test/unit/webgpu/p5.RendererWebGPU.js`
added 7 unit tests for float buffers, struct buffers, bounds checks, and type validation

## PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/src/webgpu/p5.RendererWebGPU.js
+++ b/src/webgpu/p5.RendererWebGPU.js
@@ -246,6 +246,114 @@ function rendererWebGPU(p5, fn) {
       }
       return rawCopy;
     }
+
+    /**
+     * Updates a single element in the buffer at the given index without
+     * rewriting the entire buffer. This is more efficient than `update()`
+     * when only one element needs to change.
+     *
+     * Uses WebGPU's `GPUQueue.writeBuffer()` with a byte offset, so the
+     * cost is proportional to one element rather than the whole buffer.
+     *
+     * ```js example
+     * let buf;
+     *
+     * async function setup() {
+     *   await createCanvas(100, 100, WEBGPU);
+     *
+     *   // Float buffer: update one value by index
+     *   buf = createStorage(new Float32Array([1, 2, 3, 4]));
+     *   buf.set(2, 9.5); // only index 2 changes → [1, 2, 9.5, 4]
+     *
+     *   let result = await buf.read();
+     *   print(result[2]); // 9.5
+     *   describe('Prints 9.5 to the console.');
+     * }
+     * ```
+     *
+     * ```js example
+     * let particles;
+     * const numParticles = 100;
+     *
+     * async function setup() {
+     *   await createCanvas(100, 100, WEBGPU);
+     *   particles = createStorage(makeParticles());
+     *
+     *   // Struct buffer: replace particle 42 without touching others
+     *   particles.set(42, {
+     *     position: createVector(0, 0),
+     *     velocity: createVector(1, 0),
+     *   });
+     *   describe('Updates a single particle in a storage buffer.');
+     * }
+     *
+     * function makeParticles() {
+     *   let data = [];
+     *   for (let i = 0; i < numParticles; i++) {
+     *     data.push({
+     *       position: createVector(random(width), random(height)),
+     *       velocity: createVector(random(-1, 1), random(-1, 1)),
+     *     });
+     *   }
+     *   return data;
+     * }
+     * ```
+     *
+     * @method set
+     * @for p5.StorageBuffer
+     * @beta
+     * @webgpu
+     * @webgpuOnly
+     * @param {Number} index The zero-based index of the element to update.
+     * @param {Number|Object} value The new value. Pass a number for float
+     *   buffers, or a plain object matching the original struct layout for
+     *   struct buffers.
+     */
+    set(index, value) {
+      const device = this._renderer.device;
+
+      if (this._schema !== null) {
+        // buffer was created with an array of structs
+        if (typeof value !== 'object' || value === null || Array.isArray(value)) {
+          throw new Error(
+            'set() expects a plain object matching the original struct format for this buffer'
+          );
+        }
+
+        const { stride } = this._schema;
+        const byteOffset = index * stride;
+
+        if (byteOffset + stride > this.size) {
+          throw new Error(
+            `set() index ${index} is out of bounds for this buffer ` +
+            `(buffer holds ${Math.floor(this.size / stride)} elements)`
+          );
+        }
+
+        // pack just this one element using the same logic as update()
+        const packed = this._renderer._packStructArray([value], this._schema);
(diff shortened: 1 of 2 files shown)


aashu2006:
hii @davepagurek, thanks for waiting!

as there isn’t a way to update a single buffer element without re-uploading the whole buffer, it can get expensive for large cases like 10k particles.

since `GPUQueue.writeBuffer()` already supports byte offsets, I added a `set(index, value)` method that computes the element position and writes only that chunk of data.

For struct buffers the offset is `index * schema.stride`, and for float buffers it’s `index * 4`.

I reused the existing `_packStructArray` helper to pack a single struct element (as a one item array), then passed `packed.buffer` to `writeBuffer` so the size is handled correctly in bytes.

also added the type checks with clearer errors, plus unit tests for both float and struct buffers including checks that neighbouring elements stay unchanged.

let me know if you would like anything adjusted!

src/webgpu/p5.RendererWebGPU.js
```diff
@@ -246,6 +246,114 @@ function rendererWebGPU(p5, fn) {
       }
       return rawCopy;
     }
+
+    /**
+     * Updates a single element in the buffer at the given index without
+     * rewriting the entire buffer. This is more efficient than `update()`
+     * when only one element needs to change.
+     *
+     * Uses WebGPU's `GPUQueue.writeBuffer()` with a byte offset, so the
```
davepagurek: This is good info, but is probably more technical than is necessary for readers of the docs. Maybe we could rephrase to have messaging more like, use this if you are only updating small numbers of elements, and use `.update()` (and link to its docs) if you need to update all the items at once?

src/webgpu/p5.RendererWebGPU.js
```diff
@@ -246,6 +246,114 @@ function rendererWebGPU(p5, fn) {
...
+     *
+     * async function setup() {
+     *   await createCanvas(100, 100, WEBGPU);
+     *
+     *   // Float buffer: update one value by index
+     *   buf = createStorage(new Float32Array([1, 2, 3, 4]));
+     *   buf.set(2, 9.5); // only index 2 changes → [1, 2, 9.5, 4]
+     *
+     *   let result = await buf.read();
+     *   print(result[2]); // 9.5
```
davepagurek, line 267: In the reference pages, users won't be able to see printed results easily (especially if there are multiple examples shown on the page, it isn't super clear which console log comes from which example), and there will be a blank canvas next to the code. Maybe we could either take out the `example`  tag so this renders as just a code block without a canvas next to it, or possibly do something visual with the result rather than logging?

davepagurek (commented):
Thanks for working on this, the code is looking good! I just left a few comments about the description for the p5 reference.

src/webgpu/p5.RendererWebGPU.js
aashu2006: yup, I’ll simplify the wording to focus more on when to use `.set()` vs `.update()` instead of the underlying WebGPU details.

aashu2006:
@davepagurek I've fixed the wording a bit now also I have now removed the example tag, please let me know if any tweaks are needed. Thanks!

src/webgpu/p5.RendererWebGPU.js
aashu2006: for now I’ve removed the example tag so it renders as a regular code block instead of an runnable canvas example.

a proper visual WebGPU example could be nice here though, maybe as a separate docs/example improvement issue later on. What do you think?

davepagurek: I think that's fine. Also should we do something similar for the second example since that also doesn't render anything?

aashu2006: cool, added `read()` + `print()` to the second `set()` example.

davepagurek: Like the first one, since it doesn't render, should we take out the `example` tag?

aashu2006: sorry, updated!

davepagurek (approved):
Looks good, thanks for the updates!

Issue #8694, opened 2026-04-01
Clean up p5.strands transpiler code


davepagurek:
### Increasing access

In p5.strands, one of the steps in the process of turning your code into shader code is to transpile your JavaScript into slightly different JavaScript using internal p5.strands functions instead of, for example, math operators.

This code started much simpler, with a single pass mapping different Acorn syntax node types to different structure. Now, it involves multiple passes, sometimes building up some info first before doing actual transpilation. It is stretching the organizational limits of the file and could use some cleanup to be easier for contributors to dive into.

### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [ ] Other (specify if possible)

### Feature enhancement details

Ideally I'd like to see the core transpiler function be a *lot* simpler, deferring each pass to another function. That way we can document what it's doing at a high level in that outermost function in comments, and then gradually get more detailed as we dig into the different pass functions. There's also a number of repeated patterns that we can possibly extract and find some more structure in.

We have pretty good test coverage of the features of p5.strands, so we can feel reasonably certain that a refactor is successful if the tests pass.

Note that this is not a good issue for new contributors, as it deals with relatively advanced code, but help from more established contributors is appreciated!

LalitNarayanYadav:
Hi @davepagurek,

I’ve been going through the current `p5.strands` transpiler flow, and I agree that the increasing number of passes is making the core function harder to follow and reason about.

A possible direction for cleanup could be to restructure the transpilation into clearly defined stages:

####  Suggested Direction
- **Pipeline Decomposition**
  - Separate the process into stages such as:
    1. AST preprocessing / metadata collection  
    2. Transformation passes (operator replacement, function mapping, etc.)  
    3. Final code generation  
  - The top-level transpiler function could act as an orchestrator, with high-level comments explaining each stage.

- **Pass Isolation**
  - Implement each transformation as a **self-contained function**, making the flow easier to understand and maintain.

- **Shared Utilities**
[...]

PR #8782, opened 2026-05-11
Refactor: Extract transpilation passes and callback generation from transpileStrandsToJS

LalitNarayanYadav:
Addresses #8694

### Changes

- Extracted `runNonControlFlowPass`
- Extracted `runControlFlowPass`
- Extracted `buildStrandsCallback`
- Simplified `transpileStrandsToJS` into a staged orchestration pipeline

This improves readability by separating the major transpilation stages into named functions while preserving the existing transformation order and behavior.

### Screenshots of the change

N/A (non-visual refactor)

#### PR Checklist

- [ ] `npm run lint` passes

### DIFF
--- a/src/strands/strands_transpiler.js
+++ b/src/strands/strands_transpiler.js
@@ -1682,34 +1682,30 @@ function transformHelperFunctionEarlyReturns(ast, names) {
  * This staged approach ensures correct ordering and avoids transformation conflicts.
  */
 
-export function transpileStrandsToJS(p5, sourceString, srcLocations, scope) {
-  // Reset counters at the start of each transpilation
-  blockVarCounter = 0;
-  loopVarCounter = 0;
+// Wraps each callback with a uniform context guard, eliminating the need
+// to repeat the early-return check at the top of every handler.
+function makeGuardedCallbacks(callbacks) {
+  const guarded = {};
+  for (const [name, fn] of Object.entries(callbacks)) {
+    guarded[name] = (node, state, ancestors) => {
+      if (ancestors.some(a =>
+        nodeIsUniform(a) ||
+        nodeIsUniformCallbackFn(a, state.uniformCallbackNames)
+      )) return;
+      return fn(node, state, ancestors);
+    };
+  }
+  return guarded;
+}
 
-  const ast = parse(sourceString, {
-    ecmaVersion: 2021,
-    locations: srcLocations
-  });
-
-  throwIfLoopProtectionInserted(ast);
-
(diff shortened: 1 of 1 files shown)


src/strands/strands_transpiler.js
```diff
@@ -1785,15 +1774,7 @@ export function transpileStrandsToJS(p5, sourceString, srcLocations, scope) {
   }
   const body = match[2];
   try {
-    const internalStrandsCallback = new Function(
-        // Create a parameter called __p5, not just p5, because users of instance mode
```
davepagurek, line 1789: Can we put this comment back?

src/strands/strands_transpiler.js
```diff
@@ -1682,34 +1682,30 @@ function transformHelperFunctionEarlyReturns(ast, names) {
...
-  // First pass: transform .set() calls in control flow to use intermediate variables
-  transformSetCallsInControlFlow(ast, uniformCallbackNames);
+// Wraps each callback with a uniform context guard, eliminating the need
+// to repeat the early-return check at the top of every handler.
+function makeGuardedCallbacks(callbacks) {
+  const guarded = {};
+  for (const [name, fn] of Object.entries(callbacks)) {
+    guarded[name] = (node, state, ancestors) => {
+      if (ancestors.some(a =>
+        nodeIsUniform(a) ||
```
davepagurek, line 1692: I don't see removal of these checks from the different node handlers in the diff, does that mean we're double-handling this now? Also, did every node type handler do this before, or just some? e.g. are there any that didn't before that would now? I suspect we do want this everywhere, but it would be good to double-check where the behaviour would be changing to confirm that we want the change.

src/strands/strands_transpiler.js
```diff
@@ -1802,3 +1783,31 @@ export function transpileStrandsToJS(p5, sourceString, srcLocations, scope) {
...
+
+  const ast = parse(sourceString, { ecmaVersion: 2021, locations: srcLocations });
+
+  throwIfLoopProtectionInserted(ast);
+
+  // Pre-pass: collect names of functions passed by reference as uniform callbacks
+  const uniformCallbackNames = collectUniformCallbackNames(ast);
+
+  // Pass 1: transform .set() calls in control flow to use intermediate variables
+  transformSetCallsInControlFlow(ast, uniformCallbackNames);
```
davepagurek, line 1805: Nice, this is really easy to read!

davepagurek (approved):
Thanks @LalitNarayanYadav, looks great!

Issue #8210, opened 2025-10-27
[2.x] The `cross()` method of `p5.Vector` is incorrect in the 2.x context


GregStanton:
## The core problem: A 1.x feature became a 2.x bug

I'd like to propose a fix for `p5.Vector.cross()`. I know we've discussed this method before, and at the time, a change was deferred because it was considered a breaking change rather than a fix.

However, as I've been organizing the 2.0 stabilization work for `p5.Vector`, I've realized the context has fundamentally changed. The move from 1.x's 3D-padded vectors to 2.x's true n-dimensional vectors means the old 1.x behavior is no longer just confusing—it is now demonstrably incorrect and actively hides bugs. This new context promotes the issue from a feature request to a critical bug fix.

Here's a breakdown of the problem:

* **In 1.x (correct):** All vectors were 3D, and `createVector(1, 0)` was just a shortcut for `createVector(1, 0, 0)`. The `cross()` method *always* operated on two 3D vectors and returned a 3D vector. This was 100% correct.
* **In 2.x (bugged):** Now, `createVector(1, 0)` creates a **true 2D vector**. The current `cross()` implementation "fixes" this by *silently padding* this 2D vector back to 3D to perform the calculation.

This silent, implicit padding is the bug. It is incorrect for four reasons:

1. **It hides bugs that break sketches:** This is the biggest issue. In 2.0, mixing 2D and 3D vectors is almost certainly a programming error. The current behavior *hides this bug* instead of throwing a friendly error that helps users find it.  
2. **It's mathematically and programmatically non-standard:** The vector cross product is only traditionally defined for two 3D vectors. No other major library (three.js, Godot, Blender) behaves as p5 does currently. They all correctly recognize that the 2D "cross product" is a different operation that returns a scalar. Rather than opening doors, p5-specific rules create obstacles.  
3. **It's internally inconsistent:** This "pad-by-zero" logic contradicts the simple rules already in p5.js. For example, `mult(2)` uses broadcasting (repeating the value `2`), not padding. This "pad-by-zero" rule for `cross()` introduces a *second, conflicting* system for handling dimension mismatches that no longer applies in 2.x, which is a classic source of user confusion and a violation of the [Principle of Least Surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).  
[...]

PR #8783, opened 2026-05-11
Update vector docs to indicate recommended usage

ksen0:
Addresses https://github.com/processing/p5.js/issues/8151 and https://github.com/processing/p5.js/issues/8210 

* `array()` is marked for deprecation, and should not be used (`.values`) is recommended)
* `cross()` is defined only on three-dimensional vectors. To avoid breaking older code that relied on zero-padded vectors, no change in behavior is introduced to prevent use on vectors of unexpected dimensions. However, documentation has been updated.

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -2690,14 +2690,15 @@ class Vector {
    * Returns the vector's components as an array of numbers.
    *
    * @return {Number[]} array with the vector's components.
+   * @deprecated To retrieve vector components, use `v.values`
    * @example
    * // META:norender
    * function setup() {
    *   // Create a p5.Vector object.
    *   let v = createVector(20, 30);
    *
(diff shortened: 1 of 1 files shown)


src/math/p5.Vector.js
```diff
@@ -2690,14 +2693,15 @@ class Vector {
...
    * @example
    * // META:norender
    * function setup() {
    *   // Create a p5.Vector object.
    *   let v = createVector(20, 30);
    *
-   *   // Prints "[20, 30, 0]" to the console.
-   *   print(v.array());
+   *   // Prints "[20, 30]" to the console.
+   *   print(v.values);
```
limzykenneth: Should still be `array()` since it is documenting `array()`

limzykenneth (changes requested):
`array()` don't specifically need a warning at the moment but also no harm in adding one.

PR #8801, opened 2026-05-16
Add shader examples to trigonometric functions

SOUMITRO-SAHA:
Resolves #8777

#### Changes:
Adds inline p5.strands examples to the reference documentation for 8 trigonometry functions in src/math/trigonometry.js:

- `sin()` — sine-driven color oscillation
- `cos()` — cosine-driven color oscillation  
- `tan()` — rapid tangent-based color transitions
- `asin()` — smooth color transition via asin(sin(t))
- `acos()` — pulsing color transition via acos(cos(t))
- `atan()` — sawtooth color effect via atan(tan(t))
- `degrees()` — convert radians to degrees in shader
- `radians()` — convert degrees to radians in shader

Each example uses `buildColorShader()` with `finalColor.begin()/end()` and includes a brief explanation that the function can be used in shaders with p5.strands.

#### Screenshots of the change:

Documentation-only change 

#### PR Checklist

- [x] `npm run lint` passes
- [x] [Inline reference] is included / updated
- [ ] [Unit tests] are included / updated

[Inline reference]: https://p5js.org/contribute/contributing_to_the_p5js_reference/
[Unit tests]: https://github.com/processing/p5.js/tree/main/contributor_docs#unit-tests

PR #8802, opened 2026-05-16
Add shader examples for math functions

### LINKED ISSUE #8777: Add p5.strands examples on regular p5 functions that can be used in a strands context
### Increasing access

This supports the effort of making strands more visible in the reference, and also makes it clearer what can and can't be used within p5.strands. For example, while we now support `random()`, it may not be clear from the `random()` reference page that it can be used in a shader. Having maybe one strands example for each of these would be helpful.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

We maybe want to put strands examples last, and have a bit of explanation above them saying that they can also be used in shaders. To interleave examples with explanation, examples must be written as inline example blocks: https://github.com/processing/p5.js/blob/dev-2.0/contributor_docs/contributing_to_the_p5js_reference.md#inserting-examples-within-the-description

SOUMITRO-SAHA:
Resolves #8777 (Calculation batch)

#### Changes:
Adds inline p5.strands examples to the reference documentation for 13 calculation functions in src/math/calculation.js:

- `abs()` — mirror/fold effect using abs(sin(t))
- `ceil()` — stepped color bands
- `floor()` — posterized banding effect
- `fract()` — repeating gradient pattern
- `sqrt()` — smooth easing curve
- `pow()` — gamma curve effect
- `exp()` — accelerating brightness
- `log()` — decelerating color transition
- `min()` — maximum brightness clamp
- `max()` — minimum brightness clamp
- `round()` — quantized posterized colors
- `lerp()` — color blending (maps to GLSL `mix`)
- `map()` — remapped time to color

Each example uses `buildColorShader()` with `finalColor.begin()/end()` and includes a brief explanation that the function can be used in shaders with p5.strands.

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes


src/math/calculation.js
```diff
@@ -377,6 +491,35 @@ function calculation(p5, fn){
...
+   *   createCanvas(100, 100, WEBGL);
+   *   myShader = buildColorShader(shaderCallback);
+   *   describe('A sphere that blends between teal and coral.');
+   * }
+   *
+   * function shaderCallback() {
+   *   let t = millis() * 0.001;
+   *   let value = 0.5 + 0.5 * sin(t);
+   *   finalColor.begin();
+   *   finalColor.set(lerp([0, 0.8, 0.8, 1], [1, 0.5, 0.3, 1], value));
```
nbogie: As elsewhere, i think breaking this down is clearer
```js

let mixFraction = ...
//define the two colours we'll mix between
//both as [r, g, b, a], all from 0.0 to 1.0
let color1 = [0, 0.8, 08, 1]
let color2 = [1, 0.5, 0.3, 1]
let mixedColor = lerp(color1, color2, mixFraction);
finalColor.set(mixedColor);
```

This is also has quite a lot going on in order to demonstrate lerp: (millis and sin, and its normalization), but:
* I think it's ok, if we're assuming the expected reader already understands lerp from a 2d context.
* Otherwise a static  form like `
```js
//mix a color 25% cyan, 75% magenta
let mixedColor = lerp(cyan, magenta, 0.25);
```
would be better suited.

However, your example is far more inspiring! I like it!

nbogie (commented):
looks good though  i didn't test the examples.

src/math/calculation.js
SOUMITRO-SAHA: Thanks @nbogie ! I'll apply the same pattern as PR #8801  — breaking down the `lerp()` example with named color variables and a clearer structure. I like keeping the animated example since it's more inspiring, but I'll make it more readable. 

Pushing the changes shortly!

perminder-17 (changes requested):
On the website when I run your forked branch to check the actual changes, it looks something like this:

Image

You can verify your changes by running the p5.js-website locally against your fork/branch of p5.js.


### Here's the step

1. Clone the website repository:

```
git clone https://github.com/processing/p5.js-website.git
cd p5.js-website
```

2. Switch to the 2.0 branch (since you are working on the p5.js 2.x version):
```
git checkout 2.0
```

3. Install dependencies:

```
npm install
```

Run the website against your p5.js branch using the custom:dev script.

```
npm run custom:dev https://github.com/SOUMITRO-SAHA/p5.js.git#docs-8777-strands-calc-examples
```

It would run your current changes.


Also, if you need to add new examples, you need to use the `@example` tag and then write your code instead of ```js . After the changes, you can verify by running the website locally with your changes. Thanks for your work so far :)

SOUMITRO-SAHA:
Thanks for the detailed steps and the screenshot, @perminder-17 ! I really appreciate you taking the time to verify this on the website and showing me how to set it up locally.

I wasn't aware the website renders examples differently than how they appear in the web editor. I had been testing all the examples in the p5 web editor and using the  ``js example  syntax I saw in other parts of the codebase (like `material.js`), so I didn't realise that format wasn't rendering correctly on the reference pages for these math functions.

I'll:

1. Clone the website repo and run it locally against my branch to verify the current state
2. Update all the strands examples to use @example tags instead of the code fence blocks
3. Verify everything renders correctly on the site before pushing

Thanks for the guidance

SOUMITRO-SAHA:
https://github.com/user-attachments/assets/6b4eff3f-b734-4ce6-b80b-68aa92a8aa45

Hi, @perminder-17 ,

I am currently reviewing other examples that use the following format for examples:

**```js example **

I am trying to understand the expected template and check whether this approach will work. From what I understood, we only need to move the example section before the `@method`, `@param`, and `@return` tags.

I have also added a dummy video for testing. If this looks fine, then I will raise fixes for all the examples in PR 8001, 8002, and 8003 as well.

davepagurek:
The issue is just that the `@example` tags all go at the end. If we want to interleave description and examples, there must not be any `@example` tags inside the region where descriptions appear; any existing ones have to be converted to `js example` code fences. I think your previous changes weren't rendering correctly because they appeared after an `@example` block:

image

Generally, I think it *is* preferable to introduce the example with description if it's not obvious, especially in this case because we're suddenly jumping to p5.strands shaders. So I think that means you'll need to convert the previous examples to inline `js example` code fences.

SOUMITRO-SAHA:
Thanks @nbogie , @perminder-17 , and @davepagurek  for all the feedback!

I've tested all the changes locally on the `p5.js-website` following the steps @perminder-17  shared, and everything renders correctly now. I've also improved the examples based on the review suggestions and added clearer comments for explanation.

I'll be applying a similar approach to the remaining PRs shortly.

perminder-17:
> Thanks @nbogie , @perminder-17 , and @davepagurek for all the feedback!
> 
> I've tested all the changes locally on the `p5.js-website` following the steps @perminder-17 shared, and everything renders correctly now. I've also improved the examples based on the review suggestions and added clearer comments for explanation.
> 
> I'll be applying a similar approach to the remaining PRs shortly.

thanks for the update, I'll take a look soon.

SOUMITRO-SAHA:
Hi @perminder-17 , just following up here. Any feedback on this PR, Let me know if anything else is needed.

src/math/calculation.js
```diff
@@ -377,6 +555,55 @@ function calculation(p5, fn){
...
+   *
+   *   // Each color is [R, G, B, A] with values from 0 to 1.
+   *   let teal = [0, 0.8, 0.8, 1];
+   *   let coral = [1, 0.5, 0.3, 1];
+   *
+   *   finalColor.begin();
+   *
+   *   // In p5.strands, lerp() maps to the GLSL mix() function.
+   *   // mix() blends teal (when value = 0) and coral (when value = 1).
+   *   finalColor.set(mix(teal, coral, value));
```
perminder-17: I think we should use `lerp()` here instead of `mix()` since it's in the example of `lerp()`

src/math/calculation.js
```diff
@@ -431,6 +655,55 @@ function calculation(p5, fn){
...
+   * function shaderCallback() {
+   *   // shaderCallback runs on the GPU. millis() gives ms since start; multiply by 0.001 for seconds.
+   *   let t = millis() * 0.001;
+   *
+   *   // log(1 + t) grows quickly at first, then slows down over time.
+   *   // Add 1 so we never try log(0), which is undefined.
+   *   // Multiply by 0.2 to keep it in a usable range.
+   *   // min(..., 1) caps the value at 1.
+   *   let value = log(1 + t) * 0.2;
+   *   value = min(value, 1);
```
perminder-17: Can we do `let value = min(log(1 + t * 2) * 0.4, 1);` instead. 

Also, `0.2` multiplier is too small, it will take a lot of time to reach yellow, so do you think you can use 0.4 instead?

src/math/calculation.js
```diff
@@ -790,6 +1188,54 @@ function calculation(p5, fn){
...
+   * }
+   *
+   * function shaderCallback() {
+   *   // shaderCallback runs on the GPU. millis() gives ms since start; multiply by 0.0005 to slow it.
+   *   let t = millis() * 0.0005;
+   *
+   *   // pow(t, 2) squares the time value: it starts slow then accelerates.
+   *   // Multiply by 0.001 so it doesn't reach 1 too quickly.
+   *   // min(..., 1) caps the value at 1.
+   *   let value = min(pow(t, 2) * 0.001, 1);
```
perminder-17:
```suggestion
   *   let value = pow(t, 2);
```

perminder-17 (commented):
Looks good! just some minor comments.

src/math/calculation.js
SOUMITRO-SAHA: I tried using `lerp()` initially, but it doesn't produce any visual changes on the p5.js website when rendered in a shader callback. Here's why:

In global mode, `p5.js` binds all prototype methods to **window** via `bindGlobal()` (`main.js:99`). So `window.lerp` gets the original scalar-only version. Later, when strands augments lerp to delegate to mix (`strands_api.js:290`), it uses `augmentFn()` which updates `p5.prototype.lerp` and `p5.Graphics.prototype.lerp` — but does NOT update `window.lerp`. The transpiled shader callback resolves **lerp** from the global scope, so it gets the original scalar-only version instead of the strands-aware one.

In contrast, `mix()` is a new function added by strands — there's no pre-existing `window.mix` binding, so it resolves correctly.

Do you think this is a bug? Since `lerp()` works in the playground, should I update the example to use `lerp()`, or keep using `mix()` as a workaround?


https://github.com/user-attachments/assets/aba4cf4c-a8fa-41c1-8447-36f2ee44eabb

perminder-17: You should use `lerp`, I think lerp is not defined at your library. Most likely you are using the older version of code. `lerp` is added recently in the library for strands. Can you please check what's getting on the console in your case? Is it says something like `lerp()` is not defined?

ksen0: +1 to Perminder's comment, `lerp()` is more consistently supported in recent release of 2.3. Thanks!

SOUMITRO-SAHA: Thanks @ksen0  and @perminder-17  for the context. I can confirm that after updating to `p5.js` version `2.3.0`, `lerp()` is working properly in the shader callback. I've tested it and pushed the changes.

perminder-17 (approved):
Thanks for working on everything. Looks good to me!

PR #8803, opened 2026-05-16
Add shader examples to noise random and millis functions

### LINKED ISSUE #8777: Add p5.strands examples on regular p5 functions that can be used in a strands context

SOUMITRO-SAHA:
Resolves  #8777 (Noise, Random, Time batch)

 Changes:
Adds inline `p5.strands` examples to the reference documentation for 3 functions across 3 files:

`src/math/noise.js`
- `noise()` — cloud-like texture effect using `buildFilterShader()` with `filterColor`

`src/math/random.js`
- `random()` — random RGB colors on a shape

`src/utilities/time_date.js`
- `millis()` — time-based color transitions

Each example includes a brief explanation that the function can be used in shaders with p5.strands. The `noise()` example demonstrates `buildFilterShader()` for 2D filter effects, while `random()` and `millis()` use `buildColorShader()`.

 Screenshots of the change:

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/src/math/noise.js
+++ b/src/math/noise.js
@@ -227,6 +227,32 @@ function noise(p5, fn){
    *     }
    *   }
    * }
+   *
+   * `noise()` can also be used in shaders with p5.strands. The following example
+   * uses `noise()` to create a cloud-like texture effect in a filter shader.
+   *
+   * ```js example
+   * let myFilter;
+   *
+   * function setup() {
+   *   createCanvas(100, 100, WEBGL);
+   *   myFilter = buildFilterShader(shaderCallback);
+   *   describe('A cloud-like noise pattern.');
+   * }
+   *
+   * function shaderCallback() {
+   *   filterColor.begin();
+   *   let coord = filterColor.texCoord;
+   *   let t = millis() / 2000;
+   *   let value = noise(coord.x * 5, coord.y * 5, t);
+   *   filterColor.set(mix([0.1, 0.1, 0.3, 1], [0.9, 0.9, 1, 1], value));
+   *   filterColor.end();
+   * }
+   *
+   * function draw() {
+   *   filter(myFilter);
+   * }
+   * ```
    */
   fn.noise = function(x, y = 0, z = 0) {
     if (perlin == null) {
--- a/src/math/noise.js
+++ b/src/math/noise.js
@@ -244,8 +244,10 @@ function noise(p5, fn){
    *   filterColor.begin();
    *   let coord = filterColor.texCoord;
    *   let t = millis() / 2000;
-   *   let value = noise(coord.x * 5, coord.y * 5, t);
-   *   filterColor.set(mix([0.1, 0.1, 0.3, 1], [0.9, 0.9, 1, 1], value));
+   *   let mixFraction = noise(coord.x * 5, coord.y * 5, t);
+   *   let darkBlue = [0.1, 0.1, 0.3, 1];
+   *   let lightBlue = [0.9, 0.9, 1, 1];
+   *   filterColor.set(mix(darkBlue, lightBlue, mixFraction));
    *   filterColor.end();
    * }
    *
--- a/src/math/noise.js
+++ b/src/math/noise.js
@@ -228,8 +228,9 @@ function noise(p5, fn){
    *   }
    * }
    *
-   * `noise()` can also be used in shaders with p5.strands. The following example
-   * uses `noise()` to create a cloud-like texture effect in a filter shader.
+   * `noise()` can also be used in shaders with p5.strands, where it returns
+   * values in the range 0 to 1. The following example uses `noise()` to create
+   * a cloud-like texture effect in a filter shader.
    *
    * ```js example
    * let myFilter;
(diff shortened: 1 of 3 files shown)


src/math/noise.js
```diff
@@ -227,6 +227,32 @@ function noise(p5, fn){
...
+   *   myFilter = buildFilterShader(shaderCallback);
+   *   describe('A cloud-like noise pattern.');
+   * }
+   *
+   * function shaderCallback() {
+   *   filterColor.begin();
+   *   let coord = filterColor.texCoord;
+   *   let t = millis() / 2000;
+   *   let value = noise(coord.x * 5, coord.y * 5, t);
+   *   filterColor.set(mix([0.1, 0.1, 0.3, 1], [0.9, 0.9, 1, 1], value));
```
nbogie: I think this could better be split out fully:

```js
let mixFraction = noise( /* ... omitted... */)
let color1 = [0.1, 0.1, 0.3, 1]; 
let color2 = //etc
let mixedColor = mix(color1, color2, mixFraction)
filterColor.set(mixedColor)
```

Reasoning: 
* this might be read by someone coming to strands for the first time, who loves noise() and is excited to move on with it. The more we can do to reduce the chance of overwhelming them, the better.
* shorter lines are also more readable on lower-res devices without wrap
* separate lines allow easier experimentation and replacement in the editable sketch with less chance of the user getting in a mess

src/math/noise.js
```diff
@@ -227,6 +227,32 @@ function noise(p5, fn){
...
+   *   createCanvas(100, 100, WEBGL);
+   *   myFilter = buildFilterShader(shaderCallback);
+   *   describe('A cloud-like noise pattern.');
+   * }
+   *
+   * function shaderCallback() {
+   *   filterColor.begin();
+   *   let coord = filterColor.texCoord;
+   *   let t = millis() / 2000;
+   *   let value = noise(coord.x * 5, coord.y * 5, t);
```
nbogie: I think it's worth commenting on what range of values noise will return here.

src/math/noise.js
nbogie: I wonder if a second example showing the effects of noiseDetail would be suitable here, too (perhaps driven by the mouseX)?  I guess that's for the noiseDetail() page itself but I worry about lesser discoverability of that.

nbogie (commented):
great!

src/math/noise.js
SOUMITRO-SAHA: Sure! I'll add a second inline example on the `noise()` page showing `noiseDetail()` driven by `mouseX`. That way readers see both functions together without having to hunt for `noiseDetail()`. Pushing shortly!

perminder-17 (approved):
Looks good to me. Thanks @nbogie for the review, and @SOUMITRO-SAHA for the work on it.

src/math/random.js
```diff
@@ -230,6 +230,35 @@ function random(p5, fn){
    *   // Draw the point.
    *   point(x, y);
    * }
+   *
+   * `random()` can also be used in shaders with p5.strands. The following example
```
davepagurek, line 234: I think this has the same issue as before where the examples won't render because we have description + `js example` code fences below an `@example` tag -- we may need to do a follow-up PR fixing this one up

perminder-17: aah, yes. Idk, how I missed this one. Will be opening up a follow-up PR now and will merge this one. Thanks for the finding.

davepagurek: No problem, pretty hard to catch these except by testing on the website repo, which is pretty time consuming. Thanks for checking that on some of the other PRs, which alerted me to this being a potential problem!

perminder-17: Here's the PR for the fix: https://github.com/processing/p5.js/pull/8816

SOUMITRO-SAHA: Thanks @perminder-17  for jumping on this fix! I'm sorry you had to do the extra work here. I wasn't previously aware of how to test the documentation rendering locally via the `p5.js website`, so I appreciate you explaining the setup. I've now pushed format fixes (@example →  ``js example` fences) for #8801 and #8802

PR #8807, opened 2026-05-17
Register the XYZ-D65 color space

### LINKED ISSUE #8806: [p5.js 2.0+ Bug Report]: Attempting to interpolate colors may fail with an exception if the colors are defined in different color spaces

- [ ] Accessibility
- [x] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.2.3

### Web browser and version

Browser-independent (tested on Safari 26.4, Chromium 148.0.7778.167)

### Operating system

macOS

### Steps:

For example:

1. Define a color using `oklch`: `color("oklch(0.8, 0.1, 230)")`
2. Define another color in sRGB: `color(0)`
3. Attempt to interpolate them: `lerpColor(c_0, c_1, 0.5)`

### Result

Uncaught exception from `color.js`: `TypeError: No color space found with id = "xyz-d65"`

### Snippet:

Also see: https://editor.p5js.org/dontwanttothink/sketches/pzp5YlR8t

```js
function setup() {
  createCanvas(400, 400);
}

function draw() {
  const c_0 = color("oklch(0.8 0.1 230)");
  const c_1 = color(0);
  
  printColorSpacePath(c_0);
  printColorSpacePath(c_1);
  
  background(lerpColor(c_0, c_1, 0.5));
}

function printColorSpacePath(c) {     
  console.log(c._color.space.path.map(space => space.id))
}
```

dontwanttothink:
Resolves #8806

I'm not sure whether this is the right approach. I didn't want to expose an additional color mode, so the color space is registered with `color.js` directly. The color space XYZ-D65 needs to be registered somehow, I think, in order for `lerp` to work between sRGB, Lab and Oklab colors.

Unfortunately, any time I tried to test this change, `p5.js` crashed with a message complaining about a totally different color space, HSB, being registered twice. Curiously, this happened even when I tried testing the base branch (i.e., without the change).

As a result, I'm submitting this as a draft in case it is of any use.

 Changes:
 - Fixed: Attempting to interpolate between colors defined in different color spaces could throw an exception in some cases.

#### PR Checklist

- [X] `npm run lint` passes

### DIFF
--- a/src/color/p5.Color.js
+++ b/src/color/p5.Color.js
@@ -26,7 +26,9 @@ import {
   OKLCH as OKLCHSpace,
   contrastWCAG21,
   contrastAPCA,
-  P3
+  P3,
+
+  XYZ_D65 as XYZ_D65Space
 } from 'colorjs.io/fn';
 import HSBSpace from './color_spaces/hsb.js';
 
@@ -1030,6 +1032,11 @@ function color(p5, fn, lifecycles){
   p5.Color.addColorMode(LCH, LCHSpace);
   p5.Color.addColorMode(OKLAB, OKLab);
   p5.Color.addColorMode(OKLCH, OKLCHSpace);
+
+  // The fundamental color space to which all colors can be converted (https://colorjs.io/docs/spaces#xyz-d65)
+  // This needs to be registered with color.js for interpolation to work
+  // reliably.
+  ColorSpace.register(XYZ_D65Space);
 
   lifecycles.presetup = function(){
     const pInst = this;


dontwanttothink:
I don't think this is the right solution, so I'm going to close this (plus it has conflicts now).

PR #8817, opened 2026-05-21
Add paletteLerp for p5.strands (closes #8751)

### LINKED ISSUE #8751: add a paletteLerp equivalent for p5.strands
in regular p5, paletteLerp lets you blend across multiple color stops at given positions, which is super handy for gradients. there is no equivalent inside p5.strands shaders right now, so anyone wanting a multi stop gradient on the gpu has to chain mix() calls or write their own helper.

would be cool to have something with similar ergonomics that works inside strands callbacks. exact api is up for discussion since strands has its own way of expressing things, but the goal would be feature parity with the cpu paletteLerp.

came out of the same call. opening this so it doesn't get lost.

LalitNarayanYadav:
Resolves #8751

## Changes

Implements `paletteLerp` for p5.strands, giving GPU-side feature parity with the CPU `paletteLerp`.

### `strands_transpiler.js`
Adds a pre-pass that rewrites `paletteLerp([[color, pos], ...], t)` into parallel color/position arrays before the main `ArrayExpression` pass fires. This is necessary because acorn-walk visits children before parents — the outer array would be mis-typed as a vector before the `paletteLerp` handler could split the pairs. A `_isPaletteLerpArg` flag prevents the resulting arrays from being wrapped in `strandsNode`.

### `strands_api.js`
Registers `paletteLerp` as a dual-mode runtime function. When inside a strands context, builds an IR `FUNCTION_CALL` node and injects the GLSL helper functions into the shader preamble once via `getPaletteLerpGLSL()`. CPU fallback is preserved.

### `strands_glslBackend.js`
Adds `getPaletteLerpGLSL()` which returns overloaded GLSL helpers for 2–8 color stops using typed array constructors (`vec4[N]`, `float[N]`). Overloads are required because GLSL ES 3.0 has no variadic functions.

### `strands_builtins.js`
Minor cleanup.

## Usage

```js
let s = buildFilterShader(() => {
  filterColor.begin();
  filterColor.set(paletteLerp(
    [
      [color(255, 0, 0), 0.0],
      [color(0, 255, 0), 0.5],
      [color(0, 0, 255), 1.0],
    ],
    filterColor.texCoord.x
  ));
  filterColor.end();
});

function draw() {
  background(220);
  rect(-200, -200, 400, 400);
  filter(s);
}
```

This matches the ergonomics of the CPU `paletteLerp`:

```js
// CPU equivalent
paletteLerp(
  [color(255, 0, 0), color(0, 255, 0), color(0, 0, 255)],
  [0.0, 0.5, 1.0],
  map(mouseX, 0, width, 0, 1)
);
```

## Notes

- Supports 2–8 color stops (covers the vast majority of gradient use cases; more stops can be chained)
- WGSL path not included — can be a follow-up
- All 157 existing shader tests pass with no regressions

## Screenshots

No visual rendering screenshot included — browser cache issues prevented local verification. The full pipeline (transpiler → IR → GLSL emission) was verified by code review and the existing test suite passes cleanly.

## PR Checklist

- [x] `npm run lint` passes
- [ ] Inline reference is included / updated — JSDoc to be added in a follow-up
- [ ] Unit tests are included / updated — no strands unit test infrastructure exists yet; tests to be added once the framework is in place

### DIFF
--- a/src/strands/strands_transpiler.js
+++ b/src/strands/strands_transpiler.js
@@ -3,6 +3,16 @@ import { ancestor, recursive } from 'acorn-walk';
 import escodegen from 'escodegen';
 import { UnarySymbolToName } from './ir_types';
 import * as FES from './strands_FES';
+
+// Registry of strands functions that take raw array literals as arguments.
+// Maps functionName → Set of argument indices that should NOT be
+// converted to vectors by the ArrayExpression visitor.
+// This generalizes the paletteLerp special-case so any future function
+// taking array parameters can register here without modifying ArrayExpression.
+const ARRAY_ARG_FUNCTIONS = {
+  paletteLerp: new Set([0]), // argument 0 is the [[color,pos],...] array
+};
+
 let blockVarCounter = 0;
 let loopVarCounter = 0;
 function replaceBinaryOperator(codeSource) {
@@ -563,12 +573,85 @@ const ASTCallbacks = {
       node.arguments = [];
     }
   },
+  
+  // Rewrite paletteLerp([[color,pos],...], t) → __p5.paletteLerp([c,...], [p,...], t)
+  // Must run before ArrayExpression so the child arrays carry _isPaletteLerpArg
+  // and are not wrapped in strandsNode (which would mis-type them as vectors).
+
+  CallExpression(node, state, ancestors) {
+    if (ancestors.some(a => nodeIsUniform(a) || nodeIsUniformCallbackFn(a, state.uniformCallbackNames))) {
+      return;
+    }
+    if (node.callee?.type !== 'Identifier' || node.callee?.name !== 'paletteLerp') {
+      return;
+    }
+    const args = node.arguments;
+    if (args.length !== 2) {
+      throw new Error(
+        `paletteLerp() requires 2 arguments: (colorStops[], t) — got ${args.length}.\n` +
+        `Usage: paletteLerp([[color(r,g,b), pos], ...], t)`
+      );
+    }
+    const [stopsArg, tArg] = args;
+    if (stopsArg.type !== 'ArrayExpression') {
+      throw new Error(
+        `paletteLerp() first argument must be an array literal: [[color(...), pos], ...]`
+      );
+    }
+    const stops = stopsArg.elements;
+    if (stops.length < 2 || stops.length > 8) {
+      throw new Error(
+        `paletteLerp() requires 2–8 color stops, got ${stops.length}.`
+      );
+    }
+    for (let i = 0; i < stops.length; i++) {
+      if (stops[i].type !== 'ArrayExpression' || stops[i].elements.length !== 2) {
+        throw new Error(
+          `paletteLerp() stop ${i} must be a 2-element array: [color(...), position]`
+        );
+      }
+    }
+    // Split pairs into two parallel arrays
+    const colorsArr = {
+      type: 'ArrayExpression',
+      elements: stops.map(s => s.elements[0]),
+      _isPaletteLerpArg: true,
+    };
+    const positionsArr = {
+      type: 'ArrayExpression',
+      elements: stops.map(s => s.elements[1]),
+      _isPaletteLerpArg: true,
+    };
+    // Rewrite in-place to __p5.paletteLerp(colors, positions, t)
+    node.callee = { type: 'Identifier', name: '__p5.paletteLerp' };
+    node.arguments = [colorsArr, positionsArr, tArg];
+  },
+
+
   // The callbacks for AssignmentExpression and BinaryExpression handle
   // operator overloading including +=, *= assignment expressions
+
+
   ArrayExpression(node, state, ancestors) {
     if (ancestors.some(a => nodeIsUniform(a) || nodeIsUniformCallbackFn(a, state.uniformCallbackNames))) {
       return;
     }
+    // Don't wrap arrays that are arguments to functions expecting raw arrays.
+    // Walk ancestors to find the nearest CallExpression and check the registry.
+    for (let i = ancestors.length - 1; i >= 0; i--) {
+      const a = ancestors[i];
+      if (a.type === 'CallExpression') {
+        const name = a.callee?.name;
+        if (name && ARRAY_ARG_FUNCTIONS[name]) {
+          const argIndex = a.arguments.indexOf(node);
+          if (argIndex !== -1 && ARRAY_ARG_FUNCTIONS[name].has(argIndex)) {
+            return;
+          }
+        }
+        break; // only check nearest CallExpression
+      }
+    }
+
     const original = JSON.parse(JSON.stringify(node));
     node.type = 'CallExpression';
     node.callee = {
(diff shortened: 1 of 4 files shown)


src/strands/strands_transpiler.js
```diff
@@ -560,20 +563,73 @@ const ASTCallbacks = {
...
       return;
     }
-
-    if (node.elements.length < 2 || node.elements.length > 4) {
-      FES.userError(
-        'type error',
-        `Array literals in shader functions are transpiled to vectors and must have 2-4 elements (got ${node.elements.length}).`
-      );
+    // Don't wrap paletteLerp's pre-split color/position arrays as vectors
+    if (node._isPaletteLerpArg) {
```
davepagurek: We talked a bit in https://github.com/processing/p5.js/issues/8751#issuecomment-4417863049 about a more generalized way of doing this that wouldn't be specific to `paletteLerp`, is that something we could add?

src/webgl/strands_glslBackend.js
```diff
@@ -181,6 +181,43 @@ export const glslBackend = {
...
+    vec4 _p5s_paletteLerpSeg(vec4 a, vec4 b, float pa, float pb, float t) {
+    float denom = pb - pa;
+    float localT = abs(denom) < 1e-6 ? 0.5 : clamp((t - pa) / denom, 0.0, 1.0);
+    return mix(a, b, localT);
+    }
+    vec4 paletteLerp(vec4[2] c, float[2] p, float t) {
+    if (t <= p[0]) return c[0]; if (t >= p[1]) return c[1];
+    return _p5s_paletteLerpSeg(c[0],c[1],p[0],p[1],t);
+    }
+    vec4 paletteLerp(vec4[3] c, float[3] p, float t) {
```
davepagurek: One other thought: is it necessary that these be functions? We could potentially just generate the if statement/loop and return the final result when you call `paletteLerp`. Then we wouldn't need to generate a GLSL and WGSL version. (If we do stick with this format, we'd need to create a WGSL version too.)

davepagurek (commented):
Thanks for taking this on! I had a few questions at a high level around the approach. We should probably also add some visual tests in WebGL and WebGPU mode too.

src/strands/strands_transpiler.js
LalitNarayanYadav: That's a good point.  I'll implement the parameter type registry approach we discussed. The plan is to add a `strandsParamTypes` map (or co-locate type descriptors with each function's definition) so the `ArrayExpression` visitor can look up whether a given array argument should be treated as a typed array rather than a vector, without any function-specific special-casing. I'll update the PR with that generalization.

src/webgl/strands_glslBackend.js
LalitNarayanYadav: That's a cleaner approach, generating the interpolation inline would remove the GLSL/WGSL split entirely. For N stops, the transpiler would emit something like:

```glsl
// paletteLerp with 3 stops, inlined
(t <= p0 ? c0 : t >= p2 ? c2 : t < p1 ? mix(c0, c1, clamp((t-p0)/(p1-p0), 0., 1.)) : mix(c1, c2, clamp((t-p1)/(p2-p1), 0., 1.)))
```

I'll switch to inline emission in `strands_api.js` instead of the GLSL helper functions. That also simplifies the backend significantly.

LalitNarayanYadav:
Updated. Both comments addressed:

1. Replaced `_isPaletteLerpArg` with a general `ARRAY_ARG_FUNCTIONS` registry at the top of `strands_transpiler.js`. The `ArrayExpression` visitor now walks ancestors to find the nearest `CallExpression` and checks the registry for whether that argument index should be treated as a raw array. Any future function taking array parameters can register there without touching `ArrayExpression` again.

2. Switched to inline IR emission in `strands_api.js` using `buildTernary`, `mix`, `clamp`, and `sub`/`div` — no GLSL helper functions injected, so WGSL works automatically with no separate backend needed. Removed `getPaletteLerpGLSL()` from `strands_glslBackend.js` entirely.

All 157 existing shader tests pass.

src/strands/strands_transpiler.js
```diff
@@ -1687,30 +1765,58 @@ function transformHelperFunctionEarlyReturns(ast, names) {
...
+
+  // Pre-pass: collect names of functions passed by reference as uniform callbacks
+  const uniformCallbackNames = collectUniformCallbackNames(ast);
+
+  // First pass: transform .set() calls in control flow to use intermediate variables
+  transformSetCallsInControlFlow(ast, uniformCallbackNames);
+
+  // paletteLerp pre-pass: must run before the main pass so ArrayExpression
+  // doesn't wrap [[color,pos],...] as a vector before we can split the pairs.
+  ancestor(ast, {
```
davepagurek, line 1788: Do we still need this with the `ARRAY_ARG_FUNCTIONS` check?

src/strands/strands_transpiler.js
```diff
@@ -560,18 +573,83 @@ const ASTCallbacks = {
       node.arguments = [];
     }
   },
+  
+  // Rewrite paletteLerp([[color,pos],...], t) → __p5.paletteLerp([c,...], [p,...], t)
+  // Must run before ArrayExpression so the child arrays carry _isPaletteLerpArg
+  // and are not wrapped in strandsNode (which would mis-type them as vectors).
+
+  CallExpression(node, state, ancestors) {
```
davepagurek, line 581: Is it possible to generalize this too? Ideally we don't need anything specific to one function in the transpiler, and it's all generalized enough that we could add to it for another function if needed. I see that currently this is used to split the array of pairs into two separate arrays. Is this needed if we don't do that?

davepagurek (commented):
Do you think we could add some visual tests that specifically use paletteLerp, e.g. mapping `texCoord.x` in a filter shader through a multi-stop gradient?

I also left a few comments around the transpiler -- I think the goal for that, if possible, is to make it so that `paletteLerp` isn't special-cased anywhere, and we have a way to opt other functions into the same behaviour where we keep arrays as arrays. That probably means not changing the shape of the parameters to `paletteLerp`, as that's a source of extra special-casing in the transpiler currently. I think that should be possible? but let me know if you look into it more and run into issues.

LalitNarayanYadav:
Good points. Looking at it more carefully:

1. The pre-pass can be removed - `ARRAY_ARG_FUNCTIONS` already prevents the outer array from being vectorized, so there's no race condition to guard against.

2. The `CallExpression` handler in `ASTCallbacks` can also be removed if we don't reshape the arguments. Instead of splitting `[[color,pos],...]` into two parallel arrays, `strands_api.js` can receive the pairs directly and unpack them there.

3. To protect the inner `[color, pos]` sub-arrays from being vectorized, the registry can support nested descriptors - e.g. `paletteLerp: { 0: { type: 'array', elements: 'pass-through' } }`  so `ArrayExpression` skips any array that's a descendant of a registered argument, not just the top level.

Does that direction sound right?

davepagurek:
Sorry for the delay on this, just getting back to it now if you're still interested in continuing to work on it!

I think that direction sounds good. For (3), I'm also thinking about what it would take for people to develop addons for p5.strands too. Rather than storing it in a variable like that, what if we store it on the function itself? e.g.:
```js
fn.paletteLerp = function() {
 // ...
}
fn.paletteLerp.argTypes = [
  {
    type: 'Array',
    subtype: {
      type: 'any' // In theory this could even be another array
    },
  },
  { type: 'Number' }
]
```
So then if we see an array anywhere, we see if it's a descendent of an argument to a function (or method call, for instance mode myp5.paletteLerp) node. If that base function (e.g. something like `paletteLerp` for a function or `myp5.paletteLerp` for a method) is called `fnName`, we could do`const args = eval(fnName + ' && ' + fnName + '.argTypes')` and see if that exists. Then addons can add that to their functions as well. We can basically ignore the types that aren't arrays for now, but we know that strands doesn't do FES very well currently, and possibly this could be used for that too?

PR #8822, opened 2026-05-22
feat(strands): add color() support for beginner-friendly color inputs

LalitNarayanYadav:
Part of the **Open Source Software Microgrant**: beginner-friendly p5.strands API for more p5-like color support.

## Changes

Adds beginner-friendly p5.js-style color support inside the strands context, so users can work with familiar color APIs in shaders instead of raw `vec4` literals.

### `color()` - parse any p5.js color input into a `vec4`

Previously, calling `color()` inside a `modify()` block returned a `p5.Color` object, unusable in a GPU shader context. This PR overrides `color()` in `strands_api.js` to return a `vec4` strands node instead.

**Supported input formats** (all handled by p5's existing `colorjs.io` parser):
- Hex strings: `color('#ff0000')`, `color('#f00')`
- CSS named colors: `color('red')`, `color('cornflowerblue')`
- `rgb()`/`rgba()` strings: `color('rgb(255, 0, 0)')`
- Numeric args: `color(255, 0, 0)`, `color(128)` (grayscale)

**How it works:**
1. Calls the original `p5.color()` to parse the input, no new parsing logic
2. Calls `._getRGBA()` on the result to get `[r, g, b, a]` normalized to 0–1
3. Builds a `vec4` strands node via `primitiveConstructorNode`

### `lerpColor(c1, c2, amt)` - interpolate between two colors

Maps directly to GLSL `mix()` in the strands context, since colors are `vec4`s. Falls through to normal p5 behavior outside strands.

### `red()`, `green()`, `blue()`, `alpha()` - component accessors

Extract scalar RGBA channels from a `vec4` color node via `.x/.y/.z/.w` swizzles.

### `hue()`, `saturation()`, `brightness()`, `lightness()` - HSL/HSB accessors

Inject GLSL helper functions (`_p5_rgb2hsb`, `_p5_rgb2hsl`) into the shader declarations on first use, then extract the relevant channel from the converted `vec3`. Helpers are injected into vertex, fragment, and compute declarations.

Outside a strands context, all functions fall through to normal p5 behavior.

## Example

```js
myShader = baseMaterialShader().modify(() => {
  getFinalColor((color) => {
    // hex string → vec4(1.0, 0.0, 0.0, 1.0)
    let c = color('#ff0000');

    // interpolate two colors
    let mid = lerpColor(c, color('blue'), 0.5);

    // extract channels
    let r = red(c);       // 1.0
    let h = hue(c);       // 0.0 (red hue)
    let b = brightness(c); // 1.0

    return color;
  });
});
```

## Screenshots

N/A - shader/transpiler change with no visual diff.

## PR Checklist

- [x] `npm run lint` passes


src/strands/strands_api.js
```diff
@@ -10,49 +10,49 @@ import {
...
+  height: { typeInfo: DataType.float1, get: p => p.height },
+  mouseX: { typeInfo: DataType.float1, get: p => p.mouseX },
+  mouseY: { typeInfo: DataType.float1, get: p => p.mouseY },
+  pmouseX: { typeInfo: DataType.float1, get: p => p.pmouseX },
+  pmouseY: { typeInfo: DataType.float1, get: p => p.pmouseY },
+  winMouseX: { typeInfo: DataType.float1, get: p => p.winMouseX },
+  winMouseY: { typeInfo: DataType.float1, get: p => p.winMouseY },
+  pwinMouseX: { typeInfo: DataType.float1, get: p => p.pwinMouseX },
+  pwinMouseY: { typeInfo: DataType.float1, get: p => p.pwinMouseY },
+  frameCount: { typeInfo: DataType.float1, get: p => p.frameCount },
```
perminder-17: Do you know why we’re seeing so many diffs?

LalitNarayanYadav: The extra diffs are formatting-only changes (whitespace/trailing commas) applied automatically by npx eslint --fix. They don't affect functionality. Happy to squash these into a separate cleanup commit or revert them if you'd prefer the PR to only touch the color-related lines.

ksen0: Hi @LalitNarayanYadav ! Please only include changes in the PR related to the color-related updates, for easier review. Please revert the cleanup changes

LalitNarayanYadav: Hi @ksen0 ! I've reverted all the formatting changes, the PR now only contains the color-related additions. Sorry for the noise, and thanks for the patience!

src/strands/strands_api.js
```diff
@@ -312,6 +312,156 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
...
+  const originalAlpha = fn.alpha;
+  augmentFn(fn, p5, 'alpha', function (...args) {
+    if (!strandsContext.active) {
+      return originalAlpha.apply(this, args);
+    }
+    return p5.strandsNode(args[0]).w;
+  });
+
+  // HSB/HSL accessors — inject conversion helpers and extract channel
+  const _hsbSnippet = `vec3 _p5_rgb2hsb(vec3 c) {
```
davepagurek: This is GLSL-specific. Is it possible to either build this out of existing functionality, similar to what you were doing with `paletteLerp` in #8817? Otherwise we'd need to move these snippets into the glsl and wgsl backends so that we can have a version for either, which is why it's a little less work for each backend if we can make them work without a snippet 🙂

src/strands/strands_api.js
```diff
@@ -312,6 +312,154 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
...
 
+  const originalColor = fn.color;
+  augmentFn(fn, p5, 'color', function (...args) {
+    if (!strandsContext.active) {
+      return originalColor.apply(this, args);
+    }
+    // Reuse p5's parser - handles hex strings, rgb(), CSS named colors, numerics
+    const c = originalColor.apply(this, args);
+    // _getRGBA() returns [r, g, b, a] normalized to 0-1
+    const rgba = c._getRGBA();
```
davepagurek, line 383: nice!

src/strands/strands_api.js
```diff
@@ -312,6 +312,154 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
...
+    const { id, dimension } = build.primitiveConstructorNode(
```
davepagurek, line 384: Would `return p5.strandsNode(rgba)` be equivalent to this? if so it might be a little better at reusing existing functionality.

davepagurek (commented):
This is a good start, thanks for working on this! I left a few comments, and then additionally do you think we could add unit or visual tests that make use of all the new methods in strands?

LalitNarayanYadav:
Thanks for the review Dave! I'll:

Simplify `color()` to use `p5.strandsNode(rgba)`
Rewrite the HSB/HSL conversions using existing strands math functions to avoid backend-specific GLSL snippets
Add unit/visual tests for all new methods

Quick question on point 2 - should I build the RGB→HSL/HSB conversion purely from existing strands ops (`mix`, `step`, `min`, `max`, `abs`, `mod`) similar to how `map` is implemented? Or would you prefer I add backend-specific snippets to both `strands_glslBackend.js` and `strands_wgslBackend.js`?

davepagurek:
That's a good question, my initial thought is to try to use existing strands ops. A reason why we might not want that is that this will get inlined any time we use it, so it potentially means more compiled shader code, but I feel like that's not likely to be a big issue, especially if people aren't inspecting the generated GLSL that much.

src/strands/strands_api.js
```diff
@@ -312,6 +312,143 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
...
+    );
+    return createStrandsNode(id, dimension, strandsContext);
+  });
+  const originalLerpColor = fn.lerpColor;
+  augmentFn(fn, p5, 'lerpColor', function (...args) {
+    if (!strandsContext.active) {
+      return originalLerpColor.apply(this, args);
+    }
+    // In strands, colors are vec4s - lerpColor maps directly to GLSL mix()
+    return fn.mix(...args);
```
davepagurek: Calling these functions on `fn` makes `this` no longer be correct, since they're being called as static methods here. I believe we should be calling them like `this.mix(...)`.

I *think* that means you'll need to pass in `this` as a parameter into helper functions that you declare outside of these methods, e.g. `_rgb2hsb(this, node)`, and defining that helper like: `_rgb2hsb = (instance, node) => { instance.mix(a, b) ... ` etc.

src/strands/strands_api.js
```diff
@@ -312,6 +312,143 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
...
+    return p5.strandsNode(args[0]).w;
+  });
+
+  // Helper: RGB vec3 → HSB vec3 using backend-agnostic strands ops
+  const _rgb2hsb = (colorNode) => {
+    const r = colorNode.x;
+    const g = colorNode.y;
+    const b = colorNode.z;
+    // vec4 K = vec4(0, -1/3, 2/3, -1)
+    const K = fn.vec4(0.0, -1.0/3.0, 2.0/3.0, -1.0);
```
davepagurek: Since this is js, we don't need the `.0` on these

src/strands/strands_api.js
```diff
@@ -312,6 +312,143 @@ export function initGlobalStrandsAPI(p5, fn, strandsContext) {
...
+    // p = mix(vec4(b, g, K.w, K.z), vec4(g, b, K.x, K.y), step(b, g))
+    const p = fn.mix(
+      fn.vec4(b, g, K.w, K.z),
```
davepagurek: Do we have a reference for the formulas in this helper? I was following along with https://en.wikipedia.org/wiki/HSL_and_HSV#From_RGB for the conversion to hsb, and it seems like hsl can look almost the same as that potentially. An alternative approach could be good too, but it would be good to see where it's coming from

LalitNarayanYadav:
Hey @davepagurek  I have fixed it! Updated `lerpColor` and all helper calls to use `this` instead of `fn`, added `instance` parameter to `_rgb2hsb` and `_rgb2hsl`, removed `.0` from number literals, and added formula references in comments. All tests passing

src/strands/strands_api.js
LalitNarayanYadav: I have added references as the comments above the helper functions - HSB uses the compact branching-free form from http://lolengine.net/blog/2013/07/27/rgb-to-hsv-in-glsl (which is derived from https://en.wikipedia.org/wiki/HSL_and_HSV#From_RGB), and HSL uses the same Wikipedia reference with a branchless mix/step approach for the saturation and hue calculations.

davepagurek:
While trying to get some testers for this, in the mean time could we also add some tests to make sure the things we add here continue to work?

ksen0:
Minor / future work: FES message possible improvements:

## Missing parameter --> glsl function named in message

`let mixed = lerpColor(redColor, 0.5);` (missing parameter) --> "Error: [p5.strands parameter validation error]: Function 'mix' has 3 variants which expect 3 arguments, but 2 arguments were provided."

## Wrong parameter type --> data structure named in message

`let mixed = lerpColor(redColor, "blue", 0.5);` or `let r = red("blue");` (incorrect parameter) --> "Error: [p5.strands type error]: You've tried to construct a scalar or vector type with a non-numeric value: blue

## Which string is bad?

`let redColor = color('#fv0a');` (bad hex) --> "Error: Invalid color string" (even if no line number reasonable here, maybe output the string if possible?)

LalitNarayanYadav:
> While trying to get some testers for this, in the mean time could we also add some tests to make sure the things we add here continue to work?

Added unit tests for `color()`, `lerpColor()`, `red()`/`green()`/`blue()`/`alpha()`, and `hue()` - all passing.

LalitNarayanYadav:
> Minor / future work: FES message possible improvements:
> 
> ## Missing parameter --> glsl function named in message
> `let mixed = lerpColor(redColor, 0.5);` (missing parameter) --> "Error: [p5.strands parameter validation error]: Function 'mix' has 3 variants which expect 3 arguments, but 2 arguments were provided."
> 
> ## Wrong parameter type --> data structure named in message
> `let mixed = lerpColor(redColor, "blue", 0.5);` or `let r = red("blue");` (incorrect parameter) --> "Error: [p5.strands type error]: You've tried to construct a scalar or vector type with a non-numeric value: blue
> 
> ## Which string is bad?
> `let redColor = color('#fv0a');` (bad hex) --> "Error: Invalid color string" (even if no line number reasonable here, maybe output the string if possible?)

Thanks @ksen0 for the FES improvement notes! These are good callouts. Should I address them in this PR or open a separate issue to track them as follow-up work?

ksen0:
@LalitNarayanYadav re: FES message improvements - probably separate issue, as I think that could also use some iteration/feedback before implementation, on exactly how the messages should be. Thanks!

davepagurek (approved):
Thanks for your work on this! I'm going to merge this in, and then would you be up for creating follow-up issues for FES and also to update documentation and examples to show usage in strands for these color functions, and to update existing strands functions to use these where appropriate?

PR #8823, opened 2026-05-23
Add friendly FES error for dimension mismatch on shared variable assi…

### LINKED ISSUE #8812: [p5.js 2.0+ Bug Report]: No FES error reported when assigning to strands shared variables with wrong dimensions

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [x] WebGPU
- [x] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.2.3

### Web browser and version

All

### Operating system

All

### Steps:
1. Create a `sharedVec3()`
2. Assign to it with a single value

You get a not very friendly GLSL error with dimensions not matching rather than a more directed strands error.

### Snippet:

Live: https://editor.p5js.org/davepagurek/sketches/1msmY-46C

```js
let textInput, fontFile, fontSize, extrudeDepth;
let colorA, colorB, trailLength, trailCount;
let tiltAmount, driftAmount, enableAnim, animSpeed;
let iridShader;
let textGeom = null;

async function setup() {
  textInput = createInput('TEST')
  fontFile = (await loadFont('https://fonts.gstatic.com/s/inter/v20/UcCO3FwrK3iLTeHuS_nVMrMxCp50SjIw2boKoduKmMEVuDyYMZhrib2Bg-4.ttf'))
  fontSize = createSlider(20, 200, 72, 1)
  extrudeDepth = createSlider(0, 120, 28, 1)
  colorA = createColorPicker('#ff6ec4')
  colorB = createColorPicker('#7873f5')
  trailLength = createSlider(0.1, 4, 1.8, 0.1)
  trailCount = createSlider(0, 20, 10, 1)
  tiltAmount = createSlider(0, 1, 0.38, 0.01)
  driftAmount = createSlider(0, 120, 28, 1)
  enableAnim = createCheckbox('Animate', true)
  animSpeed = createSlider(25, 400, 100, 1)

  textFont(fontFile);
  textSize(fontSize.value());
  textAlign(CENTER);
  const tw = textWidth(textInput.value()) || 200;
  const th = fontSize.value();
  createCanvas(tw + 80, th * 2.2 + 60, WEBGL);

  const txt = textInput.value() || ' ';
  const ed = extrudeDepth.value();
  try {
    textGeom = fontFile.textToModel(txt, 0, 0, { sampleFactor: 0.3, extrude: ed });
  } catch(e) {
    textGeom = null;
  }

  iridShader = buildMaterialShader(() => {
    let t = uniformFloat(() => millis() * 0.001);
    let trailAlpha = uniformFloat(() => 1.0);
    let caR = uniformFloat(() => red(color(colorA.value())) / 255);
    let caG = uniformFloat(() => green(color(colorA.value())) / 255);
    let caB = uniformFloat(() => blue(color(colorA.value())) / 255);
    let cbR = uniformFloat(() => red(color(colorB.value())) / 255);
    let cbG = uniformFloat(() => green(color(colorB.value())) / 255);
    let cbB = uniformFloat(() => blue(color(colorB.value())) / 255);

    let worldPosX = sharedVec3();
    let worldPosY = sharedVec3();
    let worldPosZ = sharedVec3();
    let myNormalX = sharedVec3();
    let myNormalY = sharedVec3();
    let myNormalZ = sharedVec3();

    worldInputs.begin();
    worldPosX = worldInputs.position.x;
    worldPosY = worldInputs.position.y;
    worldPosZ = worldInputs.position.z;
    myNormalX = worldInputs.normal.x;
    myNormalY = worldInputs.normal.y;
    myNormalZ = worldInputs.normal.z;
    worldInputs.end();

    finalColor.begin();
    let nLen = sqrt(myNormalX * myNormalX + myNormalY * myNormalY + myNormalZ * myNormalZ) + 0.0001;
    let nz = myNormalZ / nLen;
    let fresnel = pow(1.0 - abs(nz), 2.5);
    let phase = worldPosX * 0.012 + worldPosY * 0.008 + worldPosZ * 0.005 + t * 0.6 + fresnel * 0.9;
    let r = 0.5 + 0.5 * cos(6.2831 * (phase + 0.0));
    let g = 0.5 + 0.5 * cos(6.2831 * (phase + 0.333));
    let b = 0.5 + 0.5 * cos(6.2831 * (phase + 0.667));
    let baseR = mix(caR, cbR, fresnel);
    let baseG = mix(caG, cbG, fresnel);
    let baseB = mix(caB, cbB, fresnel);
    let finalR = mix(baseR, r, 0.72 + 0.28 * fresnel);
    let finalG = mix(baseG, g, 0.72 + 0.28 * fresnel);
    let finalB = mix(baseB, b, 0.72 + 0.28 * fresnel);
    let brightness = 0.85 + 0.15 * fresnel;
    finalColor.color = [finalR * brightness, finalG * brightness, finalB * brightness, trailAlpha];
    finalColor.end();
  });
}

function getPose(t) {
  const tilt = tiltAmount.value();
  const drift = driftAmount.value();
  return {
    rx: sin(t * 0.71) * tilt * 0.55,
    ry: sin(t * 0.5) * tilt,
    rz: sin(t * 0.38) * tilt * 0.2,
    px: sin(t * 0.31) * drift,
    py: cos(t * 0.43) * drift * 0.45,
  };
}

function applyPose(pose) {
  translate(pose.px, pose.py, 0);
  rotateX(pose.rx);
  rotateY(pose.ry);
  rotateZ(pose.rz);
}

function draw() {
  clear();
  orbitControl();
  noStroke();
  ambientLight(80, 80, 90);
  directionalLight(255, 240, 255, -0.5, 0.6, -0.8);
  directionalLight(180, 160, 255, 0.7, -0.4, 0.5);

  const spd = animSpeed.value() / 100;
  const T = enableAnim.checked() ? millis() * 0.001 * spd : 0;

  if (!textGeom) return;

  shader(iridShader);
  specularMaterial(255);
  shininess(60);

  const steps = floor(trailCount.value());
  const tLen = trailLength.value();
  const txt = textInput.value() || ' ';
  const ed = extrudeDepth.value();

  if (steps > 1) {
    const N = 80;
    textFont(fontFile);
    textSize(fontSize.value());
    textAlign(CENTER);
    const tw2 = textWidth(txt) * 0.5;
    const th2 = fontSize.value() * 0.25;
    const offsets = [
      [-tw2, -th2],
      [ tw2, -th2],
      [ tw2,  th2],
      [-tw2,  th2],
    ];

    for (let oi = 0; oi < offsets.length; oi++) {
      const ox = offsets[oi][0];
      const oy = offsets[oi][1];
      const pts = [];
      for (let i = 0; i <= N; i++) {
        const frac = i / N;
        const tPast = T - frac * tLen;
        const p = getPose(tPast);
        const cosRY = cos(p.ry), sinRY = sin(p.ry);
        const cosRX = cos(p.rx), sinRX = sin(p.rx);
        const cosRZ = cos(p.rz), sinRZ = sin(p.rz);
        let x = ox * cosRY + sinRY * (oy * sinRX + ed * 0.5 * cosRX);
        let y = oy * cosRX - ed * 0.5 * sinRX;
        let z = -ox * sinRY + cosRY * (oy * sinRX + ed * 0.5 * cosRX);
        const x2 = x * cosRZ - y * sinRZ;
        const y2 = x * sinRZ + y * cosRZ;
        x = x2 + p.px;
        y = y2 + p.py;
        pts.push([x, y, z]);
      }

      beginShape(TRIANGLE_STRIP);
      const halfW = max(2, fontSize.value() * 0.04);
      for (let i = 0; i < pts.length - 1; i++) {
        const frac = i / (pts.length - 2);
        const alpha = (1.0 - frac) * 0.55;
        iridShader.setUniform('trailAlpha', alpha);
        const p0 = pts[i], p1 = pts[i + 1];
        const dx = p1[0] - p0[0], dy = p1[1] - p0[1];
        const len = sqrt(dx*dx + dy*dy) + 0.0001;
        const tx2 = dx/len, ty2 = dy/len;
        const nx2 = -ty2 * halfW, ny2 = tx2 * halfW;
        vertex(p0[0] + nx2, p0[1] + ny2, p0[2]);
        vertex(p0[0] - nx2, p0[1] - ny2, p0[2]);
      }
      endShape();
    }
  }

  for (let i = steps; i >= 1; i--) {
    const frac = i / steps;
    const tPast = T - frac * tLen;
    const alpha = (1.0 - frac) * 0.48 + 0.04;
    const sc = 1.0 - frac * 0.08;
    iridShader.setUniform('trailAlpha', alpha);
    const pose = getPose(tPast);
    const depthZ = -frac * extrudeDepth.value() * 0.6;
    drawTransparent(() => {
      push();
      translate(pose.px, pose.py, depthZ);
      rotateX(pose.rx);
      rotateY(pose.ry);
      rotateZ(pose.rz);
      scale(sc);
      textFont(fontFile);
      textSize(fontSize.value());
      textAlign(CENTER);
      const tw3 = textWidth(txt);
      translate(-tw3 * 0.5, fontSize.value() * 0.3, 0);
      model(textGeom);
      pop();
    });
  }

  iridShader.setUniform('trailAlpha', 1.0);
  const mainPose = getPose(T);
  push();
  applyPose(mainPose);
  textFont(fontFile);
  textSize(fontSize.value());
  textAlign(CENTER);
  const twMain = textWidth(txt);
  translate(-twMain * 0.5, fontSize.value() * 0.3, 0);
  model(textGeom);
  pop();
}
```

harshiltewari2004:
Resolves #8812

## Changes

Adds a friendly error for the case where a value of the wrong dimension is assigned to a p5.strands shared variable. Previously, code like:

​```js
let worldPosX = sharedVec3();
worldPosX = inputs.position.x; // scalar assigned to vec3
​```

would emit invalid shader source (`vec3 = float`) and bubble up as a cryptic GPU shader-compiler message such as `ERROR: 0:98: 'assign' : cannot convert from 'highp float' to 'highp 3-component vector of float'`, with no way for the user to map it back to the JS variable that caused it.

After this PR, the same code throws at IR-construction time with:

​```
[p5.strands dimension mismatch]: Cannot assign a value of dimension 1 to `worldPosX`, which expects dimension 3.
​```

This follows the four-call-site approach discussed in the plan comment on #8812 (factor out a shared helper, throw early via the existing `userError` convention).

### Implementation

A new `dimensionMismatchError(declaredDim, actualDim, varName)` helper in `src/strands/strands_FES.js` is wired into the four sites that produce assignment IR nodes for strands-tracked variables:

1. `bridge()` in `src/strands/strands_node.js` — primary path for `sharedX = value` after the transpiler rewrites the assignment to `.bridge(...)`.
2. `bridgeSwizzle()` in `src/strands/strands_node.js` — for swizzle writes like `sharedX.xyz = value`. Declared dimension comes from `swizzlePattern.length`.
3. `swizzleTrap.set` in `src/strands/ir_builders.js` — the existing partial dimension check was refactored to use the shared helper, preserving the broadcast (`dim === 1`) and exact-match cases.
4. `Object.defineProperty` setter in `src/strands/strands_api.js` — for hook-struct field writes like `worldInputs.position = newVec`. Declared dimension comes from `propertyType.dataType.dimension`.

Broadcasting from a scalar (`dim === 1`) is preserved as before, consistent with non-shared assignment behavior.

### Tests

Three tests added in both `test/unit/webgl/p5.Shader.js` and `test/unit/webgpu/p5.Shader.js` under the `p5.strands` suite:

- `reports a friendly error when assigning a scalar to a sharedVec3 (bridge)`
- `reports a friendly error on dimension mismatch via swizzle write (bridgeSwizzle)`
- `does not error when shared variable assignment dimensions match`

### Test infrastructure note

Two small adjustments in `test/unit/webgl/p5.Shader.js` weren't part of the fix itself but were needed for the tests to run cleanly:

- The `vi.mock` factory for `strands_FES` was extended to include `dimensionMismatchError`. The existing mock only stubbed `userError` and `internalError`, so the new import broke module loading.
- The `afterEach` mock-clearing hook in the `p5.strands error messages` suite was changed to `beforeEach`. The original only cleaned up *after* each test, so the suite's first test inherited `mockUserError.mock.calls[0]` from any prior test in the file that had fired a `userError`. This was a latent fragility that the new tests happened to expose — `beforeEach` defends against it generally.

## Screenshots of the change

N/A — error-message change only. Before/after messages included above.

#### PR Checklist

- [x] `npm run lint` passes
- [ ] [Inline reference] is included / updated *(N/A — no public API added)*


src/strands/strands_api.js
```diff
@@ -671,6 +671,14 @@ function createHookArguments(strandsContext, parameters){
             return createStrandsNode(propNode.id, propNode.dimension, strandsContext, onRebind);
           },
           set(val) {
+
+            if(val?.isStrandsNode&&val.dimension!==propertyType.dataType.dimension){
```
davepagurek: There's a chance that the value is not yet a strands node, but in the if statement below on line 689 we create a strands node if it isn't one already. We could possibly move this check to happen after that, and then check for dimensions, to make sure we don't accidentally miss a scenario that should trigger a warning if e.g. the user assigned a plain number instead of a strands node?

src/strands/strands_node.js
```diff
@@ -56,6 +57,13 @@ export class StrandsNode {
 
     // For varying variables, we need both assignment generation AND a way to reference by identifier
     if (this._originalIdentifier) {
+      if(value?.isStrandsNode && value.dimension!==this._originalDimension){
```
davepagurek: Similar to before, in an if statement above here we create `newVal` if it wasn't originally a strands node. Maybe we should capture that and use it so that we don't accidentally skip this check?

src/strands/strands_node.js
```diff
@@ -108,6 +116,13 @@ export class StrandsNode {
 
     // For varying variables, create swizzle assignment
     if (this._originalIdentifier) {
+      if(value?.isStrandsNode && value.dimension!==swizzlePattern.length){
```
davepagurek: Looks like the same pattern applies here too.

davepagurek (commented):
Nice work on this so far, and thanks for adding tests! I left some comments about situations where we might not run the dimension check, which seems to be around assigning to something that isnt a strands node, which could be like `mySharedVec3 = 4`. Once we address that, it's maybe worth adding a test case for that too.

harshiltewari2004:
Thanks @davepagurek  for the review! I see the pattern — the check currently runs before the raw value gets coerced into a strands node, so plain assignments like mySharedVec3 = 4 skip it. I'll move the dimension check to after node creation at each of the sites you flagged.
One thing I want to confirm before I do: we established on #8812 that scalar broadcasting is allowed (mySharedVec3 = 4 broadcasting 4 to all components, consistent with myVec = 5 for non-shared vectors). So when the check catches a plain number, should it (a) treat a dimension-1 scalar as valid broadcast and only flag genuine mismatches like mySharedVec3 = someVec2, or (b) warn on any non-matching dimension including scalars? Want to make sure the post-coercion check preserves the broadcasting behavior we agreed on.

harshiltewari2004:
Hi @davepagurek ,gentle bump — just wanted to check if the scalar broadcast question above is settled before I start on the fix. Happy to proceed once I have a steer on that.

davepagurek:
I believe in call cases in strands, we don't distinguish between number literals and scalar strands nodes, so assigning a 1D node to a shared vec3 should still be valid. This is to be consistent with how GLSL handles float literals in operations with vectors -- when in doubt, let's default to what GLSL does.

harshiltewari2004:
Got it — so the check should flag a genuine mismatch (e.g. vec2 → vec3) but let a scalar (dim 1) broadcast through, matching GLSL. I'll move the check to after node coercion at all four sites and add the scalar exception, plus a test for the mySharedVec3 = 4 case you mentioned. Will push shortly.

harshiltewari2004:
Thanks for the review. Here's the approach I'm planning — digging into this surfaced a wrinkle worth confirming before I push.

The literal "move the check after coercion and read the coerced node's dimension" doesn't work here, because primitiveConstructorNode doesn't derive a dimension from the value — it echoes back whatever typeInfo.dimension it's handed. In bridge() that's this.dimension (the target's), so a coerced node always carries the target dimension and the check could never fire on it.

So instead the fix reads the value's true dimension directly, independent of coercion:
``const valueDim = value?.isStrandsNode
  ? value.dimension
  : (Array.isArray(value) ? value.length : 1);
if (valueDim !== this._originalDimension && valueDim !== 1) {
  dimensionMismatchError(this._originalDimension, valueDim, this._originalIdentifier);
}``

The valueDim !== 1 guard preserves the scalar-broadcast case we agreed on in #8812 — a plain scalar assigned to a higher-dimension shared var is allowed, only genuine mismatches (e.g. vec2 → vec3) throw. This also fixes the original bug you flagged: the old value?.isStrandsNode && guard skipped raw-number assignments like mySharedVec3 = 4 entirely.
I'll apply the same value-dimension pattern at the other three sites, each using its own declared dimension.

One thing I want to check before editing tests: the existing reports a friendly error when assigning a scalar to a sharedVec3 test expects a throw on scalar → vec3, but that's the broadcast case we said is valid. Should that test be retargeted to a real mismatch (vec2 → vec3), or am I misreading the intent there?

harshiltewari2004:
Hi @davepagurek gentle bump on this — no rush, just flagging it's still open whenever you have a chance. The main thing I'd value your read on is the existing scalar → sharedVec3 test: since we landed on scalar broadcast being valid, should that test be retargeted to a real mismatch (vec2 → vec3), or am I reading the intent wrong? Happy to proceed once that's clear.

davepagurek:
> One thing I want to check before editing tests: the existing reports a friendly error when assigning a scalar to a sharedVec3 test expects a throw on scalar → vec3, but that's the broadcast case we said is valid. Should that test be retargeted to a real mismatch (vec2 → vec3), or am I misreading the intent there?

I think in this case it makes sense to remove that particular assertion since scalars can "turn into" vectors safely, and add a test preventing vec2s from being used as vec3s like you said. thanks!

harshiltewari2004:
Got it, will do — thanks

harshiltewari2004:
Pushed the fix. Summary of what changed:

**Dimension check now reads the value's real dimension before coercion.** The earlier version only fired when `value?.isStrandsNode` was true, so plain-number assignments (`mySharedVec3 = 4`) slipped through. Moving the check after coercion doesn't help either, because `primitiveConstructorNode` echoes back whatever dimension it's handed rather than deriving it from the value — so a coerced node's `.dimension` isn't the value's true dimension. Instead I compute it independently up front:

```js
const valDim = value?.isStrandsNode ? value.dimension : (Array.isArray(value) ? value.length : 1);
if (valDim !== declaredDim && valDim !== 1) { /* error */ }
```

The `!== 1` keeps scalar broadcast valid, per your confirmation. Applied at three sites: `bridge()` and `bridgeSwizzle()` in `strands_node.js`, and the shared-var setter in `strands_api.js`.

**Left `swizzleTrap.set` in `ir_builders.js` untouched** — it already branches correctly on RHS shape (broadcasts dim-1, maps matching dims, errors otherwise), so it was effectively the reference for the other three. Flagging it so the untouched swizzle-trap path doesn't look like an oversight.

**Tests** (both webgl and webgpu):
- Added the vec2→vec3 mismatch test you asked for.
- For the old scalar→sharedVec3 throw test: rather than deleting it, I converted it to assert `not.toThrow()`, since scalar broadcast is now valid and it seemed worth documenting + guarding against a regression. Happy to just remove it instead if you'd prefer — your call.
- Kept the existing bridgeSwizzle mismatch and matching-dims tests as-is.

All four pass on both renderers; full suite is green (a couple of visual screenshot cases timed out under load but pass in isolation, unrelated to this change).

src/strands/strands_api.js
```diff
@@ -686,6 +686,15 @@ function createHookArguments(strandsContext, parameters){
             return createStrandsNode(propNode.id, propNode.dimension, strandsContext, onRebind);
           },
           set(val) {
+            const valDim = val?.isStrandsNode?val.dimension:(Array.isArray(val)?val.length:1);
```
davepagurek: Definitely minor, but it might be a bit easier to read to space this out a bit, like:
```js
const valDim = val?.isStrandsNode
  ? val.dimension
  : (Array.isArray(val) ? val.length : 1 );
```

test/unit/webgl/p5.Shader.js
```diff
@@ -2648,7 +2713,7 @@ test('returns numbers for builtin globals outside hooks and a strandNode when ca
       assert.include(err.message, '// noprotect');
     };
 
-    afterEach(() => {
+    beforeEach(() => {
```
davepagurek, line 2819: Was this breaking something?

test/unit/webgpu/p5.Shader.js
```diff
@@ -1361,6 +1361,62 @@ suite('WebGPU p5.Shader', function() {
           myp5.compute(s4, 4);
         }).not.toThrow();
       });
+
+      test('allows scalar broadcast when assigning a scalar to a sharedVec3 (bridge)', async () => {
+  await myp5.createCanvas(5, 5, myp5.WEBGPU);
```
davepagurek: Do you mind fixing the indentation in this file?

davepagurek (commented):
Thanks for your work on this, the logic looks good! Just some minor notes.

test/unit/webgl/p5.Shader.js
harshiltewari2004: Yeah — this fixes a pre-existing isolation issue that my new tests exposed. `mockUserError` is module-level, so its `.mock.calls` accumulate across the whole file. The `p5.strands error messages` suite reads `mock.calls[0]`, and with `afterEach` clearing it was relying on nothing upstream in the file having fired `userError` before it ran. My new bridge/bridgeSwizzle throw-tests do fire it (via `dimensionMismatchError`), so by the time that suite started, `mock.calls[0]` was a stale dimension-mismatch call and its first test failed. Switching to `beforeEach` guarantees a clean mock at the start of each test in the suite, which fixes it and guards against the same thing happening from any future test added above it.

src/strands/strands_api.js
harshiltewari2004: Done, spaced it out across all three sites (bridge, bridgeSwizzle, and the setter).

test/unit/webgpu/p5.Shader.js
harshiltewari2004: Fixed, re-indented the new tests to match the suite.

harshiltewari2004:
Hi @davepagurek — whenever you get a chance, this is ready for another look. I've addressed the second-round notes: spaced out the `valDim` ternary at all three sites, re-indented the new webgpu tests to match the suite, and replied on the `afterEach`→`beforeEach` thread explaining what it was fixing. CI is green. No rush, just flagging .Thanks!

harshiltewari2004:
Hi @ksen0 since Dave's been away a few days and this has been ready for re-review — would you be able to take a look, or know someone who could? No rush if it's better to wait for Dave. Thanks!

ksen0:
LGTM; indentation has also been fixed. Thanks both!

PR #8824, opened 2026-05-24
Replace randomness-related noise usage in strands example

### LINKED ISSUE #8776: Use random() more in p5.strands examples
### Increasing access

Previously, we didn't have `random()` supported, so many examples use `noise()` as a source of randomness. This works, but is a little less familiar from how one would use p5 outside of strands, so switching to random() makes these examples a little easier to understand.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

e.g. in the second example in https://beta.p5js.org/reference/p5/buildStrokeShader/, it uses noise for dithering. We probably need to look through all the strands reference items to get a full list of what might need to be updated.

VikasKSingh05:
Resolves #8776

## Changes:
Replaced `noise(pixelInputs.position.xy)` with `random()` in the dithering example within the `buildStrokeShader()` inline documentation in `src/webgl/material.js`.

### Reasoning:
The `noise()` function was previously used as a workaround when `random()` wasn't supported in p5.strands. For dithering (randomly setting pixel transparency), pure randomness is more appropriate than spatially-coherent noise. This change makes the example clearer and more consistent with standard p5.js usage.

### Scope:
After reviewing all p5.strands examples in the codebase, this was the only instance where `noise()` was used purely as a substitute for randomness. Other uses of `noise()` in examples involve intentional smooth spatial/temporal variation and should remain unchanged.

## Screenshots of the change:
N/A - This is a documentation change in inline code examples. The visual output will show proper random dithering instead of patterned noise-based dithering.

#### PR Checklist
- [x] `npm run lint` passes
- [ ] [Unit tests] are included / updated (N/A - documentation-only change)


perminder-17 (commented):
Hi @VikasKSingh05 , your changes are correct but it doesn't fully solves the issue. 

The issue states
> We probably need to look through all the strands reference items to get a full list of what might need to be updated.

And, I think there are some more references where it should use `random()` instead of noise, Do you mind checking them all and replacing them as did with the `buildStrokeShader` one?

VikasKSingh05:
Hi @perminder-17 ,

I've reviewed all p5.strands examples for `noise()` usage. Here's what I found:

### Changed in this PR:
- Dithering example in `buildStrokeShader()` (`src/webgl/material.js`) - replaced `noise(pixelInputs.position.xy)` with `random()`

### Kept as `noise()` (correct usage):
- **Weight scale in `buildStrokeShader()`** - creates smooth spatial/temporal variation, not just randomness
- **Animation in `buildFilterShader()`** - intentionally demonstrates the noise function
- **Roughness in `vertexProperty()` example** (`p5.Geometry.js`) - creates smooth surface deformation; `random()` would break surface continuity

### Summary:
The dithering example was the only case where `noise()` was used purely as a randomness substitute. Other instances are valid uses where noise's smooth/continuous properties are essential.

Let me know if you'd like me to reconsider any of these!

perminder-17 (approved):
Looks good to me!

PR #8828, opened 2026-05-26
Fix saveGif crash when duration <= 0

Chessing234:
### Bug
Calling `saveGif('test', 0)` or a negative duration leads to `TypeError: Cannot read properties of undefined (reading 'length')`.

### Root Cause
When the duration parameter is less than or equal to 0, no frames are recorded during the execution. However, downstream code assumes at least one frame was captured (accessing `frames[0].length` for palette generation and saving), resulting in an unhandled `TypeError`.

### Why Fix is Correct
This change introduces an early parameter validation check in `p5.prototype.saveGif` to throw a `RangeError` if the duration parameter is less than or equal to 0. This matches the existing type validations on parameters and ensures p5.js rejects invalid durations gracefully instead of crashing with a confusing internal error.

### DIFF
--- a/src/image/loading_displaying.js
+++ b/src/image/loading_displaying.js
@@ -303,6 +303,9 @@ p5.prototype.saveGif = async function(
   if (typeof duration !== 'number') {
     throw TypeError('Duration parameter must be a number');
   }
+  if (duration <= 0) {
+    throw RangeError('Duration parameter must be a positive number');
+  }
 
   // extract variables for more comfortable use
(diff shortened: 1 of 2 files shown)


perminder-17 (commented):
Can you please link the issue number in the description of your PR? So I can have context related to what it resolves?

Chessing234:
Closing this in favor of PR #8740, which addresses the same issue (issue #8710) based on recent review feedback.

Issue #8818, opened 2026-05-21
[p5.js 2.0+ Bug Report]: createVector() gives error "TypeError: this._friendlyError is not a function"


sidwellr:
### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [x] Friendly errors
- [ ] Other (specify if possible)

### p5.js version

v2.3.0-rc.1

### Web browser and version

Brave 1.90.122

### Operating system

Linux Mint 6.8.0-117

### Steps to reproduce this

I was trying v2.3.0-rc.1 with the code snippet from issue 8117 to see if this release candidate fixed that issue. The `createVector()` line started giving a warning in version 2.1 to always use arguments when calling `createVector`. But v2.3.0-rc.1 gives a hard error "TypeError: this._friendlyError is not a function".

### Snippet:

```js

function setup() {
  createCanvas(400,400);
  const v=createVector(100,0);
  const w=createVector();
  v.sub(w);
  console.log(v.toString());
  v.rotate(1);
  console.log(v.toString());
}

```

ksen0:
Thanks for finding this. Confirming it's still a problem in rc.4

perminder-17:
Fixed by #8829

PR #8836, opened 2026-05-27
Fix TypeError: _validateParameters is not a function

Chessing234:
Bug: Parameter validation was throwing a hard error when triggered, because `p5.prototype._validateParameters` was incorrectly aliased to the undefined `p5.validateParameters` instead of `p5._validateParameters`.

Root cause: The alias in `validate_params.js` referenced an undefined property instead of the static method defined earlier in the file. This prevented the FES (Friendly Error System) from correctly logging parameter errors and instead crashed the program when it attempted to execute the wrapper.

Why fix is correct: Pointing the alias to the correct static method `p5._validateParameters` restores the intended FES behavior without crashing the program, fixing Issue #8818.

### DIFF
--- a/src/core/friendly_errors/validate_params.js
+++ b/src/core/friendly_errors/validate_params.js
@@ -756,7 +756,7 @@ if (typeof IS_MINIFIED !== 'undefined') {
       }
     }
   };
-  p5.prototype._validateParameters = p5.validateParameters;
+  p5.prototype._validateParameters = p5._validateParameters;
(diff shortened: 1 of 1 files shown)


perminder-17 (commented):
Hi, I think the issue has already been resolved. https://github.com/processing/p5.js/pull/8829

Thanks for your efforts, you can explore other issues from [here](https://github.com/processing/p5.js/issues);

PR #8852, opened 2026-05-31
[1.x] Update CI node version to 22

limzykenneth:
Changes:

All CI now runs Node.js 24 where needed, this should have LTS for quite awhile yet

### DIFF
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -18,10 +18,10 @@ jobs:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           persist-credentials: false
-      - name: Use Node.js 20.x
+      - name: Use Node.js 24.x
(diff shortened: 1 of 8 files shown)


Vaivaswat2244:
Hey @limzykenneth! I pulled your branch locally to test this. The actual problem is that we have puppeteer pinned to ^18.2.1, which is from Oct 2022. This old version tries to download Chrome revision `r1045629` (Chrome 107), and that download fails under node 24 due to changes in how node handles network connections, leaving no chrome binary for the tests to launch.
Switching to the latest version of puppeteer , 25.1.0 solves the issue and tests run fine.

limzykenneth:
The test is not passing with latest version of puppeteer. Since this is 1.x, we'll not spend too much time updating it, I'll just update CI to use Node 22 for now.

Vaivaswat2244:
Makes sense! In the meantime, I can look into this more deeply and I'll try to understand why tests are failing.

ksen0 (approved):

PR #8853, opened 2026-05-31
[2.x] Update CI node version to 24

limzykenneth:
Changes:
All CI now runs Node.js 24 where needed, this should have LTS for quite awhile yet

### DIFF
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -18,10 +18,10 @@ jobs:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           persist-credentials: false
-      - name: Use Node.js 22.x
+      - name: Use Node.js 24.x
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
-          node-version: 22.x
+          node-version: 24
       - name: Get node modules
         run: npm ci
         env:
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -28,12 +28,12 @@ jobs:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           persist-credentials: false
-      
-      - name: Use Node.js 22.x
+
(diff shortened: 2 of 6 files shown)


Vaivaswat2244:
Hey @limzykenneth! Here, Chrome is present on the runner but WebDriverIO fails to establish a ChromeDriver session within the timeout. No ChromeDriver process ever starts, which is why Vitest times out waiting for the browser session.
From local testing, the root cause appears to be that WebDriverIO's automatic ChromeDriver download/unpack step is creating the destination folder but failing to place the binary inside it, so ChromeDriver never launches. The download URL itself is reachable and the zip is valid. manually unpacking it to the expected path (/tmp/chromedriver/linux-<version>/chromedriver-linux64/chromedriver) and then running the tests works fine, which confirms the test infrastructure itself (Vitest + WebDriverIO + Chrome) is sound.
I wasn't able to fully reproduce and fix this locally since the unpack failure appears to be environment-specific. However the CI runner shows the same symptom — ChromeDriver session timeout — so the underlying cause is likely the same.
A possible fix worth trying is adding an explicit ChromeDriver install step to the CI workflow before running tests, to bypass WebDriverIO's automatic download entirely:

```yaml
- name: Install ChromeDriver
  run: npx wdio install chromedriver
```

This would ensure ChromeDriver is present and executable before Vitest tries to connect to it, regardless of whether WebDriverIO's auto-download works in that environment.
Happy to help test further if you can point me to a way to reproduce the CI environment more closely.

Additionally,
This didn't fail on Node 20 because the GitHub Actions runner image tied to Node 20 shipped with an older Chrome version that WebDriverIO 9.0.7 was compatible with. Bumping to Node 24 pulled in a newer runner image with Chrome 148, which WebDriverIO 9.0.7 can't reliably download a matching ChromeDriver for or connect to.

limzykenneth:
Thanks @Vaivaswat2244! I can replicate this locally and it seems to be something flaky, I'll look into it here and might switch to playwright instead

limzykenneth:
Nevermind this for now. I plan to do a more complete dependencies updates for 2.x soon and it's more streamlined to do it all then.

Vaivaswat2244:
Sure! Please lmk if you need any help around this. Thanks!

PR #8854, opened 2026-05-31
Mark private methods as private

ksen0:
The reference incorrectly includes two private methods:

This PR marks them as private/internal.


limzykenneth (approved):
I'll see if I can manually remove it from the website as well.

PR #8878, opened 2026-06-06
fix builtin global accessors for instance mode in strands shaders

### LINKED ISSUE #8877: [p5.js 2.0+ Bug Report]: p5 globals don't update in instance mode

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [x] WebGPU
- [x] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.3.0

### Web browser and version

All

### Operating system

All

### Steps:
1. Create an instance mode sketch
2. Reference global properties like `mouseX` or `width`
3. Run the shader every frame

The values do not update live. If you log the shader code, you'll see that their current values have been hardcoded in and are not uniforms.

This is because we only override properties on `window` and not the instance here:

https://github.com/processing/p5.js/blob/549dd202c43a9524ffdff3d9c2c678ad01eb5bc3/src/strands/strands_api.js#L104-L115

### Snippet:

```js
new p5(p => {
  let myFilter
  p.setup = function () {
    p.createCanvas(400, 400, p.WEBGL)
    myFilter = p.buildFilterShader(({ p }) => {
      p.filterColor.begin()
      const v = p.map(p.mouseX, 0, p.width, 0, 1, true)
      p.filterColor.set([v, v, v, 1])
      p.filterColor.end()
    }, { p })
  }
  
  p.draw = function() {
    p.clear()
    p.filter(myFilter)
  }
})
```

https://editor.p5js.org/davepagurek/sketches/YZUwzzNOB

aashu2006:
Resolves #8877

### Changes:

Fixed `installBuiltinGlobalAccessors` in `strands_api.js` so that the builtin globals like `mouseX` and `width` work correctly in instance mode when used inside strands shader callbacks.

Previously only `window.mouseX` was intercepted, so in instance mode (`p.mouseX`), the values got hardcoded into the shader instead of becoming live uniforms.

Now the function also installs property interceptors on `p5.prototype` and `p5.Graphics.prototype`. It correctly handles both data properties (like `mouseX`) and getter based properties (like `width`, which delegates to `this._renderer?.width`) by saving the original property descriptor and delegating to it when strands isn't active.

Also I've added a unit test that verifies `myp5.mouseX` and `myp5.width` return strands nodes inside hooks and normal numbers outside.

#### PR Checklist

- [x] `npm run lint` passes


src/strands/strands_api.js
```diff
@@ -103,15 +103,75 @@ function installBuiltinGlobalAccessors(strandsContext) {
 
   for (const name of Object.keys(BUILTIN_GLOBAL_SPECS)) {
     const spec = BUILTIN_GLOBAL_SPECS[name]
+    const sym = Symbol(`_strands_${name}`)
+
+    // Define on window for global mode
     Object.defineProperty(window, name, {
```
eupthere: I think we need to check `_isGlobal` here, to avoid adding accessors to the global `window` object on instance mode.

src/strands/strands_api.js
```diff
@@ -103,15 +103,75 @@ function installBuiltinGlobalAccessors(strandsContext) {
...
+      },
+      configurable: true,
+    })
+
+    // Define on p5.prototype for instance mode
+    const originalProtoDesc = Object.getOwnPropertyDescriptor(strandsContext.p5.prototype, name);
+
+    // For data properties (like mouseX), save the initial value into Symbol-keyed store
+    if (originalProtoDesc && 'value' in originalProtoDesc) {
+      strandsContext.p5.prototype[sym] = originalProtoDesc.value;
```
davepagurek: Is there a reason for using a symbol here as opposed to a string?

test/unit/webgl/p5.Shader.js
```diff
@@ -532,6 +532,27 @@ test('returns numbers for builtin globals outside hooks and a strandNode when ca
...
 });
 
+test('returns numbers for builtin globals via instance access outside hooks and a strandNode inside hooks', () => {
+  myp5.createCanvas(5, 5, myp5.WEBGL);
+  myp5.baseMaterialShader().modify(() => {
+    myp5.getPixelInputs(inputs => {
+      //In instance mode, myp5.mouseX and myp5.width should return strands nodes
+      const mxInHook = myp5.mouseX;
+      const wInHook = myp5.width;
+      assert.isTrue(mxInHook.isStrandsNode);
```
davepagurek, line 521: nice!

src/strands/strands_api.js
aashu2006: Using a symbol here to avoid any chance of colliding with existing properties on the instance/prototype. A string like `_strands_mouseX` would also work, but symbols are guaranteed to be unique and stay out of things like `Object.keys()` and `for...in`, so the backing store remains hidden.

happy to switch to strings if you think that's simpler 🙂

src/strands/strands_api.js
aashu2006: thanks for catching it @eupthere, I've added the `_isGlobal` check so we only define window accessors in global mode now. Also updated the tests to use instance access (`myp5.width` instead of `window.width`) since the test framework runs in instance mode anyway.

aashu2006:
Hi @davepagurek, I added the `_isGlobal` check so the window accessors are only defined in global mode. Also updated the tests to use instance access (`myp5.width`) instead of `window.width`, since the test environment runs in instance mode anyway.

Also moved the original descriptor storage to `strandsContext.fnOverrides` and switched from Symbols to string keys for the backing store, like you suggested.

src/strands/strands_api.js
davepagurek: Do they need to be stored on the prototype? Elsewhere we store this on `strandsContext.fnOverrides`, e.g.: https://github.com/processing/p5.js/blob/82135bb8c6a15932e974a48fabcc94aeaf32ac0f/src/strands/strands_api.js#L143

`fn` in addons is a shortcut to `p5.prototype`, so it seems like maybe this should be using that same system.

aashu2006: switched to storing the original descriptors on `strandsContext.fnOverrides` instead of the prototype, and also switched from Symbol to a string key for the per-instance backing store, it should be cleaner now.

aashu2006:
Also, the earlier CI failure was in `p5.Framebuffer.js` (a `createElement` mock issue) unrelated to this PR's changes.

davepagurek:
One thing around `fnOverrides` is that those are automatically put back again when you init/deinit strands: https://github.com/processing/p5.js/blob/82135bb8c6a15932e974a48fabcc94aeaf32ac0f/src/strands/p5.strands.js#L32-L87

Currently we use special keys, so it doesn't quite work as expected -- it'd be adding things to `fn` with those keys on deinit. So if we're opting into this system, we may have to update the logic a little bit to make sure the name we store it as is the original so that it works with this resetting logic, and possibly updating the resetting logic to handle descriptors. Possibly it makes sense to make a similar `fnDescriptorOverrides` if that logic is different enough. But then it also resets them after  Do we also reset the flag saying whether we've installed globals, so that if they're correctly placed back, we also correctly replace them again the next time we make a strands shader?

aashu2006:
@davepagurek yes you're right `fnOverrides` gets reset by `initStrandsContext()` on every `modify()` call, which means the stored descriptors wouldn't survive. It can also lead to `deinitStrandsContext()` overwriting the getters during restore.

I've removed the `fnOverrides` storage entirely and now use the original descriptors captured in the closure directly.

davepagurek (approved):

Issue #8911, opened 2026-06-11
[p5.strands] Instancing API improvements


aashu2006:
### Increasing access

This makes GPU instancing easier to discover and use without requiring a deep understanding of the underlying rendering pipeline. The goal is to make performant rendering techniques more approachable for artists, educators, students, and beginners.

### Most appropriate sub-area of p5.js?

- [ ] Accessibility
- [ ] Color
- [x] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [x] p5.strands

### Feature enhancement details

**Summary**

This issue tracks implementation work for improving the p5.strands instancing API as part of the Microgrant project.

Instancing is a powerful GPU technique for efficiently drawing large numbers of similar objects, but the current API can be difficult for newer users to discover and understand. This project builds on the existing rendering architecture to provide a more approachable and discoverable API surface.

**Proposed Direction**

```js
let repeater = instances(100);
repeater.sphere(20);

// in a shader
let i = instanceIndex();
```

**Planned Work**

* [x] Add `instanceIndex()` as an alias for `instanceID()`
* [x] Implement `instances(count)` returning a drawable object
* [x] Integrate instancing with existing primitive rendering paths
* [x] Add tests for WebGL and WebGPU
* [x] Add documentation and examples

**Project Context**

This issue serves as a tracking thread for ongoing work and related pull requests for the p5.strands instancing API.

**References**

* Design proposal: [gist link](https://gist.github.com/aashu2006/ca13773766637bb22785f16a475b0be1)
* Visual exploration: [tldraw link](https://www.tldraw.com/f/nKqrNyqWS-8fIyC-Zcqo-?d=v-675.-363.3241.1914.page)

aashu2006:
cc @ksen0 @davepagurek

davepagurek:
Does `instanceIndex()` need to be a function? Possibly we could define a getter, similar to how other non-functions like `width` work in strands, that does the same thing.

aashu2006:
I kept it as a function in this #8912  because `instanceID()` is currently a function and the goal here was to make `instanceIndex()` a minimal alias.

[...]

PR #8912, opened 2026-06-11
Add instanceIndex alias for instanceID

aashu2006:
Addresses #8911

**Changes**

* Add `instanceIndex` as a value-based API for accessing the current instance index in p5.strands
* Preserve `instanceID()` for backwards compatibility
* Update docs and examples to use `instanceIndex` as the preferred API
* Update WebGL and WebGPU tests to use `instanceIndex`

`instanceIndex` is exposed as a property rather than a function, making it feel more consistent with other strands values such as `width`, `height`, and `mouseX`. The existing `instanceID()` API remains available for compatibility.

#### PR Checklist

* [x] `npm run lint` passes
* [x] Inline reference is included / updated
* [x] Unit tests are included / updated

### DIFF
--- a/src/strands/p5.strands.js
+++ b/src/strands/p5.strands.js
@@ -338,18 +338,20 @@ if (typeof p5 !== "undefined") {
  */
 
 /**
- * @method instanceID
+ * @property instanceIndex
  * @beta
  * @description
  * Returns the index of the current instance when drawing multiple copies of a
  * shape with <a href="#/p5/model">`model(count)`</a>. The first instance has an
- * ID of `0`, the second has `1`, and so on.
+ * index of `0`, the second has `1`, and so on.
  *
  * This lets each copy of a shape behave differently. For example, you can use
- * the ID to place instances at different positions, give them different colors,
+ * the index to place instances at different positions, give them different colors,
  * or animate them at different speeds.
  *
- * `instanceID()` can only be used inside a p5.strands shader callback.
+ * `instanceIndex` can only be used inside a p5.strands shader callback.
+ *
+ * (Note: `instanceID()` is also available as a function for compatibility.)
  *
  * ```js example
  * let instancesShader;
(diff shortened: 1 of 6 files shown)


aashu2006:
@davepagurek I have updated it based on the discussion around `instanceIndex`.

`instanceIndex` is now exposed as a value instead of a function, while `instanceID()` remains unchanged for compatibility. The implementation uses a dedicated property accessor that creates the same underlying node as `instanceID()`, and docs/examples have been updated to use the value-based API.

perminder-17 (commented):
Thanks for your work, the basic structure looks fine! I'll take a look and test it soon.

aashu2006:
I've fixed the CI failure and pushed an update, the issue was related to TypeScript type generation from the JSDoc annotations. Local `test:types` now passes.

src/strands/p5.strands.js
```diff
@@ -451,7 +453,19 @@ if (typeof p5 !== "undefined") {
...
  * to update the data being read.
  *
- * @webgpu
+ * @type {*}
+ */
+
+/**
+ * @method instanceID
+ * @beta
+ * @description
```
davepagurek, line 463: Do we want to mark this as `@deprecated Prefer using instanceIndex` or something similar?

test/unit/webgl/p5.Shader.js
```diff
@@ -532,6 +532,22 @@ test('returns numbers for builtin globals outside hooks and a strandNode when ca
   assert.strictEqual(w, myp5.width);
 });
 
+test('instanceIndex is a value and instanceID() is a compatibility alias', () => {
```
davepagurek: Looks like this + the test above are not indented, let's make them match the tests around them

davepagurek (commented):
Looks good overall, just two minor things!

src/strands/p5.strands.js
aashu2006: Done!

davepagurek (approved):

PR #8920, opened 2026-06-15
fix: lerp colors in current color mode

### LINKED ISSUE #8883: [p5.js 2.0+ Bug Report]: `lerpColor()` not affected by `colorMode()`

- [ ] Accessibility
- [x] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.3.0

### Web browser and version

browser independent, tested on Safari Version 26.5 (21624.2.5.11.4)

### Operating system

MacOS

### Steps to reproduce this

setting the `colorMode()` to anything other than RGB doesn't seem to affect [lerpColor](https://beta.p5js.org/reference/p5/lerpColor/). This used to work in previous p5 versions and the documentation reads:

The way that colors are interpolated depends on the current [colorMode()](https://beta.p5js.org/reference/p5/colorMode/).

expected (p5.js 1.x): 

Image

actual result (p5.js 2.x):

Image

### Steps:
1. Copy the code snippet into a p5.js 2.0+ version
2. Compare result with running the code in p5.js 1.x

### Snippet:

```js

function setup() {
  createCanvas(100, 100);

  background(200);

  // Create p5.Color objects to interpolate between.
  let from = color(218, 165, 32);
  let to = color(72, 61, 139);

  colorMode(HSB)
  
  // Create intermediate colors.
  let interA = lerpColor(from, to, 0.33);
  let interB = lerpColor(from, to, 0.66);

  // Draw the left rectangle.
  noStroke();
  fill(from);
  rect(10, 20, 20, 60);

  // Draw the left-center rectangle.
  fill(interA);
  rect(30, 20, 20, 60);

  // Draw the right-center rectangle.
  fill(interB);
  rect(50, 20, 20, 60);

  // Draw the right rectangle.
  fill(to);
  rect(70, 20, 20, 60);

  describe(
    'Four rectangles. From left to right, the rectangles are tan, brown, brownish purple, and purple.'
  );
}

```

pupuking723:
Resolves #8883

Changes:
- Makes `Color.lerp()` interpolate in the explicit mode passed by `lerpColor()`, instead of interpolating in the source colors' original color space and only formatting the result in the current mode.
- Keeps the existing fallback behavior when no mode is passed.
- Updates the existing HSL/HSB `lerpColor()` expectations to match current-mode interpolation. I did not add a new regression test, following the maintainer request in #8883 to start with just the fix.

Screenshots of the change:
Not applicable.

#### PR Checklist

- [x] `npm run lint` passes

Verification:
- `npm test -- --run test/unit/color/creating_reading.js test/unit/color/p5.Color.js`
- `npm run lint`

AI assistance disclosure: I used Codex assistively while preparing this change. I reviewed the issue discussion, source changes, test updates, and verification output myself, and I take responsibility for the PR.

### DIFF
--- a/test/unit/color/creating_reading.js
+++ b/test/unit/color/creating_reading.js
@@ -71,33 +71,31 @@ suite('color/CreatingReading', function() {
     });
 
     test('should correctly get lerp colors in HSL', function() {
-      // NOTE: This is equivalent to RGB case so is testing nothing new
       mockP5Prototype.colorMode(mockP5Prototype.HSL);
       var interA = mockP5Prototype.lerpColor(fromColor, toColor, 0.33);
       var interB = mockP5Prototype.lerpColor(fromColor, toColor, 0.66);
 
-      assert.closeTo(interA._color.coords[0], 37, 1);
-      assert.closeTo(interA._color.coords[1], 43, 1);
+      assert.closeTo(interA._color.coords[0], 352, 1);
+      assert.closeTo(interA._color.coords[1], 63, 1);
       assert.closeTo(interA._color.coords[2], 46, 1);
 
-      assert.closeTo(interB._color.coords[0], 345, 1);
-      assert.closeTo(interB._color.coords[1], 12, 1);
+      assert.closeTo(interB._color.coords[0], 301, 1);
+      assert.closeTo(interB._color.coords[1], 51, 1);
       assert.closeTo(interB._color.coords[2], 43, 1);
     });
 
     test('should correctly get lerp colors in HSB', function() {
-      // NOTE: This is equivalent to RGB case so is testing nothing new
       mockP5Prototype.colorMode(mockP5Prototype.HSB);
       var interA = mockP5Prototype.lerpColor(fromColor, toColor, 0.33);
       var interB = mockP5Prototype.lerpColor(fromColor, toColor, 0.66);
 
-      assert.closeTo(interA._color.coords[0], 37, 1);
-      assert.closeTo(interA._color.coords[1], 60, 1);
-      assert.closeTo(interA._color.coords[2], 66, 1);
+      assert.closeTo(interA._color.coords[0], 352, 1);
+      assert.closeTo(interA._color.coords[1], 76, 1);
+      assert.closeTo(interA._color.coords[2], 75, 1);
 
-      assert.closeTo(interB._color.coords[0], 345, 1);
-      assert.closeTo(interB._color.coords[1], 20, 1);
-      assert.closeTo(interB._color.coords[2], 47, 1);
+      assert.closeTo(interB._color.coords[0], 301, 1);
+      assert.closeTo(interB._color.coords[1], 66, 1);
+      assert.closeTo(interB._color.coords[2], 65, 1);
     });
 
     test.todo('should not extrapolate', function() {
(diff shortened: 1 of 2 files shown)


pupuking723:
Thanks for taking a look at this.

I checked the current failing CI job. The only failing test I found is in `test/unit/webgl/p5.Framebuffer.js` for `saveCanvas`, where the assertion expected `document.createElement("a")` but the recorded calls were `canvas`. That appears unrelated to the `lerpColor()` / color-mode change in this PR.

For the color-related change, I verified the affected tests locally:

- `npm test -- --run test/unit/color/creating_reading.js test/unit/color/p5.Color.js`
- `npx eslint src/color/p5.Color.js test/unit/color/creating_reading.js`

One clarification on the test diff: I did not add a new regression test, but the existing HSL/HSB `lerpColor()` expectations encoded the old RGB interpolation behavior. After making the interpolation use the current color mode, those existing expectations needed to be updated to match the corrected behavior.

If you would prefer a source-only PR first, I can adjust the test changes based on your preference.

test/unit/color/creating_reading.js
```diff
@@ -71,33 +71,31 @@ suite('color/CreatingReading', function() {
...
 
     test('should correctly get lerp colors in HSL', function() {
-      // NOTE: This is equivalent to RGB case so is testing nothing new
       mockP5Prototype.colorMode(mockP5Prototype.HSL);
       var interA = mockP5Prototype.lerpColor(fromColor, toColor, 0.33);
       var interB = mockP5Prototype.lerpColor(fromColor, toColor, 0.66);
 
-      assert.closeTo(interA._color.coords[0], 37, 1);
-      assert.closeTo(interA._color.coords[1], 43, 1);
+      assert.closeTo(interA._color.coords[0], 352, 1);
```
ksen0, line 78: Why have the RGB test cases changed?

davepagurek: It looks like it's just HSL tests that have modifications here, and the one RGB test in between these if you expand some of the code diffs is unmodified. So it might be ok!

pupuking723: Thanks for checking this.

The RGB test between these cases was not changed. The changed expectations are for the HSL/HSB `lerpColor()` cases, because those existing assertions were encoding the old behavior where interpolation still happened through RGB space.

After the fix, those cases now interpolate in the active color mode, so the expected channel values changed. I agree the diff can look confusing because the unchanged RGB case sits between the modified cases.

limzykenneth:
This PR seems to only affect the lerp space and for that I would recommend not defining a lerp space in #8806. The original issue is more about the resulting `p5.Color` object not having the defined p5 color space when being created so I'm not sure this PR addresses it.

pupuking723:
Thanks, that distinction makes sense. I’ll pause the lerp-space change here and take another look at the output color-mode path separately before pushing more changes. I don’t want to make this PR broader than the bug needs.

pupuking723:
Closing this version because the current patch changes the interpolation space rather than the output `p5.Color` mode described in the issue. I will revisit the output-mode path separately before proposing another change.

PR #8933, opened 2026-06-17
fix: add _friendlyError to Vector constructor for static methods #8930

### LINKED ISSUE #8930: [p5.js 2.0+ Bug Report]: Some Vector friendly errors don't work correctly

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.3.0

### Web browser and version

Brave 1.90.122

### Operating system

Linux Mint 22.1

### Steps:
1. Run the snippet below. It has an error; the first param should be a vector.
2. The expected output is a friendly error message "The v1 parameter should be of type Array or p5.Vector"
3. Instead, it gives an error "TypeError: this._friendlyError is not a function"

### Snippet:

```js
function setup() {
  print(p5.Vector.equals(1, createVector(1, 2)));
}
```

### Discussion

Friendly errors for vectors seem to work fine for instance methods like `setValue()`, but not for static methods like `p5.Vector.equals`.

BHARATH0153:
Resolves #8930

## Changes

Static methods like `p5.Vector.equals()` call `this._friendlyError(...)` where `this` is the Vector constructor, but `_friendlyError` was only set on `Vector.prototype`. This caused `TypeError: this._friendlyError is not a function` instead of displaying the friendly error message.
 also set `Vector._friendlyError = p5._friendlyError` and override `Vector.friendlyErrorsDisabled` to check `p5.disableFriendlyErrors`, matching the existing prototype setup.

## PR Checklist

- [x] npm run lint passes


BHARATH0153:
@perminder-17 please review

ksen0:
The issue here is not as straightforward as it looks; tests do not pass.

Unfortunately, despite multiple maintainers asking you multiple times to follow community guidelines. You have also been asked to stop making PRs without assignment, to avoid a ban; I just checked and you have 👍  that request. Nevertheless, this PR was made before assignment anyway, so a temporary (1 month) block will be applied. Thanks for understanding, and I hope you can participate in the future.

image

PR #8936, opened 2026-06-18
Fix: Add property w JSDoc block after property  #8154

### LINKED ISSUE #8154: [2.0] Stabilize `w` field of `p5.Vector`
## [2.0] Stabilize `w` field of `p5.Vector`

A task list is below.
### 1. Decide whether to release (consider confusion and utility):

There are a couple of considerations.

**Potential confusion:**
Most users, especially beginners, will be unfamiliar with homogeneous coordinates, and will therefore be unfamiliar with the usual $(x, y, z, w)$ naming convention. They may guess that the order of coordinates is alphabetical, and this impression would be reinforced by the alphabetical ordering of the reference entries. 

This confusion could be mitigated by clear documentation on the `x`, `y`, `z`, and `w` reference pages. These would need to be clearer than they currently are (right now, the full documentation is simply "The x component of the vector" for `x`, and is similar for the other components). However, it may not be necessary to release `w` into the public API (see the next section).

**Limited utility?**
A proposed `Transform` class would abstract away the need for homogeneous coordinates, with methods like `applyToPoint()` and `applyToDirection()`. These use homogeneous coordinates internally to distinguish _points_ (which can be translated) from _directions_ (which cannot be translated), so the user doesn't ever need to touch a `w` coordinate. If we think they won't need to use `w` coordinates, then it may not be worth it to release them into the public API. On the other hand, if we can find practical, concrete use cases that wouldn't already be handled by `Transform`, then it may make sense to release `w` to users, since the confusion it might cause may not be especially great.

### 2. Fix documentation

**If we decide not to release `w`:**
We could add a `@private` tag to its docs.

**If we decide to release `w`:**
Currently, a documentation error prevents this feature from appearing in the beta docs. Properties declared with `get` and `set` are typically documented by documenting the getter only, using the `@type` tag, instead of `@returns`, which is [currently used](https://github.com/processing/p5.js/blob/39ea296ec08ddee96b88c3c65f8f31c95d6ef6d9/src/math/p5.Vector.js#L190). Once the getter documentation is fixed, the setter documentation may be deleted. At least, I think this should work.

vamshi567bh:
Resolves #8154

## Changes

Add `@property w` JSDoc block for `p5.Vector` alongside the existing `x`, `y`, `z` property blocks. The `w` component was already implemented as a getter/setter but was missing from the documentation definition, so it never appeared in the generated reference docs.

## PR Checklist

- [x] npm run lint passes
- [x] npm test passes
- [x] `npm run docs` generates correct data with `w` as a p5.Vector member
 

## screenshot 

done 2

### DIFF
--- a/src/math/p5.Vector.js
+++ b/src/math/p5.Vector.js
@@ -3685,6 +3685,14 @@ function vector(p5, fn) {
    */
 
   /**
+   * The w component of the vector
+   * @type {Number}
+   * @for p5.Vector
+   * @property w
+   * @name w
+   */
+
+  /**
    * The dimensions of the vector
    * @type {Number}
    * @for p5.Vector


vamshi567bh:
@ksen0 please check when you have time thanks!

ksen0:
Hi @vamshi567bh welcome! In the future please wait to be assigned before making the PR (this is part of the [contributor guidelines](https://p5js.org/contribute/contributor_guidelines/), please review those as well)

Please update issue description with a screenshot verifying the fix worked. [This](https://beta.p5js.org/contribute/contributing_to_the_p5js_reference/) is the guide for testing locally

Thanks!

vamshi567bh:
@ksen0  thansk for the replie i have upadted discription with locally tests passed thanks!

ksen0:
@vamshi567bh I meant screenshot of the website running locally and having the valid entry for `w` (should be different than https://beta.p5js.org/reference/p5/p5.Vector/w) thanks!

vamshi567bh:
@ksen0 really sorry for misunderstanding  please check now thanks!   

updated discription  thanks ! 
done 2
with this

vamshi567bh:
@davepagurek @ksen0 please review thanks!

PR #8947, opened 2026-06-20
fix mult() inconsistency: make 3x3 and NxN use left-multiplication like 4x4

vamshi567bh:
Changes
Makes #mult3x3 and #multNxN consistent with #mult4x4 by using left-multiplication (this = arg * this) instead of right-multiplication (this = this * arg).

PR Checklist

- [x] npm run lint passes successfully.
- [x] Inline reference documentation updated.
- [x] Unit tests included and passing.

### DIFF
--- a/src/strands/p5.strands.js
+++ b/src/strands/p5.strands.js
@@ -338,18 +338,20 @@ if (typeof p5 !== "undefined") {
  */
 
 /**
- * @method instanceID
+ * @property instanceIndex
  * @beta
  * @description
  * Returns the index of the current instance when drawing multiple copies of a
  * shape with <a href="#/p5/model">`model(count)`</a>. The first instance has an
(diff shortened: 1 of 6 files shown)


vamshi567bh:
@davepagurek please review thanks!

vamshi567bh:
https://github.com/processing/p5.js-website/pull/1460 related

davepagurek:
Does this actually address https://github.com/processing/p5.js-website/issues/1571 ? i.e. does it go through the mult3x3 or multNxN code paths there? 

It sounds like if these methods have inconsistent sidedness with 4x4 matrices, that it'd maybe be a separate issue. Before continuing down this path it would be good to raise an issue for it with a sketch showing the difference between e.g. a translate applied onto a scale via 2d and 3d affine matrices to see if it's different.

PR #8952, opened 2026-06-22
Add instances() API for primitive instancing

aashu2006:
Addresses #8911

**Changes**

* Add a new `instances(count)` API for primitive instancing in WebGL.
* Support instancing for primitives including `sphere()`, `box()`, `plane()`, `ellipsoid()`, `cylinder()`, `cone()`, and `torus()`.
* Preserve existing `model(geometry, count)` behavior while supporting `instances(count).model(geometry)`
* Forward instance counts through the existing retained-geometry rendering pipeline.
* Add TypeScript support for the new InstancesWrapper API.
* Add unit tests for primitive/model instancing and explicit count precedence.
* Add WebGL and WebGPU visual tests with reference snapshots.
* Update the documentation example to use `p5.strands` and `instanceIndex()`.

#### PR Checklist

* [x] `npm run lint` passes
* [ ] Inline reference is included / updated
* [x] Unit tests are included / updated


src/core/p5.Renderer3D.js
```diff
@@ -548,6 +550,11 @@ export class Renderer3D extends Renderer {
   }
 
   model(model, count = 1) {
+    // Use _instanceCount only when count was NOT explicitly passed.
+    // arguments.length distinguishes model(geom) from model(geom, 1).
+    if (arguments.length < 2 && this._instanceCount !== undefined) {
```
davepagurek: It's a tad unintuitive seeing a default value for `count` and having this condition -- at least, it's not obvious whether or not it's included. Thoughts on leaving out the default value for `count`, and then we can do something like:
```js
count = count ?? this._instanceCount ?? 1
``

File: src/webgl/3d_primitives.js [hunk 2]
```diff
@@ -2595,6 +2595,77 @@ function primitives3D(p5, fn){
...
+   *
+   * Call a draw method on the returned object to render that primitive
+   * `count` times in a single draw call. Instance-specific transforms and
+   * attributes (position, color, etc.) are supplied through a custom shader
+   * that reads per-instance data from an instanced attribute buffer.
+   *
+   * @method instances
+   * @param  {Number} count number of instances to draw. Must be a positive
+   *   integer.
+   * @returns {Object} an object with methods `sphere`, `box`, `plane`,
```
davepagurek: For TypeScript, if we just call it `Object` then all methods chained onto it will look like type errors. Take a look at the types added for the different p5.strands hooks in https://github.com/processing/p5.js/pull/8644, they add typedefs that add the different methods

src/webgl/3d_primitives.js
```diff
@@ -2595,6 +2595,77 @@ function primitives3D(p5, fn){
...
+   *   instances(10).sphere(20);
+   * }
+   * </code>
+   * </div>
+   */
+  fn.instances = function(count) {
+    this._assert3d('instances');
+
+    if (typeof count !== 'number' || !isFinite(count) || count < 1) {
+      console.log(
```
davepagurek: This should probably be an FES message so that it's disabled when FES is disabled

src/webgl/3d_primitives.js
```diff
@@ -2595,6 +2595,77 @@ function primitives3D(p5, fn){
...
+        '🌸 p5.js says: instances() requires a positive integer count.' +
+        ' Clamping to 1.'
+      );
+      count = Math.max(1, Math.round(count) || 1);
```
davepagurek: is there an input where we don't just set it directly to 1 here? might be more robust to do that

src/webgl/3d_primitives.js
```diff
@@ -2595,6 +2595,77 @@ function primitives3D(p5, fn){
...
+    // Each wrapped method: set _instanceCount, call the renderer method with
+    // the correct `this`, clear _instanceCount in finally so it never leaks.
+    const wrap = method => function(...args) {
+      r._instanceCount = count;
+      try {
+        method.apply(r, args);
+      } finally {
+        r._instanceCount = undefined;
+      }
+      return p;
```
davepagurek: Is the intention to be able to chain multiple things? If so I think this return value would need to be the returned object below instead of `p`. If not, we probably don't need to `return p` (or track `p` at all)

src/webgl/3d_primitives.js
```diff
@@ -2595,6 +2595,77 @@ function primitives3D(p5, fn){
...
+   * <div>
+   * <code>
+   * // Draw 10 spheres in a single instanced draw call.
+   * // A custom shader reads gl_InstanceID to offset each sphere.
+   * function setup() {
+   *   createCanvas(100, 100, WEBGL);
+   * }
+   * function draw() {
+   *   background(200);
+   *   instances(10).sphere(20);
```
davepagurek: These currently all draw on top of each other since now shader is present, can we add a p5.strands shader to make use of the instances?

src/webgl/3d_primitives.js
```diff
@@ -2595,6 +2595,77 @@ function primitives3D(p5, fn){
...
+   *   integer.
+   * @returns {Object} an object with methods `sphere`, `box`, `plane`,
+   *   `ellipsoid`, `cylinder`, `cone`, `torus`, and `model`. Call one of
+   *   these to draw `count` instances of that primitive.
+   *
+   * @example
+   * <div>
+   * <code>
+   * // Draw 10 spheres in a single instanced draw call.
+   * // A custom shader reads gl_InstanceID to offset each sphere.
```
davepagurek: We should probably use strands and not mention GLSL-specific things

davepagurek (commented):
This is looking good! I left some comments. Once the example in the docs is more complete, could we add it as a visual test in both webgl and webgpu modes to have an end-to-end test for each?

src/core/p5.Renderer3D.js
aashu2006: I updated this to use `count = count ?? this._instanceCount ?? 1;` instead.

src/webgl/3d_primitives.js
aashu2006: I did follow the pattern from #8644 and updated the typedefs, this also required a small fix in the type generator so the optional parameters are generated correctly.

src/webgl/3d_primitives.js
aashu2006: updated this to use the Friendly Error System.

src/webgl/3d_primitives.js
aashu2006: simplified this to clamp directly to 1.

src/webgl/3d_primitives.js
aashu2006: updated the wording to use the strands API.

src/webgl/3d_primitives.js
aashu2006: updated the example to use `p5.strands` and `instanceIndex` so the instancing behavior is visible.

aashu2006:
@davepagurek Done! I used the updated example as the basis and added the corresponding visual tests for both WebGL and WebGPU.

src/webgl/3d_primitives.js
aashu2006: I made it a terminal draw call and removed the return value to avoid that misleading chaining behavior.

davepagurek (commented):
Thanks for the updates! Some thoughts:
- Should the parameters be recursively passed into `convertTypeToTypeScript` to handle all the same cases there? (Does doing that break anything?)
- For the visual tests to work, you'll need to run it first locally and commit the added snapshots so that CI has something to compare to in the future.
- There are some other shape drawing methods that aren't added yet, such as `plane`, `line`, etc. Maybe worth looking through the `3d_primitives` file and seeing if anything else there should be added?

aashu2006:
thanks for the quick feedback @davepagurek 

I pushed another update, also added the visual test snapshots this time, my bad for missing them in the previous push.

I looked into the `convertTypeToTypeScript()` suggestion as well. It already handles `OptionalType` recursively, but `convertFunctionTypeForInterface()` still needs the parameter name and optionality first, so I kept the unwrapping there before passing the type through `convertTypeToTypeScript()`.

I also took another pass through `3d_primitives.js`. The omitted ones like `line()` use the immediate-mode path, so they don't make use of `_instanceCount`. I did notice a few retained-geometry shapes like `triangle()`, `rect()`, `quad()`, and `ellipse()`/`arc()` could technically be supported too. Would you want `instances()` to cover those as well, or keep it focused on the current set?

davepagurek:
Sounds good, thanks for looking into the typescript stuff!

Yeah, let's see if we can support all the shapes if we can.

aashu2006:
Thanks, I updated the remaining retained-geometry shapes as well. I think that should cover everything from my side now, but happy to make any further changes if needed 🙂

davepagurek (commented):
thanks for the updates! I think we can support `line` and other APIs that go through begin/endShape by updating the `count` parameter in `endShape` to do what you've done for `model`.

As a list thing, I think we use `instanceID()` in some examples. If you search for `instanceID()` in the codebase to find those, can we update those to use your new syntax and avoid `buildGeometry` if they're just drawing a sphere or something to simplify them?

aashu2006:
I added support for immediate mode APIs by forwarding instance count through `endShape()`, so `line()`, `point()`, `bezier()`, and `spline()` now work with `instances()`. I also updated the examples to use `instanceIndex` and `instances()`, and added visual tests for both WebGL and WebGPU.

For `curve()`, I kept it conditional since `spline()` is the standard API in `2.x`, while `curve()` is only available in compatibility setups.

test/unit/visual/cases/webgl.js
```diff
@@ -1101,6 +1101,64 @@ visualSuite('WebGL', function() {
...
+      p5.fill('blue');
+      p5.shader(shader);
+      p5.instances(count).rect(-5, -5, 10, 10);
+      screenshot();
+    });
+
+    visualTest('instances() API draws immediate-mode shape primitives', (p5, screenshot) => {
+      p5.createCanvas(50, 50, p5.WEBGL);
+      const count = 4;
+      const shader = p5.baseMaterialShader().modify(() => {
```
davepagurek: material shaders don't apply to lines (you'll need `buildStrokeShader` for that), so I think this is currently only applying to the fills in the bezier/spline while the strokes stay in the center. You can pass the same callback into both a material shader and a stroke shader to make two, and then you can call `shader(myMaterialShader)` and `strokeShader(myStrokeShader)` to have both apply at once.

Looking at the image for this, it looked slightly off to me but it wasn't super clear, maybe that's a sign it should be split up into multiple images/tests? (You can also call `screenshot()` multiple times in a test after clearing and redrawing if you want to keep the test code concise; itll just put multiple screenshot images into the folder.)

aashu2006: Just pushed an update - switched the tests to use `buildStrokeShader()`, split them into separate cases so they are a bit clearer, updated the examples too, and regenerated the reference snapshots. thanks!

davepagurek (approved):
Thanks for the updates, looks good now!

PR #8954, opened 2026-06-22
Add matrix transformation builtins to strands

### LINKED ISSUE #8953: Support transformation matrices in p5.strands
### Increasing access

If you want to use WebGL/WebGPU instancing to draw many particles, then you end up having to position the particles in a p5.strands shader rather than using the usual translate/rotate/scale functions. This gets complex pretty quickly, making it hard to take advantage of the performance opportunities of shaders. It would be great to make it as close as we can to the ease of use of regular p5 positioning.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

Here are some examples of what it looks like right now:
- https://openprocessing.org/@davepagurek/2845379
- https://editor.p5js.org/davepagurek/sketches/Y4gEVBKaL
- https://openprocessing.org/@davepagurek/2838235

Notably, that last one uses rotations, and is really complicated!

So we can't make this look *exactly* like regular p5 transformations because we're working at a different level: normally we're moving space around ourselves and then drawing something there; in a shader, we're looking at an individual vertex at a time and updating its position/normal. 

However, I think we can do something fairly close, taking inspiration from the [web `DOMMatrix` API.](https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrixReadOnly) Here's a comparison of using transforms in p5 to applying a `DOMMatrix` to a point:

<table>
<tr>
<th>

p5/shape

</th><th>

`DOMMatrix`/point

</th><th>

`DOMMatrixReadOnly`/point

</th></tr>
<tr>
<td>

```js
translate(25, 25)
rotate(-PI * 0.2)
translate(-25, -25)
scale(2)
model(something)
```

</td>
<td>

```js
const t = new DOMMatrix()
t.translateSelf(25, 25)
t.rotateSelf(36) // DOMMatrix uses degrees by default
t.translateSelf(-25, -25)
t.scaleSelf(2)
const transformed = t.transformPoint(somePoint)
```

</td>
<td>

```js
const t = new DOMMatrix()
  .translate(25, 25)
  .rotate(36) // DOMMatrix uses degrees by default
  .translate(-25, -25)
  .scale(2)
const transformed = t.transformPoint(somePoint)
```

</td>
</tr>
</table>

A matrix should ideally have these methods from `DOMMatrix`:
- `translate`
- `rotate` (and `rotateAxisAngle`)
- `scale`
- `skewX` and `skewY`
- `inverse`
- `multiply`
- `transformPoint`

Additionally, I think we also need `transformNormal`. In 3D, to make lighting work, we also need to update the orientation of the surface. This involves taking the transpose of the upper-left 3x3 submatrix of the 4x4 matrix of a 3D affine transform and then applying that to the normal vector.

### Note: to mutate or not to mutate?

I don't think we'd want to have the `*Self` suffixes like in `DOMMatrix`. I feel like we should use the un-suffixed names, but decide whether they mutate in place, or return a modified version.

Here's a little pro/con list of mutation:

<table>
<tr><th>For mutation</th><th>Against mutation</th></tr>
<tr>
<td>

- `p5.Vector` methods mutate
- Method chainining is not something we see much of in p5 code, and doing `t = t.multiply(something)` each line to avoid chaining is a little more verbose

</td>
<td>

- p5.strands vectors *don't* do any mutation in place -- they mirror GLSL vectors, where they are data like numbers in js, and global functions (not methods!) like `cross(v1, v2)` or `normalize(v)` return new vectors -- so we would be consistent within strands
- avoids a common bug where you accidentally mutate an object that you intend to reuse in another calculation

</td>
</tr>
</table>

I personally lean towards *not* mutating in p5.strands for consistency within strands. To make it less confusing with `p5.Vector`, I propose **not making methods on matrices in strands**, similar to vectors in strands, and making global methods called `translate(...)`, `scale(...)`, etc. We can reuse the same methods that in regular p5 mutate the current coordinate space, and the idea would be that because there is no global coordinate space you can mutate in strands, you use those methods slightly differently in the strands context, and then there's also no confusion in p5 about whether they'll mutate the objects, since `p5.Vector` static methods also don't mutate.

In the future, matrices outside of strands might get methods that do mutate.

### Note: matrices or transforms?

While in GLSL/WGSL these are matrices, they are used almost exclusively for transforms. I think we still want to support `mat3(...)`, `mat4(...)` etc constructors for GLSL compatibility, but as aliases for `transform2D(...)` and `transform3D(...)` respectively. (Multiplying a 3D transform, which is a `mat4`, against a `vec3`, via `transformPoint` would under the hood add a `1` to its end to make it an affine point. We can maybe let the `*` operator do "raw" multiplication.)

This terminology keeps the door open for us to have more math-oriented matrices in p5, which have different assumptions (and performance tradeoffs) than ones intended for graphics. In the future we could have a non-strands transform with a similar API (a rename of the current `p5.Matrix`, keeping its transform/graphics methods?) alongside a non-strands `p5.Matrix` too.

vamshi567bh:
Resolves #8953

## Changes:
Adds 8 matrix transformation builtins (transform2D, transform3D, translate, rotate, scale, multiply, transformPoint, transformNormal) as GLSL/WGSL function snippets injected into strands shader declarations. Includes matrix type aliases (Mat4x4 → Mat4, etc.) and constructor aliases (mat4x4 → mat4, etc.), plus * operator overload support for mat×mat, mat×vec, and vec×mat multiplication.

- `src/strands/ir_builders.js` – MAT type constructor handling, matrix multiplication in binaryOpNode
- `src/strands/strands_builtins.js` – All matrix function overloads registered
- `src/strands/strands_api.js` – Matrix type and constructor aliases
- `src/strands/p5.strands.js` – Matrix snippet injection into shader declarations
- `src/webgl/shaders/functions/matrixTransformGLSL.glsl` – GLSL implementations
- `src/webgpu/shaders/functions/matrixTransformWGSL.js` – WGSL implementations

## PR Checklist
- [x] `npm run lint` passes

### DIFF
--- a/src/strands/ir_builders.js
+++ b/src/strands/ir_builders.js
@@ -101,7 +101,6 @@ export function binaryOpNode(strandsContext, leftStrandsNode, rightArg, opCode) 
   if (opCode === OpCode.Binary.MULTIPLY) {
     if (leftType.baseType === BaseType.MAT) {
       if (rightType.baseType === BaseType.MAT && rightType.dimension === leftType.dimension) {
-        // mat × mat → mat
         const nodeData = DAG.createNodeData({
           nodeType: NodeType.OPERATION,
           opCode,
@@ -114,7 +113,6 @@ export function binaryOpNode(strandsContext, leftStrandsNode, rightArg, opCode) 
         return { id, dimension: leftType.dimension };
       }
       if (rightType.baseType === BaseType.FLOAT && rightType.dimension === leftType.dimension) {
-        // mat × vec → vec (float)
         const nodeData = DAG.createNodeData({
           nodeType: NodeType.OPERATION,
           opCode,
@@ -129,7 +127,6 @@ export function binaryOpNode(strandsContext, leftStrandsNode, rightArg, opCode) 
     }
     if (rightType.baseType === BaseType.MAT && leftType.baseType === BaseType.FLOAT &&
         leftType.dimension === rightType.dimension) {
-      // vec × mat → vec (float)
       const nodeData = DAG.createNodeData({
         nodeType: NodeType.OPERATION,
         opCode,
--- a/src/strands/ir_builders.js
+++ b/src/strands/ir_builders.js
@@ -97,6 +97,52 @@ export function binaryOpNode(strandsContext, leftStrandsNode, rightArg, opCode) 
     DAG.propagateTypeToAssignOnUse(dag, rightStrandsNode.id, leftType.baseType, leftType.dimension);
     rightType = DAG.extractNodeTypeInfo(dag, rightStrandsNode.id);
   }
+
+  if (opCode === OpCode.Binary.MULTIPLY) {
+    if (leftType.baseType === BaseType.MAT) {
+      if (rightType.baseType === BaseType.MAT && rightType.dimension === leftType.dimension) {
+        // mat × mat → mat
+        const nodeData = DAG.createNodeData({
+          nodeType: NodeType.OPERATION,
+          opCode,
+          dependsOn: [leftStrandsNode.id, rightStrandsNode.id],
+          baseType: BaseType.MAT,
+          dimension: leftType.dimension,
+        });
+        const id = DAG.getOrCreateNode(dag, nodeData);
+        CFG.recordInBasicBlock(cfg, cfg.currentBlock, id);
+        return { id, dimension: leftType.dimension };
+      }
+      if (rightType.baseType === BaseType.FLOAT && rightType.dimension === leftType.dimension) {
+        // mat × vec → vec (float)
+        const nodeData = DAG.createNodeData({
+          nodeType: NodeType.OPERATION,
+          opCode,
+          dependsOn: [leftStrandsNode.id, rightStrandsNode.id],
+          baseType: BaseType.FLOAT,
+          dimension: leftType.dimension,
+        });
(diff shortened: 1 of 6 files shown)


vamshi567bh:
Hey @davepagurek, could you please take a look at this when you get a chance? We have implemented the matrix transformation builtins (transform2D, transform3D, translate, rotate, scale, multiply, transformPoint, transformNormal) for strands. please review

ksen0:
Hi @vamshi567bh thanks for the work, but please hold off from making multiple simultaneous PRs, and take care to follow [contributor guidelines](https://p5js.org/contribute/contributor_guidelines/), including the [AI usage policy](https://github.com/processing/p5.js/blob/main/AI_USAGE_POLICY.md). I'll close this one for how, because it did not wait for assignment, is much larger without explanation, and your comment includes _"We have implemented the matrix transformation builtins for strands based on the discussion here. Here is what we did:"_(who is we?). Please carefully read existing discussion and be considerate and patient with others working on the project, who are mostly volunteers. Any future PRs that do not align with the guidelines or the [code of conduct](https://p5js.org/code-of-conduct/) may result in a temporary (30 day) suspension of the account. Thanks in advance for understanding

PR #8955, opened 2026-06-24
GSOC 26: render multi-material models per part

Nixxx19:
part of my google summer of code 2026 project, "full texture support for .mtl files in p5.js". works toward #6924.

this is phase 3: the renderer side, building on the data + parser foundation in #8879. multi-material models now actually draw, one part per material, each with its own look. public api is unchanged, you still just call loadModel() and model().

project phases for context:
- phase 1: design + pipeline study
- phase 2: data model + obj/mtl parser (#8879)
- phase 3: per-part rendering (this PR)

what's in here:
- the renderer loops over a geometry's parts and draws each one
- each part applies its own material (fill, specular, ambient, shininess, texture) around its draw, restored after so nothing leaks between parts
- a single-material geometry is its own part, so that path renders exactly as before, byte for byte
- p5.GeometryPart gains what the renderer needs to treat it as a drawable (hasFillTransparency, userVertexProperties)

tests:
- a multi-material model (octa-color, 8 materials) draws one fill per part and passes the instance count through
- the existing obj+mtl diffuse-colour visual test still matches its reference screenshot
- full unit suite green, including the webgpu and visual (pixel-compare) tests

stacked on #8879, so it should go in after that one merges.


src/core/p5.Renderer3D.js
```diff
@@ -589,7 +589,28 @@ export class Renderer3D extends Renderer {
...
       ![constants.LINES, constants.POINTS].includes(mode)
     ) {
-      this._drawFills(geometry, { mode, count });
+      // draw every part. a part with no material state draws straight (no
+      // push/pop); a single-material geometry is its own part, so that case is
+      // exactly the old single draw. multi-material parts each apply their own
+      // material around the draw.
+      const parts = geometry.parts && geometry.parts.length
+        ? geometry.parts
+        : [geometry];
```
davepagurek: What's a scenario where we'd end up with no parts?

src/core/p5.Renderer3D.js
```diff
@@ -628,6 +649,31 @@ export class Renderer3D extends Renderer {
     shader.unbindShader();
   }
 
+  // apply a part's material to the renderer before it's drawn. only non-null
+  // fields are set, so an empty part state leaves the uniforms untouched.
+  _applyPartState(partState) {
+    if (!partState) return;
+    if (partState.fill) {
+      const c = partState.fill;
+      this.states.setValue('curFillColor', [c[0], c[1], c[2], 1]);
```
davepagurek: Is there a reason why it's not possible for a part to have alpha in the fill? Although the obj/mtl spec puts transparency into its own command, in p5 fills can have alpha (e.g. geometry that comes from `buildGeometry` would be able to have fills with alpha if a user calls something like `fill(255, 100)` in the middle of building it.) Maybe we can store the transparency from a mtl in the alpha field of its fill color?

test/unit/webgl/p5.GeometryPart.js
```diff
@@ -0,0 +1,112 @@
...
+      }, myp5._renderer);
+      expect(geom.parts.length).toEqual(1);
+      expect(geom.parts[0].vertices.length).toEqual(1);
+    });
+  });
+
+  suite('multi-material rendering', function() {
+    test('draws each part and passes the instance count through', async function() {
+      const renderer = myp5.createCanvas(50, 50, myp5.WEBGL);
+      const model = await new Promise(resolve =>
```
davepagurek: I think `loadModel(url)` should already returns a promise, without needing to wrap it further

test/unit/webgl/p5.GeometryPart.js
```diff
@@ -0,0 +1,112 @@
...
+        myp5.loadModel('/test/unit/assets/octa-color.obj', resolve));
+
+      // octa-color has several materials, so several parts.
+      expect(model.parts.length).toBeGreaterThan(1);
+
+      const spy = vi.spyOn(renderer, '_drawFills');
+      myp5.background(255);
+      myp5.model(model, 4);
```
davepagurek: The second parameter here is for WebGL instanced rendering, is that intended to be set in this test?

src/webgl/p5.GeometryPart.js
```diff
@@ -0,0 +1,60 @@
...
+  hasFillTransparency() {
+    for (let i = 3; i < this.vertexColors.length; i += 4) {
+      if (this.vertexColors[i] < 1) return true;
+    }
+    return false;
+  }
+}
+
+function geometryPart(p5, fn) {
+  p5.GeometryPart = GeometryPart;
```
davepagurek: I think this maybe could be done in the existing `geometry` addon, since there's not a way to use `geometry` without `geometryPart` currently. So this file could just export the class, and `p5.Geometry` can import it and add it to the `p5` object

test/unit/webgl/p5.GeometryPart.js
```diff
@@ -0,0 +1,112 @@
...
+  suite('multi-material rendering', function() {
+    test('draws each part and passes the instance count through', async function() {
+      const renderer = myp5.createCanvas(50, 50, myp5.WEBGL);
+      const model = await new Promise(resolve =>
+        myp5.loadModel('/test/unit/assets/octa-color.obj', resolve));
+
+      // octa-color has several materials, so several parts.
+      expect(model.parts.length).toBeGreaterThan(1);
+
+      const spy = vi.spyOn(renderer, '_drawFills');
```
davepagurek: We should have a similar spy and assertion for `_drawStrokes`, which I believe we still need to implement.

davepagurek (commented):
Looking good! I left a few comments. One thing that would be good to clarify is what the valid states of a geometry are (should they always have parts?) so that that can be an invariant in the code and simplifies implementation.

src/core/p5.Renderer3D.js
Nixxx19: there isn't one. the constructor always sets parts = [this] and loaders replace it, so a geometry always has at least one part. i'll make that an invariant and just loop geometry.parts directly, dropping the fallback

src/core/p5.Renderer3D.js
Nixxx19: yeah that makes sense. fills carry alpha in p5, so i'll store the mtl d/Tr transparency in the fill colour's alpha (4th component) instead of forcing 1. that also covers the buildGeometry fill(255,100) case. nice, this brings transparency back in the right place

test/unit/webgl/p5.GeometryPart.js
Nixxx19: right, will simplify to const model = await myp5.loadModel('/test/unit/assets/octa-color.obj') and drop the wrapper

test/unit/webgl/p5.GeometryPart.js
Nixxx19: the 4 was to confirm the instance count reaches each part's draw, but you're right it muddies this test, i'll split it, assert per-part drawing on its own and keep the instancing check separate

src/webgl/p5.GeometryPart.js
Nixxx19: agreed, no reason for a separate addon. i'll have p5.GeometryPart.js just export the class and let p5.Geometry import it and add it to the p5 object, dropping the extra registerAddon

test/unit/webgl/p5.GeometryPart.js
Nixxx19: good catch, strokes aren't per-part yet, only fills. i'll add per-part stroke drawing and a _drawStrokes spy + assertion alongside the fills one

Nixxx19:
> Looking good! I left a few comments. One thing that would be good to clarify is what the valid states of a geometry are (should they always have parts?) so that that can be an invariant in the code and simplifies implementation.

thanks for the review! on the invariant, yeah a geometry should always have at least one part, the constructor guarantees it (parts = [this]) and loaders replace it. i'll lean into that so the renderer just loops parts with no fallback, like you said, addressing the rest now

Nixxx19:
heads up, the failing check is just flaky ci, not anything in this pr.

src/core/p5.Renderer3D.js
```diff
@@ -589,7 +589,25 @@ export class Renderer3D extends Renderer {
...
-      this._drawFills(geometry, { mode, count });
+      // draw every part. a geometry always has at least one part (the
+      // constructor makes it its own single part when nothing else does), so no
+      // fallback is needed. a part with no material state draws straight;
+      // multi-material parts apply their own material around the draw.
+      for (const part of geometry.parts) {
+        const state = part.partState;
+        const hasMaterial = state && (
+          state.fill || state.texture || state.ambientColor ||
+          state.specularColor || state.shininess != null
```
davepagurek: Heads up, `!=` will consider 0 to be equivalent to `null` as both are falsy. Do you want `!==` here? (Also worth double checking that one of these could have valid falsy values -- to confirm, are `fill` and `ambientColor` p5.Color values or arrays?)

src/webgl/loading.js
```diff
@@ -17,6 +18,161 @@ async function fileExists(url) {
...
+
+// mtl material -> part state in p5's vocab. anything we can't draw yet is left
+// off until support lands.
+function mtlToPartState(material) {
+  const state = createPartState();
+  if (!material) return state;
+  if (material.diffuseColor) {
+    // carry the mtl transparency (d / Tr) in the fill's alpha. opaque materials
+    // stay rgb so nothing changes for them.
+    state.fill = material.opacity != null && material.opacity < 1
```
davepagurek: It might simplify things further to have the fill always be in the same format. So you could do `[...material.diffuseColor, material.opacity ?? 255]` assuming a 0-255 range. It's also not obvious what the data type and range should be for each of these so it's maybe worth adding in a comment where part state is first defined what the type and range of each field is, and then ensuring that they are always that type (e.g. `[number, number, number, number] | null` where the numbers are 0-255)

src/core/p5.Renderer3D.js
Nixxx19: these are all plain arrays, not p5.Color, fill is [r,g,b,a], ambient/specular are [r,g,b], texture's a p5.Image. shininess is the only number, and 0 != null is already true so a 0 still counts as set. made them all != null anyway so it stays consistent

src/webgl/loading.js
Nixxx19: agreed, fill's always [r,g,b,a] now. one thing, the range is 0..1 not 0..255 (curFillColor's 0..1, and Kd/d parse straight to floats), so it came out [...diffuseColor, opacity ?? 1]. also documented the type/range of each field on createPartState, and dropped the alpha padding in _applyPartState + hasFillTransparency now that fill's 4 wide

src/webgl/p5.GeometryPart.js
```diff
@@ -0,0 +1,54 @@
...
+ * @submodule 3D Primitives
+ * @for p5
+ */
+
+// fresh part state. fields use p5 names (fill, texture...), not obj/mtl tokens.
+// importers translate into this and drop anything we can't draw yet. every color
+// channel is 0..1 (same range as the renderer's curFillColor), not 0..255.
+function createPartState() {
+  return {
+    fill: null,           // Kd + d -> [r, g, b, a] | null, each 0..1
```
davepagurek, line 12: Nice!

davepagurek (approved):
Thanks for your patience, changes look good! We just did a patch release so I think we're ok to merge this + the previous PR to main now!

PR #8960, opened 2026-06-25
Fix/null undefined validation

### LINKED ISSUE #8407: [p5.js 2.0 Bug Report]: noise(null) is treated as noise(0) while noise(undefined) errors, and this behaviour is undocumented and inconsistent.

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [x] Friendly errors

### p5.js version

2.1.2

### Web browser and version

Google Chrome Version 143.0.7499.170

### Operating system

Windows 11

### Steps:
1. Run the sketch below
2. Compare the outputs of noise(null) and noise(0)
3. Observe that they are identical, while noise(undefined) throws an FES warning

### Snippet:

```js

function setup() {
  noCanvas();

  console.log("null:", noise(null));
  console.log("0:", noise(0));
  console.log("undefined:", noise(undefined));
}


```

**Explanation:**

noise(null) returns the same value as noise(0), while noise(undefined) produces a NaN and a Friendly Error message. Both null and undefined represent missing values in JavaScript, but they are handled differently here. This behaviour is not documented and can lead to silent bugs when null is passed unintentionally.

_I'm happy to take on this task, if this is approved_

jeetsdev:
Resolves #8407 

 Changes:

 This PR fixes the parameter validation behavior for noise() in two ways:

- Fixed the FES so that parameter validation warnings are emitted when all arguments are undefined.
- Updated noise() to treat null as an invalid argument by returning NaN, preventing it from being implicitly coerced to 0.

Testing: 

- noise(null) triggers a parameter validation warning.
- noise(null) returns NaN.

#### PR Checklist

- [x] `npm run lint` passes


src/core/friendly_errors/param_validator.js
```diff
@@ -551,7 +551,7 @@ function validateParams(p5, fn, lifecycles) {
       args.every(arg => arg === undefined)
     ) {
       const undefinedErrorMessage = `🌸 p5.js says: All arguments for ${func}() are undefined. There is likely an error in the code.`;
-
+      console.log(undefinedErrorMessage);
```
limzykenneth, line 554: No, this function is not meant to print any message directly.

limzykenneth (changes requested):
The parameter validation part will need a deeper review here.

PR #8965, opened 2026-07-05
Add p5.strands documentation and examples for color functions

### LINKED ISSUE #8915: Add p5.strands documentation and examples for color functions
### Increasing access

Beginners using p5.strands need to know that color functions behave differently inside shader callbacks. Without strands-specific examples in the docs, it's not obvious how to use these functions in a `modify()` block.

- [ ] Accessibility
- [x] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

Now that `color()`, `lerpColor()`, `red()`, `green()`, `blue()`, `alpha()`, `hue()`, `saturation()`, `brightness()`, and `lightness()` work inside p5.strands shader callbacks (added in #8822), the documentation should include strands-specific examples showing how to use these functions in a `modify()` callback.

Key things to document:

- In strands, `color()` returns a `vec4` instead of a `p5.Color` object
- All color values are represented as normalized RGBA values in the 0–1 range
- `colorMode()` has no effect inside strands callbacks
- Common usage patterns for `lerpColor()`, color channel accessors, and HSL/HSB accessors in shader code

Potential follow-up documentation work:

- Update existing p5.strands examples that currently use raw color vectors where the new `color()` API would improve readability
- Add example sketches demonstrating hex strings, CSS color names, and `rgb()/rgba()` inputs inside strands
- Review existing strands examples and helper functions for opportunities to showcase the newly added color APIs

**Related PR:** #8822

Yashisinghal285:
Closes #8915

Adds p5.strands-specific documentation and examples for color(), green(), blue(), alpha(), hue(), saturation(), brightness(), and lightness() in src/color/creating_reading.js. (red() and lerpColor() already had strand notes/examples from #8822.)

Each function's JSDoc now includes:
- A short note that in strands shader callbacks, the function operates on vec4 values with components normalized to the 0–1 range
- A note that colorMode() has no effect inside shader callbacks
- A minimal @example showing usage inside a baseMaterialShader().modify() callback, following the pattern established in #8822

Examples were kept minimal and focused per the style guide, as discussed in the issue thread.

cc @LalitNarayanYadav for review


LalitNarayanYadav:
Thanks for the PR @Yashisinghal285! Good start - the strands descriptions for each function are accurate. A few things to fix before this is ready:

## 1. Examples are incomplete/incorrect

- The strands examples don't have `createCanvas()` or a working sketch structure.
- `p5.getPixelInputs` should be `getPixelInputs` (no `p5.` prefix inside `modify()`).
- The examples also need `{ myp5 }` scope (or similar), following the existing strands examples in the test file.

## 2. `color()` strands example issue

The example returns `c` from inside `getPixelInputs`, but `getPixelInputs` expects the **inputs object** to be returned, not a color directly.

Instead, it should be:

```js
inputs.color = c;
return inputs;
```

## 3. Missing blank line before strands paragraph

For `brightness()` and `lightness()`, the strands paragraph is placed immediately after the existing text without a blank line. This may break the JSDoc formatting.

## 4. `lerpColor()` example is incomplete

The strands example creates `c1`, `c2`, and `mixed`, but doesn't actually do anything with `mixed`—there's no hook or return value demonstrating its use.

## 5. Style guide

Per the p5.js documentation style guide, examples should be:

- Minimal
- Self-contained
- Include a `describe()` call

Happy to clarify any of these if needed!

Yashisinghal285:
Thanks @LalitNarayanYadav  for the detailed feedback! Fixed all 5 points:
1. Examples now include createCanvas(), setup()/draw(), and getPixelInputs (no p5. prefix)
2. color() now does inputs.color = c; return inputs;
3. Added the missing blank line before the strands paragraph in brightness() and lightness()
4. lerpColor() now actually applies `mixed` via inputs.color
5. All examples are minimal, self-contained, and include describe()

On the {myp5} scope note — checked test/unit/assets/testMaterial.js, which uses global-mode getPixelInputs() (no prefix), matching what I used and consistent with the rest of this file's examples. Let me know if you meant something different.

LalitNarayanYadav:
Thanks for the updates @Yashisinghal285! Getting closer, but a few things still need fixing:

## 1. Examples still won't run

The strands examples are missing `createCanvas(100, 100, WEBGL)` and the shader setup. A minimal working example should look like:

```js
function setup() {

  let myShader = baseMaterialShader().modify(() => {
    getFinalColor((col) => {
      let c = color('#ff0000');
      col.rgb = c.rgb;
      return col;
    });
  });

  shader(myShader);
  noStroke();
  plane(100, 100);

  describe('A red plane');
}
```

## 2. `getPixelInputs` vs `getFinalColor`

The examples use `getPixelInputs`, but for color-related functions `getFinalColor` is more appropriate and consistent with the example sketch from the microgrant.

## 3. `describe()` is missing

Every example needs a `describe()` call per the p5.js documentation style guide.

## 4. Strands paragraph placement

The strands note should go just before the `@method` tag, not mixed in with the existing description text. Check how it was done in `color()` in PR #8822 as a reference.

## 5. Some functions still missing strands notes

`brightness()` and `lightness()` don't seem to have strands paragraphs added yet.

Suggest looking at the working example sketch at <https://editor.p5js.org/LalitNarayanYadav/sketches/YMy8k8QSF> as a reference for how these functions work together in practice.

Yashisinghal285:
Thanks @LalitNarayanYadav   — 
Fixed — getFinalColor needs return vec4(...) explicitly, not col.rgb = .... Verified working in the editor: [your link]. Updated all strands examples (color, red, green, blue, alpha, hue, saturation, brightness, lightness, lerpColor) to match. Also double-checked — brightness()/lightness() already have their strand notes, and every example has describe()

LalitNarayanYadav:
Thanks for addressing the feedback, @Yashisinghal285! The switch to `getFinalColor` with an explicit `vec4(...)` return looks correct, and it's great to hear you've verified it in the editor. If `describe()` has been added to all examples and the `brightness()`/`lightness()` documentation is now covered, everything looks good from my side.

@davepagurek @ksen0, this should be ready for maintainer review whenever you have a chance.

LalitNarayanYadav (approved):

src/color/creating_reading.js
```diff
@@ -396,12 +419,37 @@ function creatingReading(p5, fn){
...
    * @param {p5.Color|Number[]|String} color <a href="/reference/p5/p5.Color/">p5.Color</a> object, array of
    *                                         color components, or CSS color string.
    * @return {Number} the red value.
    *
    * @example
+   * let myShader;
+   * function setup() {
+   *   createCanvas(100, 100, WEBGL);
+   *
+   *   myShader = baseMaterialShader().modify(() => {
```
davepagurek: A p5.strands example should not be the first thing users see for `red`, let's put it at the end. We should maybe also move the explanation for how it works with p5.strands to the end too, right above it.

In order to do that, we need to switch from using these `@example` blocks to using three-backtick code fences. [See this description in our contributor docs](https://github.com/processing/p5.js/blob/main/contributor_docs/contributing_to_the_p5js_reference.md#inserting-examples-within-the-description) for what that looks like. This means we'll need to convert the existing examples for this reference into that format in order to be able to put text between the examples to add the explanation for p5.strands.

davepagurek (changes requested):
Hi! If you take a look at the other examples on beta.p5js.org/reference, you'll notice they follow some different conventions.

- We don't use `baseMaterialShader().modify(...)`; instead, we made `buildMaterialShader(...)` methods and use those in all the docs. The rationale is that the extra syntax provided more areas for users to get tripped up without giving much benefit, so we introduce people instead to the simpler single function (while still doing the more complex one under the hood internally.)
- We try not to use inline anonymous functions in shader callbacks. This works, but it's again a question of minimizing the new things readers have to understand in order to read the example. Since users will already by familiar with named top-level functions in order to create the `function setup` and `function draw` of p5, in examples, we prefer creating an additional top-level function like `function myMaterial` and then calling `buildMaterialShader(myMaterial)`.
- Similar to the above, we have moved away from using the callback syntax to fill out the hooks in a shader. Nested functions provide an additional layer of syntactic complexity for users to understand, so rather than using `getFinalColor(() => { ... })`, our p5.strands examples universally use a `finalColor.begin(); ...; finalColor.end()` syntax instead. Where a callback would return a value, we use `finalColor.set(...)`. Functions that just modify inputs (e.g. `worldInputs`) do not need to call `.set` at all.

For examples, we want to make sure that they actually run successfully and display the thing we want them to display. For that reason, rather than just writing it and trying to imagine what they do, we ask you to clone the p5.js-website repo, run it locally using your build of p5, and show us a screenshot of the results in a comment. [See our contributor docs here for instructions on how to do that.](https://github.com/processing/p5.js/blob/main/contributor_docs/contributing_to_the_p5js_reference.md#generating-and-previewing-the-reference) This lets us also discuss whether anything could be visually improved, not just in the code.

Lastly, we also want to make sure that users aren't bogged down by all the possible ways one can use a function before seeing examples. For that reason, and because p5.strands is an advanced use case that most won't initially use, we should avoid adding text to the first big block of explanations above the simple examples, and instead put the explanation of p5.strands right before the relevant p5.strands example at the end of the reference page. To be able to have text in between examples, we will need to convert the examples for a reference entry from `@example`s to three-backtick-fenced examples. [See the explanation here](https://github.com/processing/p5.js/blob/main/contributor_docs/contributing_to_the_p5js_reference.md#inserting-examples-within-the-description) for details.

Yashisinghal285:
Thanks for the detailed feedback @davepagurek — addressed all five points:

Switched to buildMaterialShader(...) instead of baseMaterialShader().modify(...)
Replaced inline anonymous callbacks with named top-level functions (e.g. myShaderBuilder)
Switched to finalColor.begin(); ...; finalColor.set(...); finalColor.end(); instead of getFinalColor(callback)
Converted examples to three-backtick fenced blocks so the p5.strands explanation sits right before its example, at the end of each function's docs
Verified examples actually run: I wasn't able to get inline examples to execute via the local custom:dev preview (this seems to be true for pre-existing, unmodified examples on the same pages too, not just my new ones), so I verified instead using the p5.js Web Editor with this PR's CDN build. blue() and red() both render correctly with no console errors, output matching their describe() text — screenshots attached. All 10 updated functions share the identical buildMaterialShader/finalColor pattern, so this confirms the approach works across the board.
Screenshot 2026-07-14 at 3 33 21 AM
Screenshot 2026-07-14 at 3 33 35 AM

davepagurek:
Hmm when I look at the diff on github now I only see unrelated changes that were pulled in from other PRs and not your docs changes any more. Is something up there?

> Verified examples actually run: I wasn't able to get inline examples to execute via the local custom:dev preview (this seems to be true for pre-existing, unmodified examples on the same pages too, not just my new ones), so I verified instead using the p5.js Web Editor with this PR's CDN build

Part of the reason for doing this is also to verify that we've done the update to inline examples correctly, so it's worth getting this working. What errors do you run into? If it's helpful, feel free to send a message on the p5.js discord in the #contributing-to-p5 channel and people can help you get things running there!

Yashisinghal285:
Thanks for catching that @davepagurek — you were right, something was off. My last commit with the buildMaterialShader/finalColor fixes (536e3c66c) was made locally but never actually got pushed to GitHub — it looks like it got left behind during a merge/rebase I did to sync with upstream. I've just pushed it now, so the diff should show the actual changes.
Apologies for the confusion, and thanks for flagging it rather than assuming it was intentional!

ksen0:
Hi @Yashisinghal285 before another round of review, could you try to test the documentation locally again please, and report any problems you run into, as Dave suggested? Please follow [this doc](https://p5js.org/contribute/contributing_to_the_p5js_reference/) and post in discrod.p5js.org #contribute-to-p5 where and how you run into problems! thanks  for all your work.

Yashisinghal285:
@davepagurek  @ksen0 
Posted in the **#contribute-to-p5** Discord channel as requested. I confirmed that the issue reproduces even with pre-existing, unmodified examples (not just the ones from my PR), so it appears to be related to the local preview rather than my documentation changes. I'll update this PR once I hear back or find the cause.

src/color/creating_reading.js
```diff
@@ -163,8 +149,9 @@ function creatingReading(p5, fn){
    *
    *   describe('A yellow square on a gray canvas.');
    * }
+   * ```
    *
-   * @example
+   * ```js
```
davepagurek: I think this needs to have the `example` tag too in order to render:
```suggestion
   * ```js example
```

Yashisinghal285:
Thanks for spotting that, @davepagurek  — you were exactly right. Added the example annotation to all 50 fenced code blocks (not just the strands ones, since all examples on this file were converted to fenced blocks). Rebuilt locally via custom:dev and confirmed both regular and p5.strands examples now render correctly, including the shader examples with buildMaterialShader/finalColor. Screenshot attached showing the blue() strands example rendering a blue sphere as expected.
Screenshot 2026-07-17 at 9 13 31 AM

davepagurek (approved):
Thanks for your patience on this, looks good to go now!

PR #8975, opened 2026-07-09
Rename HDR to P3

### LINKED ISSUE #8851: RGBHDR colorMode confuses gamut and dynamic range, and should be renamed
### Topic

The `RGBHDR` colorMode name and its documentation appear to conflate two separate concepts: **color gamut** and **dynamic range**.

As far as I can tell, `RGBHDR` currently offers access to a wider P3-derived color gamut, but I have not found a way for it to express or render luminance values above SDR white. In other words, it appears to provide **wide-gamut RGB**, but not **high-dynamic-range (HDR) RGB**.

The name `RGBHDR` therefore seems misleading. “HDR” customarily implies an extended _luminance_ range: for example, an image format or rendering pipeline that can represent values above ordinary SDR white, such as [Ultra HDR JPEG](https://en.wikipedia.org/wiki/Ultra_HDR) gain-map images or floating-point [OpenEXR](https://en.wikipedia.org/wiki/OpenEXR) images. These formats can encode "brighter-than-white" image content. By contrast, the [P3 color space](https://en.wikipedia.org/wiki/DCI-P3) expands the gamut of chromaticities available beyond sRGB, but does not by itself imply a larger luminance range.

Since `RGBHDR` appears to be [implemented as an alias to a P3 color space](https://github.com/processing/p5.js/blob/8563dd8cd59e0e3639ae20ee198cd221569ec02b/src/color/p5.Color.js#L1025), I think it would be wise to rename it before the name becomes too entrenched. A name such as `RGB_P3` would be clearer. 

Likewise, the [current documentation for `RGBHDR`](https://beta.p5js.org/reference/p5/colorMode/) says:

> `RGBHDR` - High Dynamic Range RGB defined within the Display P3 color space. Colors are expressed with an extended dynamic range. To render these colors accurately, you must use the HDR canvas.

I believe this documentation should be revised, since I have not found a way for the `RGBHDR` colorMode, by itself, to provide luminance values above SDR white. A more accurate description might be:

> `RGB_P3` - RGB values defined within the Display-P3 color space. This allows colors outside the sRGB gamut on compatible displays. 

In my testing, colors specified in `RGBHDR` were visibly able to access portions of the P3 gamut that lie outside sRGB, but were not able to produce luminance values above SDR white. By contrast, doing actual HDR output in p5 required making a separate WebGPU rendering pipeline with an `rgba16float` presentation buffer and [extended tone mapping](https://github.com/ccameron-chromium/webgpu-hdr/blob/245c838bd3069f0d264eac361dc8d33aac804a66/EXPLAINER.md). For anyone interested, I made an experimental proof-of-concept that used these techniques to demonstrate the distinction between extended dynamic range and extended color gamut. To be clear, the current `RGBHDR` implementation only allows for the graphics on the _left_ side of this sketch:

https://editor.p5js.org/golan/sketches/7n2mxXg09

<a href="https://editor.p5js.org/golan/sketches/7n2mxXg09">Image</a>

This sketch compares a 2x2 matrix of gamut and dynamic range: 

* SDR sRGB
* SDR Display-P3
* HDR sRGB
* HDR Display-P3

I was able to preview this using:

* macOS 15.6+
* Chrome 148+
* an HDR-capable MacBook Pro XDR display
* with the Settings display preset: "Apple XDR Display (P3-1600 nits)"

Thanks for your consideration.

limzykenneth:
Resolves #8851

Changes:
Per discussions in #8851, all instances of HDR in the code base is renamed to P3 for accuracy. Although this changes the behaviour, maintainers have agreed that the change is warranted for accuracy and correctness reason.

@ksen0 Do merge if it looks good to you.

### DIFF
--- a/docs/parameterData.json
+++ b/docs/parameterData.json
@@ -243,13 +243,6 @@
         []
       ]
     },
-    "abs": {
-      "overloads": [
-        [
-          "Number"
-        ]
-      ]
-    },
     "mult": {
       "overloads": [
         [
@@ -296,19 +289,19 @@
         ]
       ]
     },
-    "ceil": {
-      "overloads": [
-        [
-          "Number"
-        ]
-      ]
-    },
     "ellipseMode": {
       "overloads": [
         [
           "CENTER|RADIUS|CORNER|CORNERS"
         ],
         []
+      ]
+    },
+    "abs": {
+      "overloads": [
+        [
+          "Number"
+        ]
       ]
     },
     "nf": {
@@ -421,7 +414,7 @@
         [
           "Number?",
           "Number?",
-          "P2D|WEBGL|P2DHDR?",
+          "P2D|WEBGL|P2DP3?",
           "HTMLCanvasElement?"
         ],
         [
@@ -447,27 +440,11 @@
         ]
       ]
     },
-    "constrain": {
-      "overloads": [
-        [
-          "Number",
-          "Number",
-          "Number"
-        ]
-      ]
-    },
     "createImage": {
       "overloads": [
         [
           "Integer",
           "Integer"
-        ]
-      ]
-    },
-    "acos": {
-      "overloads": [
-        [
-          "Number"
         ]
       ]
     },
@@ -490,6 +467,13 @@
       "overloads": [
         [
           "Object?"
+        ]
+      ]
+    },
+    "ceil": {
+      "overloads": [
+        [
+          "Number"
         ]
       ]
     },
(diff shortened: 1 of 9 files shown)


davepagurek:
One thought, if we did want to avoid any breakages, we could possibly re-export `P2DHDR` as an alias of `P2DP3` (maybe marking as deprecated in the jsdoc). No strong opinion here though either way!

limzykenneth:
@davepagurek I was thinking about that, `P2DHDR` can be a getter that prints deprecation message but I'm not sure if it would be needed.

ksen0:
Thanks! I'll add it to the upcoming patch RC.

PR #8978, opened 2026-07-13
Add mat2 uniform support to WebGL setUniform, fixing silent no-op (#8…

### LINKED ISSUE #8939: [p5.js 2.0+ Bug Report]:WebGL silently ignores mat2 shader uniforms (no error, no warning)

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.3.0

### Web browser and version

Not browser-specific — the missing code path is in p5's source, so this reproduces in any WebGL-capable browser.

### Operating system

Not OS-specific (same reason).

### Steps to reproduce

1. Create a WebGL shader that declares a `uniform mat2`.
2. Set that uniform from sketch code with `setUniform(...)`.
3. Observe: **no error and no warning**, but the value never reaches the shader — the uniform stays at the GL default (zero matrix). The identical pattern with a `mat3` or `mat4` uniform works.

### Minimal illustration

> The definitive evidence is the source analysis below — this snippet just shows the symptom.

```js
let s;

function setup() {
  s = createFilterShader(`
    precision highp float;
    uniform mat2 uM;
    void main() {
      // diagonal of uM drives the red & green channels
      gl_FragColor = vec4(uM[0][0], uM[1][1], 0.0, 1.0);
    }
  `);
}

function draw() {
  s.setUniform('uM', [1.0, 0.0, 0.0, 1.0]); // identity
  filter(s);
  // Expected: yellow (uM uploaded as identity -> 1,1,0)
  // Actual:   black  (uM never uploaded -> stays at the 0 matrix -> 0,0,0)
}
```

### Root cause

The root cause is in `src/webgl/utils.js`. The `setUniform` switch handles `FLOAT_MAT3` (line 226, via `gl.uniformMatrix3fv`) and `FLOAT_MAT4` (line 229, via `gl.uniformMatrix4fv`), but there is **no `case gl.FLOAT_MAT2` and no `default:` branch** — the switch ends with `//@todo complete all types` at line 327. A `mat2` uniform falls straight through and is never uploaded.

A repo-wide check confirms this is structural, not a one-off miss: `gl.uniformMatrix2fv` — the WebGL call required to upload a 2×2 matrix — appears **zero times anywhere in `src/`**, while `uniformMatrix3fv` and `uniformMatrix4fv` each appear once. So there is no code path on which a `mat2` uniform could reach the GPU.

`getWebGLUniformMetadata` likewise omits `FLOAT_MAT2` from its `isArray` classification (around lines 360–369), consistent with `mat2` never having been wired up.

### WebGL vs WebGPU

For contrast (as motivation, not the core claim): the **WebGPU** backend handles square `mat2`. In `src/webgpu/p5.RendererWebGPU.js` it parses the matrix dimension generically (~line 2174) and maps `mat2x2<f32>` (line 3103), and throws a clear `Non-square matrices not implemented yet` for non-square (line 2172).

So `mat2` works in WebGPU but silently does nothing in WebGL — and where WebGPU at least fails loudly on what it doesn't support, WebGL fails silently. (Non-square matrices are out of scope here; WebGPU's explicit throw shows that's separately tracked.)

### Possible approaches (would like maintainer direction before implementing)

1. Add the missing `FLOAT_MAT2` case (`gl.uniformMatrix2fv`) to the switch and to the `isArray` classification — completing the `//@todo`.
2. Or, if full `mat2` support isn't wanted yet, emit a friendly warning for currently-unsupported uniform types instead of silently no-op'ing, in keeping with p5's friendly-error approach.

Happy to fix this myself once an approach is agreed on @ksen0 @davepagurek !

harshiltewari2004:
Resolves #8939

### Changes
- Added the missing `FLOAT_MAT2` case to the `setUniform` switch in
  `src/webgl/utils.js` (`gl.uniformMatrix2fv`), mirroring the existing
  mat3/mat4 cases — completing part of the `//@todo`
- Added `FLOAT_MAT2` to the `isArray` classification in
  `getWebGLUniformMetadata`, so mat2 uniforms take the array path in
  p5.Shader's value caching
- Added a visual regression test (`createFilterShader` with a `uniform mat2`
  driving the fragment color from the matrix diagonal). Verified the test
  fails on unfixed code and passes with the fix.

Note: there was previously no direct setUniform test coverage for matrix
uniforms of any size, in either backend — this test establishes the pattern.
mat2 remains untested on the WebGPU side (where it already works); happy to
add a WGSL counterpart in a follow-up if wanted.

While verifying with a full test run, I hit an intermittent failure in
`Typography > textWeight > can control non-variable fonts` — unrelated to
this change (2D-mode test, no path to the modified code; passes repeatedly
in isolation on this branch). It loads its font from Google Fonts at test
time, which is the likely flake source.

### PR Checklist
- [x] `npm run lint` passes
- [ ] [Inline reference] is included / updated — N/A, no documented API change


davepagurek (approved):
Thanks, this is good to go I think! Are you also interested in looking into p5.strands support for matrices? Currently we don't support them, so we resort to hacks like this: https://github.com/processing/p5.js/blob/5743cb5ba5aef0e1cbe01259dd1d4c17910164c4/src/core/p5.Renderer3D.js#L1929-L1943

If that's of interest, it'll help simplify the above snippet and help tie into @perminder-17's work on a transformation API in p5.strands, so we'd appreciate the help! A start would be to make an issue for it.

davepagurek:
@all-contributors please add @harshiltewari2004 for code

harshiltewari2004:
Thanks @davepagurek! Yes, I'd be glad to look into strands matrix support — that hack is a great motivating example. I'll open an issue to lay out the problem and a proposed approach so we can settle scope before any code. Since this ties into @perminder-17's transformation API work, I'll tag him there too so we can make sure the direction lines up rather than overlapping.

PR #8981, opened 2026-07-16
Add unit tests for isLooping() default, noLoop() and loop() states

### LINKED ISSUE #8979: [p5.js 2.0+ Bug Report]: Missing unit tests for isLooping() return value

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [x] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

current main (5743cb5ba)

### Web browser and version

N/A — test coverage gap, not a runtime bug

### Operating system

N/A — same reason

### Steps to reproduce this

`isLooping()` (src/core/structure.js) returns the internal `_loop` flag and is
documented in its JSDoc as supported usage — but its return value is never
tested. There are currently zero references to `isLooping` anywhere under
`test/`; the existing `loop()`/`noLoop()` suite in
`test/unit/core/structure.js` asserts only on `frameCount` behavior and never
calls `isLooping()`.

Proposed tests (test-only, no source changes):
- `isLooping()` returns `true` by default while the draw loop is running
- returns `false` after `noLoop()`
- returns `true` again after a subsequent `loop()`

I'd like to fix this myself if approved.

harshiltewari2004:
Resolves #8979

### Changes

- Added a `p5.prototype.isLooping` suite to `test/unit/core/structure.js` covering the three states scoped in the issue: `true` by default, `false` after `noLoop()`, `true` again after `loop()`.
- The suite creates its own p5 instance instead of using the file's shared one. Toggling `noLoop()`/`loop()` within the same frame on the shared instance double-pumps its draw loop — `noLoop()` doesn't cancel the pending `requestAnimationFrame` tick, and `loop()` immediately starts a new chain, so the stale tick resurrects alongside it. This made the existing timing-based "noLoop should stop" test fail (reproduced consistently; passes consistently without the new suite). A private instance isolates the toggling completely.
- The tests are synchronous, unlike the neighboring `loop`/`noLoop` tests: The tests are synchronous, unlike the neighboring loop/noLoop tests: those verify the draw loop's actual behavior, which needs real time to elapse so pending requestAnimationFrame ticks can fire (or fail to). These tests only assert the state flag, which noLoop()/loop() set synchronously before returning, so there's nothing to wait for.

### PR Checklist

- [x] `npm run lint` passes
- [ ] [Inline reference] is included / updated — N/A, test-only change


lirenjie95 (approved):
LGTM once the placeholder in the PR description is cleaned up.

harshiltewari2004:
Cleaned up-thanks for the catch!

lirenjie95 (approved):

PR #8984, opened 2026-07-16
Transformation Matrix for p5.Strands

### LINKED ISSUE #8953: Support transformation matrices in p5.strands

perminder-17:
Resolves #8953 

 Changes:

 Screenshots of the change:

#### PR Checklist

- [ ] `npm run lint` passes


src/strands/strands_api.js
```diff
@@ -837,6 +837,186 @@ const _rgb2hsl = (instance, colorNode) => {
...
+    const [t, axis, angle] = args;
+    if (!t?.isStrandsNode || t.typeInfo().baseType !== BaseType.MAT || t.dimension !== 4) {
+      FES.userError('type error',
+        'rotateAxisAngle() only works on a 3D transform created with transform3D().');
+    }
+    const n = this.normalize(axis);
+    const x = n.x, y = n.y, z = n.z;
+    const c = this.cos(angle);
+    const s = this.sin(angle);
+    const omc = p5.strandsNode(1).sub(c); // 1 - cos
```
perminder-17: Hi @davepagurek ,

Just thinking out loud. In our non-strands modules, each area of functionality lives in its own file, for example, `transform.js` keeps the JSDoc, examples, and implementation together in one place.

I was thinking of following that same convention for the strands transforms: moving the matrix and transform helpers out of `strands_api.js` into a dedicated `strands_transform.js` (something better name for the file), with each function's docs kept alongside its implementation.

One thing to check: right now the strands functions are documented centrally in `p5.strands.js` rather than inline. Would you prefer keeping that, or colocating the transform docs with the code in the new file?

ksen0:
Just starte looking at this one, really exciting! I will work on some more complex sketches to test it out, but I appreciate the docs you've included out of the box. 

Would it be a good idea to add some bench and/or unit and/or visual tests here? I think visual and unit at minimum, given the scope of the PR, would be good...

perminder-17:
> Just starte looking at this one, really exciting! I will work on some more complex sketches to test it out, but I appreciate the docs you've included out of the box.

Ah, correct. I'm also getting feedback and testing it with more complex examples myself. For now, I'm trying to find ways to break the feature and catch any issues. 

Once I'm confident with the code, I'll add tests for them.

src/strands/strands_transform.js
```diff
@@ -0,0 +1,919 @@
...
+   * let myShader;
+   *
+   * function setup() {
+   *   createCanvas(200, 200, WEBGL);
+   *   myShader = buildMaterialShader(slide);
+   * }
+   *
+   * function slide() {
+   *   getWorldInputs((inputs) => {
+   *     // translate() returns a new transform, so assign the result back.
```
davepagurek, line 348: Maybe add an inline URL here? Or add a note at the end of the `transform()` docs to mention strands + link to the `transformPoint` docs as the spot with more in depth explanation

davepagurek: Or interleave some writing before the strands example saying the same thing

src/strands/strands_transform.js
```diff
@@ -0,0 +1,919 @@
...
+   * @example
+   * let myShader;
+   *
+   * function setup() {
+   *   createCanvas(200, 200, WEBGL);
+   *   myShader = buildMaterialShader(wobble);
+   * }
+   *
+   * function wobble() {
+   *   getWorldInputs((inputs) => {
```
davepagurek, line 147: Can we use `worldInputs.begin()` and `worldInputs.end()` in the examples for consistency?

davepagurek:
Some feedback:
- We could make visual tests out of examples
- We should make sure that in the website reference, we don't accidentally recategorize the transform methods into the strands category. Maybe we need to add e.g. `@module` tags onto individual methods?
- We also want the strands parts to come last, it may be that adding our docs to the existing docs rather than making new ones and relying on the docs builder to stitch them together is the most reliable way of doing that
- Once main is merged in, could we add examples using `instances(n)` and using transforms per instance?
- Not blocking for this PR but thoughts for the future: for sketches like https://openprocessing.org/@davepagurek/2985673 when transforms vary based on position, the current `transformNormal` logic is not quite correct. That sketch has some code to figure out what the new normal should be by sampling a few nearby points, but it relies on a slightly different structure (your transform is a function based on an input position, not just a single matrix.) Any thoughts on what it could look like in the future if we wanted to support that? Just to confirm that it wouldn't clash with our current API?
- Just noticed that the current p5 transforms for skewing are called `shearX` and `shearY` but it looks like the ones we've added here use `skew` as a prefix instead, based on the DOMMatrix API. I hadn't realized we use different conventions there. Do you or @ksen0 have thoughts on whether it's better to just use what p5 already has for naming there?

davepagurek:
oh and one more thing that I forgot: add on to the logging system for experimental features in https://github.com/processing/p5.js/pull/9109. It sounds like we don't want a whole new contributor doc, but we just want to add a SECTION to the strands doc, and then link to that section via a `#sectionname` suffix (I think we do that in tables of contents in other contributor docs.) We may need to update the system a little to make a category that links to a section of a doc rather than to a whole doc of its own, + we may need to use the manual checker rather than the decorator to catch uses of these functions since we don't want it logging in regular non-strands use

PR #8994, opened 2026-07-20
Vector lerp() function now takes in generalized dimensions

### LINKED ISSUE #8927: [p5.js 2.0+ Bug Report]: Vector.lerp() is hard coded to use three dimensions

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.3.0

### Web browser and version

Brave 1.90.122

### Operating system

Linux Mint 22.1

### Steps:
1. Create two 4D vectors
2. Use lerp() to interpolate between the two
3. Note the result only works for the first three dimensions

### Snippet:

```js
function setup() {
  let v1 = createVector(0, 1, 0, 1);
  let v2 = createVector(1, 0, 1, 0);
  v1.lerp(v2, 0.5);
  
  // v1 should be vector[0.5, 0.5, 0.5, 0.5], but is actually vector[0.5, 0.5, 0.5, 1]
  print(v1.toString());
}
```

### Discussion

`lerp()` currently computes x, y, and z individually. To extend to vectors of any dimension, other vector methods can be used, like `p5.Vector.add(v1, p5.Vector.sub(v2, v1).mult(c))`. Of course, this needs to be adapted to both the instance and static methods.

mawerb:
Resolves #8927

 Changes:
-  Refactored `lerp()` to take in generalized dimensions allowing for function to be called on vectors with more than 3 dimensions and mismatched.
-  Refactored `_validatedVectorOperation` decorator to be able to take in trailingArgs in order to fit the case of `lerp()` where it follows the pattern of validating parameters its taking with ...args like the `mult()`, `add()`, `sub()` functions but requiring the extra argument `amt`
-  Added test for prioritizing smaller dimensions for lerp()
-  Refactored tests dependent on passing of 4 arguments to fit the new structure of taking in a vector, array, or number list
-  Added tests mentioned in issue #8927 and extra to ensure` lerp()` works as intended on  mismatched dimensions and vectors with dimensions > 3

 Screenshots of the change:
Screenshot 2026-07-20 at 3 57 17 PM

the 1 failed test was present before changes were made (it's the typography test)

#### PR Checklist
- [x] `npm run lint` passes


src/math/patch-vector.js
```diff
@@ -25,17 +25,99 @@ export function _defaultEmptyVector(target){
...
+            // No arguments? No action (same as other vector ops)
+            return this;
+          }
+          if (!this.friendlyErrorsDisabled()) {
+            this._friendlyError(
+              'Requires valid arguments',
+              'p5.Vector'
+            );
+          }
+          return this;
```
perminder-17: I think this code is repeated six times in this function. Can we create a helper function that takes the message as an argument? Then we can replace these repeated lines with a single helper call. That way we can reduce the code?

perminder-17 (commented):
Looks good!

src/math/patch-vector.js
mawerb: Nice catch, made the change! Thanks for reviewing my pr :)

src/math/p5.Vector.js
```diff
@@ -2350,6 +2360,32 @@ class Vector {
...
+   * // META:norender
+   * function setup() {
+   *   // Create a p5.Vector object.
+   *   let v = createVector(1, 1, 1);
+   *
+   *   // Interpolate with an array.
+   *   v.lerp([3, 3, 3], 0.5);
+   *
+   *   // Prints "p5.Vector Object : [2, 2, 2]" to the console.
+   *   print(v.toString());
```
perminder-17: One thought: if someone reads these examples, they have to open the console to see the output. Since multiple examples use the console, the messages can get mixed up. What do you think about using the text() function instead, so the output is shown directly on the canvas?

perminder-17 (commented):
Everything else looks good so far!

src/math/p5.Vector.js
mawerb: thanks, agreed, switched the lerp examples to show the result on the canvas with text() so readers don’t need the console.

ksen0:
This is looking really good, thanks both!

mawerb:
Hey! I was just curious regarding the next steps for getting my pr approved. This is my first time contributing to p5.js so I’m unsure how it works

mawerb:
@perminder-17 @ksen0 bump. How do i move forward with getting this PR approved?

davepagurek:
Hi @mawerb! Before merging we make sure that the branch is up-to-date with main, but when I did so, I think maybe a test was added elsewhere (or possibly if it was removed before git was confused about the update and it's back?) and is now failing:
```
 FAIL   unit-tests (chromium)  test/unit/math/p5.Vector.js > p5.Vector > lerp > with p5.Vector > should call lerp with 4 arguments
AssertionError: expected "Mock" to be called with arguments: [ 1, 2, 3, 1 ]

Received:

  1st Mock call:

  [
+   Vector {
+     "isVector": true,
+     "values": [
        1,
        2,
        3,
+     ],
+   },
    1,
  ]


Number of calls: 1

 ❯ test/unit/math/p5.Vector.js:1525:24
    1523|         vi.spyOn(v, 'lerp');
    1524|         v.lerp(new Vector(1, 2, 3), 1);
    1525|         expect(v.lerp).toHaveBeenCalledWith(1, 2, 3, 1);
       |                        ^
    1526|       });
    1527|     });

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

```

Do you think you could take a look at that test and help resolve the merge?

Other than that, @perminder-17 anything else need updating in the code before merging?

perminder-17 (approved):
Really sorry for the delay, everything looks good to me! Approved!

Just the same concern on the test failure,

mawerb:
Thanks for catching that @davepagurek ! That test came back when merging with main. It was checking the old 3D behavior where v.lerp(vector, amt) got rewritten to four numeric arguments (x, y, z, amt). With the N-D update that rewrite is gone, since lerp now takes (values, amt) instead, so the assertion was outdated.

I removed that test. The vector form is already covered by the behavior test that checks the resulting components. The suite is passing again on my end.

ksen0:
Hi both! sorry for the delay. The test job does not succeed (and its not one of the flaky tests): https://github.com/processing/p5.js/actions/runs/34200989101/job/102470041771?pr=8994

mawerb:
Hi @ksen0  found the cause of the error and it's me specifying that the vector to lerp can be passed in as separate values. Error root:

`@param {...Number} values  target vector components
   * @param  {Number}   amt`
   
The current solution changes the parameter types to take in a single {...Number} rest parameter instead and specifies that the last argument passed in for this parameter will always be the "amt" to interpolate by. I'm unsure if this is the best solution to this since iTypeScript can no longer tell which argument is amt, so something like v.lerp(1) type-checks even though the decorator rejects it at runtime (it treats 1 as amt and then has no target).

The other option would be named optional components, like x, [y], [z], [w], amt. That types amt correctly for the common cases, but the numeric-list form would only be typed up to 4D. Runtime would still accept higher dimensions via a vector or array.

Happy to switch to a different approach if you have a preferred pattern for this.

ksen0:
thanks @mawerb @perminder-17 ! I think the most recent update makes sense.

PR #8996, opened 2026-07-20
Expect private tags

### LINKED ISSUE #8980: [p5.js 2.0+ Bug Report]: Reference docs for RC include 2 private methods

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.3.1-rc.2

### Web browser and version

_No response_

### Operating system

_No response_

### Steps:

In `p5.js-website`:

1. `npm run build:reference` whih clones p5.js
2. `npm run dev`
3. Observe, some internal methods are visible, which should not be. Both are marked `@private` and `@internal` but this is apparently ignored? It does work for many other methods. Possibly build reference script should be updated?

Image

ksen0:
Fixes https://github.com/processing/p5.js/issues/8980

Local test ` npm run custom:dev ../p5-20/p5.js#priv` 

image

### DIFF
--- a/utils/convert.mjs
+++ b/utils/convert.mjs
@@ -8,22 +8,29 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
 const data = JSON.parse(fs.readFileSync(path.join(__dirname, '../docs/data.json')));
+const hasTag = (entry, tagName) =>
+  Array.isArray(entry?.tags) &&
+  entry.tags.some(t => String(t?.title ?? '').toLowerCase() === tagName);
 
 // Generate documentation used in the p5.js reference. This data will get read in
 // the p5.js-website repo: https://github.com/processing/p5.js-website/
 const htmlStrategy = {
-  shouldSkipEntry: (entry) =>
-    // Skip static methods on p5.Vector for now because the names clash with
-    // the non static versions
-    entry.scope === 'static' &&
-    entry.memberof === 'Vector' &&
-    ['add', 'sub', 'mult', 'div', 'copy', 'rem', 'rotate', 'dot', 'cross', 'dist', 'lerp', 'slerp',
-    'mag', 'magSq', 'normalize', 'limit', 'setMag', 'heading', 'angleBetween', 'reflect',
-    'array', 'equals'].includes(entry.name),
-  
-  processDescription: (desc) => descriptionString(desc),
-  
-  processType: (type) => typeObject(type)
+  shouldSkipEntry: entry =>
+    (
+      // Skip static methods on p5.Vector for now because the names clash with
+      // the non static versions
+      entry.scope === 'static' &&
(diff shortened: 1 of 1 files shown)


ksen0:
@limzykenneth stuck here. The original commit here fixed the reference docs but broke the tests.

So I am trying to add an "hidden" metadata that parsers/reference.ts could use to exclude but now I can't reproduce the original problem and also the tests still fail.

To be honest my mental model here makes no sense. I expect the p5.js data.json to match what p5.js-website builds and when I was getting the errors, it didn't match (on `main` on both, cleaned up). Now that's not happening either; I am not ruling out some kind of leftover data somewhere but I was manually deleting any left over reference artifacts.

I can sort of understand why changing exclusion rules in convert would break tests but I definitely dont understand why adding some metadata would.

Stopping here for the day, if you have any ideas would be helpful

perminder-17:
Hi @ksen0 , regarding the test concern, I think the failing test was this one: https://github.com/processing/p5.js/blob/a350bb39e1463e8e68c86e414b0e00663ec84545/test/unit/webgl/p5.Framebuffer.js#L672-L686

I reran the tests twice, and it passes now. I also remember discussing this particular test with Ken on Discord in the group. I'll take a look at fixing it, but for now I don't think it's related to your changes.

I observed that a test is flaky and occasionally fail (6 out of 10 times on an average) when executed in GitHub Actions.

limzykenneth:
@ksen0 I had a look, so from what I can see, everything is actually working as intended currently. The reason for the original issue showing the private `makeGuardedCallbacks` entry in the website build is because the website build automatically fetches the known version the repo is tagged with in [`src/globals/p5-version.ts`](https://github.com/processing/p5.js-website/blob/main/src/globals/p5-version.ts) and that currently points to 2.3.0. Checking out 2.3.0 in the p5.js repo where [`makeGuardedCallbacks` is called](https://github.com/processing/p5.js/blob/v2.3.0/src/strands/strands_transpiler.js#L1663-L1692), it is not marked as private, ie. before your commit marking it as private. Which means if we build the website against 2.3.0 which by default it does, it will continue to show the functions even though it has been marked private in the latest commit or RC release.

The website can be made to build against an RC version by just replacing the 2.3.0 version number in `src/globals/p5-version.ts` to `2.3.0-rc.2`.

In other words, I think there's nothing to fix here, we can try to make building against specific version of the p5.js repo by the website easier.

ksen0:
Thanks for explaining @limzykenneth . I did `npm run build:reference` with `export const p5Version = "2.3.1-rc.2" as const;` and indeed the private methods are not there.

PR #8997, opened 2026-07-21
Created a loading indicator when the canvas is being set up

### LINKED ISSUE #8922: [p5.js 2.0+ Bug Report]: Loading indication

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.0+

### Web browser and version

any

### Operating system

any

### Steps:
1. Run the below code
2. Observe that while loading, the canvas is blank with no feedback to user at all (this is a regression compared to p5.js version 1)

### Snippet:

```js
// for p5.js versions 2.0 or later only
async function setup() {	
  createCanvas(400, 400);

  // Delay to simulate loading large files (for demo only, not actual usage)
  await fakeLoad(7000);
 
  // Your usual setup things below
  background("#EB5580");
  fill(255);
  circle(width / 2, height / 2, 100);
}

function draw() {
  circle(mouseX, mouseY, 20);
}

async function fakeLoad(delay) {
  await new Promise(resolve => setTimeout(resolve, delay));
}
```

Jextic:
Resolves #8922 

 Changes:

- Created loading.js file in p5.js/src/core. loading.js handles the logic for creating a loading indicator when the canvas is being set up. When `showLoadingIndicator()` is called, a div element with a "Loading..." text will be created. When `hideLoadingIndicator()` is called, the div element will be removed from the DOM.
- Modified main.js so that `showLoadingIndicator()` and `hideLoadingIndicator()` are called before and after `await context.setup()`. The loading indicators will only appear while the canvas is loading the `setup()` function in sketch.js.
- Created a loading.js unit test file in p5.js/test/unit/core. The test is was created based on the sample code from issue #8922 to see if the loading indicator works as intended.

 Screenshots of the change:

image

image

All tests, including loading.js, pass. The only exception is a typography which always failed, even before changes were made to the project.

#### PR Checklist

- [x] `npm run lint` passes

Also worked with @Joshin on this issue.


ksen0:
Hi @Jextic thanks for putting this together and sorry for the long time to get to your comment, but please check the comment on the issue. If possible, to keep (1) non text based and (2) decoupled. I will be able to review this more deeply next week as well. Thanks for your work on this!

Jextic:
Hi @ksen0,
I decoupled loading.js by registering it as a p5 addon. Now, the loading indicator will be called and removed using `lifecycles.presetup` and `lifecycles.postsetup`. I also replaced the "loading..." text with a basic spinning circle animation. The test unit for loading.js was also updated accordingly.


https://github.com/user-attachments/assets/7535f4ab-a378-4d47-be77-d49d8f1230c5

ksen0:
Nice, thanks! Looks good to me.

ksen0:
> All tests, including loading.js, pass. The only exception is a typography which always failed, even before changes were made to the project.

Yes that one is known to be flaky, on CI it will pass when rerunning.

ksen0 (approved):

src/core/loading.js
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @private
+ * @param {HTMLElement} container The HTML element to append the indicator to
+ * @returns {HTMLElement} The loading indicator div element
+ */
+function createLoadingIndicator(container) {
+  if (!document.getElementById('loading-style')) {
+    const loadingStyle = document.createElement('style');
+    loadingStyle.id = 'loading-style';
+    loadingStyle.textContent = '@keyframes loading-spin { to { transform: rotate(360deg); } }';
```
limzykenneth: Would like this animation name to be namespaced a bit more, something like `p5-loading-spin`.

src/core/loading.js
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @param {*} p5 The p5 constructor
+ * @param {*} fn The p5 prototype object
+ * @param {*} lifecycles Lifecycle hooks for the sketch
+ */
+export default function loading(p5, fn, lifecycles) {
+  lifecycles.presetup = function() {
+    if (typeof window === 'undefined' || loadingIndicator) return;
+
+    const container = this.canvas?.parentElement || document.body;
```
limzykenneth: I've not tested this out yet but how does this look on instance mode sketches with contents already on the page and the sketch attached via `canvas.parent()`?

Also what about multiple sketches in instance mode?

Jextic: Thanks for taking the time to review. I have tested it out and if there is already existing contents on the page, they won't be affected. A loading indicator will appear during `presetup` and will only remove itself during `postsetup`. I also updated `presetup` so that the loading indicator is now independent for each sketch so it is compatible if there are multiple sketches in instance mode. 

I realize that `this.canvas?.parentElement` might never be called on since the canvas is created during `setup()` so I removed it. If the sketch is created using `new p5(sketch);`, a loading indicator will appear at the center of webpage. If the sketch is created with a \<div\> element as the second parameter (Ex. `new p5(sketch, container);`), a loading indicator will appear at the center of the \<div\> container instead. I looked at the p5.js web editor and thought this would be useful in case you wanted to only embed the loading indictor onto the preview. Is there another approach to the loading indicators that you have in mind?

src/core/loading.js
Jextic: Thank you. I've changed the name from `loading-spin` to `p5-loading-spin` in my most recent commits.

ksen0:
Hello! So after some discussion with @limzykenneth and @davepagurek, I'll go ahead and merge this but keep the issue open. I will also release a patch with this loading indication, so that it's available sooner. However I'll keep the parent issue open and add a comment there describing a followup to address the instance mode comments, which includes using a canvas-based animation rather than css.

PR #8998, opened 2026-07-21
Fix static p5 vector method

### LINKED ISSUE #8930: [p5.js 2.0+ Bug Report]: Some Vector friendly errors don't work correctly

Ayush4958:
Resolves #8930 

Changes:
- In `src/math/p5.Vector.js` :- Properly `bound _friendlyError` & `friendlyErrorsDisabled` to the static Vector. Added early `false` after the friendly error is triggered. This prevents a secondary crash.

- Added a new unit test suite to ensure the static FES validation works.

Screenshots of the change:
Screenshot 2026-07-21 071243

#### PR Checklist
- [x] `npm run lint` passes


Ayush4958:
@ksen0 
PR is ready for review 

thnks ...

src/math/p5.Vector.js
```diff
@@ -3660,6 +3661,11 @@ function vector(p5, fn) {
     return p5.disableFriendlyErrors;
   };
 
+  Vector._friendlyError = p5._friendlyError;
+  Vector.friendlyErrorsDisabled = function() {
+    return p5.disableFriendlyErrors;
+  };
```
perminder-17, line 3671: Here we are defining the fes in the vector class and here p5._firendlyError is a global static. 

What if we don't do this and everywhere in this file wherever `this._friendlyError` is called we replace it with `p5._friendlyError` it should work the same and could look a little cleaner? 

It's similar to how `setHeading()` is in this file (line no. 2028)? 

what you think?

Ayush4958: yeah, It works perfectly fine in codebase, cleaniness could be maintained and we dont have to create its duplicate method.

I could apply those changes, share ur thoughts on that

ksen0: please do @Ayush4958 thanks!

Ayush4958: @ksen0 @perminder-17 
I was trying to implement planned changes in code, but it's throwing rerference p5 is not defined.
because  vector sits above that wrapper, any code inside the class methods that tries to say p5._friendlyError will crash the entire library because p5 doesn't exist in that context.

we can go with the implementation I had proposed in PR

perminder-17 (commented):
Yes, I now understand where this error is coming from. It didn't cause any problem in the browser, but as soon as I ran the tests, the error showed up.

So about the `setHeading` reference I shared earlier, that one is an instance method, so there too we should replace `p5._friendlyError` with `this._friendlyError`. That's the correct fix.

One more thing: I noticed the Matrix functions have the same issue.
https://github.com/processing/p5.js/blob/76e4694166be2c3c515cfd5c3da3d35b7dc1e372/src/math/Matrices/Matrix.js#L249
 There as well, it's best to go through this instead of p5, because p5 isn't defined at the module level, which makes the bare `p5._friendlyError` unreliable. 

You need to route the actual p5 friendly error here too, add `Matrix.prototype._friendlyError = p5._friendlyError;` inside `matrix(p5, fn)`, so `this._friendlyError` in the Matrix methods works.


Also, very minor thing I noticed while looking at `setHeading` function, 
https://github.com/processing/p5.js/blob/76e4694166be2c3c515cfd5c3da3d35b7dc1e372/src/math/p5.Vector.js#L2026
in vector calss we don't have anything called `_values`, in constructor we define `values` so, can you please replace this._values -> this.values wherever we had? 

Thanks for your input on fixing them.

Ayush4958:
@perminder-17 , I had cover all the fixing u mentioned, u can have look to it 
@ksen0 , PR is ready for review 

thanks ....

perminder-17 (approved):
Looks good to me!

PR #8999, opened 2026-07-21
fix: Vector.equals() asymmetric result when dimensions differ

### LINKED ISSUE #8929: [p5.js 2.0+ Bug Report]: Vector.equals() is quirky when dimensions are not equal

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.3.0

### Web browser and version

Brave 1.90.122

### Operating system

Linux Mint 22.1

### Snippet:

```js
function setup() {
  let v1 = createVector(1, 2, 3);
  let v2 = createVector(1, 2, 3, 4);
  
  // Equals is symmetric so these should print the same values, but they don't
  print(v1.equals(v2));
  print(v2.equals(v1));
}
```

### Discussion

To be consistent with other Vector methods, `equals()` should log a warning if the two vectors have different lengths, then compare the components up to the smallest length.

BHARATH0153:
Fixes #8929
 
### changes 

   * components of another vector. Any missing parameters are assigned the value 0.
 * components of another vector.
  * If the two vectors have different lengths, a warning is logged and only
  * the components up to the shorter length are compared. 

- [x] `npm run lint` passes
- [x] [Inline reference] is included / updated (N/A — no public API added)


BHARATH0153:
@perminder-17 please review thanks!

src/math/p5.Vector.js
```diff
@@ -2826,8 +2828,15 @@ class Vector {
       values = args;
     }
 
-    for (let i = 0; i < this.values.length; i++) {
-      if (this.values[i] !== (values[i] || 0)) {
+    const minDimension = Math.min(this.values.length, values.length);
+    if (this.values.length !== values.length) {
+      console.warn(
+          minDimension +
```
perminder-17: The code has build failure. 

After `minDimesnion + there isn't any message, which causes the build to fail when running `npm run build`. It would be good to add one.

perminder-17 (changes requested):
Looks good, one minor thing to take care of!

BHARATH0153:
@perminder-17 done please check thanks

perminder-17 (approved):
Looks good to me!

BHARATH0153:
@perminder-17 it pre existing failure unrelated to changes

src/math/p5.Vector.js
```diff
@@ -2826,8 +2828,17 @@ class Vector {
       values = args;
     }
 
-    for (let i = 0; i < this.values.length; i++) {
-      if (this.values[i] !== (values[i] || 0)) {
+    const minDimension = Math.min(this.values.length, values.length);
+    if (this.values.length !== values.length) {
+      console.warn(
```
ksen0: ~Is there a reason this is not using FES?~ (This is correct, please ignore, sorry)

ksen0 (changes requested):
There is some duplication here of the text and the logic of the existing helper https://github.com/BHARATH0153/p5.js/blob/15167faafad5345a872a79398548cb7d60b4c206/src/math/p5.Vector.js#L12 this helper should be used, rather than duplicating behavior, unless it is not possible for a specific technical reason. Thanks for your work on this!

perminder-17:
> There is some duplication here of the text and the logic of the existing helper https://github.com/BHARATH0153/p5.js/blob/15167faafad5345a872a79398548cb7d60b4c206/src/math/p5.Vector.js#L12 this helper should be used, rather than duplicating behavior, unless it is not possible for a specific technical reason. Thanks for your work on this!

Yeah, nice catch! Also, there's a small typo in the warning message -- "linger" should be "longer". Could you please take care of that as well?

BHARATH0153:
@ksen0 @perminder-17  done please check

perminder-17 (approved):
Looks good!

BHARATH0153:
@ksen0  @perminder-17 please merge this

PR #9009, opened 2026-07-23
Fix/issue 8924

### LINKED ISSUE #8924: [p5.js 2.0+ Bug Report]: p5.Vector.x set doesn't work on 1D vector

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [x] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.3.0

### Web browser and version

Brave 1.90.122

### Operating system

Linux Mint 22.1

### Steps:
1. Create a 1D vector
2. Set v.x to change the "x" value
3. Show that setting v.x didn't work

### Snippet:

```js
function setup() {
  let v = createVector(1);
  v.x = 2;
  
  // This should print "vector[2]", but prints "vector[1]"
  print(v.toString());
}
```

### Problem
The code contains a comparison `if (this.values.length > 1)`; it doesn't work because the length is 1. The comparison should be `if (this.values.length > 0)`.

solowiz398:
Resolves #8924

## PROBLEM DESCRIPTION
The setter method of 'x' for a 1 dimension vector does not reflect any changes to its value.

Vector dimensions are stored in an array, the set method of x checks to see if the length of the array is greater than 1. The condition passes for a 2d vector but fails for a 1d vector

## SOLUTION
Updated the setter method for 'x' to check to see if the length of the array is greater than 0. This condition passes for both a 1d vector and n amount of vector

### 1. Verify the fix on this branch
 ` npm test -- test/unit/math/p5.Vector.js`

Ensure the following unit test passes:

'should show a value change for x with toString() for 1d Vector'

 Changes:

- Fixed setter method for x in p5.Vector.js to allow setting values for 1d vector `length > 0`
- Added a unit test to verify value change for 'x'

 Screenshots of the change:
Non visual code and unit test 

#### PR Checklist
- [ ] `npm run lint` passes


test/unit/math/p5.Vector.js
```diff
@@ -2184,6 +2184,13 @@ suite('p5.Vector', function () {
       v = new Vector(1, 2, 3, 4);
       expect(v.toString()).toBe('vector[1, 2, 3, 4]');
     });
+
+    test('should show a value change for x with toString() for 1d Vector', function(){
+      v = new Vector(1);
+      v.x = 2;
+      let output = v.toString();
+      assert.equal(output, 'vector[2]');
```
ksen0: Thanks! Please update to match the style of other tests using expect to be (and removing the assert dep, I don't think it would be needed?)

ksen0 (commented):
One minor thing, otherwise thanks, happy to include this in the next patch!

test/unit/math/p5.Vector.js
solowiz398: Thanks @ksen0 I just made the changes. Let me know if there's anything else

ksen0 (approved):

PR #9014, opened 2026-07-24
Fixes #9000

### LINKED ISSUE #9000: [p5.js 2.0+ Bug Report]: Large slowdown in `p5.Graphics.set()` compared to v1.11.x

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [ ] Events
- [x] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.3.0

### Web browser and version

- Chrome Version 151.0.7922.34 (Official Build) (arm64) 
- Safari Version 26.5.2 (21624.2.5.11.8)
- Firefox 152.0.4 (aarch64)

### Operating system

macOS 26.5.2 (25F84) Tahoe

### Description

Calling `set()` on a `p5.Graphics` buffer is considerably slower in p5.js 2.3.0 than in 1.11.12. Note: directly modifying `pixels[]` performs about the same in both versions.

I first noticed the issue when trying to run the following sketch in p5.js 2.3.0: https://openprocessing.org/@megamitts/2979162

### Benchmark

I made a quickly vibe coded [benchmark sketch](https://editor.p5js.org/SableRaf/sketches/LeRg5yh7C) to systematically test the issue. It updates a 100 × 100 `p5.Graphics` buffer over multiple frames and prints the averaged results to the console.

In this test in Chrome, `set()` was about 6.9 times slower in 2.3.0 than in 1.11.12. 

| p5.js version | `set()` | `pixels[]` | `set()` slowdown vs `pixels[]` |
| --- | --- | --- | --- |
| 1.11.12 | 1.85 ms | 1.10 ms | 1.7× |
| 2.3.0 | 12.79 ms | 1.16 ms | 11× |

I got comparable results, with slowdowns between ~9× and ~17×, in Safari and Firefox.

### Reproduction

Run the [benchmark](https://editor.p5js.org/SableRaf/sketches/LeRg5yh7C) sketch in p5.js 1.11.x, then in 2.x

NalinDalal:
Resolves #9000 

`set()` on `p5.Graphics` was significantly slower in 2.x vs 1.11.x because
`_getRGBA` did a full `to('srgb')` + `structuredClone` + `map()` on every
call, even when the color was already in RGB mode and being reused in a
tight loop.

 Changes:
1. **Cache `_getRGBA` results**
   - Added `#rgbaCache` Map on `Color`, keyed by `maxes` string
   - First call computes and stores; subsequent calls with same `maxes`
     return a shallow copy of the cached array
   - Cache is invalidated in `setRed`, `setGreen`, `setBlue`, `setAlpha`,
     and the `_color` setter
2. **RGB-mode fast path**
   - When `this.mode === RGB`, coords are already sRGB 0-1, so skip the
     `to('srgb')` conversion and use `[...this._color.coords, this._color.alpha]`
     directly
   - Non-RGB modes still go through colorjs conversion

 **AI usage disclosure**: I used Kilo (an AI coding assistant) to help
implement and review this change. I understand and take responsibility for
every line of code in this PR. See [AI_USAGE_POLICY.md].

#### PR Checklist

- [X] `npm run lint` passes
- [X] [Inline reference] is included / updated


ksen0:
To show that this fixes the original issue, please report results from the linked benchmark test, at minimum. Thanks!

NalinDalal:
Confirmed the fix in 88e1550a2 — benchmark (100×100 buffer, 50 iterations, Brave):
- Average: 2.41 ms
- Min: 1.90 ms
- This matches the v1.11.12 baseline of ~1.85 ms, down from the reported 12.79 ms in 2.3.0.


Screenshot 2026-07-25 at 12 59 39 PM

Hope this resolves your query @ksen0

src/color/p5.Color.js
```diff
@@ -715,33 +724,45 @@ class Color {
...
+    const result = coords.map((coord, i) => {
       return map(
         coord,
         colorjsMaxes[i][0], colorjsMaxes[i][1],
         rangeMaxes[i][0], rangeMaxes[i][1]
       );
     });
 
-    return coords;
+    this.#rgbaCache.set(cacheKey, result);
```
perminder-17: I don't think we need this Map(). The only heavy step is the 'srgb' conversion you did above, so we can just cache that part, less complexity and cleaner code.

perminder-17 (changes requested):
The results are good, just one concern for the cleaner code.

src/color/p5.Color.js
```diff
@@ -715,33 +724,45 @@ class Color {
     const newval = map(new_alpha, max[0], max[1], colorjsMax[0], colorjsMax[1]);
 
     this._color.alpha = newval;
+    this.#invalidateRGBACache();
   }
 
   _getRGBA(maxes=[1, 1, 1, 1]) {
+    const cacheKey = maxes.join(',');
```
Ayush4958: `maxes.join(',')` allocates a new string on every call inside loops. Using a fast path key check for `[255, 255, 255, 255]` avoids string allocations during pixel loops.

src/color/p5.Color.js
```diff
@@ -715,33 +724,45 @@ class Color {
...
       if(!Array.isArray(v)){
         return [0, v];
       }else{
         return v;
       }
     });
 
-    coords = coords.map((coord, i) => {
+    let coords;
+    if (this.mode === RGB) {
```
Ayush4958, line 739: Use `this._color.space.id === 'srgb'` instead of `this.mode === RGB` makes the fast-path work for any color object backed by sRGB. It ensures that any color whose underlying color space is sRGB will bypass colorjs conversion and structuredClone, even if the color mode was initialized via another path.

NalinDalal:
All feedback are addressed and solved, thanks for your input @Ayush4958 and @perminder-17

ksen0 (approved):
Nice work, thanks ! And thanks to reviewers!

perminder-17 (approved):
Looks good to me!

test/bench/set_benchmark.html
limzykenneth, line 1: This file shouldn't be committed as it does not match the required format of the benchmark harness.

limzykenneth (changes requested):
If I read the changes correctly, is it essentially memoizing the `_getRGBA` function? The implementation should be ok and functional but having a `#invalidateSrgbCache` can be a bit brittle in that we need to make sure that `#invalidateSrgbCache` is called on all possible code paths (including in the future) that changes the color somehow. If it is just some form of memoizing, can we create a memoization whose key is the color coordinates? Similar to what `toString()` uses `${this._color.space.id}-${this._color.coords.join(',')}-${this._color.alpha}-${format}`?

src/color/p5.Color.js
```diff
@@ -718,30 +719,38 @@ class Color {
...
+      } else {
+        srgbCoords = [...to(this._color, 'srgb').coords, this._color.alpha];
+      }
 
-    // Normalize everything to 0,1 or the provided range (map)
-    let coords = structuredClone(to(this._color, 'srgb').coords);
-    coords.push(this._color.alpha);
+      if (srgbCoordsCache.size > 1000) {
+        srgbCoordsCache.delete(srgbCoordsCache.keys().next().value);
+      }
```
limzykenneth: Good to have this. I wonder if 1000 is the right number? What would a typical number of colors `set()` would manipulate and what would be a sensible upper bound without it leaking memory?

limzykenneth (commented):
Looks good for now. If you can do a test with benchmarking to see if it affect performance or not that would be great. After that we can merge.

src/color/p5.Color.js
NalinDalal: Typical sketches use far fewer than 1000 unique colors. However, image processing or complex generative art could exceed this. Each entry is small (~32 bytes), so 1000 entries ≈ 32KB max. I chose 1000 as a conservative default that covers most cases without significant memory impact. Would you prefer a different value or should we make it configurable?

limzykenneth: We can keep it as 1000 for now if the above benchmark still holds.

NalinDalal:
Fresh benchmark results for commit c15b88a (key-based memoization):
=================================================================
  p5.js Color Memoization Benchmark  (commit c15b88a)
  key-based srgbCoordsCache, LRU cap=1000
=================================================================

1. Repeated RGB (100 colors x 1000 iters):
   No cache:   108.7 ms
   With cache: 35.1 ms
   Speedup:    3.1x

2. Repeated HSL (100 colors x 1000 iters):
   No cache:   122.2 ms
   With cache: 35.3 ms
   Speedup:    3.5x

3. Cache eviction (1100 unique RGB x 10 iters):
   Time:       12.5 ms
   Cache size: 1001 (cap=1000)

4. Mixed RGB+HSL (200 colors x 50 iters):
   No cache:   14.6 ms
   With cache: 3.3 ms
   Speedup:    4.4x

=================================================================

The benchmark uses the same colorjs.io to() conversion and structuredClone as the real _getRGBA() code path. Memoization gives 3-4x speedup for repeated colors, and the LRU cap of 1000 works correctly.

limzykenneth:
@NalinDalal I've just tried to validate the performance with @SableRaf sketch in #9000 and found that this PR performs worse than 2.3.2: https://editor.p5js.org/limzykenneth/sketches/a3vi9QIRy the two versions can be commented in and out in `index.html`. Can you double check what is going on here?

NalinDalal:
> @NalinDalal I've just tried to validate the performance with @SableRaf sketch in #9000 and found that this PR performs worse than 2.3.2: https://editor.p5js.org/limzykenneth/sketches/a3vi9QIRy the two versions can be commented in and out in `index.html`. Can you double check what is going on here?

ohh, seems I need to iterate over it again, will update you soon.
thanks you for inputs :-)

NalinDalal:
**Removed**:
- Module-level `srgbCoordsCache` Map
- All `#invalidateRGBACache()` calls and the `#rgbaCache` instance field

**Kept**:
- The RGB fast-path in `_getRGBA()` that skips `to('srgb')` when `this.mode === RGB`
- Non-RGB modes still go through `structuredClone(to(this._color, 'srgb').coords)`

**Why the cache was removed**:
The cache was adding overhead (string key creation, Map lookups, array spread) that outweighed its benefit for the `set()` hot loop. Benchmarking showed:

|Version	|Hz|	Mean per 50 frames|
|--------|----|-------------------------|
|With cache|	~2.8-2.9	|~348-352ms|
|Without cache (final)|	~3.06-3.07	|~325-326ms|

Removing the cache improved performance by ~7-8% for the `p5.Graphics.set()` hot path. The RGB fast-path alone fixes the original regression without the cache overhead.

**All unit test passed**

limzykenneth:
Now the improvements seems very minimal (I can't observe the difference in the example sketch) and could have been fluctuations/inaccuracy in the measurement. I think the core issues need to be re-examined on where the performance hit is and how it can be addressed.

limzykenneth:
Actually I think the problem here is that current head enables FES even when using minified version (I have still yet to finish work on the FES stuff) which causes the slow down because the bottle neck is now on FES calls. If we set `p5.disableFriendlyErrors = true;` then the performance for me mostly matches 1.x.

PR #9032, opened 2026-08-02
Pin GHA dependencies

ulgens:
Thanks to https://github.com/suzuki-shunsuke/pinact

* Follows the recommended pattern from https://github.com/processing/p5.js/issues/8674#issuecomment-4586205003
* Complementary to https://github.com/processing/p5.js/pull/8620

 Changes:
* Pins GHA dependencies

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/.github/workflows/continuous-release.yml
+++ b/.github/workflows/continuous-release.yml
@@ -19,8 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: 22
 
@@ -34,7 +34,7 @@ jobs:
         run: npx pkg-pr-new publish --no-template --json output.json --comment=off
 
       - name: Include PR info in output file
-        uses: actions/github-script@v8
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         env:
           PR_NUMBER: ${{ github.event.pull_request.number }}
         with:
@@ -49,7 +49,7 @@ jobs:
             fs.writeFileSync('output.json', JSON.stringify(output));
 
       - name: Upload output data
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: output.zip
           path: output.json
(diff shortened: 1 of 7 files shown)


.github/workflows/continuous-release.yml
```diff
@@ -19,8 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
```
Vaivaswat2244, line 22: `actions/checkout` now has two different pins. `continuous-release.yml` gets `11d5960a` (v4.4.0) because it was on the floating `@v4`, while `ci-test`, `ci-lint`, `contributors-png` and `stewards-update` stay at `34e11487` (v4.3.1). Could you bump all five to v4.4.0? Turns future bumps into one find/replace instead of two.

Vaivaswat2244 (commented):
lgtm. A small change recommended.

.github/workflows/continuous-release.yml
ulgens: Resolved.

Vaivaswat2244 (commented):
Hi @ulgens, looks good. I verified the hashes independently, all 12 pinned actions across the repo resolve to the commits their version comments claim, and no unpinned refs remain.

We're streamlining the CI work in flight, so I'd like to wait for https://github.com/processing/p5.js/pull/9020 to land and re-pin on top of it. Fair warning that it's still Draft, so if it stalls let's revisit rather than sit on this.

Tracking under https://github.com/processing/p5.js/issues/9033.

ulgens:
@Vaivaswat2244 Thanks for the response.

#9020 is a huge PR with hundreds of changed files and it looks like it will take time. We had a similar case with node-20 updates before, and handling those dependencies took months. My recommendation here would be to handle smaller PRs first, and then build the work around what is left - otherwise the process gets stuck.

Also, #9020 doesn't change the GHA dependency versions or how they are pinned and it doesn't look like an actual blocker.

Vaivaswat2244:
That's fair, and it isn't a technical blocker. You're right about that. My reason for holding is rebase direction: if this lands first, 9020 has to absorb the pin changes; if it waits, the rebase falls on this PR instead, which is a single pinact run. Cheaper on this side either way. 

Point taken on the node-20 precedent though. We're actively pushing to get 9020 merged quickly for exactly this reason — the longer it sits, the more crossover it creates with everything else in the CI area.

ulgens:
#9020 's replacement #9062 is merged now and I rebased the PR. Should be ready to merge.

Vaivaswat2244 (approved):
lgtm. Thanks a lot @ulgens for your efforts in this direction. @limzykenneth @ksen0

PR #9034, opened 2026-08-03
ci: bump contributors-png workflow to Node 22

Vaivaswat2244:
`contributors-png.yml` was the only workflow still on `node-version: 20` —
missed when #8852 moved the rest to 22. Node 20 reached EOL in April, so this
brings it in line.

Kept at 22 to match every other workflow on `main` rather than jumping to 24,
since the step below runs `npm install canvas` (a native module) and whether
`main` moves to 24 is a separate call.

Verified on my fork: `npm install canvas` and `node utils/contributors-png.js`
run cleanly on Node 22.

### DIFF
--- a/.github/workflows/contributors-png.yml
+++ b/.github/workflows/contributors-png.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
-          node-version: 20
+          node-version: 22
 
       - name: Install dependencies
         run: npm install canvas


limzykenneth:
This is fine and I would actually say go with node 24 if desired. The main CI tasks will be upgraded to node 24 once the dependencies stuff are done anyway. However, there are too many parallel things going on at the moment with the CI tasks, I would like them to be more centralized or it can be hard to keep track of things and we need to solve merge conflicts too often. @ksen0

ksen0:
Sorry for the extra noise. @Vaivaswat2244 the png task also fails for an unrelated (I think) reason: https://github.com/processing/p5.js/actions/runs/31055338747/job/92471501994

Also this workflow was pulled out into an action here https://github.com/processing/all-contributors-image/pulls that should be reused in processing to. I have unfortunately not been able to review. If you have time and interest, if you could help more the contributors image stuff forward, that would be helpful both for p5.js and processing. But no worries if not!

Vaivaswat2244:
> Sorry for the extra noise. @Vaivaswat2244 the png task also fails for an unrelated (I think) reason: https://github.com/processing/p5.js/actions/runs/31055338747/job/92471501994
> 
> Also this workflow was pulled out into an action here https://github.com/processing/all-contributors-image/pulls that should be reused in processing to. I have unfortunately not been able to review. If you have time and interest, if you could help more the contributors image stuff forward, that would be helpful both for p5.js and processing. But no worries if not!

Sure, I will gladly do that. Was trying to invest more time here anyway :) 

> This is fine and I would actually say go with node 24 if desired. The main CI tasks will be upgraded to Node 24 once the dependencies stuff are done anyway. However, there are too many parallel things going on at the moment with the CI tasks, I would like them to be more centralized or it can be hard to keep track of things, and we need to solve merge conflicts too often. @ksen0

Yes, there are mainly three threads right now. (#9020 build/deps, #9032 pinning, #8674 Node 24)

Probable Sequence: #9020 churns the most (build tooling + test runner), so it's the long pole. Smaller CI PRs wait rather than landing alongside or rebase onto it, including the Node 24 flip, which depends on #9020's Playwright swap anyway.
That ordering alone removes most of the overlap.

**Ownership per area** so two people aren't editing the same file: build/deps →@limzykenneth (#9020), pinning/hardening → @ulgens + me, Node 24 bump → me once #9020 lands. Thoughts on this plan?

PR #9051, opened 2026-08-07
fix: keep mouseIsPressed true until all touch pointers are released (fixes #9013)

### LINKED ISSUE #9013: [p5.js 2.0+ Bug Report]: mouseIsPressed is unconditionally reset on multi-touch release

- [ ] Accessibility
- [ ] Color
- [ ] Data
- [ ] DOM
- [x] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors

### p5.js version

2.x (main branch)

### Web browser and version

chrome://150

### Operating system

Windows 11

### Steps to reproduce this

##Description

While testing multi-touch events, I noticed that mouseIsPressed is becoming false even when another finger is still touching the screen.

The expected behavior is that mouseIsPressed should stay true until all active touch points are released. Releasing one finger should not reset it if there are still other active pointers.

##Steps to reproduce

Touch the screen with one finger.
mouseIsPressed becomes true.
Touch the screen with a second finger.
Now there are two active pointers.
Release the second finger while keeping the first finger pressed.
mouseIsPressed becomes false, even though one finger is still touching the screen.

##Expected behavior

mouseIsPressed should only become false when the last active pointer is released.

##Actual behavior

mouseIsPressed becomes false whenever any pointer is released, even if other pointers are still active.

Possible cause

In src/events/pointer.js, the _onpointerup function sets:

this.mouseIsPressed = false;

without checking if there are any remaining active pointers.

The _onpointercancel handler already handles this correctly by checking:

if (this._activePointers.size === 0) {
  this.mouseIsPressed = false;
}

The same check should probably be added to _onpointerup as well.

deepshekhardas:
fixes #9013

### DIFF
--- a/src/events/pointer.js
+++ b/src/events/pointer.js
@@ -1381,9 +1381,13 @@ function pointer(p5, fn, lifecycles){
    */
   fn._onpointerup = function(e) {
     let executeDefault;
-    this.mouseIsPressed = false;
 
     this._activePointers.delete(e.pointerId);
+
+    if (this._activePointers.size === 0) {
+      this.mouseIsPressed = false;
+    }
+
     this._setMouseButton(e);
 
     this._updatePointerCoords(e);
--- a/test/unit/events/touch.js
+++ b/test/unit/events/touch.js
@@ -56,6 +56,36 @@ suite('Touch Events', function() {
     });
   });
 
+  suite('p5.prototype._onpointerup', function() {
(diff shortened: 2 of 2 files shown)


ksen0:

Issue #8871, opened 2026-06-03
Switch branches on `p5.js` so that 2.x is the main branch


ksen0:
[Please see the parent issue for timeline and context.](https://github.com/processing/p5.js/issues/8870) Please feel free to raise questions, concerns, or overlooked tasks; however, the work will be done by the maintainers, so it is not open to implementation volunteers.

When this task is done, on the `p5.js` repository, the `main` branch will include p5.js version 2 latest (currently `dev-2.0`), and a `v1` branch will contain version 1 latest (currently `main`). This task does not affect the website or the editor, please see parent issue for more info.

Steps we will take:
* Merge prerequisite PR that:
  * updates contributor docs to refer to correct branch
  * updates all release workflows to refer to the correct branch
  * Note: in both cases, links refer to current website properties (beta.p5js is 2.x and p5js is 1.x)
* Switch default branch form main to dev-2.0
* Rename main to v1
* Rename dev-2.0 to main
* ~~Release patches in both branches, resolve any issues if they arise~~ Not strictly necessary, no substantive updates to contribtior docs were made

After this is completed, PRs will likely need to be rebased.

ksen0:
Note after Discord question: if by the time the switch happens there is only minor bugfixes and documentation updates in a branch, a patch will be released right after the switch. If there is more substantial changes on that branch, then the normal 1 week of RC patch release process can be followed. Leading up to the branches switch, neither `dev-2.0` nor `main` are frozen.

ksen0:
Here's some helpful guidance from GH on how to update your local clone:

The default branch has been renamed!
dev-2.0 is now named main

If you have a local clone, you can update it by running the following commands.

```sh
git branch -m main v1
git branch -m dev-2.0 main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

nbogie:
Hi @ksen0 I think these instructions are maybe  missing a first step to rename local `main` to be `v1` before trying to rename local `dev-2.0` as `main`.

ksen0:
Thanks @nbogie updated my comment

PR #9055, opened 2026-08-08
Fix Generate Contributors PNG workflow: convert script to ESM

Vaivaswat2244:
## Changes

The `Generate Contributors PNG` workflow fails on every run with `ReferenceError: require is not defined in ES module scope`. `utils/contributors-png.js` is CommonJS, but `package.json` has `"type": "module"`, so Node resolves it as ESM. It fails on every Node version — the script was added a month after the repo switched to ESM and has never worked, which is why `contributors.png` has only ever been updated by hand.

- Renamed `utils/contributors-png.js` → `utils/contributors-png.mjs` and converted it to ESM (`import` instead of `require`, top-level `await` instead of the async IIFE). Matches `utils/`, where every other script is already `.mjs`.
- Updated the `run:` step in `.github/workflows/contributors-png.yml`.

No behaviour change beyond the module format.

The workflow can't run on a PR (`on: push` filtered to `.all-contributorsrc`, gated on `refs/heads/main`), so tested on a fork: <link to successful run>. This is a stopgap — the script is meant to be replaced by processing/all-contributors-image#4.

#### PR Checklist

- [x] `npm run lint` passes

### DIFF
--- a/.github/workflows/contributors-png.yml
+++ b/.github/workflows/contributors-png.yml
@@ -26,7 +26,7 @@ jobs:
         run: npm install canvas
 
       - name: Run contributors-png generator
-        run: node utils/contributors-png.js
+        run: node utils/contributors-png.mjs
 
       - name: Reset all changes except contributors.png
         run: |
--- a/utils/contributors-png.mjs
+++ b/utils/contributors-png.mjs
@@ -1,0 +1,54 @@
+import { createCanvas, loadImage } from 'canvas';
+import fs from 'fs';
+
+const data = fs.readFileSync('.all-contributorsrc', 'utf-8');
+const parsed = JSON.parse(data);
+const contributors = parsed.contributors;
+
+const AVATAR_SIZE = 50;
+const GAP = 4;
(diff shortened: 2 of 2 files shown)


skyyash:
@Vaivaswat2244 thank you for spotting this one, good work!

traced this back to the branch switch(#8871), when 2.x became `main`. the workflow had previously been running against the `v1` branch, where `contributors-png.js` was treated as cjs. on the new `main`, `package.json` has `"type": "module"`, so node treats the `.js` generator as esm and `require()` fails.

tldr; the script worked fine before the default branch switched to v2.

also tested the fix locally, @ksen0 we can merge this one!
image

skyyash (approved):

Vaivaswat2244:
Thanks for the confirmation @skyash-dev . Though this is still a stopgap and this should be perma fixed with https://github.com/processing/all-contributors-image/pull/4

I have left some reviews there, which should help the pr converge to approval soon. Thanks again

ksen0:
Thanks all!

PR #9057, opened 2026-08-09
fix: throw on non-positive saveGif duration

deepshekhardas:
Fixes #8710

When \duration\ is <= 0, the frame-count calculation is invalid. Throw a TypeError like the existing non-number check.


ksen0:

PR #9066, opened 2026-08-09
GSOC 26: specular, ambient, and shininess texture maps

Nixxx19:
## what

extends multi-material rendering to three more mtl texture maps, following the same per-part pattern `map_Kd` already uses: specular (`map_Ks`), ambient (`map_Ka`), and shininess (`map_Ns`). each is parsed, loaded, carried on the part state in p5's vocabulary, and bound in the phong shader, gated so untextured draws are unaffected.

## the pattern (same for all three)

1. parser stores the path (`map_Ks`/`map_Ka` already parsed; added `map_Ns`)
2. `loadMaterialTextures` loads it onto the material (now driven by a small map table)
3. `mtlToPartState` copies it to the part state, defaulting the base colour to white (or shininess to 1) so a map with no matching `Ks`/`Ka`/`Ns` still shows
4. `_applyPartState` binds it into renderer state
5. `_setFillUniforms` always binds the sampler, with a bool that gates its use
6. `phong.frag` samples it and modulates the matching term

## per map

- **`map_Ks`** modulates the specular colour (`inputs.specularMaterial`)
- **`map_Ka`** modulates the ambient colour (`inputs.ambientMaterial`)
- **`map_Ns`** scales the shininess by the map's red channel (`inputs.shininess`)

## zero regression

each new sampler is always bound (to an empty texture when absent) and gated by a `uHas*Tex` bool that defaults false, so existing 3D draws render exactly as before. confirmed by the full visual + unit suite staying green.

## testing

- parser + `mtlToPartState` unit tests for each map (path parsed, lands on part state, sensible base defaults)
- full unit + visual suite green (no regression from the shader changes)

part of the gsoc multi-material .mtl work, follows #8879, #8955, and #9063.


davepagurek:
Can we do some testing to make sure that in scenes with a large number of objects with lighting turned on but no additional textures used (basically: things that use old p5 features and not this) that there is not a performance impact?

I think there is generally some impact from shader complexity so I'd like to have a clearer understanding of what that impact would be in case we need to actually switch shaders to a separate variant that uses more textures if they get used rather than branching in the same shader.

Nixxx19:
perf update. went with the separate shader variant instead of branching in the one shader. all the map sampling + uniforms are behind a #define now, so a plain lit material gets the original phong shader with none of it, and i skip binding the map uniforms on those draws.

tested the exact scenario, 512 lit spheres with specular material and no textures, on a real gpu with vsync disabled vs main built the same way. main averaged ~47.2 fps and this branch ~47.0 fps across runs, which overlap so its noise (about +/-0.6%). before the split it was around 5% slower, so the variant brings it back to no measurable impact.

